PR #14, opened 2024-04-29
[AIE][CI] Accelerate CI

stephenneuendorffer:
For the AIE CI, we will also test upstream architectures independently, so only aggressively test AIE in this workflow.
This patch adds some convenience targets as well:
check-llvm-aie
check-clang-aie
check-lld-aie


llvm/test/CMakeLists.txt
```diff
@@ -254,3 +254,17 @@ add_lit_testsuites(LLVM ${CMAKE_CURRENT_SOURCE_DIR}
...
 add_dependencies(check check-all)
 set_target_properties(check PROPERTIES FOLDER "Tests")
+
+
+add_custom_target(check-llvm-aie)
+add_dependencies(check-llvm-aie 
+  check-llvm-codegen-aie
+  check-llvm-codegen-mir-aie 
+  check-llvm-mc-aie
+  check-llvm-object-aie
```
makslevental, line 264: are these somehow automatically generated CMake targets? I don't see them anywhere (or any substring) in the code base?

makslevental: Hmm I guess they are https://github.com/Xilinx/llvm-aie/blob/f9f8ddd9b9a9799f765d3a56e09c5a55d699b6e6/.github/workflows/spirv-tests.yml#L26

which doesn't appear anywhere either but presumably is a valid target.

stephenneuendorffer: Yes, cmake generates a directory for each target.  There are 2 add_lit_testsuite() commands, one which only does the local stuff and one which generates the subdirectory targets.

makslevental:
A dumb/n00b question: why are we testing clang? Have there been patches there? Maybe to support pragmas or something?

stephenneuendorffer:
> A dumb/n00b question: why are we testing clang? Have there been patches there? Maybe to support pragmas or something?

There are clang patches to support all of the builtins and lower to LLVM IR intrinsics.

PR #17, opened 2024-04-30
[WIP] wheel distro

makslevental:


llvm/utils/TableGen/CodeGenFormat.cpp
```diff
@@ -416,11 +416,10 @@ void TGInstrLayout::addAlternateInstInMultiSlotPseudo(
...
       if (Inst.InstrName == AltInst->getName()) {
-
-        const std::vector<TGFieldLayout *> SlotsVec(Inst.slots().begin(),
-                                                    Inst.slots().end());
-        assert(SlotsVec.size() == 1 && "Real Instr should have only one slot");
-        for (auto &SlotField : SlotsVec) {
+        auto beg = Inst.slots().begin();
+        assert(beg != Inst.slots().end() && ++beg == Inst.slots().end() &&
+               "Real Instr should have only one slot");
+        for (const auto &SlotField : Inst.slots()) {
```
makslevental: @konstantinschwarz PTAL

llvm/utils/TableGen/CodeGenFormat.cpp
```diff
@@ -693,17 +692,20 @@ void TGInstrLayout::emitFormat(ConstTable &FieldsHierarchy, ConstTable &o,
...
   const std::string TargetClassName = Target + SlotsRegistry.GenSlotKindName;
 
-  for (auto &SlotField : SlotsVec) {
+  auto end = slots().end();
+  bool firstIter = true;
+  for (const auto &SlotField : slots()) {
+    if (!firstIter)
+      o << ", ";
+    else
+      firstIter = false;
```
makslevental: @konstantinschwarz PTAL

llvm/utils/TableGen/CodeGenFormat.cpp
```diff
@@ -741,10 +743,9 @@ void TGInstrLayout::emitFormat(ConstTable &FieldsHierarchy, ConstTable &o,
 
 void TGInstrLayout::emitPacketEntry(ConstTable &Packets,
                                     ConstTable &SlotData) const {
-  const std::vector<TGFieldLayout *> SlotsVec(slots().begin(), slots().end());
   // Some instructions (DUMMY96, UNKNOWN128...) are indicated as Composite but
   // they don't define a Packet Format... In that case, we don't emit anything.
-  if (SlotsVec.empty())
+  if (slots().begin() == slots().end())
```
makslevental: @konstantinschwarz PTAL

llvm/utils/TableGen/CodeGenFormat.cpp
```diff
@@ -754,7 +755,7 @@ void TGInstrLayout::emitPacketEntry(ConstTable &Packets,
   const std::string TargetSlotKindName = Target + SlotsRegistry.GenSlotKindName;
 
   SlotData << "// " << getInstrName() << " : " << SlotData.mark() << "\n";
-  for (auto &Slot : SlotsVec) {
+  for (auto Slot : slots()) {
```
makslevental: @konstantinschwarz PTAL

llvm/utils/TableGen/CodeGenFormat.cpp
makslevental: without these changes this file fails to compile on (whatever version of clang macos13 is on): https://gist.github.com/makslevental/bb5b83755f9397bcb64a137b651445d6

makslevental: Also `std::iterator` is [deprecated in C++17](https://en.cppreference.com/w/cpp/iterator/iterator) so all of these should be rewritten to use llvm ranges:

```c++
  /// Returns an iterator range ranging over all the encoding fields of the
  /// tree. The traversal of the tree is made in a pre-order way.
  iterator_range<TGFieldIterator> fields() const;
  /// Returns a leaf field iterator range (ranging only over the different
  /// leaves of the tree).
  iterator_range<TGFieldIterator> leaves() const;
  /// Returns a slot iterator range (ranging over the fields representing a
  /// slot only).
  iterator_range<TGFieldIterator> slots() const;
  /// Returns an iterator range, ranging on every field defining a MCOperand
  iterator_range<TGFieldIterator> operands() const;
```

PR #20, opened 2024-05-01
[CI] Enable code formatting check (clang-format) on PRs

konstantinschwarz:


.github/workflows/pr-code-format.yml
```diff
@@ -6,7 +6,7 @@ permissions:
 jobs:
```
konstantinschwarz, line 6: The
```
on: pull_request_target
```
makes this action run in the context of the base branch, where the updated workflow is not visible and skipped.
Not sure how to test this before merging

PR #22, opened 2024-05-02
Fix C++17 std::iterator deprecation warning in CodeGenFormat

makslevental:
I believe this is the "definitive" fix.

### DIFF
--- a/llvm/utils/TableGen/CodeGenFormat.cpp
+++ b/llvm/utils/TableGen/CodeGenFormat.cpp
@@ -416,11 +416,10 @@ void TGInstrLayout::addAlternateInstInMultiSlotPseudo(
     bool Found = false;
     for (const TGInstrLayout &Inst : InstFormats) {
       if (Inst.InstrName == AltInst->getName()) {
-
-        const std::vector<TGFieldLayout *> SlotsVec(Inst.slots().begin(),
-                                                    Inst.slots().end());
-        assert(SlotsVec.size() == 1 && "Real Instr should have only one slot");
-        for (auto &SlotField : SlotsVec) {
+        auto beg = Inst.slots().begin();
+        assert(beg != Inst.slots().end() && ++beg == Inst.slots().end() &&
+               "Real Instr should have only one slot");
+        for (const auto &SlotField : Inst.slots()) {
           const TGTargetSlot *Slot = SlotField->SlotClass;
           if (Slot) {
             // Add alternate instr only if no other instr with same slot exist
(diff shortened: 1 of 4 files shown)


llvm/utils/TableGen/CodeGenFormat.cpp
```diff
@@ -416,11 +416,10 @@ void TGInstrLayout::addAlternateInstInMultiSlotPseudo(
     bool Found = false;
     for (const TGInstrLayout &Inst : InstFormats) {
       if (Inst.InstrName == AltInst->getName()) {
-
-        const std::vector<TGFieldLayout *> SlotsVec(Inst.slots().begin(),
-                                                    Inst.slots().end());
-        assert(SlotsVec.size() == 1 && "Real Instr should have only one slot");
-        for (auto &SlotField : SlotsVec) {
+        auto beg = Inst.slots().begin();
+        assert(beg != Inst.slots().end() && ++beg == Inst.slots().end() &&
```
stephenneuendorffer: This might be better: https://llvm.org/doxygen/STLExtras_8h_source.html#l00322
hasOneElement(Inst.slots())

llvm/utils/TableGen/CodeGenFormat.cpp
```diff
@@ -693,17 +692,20 @@ void TGInstrLayout::emitFormat(ConstTable &FieldsHierarchy, ConstTable &o,
     << "/* Slots - Fields mapper */\n"
     << "      {";
 
-  const std::vector<TGFieldLayout *> SlotsVec(slots().begin(), slots().end());
   const std::string TargetClassName = Target + SlotsRegistry.GenSlotKindName;
 
-  for (auto &SlotField : SlotsVec) {
+  auto end = slots().end();
```
stephenneuendorffer: unused?

stephenneuendorffer (approved):

llvm/utils/TableGen/CodeGenFormat.cpp
makslevental:
```suggestion
```

gbossu (approved):

stephenneuendorffer (approved):

llvm/utils/TableGen/CodeGenFormat.cpp
```diff
@@ -689,21 +688,22 @@ void TGInstrLayout::emitFormat(ConstTable &FieldsHierarchy, ConstTable &o,
...
-    << "      "
-    << "/* Slots - Fields mapper */\n"
+    << "      " << "/* Slots - Fields mapper */\n"
     << "      {";
 
-  const std::vector<TGFieldLayout *> SlotsVec(slots().begin(), slots().end());
   const std::string TargetClassName = Target + SlotsRegistry.GenSlotKindName;
 
-  for (auto &SlotField : SlotsVec) {
+  bool firstIter = true;
```
keryell, line 696: I have the feeling that this is reimplementing Python/C++ `join()` with various versions is already in `llvm/include/llvm/ADT/StringExtras.h` around
```c++
/// Joins the strings in the range [R.begin(), R.end()), adding Separator
/// between the elements.
template <typename Range>
inline std::string join(Range &&R, StringRef Separator) {
```

makslevental: You're right but as this PR addresses, this `slots()` thing isn't an llvm range and doesn't successfully convert on all platforms.

PR #24, opened 2024-05-02
Loop aware scheduling

martien-de-jong:
Reopened from the rc5 one.


martien-de-jong:
CI doesn't cooperate on a fork

PR #26, opened 2024-05-03
[AIE2] Run WAW simplification DAG mutator for software pipelining

abhinay-anubola:
Add WAW DAG mutator for SWP


gbossu:
Were you able to verify QoR?

gbossu (approved):

PR #30, opened 2024-05-06
Interactive scheduler / scheduler analyzer

gbossu:
Re-opened from private repo. Already reviewed.

This is a TUI tool that allows to replay the different scheduling actions back and forth for each region. The goal is to help us understand why we end up with a certain schedule and speed up investigations. I'd be happy to get some feedback. :)

This works by parsing the --debug-only=machine-scheduler logs, as I didn't want any invasive changes to MachineScheduler at this point.

Run like this `llc -O2 -mtriple=aie2 test.ll -o - --debug-only=machine-scheduler 2>&1 | python3 llvm/utils/imisched.py -`

This requires `textual`, which can be installed with pip install textual.

image
image

### DIFF
--- a/llvm/test/CodeGen/AIE/aie2/end-to-end/Conv2D-red.ll
+++ b/llvm/test/CodeGen/AIE/aie2/end-to-end/Conv2D-red.ll
@@ -6,6 +6,9 @@
 ;
 ; (c) Copyright 2023-2024 Advanced Micro Devices, Inc. or its affiliates
 ; RUN: llc -O2 -mtriple=aie2 --enable-pipeliner=0 %s -o - | FileCheck %s --check-prefix=ASM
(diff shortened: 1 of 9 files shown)


ValentijnvdBeek (approved):
LGTM

maxodendahl:
@gbossu maybe interesting in this context?

https://github.com/shioyadan/Konata

gbossu:
> @gbossu maybe interesting in this context?
> 
> https://github.com/shioyadan/Konata

That does look nice! My initial goal was to be able to visualize the scheduling decisions so we can quickly debugs cases like "Why do I now get one more cycle in this loop?". But I think if we could get a tool like `Konata` to be a bit more interactive we could get the best of both: debugging scheduling decisions, and maybe spotting missed opportunities.

PR #31, opened 2024-05-07
AIECC-616 Fix "clang -g" failure

gbossu:
Compiling with -g shows a failure in the "Debug Variable Analysis" pass. This is due to this pass not being preserved by the `AIESuperRegRewriter` pass, which runs in our custom RA pipeline for allocation 2D and 3D registers.


llvm/lib/Target/AIE/AIESuperRegRewriter.cpp
```diff
@@ -49,6 +50,8 @@ class AIESuperRegRewriter : public MachineFunctionPass {
     AU.addPreserved<VirtRegMap>();
     AU.addRequired<SlotIndexes>();
     AU.addPreserved<SlotIndexes>();
+    AU.addRequired<LiveDebugVariables>();
+    AU.addPreserved<LiveDebugVariables>();
```
martien-de-jong, line 54: This means we previously implicitly invalidated it; how is that wrong?

llvm/test/CodeGen/AIE/aie2/ra/debug_value_2d_reg.mir
```diff
@@ -0,0 +1,139 @@
...
+# RUN: llc --start-before=greedy --stop-after=virtregrewriter --mtriple=aie2 -o - %s | FileCheck %s
+
+# Verify that the custom RA pipeline handles DBG_VALUE instructions.
+# The MIR is obtained by compiling the snippet below and stopping before greedy.
+#    void *padd_2d(void *a, int off, int size1, addr_t &count1, int inc1) {
+#      return add_2d_byte(a, off, size1, count1, inc1);
+#    }
+
+# Note that the connection to source locations is already lost after ISel, where
+# the four input parameters are combined into a single "ed" register.
```
martien-de-jong, line 21: Could you add a comment describing the main expectation of this test?

llvm/test/CodeGen/AIE/aie2/ra/debug_value_simple.mir
```diff
@@ -0,0 +1,66 @@
...
+# This file is licensed under the Apache License v2.0 with LLVM Exceptions.
+# See https://llvm.org/LICENSE.txt for license information.
+# SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+#
+# (c) Copyright 2024 Advanced Micro Devices, Inc. or its affiliates
+
+# RUN: llc --start-before=greedy --stop-after=virtregrewriter --mtriple=aie2 -o - %s | FileCheck %s
+
+# A simpler version of debug_value_2d_reg.mir to verify DBG instructions are
+# handled grafully by the custom RA pipeline.
```
martien-de-jong: nit: gracefully

martien-de-jong (approved):

llvm/lib/Target/AIE/AIESuperRegRewriter.cpp
gbossu: Greedy expects to find Debug info for the live ranges it has to allocate. If it meets a DBG_VALUE instruction that needs to be updated after live range splitting, it will have to also update the `LiveDebugVariables` analysis. Which is problematic if the analysis has not been preserved.

llvm/test/CodeGen/AIE/aie2/ra/debug_value_2d_reg.mir
gbossu: I think the main point is really to verify we do not crash. DBG instructions have mostly been ignored at this point, this is really just an initial test to see how they are handled by our RA.

andcarminati (approved):
LGTM.

PR #32, opened 2024-05-09
fixup! llvm/CodeGen: Add target hook to CallLowering to pre-lower return type

konstantinschwarz:


llvm/lib/CodeGen/GlobalISel/IRTranslator.cpp
```diff
@@ -253,11 +253,11 @@ ArrayRef<Register> IRTranslator::getOrCreateVRegs(const Value &Val) {
   return *VRegs;
 }
```
stephenneuendorffer, line 255: Needs documentation?

stephenneuendorffer (dismissed):

llvm/lib/CodeGen/GlobalISel/IRTranslator.cpp
gbossu: I remember I had outlined that part of the code into a new `getOrCreateReturnVRegs()` to avoid duplication. But what (I guess) I didn't realize back then is that `RetVal` has a larger scope than I thought, and changes to it also need to be reflected outside `getOrCreateReturnVRegs`.

In the end, I think `getOrCreateReturnVRegs` is/was a poor outlining choice, and it makes the code even more confusing given how `RetVal` is meant to be used. I'd now vouch for something where `Ret`/`RetVal` has a single point of definition, e.g.
```
const Value *getReturnValueForABI(const ReturnInst &RI) {
  const Value *RetVal = RI.getReturnValue();
  if (RetVal && DL->getTypeStoreSize(RetVal->getType()).isZero())
    return nullptr;
  return RetVal;
}

...
const Value *Ret = getReturnValueForABI(RI);
ArrayRef<Register> VRegs = getOrCreateReturnVRegs(Ret);
...
```

gbossu (approved):

PR #36, opened 2024-05-14
[AIE2] Add missing aie2_v8accfloat. Add missing undef vector tests.

SagarMaheshwari99:


clang/test/CodeGen/aie/aie2/aie2-undef-vectors.cpp
```diff
@@ -0,0 +1,328 @@
+// NOTE: Assertions have been autogenerated by utils/update_cc_test_checks.py
+//===- aie2-undef-vectors.cpp ------------------------------------*- C++ -*-===//
+//
+// This file is licensed under the Apache License v2.0 with LLVM Exceptions.
+// See https://llvm.org/LICENSE.txt for license information.
+// SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+//
+// (c) Copyright 2023-2024 Advanced Micro Devices, Inc. or its affiliates
```
gbossu: I believe this should be `(c) Copyright 2024`

clang/test/CodeGen/aie/aie2/aie2-undef-vectors.cpp
```diff
@@ -0,0 +1,328 @@
...
+// CHECK-LABEL: @_Z19test_undef_v32uint4v(
+// CHECK-NEXT:  entry:
+// CHECK-NEXT:    [[TMP0:%.*]] = tail call <4 x i32> @llvm.aie2.v4int32()
+// CHECK-NEXT:    [[TMP1:%.*]] = bitcast <4 x i32> [[TMP0]] to <16 x i8>
+// CHECK-NEXT:    ret <16 x i8> [[TMP1]]
+//
+v32uint4 test_undef_v32uint4() { return undef_v32uint4(); }
+// CHECK-LABEL: @_Z18test_undef_v16int8v(
+// CHECK-NEXT:  entry:
+// CHECK-NEXT:    [[TMP0:%.*]] = tail call noundef <16 x i8> @llvm.aie2.v16int8()
```
gbossu, line 29: This looks a bit surprising, for signed elements we use `llvm.aie2.v4int32`, but for unsigned we use `llvm.aie2.v16int8` directly and need no bitcast?

llvm/lib/Target/AIE/AIE2InstructionSelector.cpp
```diff
@@ -565,6 +565,7 @@ bool AIE2InstructionSelector::select(MachineInstr &I) {
     case Intrinsic::aie2_v16bfloat16:
     case Intrinsic::aie2_v32bfloat16:
     case Intrinsic::aie2_v64bfloat16:
+    case Intrinsic::aie2_v8accfloat:
```
gbossu, line 568: Nit: I would have preferred to see a first NFC commit where you move all the existing tests, and then a second one where you add support for `v8accfloat`, otherwise it is really hard to spot the new feature.

llvm/test/CodeGen/AIE/aie2/undef_vectors.ll
```diff
@@ -0,0 +1,522 @@
+; NOTE: Assertions have been autogenerated by utils/update_llc_test_checks.py
+;
+; This file is licensed under the Apache License v2.0 with LLVM Exceptions.
+; See https://llvm.org/LICENSE.txt for license information.
+; SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+;
+; (c) Copyright 2023-2024 Advanced Micro Devices, Inc. or its affiliates
```
gbossu: Same: `(c) Copyright 2024` seems more appropriate

llvm/lib/Target/AIE/AIE2InstructionSelector.cpp
gbossu, line 568: Could you add an ISel test as well?

gbossu:
Curious: IIRC we want to eventually eliminate all those `__builtin_aiev2_v4int32` (and friends) builtins with more generic clang builtins, is there any progress on that front?

gbossu (approved):

clang/test/CodeGen/aie/aie2/aie2-undef-vectors.cpp
SagarMaheshwari99: No, these CHECK lines are for the test below on line 32. AFAIK, CHECK lines from update_cc_test_checks.py are generated above the tests.

SagarMaheshwari99:
> Curious: IIRC we want to eventually eliminate all those `__builtin_aiev2_v4int32` (and friends) builtins with more generic clang builtins, is there any progress on that front?

I will check but I think the story was just to replace clang vector builtins for vector concat, insert, update and set with generic ones.

PR #37, opened 2024-05-17
Fix loopaware

martien-de-jong:
Two problems were solved:
A- In iterative scheduling, we need to construct the dag from the original semantical order. Negative latency scheduling in non-converged schedules will break dependency calculation. There's no convenient way to prevent reordering.
B- loop-aware scheduling should add appropriate ExitSU edge latencies. This is essential when the loop body is too short, and the dependences over the backedge have a distance of greater than one.


llvm/lib/CodeGen/MachineScheduler.cpp
```diff
@@ -875,10 +875,11 @@ void ScheduleDAGMI::schedule() {
   LLVM_DEBUG(dbgs() << "ScheduleDAGMI::schedule starting\n");
   LLVM_DEBUG(SchedImpl->dumpPolicy());
 
-  // Build the DAG.
-  buildSchedGraph(AA);
-
-  postProcessDAG();
+  if (SchedImpl->buildGraph(*this)) {
+    // Build the DAG.
```
martien-de-jong: Hmmm. This evolved to an awkward interface. Why not just have a virtual buildSchedGraph. Just say the word.

llvm/include/llvm/CodeGen/MachineScheduler.h
```diff
@@ -264,6 +264,11 @@ class MachineSchedStrategy {
   /// Tell the strategy that current MBB is done.
   virtual void leaveMBB() {}
 
+  /// This can override DAG construction and postprocessing by returning false.
+  /// This is useful for iterative scheduling, where the graph is invariant
+  /// over multiple schedule() calls.
```
martien-de-jong, line 269: Note that we still create it for each iteration. There are no nice copy constructors, and we will process multiple regions in interleaved fashion.

llvm/lib/Target/AIE/AIEMaxLatencyFinder.cpp
```diff
@@ -140,8 +140,7 @@ MaxLatencyFinder::MaxLatencyFinder(AIEScheduleDAGMI *DAG)
     : Scheduler(DAG->getSchedImpl()),
```
martien-de-jong: I think this happens to work, but it is wrong. DAG may not be an AIESchedulerDAGMI. We are saved by the fact that we will not derefence Scheduler if it is 'just' a SchedulerDAGInstrs.

llvm/lib/CodeGen/MachineScheduler.cpp
gbossu: Then I guess you'd also need to make `postProcessDAG` virtual? I don't think we can assume all the DAG mutators are idempotent, so running them again will be an issue.

I think `SchedImpl->buildGraph` is fine, I'd just rename it to `shouldBuildSchedGraph` maybe. Or as you say made `buildSchedGraph` virtual but embed `postProcessDAG` in it as well.

llvm/include/llvm/CodeGen/MachineScheduler.h
gbossu: Not sure I understand the comment. If this returns `false`, the construction would be completely skipped, and we'd get the DAG that was used in the previous region? Or is this UB?

llvm/lib/Target/AIE/AIEInterBlockScheduling.cpp
```diff
@@ -81,6 +81,7 @@ void InterBlockScheduling::leaveFunction() {
 
 void InterBlockScheduling::enterBlock(MachineBasicBlock *BB) {
   CurrentBlock = &getBlockState(BB);
+  CurrentBlock->resetRegion();
```
gbossu, line 84: Why wasn't this needed before?

llvm/lib/Target/AIE/AIEMachineScheduler.cpp
```diff
@@ -825,6 +825,23 @@ void AIEScheduleDAGMI::exitRegion() {
   ScheduleDAGMI::exitRegion();
 }
 
+void AIEScheduleDAGMI::placeDebugValues() {
```
gbossu: How does that differ from the default placement? I'm not sure I follow why this is needed

llvm/include/llvm/CodeGen/MachineScheduler.h
martien-de-jong: If it returns false, the scheduler assumes it is built correctly, and skips its own construction. We, AIE, may be constructing it multiple times from scratch in iterative scheduling. We could have stashed it away at some point and copy it here, but saving and restoring is difficult without copy constructors.

llvm/lib/Target/AIE/AIEInterBlockScheduling.cpp
martien-de-jong: Because schedule() now uses the current region to retrieve the semantic order. Previously that was only done before incrementing the current region

llvm/lib/Target/AIE/AIEMachineScheduler.cpp
martien-de-jong: I need to recheck, but this call is interleaved with the rest of dag construction and interprets the original semantic order. I can't immediately say that will be correct if the underlying basic block order changes to a negative latency order.

llvm/lib/Target/AIE/AIEInterBlockScheduling.h
```diff
@@ -123,20 +123,28 @@ class FixedpointState {
 // get enter and leave calls. We construct our Regions from the iterators
 // passed to the enterRegion call.
 class Region {
+  // The instrutions in their original order
```
andcarminati, line 126: Nit: instrutions -> instructions

llvm/lib/Target/AIE/AIEMachineScheduler.cpp
andcarminati: I think it can be a good idea to have a this change covered with a test.

andcarminati: This function can be problematic. See:
```
int foo(int* vector) {
    while (*vector != 0) {
        vector++;
    }
    return *(vector-1);
}
```

Compiled with:

`clang -target aie2 loop.c -c -g -O1`

```
[....]
#18 0x0000000004f1c05a llvm::AIEScheduleDAGMI::placeDebugValues() /scratch/llvm-aie/llvm/lib/Target/AIE/AIEMachineScheduler.cpp:835:16
#19 0x00000000067bbd10 llvm::ScheduleDAGMI::schedule() /scratch/llvm-aie/llvm/lib/CodeGen/MachineScheduler.cpp:924:3
#20 0x00000000067cd814 (anonymous namespace)::MachineSchedulerBase::scheduleRegions(llvm::ScheduleDAGInstrs&, bool) /scratch/llvm-aie/llvm/lib/CodeGen/MachineScheduler.cpp:0:0
#21 0x00000000067ce5e3 llvm::cl::opt_storage<bool, false, false>::getValue() const /scratch/llvm-aie/llvm/include/llvm/Support/CommandLine.h:1413:38
#22 0x00000000067ce5e3 llvm::cl::opt_storage<bool, false, false>::operator bool() const /scratch/llvm-aie/llvm/include/llvm/Support/CommandLine.h:1417:38
#23 0x00000000067ce5e3 (anonymous namespace)::PostMachineScheduler::runOnMachineFunction(llvm::MachineFunction&) /scratch/llvm-aie/llvm/lib/CodeGen/MachineScheduler.cpp:482:7
#24 0x0000000006740150 llvm::MachineFunctionPass::runOnFunction(llvm::Function&) /scratch/llvm-aie/llvm/lib/CodeGen/MachineFunctionPass.cpp:0:13
[....]
```

llvm/include/llvm/CodeGen/MachineScheduler.h
```diff
@@ -426,18 +435,14 @@ class ScheduleDAGMI : public ScheduleDAGInstrs {
...
-
   /// Release ExitSU predecessors and setup scheduler queues.
   void initQueues(ArrayRef<SUnit*> TopRoots, ArrayRef<SUnit*> BotRoots);
 
   /// Update scheduler DAG and queues after scheduling an instruction.
   void updateQueues(SUnit *SU, bool IsTopNode);
 
   /// Reinsert debug_values recorded in ScheduleDAGInstrs::DbgValues.
-  void placeDebugValues();
+  virtual void placeDebugValues();
```
gbossu, line 449: Does this still need to be made virtual? I do not see it being overridden.

llvm/lib/Target/AIE/AIEMachineScheduler.cpp
```diff
@@ -875,3 +900,24 @@ void AIEScheduleDAGMILive::exitRegion() {
   static_cast<AIEPreRASchedStrategy *>(SchedImpl.get())->leaveRegion(ExitSU);
   ScheduleDAGMILive::exitRegion();
 }
+
+bool llvm::AIEPostRASchedStrategy::shouldBuildGraph(ScheduleDAGMI &DAG) {
```
gbossu: (https://github.com/Xilinx/llvm-aie/pull/37/files#r1605060993) Hmmm, then I agree this is a confusing interface :)

I guess we should just have an overridable `buildGraph()` then, and by default it would do this:
```
    buildSchedGraph(AA);
    postProcessDAG();
```

llvm/lib/Target/AIE/AIEMaxLatencyFinder.cpp
martien-de-jong: I put in some dynamic checks. I think it is safe now, but still ugly.

llvm/lib/Target/AIE/AIEMachineScheduler.cpp
martien-de-jong: I changed to mimic recordDbgInstrs using the semantic order rather than do my own placement. I added this example as a test.

llvm/lib/Target/AIE/AIEInterBlockScheduling.cpp
```diff
@@ -427,10 +427,19 @@ void BlockState::clearSchedule() {
 void BlockState::classify() {
   // Detect whether this block is amenable to loop-aware scheduling.
   // We must push the safety margin to our epilogue block(s)
-  // This can only be done if the epilogue is not itself
+  // This can only be done if we have an epilogue and the epilogue is not itself
```
andcarminati, line 431: Hi @martien-de-jong , maybe we can include a test here, a very simple one.

llvm/lib/Target/AIE/AIEMachineScheduler.cpp
```diff
@@ -825,6 +825,31 @@ void AIEScheduleDAGMI::exitRegion() {
   ScheduleDAGMI::exitRegion();
 }
 
+void AIEScheduleDAGMI::recordDbgInstrs(const Region &CurrentRegion) {
```
andcarminati, line 828: I think that for this specific commit, the message is explaining a different change. The message is "[AIE] Override buildGraph to use SmanticOrder in graph building" but we are dealing with Debug instruction housekeeping. Also SmanticOrder -> SemanticOrder.

llvm/lib/Target/AIE/AIEInterBlockScheduling.cpp
```diff
@@ -427,10 +427,19 @@ void BlockState::clearSchedule() {
...
   // a loop.
   auto IsLoop = [](MachineBasicBlock *MBB) {
-    return llvm::any_of(MBB->successors(), [&](auto *S) { return S == MBB; });
+    int NumLoopEdges = 0;
```
andcarminati, line 434: NIT: maybe something more compact:

```
    int NumLoopEdges = 0;
    for (auto *S : MBB->successors())
      NumLoopEdges = (S == MBB) ? NumLoopEdges + 1 : NumLoopEdges;

    return NumLoopEdges == 1 && (MBB->succ_size() - NumLoopEdges) == 1;
```

llvm/lib/Target/AIE/AIEMachineScheduler.cpp
martien-de-jong: There was some commit confusion, which I set right. This debug handling is however part of replacing the standard DAG construction, which is interwoven with it.

llvm/lib/Target/AIE/AIEInterBlockScheduling.cpp
martien-de-jong: Perhaps less lines, but I'd rather not sacrifice readibility.

llvm/lib/CodeGen/MachineScheduler.cpp
```diff
@@ -2237,6 +2236,14 @@ void CopyConstrain::apply(ScheduleDAGInstrs *DAGInstrs) {
 
 static const unsigned InvalidCycle = ~0U;
 
+// Default virtual of MachineSchedStrategy
+void MachineSchedStrategy::buildGraph(ScheduleDAGMI &DAG, AAResults *AA,
```
gbossu, line 2240: Checking: MachineSchedStrategy is a base class for both the pre and post RA strategies.

llvm/lib/Target/AIE/AIEMachineScheduler.cpp
```diff
@@ -825,6 +825,31 @@ void AIEScheduleDAGMI::exitRegion() {
...
+      DbgMI = nullptr;
+    }
+
+    if (Prev->isDebugValue() || Prev->isDebugPHI()) {
+      DbgMI = Prev;
+    }
+  }
+  if (DbgMI)
+    FirstDbgValue = DbgMI;
+}
```
gbossu, line 851: How does this differ from the standard `recordDbgInstrs`?

gbossu (approved):
LGTM, before merging I would just like to understand why we had to write a new `recordDbgInstrs`. It looks pretty similar to the original one, maybe we can share code?

llvm/lib/CodeGen/MachineScheduler.cpp
martien-de-jong: Yes, It is the top most class in the hierarchy.

llvm/lib/Target/AIE/AIEMachineScheduler.cpp
martien-de-jong: This uses the semantic order rather than the current order in the block. As such it uses a different iterator, for the rest it is mainly the same. Sharing code would be possible through a template I guess, but I think that's rather disruptive to the main line code.  For what it's worth, we may want to tweak the debug insertion points in future, although now that I've said it, that will probably not happen

andcarminati (approved):
LGTM.

PR #38, opened 2024-05-17
G_BUILD_VECTOR & G_UNMERGE_VALUES legalization for 32-bit vectors

ValentijnvdBeek:
In this merge request, 32-bit vectors are iterated over and added iteratively to a 32-bit number. It does not support 16-bit build vectors since this type cannot occur within a properly legalized program. The intuition for this is as follows:

1. The legalizer removes code that it considers dead 
2. The legalizer considers code dead when it does a "not safe to move" operation (e.x. move, call or sequential load)
3. Build Vector does fall under those operations, therefore its result must be used somewhere
4. Global ISel uses Single Static Assignment, so variable types cannot change and must be defined exactly once
5. The legalizer works backward from the last instruction to the first instruction
6. Types must be the same when they're defined and used
7. If a 16-bit vector reaches the BV, there is another operation uses it and didn't set its type correctly. If that happens, we cannot fix it in the Build Vector since that would cause a conflict between the two types
8. Ergo such conflicts need to be resolved by legalizing them with the "not safe to move" operations

The opposite logic is done for G_UNMERGE_VALUES that iterates over a 32-bit number, shifts truncates them into separate scalar numbers and finishes once it has extracted all values.

### DIFF
--- a/llvm/lib/Target/AIE/AIELegalizerInfo.cpp
+++ b/llvm/lib/Target/AIE/AIELegalizerInfo.cpp
@@ -491,6 +491,59 @@ bool AIELegalizerInfo::legalizeCustom(LegalizerHelper &Helper,
   llvm_unreachable("Un-expected custom legalization");
 }
 
+bool AIELegalizerInfo::pack32BitVector(LegalizerHelper &Helper,
+                                       MachineInstr &MI,
+                                       Register SourceReg) const {
+  MachineIRBuilder &MIRBuilder = Helper.MIRBuilder;
+  MachineRegisterInfo &MRI = *MIRBuilder.getMRI();
+
+  const LLT SourceRegTy = MRI.getType(SourceReg);
+  const Register DstReg = MI.getOperand(0).getReg();
+  assert(SourceRegTy.getSizeInBits() == 32 &&
+         "cannot pack vectors larger or smaller than 32-bit");
+
+  const LLT S32 = LLT::scalar(32);
+  unsigned Offset = 0;
+  Register DstCastReg = MRI.createGenericVirtualRegister(S32);
+
+  // Skip the destination operand since that is where we are writing to.
+  MachineOperand *Operand = MI.operands_begin() + 1,
+                 *OperandEnd = MI.operands_end();
+  MIRBuilder.buildConstant(DstCastReg, 0);
+
+  const LLT RegTy = MRI.getType(DstReg);
+  while (Operand != OperandEnd) {
+    Register DestinationOperand = Operand->getReg();
+
+    if (RegTy.getScalarSizeInBits() != 32) {
+      const Register TmpReg32 = MRI.createGenericVirtualRegister(S32);
+      MIRBuilder.buildInstr(AIE2::G_ZEXT, {TmpReg32}, {DestinationOperand});
+      DestinationOperand = TmpReg32;
+    }
+
+    // Avoid a useless shift for the first element, since it doesn't get
+    // optimized out in O0.
+    const Register AccumulatorReg = MRI.createGenericVirtualRegister(S32);
+    if (Offset != 0) {
+      const MachineInstrBuilder ShiftConstant =
+          MIRBuilder.buildConstant(S32, Offset);
+      const MachineInstrBuilder Masked =
+          MIRBuilder.buildShl(S32, DestinationOperand, ShiftConstant);
+      MIRBuilder.buildOr(AccumulatorReg, DstCastReg, Masked);
+    } else {
+      MIRBuilder.buildOr(AccumulatorReg, DstCastReg, DestinationOperand);
+    }
+
+    DstCastReg = AccumulatorReg;
+    Offset += RegTy.getScalarSizeInBits();
+    ++Operand;
+  }
+
+  MIRBuilder.buildBitcast(DstReg, DstCastReg);
+  MI.eraseFromParent();
+  return true;
+}
+
 bool AIELegalizerInfo::legalizeG_BUILD_VECTOR(LegalizerHelper &Helper,
                                               MachineInstr &MI) const {
   MachineIRBuilder &MIRBuilder = Helper.MIRBuilder;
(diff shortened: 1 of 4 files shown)


ValentijnvdBeek:
To ease the reviewing process and to help prove correctness of the transformation, I have written property tests for the outputted code. Those are tests where you give an invariant which must hold, the test framework then uses a random/all options to see if the invariant holds. If not, it finds the minimal case where the test fails. The main invariant here is that `unpack(pack(x)) = x`.

Implementation is done using the [rapidcheck](https://github.com/emil-e/rapidcheck) testing framework, you can run it on a local machine if you'd like. 

```C++

#include <cstdint>
#include <rapidcheck.h>
#include <vector>
#include <cstdio>

#define G_TRUNC(val) ((val) & 0xFF)
#define G_ANYEXT(val) (static_cast<uint32_t>(val))

std::vector<uint8_t> unpack_vector(uint32_t source) {
    uint32_t offset = 0;
    uint32_t dstcast = source;
    std::vector<uint8_t> abc;

    for (int i = 0; i < 4; i++) {
        if (offset != 0) {
            uint32_t mask = dstcast >> offset;
            uint32_t cut = G_TRUNC(mask);
            abc.push_back(cut);
        } else {
            uint32_t cut = G_TRUNC(dstcast);
            abc.push_back(cut);
        }
        offset += 8;
    }
    return abc;
}

uint32_t pack_vector(const std::vector<uint8_t> abc) {
    uint32_t offset = 0;
    uint32_t dstcast = 0;
    uint32_t acc = 0;

    for (int i = 0; i < 4; i++) {
        uint8_t op = abc.at(i);
        uint32_t destination_operand = G_ANYEXT(op);

        if (offset != 0) {
            uint32_t mask = destination_operand << offset;
            dstcast |= mask;
            acc = dstcast | mask;
        } else {
            acc = dstcast | destination_operand;
        }

        dstcast = acc;
        offset += 8;
    }
    return dstcast;
}

int main() {
    rc::check("pack unsigned numbers", []() {
            const uint32_t value = 0xDEADBEEF;
            const std::vector<uint8_t> split {0xEF, 0xBE, 0xAD, 0xDE};
            const uint32_t packed = pack_vector(split);
            RC_ASSERT(packed == value);
        });

    rc::check("pack unsigned numbers random",
        [](const uint8_t a, const uint8_t b, const uint8_t c, const uint8_t d) {
           std::vector<uint8_t> vec {a, b, c, d};
           const uint32_t packed = pack_vector(vec);
           const uint32_t custom = a | b << 8 | c << 16 | d << 24;
           RC_ASSERT(packed == custom);
        });

    rc::check("unpack unsigned numbers", []() {
            const uint32_t values = 0xDEADBEEF;
            const std::vector<uint8_t> unpacked = unpack_vector(values);
            const std::vector<uint8_t> custom { 0xEF, 0xBE, 0xAD, 0xDE };
            RC_ASSERT(unpacked == custom);
            for (size_t i = 0; i < unpacked.size(); ++i)
                RC_ASSERT(unpacked.at(i) == custom.at(i));
        });

    rc::check("unpack unsigned numbers random",
        [](const uint8_t a, const uint8_t b, const uint8_t c, const uint8_t d) {
            const uint32_t value = a | b << 8 | c << 16 | d << 24;
            const std::vector<uint8_t> packed = unpack_vector(value);
            const std::vector<uint8_t> custom {a, b, c, d};
            RC_ASSERT(packed == custom);
        });

  rc::check("pack and unpack cancel each other out",
        [](const uint8_t a, const uint8_t b, const uint8_t c, const uint8_t d) {
            const std::vector<uint8_t> vec {a, b, c, d};
            std::vector newvec = unpack_vector(pack_vector(vec));
            RC_ASSERT(vec == newvec);
        });

  rc::check("Signedness is preserved whenever values are cast",
        [](const int8_t a, const int8_t b, const int8_t c, const int8_t d) {
            const std::vector<uint8_t> vec { static_cast<uint8_t>(a),
                static_cast<uint8_t>(b), static_cast<uint8_t>(c),
                static_cast<uint8_t>(d) };
           std::vector newvec = unpack_vector(pack_vector(vec));
           RC_ASSERT(vec == newvec);
           RC_ASSERT(static_cast<int8_t>(newvec.at(0)) == a);
           RC_ASSERT(static_cast<int8_t>(newvec.at(1)) == b);
           RC_ASSERT(static_cast<int8_t>(newvec.at(2)) == c);
           RC_ASSERT(static_cast<int8_t>(newvec.at(3)) == d);
        });
}
```

ValentijnvdBeek:
This code has already been reviewed off-repo before the move to Github. Some of the issues there an incorrect opcode was used (ANYEXT instead of ZEXT), lack of tests for negative numbers and the lack of tests where operands are from registers.

llvm/lib/Target/AIE/AIELegalizerInfo.cpp
```diff
@@ -491,6 +491,60 @@ bool AIELegalizerInfo::legalizeCustom(LegalizerHelper &Helper,
...
+  const LLT SourceRegTy = MRI.getType(SourceReg);
+  const Register DstReg = MI.getOperand(0).getReg();
+  assert(SourceRegTy.getSizeInBits() == 32 &&
+         "cannot pack or unpack vectors larger or smaller than 32-bit");
+
+  const LLT S32 = LLT::scalar(32);
+  unsigned Offset = 0;
+  Register DstCastReg = MRI.createGenericVirtualRegister(S32);
+
+  // We want to skip the destination vector for unmerging and the source vector
```
konstantinschwarz: Nit: update comment now that we agreed to have separate functions for packing/unpacking
Same in the `assert` above

llvm/lib/Target/AIE/AIELegalizerInfo.cpp
```diff
@@ -491,6 +491,60 @@ bool AIELegalizerInfo::legalizeCustom(LegalizerHelper &Helper,
...
+  // for merging
+  MachineOperand *Operand = MI.operands_begin() + 1,
+                 *OperandEnd = MI.operands_end();
+  MIRBuilder.buildConstant(DstCastReg, 0);
+
+  while (Operand != OperandEnd) {
+    Register DestinationOperand = Operand->getReg();
+    const LLT RegTy = MRI.getType(DestinationOperand);
```
konstantinschwarz: Nit: RegTy is required to be the same for all operands of a G_BUILD_VECTOR
Hence, you can get `RegTy` from the first `Operand` outside of the loop. Same for the size of `RegTy` used to increase the `Offset` at the end of the loop

ValentijnvdBeek: Yes, that makes sense. Doing it in the loop was needlessly defensive, thanks.

llvm/lib/Target/AIE/AIELegalizerInfo.cpp
ValentijnvdBeek: The comment is in error, the assert is iffy. Personally, I would lean on keeping it since the error message is still true and combining them isn't crazy. I did change it since I was modifying the file anyways.

llvm/lib/Target/AIE/AIELegalizerInfo.cpp
```diff
@@ -630,6 +671,9 @@ bool AIELegalizerInfo::legalizeG_UNMERGE_VALUES(LegalizerHelper &Helper,
     assert(CurrentTy.isScalar() &&
            "this operation is only supported for scalar types");
 
+    if (LastTy.getSizeInBits() == 32)
```
konstantinschwarz: Move outside of the `for` loop, below the assert

ValentijnvdBeek: Thanks for the review, that makes sense. I will update it, thank you Konstantin.

konstantinschwarz (approved):

PR #39, opened 2024-05-17
Support G_EXTRACT_VALUES for 16, 32 and 128-bit vectors 

ValentijnvdBeek:
Implements supports for the extraction of vector smaller than 256-bit which was missing for in previous versions of the code base. It relies on #38 since it needs to store the 32-bit vectors into integers, which is implemented there. The implementation is mainly differences in the legalization code, so it is all smashed together since otherwise it will be mostly two line commits.


llvm/lib/Target/AIE/AIELegalizerInfo.cpp
```diff
@@ -407,25 +431,28 @@ AIELegalizerInfo::AIELegalizerInfo(const AIEBaseSubtarget &ST) {
         // TODO: can be implemented by unpadding the source vector into Src1,
```
konstantinschwarz: This TODO is fixed by the next commit?

konstantinschwarz (approved):

PR #40, opened 2024-05-17
[AIE2] Implement simple (and inefficient) Shuffle Vector legalization

ValentijnvdBeek:
This implements the simple legalization that lowers G_SHUFFLE_VECTOR into extracts of the elements based on the mask and then combining them using a G_BUILD_VECTOR. This is done as a basic feature of the LLVM framework, but this causes very inefficient legalizations. You can also optimize out some of the patterns as done by the AArch64 architecture, which replaces it with some opcode. What is nice about our architecture is that we could do the same thing using the `vshuffle` instruction, which might lead to more fast patterns.


llvm/lib/Target/AIE/AIELegalizerInfo.cpp
```diff
@@ -493,6 +493,24 @@ AIELegalizerInfo::AIELegalizerInfo(const AIEBaseSubtarget &ST) {
         .clampMaxNumElements(0, S16, 32)
         .clampMaxNumElements(0, S32, 16)
         .custom();
+
+    getActionDefinitionsBuilder(G_SHUFFLE_VECTOR)
+        .unsupportedIf(IsNotValidDestinationVector)
```
konstantinschwarz, line 510: As discussed, this should always call `.lower()`. We'll detect legal patterns before legalization and turn them into alternative operations

llvm/test/CodeGen/AIE/aie2/GlobalISel/legalize-shuffle-vector.mir
```diff
@@ -0,0 +1,1007 @@
...
+    ; CHECK-NEXT: [[AIE_ADD_VECTOR_ELT_LEFT28:%[0-9]+]]:_(<16 x s32>) = G_AIE_ADD_VECTOR_ELT_LEFT [[AIE_ADD_VECTOR_ELT_LEFT27]], [[AIE_SEXT_EXTRACT_VECTOR_ELT19]](s32)
+    ; CHECK-NEXT: [[AIE_ADD_VECTOR_ELT_LEFT29:%[0-9]+]]:_(<16 x s32>) = G_AIE_ADD_VECTOR_ELT_LEFT [[AIE_ADD_VECTOR_ELT_LEFT28]], [[AIE_SEXT_EXTRACT_VECTOR_ELT18]](s32)
+    ; CHECK-NEXT: [[AIE_ADD_VECTOR_ELT_LEFT30:%[0-9]+]]:_(<16 x s32>) = G_AIE_ADD_VECTOR_ELT_LEFT [[AIE_ADD_VECTOR_ELT_LEFT29]], [[AIE_SEXT_EXTRACT_VECTOR_ELT17]](s32)
+    ; CHECK-NEXT: [[AIE_ADD_VECTOR_ELT_LEFT31:%[0-9]+]]:_(<16 x s32>) = G_AIE_ADD_VECTOR_ELT_LEFT [[AIE_ADD_VECTOR_ELT_LEFT30]], [[AIE_SEXT_EXTRACT_VECTOR_ELT16]](s32)
+    ; CHECK-NEXT: [[CONCAT_VECTORS:%[0-9]+]]:_(<32 x s32>) = G_CONCAT_VECTORS [[AIE_ADD_VECTOR_ELT_LEFT15]](<16 x s32>), [[AIE_ADD_VECTOR_ELT_LEFT31]](<16 x s32>)
+    ; CHECK-NEXT: PseudoRET implicit $lr, implicit [[CONCAT_VECTORS]](<32 x s32>)
+    %0:_(<16 x s32>) = G_IMPLICIT_DEF
+    %1:_(<32 x s32>) = G_SHUFFLE_VECTOR %0(<16 x s32>), %0, shufflemask(0, 16, 1, 17, 2, 18, 3, 19, 4, 20, 5, 21, 6, 22, 7, 23, 8, 24, 9, 25, 10, 26, 11, 27, 12, 28, 13, 29, 14, 30, 15, 31)
```
konstantinschwarz: These tests for valid AIE VSHUFFLE shufflemasks should be moved to the pre-legalizer combiner where we match these masks

llvm/lib/Target/AIE/AIELegalizerInfo.cpp
ValentijnvdBeek: This line is there to reject the patterns that we indicate in a xfail test. So either we need to decide that these test are not valid or reject them before the legalizer.

https://github.com/Xilinx/llvm-aie/blob/aie-public/llvm/test/CodeGen/AIE/GlobalISel/xfail-legalize-shufflevector.mir

llvm/test/CodeGen/AIE/aie2/GlobalISel/legalize-shuffle-vector.mir
ValentijnvdBeek: Hi Konstantin, thank you for your review. I am working on improving the `vshuffle` implementation and I will split it off into a separate commit. I will be sure to take this into account when I push that.

llvm/lib/Target/AIE/AIELegalizerInfo.cpp
konstantinschwarz: There is a final `.lower()`, so no need for the `lowerIf` anymore.
Do we need the clampMinNumElements?

ValentijnvdBeek: The `clampMinNumElements` do something useful in that it avoids generating bad IR. Lowering it directly will cause them to crash since you are moving the buck down. It is not that useful, but actually what it should do is clamp the arguments instead. I will change it to that.

`lowerIf` is mixed. I could remove most of them, but I also see the point in keeping them there as documentation. What a shufflevector looks like isn't very clear and it is non-obvious to someone new to LLVM that it can turn scalars into vectors or that there is a "canonical" vector with different properties. Lowering them separately makes it clear that they are different cases. So I am a bit mixed on that.

konstantinschwarz: I removed all but the `.lower()` action and all regression tests pass, except [xfail-legalize-shufflevector.mir](https://github.com/Xilinx/llvm-aie/blob/aie-public/llvm/test/CodeGen/AIE/GlobalISel/xfail-legalize-shufflevector.mir) . Seems like there are missing tests for the `clampMinNumElements` cases?
Could you either add a test that shows these clamps are needed, or remove all other legalization actions.

The xfail test fails because `.lower` actually generates the expected MIR instead of crashing. We can remove the RUN line for aie2 for that test

ValentijnvdBeek: Yeah, fair enough. Clamping for the input vector can be useful, which is different than what it does now, however recently 64-bit vectors support was added in #77 which is the main missing part. I'll remove it for now, but keep in mind that this could lead to errors in the form of: `cannot legalize: %0:_<1 x i16> = G_SOME_FUNC`.

I am keeping two checks: canonical and the check whether it has an AIE2 vector output. The former does weird things to the Combiner that I would like people to be aware and the latter gives some flexibility for future architectures.

konstantinschwarz: ok fine to me. G_SHUFFLE_VECTOR is able to produce scalar results, and the LegalizerHelper has support for lowering those. Also, [xfail-legalize-shufflevector.mir](https://github.com/Xilinx/llvm-aie/blob/aie-public/llvm/test/CodeGen/AIE/GlobalISel/xfail-legalize-shufflevector.mir) contains many cases that we actually support with this (256/512-bit vectors). Could you remove those?

llvm/test/CodeGen/AIE/GlobalISel/xfail-legalize-shufflevector.mir
```diff
@@ -28,48 +28,3 @@ body:             |
...
-  bb.0:
-    %1:_(<32 x s16>) = G_IMPLICIT_DEF
-    %2:_(p0) = G_IMPLICIT_DEF
-    %0:_(<32 x s16>) = G_SHUFFLE_VECTOR %1(<32 x s16>), %1, shufflemask(undef, undef, undef, undef, undef, undef, undef, undef, undef, undef, undef, undef, undef, undef, undef, undef, undef, undef, undef, undef, undef, undef, undef, undef, undef, undef, undef, undef, undef, undef, undef, undef)
-    G_STORE %0(<32 x s16>), %2(p0) :: (store (<32 x s16>) into `<32 x i16>* undef`, align 2)
-    PseudoRET implicit $lr
-...
-
----
-name:            f_2x64
```
konstantinschwarz, line 67: Super nit: this currently fails in
```
bool llvm::AIELegalizerInfo::legalizeG_BUILD_VECTOR(llvm::LegalizerHelper &, llvm::MachineInstr &) const: Assertion `(EltSize == 8 || EltSize == 16 || EltSize == 32) && "non-existent integer size"' failed.
```

konstantinschwarz (dismissed):
LGTM!

llvm/test/CodeGen/AIE/GlobalISel/xfail-legalize-shufflevector.mir
ValentijnvdBeek: This doesn't fail, it just seems like it would. The precombiner does some magic and turns it into a copy, which can be legalized completely.

Snippet from the output using commit e503a9c8a65c2a0236d891f64c06b2b5795971ac
```
Try combining %0:_(<2 x s64>) = G_SHUFFLE_VECTOR %1:_(<2 x s64>), %1:_, shufflemask(undef, undef)
10: GIM_SwitchOpcode(MIs[0], [19, 225), Default=5243, JumpTable...) // Got=213
5188: Begin try-block
5195: GIM_CheckSimplePredicate(Predicate=21)
5199: GIM_CheckCxxPredicate(MIs[0], Predicate=15)
5202: GIR_CustomAction(FnID=14)
CSEInfo::Found Instr %1:_(<2 x s64>) = G_IMPLICIT_DEF
Creating: COPY

Creating: COPY

Erasing: %0:_(<2 x s64>) = G_SHUFFLE_VECTOR %1:_(<2 x s64>), %1:_, shufflemask(undef, undef)

5203: GIR_Done
Created: %0:_(<2 x s64>) = COPY %1:_(<2 x s64>)     
```

konstantinschwarz: The pre-legalizer combiner has nothing to do with this, this test runs the legalizer.
Please add it back, then this PR can be merged

ValentijnvdBeek: Sure will do, it is strange because I did run the update script multiple times before committing the change. I'll put that down to "works on my machine"

ValentijnvdBeek: It is added back to the commit

konstantinschwarz (approved):

PR #41, opened 2024-05-17
[LLVM] Optimize G_SHUFFLE_VECTOR into more efficient generic opcodes

ValentijnvdBeek:
This merge requests implements an infrastructure based on streams that allows for the creation of shuffle vector pattern masks that can be used to replace inefficient scalarizations with faster generic opcodes. It implements five patterns that are useful for the AIE going forwards, namely one that matches the concatenation of vectors, the same in reverse, the insertion of a vector and the merging of two vectors. 

Some todos in the merge request that can be done after the rest has been OK'ed
- [ ] The assertions that check the initial state are done a bit adhoc, it might be nice to pass them as function pointers as well
- [ ] Inefficient matching for larger groups of patterns. The patterns now match the minimum of what they can. For example, only the first two parts of vectors are transformed. Most of these can be extended, but this leads to very large set of patterns quickly
- [x] There are two matching functions, since the concat version does a lot of additional work that I am not sure what it does. I should probably figure out if those can be fully merged or whether just the simply variant ought to survive
- [x] Use llvm::SmallVector and not std::vector


llvm/lib/CodeGen/GlobalISel/CombinerHelper.cpp
```diff
@@ -301,17 +303,55 @@ void CombinerHelper::applyCombineConcatVectors(
...
+  // Turns a shuffle vector that only increments into a concat vector
+  // instruction
+  std::function<std::optional<int32_t>()> CountUp =
+      adderGenerator(0, DstNumElts - 1, 1);
+  if (matchCombineShuffleVector(MI, Ops, CountUp, 2 * SrcNumElts)) {
     applyCombineShuffleVector(MI, Ops);
     return true;
   }
+
+  // {n, n-1, ..., 2, 1} -> G_CONCAT_VECTOR
```
konstantinschwarz: This requires a test.
Also, this reverses the MSB and LSB elements within each sub-vector, this cannot be matched to G_CONCAT_VECTOR

ValentijnvdBeek: Thanks Konstantin, the pattern indeed is in error. I'll update so it makes sense `(n/2, ..., n, 1, ..., n/2-1)`, the opcode replacement is still a good idea, so I will keep that in.

I did write a test, but I see now that is in the wrong commit. I will fix that as well. Thanks for the comment.

llvm/test/CodeGen/AArch64/GlobalISel/prelegalizercombiner-undef.mir
```diff
@@ -216,16 +216,16 @@ body:             |
...
-    ; CHECK-NEXT: [[DEF:%[0-9]+]]:_(<2 x s32>) = G_IMPLICIT_DEF
-    ; CHECK-NEXT: [[COPY:%[0-9]+]]:_(<2 x s32>) = COPY $d0
-    ; CHECK-NEXT: [[SHUF:%[0-9]+]]:_(<2 x s32>) = G_SHUFFLE_VECTOR [[DEF]](<2 x s32>), [[COPY]], shufflemask(0, 1)
-    ; CHECK-NEXT: $d0 = COPY [[SHUF]](<2 x s32>)
-    ; CHECK-NEXT: RET_ReallyLR implicit $d0
-    %1:_(<2 x s32>) = G_IMPLICIT_DEF
-    %2:_(<2 x s32>) = COPY $d0
-    %0:_(<2 x s32>) = G_SHUFFLE_VECTOR %1(<2 x s32>), %2(<2 x s32>), shufflemask(0, 1)
-    $d0 = COPY %0(<2 x s32>)
-    RET_ReallyLR implicit $d0
```
konstantinschwarz: Why did you change this AArch64 test input?

llvm/test/CodeGen/AArch64/ext-narrow-index.ll
```diff
@@ -118,11 +117,16 @@ entry:
...
-; CHECK-NEXT:    ret
+; CHECK-SD-LABEL: i16_off8:
+; CHECK-SD:       // %bb.0: // %entry
+; CHECK-SD-NEXT:    mov v0.16b, v1.16b
+; CHECK-SD-NEXT:    // kill: def $d0 killed $d0 killed $q0
+; CHECK-SD-NEXT:    ret
+;
+; CHECK-GISEL-LABEL: i16_off8:
+; CHECK-GISEL:       // %bb.0: // %entry
+; CHECK-GISEL-NEXT:    // kill: def $d0 killed $d0 killed $q0
```
konstantinschwarz: This exposes a bug:
```
Erasing: %2:_(<4 x s16>) = G_SHUFFLE_VECTOR %0:_(<8 x s16>), %1:_, shufflemask(8, 9, 10, 11)

Created: %2:_(<4 x s16>), %3:_(<4 x s16>) = G_UNMERGE_VALUES %0:_(<8 x s16>)
```

The shufflemask selects the 4 lower elements of `%1`, while the `G_UNMERGE_VALUES` gets them from `%0`.

This is the incoming MIR:
```
bb.1.entry:
  liveins: $q0, $q1
  %0:_(<8 x s16>) = COPY $q0
  %1:_(<8 x s16>) = COPY $q1
  %2:_(<4 x s16>) = G_SHUFFLE_VECTOR %0:_(<8 x s16>), %1:_, shufflemask(8, 9, 10, 11)
  $d0 = COPY %2:_(<4 x s16>)
  RET_ReallyLR implicit $d0
```

llvm/test/CodeGen/AArch64/vecreduce-add.ll
```diff
@@ -6305,69 +6305,16 @@ entry:
...
-; CHECK-GI-DOT-NEXT:    add v1.4s, v5.4s, v1.4s
-; CHECK-GI-DOT-NEXT:    add v2.4s, v6.4s, v2.4s
-; CHECK-GI-DOT-NEXT:    add v3.4s, v7.4s, v3.4s
-; CHECK-GI-DOT-NEXT:    add v0.4s, v0.4s, v1.4s
-; CHECK-GI-DOT-NEXT:    add v1.4s, v2.4s, v3.4s
-; CHECK-GI-DOT-NEXT:    add v0.4s, v0.4s, v1.4s
-; CHECK-GI-DOT-NEXT:    addv s0, v0.4s
-; CHECK-GI-DOT-NEXT:    fmov w0, s0
-; CHECK-GI-DOT-NEXT:    ret
+; CHECK-LABEL: add_pair_v8i16_v4i32_double_sext_zext_shuffle:
```
konstantinschwarz: Yay!

llvm/test/CodeGen/AIE/aie2/intrinsics-shufflevec.ll
```diff
@@ -15,69 +15,18 @@ define <8 x i32> @test_extract_vector(<16 x i32> noundef %a, i32 noundef %idx) {
...
-; CHECK-NEXT:    mova r16, #11
-; CHECK-NEXT:    vextract.s32 r3, x2, r16
-; CHECK-NEXT:    nop
-; CHECK-NEXT:    mova r16, #12
-; CHECK-NEXT:    vextract.s32 r4, x2, r16
-; CHECK-NEXT:    nop
-; CHECK-NEXT:    mova r16, #13
-; CHECK-NEXT:    vextract.s32 r5, x2, r16
-; CHECK-NEXT:    j #.LBB0_3
+; CHECK-NEXT:    nopb ; nopa ; nops ; nopx ; vmov wl0, wh0; nopv
```
konstantinschwarz, line 21: 🚀

llvm/lib/CodeGen/GlobalISel/CombinerHelper.cpp
```diff
@@ -408,6 +408,30 @@ bool CombinerHelper::tryCombineShuffleVector(MachineInstr &MI) {
     return true;
   }
 
+  // {1, 2, ..., |DstVector|} -> G_UNMERGE_VALUES
+  // Extracts the first chunk of the same size of the destination vector from
+  // the source
+  GeneratorType FirstQuarter = adderGenerator(0, DstNumElts - 1, 1);
+  if (matchCombineShuffleVector(MI, FirstQuarter, DstNumElts - 1)) {
+    if (SrcTy == DstTy || ((SrcNumElts / 2) % 2) != 0 || SrcNumElts % DstNumElts != 0 )
```
konstantinschwarz: Needs some documentation

konstantinschwarz (changes requested):

llvm/test/CodeGen/AArch64/GlobalISel/prelegalizercombiner-undef.mir
ValentijnvdBeek: This test is the "shufflevector_not_all_ops_undef", so it tests specifically their precombiner whenever one of the two is not undefined. Our combiner will optimize this line of code into a simple unmerge, this is a win for IR, but this has nothing to do with what this test actually does. So if I don't change it, then this test stops being useful and the overall test coverage goes down.

llvm/test/CodeGen/AArch64/ext-narrow-index.ll
ValentijnvdBeek: Good catch, I will look at it. Thanks!

ValentijnvdBeek: This was resolved in the latest push

llvm/lib/CodeGen/GlobalISel/CombinerHelper.cpp
```diff
@@ -316,6 +316,54 @@ adderGenerator(const int32_t From, const int32_t To, const int32_t StepSize) {
   };
 }
 
+Register CombinerHelper::createUnmergeValue(
+    MachineInstr &MI, const Register SrcReg, const Register DstReg,
+    const uint8_t DestinationIndex, const uint32_t Start, const uint32_t End) {
+  Builder.setInsertPt(*MI.getParent(), MI);
+  const LLT DstTy = MRI.getType(DstReg);
+  const LLT SrcTy = MRI.getType(SrcReg);
+  assert((!DstTy.isVector() ||
```
konstantinschwarz: De-Morgan this?

llvm/test/CodeGen/AArch64/GlobalISel/prelegalizercombiner-undef.mir
konstantinschwarz: I think this test was running into the same bug https://github.com/Xilinx/llvm-aie/pull/41#discussion_r1649336200. In fact, there is nothing to unmerge here.

If I revert the test input back to its original source, the CHECK lines do not change. Thus, we should not modify these upstream tests.

There seems to be a missing optimization though: this shuffle vector simply passes through the lower vector (%1), hence this could be optimized to a simple `COPY`. That should be a separate PR though.

llvm/test/CodeGen/AIE/aie2/GlobalISel/prelegalizercombiner-shufflevector.mir
```diff
@@ -258,3 +297,117 @@ body:             |
...
+    ; CHECK-NEXT: [[COPY:%[0-9]+]]:_(<128 x s8>) = COPY $y2
+    ; CHECK-NEXT: [[UV:%[0-9]+]]:_(<64 x s8>), [[UV1:%[0-9]+]]:_(<64 x s8>) = G_UNMERGE_VALUES [[COPY]](<128 x s8>)
+    ; CHECK-NEXT: [[UV2:%[0-9]+]]:_(<32 x s8>), [[UV3:%[0-9]+]]:_(<32 x s8>) = G_UNMERGE_VALUES [[UV]](<64 x s8>)
+    ; CHECK-NEXT: [[AIE_UNPAD_VECTOR:%[0-9]+]]:_(<16 x s8>) = G_AIE_UNPAD_VECTOR [[UV2]](<32 x s8>)
+    ; CHECK-NEXT: [[UV4:%[0-9]+]]:_(<8 x s8>), [[UV5:%[0-9]+]]:_(<8 x s8>) = G_UNMERGE_VALUES [[AIE_UNPAD_VECTOR]](<16 x s8>)
+    ; CHECK-NEXT: [[UV6:%[0-9]+]]:_(<4 x s8>), [[UV7:%[0-9]+]]:_(<4 x s8>) = G_UNMERGE_VALUES [[UV4]](<8 x s8>)
+    ; CHECK-NEXT: PseudoRET implicit $lr, implicit [[UV7]](<4 x s8>)
+    %1:_(<128 x s8>) = COPY $y2
+    %2:_(<4 x s8>) = G_SHUFFLE_VECTOR %1:_(<128 x s8>), %1:_(<128 x s8>), shufflemask(4, 5, 6, 7)
+    PseudoRET implicit $lr, implicit %2
```
konstantinschwarz, line 582: Could you add some tests where we extract from the second input, coming from a separate register?
e.g.
```
---
name:            extract_vector_third_half_1024
legalized:       false
body:             |
  bb.1.entry:
    liveins: $y2, $y3
    %1:_(<32 x s32>) = COPY $y2
    %2:_(<32 x s32>) = COPY $y3
    %0:_(<16 x s32>) = G_SHUFFLE_VECTOR %1:_(<32 x s32>), %2:_, shufflemask(32,33,34,35,36,37,38,39,40,41,42,43,44,45,46,47)
    $x0 = COPY %0:_(<16 x s32>)
    PseudoRET implicit $lr, implicit $x0
...
```

ValentijnvdBeek: That is a great suggestion, thanks Konstantin

llvm/lib/CodeGen/GlobalISel/CombinerHelper.cpp
ValentijnvdBeek: DeMorgan doesn't do much, since `(!x || y) && z` doesn't really turn into something nice if you negate it. Instead I think it is better if I check for scalars directly instead of checking if it is not a vector.

llvm/lib/CodeGen/GlobalISel/CombinerHelper.cpp
```diff
@@ -384,17 +385,265 @@ void CombinerHelper::applyCombineShuffleConcat(MachineInstr &MI,
...
+  const LLT SrcTy = MRI.getType(MI.getOperand(1).getReg());
+
+  const unsigned DstNumElts = DstTy.isVector() ? DstTy.getNumElements() : 1;
+  const unsigned SrcNumElts = SrcTy.isVector() ? SrcTy.getNumElements() : 1;
+
+  // This test is a bit silly, but it is required because some tests rely on
+  // the legalizer changing the type of the shufflevector.
+  if (DstTy.getScalarSizeInBits() == 1)
+    return false;
+
```
ValentijnvdBeek, line 480: Yeah, this is a bit weird. There is an ARM64 test which creates an array of 8 1-bit numbers that the legalizer turns into 8-bit numbers. At the moment, their legalizer doesn't do this for some of the replacement op codes.

llvm/lib/Target/AIE/AIECombine.td
```diff
@@ -63,7 +63,7 @@ def AIE2PreLegalizerCombiner
                                                                  combine_globalval_offset,
                                                                  combine_extract_vector_elt_and_zsa_ext,
                                                                  combine_splat_vector ]> {
-  let CombineAllMethodName = "tryCombineAllImpl";
+    let CombineAllMethodName = "tryCombineAllImpl";
```
konstantinschwarz: nit: unrelated change

llvm/lib/CodeGen/GlobalISel/CombinerHelper.cpp
```diff
@@ -371,7 +373,6 @@ void CombinerHelper::applyCombineShuffleConcat(MachineInstr &MI,
                                                SmallVector<Register> &Ops) {
   LLT SrcTy = MRI.getType(Ops[0]);
   Register UndefReg = 0;
-
```
konstantinschwarz: Unrelated change. Since these changes cause unnecessary conflicts with upstream merges, I'd like to avoid them.

llvm/lib/CodeGen/GlobalISel/CombinerHelper.cpp
```diff
@@ -421,51 +483,24 @@ bool CombinerHelper::matchCombineShuffleVector(MachineInstr &MI,
...
     int Idx = Mask[i];
+    const int32_t ShiftIndex = Generator().value_or(-1);
+
     // Undef value.
-    if (Idx < 0)
+    if (Idx < 0 || ShiftIndex < 0)
       continue;
-    // Ensure the indices in each SrcType sized piece are sequential and that
+
+    // Ensure the indices in each SrcType sized piece are seqential and that
```
konstantinschwarz: nit: its really seq*u*ential

konstantinschwarz (commented):
There are plenty of upstream tests failing. Could you please take a look?

llvm/lib/CodeGen/GlobalISel/CombinerHelper.cpp
ValentijnvdBeek: Thanks, indeed it is not a conseqential but important change :)

llvm/lib/CodeGen/GlobalISel/CombinerHelper.cpp
```diff
@@ -500,6 +509,38 @@ bool CombinerHelper::tryCombineShuffleVector(MachineInstr &MI) {
...
 
+  // {1, 2, ..., |DstVector|} -> G_UNMERGE_VALUES
+  // Extracts the first chunk of the same size of the destination vector from
+  // the source
+  GeneratorType FirstQuarter = adderGenerator(0, DstNumElts - 1, 1);
+  if (matchCombineShuffleVector(MI, FirstQuarter, DstNumElts - 1)) {
+    // This optimization does not work if the target type is not a power of two,
+    // this can happen in some backends that support uneven vector types. We
+    // also need to make sure that the vector can be split into two.
+    if (SrcTy == DstTy || ((SrcNumElts / 2) % 2) != 0 ||
```
konstantinschwarz, line 534: Maybe use `isPowerOf2_32`?
`((SrcNumElts / 2) % 2) != 0` looks overly complicated

llvm/lib/CodeGen/GlobalISel/CombinerHelper.cpp
```diff
@@ -500,6 +509,38 @@ bool CombinerHelper::tryCombineShuffleVector(MachineInstr &MI) {
...
+    return true;
+  }
+
+  // {|DstVector|, |DstVector|+1, ..., 2 * |DstVector|} -> G_UNMERGE_VALUES
+  // Extracts the second chunk of the same size of the destination vector from
+  // the source
+  GeneratorType SecondQuarter =
+      adderGenerator(DstNumElts, (DstNumElts * 2) - 1, 1);
+  if (matchCombineShuffleVector(MI, SecondQuarter, DstNumElts - 1)) {
+    if (((SrcNumElts / 2) % 2) != 0 || SrcNumElts % DstNumElts != 0)
```
konstantinschwarz, line 550: Same here

llvm/test/CodeGen/AIE/aie2/GlobalISel/prelegalizercombiner-shufflevector.mir
```diff
@@ -682,3 +682,172 @@ body:             |
...
+    ; CHECK: liveins: $wl2, $wl4
+    ; CHECK-NEXT: {{  $}}
+    ; CHECK-NEXT: [[COPY:%[0-9]+]]:_(<8 x s32>) = COPY $wl2
+    ; CHECK-NEXT: [[COPY1:%[0-9]+]]:_(<8 x s32>) = COPY $wl4
+    ; CHECK-NEXT: [[CONCAT_VECTORS:%[0-9]+]]:_(<16 x s32>) = G_CONCAT_VECTORS [[COPY1]](<8 x s32>), [[COPY]](<8 x s32>)
+    ; CHECK-NEXT: $x0 = COPY [[CONCAT_VECTORS]](<16 x s32>)
+    ; CHECK-NEXT: PseudoRET implicit $lr, implicit $x0
+    %1:_(<8 x s32>) = COPY $wl2
+    %2:_(<8 x s32>) = COPY $wl4
+    %0:_(<16 x s32>) = G_SHUFFLE_VECTOR %1:_(<8 x s32>), %2:_, shufflemask(8, 9, 10, 11, 12, 13, 14, 15, 0, 1, 2, 3, 4, 5, 6, 7)
```
konstantinschwarz: Do we ever see this pattern? I'm curious, because it looks to me like a generic canonicalization to just swap the two inputs and rewrite the shufflemask to run from 0-15.
Do you know whether that canonicalization exists for LLVM IR?

llvm/lib/CodeGen/GlobalISel/CombinerHelper.cpp
ValentijnvdBeek: I think it is a case of an outdated comment. Using `isPoweredOf2_32` does make all the tests pass, but adds the restriction that all target types are powers of 2. This is not true for all architectures. 

For example, this test allows for the following optimization which `isPowerOf2_32` would miss
```
bb.0.entry:
  liveins: $wl0, $wl1
  %2:_(s32) = G_CONSTANT i32 4
  %3:_(s32) = G_CONSTANT i32 12
  %4:_(s32) = G_CONSTANT i32 9
  %5:_(s32) = G_CONSTANT i32 0
  %1:_(<12 x s32>) = G_BUILD_VECTOR %2:_(s32), %3:_(s32), %4:_(s32), %5:_(s32), %5:_(s32), %5:_(s32), %2:_(s32), %3:_(s32), %4:_(s32), %5:_(s32), %5:_(s32), %5:_(s32)
  %6:_(<6 x s32>), %11:_(<6 x s32>) = G_UNMERGE_VALUES %1:_(<12 x s32>)
  PseudoRET implicit $lr, implicit %6:_(<6 x s32>)
```

llvm/test/CodeGen/AIE/aie2/GlobalISel/prelegalizercombiner-shufflevector.mir
ValentijnvdBeek: As far as, I can see, the answer is no. There is some analysis done `InstructionSimplify.cpp`, but they don't cover this case. If you feed it just LLVM IR for a shuffle vector, it will also not reorder it. But actually, while looking at it now, this code can be dropped since one of the earlier changes during a rewrite made this case be covered under an earlier commit

konstantinschwarz: Could you please move the tests from this commit to that earlier change then and drop the code change?

ValentijnvdBeek: Yes, of course, I was just running the tests before I pushed it. Never can be too sure

llvm/include/llvm/CodeGen/GlobalISel/CombinerHelper.h
```diff
@@ -256,8 +257,11 @@ class CombinerHelper {
   /// concat_vectors.
   ///
   /// \pre MI.getOpcode() == G_SHUFFLE_VECTOR.
-  bool matchCombineShuffleVector(MachineInstr &MI,
-                                 SmallVectorImpl<Register> &Ops);
+  using GeneratorType = std::function<std::optional<int32_t>()>;
+
+  bool matchCombineShuffleVector(MachineInstr &MI, GeneratorType Generator,
```
gbossu, line 263: The docstring for `matchCombineShuffleVector` is now physically disconnected from the function. Can you fix it (and update it)?

I think some documentation for `GeneratorType` would also be helpful, it's not immediately clear to me what it represents.

llvm/lib/CodeGen/GlobalISel/CombinerHelper.cpp
```diff
@@ -384,17 +386,78 @@ void CombinerHelper::applyCombineShuffleConcat(MachineInstr &MI,
...
+  return [Counter, To, StepSize]() mutable {
+    std::optional<int32_t> OldCount = std::optional<int32_t>(Counter);
+    Counter += StepSize;
+    if (OldCount == (To + StepSize))
+      OldCount = {};
+    return OldCount;
+  };
+}
+
 bool CombinerHelper::tryCombineShuffleVector(MachineInstr &MI) {
```
gbossu, line 465: From what I understand, this isn't an NFC change. I think it would really help readability if you first introduce `GeneratorType` in an NFC change, and then in a follow-up commit add extra features/combines.

llvm/lib/CodeGen/GlobalISel/CombinerHelper.cpp
```diff
@@ -399,6 +399,54 @@ adderGenerator(const int32_t From, const int32_t To, const int32_t StepSize) {
   };
 }
 
+Register CombinerHelper::createUnmergeValue(
```
gbossu, line 417: Curious: Maybe `IRTranslator` already has a similar piece of code that we can re-use?

llvm/lib/CodeGen/GlobalISel/CombinerHelper.cpp
```diff
@@ -399,6 +399,21 @@ adderGenerator(const int32_t From, const int32_t To, const int32_t StepSize) {
   };
 }
 
+// Move to the next generator if it is exhausted allowing to chain generators
+std::function<std::optional<int32_t>()> concatGenerators(
```
gbossu: Didn't you cerate a nice type alias for `std::function<std::optional<int32_t>()>`? Maybe you can re-use it?

llvm/lib/CodeGen/GlobalISel/CombinerHelper.cpp
ValentijnvdBeek: Not from what I can see, `G_UNMERGE_VALUES` isn't meant for vectors to begin with. It only supports scalars, according to the documentation. Most backends just support this as an extension since there is no other opcode to do it. In theory, this will be ripped out in favour of a simple `G_SUBREGISTER_EXTRACT` in the near future if that opcode gets proper support. `Utils.cpp` does have something that is a bit similar, called `llvm::extractParts`, but that splits into two parts while this code goes through it like a binary tree.

In any case, I would keep it like this since this code should be replaced before upstreaming and that is easier if we are the only ones depending on it.

llvm/lib/CodeGen/GlobalISel/CombinerHelper.cpp
ValentijnvdBeek: I fixed this, but I do struggle a bit with seeing the difference between NFC and not. Do you have good resources for me to look at for that? I think that in the rest of the commits I did a good job of keeping them separate, but I will double-check them.

llvm/include/llvm/CodeGen/GlobalISel/CombinerHelper.h
ValentijnvdBeek: Updated the documentation and added some for GeneratorType. Thanks for pointing it out :)

llvm/lib/CodeGen/GlobalISel/CombinerHelper.cpp
ValentijnvdBeek: It used to have it, but it got lost in the see of git rebases. I modified it back again, thanks :)

llvm/include/llvm/CodeGen/GlobalISel/CombinerHelper.h
```diff
@@ -246,6 +247,12 @@ class CombinerHelper {
   void applyCombineShuffleConcat(MachineInstr &MI, SmallVector<Register> &Ops);
 
   /// Try to combine G_SHUFFLE_VECTOR into G_CONCAT_VECTORS.
+  /// A function type that returns either the next value in a
+  /// shufflemask or an empty value. Each iteration should return
+  /// one value, like a Python iterator or a Lisp stream.
+  using GeneratorType = std::function<std::optional<int32_t>()>;
+
+  /// Try to combine G_SHUFFLE_VECTOR into more efficient opcodes.
```
ValentijnvdBeek, line 254: Nit: this should be in another commit

llvm/lib/CodeGen/GlobalISel/CombinerHelper.cpp
```diff
@@ -400,16 +400,64 @@ adderGenerator(const int32_t From, const int32_t To, const int32_t StepSize) {
...
   SmallVector<Register, 4> Ops;
-  if (matchCombineShuffleVector(MI, Ops)) {
+
+  if (matchCombineShuffleVector(MI, CountUp, 2 * SrcNumElts)) {
+    // The shuffle is concatenating multiple vectors together.
+    // Collect the different operands for that.
+    Register UndefReg;
+    const Register Src1 = MI.getOperand(1).getReg();
+    const Register Src2 = MI.getOperand(2).getReg();
+
```
ValentijnvdBeek: Nit: useless newline

llvm/lib/CodeGen/GlobalISel/CombinerHelper.cpp
```diff
@@ -400,16 +400,64 @@ adderGenerator(const int32_t From, const int32_t To, const int32_t StepSize) {
...
   return false;
 }
 
 bool CombinerHelper::matchCombineShuffleVector(MachineInstr &MI,
-                                               SmallVectorImpl<Register> &Ops) {
+                                               GeneratorType Generator,
+                                               const size_t TargetDstSize) {
   assert(MI.getOpcode() == TargetOpcode::G_SHUFFLE_VECTOR &&
          "Invalid instruction kind");
```
ValentijnvdBeek, line 604: Maybe change this to pass by copy somehow? Now a new generator needs to be generated whenever it is being used. This could be a nice feature to split combiners, but just being able to pass it and reusing it again would be nicer.

llvm/include/llvm/CodeGen/GlobalISel/CombinerHelper.h
```diff
@@ -263,6 +263,16 @@ class CombinerHelper {
...
+  /// Create G_UNMERGE_VECTOR instructions until the source has reached a
+  /// target vector size.
+  ///
+  /// Requires that the destination fits evenly in the source register. It
+  /// allows you to pass which of the different destination sized slices
+  /// you require.
+  Register createUnmergeValue(MachineInstr &MI, const Register SrcReg,
+                              const Register DstReg, uint8_t DestinationIndex,
+                              const uint32_t Start, const uint32_t End);
+
```
ValentijnvdBeek, line 275: `G_UNMERGE_VALUE` isn't really meant to be used for vectors like this, it is actually defined as a [scalar opcode](https://llvm.org/docs/GlobalISel/GenericOpcode.html#g-unmerge-values). There is a nice drop-in replacement opcode called `G_EXTRACT_SUBVECTOR` which has landed in our tree in early August, but there is no support in other backends for this yet. So this implementation is a necessary evil until the other opcode is able to replace it.

llvm/test/CodeGen/AArch64/ext-narrow-index.ll
```diff
@@ -42,8 +43,7 @@ define <8 x i8> @i8_off8(<16 x i8> %arg1, <16 x i8> %arg2) {
 ;
 ; CHECK-GISEL-LABEL: i8_off8:
 ; CHECK-GISEL:       // %bb.0: // %entry
-; CHECK-GISEL-NEXT:    ext v0.16b, v0.16b, v1.16b, #8
-; CHECK-GISEL-NEXT:    // kill: def $d0 killed $d0 killed $q0
+; CHECK-GISEL-NEXT:    mov d0, v0.d[1]
 ; CHECK-GISEL-NEXT:    ret
```
ValentijnvdBeek, line 47: This test is:
```
define <8 x i8> @i8_off8(<16 x i8> %arg1, <16 x i8> %arg2) {
entry:
  %shuffle = shufflevector <16 x i8> %arg1, <16 x i8> %arg2, <8 x i32> <i32 8, i32 9, i32 10, i32 11, i32 12, i32 13, i32 14, i32 15>
  ret <8 x i8> %shuffle
}
```

This turns into:
```
bb.1.entry:
  liveins: $q0, $q1
  %0:_(<16 x s8>) = COPY $q0
  %3:_(<8 x s8>), %2:_(<8 x s8>) = G_UNMERGE_VALUES %0:_(<16 x s8>)
  $d0 = COPY %2:_(<8 x s8>)
  RET_ReallyLR implicit $d0
```

This is expected

llvm/test/CodeGen/AArch64/ext-narrow-index.ll
```diff
@@ -254,9 +254,7 @@ define <8 x i8> @i8_zero_off8(<16 x i8> %arg1) {
 ;
 ; CHECK-GISEL-LABEL: i8_zero_off8:
 ; CHECK-GISEL:       // %bb.0: // %entry
-; CHECK-GISEL-NEXT:    movi v1.2d, #0000000000000000
-; CHECK-GISEL-NEXT:    ext v0.16b, v0.16b, v1.16b, #8
-; CHECK-GISEL-NEXT:    // kill: def $d0 killed $d0 killed $q0
+; CHECK-GISEL-NEXT:    mov d0, v0.d[1]
 ; CHECK-GISEL-NEXT:    ret
 entry:
   %shuffle = shufflevector <16 x i8> %arg1, <16 x i8> zeroinitializer, <8 x i32> <i32 8, i32 9, i32 10, i32 11, i32 12, i32 13, i32 14, i32 15>
```
ValentijnvdBeek, line 260: This is:
```
define <8 x i8> @i8_zero_off8(<16 x i8> %arg1) {
entry:
  %shuffle = shufflevector <16 x i8> %arg1, <16 x i8> zeroinitializer, <8 x i32> <i32 8, i32 9, i32 10, i32 11, i32 12, i32 13, i32 14, i32 15>
  ret <8 x i8> %shuffle
}
```

And turns into:
```
bb.1.entry:
  liveins: $q0, $q1
  %0:_(<16 x s8>) = COPY $q0
  %5:_(<8 x s8>), %4:_(<8 x s8>) = G_UNMERGE_VALUES %0:_(<16 x s8>)
  $d0 = COPY %4:_(<8 x s8>)
  RET_ReallyLR implicit $d0

```

this is expected

llvm/test/CodeGen/AArch64/vecreduce-add.ll
```diff
@@ -3744,17 +3745,13 @@ define i32 @add_pair_v8i16_v4i32_double_sext_zext_shuffle(<8 x i16> %ax, <8 x i1
...
-; CHECK-GI-NEXT:    add v1.4s, v5.4s, v1.4s
-; CHECK-GI-NEXT:    add v2.4s, v6.4s, v2.4s
-; CHECK-GI-NEXT:    add v3.4s, v7.4s, v3.4s
+; CHECK-GI-NEXT:    uaddw2 v0.4s, v4.4s, v0.8h
+; CHECK-GI-NEXT:    uaddw2 v1.4s, v5.4s, v1.8h
+; CHECK-GI-NEXT:    uaddw2 v2.4s, v6.4s, v2.8h
+; CHECK-GI-NEXT:    uaddw2 v3.4s, v7.4s, v3.8h
 ; CHECK-GI-NEXT:    add v0.4s, v0.4s, v1.4s
 ; CHECK-GI-NEXT:    add v1.4s, v2.4s, v3.4s
 ; CHECK-GI-NEXT:    add v0.4s, v0.4s, v1.4s
```
ValentijnvdBeek, line 3757: this is for me the most surprising test result and hardest to parse. The input LLVM IR looks like:

```
define i32 @add_pair_v8i16_v4i32_double_sext_zext_shuffle(<8 x i16> %ax, <8 x i16> %ay, <8 x i16> %bx, <8 x i16> %by) {
entry:
  %axx = zext <8 x i16> %ax to <8 x i32>
  %s1h = shufflevector <8 x i32> %axx, <8 x i32> undef, <4 x i32> <i32 0, i32 1, i32 2, i32 3>
  %s1l = shufflevector <8 x i32> %axx, <8 x i32> undef, <4 x i32> <i32 4, i32 5, i32 6, i32 7>
  %axs = add <4 x i32> %s1h, %s1l
  %ayy = zext <8 x i16> %ay to <8 x i32>
  %s2h = shufflevector <8 x i32> %ayy, <8 x i32> undef, <4 x i32> <i32 0, i32 1, i32 2, i32 3>
  %s2l = shufflevector <8 x i32> %ayy, <8 x i32> undef, <4 x i32> <i32 4, i32 5, i32 6, i32 7>
  %ays = add <4 x i32> %s2h, %s2l
  %az = add <4 x i32> %axs, %ays
  %bxx = zext <8 x i16> %bx to <8 x i32>
  %s3h = shufflevector <8 x i32> %bxx, <8 x i32> undef, <4 x i32> <i32 0, i32 1, i32 2, i32 3>
  %s3l = shufflevector <8 x i32> %bxx, <8 x i32> undef, <4 x i32> <i32 4, i32 5, i32 6, i32 7>
  %bxs = add <4 x i32> %s3h, %s3l
  %byy = zext <8 x i16> %by to <8 x i32>
  %s4h = shufflevector <8 x i32> %byy, <8 x i32> undef, <4 x i32> <i32 0, i32 1, i32 2, i32 3>
  %s4l = shufflevector <8 x i32> %byy, <8 x i32> undef, <4 x i32> <i32 4, i32 5, i32 6, i32 7>
  %bys = add <4 x i32> %s4h, %s4l
  %bz = add <4 x i32> %bxs, %bys
  %z = add <4 x i32> %az, %bz
  %z2 = call i32 @llvm.vector.reduce.add.v4i32(<4 x i32> %z)
  ret i32 %z2
}
```
This is translated into the following global isel IR:
```

bb.1.entry:                                                                                                                                                                                                                                                        liveins: $q0, $q1, $q2, $q3
  %0:_(<8 x s16>) = COPY $q0
  %1:_(<8 x s16>) = COPY $q1
  %2:_(<8 x s16>) = COPY $q2
  %3:_(<8 x s16>) = COPY $q3
  %33:_(<4 x s16>), %34:_(<4 x s16>) = G_UNMERGE_VALUES %0:_(<8 x s16>)
  %35:_(<4 x s32>) = G_ZEXT %33:_(<4 x s16>)
  %36:_(<4 x s32>) = G_ZEXT %34:_(<4 x s16>)
  %8:_(<4 x s32>) = G_ADD %35:_, %36:_
  %37:_(<4 x s16>), %38:_(<4 x s16>) = G_UNMERGE_VALUES %1:_(<8 x s16>)
  %39:_(<4 x s32>) = G_ZEXT %37:_(<4 x s16>)
  %40:_(<4 x s32>) = G_ZEXT %38:_(<4 x s16>)
  %12:_(<4 x s32>) = G_ADD %39:_, %40:_
  %13:_(<4 x s32>) = G_ADD %8:_, %12:_
  %41:_(<4 x s16>), %42:_(<4 x s16>) = G_UNMERGE_VALUES %2:_(<8 x s16>)
  %43:_(<4 x s32>) = G_ZEXT %41:_(<4 x s16>)
  %44:_(<4 x s32>) = G_ZEXT %42:_(<4 x s16>)
  %17:_(<4 x s32>) = G_ADD %43:_, %44:_
  %45:_(<4 x s16>), %46:_(<4 x s16>) = G_UNMERGE_VALUES %3:_(<8 x s16>)
  %47:_(<4 x s32>) = G_ZEXT %45:_(<4 x s16>)
  %48:_(<4 x s32>) = G_ZEXT %46:_(<4 x s16>)
  %21:_(<4 x s32>) = G_ADD %47:_, %48:_
  %22:_(<4 x s32>) = G_ADD %17:_, %21:_
  %23:_(<4 x s32>) = G_ADD %13:_, %22:_
  %24:_(s32) = G_VECREDUCE_ADD %23:_(<4 x s32>)
  $w0 = COPY %24:_(s32)
  RET_ReallyLR implicit $w0
                                                                                                                                               
```
This corresponds to the the shufflevectors perfectly and actually shows that they simplify the unmerges as well, which is a welcome bonus.

llvm/lib/CodeGen/GlobalISel/CombinerHelper.cpp
```diff
@@ -556,6 +556,45 @@ bool CombinerHelper::tryCombineShuffleVector(MachineInstr &MI) {
...
+  // vector.
+  GeneratorType FirstEightA = adderGenerator(0, (DstNumElts / 2) - 1, 1);
+  GeneratorType FirstEightB =
+      adderGenerator(DstNumElts, DstNumElts + (DstNumElts / 2) - 1, 1);
+
+  auto UnmergeMatcher = SmallVector<GeneratorType>{FirstEightA, FirstEightB};
+  GeneratorType FirstAndThird = concatGenerators(UnmergeMatcher);
+  if (matchCombineShuffleVector(MI, FirstAndThird, (DstNumElts / 2) - 1)) {
+    if (DstNumElts <= 2)
+      return false;
```
ValentijnvdBeek, line 575: Maybe do these tests in a callback? Now, if one fails, it will exit the combiner completely rather than continuing on. For now, this is fine, but with more patterns that are close, this might be problems.

llvm/test/CodeGen/AArch64/arm64-neon-copy.ll
```diff
@@ -1814,14 +1804,13 @@ define <16 x i8> @test_concat_v16i8_v8i8_v16i8(<8 x i8> %x, <16 x i8> %y) #0 {
...
 ; CHECK-GI-NEXT:    mov v0.b[1], v2.b[0]
-; CHECK-GI-NEXT:    ldr q2, [x8, :lo12:.LCPI127_0]
 ; CHECK-GI-NEXT:    mov v0.b[2], v3.b[0]
 ; CHECK-GI-NEXT:    mov v0.b[3], v4.b[0]
 ; CHECK-GI-NEXT:    mov v0.b[4], v5.b[0]
 ; CHECK-GI-NEXT:    mov v0.b[5], v6.b[0]
 ; CHECK-GI-NEXT:    mov v0.b[6], v7.b[0]
 ; CHECK-GI-NEXT:    mov v0.b[7], v16.b[0]
-; CHECK-GI-NEXT:    tbl v0.16b, { v0.16b, v1.16b }, v2.16b
+; CHECK-GI-NEXT:    mov v0.d[1], v1.d[0]
```
ValentijnvdBeek, line 1813: The input LLVM IR is: 
```
──────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────define <16 x i8> @test_concat_v16i8_v8i8_v16i8(<8 x i8> %x, <16 x i8> %y) #0 {
entry:
  %vecext = extractelement <8 x i8> %x, i32 0
  %vecinit = insertelement <16 x i8> undef, i8 %vecext, i32 0
  %vecext1 = extractelement <8 x i8> %x, i32 1
  %vecinit2 = insertelement <16 x i8> %vecinit, i8 %vecext1, i32 1
  %vecext3 = extractelement <8 x i8> %x, i32 2
  %vecinit4 = insertelement <16 x i8> %vecinit2, i8 %vecext3, i32 2
  %vecext5 = extractelement <8 x i8> %x, i32 3
  %vecinit6 = insertelement <16 x i8> %vecinit4, i8 %vecext5, i32 3
  %vecext7 = extractelement <8 x i8> %x, i32 4
  %vecinit8 = insertelement <16 x i8> %vecinit6, i8 %vecext7, i32 4
  %vecext9 = extractelement <8 x i8> %x, i32 5
  %vecinit10 = insertelement <16 x i8> %vecinit8, i8 %vecext9, i32 5
  %vecext11 = extractelement <8 x i8> %x, i32 6
  %vecinit12 = insertelement <16 x i8> %vecinit10, i8 %vecext11, i32 6
  %vecext13 = extractelement <8 x i8> %x, i32 7
  %vecinit14 = insertelement <16 x i8> %vecinit12, i8 %vecext13, i32 7
  %vecinit30 = shufflevector <16 x i8> %vecinit14, <16 x i8> %y, <16 x i32> <i32 0, i32 1, i32 2, i32 3, i32 4, i32 5, i32 6, i32 7, i32 16, i32 17, i32 18, i32 19, i32 20, i32 21, i32 22, i32 2>
  ret <16 x i8> %vecinit30
}
```
Our outputted post-combiner IR is:
```
bb.1.entry:
  liveins: $d0, $q1
  %0:_(<8 x s8>) = COPY $d0
  %1:_(<16 x s8>) = COPY $q1
  %3:_(s64) = G_CONSTANT i64 0
  %2:_(s8) = G_EXTRACT_VECTOR_ELT %0:_(<8 x s8>), %3:_(s64)
  %7:_(s64) = G_CONSTANT i64 1
  %6:_(s8) = G_EXTRACT_VECTOR_ELT %0:_(<8 x s8>), %7:_(s64)
  %10:_(s64) = G_CONSTANT i64 2
  %9:_(s8) = G_EXTRACT_VECTOR_ELT %0:_(<8 x s8>), %10:_(s64)
  %13:_(s64) = G_CONSTANT i64 3
  %12:_(s8) = G_EXTRACT_VECTOR_ELT %0:_(<8 x s8>), %13:_(s64)
  %16:_(s64) = G_CONSTANT i64 4
  %15:_(s8) = G_EXTRACT_VECTOR_ELT %0:_(<8 x s8>), %16:_(s64)
  %19:_(s64) = G_CONSTANT i64 5
  %18:_(s8) = G_EXTRACT_VECTOR_ELT %0:_(<8 x s8>), %19:_(s64)
  %22:_(s64) = G_CONSTANT i64 6
  %21:_(s8) = G_EXTRACT_VECTOR_ELT %0:_(<8 x s8>), %22:_(s64)
  %25:_(s64) = G_CONSTANT i64 7
  %24:_(s8) = G_EXTRACT_VECTOR_ELT %0:_(<8 x s8>), %25:_(s64)
  %34:_(<8 x s8>) = G_BUILD_VECTOR %2:_(s8), %6:_(s8), %9:_(s8), %12:_(s8), %15:_(s8), %18:_(s8), %21:_(s8), %24:_(s8)
  %31:_(<8 x s8>), %32:_(<8 x s8>) = G_UNMERGE_VALUES %1:_(<16 x s8>)
  %27:_(<16 x s8>) = G_CONCAT_VECTORS %34:_(<8 x s8>), %31:_(<8 x s8>)
  $q0 = COPY %27:_(<16 x s8>)
  RET_ReallyLR implicit $q0
``` 

The IR is alright, it correctly puts the vector together, however there is a combiner missing that takes the extractions, turns it into a shufflevector and then optimizes it out.

llvm/test/CodeGen/AArch64/arm64-neon-copy.ll
```diff
@@ -2169,13 +2137,10 @@ define <4 x i32> @test_concat_v4i32_v2i32_v4i32(<2 x i32> %x, <4 x i32> %y) #0 {
...
-; CHECK-GI-NEXT:    // kill: def $d0 killed $d0 killed $q0_q1 def $q0_q1
-; CHECK-GI-NEXT:    adrp x8, .LCPI135_0
-; CHECK-GI-NEXT:    // kill: def $q1 killed $q1 killed $q0_q1 def $q0_q1
+; CHECK-GI-NEXT:    // kill: def $d0 killed $d0 def $q0
 ; CHECK-GI-NEXT:    mov s2, v0.s[1]
 ; CHECK-GI-NEXT:    mov v0.s[1], v2.s[0]
-; CHECK-GI-NEXT:    ldr q2, [x8, :lo12:.LCPI135_0]
-; CHECK-GI-NEXT:    tbl v0.16b, { v0.16b, v1.16b }, v2.16b
+; CHECK-GI-NEXT:    mov v0.d[1], v1.d[0]
 ; CHECK-GI-NEXT:    ret
```
ValentijnvdBeek, line 2144: Same as above

ValentijnvdBeek:
To explain the lateness of the reviews: ASUS managed to lose my hawkpoint laptop and have been promising that they will ship it soon (tm) for the past three weeks. I will see if I can get some progress done on my old 4 core

llvm/lib/CodeGen/GlobalISel/CombinerHelper.cpp
ValentijnvdBeek: Removed

llvm/lib/CodeGen/GlobalISel/CombinerHelper.cpp
ValentijnvdBeek: Not relevant now, better to merge it first, but a good addition later.

ValentijnvdBeek:
@konstantinschwarz @gbossu friendly review bump. Hopefully the tests don't fail (I forgot where the git-clang-format hook was again) :)

Changes from the last time:
- Split some changes up into different commits
- Fixed a few nits (the ones recently in a nit commit as I am bettering my life)
- Explained the assembly changes outside AIE2
- Rebased onto current `Main`

It is a bit late, so I hope I haven't missed anything. Good luck today!

ValentijnvdBeek:
FYI: there will be a conflict issue whenever LLVM main is merged into this branch due to ` CombinerHelper::tryCombineShuffleVector` being split up into two parts. To incorporate it into this branch, you simply put the stream matching code in the new `matchVectorMaskSequence` function and it should be fine.

The relevant PR is: https://github.com/llvm/llvm-project/pull/110545

@jsetoain the above PR is interesting for you since it sidesteps the need to convert into shufflevector explicitly. It will optimize sequences of inserts/extracts/buildvector automatically. You may want to [cherry-pick the original version](https://github.com/llvm/llvm-project/pull/110545/commits/1c9d534be76ce944d54c5f166c0c93b176187fe9) which turns into a `G_SHUFFLE_VECTOR` and then optimizes that rather than the current one that runs the optimizer directly.

Cheers!

konstantinschwarz:
Hi @ValentijnvdBeek, I finally came back to this, sorry for the long delay.
In https://github.com/Xilinx/llvm-aie/pull/224 I implemented an alternative combine for matching the G_SHUFFLE_VECTOR -> G_EXTRACT_SUBVECTOR (still G_UNMERGE_VALUES until the next upstream sync).

It uses the already available helper functions for creating contiguous shuffle masks. WDYT?

ValentijnvdBeek:
> Hi @ValentijnvdBeek, I finally came back to this, sorry for the long delay. In #224 I implemented an alternative combine for matching the G_SHUFFLE_VECTOR -> G_EXTRACT_SUBVECTOR (still G_UNMERGE_VALUES until the next upstream sync).
> 
> It uses the already available helper functions for creating contiguous shuffle masks. WDYT?

It is fine, I am just happy if the work doesn't totally bit rot. I like the linked code, for the most part, it takes my approach and makes it less of a tech demo (which this sort of is a bit). You do cover a part of the approach though, which admittedly the one that this implements, but there are some worries that I have about the general cost of this combiner in the future. It is pretty expensive and putting it in many different combiners does add a lot of code (and causes it to check the same values a lot). But as an approach within AIE, that is probably fine.

So as of this PR, what could be an idea is that I close it and you guys pick it over for those combiners. I will then take it back, be inspired by your approach and rewrite it with some of my crazier ideas to reduce the complexity so that I can push it upstream directly. Together with maybe https://github.com/Xilinx/llvm-aie/pull/129, that way it is also out of your hair and review todo list. Is that an idea @konstantinschwarz?

PR #46, opened 2024-05-21
Introduce zero overhead loop

martien-de-jong:
Final verdict: hwloop mostly causes significant PMsize expansion and frequent slight instruction count regressions (~ 10-100 cycles)

We have a few significant wins, e.g. GEMM_int8_1 InsnCount 55877 -> 53775
I didn't find any functional incorrectness and it is switched off by default. 

I would like to commit now, and propose a follow-up to take the loop size into account at e.g. legalization time, where it is relatively easy to allocate a virtual loopcount register. This on the basis that we don't gain on big loops, since there the loop body is dominated by memory, move and vector instructions.

This is a direct port of Abnikant's  original aie-private PR.
The representation of PseudoLoopEnd for analyzeBranch has changed significantly; We always push two components. The first is the opcode, the second the additional operand which can not be derived from the target block.

It should be noted that ZOL does not handle zero or negative loopcounts correctly. As such we need to establish that it is positive, e.g. by loop guarding or by interpreting pragma-like directives.


llvm/lib/Target/AIE/AIELegalizerInfo.cpp
```diff
@@ -1164,3 +1164,29 @@ bool AIELegalizerInfo::legalizeG_FADDSUB(LegalizerHelper &Helper,
   MI.eraseFromParent();
   return true;
 }
+
+bool AIELegalizerInfo::legalizeIntrinsic(LegalizerHelper &Helper,
+                                         MachineInstr &MI) const {
+  switch (MI.getIntrinsicID()) {
```
martien-de-jong: There's a fix in a later commit that should go here.

llvm/lib/Target/AIE/AIEBaseInstrInfo.cpp
```diff
@@ -78,7 +78,11 @@ void AIEBaseInstrInfo::insertNoop(MachineBasicBlock &MBB,
 
 bool AIEBaseInstrInfo::reverseBranchCondition(
     SmallVectorImpl<MachineOperand> &Cond) const {
-  assert((Cond.size() == 2) && "Invalid branch condition!");
+  assert((Cond.size() == 2 || Cond.size() == 3) && "Invalid branch condition!");
```
martien-de-jong: I think this is wrong. There's no opposite branchopcode for a hardware loop. Perhaps we can swap the branch targets, but it sounds iffy.

llvm/lib/Target/AIE/AIE2AsmPrinter.cpp
```diff
@@ -70,6 +70,39 @@ bool AIE2AsmPrinter::lowerOperand(const MachineOperand &MO,
   return LowerAIEMachineOperandToMCOperand(MO, MCOp, *this);
 }
 
+void AIE2AsmPrinter::emitInstruction(const MachineInstr *MI) {
+
+  // Insert a label before the instruction preceeding the PseudoLoopEnd.
+  if (MI->getOpcode() == AIE2::PseudoLoopEnd)
+    return;
```
gbossu: Comment says we insert something, but this is actually an early exit?

llvm/lib/Target/AIE/AIE2AsmPrinter.cpp
```diff
@@ -70,6 +70,39 @@ bool AIE2AsmPrinter::lowerOperand(const MachineOperand &MO,
...
+      }
+    }
+  }
+  if (MIPred == MI) {
+    MCSymbol *S = LoopEnd->getOperand(1).getMCSymbol();
+    OutStreamer->emitLabel(S);
+    OutStreamer->getCommentOS() << "Loop end."
+                                << "\n";
+  }
+  AIEBaseAsmPrinter::emitInstruction(MI);
```
gbossu: I'm a bit confused by this code. I think it tries to find a `PseudoLoopEnd`, and emit a label for the instruction before it.

Maybe change into something like:
```
 void AIE2AsmPrinter::emitInstruction(const MachineInstr *MI) {
  if (isPrecedingLoopEnd(MI)) {
    // Emit label and comment
  }
  AIEBaseAsmPrinter::emitInstruction(MI);
}
```

Also, I actually wonder if we should really trust the position of `PseudoLoopEnd`. I'm not sure it's guaranteed to really mark the end of the HW Loop with all the optimizations we run. Isn't the end of the loop always the end of the BB?

llvm/lib/Target/AIE/AIE2InstrInfo.cpp
```diff
@@ -940,7 +940,8 @@ bool AIE2InstrInfo::isLock(unsigned Opc) const {
 }
 
 bool AIE2InstrInfo::isDelayedSchedBarrier(const MachineInstr &MI) const {
-  return MI.getOpcode() == AIE2::DelayedSchedBarrier;
+  return MI.getOpcode() == AIE2::DelayedSchedBarrier ||
+         MI.getOpcode() == AIE2::PseudoLoopEnd;
```
gbossu: This would be better in the function below: `AIE2InstrInfo::isSchedBarrier`

llvm/lib/Target/AIE/AIE2InstructionSelector.cpp
```diff
@@ -214,6 +216,7 @@ class AIE2InstructionSelector : public InstructionSelector {
   // Select a control packet header stream write instruction from the
   // corresponding intrinsic.
   bool selectPutMS_CPH(MachineInstr &I, MachineRegisterInfo &MRI);
+  bool selectLoopDecrement(MachineInstr &I, MachineRegisterInfo &MRI);
```
gbossu: I'd be happy to see .mir ISel tests for `LoopDecrement`, `LoopEnd`, `loop_iterations`

llvm/lib/Target/AIE/AIE2InstructionSelector.cpp
```diff
@@ -959,9 +1001,8 @@ bool AIE2InstructionSelector::selectG_BRCOND(MachineInstr &I,
     Args->IntrinInst->eraseFromParent();
     return constrainSelectedInstRegOperands(*LoopDec, TII, TRI, RBI);
   }
-
-  // resort to TableGen'ed selection patterns
-  return selectImpl(I, *CoverageInfo);
+  // select loop_decrement or resort to TableGen'ed selection patterns
+  return selectLoopDecrement(I, MRI);
```
gbossu: I think something like `return selectLoopDecrement(I, MRI) || selectImpl(I, *CoverageInfo);` would be clearer

llvm/lib/Target/AIE/AIE2AsmPrinter.cpp
martien-de-jong: Good question, PseudoLoopEnd is a terminator that lists both successors. I would hope it stays in place.

ehh, not true. It has a successor and the last-bundle symbol. But it is still a terminator.

llvm/lib/Target/AIE/AIE2InstrInfo.cpp
martien-de-jong: Fully agree. I think we're looking at a lot of ad-hoc adaptations.

llvm/lib/Target/AIE/AIE2InstructionSelector.cpp
```diff
@@ -889,6 +897,40 @@ bool AIE2InstructionSelector::selectG_CONCAT_VECTORS(MachineInstr &I,
...
+  Register CondReg = MO.getReg();
+  MachineBasicBlock *DestMBB = I.getOperand(1).getMBB();
+  MachineBasicBlock *MBB = MO.getParent()->getParent();
+  MachineBasicBlock::instr_iterator it_begin = MBB->instr_begin();
+  MachineBasicBlock::instr_iterator it =
+      static_cast<MachineBasicBlock::instr_iterator>(I);
+  auto LoopDec = getDefIgnoringCopies(CondReg, MRI);
+  // Look for LoopDec Insn in the BB with G_BRCOND, the LoopDec is
+  // Selected else the loop break when it reaches first insn in the BB.
+  while (it != it_begin) {
```
martien-de-jong: @gbossu I don't understand this loop. Why not trust the getDefIgnoringCopies, do you see what is going on here?

llvm/lib/Target/AIE/AIE2InstructionSelector.cpp
martien-de-jong: And I don't like the fact that we feed a BRCOND as instruction into selectLoopDecrement, that's different from all other selectXXX functions.

llvm/lib/Target/AIE/AIEBaseInstrInfo.cpp
martien-de-jong: This should not be called on a PseudoLoopEnd, which is represented by Cond.size() == 3. This change can be dropped, sanity assert will come from reversBranchCondition() which doesn't recognize PseudoLoopEnd

llvm/lib/Target/AIE/AIE2InstructionSelector.cpp
martien-de-jong: Loop could be removed without breaking existing tests

llvm/lib/Target/AIE/AIE2AsmPrinter.cpp
martien-de-jong: PseudoLoopEnd is a schedbarrier.

llvm/lib/Target/AIE/AIE2InstructionSelector.cpp
```diff
@@ -897,38 +898,29 @@ bool AIE2InstructionSelector::selectG_CONCAT_VECTORS(MachineInstr &I,
...
+      LoopDec->getOpcode() != TargetOpcode::G_INTRINSIC_W_SIDE_EFFECTS) {
+    return false;
   }
-  // resort to TableGen'ed selection patterns
-  return selectImpl(I, *CoverageInfo);
+  auto *LoopDecIntrinsic = cast<GIntrinsic>(LoopDec);
+  if (LoopDecIntrinsic->getIntrinsicID() == Intrinsic::loop_decrement) {
+    MachineBasicBlock *DestMBB = BrCond.getOperand(1).getMBB();
+    MCContext &Context = BrCond.getParent()->getParent()->getContext();
+    MCSymbol *EndLabel = Context.createNamedTempSymbol("_112");
```
martien-de-jong: 112: this label should be 112 bytes from the loopstart. We are not actually using it for that purpose yet.

llvm/lib/Target/AIE/AIE2TargetTransformInfo.cpp
```diff
@@ -57,7 +57,23 @@ bool AIE2TTIImpl::isHardwareLoopProfitable(Loop *L, ScalarEvolution &SE,
...
+  // zero-overhead loop.
+  if (L->getNumBlocks() > 1)
+    return false;
+
+  // Scan the loop: loops with calls - make it unprofitable
+  for (BasicBlock *BB : L->blocks()) {
+    for (Instruction &I : *BB) {
+      if (isa<CallInst>(I) || isa<InvokeInst>(I)) {
+        if (const Function *F = cast<CallBase>(I).getCalledFunction()) {
+          if (!isLoweredToCall(F))
```
gbossu, line 70: Curious: Where does `isLoweredToCall` come from? Is that a generic LLVM function?

llvm/test/CodeGen/AIE/aie2/hardware-loops/irtranslator-zol.ll
```diff
@@ -0,0 +1,57 @@
+; NOTE: Assertions have been autogenerated by utils/update_mir_test_checks.py
+; RUN: llc -O2 -mtriple=aie2 -stop-after=irtranslator --enable-aie-hardware-loops --enable-aie-zero-overhead-loops %s -o - | FileCheck %s
+
```
gbossu, line 3: Nit: Add a small description to tell about the test and what is expected.

llvm/test/CodeGen/AIE/aie2/hardware-loops/nested.ll
```diff
@@ -17,31 +17,19 @@ define void @nested(ptr nocapture %out, ptr nocapture readonly %in, i32 noundef
...
-; CHECK-NEXT:    // =>This Loop Header: Depth=1
-; CHECK-NEXT:    // Child Loop BB0_2 Depth 2
-; CHECK-NEXT:    nopa ; lshl r5, r3, r4; nopm
-; CHECK-NEXT:    mov dj0, r5
-; CHECK-NEXT:    lda p4, [p1, dj0]
-; CHECK-NEXT:    nop
-; CHECK-NEXT:    nop
-; CHECK-NEXT:    nop
-; CHECK-NEXT:    mova r6, #0
-; CHECK-NEXT:    add.nc r5, r1, #-1
```
gbossu, line 37: Do you understand the changes?

llvm/lib/Target/AIE/AIE2TargetTransformInfo.cpp
martien-de-jong: It's generic, with a note that it should be moved to a target-specific hook.

llvm/lib/Target/AIE/AIELegalizerInfo.cpp
```diff
@@ -1164,3 +1165,29 @@ bool AIELegalizerInfo::legalizeG_FADDSUB(LegalizerHelper &Helper,
   MI.eraseFromParent();
   return true;
 }
+
+bool AIELegalizerInfo::legalizeIntrinsic(LegalizerHelper &Helper,
+                                         MachineInstr &MI) const {
+  switch (cast<GIntrinsic>(MI).getIntrinsicID()) {
```
gbossu, line 1351: Nit: Maybe leave a comment to say that other AIE-specific intrinsics are already created with legal operand types, and therefore do not need to be legalized here.

llvm/test/CodeGen/AIE/aie2/hardware-loops/legalize-zol.mir
```diff
@@ -38,9 +38,7 @@ body:             |
   ; CHECK-NEXT:   [[ADD1:%[0-9]+]]:_(s32) = nuw nsw G_ADD [[PHI1]], [[C]]
   ; CHECK-NEXT:   [[INT:%[0-9]+]]:_(s32) = G_INTRINSIC_W_SIDE_EFFECTS intrinsic(@llvm.loop.decrement), [[C]](s32)
   ; CHECK-NEXT:   [[ASSERT_ZEXT:%[0-9]+]]:_(s32) = G_ASSERT_ZEXT [[INT]], 1
-  ; CHECK-NEXT:   [[C3:%[0-9]+]]:_(s32) = G_CONSTANT i32 1
-  ; CHECK-NEXT:   [[AND:%[0-9]+]]:_(s32) = G_AND [[ASSERT_ZEXT]], [[C3]]
-  ; CHECK-NEXT:   G_BRCOND [[AND]](s32), %bb.2
+  ; CHECK-NEXT:   G_BRCOND [[ASSERT_ZEXT]](s32), %bb.2
```
gbossu, line 41: You did not change the legalizer, do you know why that test needed to be updated?
~Edit: I didn't see it's not your change. But I think it still makes sense to move that diff to the first commit if that's where it belongs.~

llvm/lib/Target/AIE/AIE2InstructionSelector.cpp
gbossu: Should this commit maybe be squashed with [[AIE2] Clean up matching of ZOL End and add test](https://github.com/Xilinx/llvm-aie/pull/46/commits/a37cc925d1c2900944f875521ee80f0cb43939eb)?

llvm/lib/Target/AIE/AIE2AsmPrinter.cpp
```diff
@@ -70,6 +70,34 @@ bool AIE2AsmPrinter::lowerOperand(const MachineOperand &MO,
...
+      break;
+    }
+    Prev = &MII;
+  }
+
+  if (MI == LastBundle) {
+    MCSymbol *S = LoopEnd->getOperand(1).getMCSymbol();
+    OutStreamer->emitLabel(S);
+    OutStreamer->getCommentOS() << "Loop end.\n";
+  }
```
gbossu: Could you create a test for that change? Even if it's an end-to-end test, we could still use one as a baseline, and see that there is now a label.

llvm/test/CodeGen/AIE/aie2/hardware-loops/nested.ll
martien-de-jong: As far as I can see, profitability of low overhead loops was reduced to single block loops. I removed the issue-limit=1, which probably wasn't very clever for this particular test.

llvm/lib/Target/AIE/AIE2AsmPrinter.cpp
```diff
@@ -70,6 +70,34 @@ bool AIE2AsmPrinter::lowerOperand(const MachineOperand &MO,
   return LowerAIEMachineOperandToMCOperand(MO, MCOp, *this);
 }
 
+void AIE2AsmPrinter::emitInstruction(const MachineInstr *MI) {
+  // PseudoLoopEnd is just here to carry some extra information. No need to
+  // emit anything
+  if (MI->getOpcode() == AIE2::PseudoLoopEnd)
```
gbossu: Could we maybe just declare `PseudoLoopEnd` as a meta instruction to avoid the special case here?

llvm/lib/Target/AIE/AIE2RegisterInfo.cpp
```diff
@@ -108,6 +108,11 @@ BitVector AIE2RegisterInfo::getReservedRegs(const MachineFunction &MF) const {
   ReserveRegs(AIE2::eRRegClass, ReservedGPRs);
   ReserveRegs(AIE2::eDRegClass, ReservedMODs);
 
+  // Mark the hardware loop related register as reserved,
+  // otherwise they are considered dead
+  markSuperRegs(Reserved, AIE2::LC);
+  markSuperRegs(Reserved, AIE2::LS);
+  markSuperRegs(Reserved, AIE2::LE);
```
gbossu, line 118: That means we were deleting copies to those registers?

llvm/lib/Target/AIE/AIEBaseHardwareLoops.cpp
```diff
@@ -164,8 +169,15 @@ bool AIEBaseHardwareLoops::ProcessLoop(MachineLoop *ML) {
   for (MachineLoop *L : *ML)
     Changed |= ProcessLoop(L);
 
-  LowOverheadLoop LoLoop(*ML, *MLI, *TRI, *TII);
+  MachineBasicBlock *LastMBB = ML->findLoopControlBlock();
+  // More than one exit, don't generate ZOL.
```
gbossu, line 205: Can that happen?

llvm/lib/Target/AIE/AIE2RegisterInfo.cpp
martien-de-jong: It looks that way. Didn't check, but stands to reason.

llvm/lib/Target/AIE/AIEBaseHardwareLoops.cpp
martien-de-jong: That is probably a remote possibility involving introducing a more-than-two way branch.

llvm/lib/Target/AIE/AIE2TargetTransformInfo.cpp
andcarminati: I think we should have a custom implementation for this PR. See: 

```
void sum(double *a, double *b, double *c) {
    for(int i = 0; i < 30; i++) {
       c[i] = a[i] + b[i];
    }
}
```

It should walk in sync with GISel legalization rules for libcalls.

llvm/test/CodeGen/AIE/aie2/hardware-loops/legalize-zol.mir
martien-de-jong: I have assumed that the bit analysis has improved in upstream llvm, and now recognises that booleans are inrange.

andcarminati:
Hi, the following code can cause some problem in this PR:

```
target datalayout = "e-m:e-p:20:32-i1:8:32-i8:8:32-i16:16:32-i32:32:32-f32:32:32-i64:32-f64:32-a:0:32-n32"
target triple = "aie2"

define i32 @main() {
entry:
  br label %for.body

for.body:                                  ; preds = %for.body, %entry
  %i = phi i16 [ 0, %entry ], [ 1, %for.body ]
  %cmp = icmp ult i16 %i, 1
  br i1 %cmp, label %for.body, label %if.end

if.end:                                     ; preds = %for.body
  ret i32 0
}
```

With:

```
llc --march=aie2 sample.ll  --enable-aie-hardware-loops  --enable-aie-zero-overhead-loops
```

Gives: 

```
llc: ../llvm/include/llvm/ADT/ilist_iterator.h:138: llvm::ilist_iterator::reference llvm::ilist_iterator<llvm::ilist_detail::node_options<llvm::MachineInstr, true, true, void, false>, false, false>::operator*() const [OptionsT = llvm::ilist_detail::node_options<llvm::MachineInstr, true, true, void, false>, IsReverse = false, IsConst = false]: Assertion `!NodePtr->isKnownSentinel()' failed.
PLEASE submit a bug report to https://github.com/llvm/llvm-project/issues/ and include the crash backtrace.
Stack dump:
0.      Program arguments: ./bin/llc --march=aie2 reduced.ll --enable-aie-hardware-loops --enable-aie-zero-overhead-loops
1.      Running pass 'Function Pass Manager' on module 'reduced.ll'.
2.      Running pass 'InstructionSelect' on function '@main'
 #0 0x0000000005d38f97 llvm::sys::PrintStackTrace(llvm::raw_ostream&, int) ./../llvm/lib/Support/Unix/Signals.inc:723:13
 #1 0x0000000005d36f70 llvm::sys::RunSignalHandlers() ./../llvm/lib/Support/Signals.cpp:106:18
 #2 0x0000000005d3964a SignalHandler(int) ./../llvm/lib/Support/Unix/Signals.inc:413:1
 #3 0x00007f6f6c1a5520 (/lib/x86_64-linux-gnu/libc.so.6+0x42520)
 #4 0x00007f6f6c1f99fc __pthread_kill_implementation ./nptl/pthread_kill.c:44:76
 #5 0x00007f6f6c1f99fc __pthread_kill_internal ./nptl/pthread_kill.c:78:10
 #6 0x00007f6f6c1f99fc pthread_kill ./nptl/pthread_kill.c:89:10
 #7 0x00007f6f6c1a5476 gsignal ./signal/../sysdeps/posix/raise.c:27:6
 #8 0x00007f6f6c18b7f3 abort ./stdlib/abort.c:81:7
 #9 0x00007f6f6c18b71b _nl_load_domain ./intl/loadmsgcat.c:1177:9
#10 0x00007f6f6c19ce96 (/lib/x86_64-linux-gnu/libc.so.6+0x39e96)
```

llvm/lib/Target/AIE/AIE2InstructionSelector.cpp
```diff
@@ -889,6 +898,31 @@ bool AIE2InstructionSelector::selectG_CONCAT_VECTORS(MachineInstr &I,
...
+      LoopDec->getOpcode() != TargetOpcode::G_INTRINSIC_W_SIDE_EFFECTS) {
+    return false;
+  }
+  auto *LoopDecIntrinsic = cast<GIntrinsic>(LoopDec);
+  if (LoopDecIntrinsic->getIntrinsicID() == Intrinsic::loop_decrement) {
+    MachineBasicBlock *DestMBB = BrCond.getOperand(1).getMBB();
+    MCContext &Context = BrCond.getParent()->getParent()->getContext();
+    MCSymbol *EndLabel = Context.createNamedTempSymbol("_112");
+    MIB.buildInstr(AIE2::PseudoLoopEnd).addMBB(DestMBB).addSym(EndLabel);
+    LoopDec->eraseFromParent();
```
andcarminati: After the second erase, iterator got corrupted.

andcarminati:
Hi @martien-de-jong , another interesting case (sample.ll):

```
target datalayout = "e-m:e-p:20:32-i1:8:32-i8:8:32-i16:16:32-i32:32:32-f32:32:32-i64:32-f64:32-a:0:32-n32"
target triple = "aie2"

define i32 @main() {
entry:
  br label %for.body

for.body:                                     ; preds = %for.body, %entry
  %i = phi i32 [ 0, %entry ], [ %inc, %for.body ]
  %inc = add i32 %i, 1
  %exitcond = icmp eq i32 %inc, 0
  br i1 %exitcond, label %label, label %for.body

label:                                     ; preds = %label, %for.body
  br label %label
}
```

However, you need the following options (Elf emission, no loop scheduling):

```
llc --march=aie2 sample.ll  --enable-aie-hardware-loops  --enable-aie-zero-overhead-loops -aie-loop-aware=false -filetype=obj
```

Result:
```
<unknown>:0: error: Undefined temporary symbol .L_1120
```

llvm/lib/Target/AIE/AIEFinalizeBundle.cpp
```diff
@@ -20,8 +20,17 @@ using namespace llvm;
...
+  const MachineBasicBlock *MBB = MII->getParent();
+  auto *MF = MBB->getParent();
+  auto &Subtarget = MF->getSubtarget();
+  const AIEBaseInstrInfo *TII =
+      static_cast<const AIEBaseInstrInfo *>(Subtarget.getInstrInfo());
   MachineInstr *MI = &*MII;
-  if (MI->isMetaInstruction() || MII->isBundled())
+  // LoopStart should be lowered before reaching here
+  assert(!TII->isHardwareLoopStart(MI->getOpcode()) &&
+         "LoopStart is not lowered \n\r");
```
andcarminati, line 31: Hi @martien-de-jong,

I can trigger this assert with the following code:

```
target datalayout = "e-m:e-p:20:32-i1:8:32-i8:8:32-i16:16:32-i32:32:32-f32:32:32-i64:32-f64:32-a:0:32-n32"
target triple = "aie2"

define i32 @main(i1 %cmp) {
entry:
  %call = tail call i32 null(ptr null, ptr null)
  br i1 %cmp, label %if.else, label %for.cond.preheader

for.cond.preheader:                          ; preds = %entry
  store i8 0, ptr null, align 4
  br label %for.body

if.else:                                      ; preds = %entry
  %bf = load i16, ptr null, align 4
  br label %for.body

exit:                        ; preds = %for.body
  ret i32 0

for.body:                                  ; preds = %for.body, %if.else, %for.cond.preheader
  %j = phi i32 [ %inc, %for.body ], [ 0, %if.else ], [ 0, %for.cond.preheader ]
  %0 = trunc i32 %j to i20
  %inc = add i32 %j, 1
  %exitcond = icmp eq i32 %inc, 0
  br i1 %exitcond, label %exit, label %for.body
}

```

llvm/lib/Target/AIE/AIEBaseHardwareLoops.cpp
```diff
@@ -206,48 +279,73 @@ bool AIEBaseHardwareLoops::ProcessLoop(MachineLoop *ML) {
...
   LLVM_DEBUG(dbgs() << "AIE Loops: Expanding LoopStart.\n");
 
   if (LoLoop.isJNZDLoop()) {
     LLVM_DEBUG(dbgs() << "AIE Loops: JNZD loop. Nothing to be done.\n");
     return;
   }
 
-  // Future: Handle true hardware loop setup code
-  llvm_unreachable("Implement true hardware loop support");
+  if (LoLoop.isZeroOverheadLoop()) {
```
andcarminati: Do we really need to test this condition? Is there any situation where this if will evaluate to false?

llvm/lib/Target/AIE/AIEBaseInstrInfo.cpp
```diff
@@ -267,17 +276,20 @@ unsigned AIEBaseInstrInfo::insertBranch(
...
-    return 1;
-  }
-
-  // Either a one or two-way conditional branch.
+  // Cond[0] contains the branch opcode (JZ, JNZ, PseudoLoopEnd)
+  // Cond[1] contains opcode dependent information, the last-bundle symbol for
+  // PseudoLoopEnd, the condition register for JZ/JNZ
   unsigned Opc = Cond[0].getImm();
-  MachineInstr &CondMI = *BuildMI(&MBB, DL, get(Opc)).add(Cond[1]).addMBB(TBB);
+  MachineInstrBuilder CBB = BuildMI(&MBB, DL, get(Opc));
```
andcarminati: As we already have TBB, FBB as BB, CBB seems to be a BB as well because of the name. Maybe a different name is enough to state that this is an instr.

llvm/lib/Target/AIE/AIEBaseHardwareLoops.cpp
martien-de-jong: I'll check. Modify it to an assert if it can be.

llvm/lib/Target/AIE/AIEFinalizeBundle.cpp
martien-de-jong: I'm on it. Probably an assumption on seeing LoopStart before LoopEnd.

llvm/lib/Target/AIE/AIE2AsmPrinter.cpp
martien-de-jong: That worked, after reordering some condition logic.

llvm/test/CodeGen/AIE/aie2/hardware-loops/nested.ll
martien-de-jong: have restored issue-limit=1.

llvm/lib/Target/AIE/AIEFinalizeBundle.cpp
martien-de-jong: The true story: tail duplication copied LoopStart out into predecessor blocks. This scares hwloop expansion.

llvm/lib/Target/AIE/AIE2InstrInfo.td
```diff
@@ -557,7 +557,7 @@ def LoopJNZ :
     Pseudo<(outs), (ins eR:$tc, eP:$target), "loop_jnz", "${tc}, ${target}">;
 
 // Zero overhead loops
-let hasSideEffects = 1, mayLoad = 0, mayStore = 0 in
+let hasSideEffects = true, mayLoad = 0, mayStore = 0, isNotDuplicable = true in
```
gbossu: Super nit: While you're at it, also change the `0`s to `false`?

llvm/test/CodeGen/AIE/aie2/hardware-loops/legalize-zol.mir
```diff
@@ -0,0 +1,76 @@
...
+    successors: %bb.1(0x80000000)
+    liveins: $p0, $p1, $r0
+
+    %0:_(p0) = COPY $p0
+    %1:_(p0) = COPY $p1
+    %2:_(s32) = COPY $r0
+    %13:_(s32) = G_CONSTANT i32 1
+    %16:_(s32) = G_CONSTANT i32 0
+    %3:_(s32) = G_LOAD %0(p0) :: (load (s32))
+    G_INTRINSIC_W_SIDE_EFFECTS intrinsic(@llvm.set.loop.iterations), %2(s32)
```
gbossu, line 53: Btw what are the semantics of `llvm.set.loop.iterations`? Is it always assumed to sit in a BB with a single successor? And in this case this gives the number of iterations for that successor?

llvm/lib/Target/AIE/AIE2InstructionSelector.cpp
```diff
@@ -214,6 +215,8 @@ class AIE2InstructionSelector : public InstructionSelector {
   // Select a control packet header stream write instruction from the
   // corresponding intrinsic.
   bool selectPutMS_CPH(MachineInstr &I, MachineRegisterInfo &MRI);
+  bool selectBrCondLoopDecrement(MachineInstr &BrCond,
+                                 MachineRegisterInfo &MRI);
```
gbossu, line 218: Nit: Commit says it's cleaning something up, but I think it's only adding things now.

llvm/lib/Target/AIE/AIE2InstructionSelector.cpp
```diff
@@ -889,6 +897,31 @@ bool AIE2InstructionSelector::selectG_CONCAT_VECTORS(MachineInstr &I,
   return true;
 }
 
+// Try to match BRCOND(Intrinsic::loop_decrement)
+bool AIE2InstructionSelector::selectBrCondLoopDecrement(
+    MachineInstr &BrCond, MachineRegisterInfo &MRI) {
+  MachineOperand &MO = BrCond.getOperand(0);
```
gbossu, line 845: Super-nit: Maybe have an assert to verify the `BrCond.getOpcode() == G_BRCOND` precoindition?

llvm/lib/Target/AIE/AIE2InstructionSelector.cpp
```diff
@@ -889,6 +897,31 @@ bool AIE2InstructionSelector::selectG_CONCAT_VECTORS(MachineInstr &I,
...
+  Register CondReg = MO.getReg();
+
+  // Check if the condition is a LoopDecrement
+  auto *LoopDec = getDefIgnoringCopies(CondReg, MRI);
```
gbossu, line 849: Can it actually happen that `LoopDec == nullptr`?

llvm/test/CodeGen/AIE/aie2/hardware-loops/instruction-select-zol.mir
```diff
@@ -0,0 +1,73 @@
...
+    %22:gprregbank(s32) = G_SHL %8, %23(s32)
+    %12:modregbank(s20) = G_TRUNC %22(s32)
+    %15:gprregbank(s32) = G_AIE_OFFSET_LOAD %1(p0), %12(s20) :: (load (s32))
+    %7:gprregbank(s32) = nsw G_ADD %6, %15
+    G_STORE %7(s32), %0(p0) :: (store (s32))
+    %9:gprregbank(s32) = nuw nsw G_ADD %8, %3
+    %19:gprregbank(s32) = G_INTRINSIC_W_SIDE_EFFECTS intrinsic(@llvm.loop.decrement), %3(s32)
+    %20:gprregbank(s32) = G_ASSERT_ZEXT %19, 1
+    G_BRCOND %20(s32), %bb.2
+    G_BR %bb.1
```
gbossu: I think that test could be simplified/split to really only focus on:
 - Isel for `llvm.set.loop.iterations`
 - ISel for `llvm.loop.decrement`

llvm/lib/Target/AIE/AIE2AsmPrinter.cpp
```diff
@@ -70,6 +70,34 @@ bool AIE2AsmPrinter::lowerOperand(const MachineOperand &MO,
...
+    return;
+
+  // Check whether there's a PseudoLoopEnd following. If so, emit the
+  // last-bundle-label that it carries to designate this MI as the last bundle.
```
gbossu: Note: The AsmPrinter has a notion of "post-instruction symbol", and it can trigger the emission of a symbol after an instruction is emitted. See:
```
      // If there is a post-instruction symbol, emit a label for it here.
      if (MCSymbol *S = MI.getPostInstrSymbol())
        OutStreamer->emitLabel(S);
```

Maybe we could hook into this mechanism?

gbossu: And even:
```
      // If there is a pre-instruction symbol, emit a label for it here.
      if (MCSymbol *S = MI.getPreInstrSymbol())
        OutStreamer->emitLabel(S);
```
before emitting the instruction

llvm/lib/Target/AIE/AIE2AsmPrinter.cpp
```diff
@@ -70,6 +70,34 @@ bool AIE2AsmPrinter::lowerOperand(const MachineOperand &MO,
...
+  const MachineInstr *LoopEnd = nullptr;
+  const MachineInstr *Prev = nullptr;
+  for (auto &MII : *MI->getParent()) {
+    if (MII.getOpcode() == AIE2::PseudoLoopEnd) {
+      LoopEnd = &MII;
+      LastBundle = Prev;
+      break;
+    }
+    Prev = &MII;
+  }
```
gbossu: I think that analysis could be done a single time before emitting any instruction is called. If we can't use pre/post emission symbols, maybe move that analysis into e.g. `AIEBaseAsmPrinter::emitFunctionBodyStart`?

llvm/test/CodeGen/AIE/aie2/hardware-loops/loop-with-call.ll
```diff
@@ -0,0 +1,132 @@
+; NOTE: Assertions have been autogenerated by utils/update_llc_test_checks.py
+
+; This file is licensed under the Apache License v2.0 with LLVM Exceptions.
+; See https://llvm.org/LICENSE.txt for license information.
+; SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+;
+; (c) Copyright 2024 Advanced Micro Devices, Inc. or its affiliates
+
+; RUN: llc -O2 -mtriple=aie2 --enable-aie-hardware-loops \
+; RUN:   --enable-aie-zero-overhead-loops %s -o - | FileCheck %s
```
gbossu, line 10: I'd be happy to have those tests added earlier as baseline, so we can see how they get updated with each commit.

llvm/lib/Target/AIE/AIEFinalizeBundle.cpp
```diff
@@ -20,8 +20,17 @@ using namespace llvm;
...
+  if (MI->isMetaInstruction() || MII->isBundled() ||
+      TII->isHardwareLoopEnd(MI->getOpcode()))
```
gbossu, line 33: Can you add a test for the alignment after LoopStart?

llvm/lib/Target/AIE/AIE2InstrInfo.td
martien-de-jong: Yes!

llvm/test/CodeGen/AIE/aie2/hardware-loops/legalize-zol.mir
martien-de-jong: It is not duplicable, and it is placed in the preheader of the loop. I don't find much specified about nesting, or 1-1 use of setup and decrement. I assume that a set.loop.iterations will effectively push a loop level, decrement will maintain that loop level until the iterations run out, then it will pop. I think AIE could handle multiple set.loop.iterations for the same loop, but needs a single endloop.

llvm/lib/Target/AIE/AIE2InstructionSelector.cpp
martien-de-jong: Will fix.

llvm/lib/Target/AIE/AIE2AsmPrinter.cpp
martien-de-jong: Sounds attractive.

llvm/lib/Target/AIE/AIEFinalizeBundle.cpp
martien-de-jong: Perhaps, but we don't need alignment. we piggy-back padding onto the alignment mechanism.

llvm/lib/Target/AIE/AIE2InstructionSelector.cpp
martien-de-jong: I think it's very unlikely, CondReg would need to have an invalid type.

llvm/test/CodeGen/AIE/aie2/hardware-loops/instruction-select-zol.mir
martien-de-jong: ok.

llvm/lib/Target/AIE/AIE2AsmPrinter.cpp
martien-de-jong: Yep. Works like a charm.

llvm/lib/Target/AIE/AIEFinalizeBundle.cpp
martien-de-jong: I think there is already a test for the padding added in the bundle alignment PR. But I will check.

martien-de-jong: Indeed, in CodeGen/AIE/aie2/elongate/*

llvm/lib/Target/AIE/AIEBaseAsmPrinter.cpp
```diff
@@ -61,14 +61,32 @@ void AIEBaseAsmPrinter::emitFunctionBodyStart() {
...
       for (ConstMIBundleOperands OP(MI); OP.isValid(); ++OP) {
         if (OP->isMBB()) {
           FunctionTakenMBBAddresses.insert(OP->getMBB());
         }
       }
     }
+
+    // Scan for endloops to propagate the last-bundle label.
+    // The loop is separated from the previous one because it needs a bundle
+    // iterator instead of an instruction iterator.
```
gbossu, line 77: That turned out pretty nice!

llvm/test/CodeGen/AIE/aie2/hardware-loops/endlabel.mir
```diff
@@ -0,0 +1,34 @@
+# NOTE: Assertions have been autogenerated by utils/update_mir_test_checks.py
```
gbossu, line 1: Liar 😏

llvm/test/CodeGen/AIE/aie2/hardware-loops/noduplication.mir
```diff
@@ -0,0 +1,152 @@
+# NOTE: Assertions have been autogenerated by utils/update_mir_test_checks.py UTC_ARGS: --version 4
+# RUN: llc -mtriple=aie2 --start-after=instruction-select \
+# RUN:   --stop-before=aie-finalize-mi-bundles %s -o - | FileCheck %s
```
gbossu: Super-Nit: I'm changing the scheduler so it outputs "correct" Bundles. Could you change that line to `--stop-after=aie-finalize-mi-bundles`? This way there won't be test updates.

llvm/test/CodeGen/AIE/aie2/hardware-loops/complex-flow.mir
```diff
@@ -0,0 +1,205 @@
...
+# SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+#
+# (c) Copyright 2024 Advanced Micro Devices, Inc. or its affiliates
+
+# RUN: llc --mtriple=aie2 -O2 %s --run-pass=block-placement \
+# RUN:   -verify-machineinstrs -o - | FileCheck %s
+
+# This stresses block placement to a point where a ZOL doesn't fall-through
+# to its exit point. This gives rise to insertBranch of a PseudoLoopEnd with
+# a non-null FBB, which should generate an unconditional branch
```
gbossu, line 14: Nit: I'd be happy with an ascii-art CFG :)

llvm/lib/Target/AIE/AIEBaseInstrInfo.cpp
```diff
@@ -172,8 +180,11 @@ bool AIEBaseInstrInfo::analyzeBranch(MachineBasicBlock &MBB,
   if (NumTerminators == 1 && I->getDesc().isConditionalBranch()) {
     if (isHardwareLoopEnd(I->getOpcode())) {
       TBB = I->getOperand(0).getMBB();
+      // The machine verifier expects a non-empty Cond array, and we need to
+      // supply enough information to reconstruct this instruction.
+      // The basic block target is provided in the interfaces so we don't need
+      // to push that
       Cond.push_back(MachineOperand::CreateImm(I->getOpcode()));
-      Cond.push_back(I->getOperand(0));
```
gbossu, line 176: Did that hurt to keep `Cond.push_back(I->getOperand(0));`? I'm still struggling to understand what kind of API `analyzeBranch` has.

llvm/lib/Target/AIE/AIEBaseInstrInfo.cpp
```diff
@@ -269,17 +278,20 @@ unsigned AIEBaseInstrInfo::insertBranch(
...
+  // Cond[0] contains the branch opcode (JZ, JNZ, PseudoLoopEnd)
+  // Cond[1] contains opcode dependent information, the last-bundle symbol for
+  // PseudoLoopEnd, the condition register for JZ/JNZ
   unsigned Opc = Cond[0].getImm();
-  MachineInstr &CondMI = *BuildMI(&MBB, DL, get(Opc)).add(Cond[1]).addMBB(TBB);
+  MachineInstrBuilder CBranchBuilder = BuildMI(&MBB, DL, get(Opc));
+  if (isHardwareLoopEnd(Opc)) {
+    CBranchBuilder.addMBB(TBB).add(Cond[1]);
+  } else {
+    CBranchBuilder.add(Cond[1]).addMBB(TBB);
```
gbossu: Super-nit: Maybe we could define `PseudoLoopEnd` to have the same operand order as other branches?

llvm/lib/Target/AIE/AIEBaseInstrInfo.cpp
```diff
@@ -184,18 +195,18 @@ bool AIEBaseInstrInfo::analyzeBranch(MachineBasicBlock &MBB,
...
+      Cond.push_back(Prev->getOperand(1));
+    } else {
       // JNZD
-      if (!(std::prev(I)->getOperand(1).isMBB()))
+      if (!Prev->getOperand(1).isMBB())
         return Unhandled;
-      else
-        parseCondBranch(*std::prev(I), TBB, Cond);
+      parseCondBranch(*Prev, TBB, Cond);
+    }
```
gbossu, line 199: Nit: maybe let `parseCondBranch` determine whether it can actually parse the branch? E.g.
```
else if (!parseCondBranch(*Prev, TBB, Cond)) {
  return Unhandled;
}
```

llvm/test/CodeGen/AIE/aie2/hardware-loops/zol-loop.mir
```diff
@@ -0,0 +1,82 @@
+# NOTE: Assertions have been autogenerated by utils/update_mir_test_checks.py
+
+# This file is licensed under the Apache License v2.0 with LLVM Exceptions.
+# See https://llvm.org/LICENSE.txt for license information.
+# SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+#
+# (c) Copyright 2024 Advanced Micro Devices, Inc. or its affiliates
+
+# RUN: llc -mtriple aie2 -run-pass=aie-hardware-loops \
```
gbossu, line 9: Can you move that test to the previous commit where you add the code for `LoopStart` expansion?

llvm/lib/Target/AIE/AIEFinalizeBundle.cpp
gbossu: But is there one specifically for HW Loops and the handling of "loop setup instructions"?

gbossu (commented):
I'm done with the review. I think it looks good! Please go through the remaining comments, I'd be happy if some tests are moved around, but it's not such a big deal :)

gbossu (dismissed):

llvm/lib/Target/AIE/AIEFinalizeBundle.cpp
martien-de-jong: yes

llvm/lib/Target/AIE/AIEBaseInstrInfo.cpp
martien-de-jong: Deal!

llvm/lib/Target/AIE/AIEBaseInstrInfo.cpp
martien-de-jong: The api is that the target can push whatever it needs to reconstruct/invert a branch. We don't actually need that third operand, and I like occam's razor.

andcarminati (dismissed):
LGTM. Nice work, credits to you and the origin author.

llvm/lib/Target/AIE/AIEBaseAsmPrinter.cpp
martien-de-jong: Yes, sweet and lovely.

gbossu (approved):

andcarminati (approved):

llvm/test/CodeGen/AIE/aie2/hardware-loops/instruction-select-zol-end.mir
```diff
@@ -0,0 +1,27 @@
+# NOTE: Assertions have been autogenerated by utils/update_mir_test_checks.py
+# RUN: llc -O2 -mtriple=aie2 -run-pass=instruction-select %s -verify-machineinstrs -o - | FileCheck %s
+
+---
```
konstantinschwarz, line 4: A couple of tests here are missing the license & copyright header. Could you please add those in a fixup PR @martien-de-jong?

PR #47, opened 2024-05-22
Startup code

mchokshi091101:
- Rebased Startup Code PR 
- Added _Exit() function in _crt1.cc_

### DIFF
--- a/libc/startup/baremetal/CMakeLists.txt
+++ b/libc/startup/baremetal/CMakeLists.txt
@@ -1,0 +1,75 @@
+#
+# This file is licensed under the Apache License v2.0 with LLVM Exceptions.
+# See https://llvm.org/LICENSE.txt for license information.
+# SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+#
+# (c) Copyright 2024 Advanced Micro Devices, Inc. or its affiliates
+function(add_startup_object name)
+  cmake_parse_arguments(
+    "ADD_STARTUP_OBJECT"
+    "ALIAS" # Option argument
+    "SRC"   # Single value arguments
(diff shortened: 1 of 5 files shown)


libc/startup/baremetal/CMakeLists.txt
```diff
@@ -0,0 +1,75 @@
...
+else()
+  set(LIBC_INSTALL_LIBRARY_DIR lib${LLVM_LIBDIR_SUFFIX})
+endif()
+
+add_custom_target(libc-startup)
+set(startup_components crt0 crt1)
+foreach(target IN LISTS startup_components)
+  set(fq_target_name libc.startup.baremetal.${target})
+  add_dependencies(libc-startup ${fq_target_name})
+  install(FILES $<TARGET_OBJECTS:${fq_target_name}>
```
konstantinschwarz, line 71: Until we update to the most recent libc build system, we need to add the following to fix the build & install targets of the startup files:

```
diff --git a/libc/lib/CMakeLists.txt b/libc/lib/CMakeLists.txt
index 652ebc9ffb46..cffe47dbf931 100644
--- a/libc/lib/CMakeLists.txt
+++ b/libc/lib/CMakeLists.txt
@@ -41,6 +41,9 @@ foreach(archive IN ZIP_LISTS
   # with custom targets and in older version of cmake (~pre-3.20, apparently)
   # there is not good support for describing install dependencies on custom targets.
   add_dependencies(${archive_1} libc-headers)
+  if(TARGET libc-startup)
+    add_dependencies(${archive_1} libc-startup)
+  endif()
 
   set_target_properties(
     ${archive_1}
```

konstantinschwarz (approved):
Please squash the commits during rebase&merge

PR #48, opened 2024-05-22
[AIE2] Rewrite G_CONCAT_VECTOR selection as table-gen

ValentijnvdBeek:
Rewrites the selection code for the `G_CONCAT_VECTORS` opcode so that is done in `tablegen` instead of the C code.


llvm/lib/Target/AIE/AIE2InstrPatterns.td
```diff
@@ -961,6 +961,22 @@ def : Concat256<int_aie2_concat_512_256_acc, ACC512, ACC256>;
 def : Concat512<int_aie2_concat_1024_512_acc, ACC1024, ACC512>;
 def : Concat1024<int_aie2_concat_1024_256_acc, ACC1024, ACC256>;
 
+// concat_vector generic opcode
+def : Pat<(v16i32 (concat_vectors (v8i32 VEC256:$src0), (v8i32 VEC256:$src1))),
+  (v16i32 (REG_SEQUENCE VEC512, VEC256:$src0, sub_256_lo, VEC256:$src1, sub_256_hi))>;
+def : Pat<(v32i32 (concat_vectors (v16i32 VEC512:$src0), (v16i32 VEC512:$src1))),
+  (v32i32 (REG_SEQUENCE VEC1024, VEC512:$src0, sub_512_lo, VEC512:$src1, sub_512_hi))>;
```
gbossu, line 968: TIL `G_CONCAT_VECTOR` exists. Where are such opcodes created, is this from a generic combiner that identifies a certain `G_SHUFFLE_VECTOR` pattern?

ValentijnvdBeek: `G_CONCAT_VECTOR` is created by the legalizer whenever it needs to combine two vectors together. So, for example, if you ask for 1024-bit vector you need to build 2 512-bit vectors and concat them together. There is also a combiner for it, which I enable in #41.

llvm/lib/Target/AIE/AIE2InstrPatterns.td
```diff
@@ -961,6 +961,22 @@ def : Concat256<int_aie2_concat_512_256_acc, ACC512, ACC256>;
 def : Concat512<int_aie2_concat_1024_512_acc, ACC1024, ACC512>;
 def : Concat1024<int_aie2_concat_1024_256_acc, ACC1024, ACC256>;
 
+// concat_vector generic opcode
+def : Pat<(v16i32 (concat_vectors (v8i32 VEC256:$src0), (v8i32 VEC256:$src1))),
```
SagarMaheshwari99, line 965: There are patterns for CONCAT above. Can we replace the intrinsics `int_aie2_concat_I512_I256` and friends with the generic ones?

ValentijnvdBeek: Yes, they should be mostly be the same. Would you like me to update the intrinsics definition so that they rely on `concat_vectors` or just replace them out right? The former is a bit of breaking change for people that might depend on those intrinsics.

ValentijnvdBeek: You cannot define intrinsics like that these `concat_vectors` cannot be on the output side of a `table-gen` file. So my suggestion would be just keeping it place to give a transition time and remove it later.

SagarMaheshwari99:
@ValentijnvdBeek Can you check the code formatting?

ValentijnvdBeek:
> @ValentijnvdBeek Can you check the code formatting?

Yes, I rerun the job. That workflow wasn't working yet when I setup this PR

SagarMaheshwari99 (approved):

konstantinschwarz (approved):

PR #50, opened 2024-05-27
[AIE2] Supply interblock liveness information to WAW rewriter

abhinay-anubola:
Add liveness analysis to InterBlockScheduling


llvm/lib/Target/AIE/AIEInterBlockScheduling.cpp
```diff
@@ -55,15 +56,84 @@ void InterBlockScheduling::markEpilogueBlocks() {
...
+void calculateLiveOuts(MachineFunction *MF,
+                       std::map<MachineBasicBlock *, LivePhysRegs> &LiveOuts) {
+  std::queue<MachineBasicBlock *> WorkList;
+  std::set<MachineBasicBlock *> WorkListSet;
+  std::map<MachineBasicBlock *, LivePhysRegs> LiveIns;
+  const TargetRegisterInfo *TRI = MF->getSubtarget().getRegisterInfo();
+  for (MachineBasicBlock &MBB : *MF) {
+    LiveIns[&MBB].init(*TRI);
+    LiveOuts[&MBB].init(*TRI);
+    if (WorkListSet.find(&MBB) == WorkListSet.end()) {
```
martien-de-jong: I think this test is redundant. All basic blocks are distinct, so will not be inserted yet.

llvm/lib/Target/AIE/AIEInterBlockScheduling.cpp
```diff
@@ -55,15 +56,84 @@ void InterBlockScheduling::markEpilogueBlocks() {
...
+    LivePhysRegs OldLiveIns;
+    OldLiveIns.init(*TRI);
+    // Copying LiveIns of MBB into OldLiveIns
+    for (MCPhysReg Reg : LiveIns[MBB]) {
+      OldLiveIns.addReg(Reg);
+    }
+    // Calculating LiveOuts by union of LiveIns of successors of MBB
+    for (MachineBasicBlock *Succ : MBB->successors()) {
+      for (MCPhysReg Reg : LiveIns[Succ]) {
+        LiveOuts[MBB].addReg(Reg);
```
martien-de-jong: CHECK: this is extending the one LiveOut set we maintain for MBB. That means that we compute the union of the current LiveOut and LiveIns of the successors. That is fine if we start from empty LiveIns, but I would explictly mention this in the comment.

llvm/lib/Target/AIE/AIEInterBlockScheduling.cpp
```diff
@@ -55,15 +56,84 @@ void InterBlockScheduling::markEpilogueBlocks() {
...
+    }
+    for (MachineInstr &MI : llvm::reverse(*MBB)) {
+      LiveIns[MBB].stepBackward(MI);
+    }
+    // Check for convergence of LiveIns of MBB
+    bool change =
+        std::distance(OldLiveIns.begin(), OldLiveIns.end()) !=
+            std::distance(LiveIns[MBB].begin(), LiveIns[MBB].end()) ||
+        std::any_of(OldLiveIns.begin(), OldLiveIns.end(),
+                    [&](MCPhysReg Reg) { return !LiveIns[MBB].contains(Reg); });
```
martien-de-jong: Oh dear, what an awkward interface. I would suggest to use a local LivePhysReg CurrentLive. This gets iniitialized with the union of the successors, and updated using stepBackward, Then, we update the new LiveIns from this CurrentLive in a loop that checks change: Record the old size, run over CurrentLive, recording change, update LiveIns and count and finally compare the count with the old size.

llvm/lib/Target/AIE/AIEInterBlockScheduling.cpp
```diff
@@ -55,15 +56,84 @@ void InterBlockScheduling::markEpilogueBlocks() {
...
+    // Calculating LiveIns by traversing the MBB
+    LiveIns[MBB].init(*TRI);
+    for (MCPhysReg Reg : LiveOuts[MBB]) {
+      LiveIns[MBB].addReg(Reg);
+    }
+    for (MachineInstr &MI : llvm::reverse(*MBB)) {
+      LiveIns[MBB].stepBackward(MI);
+    }
+    // Check for convergence of LiveIns of MBB
+    bool change =
```
martien-de-jong: nit: capitalized Change

llvm/lib/Target/AIE/AIEInterBlockScheduling.cpp
```diff
@@ -55,15 +56,84 @@ void InterBlockScheduling::markEpilogueBlocks() {
   }
 }
 
+void calculateLiveOuts(MachineFunction *MF,
+                       std::map<MachineBasicBlock *, LivePhysRegs> &LiveOuts) {
+  std::queue<MachineBasicBlock *> WorkList;
+  std::set<MachineBasicBlock *> WorkListSet;
+  std::map<MachineBasicBlock *, LivePhysRegs> LiveIns;
+  const TargetRegisterInfo *TRI = MF->getSubtarget().getRegisterInfo();
+  for (MachineBasicBlock &MBB : *MF) {
```
martien-de-jong: For fastest convergence, I would push them in a postorder traversal.

llvm/lib/Target/AIE/AIEInterBlockScheduling.cpp
```diff
@@ -55,15 +56,84 @@ void InterBlockScheduling::markEpilogueBlocks() {
...
 
+  std::map<MachineBasicBlock *, LivePhysRegs> LiveOuts;
+  const TargetRegisterInfo *TRI = MF->getSubtarget().getRegisterInfo();
+  calculateLiveOuts(MF, LiveOuts);
+
   // Define our universe of blocks
   for (MachineBasicBlock &MBB : *MF) {
     Blocks.emplace(&MBB, &MBB);
+    // Assigning LiveOuts to BlockState
+    auto &BS = getBlockState(&MBB);
```
martien-de-jong: I think you can get an iterator from emplace().first

llvm/lib/Target/AIE/AIEInterBlockScheduling.cpp
```diff
@@ -55,15 +56,84 @@ void InterBlockScheduling::markEpilogueBlocks() {
...
+
 void InterBlockScheduling::enterFunction(MachineFunction *MF) {
   DEBUG_BLOCKS(dbgs() << ">> enterFunction " << MF->getName() << "\n");
 
   // Get ourselves a hazard recognizer
   HR = std::make_unique<AIEHazardRecognizer>(MF->getSubtarget());
 
+  std::map<MachineBasicBlock *, LivePhysRegs> LiveOuts;
+  const TargetRegisterInfo *TRI = MF->getSubtarget().getRegisterInfo();
+  calculateLiveOuts(MF, LiveOuts);
```
martien-de-jong: Perhaps change calculateLiveOuts to return the map and write
`const std::map<...> LiveOuts = calculateLiveOuts(...)`

martien-de-jong: Or perhaps just return LiveIns, and compute LiveOuts from it in the loop below.

llvm/test/CodeGen/AIE/aie2/schedule/interblock/liveness.mir
```diff
@@ -0,0 +1,151 @@
...
+
+#     bb.0-----*
+#     /  \     |
+# bb.1    bb.2 |
+#     \  /     |
+#     bb.3-----*
+#       |
+#     bb.4
+---
+name:            test_liveness_loop_2
```
martien-de-jong: This test proves that srcarry doesn't escape beyond its definition. I would also like to see a case where it does escape and gets propagated to all reachable paths. I suggest to put it in a separate file and make a pretty complicated flow graph, eg. loop with two nested loops, and perhaps one path where a definition blocks further propagation

abhinay-anubola: Why in separate file?

llvm/lib/Target/AIE/AIEInterBlockScheduling.cpp
```diff
@@ -55,15 +56,91 @@ void InterBlockScheduling::markEpilogueBlocks() {
   }
 }
 
+std::map<MachineBasicBlock *, LivePhysRegs> calculateLiveOuts(MachineFunction *MF) {
+  std::queue<MachineBasicBlock *> WorkList;
+  std::set<MachineBasicBlock *> WorkListSet;
+  std::map<MachineBasicBlock *, LivePhysRegs> LiveIns;
+  std::map<MachineBasicBlock *, LivePhysRegs> LiveOuts;
+  const TargetRegisterInfo *TRI = MF->getSubtarget().getRegisterInfo();
+  for (MachineBasicBlock *MBB : post_order(MF)) {
```
gbossu: I get nervous when I see `post_order` on a DAG. There can be cycles, so I don't think we have the guarantee that all BB successors have been analyzed before we analyze BB itself.

I assume this is why you have a worklist and potentially run over the same BB multiple times. Maybe mention that using `post_order` is not about correctness, but rather optimizing the number of times we need to "re-run".

llvm/lib/Target/AIE/AIEInterBlockScheduling.cpp
```diff
@@ -55,15 +56,91 @@ void InterBlockScheduling::markEpilogueBlocks() {
...
+      }
+      LiveInCount++;
+    }
+    LiveIns[MBB].init(*TRI);
+    // Updates LiveIns of MBB with CurrentLive
+    for (MCPhysReg Reg : CurrentLive) {
+      LiveInCount--;
+      LiveIns[MBB].addReg(Reg);
+    }
+    if (LiveInCount != 0) {
```
gbossu: I don't understand the point of `LiveInCount`. Why do we need to track this? As far as I understand, we just need to re-run the analysis for MBB's predecessors if the liveins of MBB change. That is, if `CurrentLive` differs from `LiveIns[MBB]`. Is there something I'm missing?

llvm/lib/Target/AIE/AIEInterBlockScheduling.cpp
```diff
@@ -55,15 +56,91 @@ void InterBlockScheduling::markEpilogueBlocks() {
...
+        LiveOuts[MBB].addReg(Reg);
+        CurrentLive.addReg(Reg);
+      }
+    }
+    // Calculates CurrentLive by bottom-up traversal the MBB
+    for (MachineInstr &MI : llvm::reverse(*MBB)) {
+      CurrentLive.stepBackward(MI);
+    }
+    // Checks for convergence of LiveIns of MBB
+    bool Change = false;
```
gbossu: I would suggest moving the comparison logic for LiveRegs to a separate function. E.g.
```
bool liveRegsAreEqual(const LivePhysRegs &Lhs, const LivePhysRegs &Rhs) {
  return std::equal(Lhs.begin(), Lhs.end(), Rhs.begin(), Rhs.end());
}
```

That could even become an `operator==` for `LivePhysRegs`. With that function, I think we can make the code more concise, e.g.
```
    if (!liveRegsAreEqual(CurrentLive, LiveIns[MBB]) {
      // The liveins for MBB have changed, trigger a recomputation for MBB's preds
      for (MachineBasicBlock *Pred : MBB->predecessors()) {
        if (WorkListSet.find(Pred) == WorkListSet.end()) {
          WorkListSet.insert(Pred);
          WorkList.push(Pred);
        }
      }
    }
```

Edit: See my comment https://github.com/Xilinx/llvm-aie/pull/50/commits/eb7f87269663af12eaf129e3f61b9ea262c15c3f#r1632940914: I think we can go a bit further to improve readability.

llvm/lib/Target/AIE/AIEInterBlockScheduling.cpp
```diff
@@ -55,15 +56,91 @@ void InterBlockScheduling::markEpilogueBlocks() {
...
+    WorkListSet.erase(MBB);
+    LivePhysRegs CurrentLive;
+    CurrentLive.init(*TRI);
+
+    // Calculates LiveOuts by iterating over each successor of the current MBB
+    // and adding each successor's LiveIns to the current MBB's LiveOuts.
+    // This ensures LiveOuts is the union of LiveIns of all successor MBBs.
+    for (MachineBasicBlock *Succ : MBB->successors()) {
+      for (MCPhysReg Reg : LiveIns[Succ]) {
+        LiveOuts[MBB].addReg(Reg);
```
gbossu: Nit: We could compute the LiveOuts in a final step, and keep that loop only about liveins computation.

llvm/lib/Target/AIE/AIEInterBlockScheduling.cpp
```diff
@@ -55,15 +56,91 @@ void InterBlockScheduling::markEpilogueBlocks() {
...
+    }
+    if (Change) {
+      for (MachineBasicBlock *Pred : MBB->predecessors()) {
+        if (WorkListSet.find(Pred) == WorkListSet.end()) {
+          WorkListSet.insert(Pred);
+          WorkList.push(Pred);
+        }
+      }
+    }
+  }
```
gbossu: That's really a lot of indentation, do you think it would be possible to re-write the main loop into something like below?
```
  while (!WorkList.empty()) {
    MachineBasicBlock *MBB = popFromWorkList();
    LivePhysRegs CurrentLive = computeBlockLiveins(MBB, LiveIns);

    // If MBB's liveins changed, force a recomputation for MBB's preds
    if (liveRegsAreEqual(CurrentLive, LiveIns[MBB])
      continue;
    for (MachineBasicBlock *Pred : MBB->predecessors())
      addToWorkList(MBB);
  }
```

llvm/lib/Target/AIE/AIEInterBlockScheduling.h
```diff
@@ -137,6 +138,7 @@ class Region {
     if (End != BB->end()) {
       ExitInstr = &*End;
     }
+    LiveIns = new LivePhysRegs();
```
gbossu: Isn't that leaking memory? Why do we need dynamic allocation here?

llvm/lib/Target/AIE/AIEBaseSubtarget.cpp
```diff
@@ -444,10 +444,35 @@ class WAWEdges : public ScheduleDAGMutation {
     auto *RI = static_cast<const AIEBaseRegisterInfo *>(TRI);
     LivePhysRegs LiveRegs;
     LiveRegs.init(*TRI);
-    // Reserved registers are considered always live
-    for (MCPhysReg PhysReg : MRI.getReservedRegs().set_bits()) {
-      if (RI->isSimplifiableReservedReg(PhysReg))
-        LiveRegs.addReg(PhysReg);
+    auto Scheduler = DAG->getBB()
+                         ? static_cast<AIEScheduleDAGMI *>(DAG)->getSchedImpl()
+                         : nullptr;
```
gbossu: Can't we just do a `dynamic_cast`?

llvm/lib/Target/AIE/AIEBaseSubtarget.cpp
```diff
@@ -444,10 +444,35 @@ class WAWEdges : public ScheduleDAGMutation {
...
+    const BlockState *BS;
```
gbossu: Please avoid un-initialized pointers, they are very painful to debug :D Maybe we can make the pointer local to the `if (Scheduler)` scope?

llvm/lib/Target/AIE/AIEBaseSubtarget.cpp
```diff
@@ -444,10 +444,35 @@ class WAWEdges : public ScheduleDAGMutation {
...
+      auto BottomRegion = BS->getBottom();
+      if (*Region.begin() == *BottomRegion.begin()) {
+        // If the region is bottom region, liveouts of region are same as
+        // liveouts of MBB
+        for (MCPhysReg Reg : BS->LiveOuts) {
+          LiveRegs.addReg(Reg);
+        }
+      } else {
+        // Else, liveins of succesor region will be the liveouts of current
+        // region
```
gbossu: Not sure it's that easy, there can be a call in between the two regions which clobbers most of the registers. Only the liveins that are also CSRs will remain liveins. However, for our reserved registers, I'm not too sure about their scope, and we should probably assume they are made live by any function call.

For now, I'd suggest to only update the bottom region. If that proves to be insufficient, we can have a follow-up commit that accumulates the liveins of all the regions if there is no function call in between.

llvm/lib/Target/AIE/AIEBaseSubtarget.cpp
```diff
@@ -469,6 +494,20 @@ class WAWEdges : public ScheduleDAGMutation {
       }
       LiveRegs.stepBackward(MI);
     }
+    // Updating liveins of current region
```
gbossu: As said above, I believe this can be dangerous. I'd at least like to move that to another commit to keep the code changes as small as possible.

llvm/test/CodeGen/AIE/aie2/schedule/control_regs/output_dependency.mir
```diff
@@ -19,8 +19,9 @@ body:             |
     ; CHECK-LABEL: name: test_waw
     ; CHECK: liveins: $r0, $r1
     ; CHECK-NEXT: {{  $}}
-    ; CHECK-NEXT: $crsat = MOVX_mvx_scl killed $r0
-    ; CHECK-NEXT: $crsat = MOV_mv_scl killed $r1
+    ; CHECK-NEXT: $crsat = MOVX_mvx_scl $r0 {
+    ; CHECK-NEXT:   $crsat = MOV_mv_scl killed $r1
+    ; CHECK-NEXT: }
     $crsat = MOVX_mvx_scl $r0
     $crsat = MOV_mv_scl $r1
```
gbossu, line 27: The test lost its meaning, as there is no more WAW dependency. Can you add an instruction that reads `$crsat` after the two MOVs? This should force an output dependency on the latter.

llvm/lib/Target/AIE/AIEInterBlockScheduling.cpp
abhinay-anubola: We need to check for convergence of LiveIns (CurrentLive and LiveIns[MBB]) for termination of loop, as there no '==' operator for LivePhysRegs. For comparison I used LiveInCount. LiveIns[MBB] is old liveins and CurrentLive is new liveins.

llvm/lib/Target/AIE/AIEBaseSubtarget.cpp
abhinay-anubola: There are 2 if (Scheduler) scopes, so to avoid re-computation of BlockState I chose to do this way. Shall I initialize it with nullptr?

llvm/lib/Target/AIE/AIEBaseSubtarget.cpp
```diff
@@ -444,10 +444,35 @@ class WAWEdges : public ScheduleDAGMutation {
...
+      auto Region = BS->getCurrentRegion();
+      auto BottomRegion = BS->getBottom();
+      if (*Region.begin() == *BottomRegion.begin()) {
+        // If the region is bottom region, liveouts of region are same as
+        // liveouts of MBB
+        for (MCPhysReg Reg : BS->LiveOuts) {
+          LiveRegs.addReg(Reg);
+        }
+      } else {
+        // Else, liveins of succesor region will be the liveouts of current
```
SagarMaheshwari99: ... liveins of **_the succes`s`or_** region ...

llvm/lib/Target/AIE/AIEInterBlockScheduling.cpp
```diff
@@ -55,15 +56,91 @@ void InterBlockScheduling::markEpilogueBlocks() {
...
+    // Calculates LiveOuts by iterating over each successor of the current MBB
+    // and adding each successor's LiveIns to the current MBB's LiveOuts.
+    // This ensures LiveOuts is the union of LiveIns of all successor MBBs.
+    for (MachineBasicBlock *Succ : MBB->successors()) {
+      for (MCPhysReg Reg : LiveIns[Succ]) {
+        LiveOuts[MBB].addReg(Reg);
+        CurrentLive.addReg(Reg);
+      }
+    }
+    // Calculates CurrentLive by bottom-up traversal the MBB
```
SagarMaheshwari99: _of_ the MBB

llvm/lib/Target/AIE/AIEInterBlockScheduling.cpp
```diff
@@ -55,15 +56,91 @@ void InterBlockScheduling::markEpilogueBlocks() {
   }
 }
 
+std::map<MachineBasicBlock *, LivePhysRegs> calculateLiveOuts(MachineFunction *MF) {
```
andcarminati: I think that this function can also be refactored to a different file to improve readability.

llvm/lib/Target/AIE/AIEInterBlockScheduling.cpp
```diff
@@ -55,15 +56,91 @@ void InterBlockScheduling::markEpilogueBlocks() {
...
+  std::queue<MachineBasicBlock *> WorkList;
```
andcarminati: As there is no change withing basic blocks and instructions, you can use `const` everywhere.

llvm/lib/Target/AIE/AIEInterBlockScheduling.cpp
```diff
@@ -55,15 +56,91 @@ void InterBlockScheduling::markEpilogueBlocks() {
...
+    if (Change) {
+      for (MachineBasicBlock *Pred : MBB->predecessors()) {
+        if (WorkListSet.find(Pred) == WorkListSet.end()) {
+          WorkListSet.insert(Pred);
+          WorkList.push(Pred);
+        }
+      }
+    }
+  }
+  return LiveOuts;
```
andcarminati: I think it is better to use a smart pointer here to avoid the whole copy of the map to the caller.

llvm/lib/Target/AIE/AIEInterBlockScheduling.cpp
martien-de-jong: To wit: entering them in forward order makes the algorithm weakly quadratic in the number of basic blocks. In forward flow you would normally just push the entry point and run the worklist. In backward flow this is problematic, because backward flow is not guaranteed to reach all blocks

llvm/lib/Target/AIE/AIEInterBlockScheduling.cpp
martien-de-jong: It won't be copied I think. It's the last use of local variable that is returned. It will be moved.

llvm/lib/Target/AIE/AIEInterBlockScheduling.h
```diff
@@ -22,6 +22,7 @@
 #include "llvm/CodeGen/MachineScheduler.h"
 #include "llvm/CodeGen/ScheduleHazardRecognizer.h"
 #include <memory>
+#include <queue>
```
gbossu: is this used somewhere?

llvm/lib/Target/AIE/AIELiveRegs.cpp
```diff
@@ -0,0 +1,97 @@
+//===- AIELiveRegs.cpp - Liveness Analysis infrastructure ===//
+//
+// This file is licensed under the Apache License v2.0 with LLVM Exceptions.
+// See https://llvm.org/LICENSE.txt for license information.
+// SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+//
+//===----------------------------------------------------------------------===//
+// Implementations of the classes used to support Liveness Analysis
```
gbossu: Nit: Maybe mention this tracks liveness of all physical registers, including reserved ones.

gbossu:
I think this is looking good! Did you have a chance to try QoR?

abhinay-anubola:
> I think this is looking good! Did you have a chance to try QoR?

I am running it

llvm/test/CodeGen/AIE/aie2/schedule/interblock/liveness.mir
martien-de-jong: It might give you a pass for the simple case immediately. Less investigation when the complicated test fails. smaller test files cause less merge conflicts. lit tests run in milliseconds

llvm/lib/Target/AIE/AIEInterBlockScheduling.cpp
```diff
@@ -111,9 +112,24 @@ void InterBlockScheduling::enterFunction(MachineFunction *MF) {
...
+  const TargetRegisterInfo *TRI = MF->getSubtarget().getRegisterInfo();
+  LiveRegs MBBLiveness(MF);
+
   // Define our universe of blocks
   for (MachineBasicBlock &MBB : *MF) {
-    Blocks.emplace(&MBB, &MBB);
+    auto Itr = Blocks.emplace(&MBB, &MBB).first;
+    auto &BS = Itr->second;
+    BS.LiveOuts.init(*TRI);
+    // Calculating Liveouts by aAssigns LiveOuts to BlockState
```
martien-de-jong: nit: comment garbled

llvm/lib/Target/AIE/AIELiveRegs.cpp
```diff
@@ -0,0 +1,98 @@
...
+      return false;
+    }
+    LiveInCount++;
+  }
+
+  // Updates LiveIns of MBB with CurrentLive
+  LiveIns[MBB].init(*TRI);
+  for (MCPhysReg Reg : CurrentLive) {
+    LiveInCount--;
+    LiveIns[MBB].addReg(Reg);
```
martien-de-jong: Could you please grab a reference to Liveins[MBB] at the start of the function?

martien-de-jong: The name of this function doesn't suggest it's updating. Also, the early return above skips updating. As far as I know this is supposed to update the liveset and indicate a change, not indicate a change and possibly update.

llvm/lib/Target/AIE/AIELiveRegs.cpp
```diff
@@ -0,0 +1,98 @@
...
+  }
+  return LiveInCount == 0;
+}
+
+void LiveRegs::addToWorkList(const MachineBasicBlock *MBB) {
+  if (WorkListSet.find(MBB) == WorkListSet.end()) {
+    WorkListSet.insert(MBB);
+    WorkList.push(MBB);
+  }
```
martien-de-jong, line 68: Nice to have this isolated. Later, if we see that worklists are usually short, we may consider dropping the set and use std::find in the worklist itself. also, erasing from a set may be expensive. We could use a map<MBB, bool> and clear the boolean to erase it.

llvm/lib/Target/AIE/AIELiveRegs.h
```diff
@@ -0,0 +1,62 @@
...
+                        const MachineBasicBlock *MBB);
+
+  // Adds a Machine Basic Block to the work list if it is not already present.
+  void addToWorkList(const MachineBasicBlock *MBB);
+
+public:
+  // Mapping from Machine Basic Blocks to their livein registers.
+  std::map<const MachineBasicBlock *, LivePhysRegs> LiveIns;
+
+  LiveRegs(const MachineFunction *MF);
```
martien-de-jong, line 61: Future: It may be a bit expensive, but you could call this lazily from the DAG mutator if you can't reach InterblockScheduling

llvm/lib/Target/AIE/AIEInterBlockScheduling.h
```diff
@@ -177,7 +178,7 @@ class BlockState {
     CurrentRegion = Regions.size();
     Regions.emplace_back(BB, RegionBegin, RegionEnd);
   }
-  auto getCurrentRegion() const { return Regions.at(CurrentRegion); }
+  const Region &getCurrentRegion() const { return Regions.at(CurrentRegion); }
```
martien-de-jong: Interesting that this was returned by value. I don't think that had been my intention.

llvm/lib/Target/AIE/AIELiveRegs.cpp
```diff
@@ -0,0 +1,99 @@
+//===- AIELiveRegs.cpp - Liveness Analysis infrastructure ===//
```
andcarminati: nit1: introduce some more `==` to equalize the number of columns.
nit2: check copyright.

llvm/lib/Target/AIE/AIELiveRegs.h
```diff
@@ -0,0 +1,62 @@
+//===- AIELiveRegs.h - Liveness Analysis logic -*- C++ -*-===//
```
andcarminati: nit: same comment as for .cpp file.

llvm/lib/Target/AIE/AIELiveRegs.h
```diff
@@ -0,0 +1,62 @@
...
+                           LivePhysRegs &CurrentLive);
+
+  // Checks if the CurrentLive is the same as the liveins of the given MBB.
+  // If they are different, it updates the liveins of the MBB and returns false.
+  bool compareAndUpdateLiveRegs(LivePhysRegs &CurrentLive,
+                                LivePhysRegs &OldLive);
+
+public:
+  // Mapping from Machine Basic Blocks to their livein registers.
+  std::map<const MachineBasicBlock *, LivePhysRegs> LiveIns;
```
andcarminati: I think that this refactoring is very nice! Maybe we can encapsulate the access to this map using a getter, for example.

llvm/lib/Target/AIE/AIEBaseSubtarget.cpp
```diff
@@ -446,17 +446,39 @@ class WAWEdges : public ScheduleDAGMutation {
       }
     }
   }
-  void apply(ScheduleDAGInstrs *DAG) override {
+
+public:
+  void apply(ScheduleDAGInstrs *DAG, AIEPostRASchedStrategy *Scheduler) {
```
andcarminati: Maybe it is better to not pass Scheduler as parameter here because hidden overloads causing warnings.

One idea to avoid this warning:

```
--- a/llvm/lib/Target/AIE/AIEBaseSubtarget.cpp
+++ b/llvm/lib/Target/AIE/AIEBaseSubtarget.cpp
@@ -447,8 +447,15 @@ class WAWEdges : public ScheduleDAGMutation {
     }
   }
 
+  AIEPostRASchedStrategy *Scheduler = nullptr;
+
 public:
-  void apply(ScheduleDAGInstrs *DAG, AIEPostRASchedStrategy *Scheduler) {
+
+  void setScheduler(AIEPostRASchedStrategy *Scheduler) {
+    this->Scheduler = Scheduler;
+  }
+
+  void apply(ScheduleDAGInstrs *DAG) override {
     MachineFunction &MF = DAG->MF;
     MachineRegisterInfo &MRI = MF.getRegInfo();
     const TargetRegisterInfo *TRI = MF.getSubtarget().getRegisterInfo();
@@ -513,7 +520,8 @@ class SchedWAWEdges : public WAWEdges {
     AIEPostRASchedStrategy *Scheduler =
         DAG->getBB() ? static_cast<AIEScheduleDAGMI *>(DAG)->getSchedImpl()
                      : nullptr;
-    WAWEdges::apply(DAG, Scheduler);
+    setScheduler(Scheduler);
+    WAWEdges::apply(DAG);
   }
 };
 
@@ -522,8 +530,7 @@ class SchedWAWEdges : public WAWEdges {
 // independently of the Scheduler.
 class SWPWAWEdges : public WAWEdges {
-    AIEPostRASchedStrategy *Scheduler = nullptr;
-    WAWEdges::apply(DAG, Scheduler);
+    WAWEdges::apply(DAG);
   }
 };
 } // namespace
```

llvm/lib/Target/AIE/AIEBaseSubtarget.cpp
```diff
@@ -446,17 +446,39 @@ class WAWEdges : public ScheduleDAGMutation {
...
+      auto BottomRegion = BS.getBottom();
+      if (*Region.begin() == *BottomRegion.begin()) {
+        // If the region is bottom region, liveouts of region are same as
+        // liveouts of the MBB
+        for (MCPhysReg Reg : BS.LiveOuts) {
+          LiveRegs.addReg(Reg);
+        }
+      } else {
+        // Reserved registers are considered always live
+        for (MCPhysReg PhysReg : MRI.getReservedRegs().set_bits()) {
```
andcarminati: We can simplify a bit to avoid the same code con two different places, like:

```
    bool AddReservedRegs = true;
    if (Scheduler) {
      const BlockState &BS = Scheduler->getInterBlock().getBlockState(MBB);
      auto Region = BS.getCurrentRegion();
      auto BottomRegion = BS.getBottom();
      if (*Region.begin() == *BottomRegion.begin()) {
        // If the region is bottom region, liveouts of region are same as
        // liveouts of the MBB
        for (const MCPhysReg Reg : BS.LiveOuts)
          LiveRegs.addReg(Reg);

        AddReservedRegs = false;
      }
    } 

    if(AddReservedRegs) {
      // Reserved registers are considered always live
      for (const MCPhysReg PhysReg : MRI.getReservedRegs().set_bits()) {
        if (RI->isSimplifiableReservedReg(PhysReg))
          LiveRegs.addReg(PhysReg);
      }
    }
```

llvm/lib/Target/AIE/AIEInterBlockScheduling.cpp
```diff
@@ -111,9 +112,25 @@ void InterBlockScheduling::enterFunction(MachineFunction *MF) {
...
+
   // Define our universe of blocks
   for (MachineBasicBlock &MBB : *MF) {
-    Blocks.emplace(&MBB, &MBB);
+    auto Itr = Blocks.emplace(&MBB, &MBB).first;
+    auto &BS = Itr->second;
+    BS.LiveOuts.init(*TRI);
+    // Calculating LiveOuts by iterating over each successor of the MBB and
+    // adding each successor's LiveIns to LiveOuts.
+    for (MachineBasicBlock *Succ : MBB.successors()) {
```
andcarminati: You can use `const`:

```
for (const MachineBasicBlock *Succ : MBB.successors()) {
   for (const MCPhysReg Reg : MBBLiveness.LiveIns[Succ]) {
```

llvm/lib/Target/AIE/AIELiveRegs.h
```diff
@@ -0,0 +1,62 @@
...
+  // list set, and returns the popped MBB.
+  const MachineBasicBlock *popFromWorkList();
+
+  // Computes the live-in registers for a given Machine Basic Block.
+  void computeBlockLiveIns(const MachineBasicBlock *MBB,
+                           LivePhysRegs &CurrentLive);
+
+  // Checks if the CurrentLive is the same as the liveins of the given MBB.
+  // If they are different, it updates the liveins of the MBB and returns false.
+  bool compareAndUpdateLiveRegs(LivePhysRegs &CurrentLive,
```
andcarminati: nit: compareAndUpdateLiveInRegs

llvm/lib/Target/AIE/AIELiveRegs.cpp
```diff
@@ -0,0 +1,99 @@
...
+    CurrentLive.stepBackward(MI);
+  }
+}
+
+bool LiveRegs::compareAndUpdateLiveRegs(LivePhysRegs &CurrentLive,
+                                        LivePhysRegs &OldLive) {
+  int LiveInCount = 0;
+  bool Equal = true;
+
+  for (MCPhysReg Reg : OldLive) {
```
andcarminati: nit: `const`.

llvm/lib/Target/AIE/AIELiveRegs.cpp
```diff
@@ -0,0 +1,99 @@
...
+    if (!CurrentLive.contains(Reg)) {
+      Equal = false;
+    }
+    LiveInCount++;
+  }
+
+  // Updates LiveIns of MBB with CurrentLive
+  OldLive.init(*TRI);
+  for (MCPhysReg Reg : CurrentLive) {
```
andcarminati: nit: `const`.

llvm/test/CodeGen/AIE/aie2/schedule/interblock/liveness.mir
```diff
@@ -0,0 +1,229 @@
...
+# SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+#
+# (c) Copyright 2024 Advanced Micro Devices, Inc. or its affiliates
+
+# RUN: llc --mtriple=aie2 --run-pass=postmisched --issue-limit=1 -debug-only=machine-scheduler %s -o - 2>%t.log
+# RUN: cat %t.log | FileCheck %s --check-prefix=CHECK-LIVENESS
+# REQUIRES: asserts
+
+# These tests verify liveouts of every MachineBasicBlock in MachineFunction and 
+# liveins of every Region in MachineBasicBlock.
```
gbossu, line 14: Could you add tests with function calls? Those are important to test for two reasons:

- They spilt a BB in multiple sched regions
- They come with a regmask operand like `csr_aie2` which clobbers most registers, including reserved ones

llvm/lib/Target/AIE/AIEBaseSubtarget.cpp
```diff
@@ -446,17 +448,41 @@ class WAWEdges : public ScheduleDAGMutation {
...
+    bool AddReservedRegs = true;
+    if (Scheduler) {
+      MachineBasicBlock *MBB = DAG->getBB();
+      const BlockState &BS = Scheduler->getInterBlock().getBlockState(MBB);
+      auto Region = BS.getCurrentRegion();
+      auto BottomRegion = BS.getBottom();
+      if (*Region.begin() == *BottomRegion.begin()) {
+        // If the region is bottom region, liveouts of region are same as
+        // liveouts of the MBB
+        for (MCPhysReg Reg : BS.LiveOuts) {
```
andcarminati: nit: `const`.

llvm/lib/Target/AIE/AIEBaseSubtarget.cpp
```diff
@@ -446,17 +448,41 @@ class WAWEdges : public ScheduleDAGMutation {
...
+          LiveRegs.addReg(Reg);
+        }
+        AddReservedRegs = false;
+      }
+    }
+
+    if (AddReservedRegs) {
+      // Reserved registers are considered always live
+      for (MCPhysReg PhysReg : MRI.getReservedRegs().set_bits()) {
```
andcarminati: nit: `const`.

abhinay-anubola:
[QoR.zip](https://github.com/user-attachments/files/15839015/QoR.zip)

andcarminati:
> [QoR.zip](https://github.com/user-attachments/files/15839015/QoR.zip)

Hi @abhinay-anubola, it will be nice if you can summarize the results in a visual way, maybe a comparison.

llvm/lib/Target/AIE/AIELiveRegs.cpp
```diff
@@ -0,0 +1,106 @@
...
+  // way that naturally accommodates dependencies and minimizes redundant work.
+  for (const MachineBasicBlock *MBB : post_order(MF)) {
+    LiveIns[MBB].init(*TRI);
+    WorkList.push(MBB);
+    InWorkList[MBB] = true;
+  }
+  while (!WorkList.empty()) {
+    const MachineBasicBlock *MBB = popFromWorkList();
+    LivePhysRegs CurrentLive;
+    computeBlockLiveIns(MBB, CurrentLive);
```
gbossu, line 100: Nit: I'd prefer something like `LivePhysRegs CurrentLive = computeBlockLiveIns(MBB);`

llvm/lib/Target/AIE/AIELiveRegs.cpp
```diff
@@ -0,0 +1,106 @@
...
+    }
+  }
+  // Calculates CurrentLive by bottom-up traversal of the MBB
+  for (const MachineInstr &MI : llvm::reverse(*MBB)) {
+    CurrentLive.stepBackward(MI);
+  }
+}
+
+bool LiveRegs::compareAndUpdateLiveInRegs(LivePhysRegs &CurrentLive,
+                                          LivePhysRegs &OldLive) {
```
gbossu: Did you try the suggestion at https://github.com/Xilinx/llvm-aie/pull/50#discussion_r1632923512?


I.e. defining
```
bool liveRegsAreEqual(const LivePhysRegs &Lhs, const LivePhysRegs &Rhs) {
  return std::equal(Lhs.begin(), Lhs.end(), Rhs.begin(), Rhs.end());
}
```

And then the worklist loop would be:
```
   while (!WorkList.empty()) {
    const MachineBasicBlock *MBB = popFromWorkList();
    LivePhysRegs CurrentLive = computeBlockLiveIns(MBB);

    // If MBB's liveins changed, force a recomputation for MBB's preds
    if (liveRegsAreEqual(CurrentLive, LiveIns[MBB]))
      continue;
    LiveIns[MBB] = CurrentLive;
    for (const MachineBasicBlock *Pred : MBB->predecessors())
      addToWorkList(Pred);
  }
```

But maybe that isn't possible because `LivePhysRegs` has no `operator=`? If that's the case, I'd still be happy to simplify `compareAndUpdateLiveInRegs` into:
```
  if (liveRegsAreEqual(CurrentLive, LiveIns[MBB]))
    return true;

  // Live regs changed, reset OldLive to have the same content as CurrentLive
  OldLive.init(*TRI);
  for (const MCPhysReg Reg : CurrentLive) {
    OldLive.addReg(Reg);
  }
  return false;
```

llvm/lib/Target/AIE/AIELiveRegs.cpp
abhinay-anubola: This cannot be done because there is no '=' operator for LivePhysRegs

llvm/lib/Target/AIE/AIEInterBlockScheduling.cpp
```diff
@@ -111,9 +112,27 @@ void InterBlockScheduling::enterFunction(MachineFunction *MF) {
...
 
+  const TargetRegisterInfo *TRI = MF->getSubtarget().getRegisterInfo();
+  LiveRegs MBBLiveness(MF);
+  auto &LiveIns = MBBLiveness.getLiveIns();
+
   // Define our universe of blocks
   for (MachineBasicBlock &MBB : *MF) {
-    Blocks.emplace(&MBB, &MBB);
+    auto Itr = Blocks.emplace(&MBB, &MBB).first;
+    auto &BS = Itr->second;
```
gbossu: Please expand the `auto` when it helps readability. Here I think it could be `BlockState &BS = Itr->second`;

llvm/lib/Target/AIE/AIELiveRegs.h
```diff
@@ -0,0 +1,66 @@
...
+  // Pops a Machine Basic Block from the work list, removes it from the work
+  // list set, and returns the popped MBB.
+  const MachineBasicBlock *popFromWorkList();
+
+  // Computes the live-in registers for a given Machine Basic Block.
+  void computeBlockLiveIns(const MachineBasicBlock *MBB,
+                           LivePhysRegs &CurrentLive);
+
+  // Checks if the CurrentLive is the same as the liveins of the given MBB.
+  bool liveRegsAreEqual(LivePhysRegs &CurrentLive, LivePhysRegs &OldLive);
```
gbossu: That can be made `static` I think. Maybe we can even name the function simply `equal`, because calling it would be `LiveRegs::equal(Lhs, Rhs)` so the context is clear.

llvm/lib/Target/AIE/AIEInterBlockScheduling.cpp
```diff
@@ -111,9 +112,27 @@ void InterBlockScheduling::enterFunction(MachineFunction *MF) {
   // Get ourselves a hazard recognizer
   HR = std::make_unique<AIEHazardRecognizer>(MF->getSubtarget());
 
+  const TargetRegisterInfo *TRI = MF->getSubtarget().getRegisterInfo();
+  LiveRegs MBBLiveness(MF);
+  auto &LiveIns = MBBLiveness.getLiveIns();
```
gbossu: Nit: add `const`

llvm/lib/Target/AIE/AIEInterBlockScheduling.cpp
```diff
@@ -111,9 +112,27 @@ void InterBlockScheduling::enterFunction(MachineFunction *MF) {
...
   // Define our universe of blocks
   for (MachineBasicBlock &MBB : *MF) {
-    Blocks.emplace(&MBB, &MBB);
+    auto Itr = Blocks.emplace(&MBB, &MBB).first;
+    auto &BS = Itr->second;
+    BS.LiveOuts.init(*TRI);
+    // Calculating LiveOuts by iterating over each successor of the MBB and
+    // adding each successor's LiveIns to LiveOuts.
+    for (const MachineBasicBlock *Succ : MBB.successors()) {
+      auto It = LiveIns.find(Succ);
```
gbossu: Please use `const LivePhysRegs &MBBLiveins = LiveIns.at(Succ)` if you don't manually check the iterator.

llvm/lib/Target/AIE/AIEBaseSubtarget.cpp
```diff
@@ -482,6 +508,26 @@ class WAWEdges : public ScheduleDAGMutation {
   };
 };
 
+// Adds WAW edges for scheduling in the context of the Scheduler.
+// This class extends WAWEdges to apply WAW edges using a Scheduler if available
+// It overrides the apply method to retrieve the Scheduler from the DAG if a
+// BasicBlock is present, otherwise, it uses nullptr.
+class SchedWAWEdges : public WAWEdges {
```
gbossu: Nit: Maybe name this `MachineSchedWAWEdges` to be a bit more explicit. The `SWP` is also a scheduler.

llvm/test/CodeGen/AIE/aie2/schedule/interblock/liveness-region.mir
```diff
@@ -0,0 +1,248 @@
...
+name:            liveness_loop_regions
+alignment:       16
+tracksRegLiveness: true
+body:             |
+  ; CHECK-LIVENESS-LABEL: liveness_loop_regions:BB0 LiveOuts
+  ; CHECK-LIVENESS-NEXT:   Live Registers: $bml0 $amll0 $amlh0 $bml2 $amll2 $amlh2 $r1 $crfpmask
+  ; CHECK-LIVENESS-NEXT: liveness_loop_regions:BB1 LiveOuts
+  ; CHECK-LIVENESS-NEXT:   Live Registers: $bml0 $amll0 $amlh0 $bml2 $amll2 $amlh2 $r1 $crfpmask $lr
+  ; CHECK-LIVENESS-NEXT: liveness_loop_regions:BB2 LiveOuts
+  ; CHECK-LIVENESS-NEXT:   Live Registers: (empty)
```
gbossu, line 127: The log is a bit confusing, `$bml2` is live in `bb.2`, but I assume what we are dumping are the block liveouts? Maybe drop the `Live Registers:` part on the second line and just list the registers?

abhinay-anubola: This cannot be done, as `->dump()` is only way to dump the LiveRegs. `Live Registers:` is logged by inbuilt `dump()`

llvm/test/CodeGen/AIE/aie2/schedule/interblock/liveness-region.mir
```diff
@@ -0,0 +1,248 @@
+# NOTE: Assertions have been autogenerated by utils/update_mir_test_checks.py
+
+# This file is licensed under the Apache License v2.0 with LLVM Exceptions.
+# See https://llvm.org/LICENSE.txt for license information.
+# SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+#
+# (c) Copyright 2024 Advanced Micro Devices, Inc. or its affiliates
+
+# RUN: llc --mtriple=aie2 --run-pass=postmisched --issue-limit=1 -debug-only=machine-scheduler %s -o - 2>%t.log
```
gbossu: I think we should run the verifier, there are some regs which I think aren't properly defined.

llvm/test/CodeGen/AIE/aie2/schedule/interblock/liveness-region.mir
```diff
@@ -0,0 +1,248 @@
...
+  bb.1:
+    successors: %bb.2
+    liveins: $r0, $r1
+    $r3 = nsw ADD $r0, $r1, implicit-def $srcarry
+    JL @callee, implicit-def $lr
+    DelayedSchedBarrier csr_aie2, implicit-def $r1, implicit-def $r3
+    $r1 = AND $r1, $r3
+  bb.2:
+    successors: %bb.3, %bb.4
+    liveins: $r2, $r3
```
gbossu, line 207: When we come from `bb.5`, I think `$r2` is actually undefined. Maybe extend `DelayedSchedBarrier csr_aie2, implicit-def $r6` to also define `$r2`? (And maybe others, I didn't check all the liveins, but the MachineVerifier should 😉)

llvm/test/CodeGen/AIE/aie2/schedule/interblock/liveness.mir
```diff
@@ -0,0 +1,228 @@
...
+
+# These tests verify liveouts of every MachineBasicBlock in MachineFunction
+
+#     bb.0
+#     /  \
+# bb.1    bb.2
+#     \  /
+#     bb.3
+
+---
```
gbossu, line 22: I think some short description for each test would help. I think here what is important is that `$srcarry` is found not to be live, because no-one reads it in the function.

llvm/lib/Target/AIE/AIEBaseSubtarget.cpp
```diff
@@ -482,6 +508,26 @@ class WAWEdges : public ScheduleDAGMutation {
...
+  void apply(ScheduleDAGInstrs *DAG) override {
```
gbossu, line 516: Do we run this more the pre-ra machinescheduler as well? If we do, I think casting to `AIEScheduleDAGMI` is invalid.

llvm/test/CodeGen/AIE/aie2/hardware-loops/sibling.ll
```diff
@@ -45,21 +40,16 @@ define void @sibling(ptr nocapture %out, ptr nocapture readonly %in, i32 noundef
...
-; CHECK-NEXT:    nop
-; CHECK-NEXT:    add r2, r2, #1
-; CHECK-NEXT:    add r0, r0, r4
 ; CHECK-NEXT:    jnzd r1, r1, p2
 ; CHECK-NEXT:    nop // Delay Slot 5
 ; CHECK-NEXT:    nop // Delay Slot 4
-; CHECK-NEXT:    nop // Delay Slot 3
-; CHECK-NEXT:    nop // Delay Slot 2
+; CHECK-NEXT:    add r2, r2, #1 // Delay Slot 3
+; CHECK-NEXT:    add r0, r0, r4 // Delay Slot 2
```
gbossu, line 52: Checking: Those could be moved past `JNZD` because the latter also writes `srCarry` but we proved the dependency can be removed.

llvm/lib/Target/AIE/AIEBaseSubtarget.cpp
abhinay-anubola: It will run only in post ra and also in interblock dag construction. It won’t run in pre ra machine scheduler. I think `AIEScheduleDAGMI` will be valid.

llvm/test/CodeGen/AIE/aie2/schedule/interblock/liveness-region.mir
```diff
@@ -0,0 +1,248 @@
+# NOTE: Assertions have been autogenerated by utils/update_mir_test_checks.py
+
+# This file is licensed under the Apache License v2.0 with LLVM Exceptions.
+# See https://llvm.org/LICENSE.txt for license information.
+# SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+#
+# (c) Copyright 2024 Advanced Micro Devices, Inc. or its affiliates
+
+# RUN: llc --mtriple=aie2 -verify-machineinstrs --run-pass=postmisched --issue-limit=1 -debug-only=machine-scheduler %s -o - 2>%t.log
```
gbossu, line 9: Nit: I think we need the to run the liveregisters (or similar, I don't remember the name) to get verification of live ranges.  You'll probably need to add a couple of implicit-defs. But I think what is found now is correct.

gbossu (approved):
LGTM, let's get this in :)

llvm/test/CodeGen/AIE/aie2/schedule/interblock/liveness-region.mir
abhinay-anubola: I could not find the required flag for verifying liveregisters. For now I will merge the changes

PR #51, opened 2024-05-29
Add InterBlock loop/epilogue analysis

andcarminati:
Implement loop/epilogue analysis to reduce pessimism related to the propagation of loop latencies and resource conflicts to epilogue.
This PR adds:
* MemoryEdges update to ignore bundles.
* Latency analysis using a DDG.
* Conflict analysis using two scoreboards.
* Code refactoring.
* Test additions/updates.


andcarminati:
Benchmark evaluation:

With option `--aie-loop-min-tripcount=4`

First results are without the change and the second are with.
```
+----------------+-------------------+------------------+--+-------------------+------------------+--+--------------------+-------------------+
|     Design     | Total Cycle Count | Total Insn Count |  | Total Cycle Count | Total Insn Count |  | % Dec. Cycle Count | % Dec. Insn Count |
+----------------+-------------------+------------------+--+-------------------+------------------+--+--------------------+-------------------+
| GEMM_bf16_0    |             10994 |             4670 |  |             10898 |             4574 |  | 0,87%              | 2,06%             |
| GEMV_0         |              3163 |              689 |  |              3163 |              689 |  | 0,00%              | 0,00%             |
| Mul2D_0        |              4807 |             1643 |  |              4807 |             1643 |  | 0,00%              | 0,00%             |
| AvgPool2D_0    |              5153 |             3299 |  |              5153 |             3299 |  | 0,00%              | 0,00%             |
| Pad2D_0        |              4522 |             2190 |  |              4522 |             2190 |  | 0,00%              | 0,00%             |
| MaxPool2D_0    |              3832 |             2483 |  |              3832 |             2483 |  | 0,00%              | 0,00%             |
| GEMV_1         |              1969 |              587 |  |              1969 |              587 |  | 0,00%              | 0,00%             |
| Add2D_0        |              8157 |             3478 |  |              8157 |             3478 |  | 0,00%              | 0,00%             |
| Conv2D_0       |             20649 |            10943 |  |             20239 |            10533 |  | 1,99%              | 3,75%             |
| Conv2D_ReLU_0  |              3391 |             1716 |  |              3331 |             1656 |  | 1,77%              | 3,50%             |
| Conv2D_LReLU_0 |              4768 |             3094 |  |              4708 |             3034 |  | 1,26%              | 1,94%             |
+----------------+-------------------+------------------+--+-------------------+------------------+--+--------------------+-------------------+


```

llvm/lib/Target/AIE/AIEInterBlockScheduling.cpp
```diff
@@ -369,14 +407,168 @@ int InterBlockScheduling::getNumEntryNops(const BlockState &BS) const {
...
-  MachineBasicBlock *Loop = *BB.predecessors().begin();
+  MachineBasicBlock *Loop = getSinglePredecessor(BB);
   auto &LBS = getBlockState(Loop);
 
-  // TODO: we can do better by doing full interblock analysis
-  // between BS and LBS
+  // We can only analyze non-empty epilogue blocks because we need
+  // to build a DDG, which is not possible.
+  // For empty ones, we need to be conservative because we are not aware of
+  // content of epilogues' sucessor.
```
martien-de-jong: nit: successor

martien-de-jong: I'm wondering about the above comment, because there's no real distinction between 'empty' or 'too short'. 
The DAG is created anyway, and the RegionEndEdges should push ExitSU out by the right amount. I guess the problem is because we will don't inspect edges to ExitSU

llvm/lib/Target/AIE/AIEInterBlockScheduling.cpp
```diff
@@ -369,14 +407,168 @@ int InterBlockScheduling::getNumEntryNops(const BlockState &BS) const {
...
+          Latency = AIE::maxLatency(MI, *TII, *ItinData, false);
+          DEBUG_LOOPAWARE(dbgs() << "Sched boundary dependency found:\n"
+                                 << " Loop instruction: " << *MI
+                                 << "\n Latency: " << Latency
+                                 << "\n Distance: " << Distance << "\n");
+
+        } else {
+          // We need to skip instructions like debug ones.
+          auto DepthIt = Depths.find(Succ->getInstr());
+          if (DepthIt == Depths.end()) {
```
martien-de-jong: Curious: since we only enter instructions from bundles, I don't expect any debug instructions. Do we ever take this path?

llvm/lib/Target/AIE/AIEInterBlockScheduling.cpp
```diff
@@ -369,14 +407,168 @@ int InterBlockScheduling::getNumEntryNops(const BlockState &BS) const {
...
+                         << EpilogueMBB << "\n");
+
+  emitBundlesInScoreboard(LoopBS.getBottom().Bundles, Bottom, HR.get());
+
+  // We know how many latency cycles we need to respect, and we can advance
+  // the scoreboard to the first possible cycle that can accommodate another
+  // instruction and start the resource verification from this point, tracking
+  // the number of NOPS.
+  int NopCounter = ExistingLatency;
+  for (NopCounter = 0; NopCounter < ExistingLatency; ++NopCounter)
```
martien-de-jong, line 528: nit: I would have `int NopCounter=0;` and leave the first for-clause empty, or write it as a while loop. That while loop will look nicely similar to the one below.

martien-de-jong (dismissed):
Nothing wrong.  Just a few nits you may want to consider.

llvm/lib/Target/AIE/AIEInterBlockScheduling.cpp
andcarminati: This is because we are handling vectors that cannot be empty. This prevents vectors' assertion's from kick in.

llvm/lib/Target/AIE/AIEInterBlockScheduling.cpp
andcarminati: You are right, it is safe to not expect debug instructions here.

llvm/lib/Target/AIE/AIEInterBlockScheduling.cpp
```diff
@@ -32,8 +32,56 @@ static cl::opt<bool>
...
+    for (MachineInstr *MI : Bundles[i].getInstrs())
+      HR->emitInScoreboard(Scoreboard, MI->getDesc(), 0);
+
+    Scoreboard.advance();
+  }
+}
+
+void emitBundlesInScoreboardDelta(
+    const std::vector<MachineBundle> &Bundles,
+    ResourceScoreboard<FuncUnitWrapper> &Scoreboard, int &Cycle,
```
gbossu: Nit: Using the name `Cycle` is confusing me a bit, as this does not correspond to an actual emission cycle of the instructions we are emitting. This is really just a Delta from the `CurrCycle` of `Scoreboard`, and when that Delta reaches `0`, we know there's no need to replay more instructions and identify resource conflicts.

Maybe just rename to `Delta`, and add a comment where the `break` is to explain why we can exit?

llvm/lib/Target/AIE/AIEInterBlockScheduling.cpp
```diff
@@ -369,14 +407,162 @@ int InterBlockScheduling::getNumEntryNops(const BlockState &BS) const {
...
+  Edges.markBoundary();
+  // Second part is the epilogue itself
+  AddRegionToEdges(EpilogueBS.getTop());
+  Edges.buildEdges();
+
+  DEBUG_LOOPAWARE(dumpInterBlock(Edges));
+  // Check cross-boundary latencies.
+  int Height = 1;
+  for (auto &Bundle : reverse(LoopBS.getBottom().Bundles)) {
+    for (auto *MI : Bundle.getInstrs()) {
```
gbossu: Nit: Rename to `LoopMI` or `PreBoundaryMI`

llvm/lib/Target/AIE/AIEInterBlockScheduling.cpp
```diff
@@ -369,14 +407,162 @@ int InterBlockScheduling::getNumEntryNops(const BlockState &BS) const {
...
+
+  InterBlockEdges Edges(*Context);
+  std::map<const MachineInstr *, int> Depths;
+  int Depth = 0;
+  int EntryNops = 0;
+
+  auto AddRegionToEdges = [&](const Region &R) {
+    for (auto &Bundle : R.Bundles) {
+      for (MachineInstr *MI : Bundle.getInstrs()) {
+        Depths[MI] = Depth;
```
gbossu: Nit: The term `Depth` is extremely overloaded and it's hard for me to keep track of all the meanings. Here I think we use it to access the distance from the "pre-boundary" region. Maybe name it `DistFromLoopEntry` or `DistFromFirstBundle`?

llvm/lib/Target/AIE/AIEInterBlockScheduling.cpp
```diff
@@ -369,14 +407,162 @@ int InterBlockScheduling::getNumEntryNops(const BlockState &BS) const {
...
+        // We need to have special handling for instruction that have
+        // as successor ExitSU. In this case, if ExitSU is not far away
+        // enough to ensure the correct behavior considering the
+        // instructions that follow the epilogue (epilogue's successor),
+        // we need to insert nops. This affects spacially small epilogues.
+        if (Succ->isBoundaryNode()) {
+          // We can consider the maximum depth+1 as depth of the
+          // ExitSU.
+          Distance = Depth - Depths[MI];
+          Latency = AIE::maxLatency(MI, *TII, *ItinData, false);
```
gbossu: Do we still need the special casing now that `Edges.buildEdges();` also runs the `RegionEndEdges` mutator? I would expect that the latency is already correctly computed.

llvm/lib/Target/AIE/AIEInterBlockScheduling.cpp
```diff
@@ -369,14 +407,162 @@ int InterBlockScheduling::getNumEntryNops(const BlockState &BS) const {
...
+      ++Depth;
+    }
+  };
+
+  // Construction of the superblock containing Loop+Epilogue
+  // First part is the loop
+  AddRegionToEdges(LoopBS.getBottom());
+  Edges.markBoundary();
+  // Second part is the epilogue itself
+  AddRegionToEdges(EpilogueBS.getTop());
```
gbossu, line 458: Nit: Maybe also add the `ExitSU` node of `EpilogueBS` to the map of distances so we can avoid the `if (Succ->isBoundaryNode())` corner case in the loop below.

gbossu (commented):
This looks good, I think the last remaining thing would be to fix the assertion that now triggers in the `RegionEndEdges` mutator for some benchmarks, and add a small reproducer test for it.

On top of that, please look at the Nit comments and address up to taste :) Generally, I'd be happy if we can get rid of the `if (Succ->isBoundaryNode())` special casing.

llvm/lib/Target/AIE/AIEInterBlockScheduling.cpp
andcarminati: In the current implementation, this mapping maps instructions to depths, in this way we need to change the mapping logic as well.

llvm/lib/Target/AIE/AIEInterBlockScheduling.cpp
andcarminati: In this case, it is necessary the identification of the boundary node because there is no instruction associated (no mapping). But, as the mutation is running now, I simplified the code to use the already computed latency.

llvm/lib/Target/AIE/AIEInterBlockScheduling.cpp
andcarminati: I updated this point, thank you!

llvm/lib/Target/AIE/AIEInterBlockScheduling.cpp
andcarminati: I updated the named to reflect better the semantics.

andcarminati:
Hi @martien-de-jong and @gbossu, I addressed all the comments and suggestions. Benchmarks are running again with no problems. I you can take a look, it would be nice.

andcarminati:
Here the most recent results:

```

+----------------+---+------------+---------------+-------------+---+---+------------+---------------+-------------+---+---+-------------------+----------------+--------------------+
|       #        | # | Without EA |       #       |      #      | # | # |  With EA   |       #       |      #      | # | # |       Comp.       |       #        |         #          |
+----------------+---+------------+---------------+-------------+---+---+------------+---------------+-------------+---+---+-------------------+----------------+--------------------+
| Design         |   | Tot. PM S. | Tot. Cycle C. | Tot. Ins C. |   |   | Tot. PM S. | Tot. Cycle C. | Tot. Ins C. |   |   | % Diff Tot. PM S. | % Diff Tot. C. | % Diff Tot. Ins C. |
| GEMM_bf16_0    |   | 4160       | 10461         | 4160        |   |   | 4144       | 10365         | 4064        |   |   | -0,38%            | -0,92%         | -2,31%             |
| GEMV_0         |   | 3488       | 3126          | 650         |   |   | 3488       | 3126          | 650         |   |   | 0,00%             | 0,00%          | 0,00%              |
| Mul2D_0        |   | 2672       | 4857          | 1686        |   |   | 2672       | 4857          | 1686        |   |   | 0,00%             | 0,00%          | 0,00%              |
| AvgPool2D_0    |   | 3552       | 4159          | 2811        |   |   | 3552       | 4159          | 2811        |   |   | 0,00%             | 0,00%          | 0,00%              |
| Pad2D_0        |   | 3456       | 4512          | 2184        |   |   | 3456       | 4512          | 2184        |   |   | 0,00%             | 0,00%          | 0,00%              |
| MaxPool2D_0    |   | 3344       | 3798          | 2449        |   |   | 3344       | 3798          | 2449        |   |   | 0,00%             | 0,00%          | 0,00%              |
| GEMV_1         |   | 3488       | 1948          | 560         |   |   | 3488       | 1948          | 560         |   |   | 0,00%             | 0,00%          | 0,00%              |
| Add2D_0        |   | 3600       | 8157          | 3478        |   |   | 3600       | 8157          | 3478        |   |   | 0,00%             | 0,00%          | 0,00%              |
| Conv2D_0       |   | 5952       | 19018         | 10543       |   |   | 5920       | 18538         | 10063       |   |   | -0,54%            | -2,52%         | -4,55%             |
| Conv2D_ReLU_0  |   | 5328       | 3377          | 1704        |   |   | 5296       | 3317          | 1644        |   |   | -0,60%            | -1,78%         | -3,52%             |
| Conv2D_LReLU_0 |   | 6000       | 4767          | 3095        |   |   | 5968       | 4707          | 3035        |   |   | -0,53%            | -1,26%         | -1,94%             |
+----------------+---+------------+---------------+-------------+---+---+------------+---------------+-------------+---+---+-------------------+----------------+--------------------+


```

llvm/lib/Target/AIE/AIEInterBlockScheduling.cpp
gbossu: What I meant is something like: `DistancesFromLoopEntry[EpilogueBS.getTop().getExitSU()] = DistFromLoopEntry;`, this way there is no `if (Succ->isBoundaryNode())` special casing in the loop below.

llvm/lib/Target/AIE/AIEInterBlockScheduling.cpp
```diff
@@ -369,14 +409,160 @@ int InterBlockScheduling::getNumEntryNops(const BlockState &BS) const {
...
+  emitBundlesInScoreboardDelta(EpilogueBS.getBottom().Bundles, Top, Cycle,
+                               HR.get());
+
+  DEBUG_LOOPAWARE(dbgs() << "Epilogue scoreboard\n"; Top.dump());
+
+  // Use scoreboard comparison to calculate the number of nops
+  while (Bottom.conflict(Top, Depth)) {
+    Bottom.advance();
+    NopCounter++;
+  }
```
gbossu, line 548: Nit: That looks a lot like `AIEPostRASchedStrategy::handleRegionConflicts`, maybe there's code we can share.

gbossu (dismissed):
LGTM, I think it's great work!

Nit: I'd be even happier if we could remove the `if (Succ->isBoundaryNode())` special case in `getCyclesToRespectTiming()`, maybe by extending the map of depths, maybe by using a conditional assignment instead.

llvm/lib/Target/AIE/AIEInterBlockScheduling.cpp
andcarminati: Hi @gbossu, I think it could be a bit complicated, because `handleRegionConflicts` uses two hazard recognizers, while we use just one, plus scoreboard comparison. I am afraid that a refactor could create a more confusing code.

llvm/lib/Target/AIE/AIEInterBlockScheduling.cpp
andcarminati: Hi, I simplified a bit more the code.

andcarminati:
> LGTM, I think it's great work!
> 
> Nit: I'd be even happier if we could remove the `if (Succ->isBoundaryNode())` special case in `getCyclesToRespectTiming()`, maybe by extending the map of depths, maybe by using a conditional assignment instead.

Thank you for this suggestion, I addressed this special case.

llvm/lib/Target/AIE/AIEInterBlockScheduling.cpp
```diff
@@ -369,14 +409,152 @@ int InterBlockScheduling::getNumEntryNops(const BlockState &BS) const {
...
+        auto *Succ = SDep.getSUnit();
+
+        int Distance;
+        int Latency = SDep.getSignedLatency();
+
+        if (!Edges.isPostBoundaryNode(Succ))
+          continue;
+        const MachineInstr *PostBoundaryMI = Succ->getInstr();
+
+        int PostBondOrExitDist =
```
gbossu: Nit `PostBoundOrExitDist`

llvm/lib/Target/AIE/AIEInterBlockScheduling.cpp
gbossu: The implementation is different, but ultimately the goal is the same, we want to insert NOPs until there is no resource hazard. We can do that in a follow-up PR.

gbossu (dismissed):
LGTM

gbossu (approved):

gbossu (approved):

martien-de-jong (approved):
LGTM

PR #53, opened 2024-05-30
Some fixups.

SagarMaheshwari99:


clang/cmake/caches/Peano-AIE-runtime-libraries.cmake
```diff
@@ -22,7 +22,6 @@ foreach(target ${LLVM_BUILTIN_TARGETS})
   set(BUILTINS_${target}_CMAKE_BUILD_TYPE Release CACHE STRING "")
   set(RUNTIMES_${target}_CMAKE_BUILD_TYPE Release CACHE STRING "")
   set(BUILTINS_${target}_LLVM_USE_LINKER lld CACHE STRING "")
```
konstantinschwarz, line 24: The same check is not performed for BUILTINS?

SagarMaheshwari99: No.

konstantinschwarz (approved):

PR #54, opened 2024-05-31
[AIE2] Disable the findPostIncMatch combiner for 128-bit stores

ValentijnvdBeek:
In the AIE2 architecture, there is not a complete correspondence between the capacities for the full-sized vector load instruction and the 128-bit vector load instruction. For this combiner we make use of the Shift-Round-Saturate variant, but this doesn't exist for the 128-bit instruction.

### DIFF
--- a/llvm/lib/Target/AIE/AIECombinerHelper.cpp
+++ b/llvm/lib/Target/AIE/AIECombinerHelper.cpp
@@ -12,6 +12,7 @@
 #include "AIE2TargetMachine.h"
 #include "AIEBaseInstrInfo.h"
 #include "MCTargetDesc/AIE2MCTargetDesc.h"
+#include "MCTargetDesc/AIEMCTargetDesc.h"
 #include "llvm/ADT/SetVector.h"
 #include "llvm/CodeGen/GlobalISel/GenericMachineInstrs.h"
 #include "llvm/CodeGen/GlobalISel/Utils.h"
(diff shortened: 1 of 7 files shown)


llvm/lib/Target/AIE/AIECombinerHelper.cpp
```diff
@@ -218,7 +218,16 @@ MachineInstr *findPostIncMatch(MachineInstr &MemI, MachineRegisterInfo &MRI,
...
+
+  const Register DstReg = MemI.getOperand(0).getReg();
+  const unsigned DstSize = MRI.getType(DstReg).getSizeInBits();
+  if (DstSize >= 1024)
+    return nullptr;
+
+  // Disable the optimizations for 128 G_STORE since AIE2 lacks
+  // support for all the same variants of vector loading for 128
+  // as it does for register-sized vectors.
+  if (DstSize == 128 && MemI.getOpcode() == TargetOpcode::G_STORE)
```
konstantinschwarz: Could you add a test?
Also, VST.128 supports post-increment store using normal pointers. We should still support that mode for 128-bit stores

ValentijnvdBeek: Thanks for the review, I have updated the code so that is a bit more specific with regards to only skipping 2D and 3D loads. Do you know if it is an issue for both AIE1 & AIE2?

PR #55, opened 2024-06-04
Adding libc, libm support and test

mchokshi091101:
Adding
- Library search path
- `-lc -lm` options to look for `libc` and `libm` libraries in the provided search paths.
- Added test checking the same.

### DIFF
--- a/clang/lib/Driver/ToolChains/AIE.cpp
+++ b/clang/lib/Driver/ToolChains/AIE.cpp
@@ -37,9 +37,11 @@ void aie::Linker::ConstructJob(Compilation &C, const JobAction &JA,
       !Args.hasArg(options::OPT_nodefaultlibs)) {
     AddRunTimeLibs(ToolChain, ToolChain.getDriver(), CmdArgs, Args);
   }
-
+  ToolChain.AddFilePathLibArgs(Args, CmdArgs);
   if (!Args.hasArg(options::OPT_nostdlib, options::OPT_nostartfiles,
                    options::OPT_r)) {
+    CmdArgs.push_back(Args.MakeArgString("-lc"));   
+    CmdArgs.push_back(Args.MakeArgString("-lm"));            
     CmdArgs.push_back(Args.MakeArgString(ToolChain.GetFilePath("crt0.o")));
     CmdArgs.push_back(Args.MakeArgString(ToolChain.GetFilePath("crt1.o")));
   }
--- a/clang/lib/Driver/ToolChains/AIE.cpp
+++ b/clang/lib/Driver/ToolChains/AIE.cpp
@@ -33,15 +33,16 @@ void aie::Linker::ConstructJob(Compilation &C, const JobAction &JA,
   ArgStringList CmdArgs;
   AddLinkerInputs(ToolChain, Inputs, Args, CmdArgs, JA);
   // CmdArgs.push_back("-shared");
+  ToolChain.AddFilePathLibArgs(Args, CmdArgs);
   if (!Args.hasArg(options::OPT_nostdlib) &&
       !Args.hasArg(options::OPT_nodefaultlibs)) {
     AddRunTimeLibs(ToolChain, ToolChain.getDriver(), CmdArgs, Args);
+    CmdArgs.push_back(Args.MakeArgString("-lc"));
+    CmdArgs.push_back(Args.MakeArgString("-lm"));
   }
-  ToolChain.AddFilePathLibArgs(Args, CmdArgs);
+
   if (!Args.hasArg(options::OPT_nostdlib, options::OPT_nostartfiles,
                    options::OPT_r)) {
-    CmdArgs.push_back(Args.MakeArgString("-lc"));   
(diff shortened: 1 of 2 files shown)


clang/lib/Driver/ToolChains/AIE.cpp
```diff
@@ -37,9 +37,11 @@ void aie::Linker::ConstructJob(Compilation &C, const JobAction &JA,
       !Args.hasArg(options::OPT_nodefaultlibs)) {
     AddRunTimeLibs(ToolChain, ToolChain.getDriver(), CmdArgs, Args);
   }
-
+  ToolChain.AddFilePathLibArgs(Args, CmdArgs);
   if (!Args.hasArg(options::OPT_nostdlib, options::OPT_nostartfiles,
                    options::OPT_r)) {
+    CmdArgs.push_back(Args.MakeArgString("-lc"));   
+    CmdArgs.push_back(Args.MakeArgString("-lm"));
```
stephenneuendorffer: It's unclear to me how this should be guarded.. should it just go in the above case with nostdlib, nodefaultlibs?

stephenneuendorffer (dismissed):

clang/lib/Driver/ToolChains/AIE.cpp
konstantinschwarz: From GNU linker docs:

>-nodefaultlibs
Do not use the standard system libraries when linking. Only the libraries you specify are passed to the linker, and options specifying linkage of the system libraries, such as -static-libgcc or -shared-libgcc, are ignored. The standard startup files are used normally, unless -nostartfiles is used.

>-nostdlib
Do not use the standard system startup files or libraries when linking. No startup files and only the libraries you specify are passed to the linker, and options specifying linkage of the system libraries, such as -static-libgcc or -shared-libgcc, are ignored.

>-nostartfiles
Do not use the standard system startup files when linking. The standard system libraries are used normally, unless -nostdlib, -nolibc, or -nodefaultlibs is used.


So yes, I think these new options should be moved into the 
```
if (!Args.hasArg(options::OPT_nostdlib) &&
      !Args.hasArg(options::OPT_nodefaultlibs)) {
```
block.

mchokshi091101: Sure, will move it

clang/test/Driver/aie2/aie2-stdlib.c
```diff
@@ -0,0 +1,25 @@
...
+// LIBS-SAME: "{{.*}}/lib/clang/{{.*}}/lib/aie2-none-unknown-elf/libclang_rt.builtins.a"
+// LIBS-SAME: "-lc"
+// LIBS-SAME: "-lm"
+
+// ... but not if we have --nostdlib or a target without a library.
+// RUN: %clang %s -### --target=aie2-none-unknown-elf -nostdlib 2>&1 \
+// RUN:   | FileCheck -check-prefix=NOLIBS %s
+// NOLIBS-NOT: ld.lld{{.*}}libclang_rt.builtins.a
+// NOLIBS-NOT: ld.lld{{.*}}-lm
+// NOLIBS-NOT: ld.lld{{.*}}-lc
```
SagarMaheshwari99: Nit: add new empty line.

clang/test/Driver/aie2/aie2-stdlib.c
```diff
@@ -0,0 +1,25 @@
...
+// LIBS-SAME: "-L{{[^"]*}}/lib/aie2-none-unknown-elf"
+// LIBS-SAME: "-L{{[^"]*}}lib/clang/{{..}}/lib/aie2-none-unknown-elf
+// LIBS-SAME: "{{.*}}/lib/clang/{{.*}}/lib/aie2-none-unknown-elf/libclang_rt.builtins.a"
+// LIBS-SAME: "-lc"
+// LIBS-SAME: "-lm"
+
+// ... but not if we have --nostdlib or a target without a library.
+// RUN: %clang %s -### --target=aie2-none-unknown-elf -nostdlib 2>&1 \
+// RUN:   | FileCheck -check-prefix=NOLIBS %s
+// NOLIBS-NOT: ld.lld{{.*}}libclang_rt.builtins.a
```
SagarMaheshwari99, line 23: libclang_rt.builtins.a is also under stdlibs?

konstantinschwarz: Yes, see https://github.com/Xilinx/llvm-aie/pull/55#discussion_r1627243117

clang/test/Driver/aie2/aie2-stdlib.c
```diff
@@ -0,0 +1,25 @@
+//===- aie2-stdlib.c ------------------------------------------*- C++ -*-===//
```
stephenneuendorffer, line 1: This is basically the same as aie2-startup.c.  Does this need to be separately?

mchokshi091101: Maybe add all tests in one and rename it to `aie2-toolchain.c`?

PR #58, opened 2024-06-05
[AIE2] Mark core_id as reserved.

SagarMaheshwari99:


llvm/lib/Target/AIE/AIE2RegisterInfo.cpp
```diff
@@ -108,6 +108,9 @@ BitVector AIE2RegisterInfo::getReservedRegs(const MachineFunction &MF) const {
   ReserveRegs(AIE2::eRRegClass, ReservedGPRs);
   ReserveRegs(AIE2::eDRegClass, ReservedMODs);
 
+  // CORE_ID is reserved.
+  markSuperRegs(Reserved, AIE2::CORE_ID);
```
gbossu, line 112: Could maybe add an end-to-end .ll test with an intrinsic that uses this register?

SagarMaheshwari99: There is one, it was missing the `--verify-machineinstrs`, which I added now.

konstantinschwarz (approved):

llvm/lib/Target/AIE/AIE2RegisterInfo.cpp
gbossu: Even better :)

gbossu (approved):

andcarminati (approved):

PR #64, opened 2024-06-06
[AIE2] NFC: Cleanup the definitions of _split instructions

gbossu:
As a reminder, `_split` instructions variants were introduced to expose the sub-registers of our 2D and 3D registers. This is then easier for the staged RA pipeline to allocate and give new VRegs to each lane.

E.g.
```
%20:ep, %100.sub_dim_count:ed = PADDA_2D killed %20, killed %100
; Turned into:
%20:ep, %100.sub_dim_count:ed = PADDA_2D_split killed %20, killed %100.sub_mod, killed %100.sub_dim_size, killed %100.sub_dim_stride, killed %100.sub_dim_count
```

In this PR, I'm using the new `!dag` bang operator to simplify our .td files, and there is a new TableGen generator that reduces the amount of manual code that has to be written.


llvm/lib/Target/AIE/AIE2InstrInfo.td
```diff
@@ -573,5 +573,52 @@ let Itinerary = II_MOV_SCL, hasSideEffects = 0,  mayLoad = 0, mayStore = 0 in {
       Pseudo<(outs OP_mMvSclDst:$mMvSclDst), (ins OP_mMvSclSrc:$mMvSclSrc), "copy","$mMvSclDst, $mMvSclSrc">;
 }
 
+// Define _split variants for instructions using 2D registers
+class Split2DInstr<Instruction RealInst, int opidx> : SplitPseudo<RealInst,
+    opidx, (ins eM:$mod, eDN:$dim_size, eDJ:$dim_stride, eDC:$dim_count)> {}
```
martien-de-jong, line 578: Any plans to get a single-point of definition for this tuple from register info, also creating the tied register arrays?

llvm/lib/Target/AIE/AIEBaseInstrFormats.td
```diff
@@ -34,3 +34,37 @@ class AIEPseudoInstrAttributes {
...
+// DagAfterSplit. Note that the element at index idx won't be in either of those.
+class SplitDAGOps<dag DAG, int idx> {
+  list<string> DagKeysBefore = !foreach(i, !range(0, idx), !getdagname(DAG, i));
+  list<DAGOperand> DagArgsBefore = !foreach(key, DagKeysBefore,
+      !getdagarg<DAGOperand>(DAG, key));
+  list<string> DagKeysAfter = !foreach(i, !range(!add(idx, 1), !size(DAG)), !getdagname(DAG, i));
+  list<DAGOperand> DagArgsAfter = !foreach(key, DagKeysAfter,
+      !getdagarg<DAGOperand>(DAG, key));
+  dag DagBeforeSplit = !dag(!getdagop(DAG), DagArgsBefore, DagKeysBefore);
+  dag DagAfterSplit = !dag(!getdagop(DAG), DagArgsAfter, DagKeysAfter);
```
martien-de-jong, line 49: Add a dag DagSplit to be more generally useful?

llvm/lib/Target/AIE/AIEBaseInstrFormats.td
```diff
@@ -34,3 +34,37 @@ class AIEPseudoInstrAttributes {
...
+
+  // Copy properties over from RealInst
+  let Constraints = RealInst.Constraints;
+  let mayStore = RealInst.mayStore;
+  let mayLoad = RealInst.mayLoad;
+  let hasSideEffects = RealInst.hasSideEffects;
+  let Defs = RealInst.Defs;
+  let Uses = RealInst.Uses;
+  let Namespace = RealInst.Namespace;
+}
```
martien-de-jong, line 73: I'm not sure how and where to fit it, but I guess there will also be a translation to be made for the operand latencies array at some point. I could  think of [L=4] -> [L+0, L+1, L+0, L+0]. implicitly coded as [0, 1, 0, 0].

llvm/lib/Target/AIE/AIE2InstrInfo.cpp
```diff
@@ -1158,392 +1159,6 @@ AIE2InstrInfo::getTiedRegInfo(unsigned Opcode) const {
...
-  case AIE2::VLDB_3D_UNPACK_D16_D8_split:
-    return AIE2::VLDB_3D_UNPACK_D16_D8;
-  case AIE2::LDA_2D_dmv_lda_q_split:
-    return AIE2::LDA_2D_dmv_lda_q;
-  case AIE2::LDA_3D_dmv_lda_q_split:
-    return AIE2::LDA_3D_dmv_lda_q;
-  default:
-    return {};
-  }
-}
```
martien-de-jong, line 1545: Ohh, nice!

llvm/lib/Target/AIE/AIE2InstrInfo.td
gbossu: I don't think it's easily possible to compute a list of sub-registers for an arbitrary regclass entirely in TableGen constructs. So yes, a single point of definition is sensible once we start to generate more of RegisterInfo/InstrInfo.

Future candidates for auto generation would be:
 - `AIE2InstrInfo::getTiedRegInfo(unsigned Opcode)`, although I'd first like to discuss with the community about a common syntax like `Constraints = "$count_out = $mod.sub_count"` would be acceptable before spending more time on our fully-internal solution for tying a register with a sub-register
 - `const std::set<int> &AIE2RegisterInfo::getSubRegSplit(int RegClassId)`
 - `SmallSet<int, 8> AIE2RegisterInfo::getCoveringSubRegs(const TargetRegisterClass &RC)`
 - `AIE2InstrInfo::getSpillPseudoExpandInfo(const MachineInstr &MI)`

llvm/lib/Target/AIE/AIEBaseInstrFormats.td
gbossu: The whole point of the `_split` instructions is that they only shortly live during the RA pipeline. They are then rewritten as "fully-fledged" instructions with encoding + timing info after RA. That means their attributes can be kept minimal. Or do you have something else in mind that could use `SplitPseudo`?

llvm/lib/Target/AIE/AIEBaseInstrFormats.td
gbossu: What would you like in there?

martien-de-jong: Well, something so that "DagSplitBefore; DegSplit; DagSplitAfter" represents the original. No big deal, certainly not now.

llvm/lib/Target/AIE/AIEBaseInstrFormats.td
martien-de-jong: Perhaps I misunderstood. If there ever is a different access cycle for one of the subregisters, we may want to be able to describe that at least for post-RA sched. It also means that the non-split form will have an inaccuracy in its operand latencies. I worry too much ;-)

gbossu: Currently, the `_split` instructions have no itinerary, this is for two reasons:
 - It would be a lot of work to write or craft so many new itineraries
 - We really do not need scheduling info, as those instructions are seen by neither the premisched, nor the postmisched

llvm/lib/Target/AIE/AIE2InstrInfo.td
martien-de-jong: I would also like to generate the decoder functions

llvm/lib/Target/AIE/AIEBaseInstrFormats.td
```diff
@@ -34,3 +34,37 @@ class AIEPseudoInstrAttributes {
...
+  SplitDAGOps SplitOps = SplitDAGOps<RealInst.InOperandList, !sub(idx, !size(OutOperandList))>;
+  let InOperandList = !con(SplitOps.DagBeforeSplit, NewSplitIns, SplitOps.DagAfterSplit);
+
+  // Copy properties over from RealInst
+  let Constraints = RealInst.Constraints;
+  let mayStore = RealInst.mayStore;
+  let mayLoad = RealInst.mayLoad;
+  let hasSideEffects = RealInst.hasSideEffects;
+  let Defs = RealInst.Defs;
+  let Uses = RealInst.Uses;
```
martien-de-jong, line 68: CHECK: All the instruction defs that were removed are now generated through this rule, and all of them had no other attributes than the ones listed here. -- I guess so, we would have seen the fallout.

llvm/lib/Target/AIE/AIE2InstrInfo.cpp
```diff
@@ -1158,392 +1159,6 @@ AIE2InstrInfo::getTiedRegInfo(unsigned Opcode) const {
   }
 }
 
-std::optional<unsigned>
-AIE2InstrInfo::getOpcodeWithAtomicOperands(unsigned Opcode) const {
-  // TODO: This should be tablegen-erated.
```
martien-de-jong, line 1163: WELLDONE: This has become tablegen-generated

martien-de-jong (approved):

llvm/lib/Target/AIE/AIE2InstrInfo.cpp
gbossu: This was my masterplan all along 😈

llvm/lib/Target/AIE/AIEBaseInstrFormats.td
gbossu: The name is misleading, these are implicit defs and uses. They are just copied over.

PR #66, opened 2024-06-07
[AIEX] Fix premisched delaying SUs indefinitely

gbossu:
Under rare circumstances, a cycle of SUnits that keeps delaying each other indefinitely can be formed. This PR ensures the cycle is detected, and the nodes are not being delayed.

The first two commits are about changing the `isAvailableNode` interface to be non `const`, allowing `MachineSchedStrategy` to keep a bit of state around before rendering its decision. We use it to keep track of the cycles of delayed instructions.

TODO: ~I'm trying to add a test, and running the QoR suite~

As expected, there are no QoR changes other than 3 newly-passing tests.

### DIFF
--- a/llvm/include/llvm/CodeGen/MachineScheduler.h
+++ b/llvm/include/llvm/CodeGen/MachineScheduler.h
@@ -300,7 +300,7 @@ public:
   /// overridden by schedulers that try to release successors/predecessors early
   /// to schedule "out of order".
   virtual bool isAvailableNode(SUnit &SU, SchedBoundary &Zone,
-                               bool VerifyReadyCycle) const;
+                               bool VerifyReadyCycle);
 
   /// Scheduler callback to notify that a new subtree is scheduled.
   virtual void scheduleTree(unsigned SubtreeID) {}
@@ -918,7 +918,7 @@ public:
   };
 
   ScheduleDAGMI *DAG = nullptr;
-  const MachineSchedStrategy *SchedImpl = nullptr;
+  MachineSchedStrategy *SchedImpl = nullptr;
   const TargetSchedModel *SchedModel = nullptr;
   SchedRemainder *Rem = nullptr;
 
@@ -1030,7 +1030,7 @@ public:
 
   void reset();
 
-  void init(ScheduleDAGMI *DAG, const MachineSchedStrategy *SchedImpl,
+  void init(ScheduleDAGMI *DAG, MachineSchedStrategy *SchedImpl,
             const TargetSchedModel *SModel, SchedRemainder *Rem);
 
   bool isTop() const {
--- a/llvm/lib/Target/AIE/AIEMachineScheduler.cpp
+++ b/llvm/lib/Target/AIE/AIEMachineScheduler.cpp
@@ -469,7 +469,7 @@ int AIEPostRASchedStrategy::getMaxDeltaCycles(const SchedBoundary &Zone) const {
 }
 
 bool AIEPostRASchedStrategy::isAvailableNode(SUnit &SU, SchedBoundary &Zone,
-                                             bool /*VerifyReadyCycle*/) const {
+                                             bool /*VerifyReadyCycle*/) {
   // Whether or not the zone is Top or Bot, verify if SU is ready to be
   // scheduled in terms of cycle.
   if (Zone.isTop())
@@ -855,7 +855,7 @@ const SUnit *findPressureReducer(unsigned CriticalPSet, ArrayRef<SUnit *> Nodes,
 }
 
 bool AIEPreRASchedStrategy::isAvailableNode(SUnit &SU, SchedBoundary &Zone,
-                                            bool /*VerifyReadyCycle*/) const {
+                                            bool /*VerifyReadyCycle*/) {
   // Force verifying if SU is ready to be scheduled in terms of cycle.
   bool Avail = MachineSchedStrategy::isAvailableNode(SU, Zone,
                                                      /*VerifyReadyCycle=*/true);
--- a/llvm/lib/Target/AIE/AIEMachineScheduler.cpp
+++ b/llvm/lib/Target/AIE/AIEMachineScheduler.cpp
@@ -80,6 +80,11 @@ static cl::opt<bool>
     InterBlockAlignment("aie-interblock-alignment", cl::init(true),
                         cl::desc("Allow for alignment of successor blocks"));
 
+namespace {
+// A sentinel value to represent an unknown SUnit.
+const constexpr unsigned UnknownSUNum = ~0;
+} // namespace
+
 static AIEHazardRecognizer *getAIEHazardRecognizer(const SchedBoundary &Zone) {
   return static_cast<AIEHazardRecognizer *>(Zone.HazardRec);
 }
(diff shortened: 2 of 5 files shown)


llvm/include/llvm/CodeGen/MachineScheduler.h
```diff
@@ -918,7 +918,7 @@ class SchedBoundary {
   };
 
   ScheduleDAGMI *DAG = nullptr;
-  const MachineSchedStrategy *SchedImpl = nullptr;
+  MachineSchedStrategy *SchedImpl = nullptr;
```
martien-de-jong, line 921: It makes sense, but it looks a bit fragile.

llvm/lib/Target/AIE/AIEMachineScheduler.cpp
```diff
@@ -80,6 +80,11 @@ static cl::opt<bool>
     InterBlockAlignment("aie-interblock-alignment", cl::init(true),
                         cl::desc("Allow for alignment of successor blocks"));
 
+namespace {
+// A sentinel value to represent an unknown SUnit.
+const constexpr unsigned UnknownSUNum = ~0;
```
martien-de-jong, line 85: Doesn't this conflict with BoundaryID?

llvm/lib/Target/AIE/AIEMachineScheduler.cpp
```diff
@@ -883,8 +890,34 @@ bool AIEPreRASchedStrategy::isAvailableNode(SUnit &SU, SchedBoundary &Zone,
...
+  std::function<bool(unsigned)> Impl = [&](unsigned SUNum) {
+    if (SUNum == UnknownSUNum)
+      return false;
+    if (SUNum == DelayedSU.NodeNum)
+      return true;
+    return Impl(SUDelayerMap[SUNum]);
+  };
+  // If SU is delayed by another instruction that is eventually waiting on SU
+  // itself, do not keep delaying SU otherwise this creates an infinite loop.
+  return !Impl(Delayer.NodeNum);
```
martien-de-jong: I can see it working for a cycle of two, but I'm not so sure about larger cycles. Could these occur? Oh, but you go recursive to check the entire chain, sorry.
Curious, why not make canBeDelayed recursive directly?

gbossu: I could imagine something like:
- An instr decreases vec pressure but increases acc pressure
- An instr decreases acc pressure but increases gpr pressure
- An instr decreases gpr pressure but increases vec pressure

llvm/include/llvm/CodeGen/MachineScheduler.h
gbossu: I still need to convince myself that it is fine to have a non-const `isAvailableNode()`, but in the end I think it makes sense to allow the SchedStrategies to keep some state so as to render better decisions given the global context of the region.

llvm/lib/Target/AIE/AIEMachineScheduler.cpp
```diff
@@ -883,8 +890,34 @@ bool AIEPreRASchedStrategy::isAvailableNode(SUnit &SU, SchedBoundary &Zone,
...
+}
+
+bool AIEPreRASchedStrategy::canBeDelayed(const SUnit &DelayedSU,
+                                         const SUnit &Delayer) const {
+  std::function<bool(unsigned)> Impl = [&](unsigned SUNum) {
+    if (SUNum == UnknownSUNum)
+      return false;
+    if (SUNum == DelayedSU.NodeNum)
+      return true;
+    return Impl(SUDelayerMap[SUNum]);
```
martien-de-jong, line 916: use .at() here?

andcarminati:
Hi @gbossu, this not really important, but I was trying to figure out the reason for the second commit message (fixup! [AIE2] Define target as out-of-order in .td)

llvm/lib/Target/AIE/AIEMachineScheduler.cpp
andcarminati: I think that you can flip the returns of the implementation. In this case you do not need to negate the main return. In this way the idea becomes more clear.

llvm/lib/Target/AIE/AIEMachineScheduler.cpp
martien-de-jong: Perhaps, but I don't think that would hurt.

llvm/test/CodeGen/AIE/aie2/schedule/pre_ra/reduce_pressure_cycle.mir
```diff
@@ -0,0 +1,50 @@
...
+    %6:vec1024 = COPY $y2
+    %11:acc1024 = COPY $y2
+    %12:acc1024 = COPY $y2
+    %13:acc1024 = COPY $y2
+    %14:acc1024 = COPY $y2
+    %15:acc1024 = COPY $y2
+    %16:acc1024 = COPY $y2
+    %17:acc1024 = COPY $y2
+    %18:acc1024 = COPY $y2
+    %19:acc1024 = COPY $y2
```
martien-de-jong, line 39: Carefully counted  each to reproduce the pressure situation I guess. How stable is this, should we randomize a bit around it? Perhaps give a rationale in comment of these counts?

llvm/include/llvm/CodeGen/MachineScheduler.h
martien-de-jong: I agree it makes sense. If it weren't for legacy in the code base you might want to separate query and state update. But whatever you do, there's probably some consts to be dropped, so I think its fine.

llvm/lib/Target/AIE/AIEMachineScheduler.cpp
martien-de-jong: paper, scissors, stone

martien-de-jong (dismissed):
LGTM.

llvm/test/CodeGen/AIE/aie2/schedule/pre_ra/reduce_pressure_cycle.mir
gbossu: Not so carefully, I'm just maxing out the number of available acc and vec registers, this way I am sure the pressure is at a critical stage.

gbossu:
> Hi @gbossu, this not really important, but I was trying to figure out the reason for the second commit message (fixup! [AIE2] Define target as out-of-order in .td)

This is the commit that introduced the `isAvailableNode` callback for AIE2, so I prefer to "patch" it instead of creating a "brand new" commit. When we try to upstream some of those scheduler changes, I hope it will be easier if we can just follow the `!fixup` commits to constitute a patch.

llvm/lib/Target/AIE/AIEMachineScheduler.cpp
gbossu: This is actually an `std::vector` 🙈 I just use it as a flat map.

martien-de-jong: std::vector has .at too, giving you bounds checks?

martien-de-jong (approved):

andcarminati (approved):

PR #72, opened 2024-06-10
[AIE2] Support broadcast_to_v16accfloat and broadcast_to_v16float

gbossu:
These "frontend intrinsics" were not properly supported and would crash the compiler if used. They now use __builtin_aiev2_vbroadcast32_I512 and are eventually selected to a VBCST_32 instruction. There is already an ISel test for this. See aie2/GlobalISel/inst-select-vbcst.mir.

### DIFF
--- a/clang/include/clang/Basic/BuiltinsAIE.def
+++ b/clang/include/clang/Basic/BuiltinsAIE.def
@@ -272,7 +272,6 @@ BUILTIN(__builtin_aiev2_vbroadcast64_I512, "V16iV2i", "nc")
 BUILTIN(__builtin_aiev2_vbroadcast_bf16_bf512, "V32yy", "nc")
 BUILTIN(__builtin_aiev2_vbroadcast_bf32_bf512, "V32yV2y", "nc")
 BUILTIN(__builtin_aiev2_vbroadcast_bf64_bf512, "V32yV4y", "nc")
-BUILTIN(__builtin_aiev2_vbroadcastfloat_I512, "V16ff", "nc")
 BUILTIN(__builtin_aiev2_vbroadcast_zero_acc1024, "V16m", "nc")
 //scl2vec::broadcast_elem
 BUILTIN(__builtin_aiev2_vextract_broadcast8_I512, "V64cV64ci", "nc")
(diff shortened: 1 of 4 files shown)


clang/include/clang/Basic/BuiltinsAIE.def
```diff
@@ -272,7 +272,6 @@ BUILTIN(__builtin_aiev2_vbroadcast64_I512, "V16iV2i", "nc")
 BUILTIN(__builtin_aiev2_vbroadcast_bf16_bf512, "V32yy", "nc")
 BUILTIN(__builtin_aiev2_vbroadcast_bf32_bf512, "V32yV2y", "nc")
 BUILTIN(__builtin_aiev2_vbroadcast_bf64_bf512, "V32yV4y", "nc")
```
gbossu, line 274: I'm tempted to also clean up the `bf` builtins and have our frontend intrinsics for `bf16` use the standard `__builtin_aiev2_vbroadcastXXXX_I512` builtins instead.

gbossu: @konstantinschwarz @SagarMaheshwari99 Give me a sign and I'll do it :)

clang/lib/Headers/aiev2_scl2vec.h
```diff
@@ -521,13 +521,15 @@ broadcast_to_v32bfloat16(v4bfloat16 b) {
 
 INTRINSIC(v16accfloat)
 broadcast_to_v16accfloat(float b) {
-  return __builtin_aiev2_vbroadcastfloat_I512(b);
+  int as_int = *reinterpret_cast<int *>(&b);
```
konstantinschwarz: nit: I'd prefer `__builtin_bit_cast(int, b);`

clang/include/clang/Basic/BuiltinsAIE.def
konstantinschwarz: We had a similar discussion about the vshuffle intrinsics. Happy to take a cleanup PR for these

konstantinschwarz (dismissed):

andcarminati (dismissed):

clang/lib/Headers/aiev2_scl2vec.h
gbossu: Same, I just forgot about it :/

gbossu: Updated

konstantinschwarz (approved):

PR #73, opened 2024-06-11
[AIE] Internalize non-kernel symbols.

SagarMaheshwari99:
This saves compilation time, size of the final
executable, and size of any intermediate dumps.

Run Internalize pass early in the opt pipeline followed by global DCE pass. To enable it RT can pass -aie-internalize-symbols option.

Edit: Also some refactoring.

### DIFF
--- a/llvm/lib/Target/AIE/AIE2TargetMachine.cpp
+++ b/llvm/lib/Target/AIE/AIE2TargetMachine.cpp
@@ -34,18 +34,11 @@
 #include "llvm/InitializePasses.h"
 #include "llvm/MC/TargetRegistry.h"
 #include "llvm/PassRegistry.h"
-#include "llvm/Passes/CodeGenPassBuilder.h"
-#include "llvm/Passes/PassBuilder.h"
 #include "llvm/Support/FormattedStream.h"
 #include "llvm/Target/TargetOptions.h"
 #include "llvm/Transforms/Scalar.h"
 
 using namespace llvm;
-
-static cl::opt<bool>
-    EnableCustomAliasAnalysis("aie-enable-alias-analysis",
-                              cl::desc("Enable AIE alias analysis pass"),
-                              cl::init(true), cl::Hidden);
 
 static cl::opt<bool>
     EnableStagedRA("aie-staged-ra", cl::Hidden, cl::init(true),
@@ -102,16 +95,10 @@ public:
   void addBlockPlacement() override;
   void addPreLegalizeMachineIR() override;
   void addPreRegBankSelect() override;
-  void addIRPasses() override;
 };
 
 TargetPassConfig *AIE2TargetMachine::createPassConfig(PassManagerBase &PM) {
   return new AIE2PassConfig(*this, PM);
-}
-
-void AIE2TargetMachine::registerDefaultAliasAnalyses(AAManager &AAM) {
-  if (EnableCustomAliasAnalysis)
-    AAM.registerFunctionAnalysis<AIEBaseAA>();
 }
 
 bool AIE2PassConfig::addPreISel() {
@@ -121,26 +108,6 @@ bool AIE2PassConfig::addPreISel() {
 }
 
 void AIE2PassConfig::addPreEmitPass() {}
-
-void AIE2PassConfig::addIRPasses() {
-  // Always expand atomic operations, we don't deal with atomicrmw or cmpxchg
-  // ourselves.
-  addPass(createAtomicExpandLegacyPass());
-
-  if (TM->getOptLevel() > CodeGenOptLevel::None) {
-
-    if (EnableCustomAliasAnalysis) {
-      addPass(createAIEBaseAAWrapperPass());
-      addPass(
-          createExternalAAWrapperPass([](Pass &P, Function &, AAResults &AAR) {
-            if (auto *WrapperPass =
-                    P.getAnalysisIfAvailable<AIEBaseAAWrapperPass>())
-              AAR.addAAResult(WrapperPass->getResult());
-          }));
-    }
-  }
-  TargetPassConfig::addIRPasses();
-}
 
 void AIE2PassConfig::addPreLegalizeMachineIR() {
   addPass(createAIEAddressSpaceFlattening());
@@ -283,20 +250,4 @@ bool AIE2PassConfig::addInstSelector() {
   if (AIEDumpArtifacts)
     addPass(createMachineFunctionDumperPass(/*Suffix=*/"after-isel"));
   return false;
-}
-
-void AIE2TargetMachine::registerPassBuilderCallbacks(
-    PassBuilder &PB, bool PopulateClassToPassNames) {
-  if (EnableCustomAliasAnalysis) {
-    PB.registerAnalysisRegistrationCallback([](FunctionAnalysisManager &FAM) {
-      FAM.registerPass([&] { return AIEBaseAA(); });
-    });
-    PB.registerParseAACallback([](StringRef AAName, AAManager &AAM) {
-      if (AAName == "aie-aa") {
-        AAM.registerFunctionAnalysis<AIEBaseAA>();
-        return true;
-      }
-      return false;
-    });
-  }
 }
--- a/llvm/lib/Target/AIE/AIETargetMachine.cpp
+++ b/llvm/lib/Target/AIE/AIETargetMachine.cpp
@@ -15,6 +15,7 @@
 #include "AIETargetMachine.h"
 #include "AIE.h"
 #include "AIE2TargetMachine.h"
+#include "AIEBaseAliasAnalysis.h"
 #include "AIEDumpArtifacts.h"
 #include "AIEFinalizeBundle.h"
 #include "AIEMachineBlockPlacement.h"
@@ -38,10 +39,16 @@
 #include "llvm/InitializePasses.h"
 #include "llvm/MC/TargetRegistry.h"
 #include "llvm/PassRegistry.h"
+#include "llvm/Passes/PassBuilder.h"
 #include "llvm/Support/CodeGen.h"
 #include "llvm/Support/FormattedStream.h"
 #include "llvm/Target/TargetOptions.h"
 using namespace llvm;
+
+static cl::opt<bool>
+    EnableCustomAliasAnalysis("aie-enable-alias-analysis",
+                              cl::desc("Enable AIE alias analysis pass"),
+                              cl::init(true), cl::Hidden);
 
 extern bool AIEDumpArtifacts;
 
@@ -118,7 +125,21 @@ TargetPassConfig *AIETargetMachine::createPassConfig(PassManagerBase &PM) {
 }
 
 void AIEPassConfig::addIRPasses() {
+  // Always expand atomic operations, we don't deal with atomicrmw or cmpxchg
+  // ourselves.
   addPass(createAtomicExpandLegacyPass());
+
+  if (TM->getOptLevel() > CodeGenOptLevel::None) {
+    if (EnableCustomAliasAnalysis) {
+      addPass(createAIEBaseAAWrapperPass());
+      addPass(
+          createExternalAAWrapperPass([](Pass &P, Function &, AAResults &AAR) {
+            if (auto *WrapperPass =
+                    P.getAnalysisIfAvailable<AIEBaseAAWrapperPass>())
+              AAR.addAAResult(WrapperPass->getResult());
+          }));
+    }
+  }
   TargetPassConfig::addIRPasses();
 }
 
@@ -236,4 +257,24 @@ std::unique_ptr<CSEConfigBase> AIEPassConfig::getCSEConfig() const {
   if (TM->getOptLevel() == CodeGenOptLevel::None)
     return std::make_unique<CSEConfigBase>();
   return getStandardCSEConfigForOpt(TM->getOptLevel());
+}
+
+void AIEBaseTargetMachine::registerDefaultAliasAnalyses(AAManager &AAM) {
+  if (EnableCustomAliasAnalysis)
+    AAM.registerFunctionAnalysis<AIEBaseAA>();
+}
+void AIEBaseTargetMachine::registerPassBuilderCallbacks(
+    PassBuilder &PB, bool PopulateClassToPassNames) {
+  if (EnableCustomAliasAnalysis) {
+    PB.registerAnalysisRegistrationCallback([](FunctionAnalysisManager &FAM) {
+      FAM.registerPass([&] { return AIEBaseAA(); });
+    });
+    PB.registerParseAACallback([](StringRef AAName, AAManager &AAM) {
+      if (AAName == "aie-aa") {
+        AAM.registerFunctionAnalysis<AIEBaseAA>();
+        return true;
+      }
+      return false;
+    });
+  }
 }
--- a/llvm/lib/Target/AIE/AIETargetMachine.cpp
+++ b/llvm/lib/Target/AIE/AIETargetMachine.cpp
@@ -43,12 +43,27 @@
 #include "llvm/Support/CodeGen.h"
 #include "llvm/Support/FormattedStream.h"
 #include "llvm/Target/TargetOptions.h"
+#include "llvm/Transforms/IPO/GlobalDCE.h"
+#include "llvm/Transforms/IPO/Internalize.h"
 using namespace llvm;
 
 static cl::opt<bool>
     EnableCustomAliasAnalysis("aie-enable-alias-analysis",
                               cl::desc("Enable AIE alias analysis pass"),
                               cl::init(true), cl::Hidden);
+
+// Option to run internalize pass.
+static cl::opt<bool> InternalizeSymbols(
+    "aie-internalize-symbols",
+    cl::desc("Enable elimination of non-kernel functions and unused globals"),
+    cl::init(false), cl::Hidden);
+
+// Option to skip the functions we don't want to internalize.
+static cl::list<std::string>
+    FunctionSkipList("aie-internalize-skip-functions",
+                     cl::desc("List of function names to skip internalization"),
+                     cl::Hidden, cl::list_init<std::string>({"main"}),
+                     cl::CommaSeparated);
 
 extern bool AIEDumpArtifacts;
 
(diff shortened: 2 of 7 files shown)


andcarminati:
Hi @SagarMaheshwari99, is there any other specific use apart of time/space reduction? I think that this cause problems at linking stage.

SagarMaheshwari99:
> Hi @SagarMaheshwari99, is there any other specific use apart of time/space reduction? I think that this cause problems at linking stage.

@andcarminati The current use is only time/space reduction. Can you please elaborate on what kind of problems can be caused at the linking stage?

llvm/lib/Target/AIE/CMakeLists.txt
```diff
@@ -114,6 +114,7 @@ add_llvm_target(AIECodeGen
    Core
    CodeGen
    CodeGenTypes
+   IPO
```
martien-de-jong, line 120: Check: this is the library where the internalize pass lives?

llvm/lib/Target/AIE/AIE2TargetMachine.cpp
```diff
@@ -82,6 +90,15 @@ AIE2TargetMachine::AIE2TargetMachine(const Target &T, const Triple &TT,
   setGlobalISelAbort(GlobalISelAbortMode::Enable);
 }
 
+/// Predicate for Internalize pass.
+static bool mustPreserveGV(const GlobalValue &GV) {
+  if (const Function *F = dyn_cast<Function>(&GV))
+    return F->isDeclaration() || F->getName().starts_with("_main");
```
martien-de-jong: I just pushed a PR to change this to "main" in the startup code. I don't know the rationale for _main, nor did @konstantinschwarz. Everything would look more familiar if we could just use main, like the rest of the world.

andcarminati:
> > Hi @SagarMaheshwari99, is there any other specific use apart of time/space reduction? I think that this cause problems at linking stage.
> 
> @andcarminati The current use is only time/space reduction. Can you please elaborate on what kind of problems can be caused at the linking stage?

Hi @SagarMaheshwari99 , I was thinking in a work flow where the final executable Elf would be created by linking several different object files with different symbols. If this option is off by default then the user can use it for self contained CUs. For space reduction, another alternative we could think is the default use of `-ffunction-sections` and `-fdata-sections` in the frontend plus linker garbage collection.

llvm/lib/Target/AIE/AIE2TargetMachine.cpp
andcarminati: Hi, maybe we can have a option like this for flexibility:
```
cl::opt<std::string> EntryFunctionName ...
```

konstantinschwarz: Having an option for this sounds good! It could default to `main`, but would allow users to specify their own entry function

konstantinschwarz: Instead of `.starts_with`, change to `equals`? Otherwise we might get false positives

konstantinschwarz: Instead of .starts_with, change to equals? Otherwise we might get false positives

llvm/lib/Target/AIE/CMakeLists.txt
SagarMaheshwari99: Yes, and GlobalDCE.

llvm/lib/Target/AIE/AIE2TargetMachine.cpp
SagarMaheshwari99: `.starts_with` was also useful for `_main_init`... so I didn't change it. But I think a CLI option to add a list of functions is convenient... 

@martien-de-jong : The _main was used because "our internal linker" expected it (somewhere while creating object, the internal compiler converted main to _main) and we had no way other than use _main directly in the cc file. I agree to change it conditionally for the lld flow back to `main` in aiecompiler. 

I am thinking of also removing the extern "C", since we should be able to handle the mangling? Any opinions?

SagarMaheshwari99:
> > > Hi @SagarMaheshwari99, is there any other specific use apart of time/space reduction? I think that this cause problems at linking stage.
> > 
> > 
> > @andcarminati The current use is only time/space reduction. Can you please elaborate on what kind of problems can be caused at the linking stage?
> 
> Hi @SagarMaheshwari99 , I was thinking in a work flow where the final executable Elf would be created by linking several different object files with different symbols. If this option is off by default then the user can use it for self contained CUs. For space reduction, another alternative we could think is the default use of `-ffunction-sections` and `-fdata-sections` in the frontend plus linker garbage collection.

Hi @andcarminati, Yes, we should enable function and data sections in frontend and garbage collection in the linker by default. But the internalize pass was also useful to save up on the compile time plus help make debugging less noisy...

llvm/lib/Target/AIE/AIE2TargetMachine.cpp
SagarMaheshwari99: Made the changes.

konstantinschwarz: I think its sensible for the `FunctionSkipList` to contain `main` by default. Now we require both options to be set

llvm/lib/Target/AIE/AIE2TargetMachine.cpp
```diff
@@ -64,6 +66,17 @@ static cl::opt<bool> EnablePreMISchedCoalescer(
...
     cl::desc("Run the coalescer again after the pre-RA scheduler"));
 
+// Option to run internalize pass.
+static cl::opt<bool> InternalizeSymbols(
+    "aie2-internalize-symbols",
+    cl::desc("Enable elimination of non-kernel functions and unused globals"),
+    cl::init(false), cl::Hidden);
+// Option to skip the functions we don't want to internalize.
+static cl::list<std::string>
+    FunctionSkipList("aie2-internalize-skip-function",
```
andcarminati: Maybe `aie2-internalize-skip-functions`

llvm/lib/Target/AIE/AIE2TargetMachine.cpp
```diff
@@ -64,6 +66,17 @@ static cl::opt<bool> EnablePreMISchedCoalescer(
...
+                     cl::desc("List of symbols to skip internalization"),
```
andcarminati: I think it is better "List of function names to skip internalization".

llvm/test/CodeGen/AIE/aie2/internalize.ll
```diff
@@ -0,0 +1,58 @@
+; RUN: opt -O1 -S -mtriple=aie2-none-unknown-elf -aie2-internalize-symbols -aie2-internalize-skip-function="_main" < %s | FileCheck %s
```
andcarminati: I think we can have an additional test considering two function in the list.

SagarMaheshwari99: Added..

llvm/test/CodeGen/AIE/aie2/internalize.ll
```diff
@@ -0,0 +1,123 @@
+; RUN: opt -O1 -S -mtriple=aie2-none-unknown-elf -aie2-internalize-symbols < %s | FileCheck --check-prefix=ONLYMAIN %s
+; RUN: opt -O1 -S -mtriple=aie2-none-unknown-elf -aie2-internalize-symbols -aie2-internalize-skip-functions="skipfunc1" -aie2-internalize-skip-functions="skipfunc2" -aie2-internalize-skip-functions="skipfunc3" < %s | FileCheck --check-prefix=FUNCLIST %s
```
konstantinschwarz: Maybe use `cl::CommaSeparated` to keep the command line shorter?
i.e. allow to pass the symbols as
```
-aie2-internalize-skip-functions="skipfunc1,skipfunc2,skipfunc3"
```

llvm/lib/Target/AIE/AIE2TargetMachine.cpp
```diff
@@ -64,6 +66,17 @@ static cl::opt<bool> EnablePreMISchedCoalescer(
     "aie-premisched-coalescer", cl::Hidden, cl::init(true),
     cl::desc("Run the coalescer again after the pre-RA scheduler"));
 
+// Option to run internalize pass.
+static cl::opt<bool> InternalizeSymbols(
+    "aie2-internalize-symbols",
+    cl::desc("Enable elimination of non-kernel functions and unused globals"),
+    cl::init(false), cl::Hidden);
```
gbossu: Do we have a common pass pipeline that can be shared between different AIE versions?

SagarMaheshwari99: I am not sure. They ultimately derive from TargetPassConfig, so maybe we could have a generic one...

gbossu: Maybe have a chat with @abnikant ? It would be sad to just duplicate all those files for every AIE version.

llvm/lib/Target/AIE/AIE2TargetMachine.cpp
```diff
@@ -82,6 +96,19 @@ AIE2TargetMachine::AIE2TargetMachine(const Target &T, const Triple &TT,
   setGlobalISelAbort(GlobalISelAbortMode::Enable);
 }
 
+/// Predicate for Internalize pass.
```
gbossu: Nit: I feel the comment does not help much. Instead, maybe give a rationale about what should be preserved and why?

llvm/lib/Target/AIE/AIE2TargetMachine.cpp
```diff
@@ -64,6 +66,18 @@ static cl::opt<bool> EnablePreMISchedCoalescer(
...
+// Option to run internalize pass.
+static cl::opt<bool> InternalizeSymbols(
+    "aie2-internalize-symbols",
+    cl::desc("Enable elimination of non-kernel functions and unused globals"),
+    cl::init(false), cl::Hidden);
+// Option to skip the functions we don't want to internalize.
+static cl::list<std::string>
+    FunctionSkipList("aie2-internalize-skip-functions",
+                     cl::desc("List of function names to skip internalization"),
+                     cl::Hidden, cl::list_init<std::string>({"main"}),
```
gbossu: We sometimes also have a `_main` symbol. I guess we don't have to include it here because `main` is the real entry point?

llvm/test/CodeGen/AIE/aie2/internalize.ll
```diff
@@ -0,0 +1,123 @@
...
+; ONLYMAIN-NOT: define{{.*}}void @func_used_skipfunc1
+; FUNCLIST: define{{.*}}void @func_used_skipfunc1
+define void @func_used_skipfunc1(ptr addrspace(1) %out, i32 %val) #0 {
+entry:
+  store i32 %val, ptr addrspace(1) %out
+  ret void
+}
+
+; ONLYMAIN-NOT: define{{.*}}void @skipfunc1()
+; ONLYMAIN-NOT: tail call fastcc void @func_used_skipfunc1
```
gbossu: Maybe just `call{{.*}}void @func_used_skipfunc1`? I'm nervous about all those attributes that can change when we bump the LLVM version. (Same for other check lines)

llvm/test/CodeGen/AIE/aie2/internalize.ll
```diff
@@ -0,0 +1,123 @@
...
+; ONLYMAIN-NOT: ret void
+; FUNCLIST: define{{.*}}void @skipfunc3()
+; FUNCLIST: tail call fastcc void @func_used_skipfunc3
+; FUNCLIST: ret void
+define void @skipfunc3() {
+entry:
+    %val = add i32 42, 0
+    tail call fastcc void @func_used_skipfunc3(ptr addrspace(1) @gvar_used, i32 %val)
+    ret void
+}
```
gbossu: Maybe add a test where the address of a function escapes (e.g. it gets stored somewhere)?

llvm/lib/Target/AIE/AIE2TargetMachine.cpp
abnikant: In the aie-private branch the AIE2PPassConfig derives from AIE2PassConfig. Most of passes are shared between AIE2 and AIE2p.
class AIE2PassConfig : public AIEPassConfig {
}
class AIEPassConfig : public TargetPassConfig {
}

I suppose you can setup your pass in AIEPassConfig.

llvm/lib/Target/AIE/AIE2TargetMachine.cpp
SagarMaheshwari99: When using the peano flow, the `main` will be the "real" entry point(from today). In our other flow, we have `_main` as the entry point bcoz that's what the linker expects... For internalizing _main, we need to pass it on the command line. This will override `main` as the entry point.

llvm/test/CodeGen/AIE/aie2/internalize.ll
SagarMaheshwari99: yeah i should probably change that.

llvm/lib/Target/AIE/AIE2TargetMachine.cpp
SagarMaheshwari99: Refactored the code to be in AIETargetMachine.cpp

llvm/lib/Target/AIE/AIETargetMachine.cpp
```diff
@@ -43,13 +43,27 @@
...
     EnableCustomAliasAnalysis("aie-enable-alias-analysis",
                               cl::desc("Enable AIE alias analysis pass"),
                               cl::init(true), cl::Hidden);
 
+// Option to run internalize pass.
+static cl::opt<bool> InternalizeSymbols(
+    "aie-internalize-symbols",
+    cl::desc("Enable elimination of non-kernel functions and unused globals"),
+    cl::init(false), cl::Hidden);
+// Option to skip the functions we don't want to internalize.
```
gbossu, line 61: Nit: Skip a line for readability?

llvm/lib/Target/AIE/AIETargetMachine.cpp
```diff
@@ -261,6 +275,26 @@ void AIEBaseTargetMachine::registerDefaultAliasAnalyses(AAManager &AAM) {
...
     AAM.registerFunctionAnalysis<AIEBaseAA>();
 }
+
+/// Predicate for Internalize pass.
+/// Preserve functions for names passed via FunctionSkipList. These are usually
+/// entry points which shouldn't be internalized to let linker know of them
+/// (default entry point is 'main'). Preserve function declarations, as if we
+/// internalize a declaration, it would not be accessible outside of its
+/// translation unit, which could lead to linker errors if the function is
+/// used/defined elsewhere.
```
gbossu: I think the comment is a bit convoluted, maybe simply write `Preserve functions that can be an entry point or that have uses within the Module`?

llvm/test/CodeGen/AIE/internalize.ll
```diff
@@ -0,0 +1,126 @@
...
+; ONLYMAIN: define{{.*}}void @main()
+; ONLYMAIN: call{{.*}}void @func_used_noinline
+; ONLYMAIN: store volatile
+; ONLYMAIN: call{{.*}}void @decl_func()
+; ONLYMAIN: ret void
+; FUNCLIST-NOT: define{{.*}}void @main()
+; FUNCLIST-NOT: call{{.*}}void @func_used_noinline
+; FUNCLIST-NOT: store volatile
+; FUNCLIST-NOT: call{{.*}}void @decl_func()
+; FUNCLIST-NOT: ret void
```
gbossu: Nit: I think just `; FUNCLIST-NOT: define{{.*}}void @main()` is fine

llvm/test/CodeGen/AIE/internalize.ll
```diff
@@ -0,0 +1,126 @@
...
+
+; ONLYMAIN-NOT: define{{.*}}void @func_used_skipfunc3
+; FUNCLIST: define{{.*}}void @func_used_skipfunc3
+define void @func_used_skipfunc3(ptr addrspace(1) %out, i32 %val) #0 {
+entry:
+  store i32 %val, ptr addrspace(1) %out
+  ret void
+}
+
+; ONLYMAIN-NOT: define{{.*}}void @skipfunc3()
```
gbossu: Nit: Same, I think when we expect a function **not** to be there, something like `CHECK-NOT: define{{.*}} @thefunction` is enough

gbossu (dismissed):
LGTM, please address comments as you please. Did you verify QoR? I'd be curious about PM size reduction.

llvm/test/CodeGen/AIE/internalize.ll
SagarMaheshwari99: Done

llvm/test/CodeGen/AIE/internalize.ll
SagarMaheshwari99: done

llvm/lib/Target/AIE/AIETargetMachine.cpp
SagarMaheshwari99: done

llvm/lib/Target/AIE/AIETargetMachine.cpp
SagarMaheshwari99: done

SagarMaheshwari99:
> LGTM, please address comments as you please. Did you verify QoR? I'd be curious about PM size reduction.

I did run it a while back but I lost the results, I will run it again.

llvm/lib/Target/AIE/AIETargetMachine.cpp
```diff
@@ -38,11 +39,32 @@
...
+static cl::opt<bool>
+    EnableCustomAliasAnalysis("aie-enable-alias-analysis",
+                              cl::desc("Enable AIE alias analysis pass"),
+                              cl::init(true), cl::Hidden);
+
+// Option to run internalize pass.
+static cl::opt<bool> InternalizeSymbols(
+    "aie-internalize-symbols",
+    cl::desc("Enable elimination of non-kernel functions and unused globals"),
+    cl::init(false), cl::Hidden);
```
gbossu, line 59: As discussed offline, I think this should be enabled by default.

gbossu: Discussed offline, static libraries can be a problem. We should only internalize and simplify if we are producing an "final executable". Let's keep it disabled by default.

konstantinschwarz (approved):
All downstream issues got resolved, let's land this!

PR #74, opened 2024-06-13
Provide default ldfile; call main, not _main

martien-de-jong:
Your favourite "hello world" can now be compiled with

`clang --target=aie2-none-unknown-elf <path-to-ldfile> hello.c -o hello`

provided you don't call printf; it won't fit in program memory.

### DIFF
--- a/libc/startup/baremetal/aie2/crt1.cc
+++ b/libc/startup/baremetal/aie2/crt1.cc
@@ -6,10 +6,10 @@
 // (c) Copyright 2024 Advanced Micro Devices, Inc. or its affiliates
 
 extern "C" {
-extern int _main(int, char **);
+extern int main(int, char **);
 void _Exit(int val) {
   (void)val;
   done();
 }
-void _main_init() { _Exit(_main(0, nullptr)); }
+void _main_init() { _Exit(main(0, nullptr)); }
 }
--- a/libc/startup/baremetal/aie2/ldfile
+++ b/libc/startup/baremetal/aie2/ldfile
@@ -1,0 +1,59 @@
+/*
+ * This file is licensed under the Apache License v2.0 with LLVM Exceptions.
+ * See https://llvm.org/LICENSE.txt for license information.
+ * SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+ *
+ * (c) Copyright 2024 Advanced Micro Devices, Inc. or its affiliates
+ */
+
+/* This is an example linker script that defines a simple memory layout for
+ * a binary running on aie2. It can be passed as a source file on clang's
+ * command line; it will pass it on to ld.lld, which interprets it as a
+ * linker script.
+ */
+
+/*
+ * The extent of program and data memory
+ */
+MEMORY
+{
+   program (RX) : ORIGIN = 0, LENGTH = 0x0020000
+   data (!RX) : ORIGIN = 0x20000, LENGTH = 0x0020000
+}
+
+/*
+ * We allow more than one text, data, etc section, provided that they
+ * have the right prefix '.text' etc.
+ * We force the start of the text section to address 0x0.
+ * The standard clang compiler driver supplies crt0.o and crt1.o which together
+ * define the C/C++ startup.
+ * crt0.o defines __start, the low-level entry point of execution.
+ * It initializes the stack pointer from the symbol _sp_start_value_DM_stack
+ * defined here and calls _main_init.
+ * crt1.o defines _main_init, which calls _Exit(main(0, nullptr));
+ * _Exit should not return
+ */
+ENTRY(__start)
+SECTIONS
+{
+  . = 0x0;
+  .text : { 
+    *crt0.o(.text*)
+    *(.text*)
+  } > program
+
+  .stack : {
+    . = 0x20000;
+    _sp_start_value_DM_stack = .;
+    . += 0x800;
+  } > data
+  .rodata : {
+    *(.rodata*)
+  } > data
+  .data : {
+    *(.data*)
+  } > data
+  .bss : {
+    *(.bss*)
+  } > data
+}
--- a/libc/startup/baremetal/aie2/crt0.S
+++ b/libc/startup/baremetal/aie2/crt0.S
@@ -9,9 +9,8 @@
 .global _sp_start_value_DM_stack
 .global _main_init
 __start:
-        MOVXM sp, #_sp_start_value_DM_stack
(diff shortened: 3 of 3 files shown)


libc/startup/baremetal/aie2/ldfile
```diff
@@ -0,0 +1,37 @@
...
+  . = 0x0;
+  .text : { 
+    *crt0.o(.text)
+    *(.text)
+  } > program
+
+  .stack : {
+    . = 0x20000;
+    _sp_start_value_DM_stack = .;
+    . += 0x2000;
```
martien-de-jong: Hmm. this is more than 2k. Anyway, it's just a number that can be adjusted.

libc/startup/baremetal/aie2/ldfile
```diff
@@ -0,0 +1,37 @@
...
+ *
+ * (c) Copyright 2024 Advanced Micro Devices, Inc. or its affiliates
+ */
+
+MEMORY
+{
+   program (RX) : ORIGIN = 0, LENGTH = 0x0020000
+   data (!RX) : ORIGIN = 0x20000, LENGTH = 0x0020000
+}
+ENTRY(_main_init)
```
SagarMaheshwari99: CHECK : shouldn't the entry be `__start` ?

libc/startup/baremetal/aie2/ldfile
```diff
@@ -0,0 +1,37 @@
...
+SECTIONS
+{
+  . = 0x0;
+  .text : { 
+    *crt0.o(.text)
```
SagarMaheshwari99: don't we also need `*crt1.o(.text)` ?

libc/startup/baremetal/aie2/ldfile
```diff
@@ -0,0 +1,37 @@
...
+  .stack : {
+    . = 0x20000;
+    _sp_start_value_DM_stack = .;
+    . += 0x2000;
+  } > data
+  .rodata : {
+    *(.rodata.*)
+  } > data
+  .data : {
+    *(.data)
```
SagarMaheshwari99: I think this should be *(.data`*`) ?

SagarMaheshwari99: I am not sure if we require `*` for .bss as well. Safe to keep it?

libc/startup/baremetal/aie2/ldfile
SagarMaheshwari99: Also, `*` after the `.text`s ?

SagarMaheshwari99 (dismissed):

libc/startup/baremetal/aie2/ldfile
martien-de-jong: Well spotted! Yes, dead wrong, this.

libc/startup/baremetal/aie2/ldfile
martien-de-jong: We only need crt0.o to be in place, for loaders that expect the program to start at address 0. crt1 can be anywhere, it will be jumped to. Some might want to rely on fallthrough, not me.
Wildcarding text sections sounds like a good idea. My thought was that this is a vanilla for a vanilla invocation, and everything gets directed to .text if you don't pass extra options to the compiler.

libc/startup/baremetal/aie2/ldfile
martien-de-jong: I'll add wildcards everywhere. I think we have different sections for different alignments.

SagarMaheshwari99 (dismissed):

andcarminati:
Hi @martien-de-jong, for me it it is not clear how the linker script is forwarded to the linker. How it differs from the common `clang  -Wl,-T script.ld` ?

martien-de-jong:
> Hi @martien-de-jong, for me it it is not clear how the linker script is forwarded to the linker. How it differs from the common `clang -Wl,-T script.ld` ?

It's not forwarded. it is there for reference only. I think you can just add it without any options on the clang commandline; it will pass it on to ld.lld, which will interpret it as an ldfile by default (it has no extension)

andcarminati:
> > Hi @martien-de-jong, for me it it is not clear how the linker script is forwarded to the linker. How it differs from the common `clang -Wl,-T script.ld` ?
> 
> It's not forwarded. it is there for reference only. I think you can just add it without any options on the clang commandline; it will pass it on to ld.lld, which will interpret it as an ldfile by default (it has no extension)

Understood, thank you!

libc/startup/baremetal/aie2/ldfile
stephenneuendorffer: Note that this needs to be a published standard, because other tools need to generate ldfiles.   So you get to change it.... once.  Then document it and set it in stone.

libc/startup/baremetal/aie2/crt1.cc
```diff
@@ -6,10 +6,10 @@
 // (c) Copyright 2024 Advanced Micro Devices, Inc. or its affiliates
 
 extern "C" {
-extern int _main(int, char **);
+extern int main(int, char **);
 void _Exit(int val) {
   (void)val;
   done();
 }
-void _main_init() { _Exit(_main(0, nullptr)); }
```
stephenneuendorffer, line 14: Is this compatible with existing code?  My recollection here is having to do some awkward things to get the same code to compile with existing tools.

libc/startup/baremetal/aie2/ldfile
martien-de-jong: I'm not afraid to document __start as entry point, which will be recognizable by unix system programmers. It is actually only documenting crt0.o. If you write your own ldfile, you can also choose to supply your own startup code and place it at address 0x0. And even 0x0 is only necessary if you want to start executing at zero from hardware reset, which would require ROM PM.
I will add some comments here as documentation.

libc/startup/baremetal/aie2/crt1.cc
martien-de-jong: The main problem I see would be the organization of static constructors, which are currently left out. In this set-up we control both startup code and _Exit, and compatibility hinges on C 'main' being compiled to link 'main'.
What I have seen is that the structure of static constructors is compatible, i.e. an array of pointers to void(void) initialization routines. When we want to add them I hope to glue them together by a symbol equate in the ldfile.
I don't think we want to add argc/argv at this point.
If there are any insights I've missed, please let me know.

SagarMaheshwari99 (approved):

PR #75, opened 2024-06-13
[AIE] Pass -nostdlib to the Peano builtins and runtimes build

konstantinschwarz:
When building Peano with a custom linker, e.g. ld.lld, the option LLVM_ENABLE_LLD is also passed to the runtimes build. That triggers a CMake compile check while configuring the runtimes build, which fails because we try to link the standard libraries.

This patch adds the -nostdlib option to the runtimes linker options, allowing the check to succeed.

### DIFF
--- a/clang/cmake/caches/Peano-AIE-runtime-libraries.cmake
+++ b/clang/cmake/caches/Peano-AIE-runtime-libraries.cmake
@@ -25,6 +25,13 @@ foreach(target ${LLVM_BUILTIN_TARGETS})
   set(BUILTINS_${target}_CMAKE_C_FLAGS "--target=${target}" CACHE STRING "")
   set(BUILTINS_${target}_CMAKE_CXX_FLAGS "--target=${target}" CACHE STRING "")
   set(BUILTINS_${target}_CMAKE_ASM_FLAGS "--target=${target}" CACHE STRING "")
+  # The runtimes CMake system passes the LLVM_USE_LINKER option set for the host
+  # compilation down to the runtimes build. That triggers a compile & link check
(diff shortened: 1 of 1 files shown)


clang/cmake/caches/Peano-AIE-runtime-libraries.cmake
```diff
@@ -25,6 +25,10 @@ foreach(target ${LLVM_BUILTIN_TARGETS})
   set(BUILTINS_${target}_CMAKE_C_FLAGS "--target=${target}" CACHE STRING "")
   set(BUILTINS_${target}_CMAKE_CXX_FLAGS "--target=${target}" CACHE STRING "")
   set(BUILTINS_${target}_CMAKE_ASM_FLAGS "--target=${target}" CACHE STRING "")
+  # Pass -nostdlib to linker flags. The standard libraries don't exist yet,
+  # prevent linking them
```
martien-de-jong: I would like to understand from this comment why someone would try to link at all when building libraries. I *think* that is because of a misconstrued use of LLVM_ENABLE_LLD. To me it makes sense to possibly use it for building library generation tooling, but then it should be referring to the host linker, not the cross linker.
Therefore, mention this CMake quirk in the comment?

konstantinschwarz: I mentioned it in the commit message. Would you prefer to have that explanation in the code?

martien-de-jong: I think I would. As I said, this rather terse comment raises more questions than it answers.

martien-de-jong (approved):

PR #76, opened 2024-06-15
[AIE2] Combiners for 8x8->8x8 and 8x4->4x8 matrix transposes

ValentijnvdBeek:
This merge requests adds the same framework for Shuffle Vector combinations used in #41 to the AIE2 backend. It also defines a new generator that generates the pattern used for a decent for subset of the shuffle vector modes. Concretely it matches mode 35 (4x4 -> 4x4 matrix transpose) and 29 (8x4->4x8 matrix transpose). Finally, it adds a generic opcode for the AIE vshuffle instruction.


llvm/test/CodeGen/AIE/aie2/GlobalISel/prelegalizercombiner-shufflevector.mir
```diff
@@ -642,4 +642,69 @@ body:             |
...
+legalized:       false
+body:             |
+  bb.1.entry:
+    liveins: $x0, $x1
+    ; CHECK-LABEL: name: shuffle_vector_32
+    ; CHECK: liveins: $x0, $x1
+    ; CHECK-NEXT: {{  $}}
+    ; CHECK-NEXT: [[COPY:%[0-9]+]]:_(<16 x s32>) = COPY $x0
+    ; CHECK-NEXT: [[COPY1:%[0-9]+]]:_(<16 x s32>) = COPY $x1
+    ; CHECK-NEXT: [[C:%[0-9]+]]:_(s32) = G_CONSTANT i32 35
```
konstantinschwarz: I think this should be mode 34 (Element Size 32-bit, 4x4 -> 4x4 transpose)?

llvm/test/CodeGen/AIE/aie2/GlobalISel/prelegalizercombiner-shufflevector.mir
```diff
@@ -642,4 +642,69 @@ body:             |
...
+legalized:       false
+body:             |
+  bb.1.entry:
+    liveins: $x0, $x1
+    ; CHECK-LABEL: name: shuffle_vector_16
+    ; CHECK: liveins: $x0, $x1
+    ; CHECK-NEXT: {{  $}}
+    ; CHECK-NEXT: [[COPY:%[0-9]+]]:_(<32 x s16>) = COPY $x0
+    ; CHECK-NEXT: [[COPY1:%[0-9]+]]:_(<32 x s16>) = COPY $x1
+    ; CHECK-NEXT: [[C:%[0-9]+]]:_(s32) = G_CONSTANT i32 35
```
konstantinschwarz: This should be mode 39 (Element Size 16-bit, 4x4 -> 4x4 transpose)
However, the mask should be the same as in `shuffle_vector_32`

llvm/test/CodeGen/AIE/aie2/GlobalISel/prelegalizercombiner-shufflevector.mir
```diff
@@ -642,4 +642,69 @@ body:             |
...
+    ; CHECK-NEXT: {{  $}}
+    ; CHECK-NEXT: [[COPY:%[0-9]+]]:_(<64 x s8>) = COPY $x0
+    ; CHECK-NEXT: [[COPY1:%[0-9]+]]:_(<64 x s8>) = COPY $x1
+    ; CHECK-NEXT: [[C:%[0-9]+]]:_(s32) = G_CONSTANT i32 35
+    ; CHECK-NEXT: [[AIE_VSHUFFLE:%[0-9]+]]:_(<64 x s8>) = G_AIE_VSHUFFLE [[COPY]], [[COPY1]], [[C]](s32)
+    ; CHECK-NEXT: $x2 = COPY [[AIE_VSHUFFLE]](<64 x s8>)
+    ; CHECK-NEXT: PseudoRET implicit $lr, implicit $x2
+    %1:_(<64 x s8>) = COPY $x0
+    %2:_(<64 x s8>) = COPY $x1
+    %0:_(<64 x s8>) = G_SHUFFLE_VECTOR %1:_(<64 x s8>), %2:_, shufflemask(0, 16, 32, 48, 1, 17, 33, 49, 2, 18, 34, 50, 3, 19, 35, 51, 4, 20, 36, 52, 5, 21, 37, 53, 6, 22, 38, 54, 7, 23, 39, 55, 8, 24, 40, 56, 9, 25, 41, 57, 10, 26, 42, 58, 11, 27, 43, 59, 12, 28, 44, 60, 13, 29, 45, 61, 14, 30, 46, 62, 15, 31, 47, 63)
```
konstantinschwarz, line 938: Mode 35 operates on a 8-bit element 8x8 matrix.
That would need to match `shufflemask(0, 8, 16, 24, 1, 9, 17, 25, ...)`?

konstantinschwarz (changes requested):
The shuffle modes description is difficult to read at first. Each element type and matrix dimension uses its own shuffle mode.

ValentijnvdBeek:
> The shuffle modes description is difficult to read at first. Each element type and matrix dimension uses its own shuffle mode.

Yeah, it is a strange set of the ISA. At the moment I am orientating myself by looking at the image descriptions since from the perspective of `G_SHUFFLEVECTOR` I don't really care about what the original input was. I'll double check if I got that right, but being a bit more stringent on the size requirements seems prudent to me.

llvm/test/CodeGen/AIE/aie2/GlobalISel/prelegalizercombiner-shufflevector.mir
ValentijnvdBeek: I look wrong on the table, it is supposed to be 8x8 -> 8x8. But that is a mistake in the merge name. The content itself seems fine, at least according to the associated image.

llvm/test/CodeGen/AIE/aie2/GlobalISel/prelegalizercombiner-shufflevector.mir
ValentijnvdBeek: I think this is me just forgetting about this not being LLVM IR. I'll clasp the types, that should help.

llvm/test/CodeGen/AIE/aie2/GlobalISel/prelegalizercombiner-shufflevector.mir
ValentijnvdBeek: Yes, it does but vshuffle takes 1024 bytes of input. It ignores the higher bits of the input. 35 operates on a 8x8 8-bit element matrix, 64x8 vector, which is 512-bits. In the first case, those two are split into two 4x8 8-bit vectors and the second case is the "common" case where we just ignore the higher order bits.

What you propose would be a 4x8 8-bit match which is a 32x8 vector or 256-bit.

llvm/test/CodeGen/AIE/aie2/GlobalISel/prelegalizercombiner-shufflevector.mir
```diff
@@ -300,173 +300,6 @@ body:             |
...
-legalized:       false
-body:             |
-  bb.1.entry:
-    liveins: $wl2, $wl4
-    ; CHECK-LABEL: name: concat_vector_reverse_32_512_random
-    ; CHECK: liveins: $wl2, $wl4
-    ; CHECK-NEXT: {{  $}}
-    ; CHECK-NEXT: [[COPY:%[0-9]+]]:_(<8 x s32>) = COPY $wl2
-    ; CHECK-NEXT: [[COPY1:%[0-9]+]]:_(<8 x s32>) = COPY $wl4
-    ; CHECK-NEXT: [[CONCAT_VECTORS:%[0-9]+]]:_(<16 x s32>) = G_CONCAT_VECTORS [[COPY1]](<8 x s32>), [[COPY]](<8 x s32>)
```
ValentijnvdBeek: These got moved somehow, I will fix it when i get to it

PR #77, opened 2024-06-17
[AIE2] Support for 64-bit G_BUILD_VECTOR and G_UNMERGE_VALUES

ValentijnvdBeek:
Implements support for 64-bit vector creation and unmerging for AIE2 by extending the code for the 32-bit case. This will create mildly funky GMIR code since the rest of the legalizer splits 64-bit down into 32-bit numbers, but this all gets untangled later in the pipeline into nice(r) IR.

### DIFF
--- a/llvm/lib/Target/AIE/AIELegalizerInfo.cpp
+++ b/llvm/lib/Target/AIE/AIELegalizerInfo.cpp
@@ -89,7 +89,7 @@ static LegalityPredicate isValidVectorAIE2(const unsigned TypeIdx) {
   return [=](const LegalityQuery &Query) {
     const LLT DstTy = Query.Types[TypeIdx];
     const unsigned DstSize = DstTy.getSizeInBits();
-    return DstTy.isVector() && (DstSize == 32 || DstSize > 64);
+    return DstTy.isVector() && (DstSize >= 32);
   };
 }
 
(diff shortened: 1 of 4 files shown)


llvm/lib/Target/AIE/AIELegalizerInfo.cpp
```diff
@@ -89,7 +89,7 @@ static LegalityPredicate isValidVectorAIE2(const unsigned TypeIdx) {
   return [=](const LegalityQuery &Query) {
     const LLT DstTy = Query.Types[TypeIdx];
     const unsigned DstSize = DstTy.getSizeInBits();
-    return DstTy.isVector() && (DstSize == 32 || DstSize > 64);
+    return DstTy.isVector() && (DstSize == 32 || DstSize >= 64);
```
konstantinschwarz: `DstSize >= 32`?

llvm/lib/Target/AIE/AIELegalizerInfo.cpp
```diff
@@ -551,44 +551,46 @@ bool AIELegalizerInfo::legalizeCustom(LegalizerHelper &Helper,
   llvm_unreachable("Un-expected custom legalization");
 }
 
-bool AIELegalizerInfo::pack32BitVector(LegalizerHelper &Helper,
-                                       MachineInstr &MI,
-                                       Register SourceReg) const {
+bool AIELegalizerInfo::packBitVector(LegalizerHelper &Helper, MachineInstr &MI,
+                                     Register SourceReg) const {
```
konstantinschwarz: Nit: `SourceReg` is an artifact of having a single function for packing/unpacking. Since we are anyway passing `MI` and access `DstReg` in the function, I'd suggest to remove this parameter to make the code easier to understand. 
For `packBitVector`, we have `SourceReg` == `DstReg` anyway

ValentijnvdBeek: Yeah, that makes sense. I changed it, thanks for the review.

PR #78, opened 2024-06-17
[AIE2] Add ISel rules for 128-bit 2D/3D stores

gbossu:
We did not have ISel rules for 128-bit stores with 2D or 3D post-increments.

### DIFF
--- a/llvm/lib/Target/AIE/AIE2InstructionSelector.cpp
+++ b/llvm/lib/Target/AIE/AIE2InstructionSelector.cpp
@@ -3473,6 +3473,14 @@ LoadStoreOpcodes AIE2InstructionSelector::getLoadStoreOpcode(
                 /*OffsetOpcode=*/AIE2::VST_dmw_sts_w_ag_idx_imm};
       llvm_unreachable("Vector type not in AccRegBank nor VRegBank");
     }
+    if (getLoadStoreSize(I) == 128) {
+      unsigned RBID = deriveRegBankID(I.getOperand(2).getReg(), MRI, RBI);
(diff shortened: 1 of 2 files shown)


andcarminati (approved):
LGTM.

llvm/lib/Target/AIE/AIE2InstructionSelector.cpp
```diff
@@ -3473,6 +3473,14 @@ LoadStoreOpcodes AIE2InstructionSelector::getLoadStoreOpcode(
                 /*OffsetOpcode=*/AIE2::VST_dmw_sts_w_ag_idx_imm};
       llvm_unreachable("Vector type not in AccRegBank nor VRegBank");
     }
+    if (getLoadStoreSize(I) == 128) {
+      unsigned RBID = deriveRegBankID(I.getOperand(2).getReg(), MRI, RBI);
+      if (RBID == AIE2::VRegBankID) {
```
martien-de-jong, line 3478: nit: below for scalars, we don't check the register bank for scalars. Above we need it to select the right opcode. why here?

gbossu: For scalars, our load/store instructions can pretty much handle any bank. For 128-bit, we actually do have two valid instructions: One for mask registers, one for vector registers. However, at this point the mask registers unfortunately don't have their own bank yet. (They should, but instead they are considered part of VRegBank). So this means the only allowed bank is VRegBank. Once we have the correct banks and assignments, we can pick between the two opcodes.

SagarMaheshwari99 (approved):

PR #82, opened 2024-06-19
Improve support for our i20 pointers

gbossu:
Our target introduced "support" for i20 pointer types, but really this is just defining a new types, and optimizations like SROA barely know how to deal with "non-multiple-of-8" types.

This PR brings two changes to InstCombine and SROA so that `struct { ptr }` allocas can be simplified into simple `ptr` regs.

Essentially, those changes are:
 - Consider `ptr` always copy-able through load/stores in SROA, instead of requiring a memcpy because the storage size of `ptr` is different from the type size.
 - Simplify `trunc i20 (load ...)` into `load i20` because otherwise SROA cannot easily rewrite `alloca {ptr}` into `alloca ptr`


andcarminati:
I like this implementation, it is very clean! Is it possible to include some benchmark numbers? it can be a small selection of benchmarks (only if you have them available).

andcarminati (approved):

gbossu:
> I like this implementation, it is very clean! Is it possible to include some benchmark numbers? it can be a small selection of benchmarks (only if you have them available).

I don't have numbers specifically for these changes anymore, but I remember that the `BatchNorm` benchmarks really benefited from these changes, cutting the number of cycles by like 2x. The reason is that they use `vector_iterator` from AIE-API, which is just a `{ {ptr} }` in IR. That type was copied around with memcpy (due to copy constructors), and we would never simplify the allocas of such types or memcpy instructions due to the problems mentioned in the intro. Now, `vector_iterator` is pretty much just a `ptr` register, and increments to it are simple `geps`.

llvm/lib/Transforms/Scalar/SROA.cpp
```diff
@@ -3280,8 +3283,10 @@ class AllocaSliceRewriter : public InstVisitor<AllocaSliceRewriter, bool> {
 
     // If this doesn't map cleanly onto the alloca type, and that type isn't
     // a single value type, just emit a memcpy.
+    // We assume that ptr types are always "nice" types, which can use normal
+    // load/store pairs instead of memcpy.
     bool EmitMemCpy =
-        !VecTy && !IntTy &&
+        !VecTy && !IntTy && !NewAllocaTy->isPointerTy() &&
```
martien-de-jong, line 3289: nit: is it worth it to have an analogous `!PtrTy`?

llvm/test/Transforms/SROA/pointer-struct-20bit.ll
```diff
@@ -0,0 +1,52 @@
...
+; CHECK-LABEL: @copy_it_as_ptr(
+; CHECK-NEXT:  entry:
+; CHECK-NEXT:    [[RETVAL_SROA_0_0_COPYLOAD:%.*]] = load ptr, ptr [[THIS:%.*]], align 4
+; CHECK-NEXT:    [[RETVAL_SROA_2_0_THIS_SROA_IDX:%.*]] = getelementptr inbounds i8, ptr [[THIS]], i20 3
+; CHECK-NEXT:    [[RETVAL_SROA_2_SROA_0_0_COPYLOAD:%.*]] = load i8, ptr [[RETVAL_SROA_2_0_THIS_SROA_IDX]], align 1
+; CHECK-NEXT:    ret ptr [[RETVAL_SROA_0_0_COPYLOAD]]
+;
+entry:
+  %retval = alloca %vector_iterator, align 4
+  call void @llvm.memcpy.p0.p0.i32(ptr align 4 %retval, ptr align 4 %this, i32 4, i1 false)
```
martien-de-jong, line 46: Could it also occur as a direct struct assignment?

llvm/lib/Transforms/InstCombine/InstCombineLoadStoreAlloca.cpp
```diff
@@ -684,6 +687,22 @@ static Instruction *combineLoadToOperationType(InstCombinerImpl &IC,
...
+
+    // Fold trunc intPtrType (load intType) to load intPtrType if the load is
+    // only used by a IntToPtr instruction.
+    if (auto *TruncUser = dyn_cast<TruncInst>(Load.user_back());
+        TruncUser && TruncUser->hasOneUser()) {
+      Type *TruncTy = TruncUser->getDestTy();
+      const DataLayout &DL = IC.getDataLayout();
+      if (auto *IntToPtrUser = dyn_cast<IntToPtrInst>(TruncUser->user_back());
+          IntToPtrUser &&
+          TruncTy == DL.getIntPtrType(IntToPtrUser->getType())) {
```
martien-de-jong, line 699: Do we routinely see IntToPtrInst? Or are they introduced by the 20->32 bit SROA trick?

gbossu: They can be produced by a chain of events when a `{ ptr }` struct is copied through a 4-byte memcpy, that gets simplified in load/stores of i32. And then to get a pointer out of the i32 alloca, one needs a trunc+inttoptr.

llvm/test/Transforms/SROA/pointer-struct-20bit.ll
gbossu: Do you mean through a store? It happens as well. The problem here is really that memcpy were not correctly simplified.

martien-de-jong: Ok.

PR #91, opened 2024-06-24
utils/update_xxxx_test_checks.py: Support lit substitutions

gbossu:
This updates the following tools to support basic substitutions in RUN lines by reading lit.cfg files:
 - llvm/utils/update_cc_test_checks.py
 - llvm/utils/update_mir_test_checks.py
 - llvm/utils/update_llc_test_checks.py

This uses lit.discovery.getTests() to find the parent test suite for each test to update. Note that because the update scripts aren't configured at build time like llvm-lit, it is not always possible to read the lit.cfg files completely. In particular, lit.site.cfg files are missing. However, even a partial understanding of the config is enough to collect most user substitutions and apply them in RUN lines.

A lot of our tests use substitutions like `%topdown-multi`. This teaches the update scripts how to handle them.


llvm/utils/UpdateTestChecks/common.py
```diff
@@ -340,9 +349,57 @@ def get_checks_for_unused_prefixes(
...
+    lit_config.targets = []
+
+    tests = lit.discovery.getTests(
+        test_path, lit_config, test_suite_cache, local_config_cache
+    )[1]
+    try:
+        return next(tests).config.substitutions
+    except StopIteration:
+        return []
+
```
ValentijnvdBeek: This would be more readable as a `for`-loop. One of the weird quirks of Python is that for loops are just try-catch wrappers that look for the `StopIteration` exception.

llvm/utils/UpdateTestChecks/common.py
```diff
@@ -437,10 +500,22 @@ def applySubstitutions(s, substitutions):
...
 
 
+def applyArgArraySubstitutions(args: list, substitutions):
+    new_args = []
+
+    # After expanding substitutions, args might be turned into multiple ones.
+    # Make sure to re-split.
+    for arg in args:
+        new_args += shlex.split(applySubstitutions(arg, substitutions))
+    return new_args
```
ValentijnvdBeek: Nit (?): Alternatively, this could also be a map or list comprehension instead. Something on the lines of `[shlex.split(applySubstitions(args, substitution)) for arg in `args]`` or `args.map(lambda arg: ...)`

llvm/utils/update_cc_test_checks.py
```diff
@@ -309,11 +309,11 @@ def main():
...
-        subs = {
-            "%s": ti.path,
-            "%t": tempfile.NamedTemporaryFile().name,
-            "%S": os.path.dirname(ti.path),
-        }
+        subs = [
+            ("%s", ti.path),
+            ("%t", tempfile.NamedTemporaryFile().name),
+            ("%S", os.path.dirname(ti.path)),
+        ] + ti.lit_substitutions
```
ValentijnvdBeek: Nit: this could also be written as `[..., *ti.lit_substitutions]`.

llvm/utils/UpdateTestChecks/common.py
gbossu: I have changed that bit in follow-up fixups to support Windows as well, maybe it's a bit more readable.

llvm/utils/UpdateTestChecks/common.py
gbossu: The thing is, I'm only interested in one of the tests, because I want to access the configuration for them, and this is shared across all. Ultimately, i'd like something like `tests[0]` or `tests.front()` but I don't think python provides this :/

llvm/utils/update_cc_test_checks.py
gbossu: Oh, TIL

llvm/utils/UpdateTestChecks/common.py
gbossu: I guess I could write the following:
```
    for test in tests:
        return test.config.substitutions
    return []
```

I'm undecided whether it's clearer. What do you think?

ValentijnvdBeek: I think that you can also write something line `list(tests)[0]` which exhausts the iterator and then gets the first element. 

I agree with that it is a difficult case and I am not sure if there is a strictly better one. The for loop feels nicer to me since it doesn't have the try catch mechanism, which takes more knowledge to understand (ie StopIteration is not a fault).

andcarminati (approved):
LGTM.

PR #92, opened 2024-06-24
[AIE] legalize saturating arith ops by lowering

martien-de-jong:
This came up in a supertest rune, where an USUBSAT was distilled from some mixed sign comparison context.


konstantinschwarz (approved):

martien-de-jong:
One less Supertest failure

PR #93, opened 2024-06-24
[AIE] Memory bank modelling in the frontend/backend

krishnamtibrewala:
Provide a mechanism for adding bank annotations


clang/lib/Headers/aiev2_core.h
```diff
@@ -1566,4 +1566,300 @@ inline v16accfloat addmsc_4x8_8x4( v32float v1, v32float v2, v16accfloat acc1, v
 
 #endif
 
+enum class aie_dm_resource {
+  none,
+  PM,
+  DM,
+  DM_test,
+  stack,
```
martien-de-jong, line 1574: Do we have an overlap map somewhere?

clang/lib/Headers/aiev2_core.h
```diff
@@ -1566,4 +1566,300 @@ inline v16accfloat addmsc_4x8_8x4( v32float v1, v32float v2, v16accfloat acc1, v
...
+  a,
+  b,
+  c,
+  d,
+  ab,
```
martien-de-jong, line 1579: These are "a or b" etc?

clang/lib/Headers/aiev2_core.h
```diff
@@ -1566,4 +1566,300 @@ inline v16accfloat addmsc_4x8_8x4( v32float v1, v32float v2, v16accfloat acc1, v
...
+#define __aie_dm_resource_cd                                                   \
+  __attribute__((address_space(static_cast<unsigned>(aie_dm_resource::cd))))
+#define __aie_dm_resource_TM                                                   \
+  __attribute__((address_space(static_cast<unsigned>(aie_dm_resource::TM))))
+
+template <typename T> struct aie_dm_resource_remove {
+  using type = T;
+};
+
+template <typename T, int N>
```
martien-de-jong, line 1607: Above we cast to unsigned, here we use int as address space identifier. Why the difference? (I guess the enum constants would evaluate to int, which then would not need a cast?)

clang/include/clang/AST/Type.h
```diff
@@ -520,7 +520,9 @@ class Qualifiers {
            // to implicitly cast into the default address space.
            (A == LangAS::Default &&
             (B == LangAS::cuda_constant || B == LangAS::cuda_device ||
-             B == LangAS::cuda_shared));
+             B == LangAS::cuda_shared)) ||
+           // TODO make following check specific to AIE target only
```
martien-de-jong, line 524: It seems to say: any non-default address space is a subspace of the default address space. I like that, and I think it would make sense to get that option into a target plugin.

clang/test/CodeGen/aie/aie2/aie2-bank-annotation.cpp
```diff
@@ -0,0 +1,57 @@
+// NOTE: Assertions have been autogenerated by utils/update_cc_test_checks.py UTC_ARGS: --version 4
+//===- aie2-bank-annotation.cpp----------------------------------*- C++ -*-===//
+//
+// This file is licensed under the Apache License v2.0 with LLVM Exceptions.
+// See https://llvm.org/LICENSE.txt for license information.
+// SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+//
+// (c) Copyright 2023-2024 Advanced Micro Devices, Inc. or its affiliates
```
martien-de-jong: nit: only 2024?

clang/test/CodeGen/aie/aie2/aie2-bank-annotation.cpp
```diff
@@ -0,0 +1,57 @@
...
+
+// CHECK-LABEL: define dso_local noundef i32 @_Z6squarePU3AS5iS0_(
+// CHECK-SAME: ptr addrspace(5) nocapture readonly [[NUM:%.*]], ptr addrspace(5) nocapture writeonly [[MEM:%.*]]) local_unnamed_addr #[[ATTR0:[0-9]+]] {
+// CHECK-NEXT:  entry:
+// CHECK-NEXT:    [[TMP0:%.*]] = load i32, ptr addrspace(5) [[NUM]], align 4, !tbaa [[TBAA2:![0-9]+]]
+// CHECK-NEXT:    [[MUL:%.*]] = mul nsw i32 [[TMP0]], [[TMP0]]
+// CHECK-NEXT:    store i32 [[MUL]], ptr addrspace(5) [[MEM]], align 4, !tbaa [[TBAA2]]
+// CHECK-NEXT:    ret i32 [[MUL]]
+//
+int square(int __aie_dm_resource_a *num, int __aie_dm_resource_a* mem) {
```
martien-de-jong: nit: spacing around '*' different for num and mem

llvm/lib/Target/AIE/AIEAddressSpaceLowering.cpp
```diff
@@ -0,0 +1,113 @@
+//===--------- AIEAddressSpaceLowering.cpp - Retype pointer to P0  --------===//
+//
+// Part of the LLVM Project, under the Apache License v2.0 with LLVM Exceptions.
+// See https://llvm.org/LICENSE.txt for license information.
+// SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+//
```
martien-de-jong: AMD clause?

llvm/lib/Target/AIE/AIE2TargetMachine.cpp
```diff
@@ -143,6 +143,7 @@ void AIE2PassConfig::addIRPasses() {
 }
 
 void AIE2PassConfig::addPreLegalizeMachineIR() {
+  addPass(createAIEAddressSpaceLowering());
```
martien-de-jong: CHECK: we remove the extra information from the LLTs before doing anything else, which means that all our optimizations are based on the high level types as supplied in MMOs. As such this pass is 'repairing' the irtranslator to translate the annotated pointers to default LLT ptr types. We are pretty confident that we don't need more elaborate LLT types.
I can imagine we could keep distinct LLTs for PM, DM and TM, but that is not strictly necessary.

llvm/lib/Target/AIE/AIEAddressSpaceLowering.cpp
```diff
@@ -0,0 +1,113 @@
...
+
+class AIEAddressSpaceLowering : public MachineFunctionPass {
+public:
+  static char ID;
+  AIEAddressSpaceLowering() : MachineFunctionPass(ID) {}
+
+  bool runOnMachineFunction(MachineFunction &MF) override {
+    MachineRegisterInfo &MRI = MF.getRegInfo();
+    TargetPassConfig &TPC = getAnalysis<TargetPassConfig>();
+    // Enable CSE.
```
martien-de-jong: I could do with some more comment. What is the function of CSE here? I can imagine cse-ing default address space expressions with specific ones?

llvm/lib/Target/AIE/AIEAddressSpaceLowering.cpp
```diff
@@ -0,0 +1,113 @@
...
+static const char AIE_POINTER_RETYPE[] = "AIE Address Space Lowering Pass";
+
+namespace {
+
+class AIEAddressSpaceLowering : public MachineFunctionPass {
+public:
+  static char ID;
+  AIEAddressSpaceLowering() : MachineFunctionPass(ID) {}
+
+  bool runOnMachineFunction(MachineFunction &MF) override {
```
martien-de-jong: I wouldn't mind defining the bodies of these main functions outline of the class.

llvm/lib/Target/AIE/AIEAddressSpaceLowering.cpp
```diff
@@ -0,0 +1,113 @@
...
+                 MachineIRBuilder &MIB, GISelObserverWrapper &Observer) {
+
+    // Lower all pointers to pointers with default address space
+    for (llvm::MachineOperand &MO : MI.operands()) {
+      if (!MO.isReg())
+        continue;
+      Register Reg = MO.getReg();
+      if (MRI.getType(Reg).isPointer() && Reg.isVirtual()) {
+        Observer.changingInstr(MI);
+        MRI.setType(Reg, LLT::pointer(0, 20));
```
martien-de-jong: Perhaps get the 20 from some architecture-dependent api?

llvm/lib/Target/AIE/AIEAddressSpaceLowering.cpp
```diff
@@ -0,0 +1,113 @@
...
+        Observer.changedInstr(MI);
+      }
+    }
+
+    // Since we retypr the pointer to P0, G_ADDRSPACE_CAST with same i/p & o/p
```
martien-de-jong: nit: *retype*

clang/lib/Headers/aiev2_core.h
krishnamtibrewala: Not as part of this PR, those will come as part of follow up PR,

clang/lib/Headers/aiev2_core.h
krishnamtibrewala: Yes "ab" are compound bank annotation, which represent the pointer could access memory in Bank A or B, etc.

clang/include/clang/AST/Type.h
krishnamtibrewala: Yes the plan is have a target plugin, we can discuss on IM more. 
But thats a big change in clang frontend and based on discussion with Gaetan & Konstantin, on side I work on that to get upstream to LLVM

llvm/lib/Target/AIE/AIE2TargetMachine.cpp
krishnamtibrewala: PM, DM & TM has a distinct address space number, I would prefer keeping only P0 with different address space.

clang/lib/Headers/aiev2_core.h
stephenneuendorffer: If the point is to describe memory banks, then shouldn't we be focusing only on the memory bank aspect?  I assume you're modeling it this way to get compatibility with existing code?

clang/lib/Headers/aiev2_core.h
```diff
@@ -1566,4 +1566,300 @@ inline v16accfloat addmsc_4x8_8x4( v32float v1, v32float v2, v16accfloat acc1, v
...
+};
+
+template <typename T>
+using aie_dm_resource_remove_t = typename aie_dm_resource_remove<T>::type;
+
+template <typename T, aie_dm_resource Resource> struct aie_dm_resource_set {
+  using type = T;
+};
+
+template <typename T> struct aie_dm_resource_set<T, aie_dm_resource::a> {
```
stephenneuendorffer, line 1632: It seems like these templates could be written much more compactly with some preprocessing magic.   I assume the problem here is that you can't put a template reference in an __attribute declaration?

clang/lib/Headers/aiev2_core.h
```diff
@@ -1566,4 +1566,300 @@ inline v16accfloat addmsc_4x8_8x4( v32float v1, v32float v2, v16accfloat acc1, v
...
+  ab,
+  ac,
+  ad,
+  bc,
+  bd,
+  cd,
+  TM
+};
+
+#define __aie_dm_resource_PM                                                   \
```
stephenneuendorffer: Should these #defines remain defined?

clang/lib/Headers/aiev2_core.h
krishnamtibrewala: have updated it to signed, we would still need a cast for the define to be used in the AIE API or user code.

clang/lib/Headers/aiev2_core.h
stephenneuendorffer: How about:
```
#define FOO(x) __attribute__((address_space(static_cast<signed>(aie_dm_resource::x))))
#define __aie_dm_resource_DM FOO(DM)
#undef FOO
```
I find this much easier to see what the regular pattern is, especially given the awkward line breaks.

llvm/lib/Target/AIE/AIEAddressSpaceLowering.cpp
gbossu: Can't we just get the size from the type we are lowering? If we run this pass, it means we know our pointers all point to aliasing/equivalent spaces. I guess we can just convert `LLT::pointer(x, y)` to `LLT::pointer(0, y)`?

llvm/lib/Target/AIE/AIEAddressSpaceLowering.cpp
```diff
@@ -0,0 +1,115 @@
...
+  }
+
+  StringRef getPassName() const override { return AIE_POINTER_RETYPE; }
+
+private:
+  void processMI(MachineInstr &MI, MachineRegisterInfo &MRI,
+                 MachineIRBuilder &MIB, GISelObserverWrapper &Observer) {
+
+    // Lower all pointers to pointers with default address space
+    for (llvm::MachineOperand &MO : MI.operands()) {
```
gbossu: The code is a lot more compact now, great :)

llvm/lib/Target/AIE/AIEAddressSpaceLowering.cpp
```diff
@@ -0,0 +1,115 @@
...
+      Observer.addObserver(CSEInfo);
+      MIB.setChangeObserver(Observer);
+    }
+
+    for (MachineBasicBlock &MBB : MF) {
+      for (MachineInstr &MI : MBB) {
+        processMI(MI, MRI, MIB, Observer);
+      }
+    }
+    return true;
```
andcarminati: Here it is better to return true only if you change some instruction. In this way, it is better to have a bool return type for `processMI`.

llvm/lib/Target/AIE/AIEAddressSpaceLowering.cpp
```diff
@@ -0,0 +1,115 @@
+//===--------- AIEAddressSpaceLowering.cpp - Retype pointer to P0  --------===//
+//
+// Part of the LLVM Project, under the Apache License v2.0 with LLVM Exceptions.
+// See https://llvm.org/LICENSE.txt for license information.
+// SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+//
+// (c) Copyright 2024 Advanced Micro Devices, Inc. or its affiliates
+//
+//===----------------------------------------------------------------------===//
+//
```
andcarminati: I think we can have a summary for this pass here.

llvm/lib/Target/AIE/AIEAddressSpaceLowering.cpp
```diff
@@ -0,0 +1,115 @@
...
+  }
+
+  void getAnalysisUsage(AnalysisUsage &AU) const override {
+    AU.addRequired<MachineModuleInfoWrapperPass>();
+    AU.addRequired<GISelCSEAnalysisWrapperPass>();
+    AU.addRequired<TargetPassConfig>();
+    AU.setPreservesAll();
+  }
+
+  StringRef getPassName() const override { return AIE_POINTER_RETYPE; }
```
andcarminati: I think this AIE_POINTER_RETYPE escapes from the convention, from here it looks like a macro or enum item. Maybe it is better to just return the string directly.

clang/test/Sema/address_spaces.c
```diff
@@ -1,3 +1,4 @@
+// XFAIL: *
```
andcarminati, line 1: Do you think we can handle these tests without XFAIL? I mean, they support the common behavior of `address_space`, that is required for mainstream compatibility.

llvm/lib/Target/AIE/AIEAddressSpaceLowering.cpp
krishnamtibrewala: I am following how other passes that were introduced for AIE target, handles `getPassName()` .
Kindly confirm if we want to change the way.

clang/test/Sema/address_spaces.c
krishnamtibrewala: Currently we have created a copy of these testes under AIE target (`clang/test/CodeGen/aie/address_space/`) and supported most of them. 
Based on discussion on LLVM forum & Internally (Documented internally on JIRA). Plan is to extend the support of `isAddressSpaceSupersetOf ` clang address_space to support requirement of AIE in parallel.

llvm/lib/Target/AIE/AIEAddressSpaceLowering.cpp
andcarminati: In this way I think you can keep as is.

clang/test/Sema/address_spaces.c
andcarminati: Hi @krishnamtibrewala, understood, thank you for the explanation.

llvm/lib/Target/AIE/AIEAddressSpaceLowering.cpp
```diff
@@ -0,0 +1,122 @@
...
+
+private:
+  bool processMI(MachineInstr &MI, MachineRegisterInfo &MRI,
+                 MachineIRBuilder &MIB, GISelObserverWrapper &Observer) {
+    bool Changed = false;
+    // Lower all pointers to pointers with default address space
+    for (llvm::MachineOperand &MO : MI.operands()) {
+      if (!MO.isReg())
+        continue;
+      Register Reg = MO.getReg();
```
andcarminati: nit: `const Register Reg = MO.getReg();`

llvm/lib/Target/AIE/AIEAddressSpaceLowering.cpp
```diff
@@ -0,0 +1,122 @@
+//===--------- AIEAddressSpaceLowering.cpp - Retype pointer to P0  --------===//
```
andcarminati: Not really important, but maybe "Lowering" is a too heavy term that give us a impression that operations will be broken into smaller ones. What happens here is retype or "address space flattening".

gbossu: I proposed this new name back then, but I agree something like `AddressSpaceFlattening` or `AddressSpaceRetyping` sounds better. I would be fine to keep the name like this though, as there is a comment at the top of the file that briefly explains what the pass does.

clang/lib/Headers/aiev2_core.h
gbossu: I hate macros, but I have to agree here. @krishnamtibrewala do you think there's something we could do to improve the readability?

clang/test/CodeGen/aie/aie2/aie2-bank-annotation.cpp
```diff
@@ -0,0 +1,57 @@
...
+//===- aie2-bank-annotation.cpp----------------------------------*- C++ -*-===//
+//
+// This file is licensed under the Apache License v2.0 with LLVM Exceptions.
+// See https://llvm.org/LICENSE.txt for license information.
+// SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+//
+// (c) Copyright 2024 Advanced Micro Devices, Inc. or its affiliates
+//
+//===----------------------------------------------------------------------===//
+// RUN: %clang -O1 --target=aie2 -S -emit-llvm %s -o - | FileCheck %s
```
gbossu, line 13: I would be happy with a top-level comment to explain what we are testing.

clang/lib/Headers/aiev2_core.h
```diff
@@ -1566,4 +1566,283 @@ inline v16accfloat addmsc_4x8_8x4( v32float v1, v32float v2, v16accfloat acc1, v
...
+#define __aie_dm_resource_b AIE_DM_RESOURCE(b)
+#define __aie_dm_resource_c AIE_DM_RESOURCE(c)
+#define __aie_dm_resource_d AIE_DM_RESOURCE(d)
+#define __aie_dm_resource_ab AIE_DM_RESOURCE(ab)
+#define __aie_dm_resource_ac AIE_DM_RESOURCE(ac)
+#define __aie_dm_resource_ad AIE_DM_RESOURCE(ad)
+#define __aie_dm_resource_bc AIE_DM_RESOURCE(bc)
+#define __aie_dm_resource_bd AIE_DM_RESOURCE(bd)
+#define __aie_dm_resource_cd AIE_DM_RESOURCE(cd)
+#define __aie_dm_resource_TM AIE_DM_RESOURCE(TM)
```
gbossu, line 1601: Could you add test for these to see how they get translated into IR? Not sure we need everything, but at least a. b, c and d would be nice.

llvm/test/CodeGen/AIE/GlobalISel/addrspace-translate.ll
```diff
@@ -0,0 +1,67 @@
+; NOTE: Assertions have been autogenerated by utils/update_mir_test_checks.py UTC_ARGS: --version 4
+;
+; This file is licensed under the Apache License v2.0 with LLVM Exceptions.
+; See https://llvm.org/LICENSE.txt for license information.
+; SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+;
+; (c) Copyright 2023-2024 Advanced Micro Devices, Inc. or its affiliates
+; RUN: llc -mtriple=aie2 -stop-after=aie-addrspace-lowering %s -o - 2>&1 | FileCheck %s
```
gbossu: Thanks, I think it's great to have an "integration" test like this as well :)

clang/test/Sema/address_spaces.c
gbossu: I believe that commit should be squashed with the one where we actually make the change to allow implicit address space casts?

krishnamtibrewala: have merged it as suggested above.

llvm/lib/Target/AIE/AIEAddressSpaceLowering.cpp
krishnamtibrewala: have renamed to `AddressSpaceFlattening`

clang/lib/Headers/aiev2_core.h
krishnamtibrewala: Sure, have added to IR translation test for all the dm_resources.

clang/lib/Headers/aiev2_core.h
krishnamtibrewala: I tried ways, but none of the ways worked well for AIE API.

clang/test/CodeGen/aie/address_space/address-space-conversion.cpp
```diff
@@ -0,0 +1,207 @@
...
+  (void)static_cast<A_ptr>(bp);
+  (void)static_cast<A_ptr_1>(bp1);
+  (void)static_cast<A_ptr_2>(bp2);
+
+  // Well-formed downcast
+  (void)static_cast<B_ptr>(ap);
+  (void)static_cast<B_ptr_1>(ap1);
+  (void)static_cast<B_ptr_2>(ap2);
+
+  // Well-formed cast to/from void
```
martien-de-jong, line 73: nit: *void pointer*

clang/test/CodeGen/aie/address_space/address_space-dependent.cpp
```diff
@@ -0,0 +1,129 @@
...
+template <int I>
+struct HasASTemplateFields {
+  __attribute__((address_space(I))) int as_field; // expected-error {{field may not be qualified with an address space}}
+};
+
+template <int I, int J>
+void foo(__attribute__((address_space(I))) float *a, // expected-note {{candidate template ignored: substitution failure [with I = 1, J = 2]: parameter may not be qualified with an address space}}
+         __attribute__((address_space(J))) float b) {
+  *a = 5.0f + b;
+}
```
martien-de-jong, line 45: I guess that a global declaration `mytype AS1 * AS2 Ptr;`, which is a Ptr located in address space AS2 pointing to AS1.

llvm/lib/Target/AIE/AIEAddressSpaceFlattening.cpp
```diff
@@ -0,0 +1,124 @@
...
+                 MachineIRBuilder &MIB, GISelObserverWrapper &Observer) {
+    bool Changed = false;
+    // Lower all pointers to pointers with default address space
+    for (llvm::MachineOperand &MO : MI.operands()) {
+      if (!MO.isReg())
+        continue;
+      const Register Reg = MO.getReg();
+      if (MRI.getType(Reg).isPointer() && Reg.isVirtual()) {
+        Observer.changingInstr(MI);
+        MRI.setType(Reg, LLT::pointer(0, MRI.getType(Reg).getSizeInBits()));
```
martien-de-jong, line 91: Use something like `const int DefaultAddrSpace = 0;` here?

clang/test/CodeGen/aie/address_space/address-space-conversion.cpp
```diff
@@ -0,0 +1,207 @@
+//===- address-space-conversion.cpp ------------------------------*- C++ -*-===//
```
gbossu, line 1: Super-nit: Maybe have a `README.md` under `clang/test/CodeGen/aie/address_space/` that actually says those tests come from `CodeGen/Sema` but that been updated with the new rules introduced for AIE?

llvm/test/CodeGen/AIE/GlobalISel/addrspace-lowering.mir
```diff
@@ -0,0 +1,161 @@
+# NOTE: Assertions have been autogenerated by utils/update_mir_test_checks.py UTC_ARGS: --version 4
+#
+# This file is licensed under the Apache License v2.0 with LLVM Exceptions.
+# See https://llvm.org/LICENSE.txt for license information.
+# SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+#
+# (c) Copyright 2023-2024 Advanced Micro Devices, Inc. or its affiliates
```
martien-de-jong, line 7: 2024 only?

llvm/lib/Target/AIE/AIEAddressSpaceFlattening.cpp
```diff
@@ -0,0 +1,124 @@
+//===--------- AIEAddressSpaceFlattening.cpp - Retype pointer to P0
+//--------===//
```
gbossu, line 2: Nit: Weird formatting

llvm/test/CodeGen/AIE/GlobalISel/addrspace-lowering.mir
```diff
@@ -0,0 +1,161 @@
...
+  bb.1:
+    ; CHECK-LABEL: name: chain_G_ADDRSPACE_CAST
+    ; CHECK: [[C:%[0-9]+]]:_(p0) = G_CONSTANT i20 0
+    ; CHECK-NEXT: [[COPY:%[0-9]+]]:_(p0) = COPY [[C]](p0)
+    ; CHECK-NEXT: [[COPY1:%[0-9]+]]:_(p0) = COPY [[COPY]](p0)
+    ; CHECK-NEXT: [[LOAD:%[0-9]+]]:_(<32 x s8>) = G_LOAD [[COPY1]](p0) :: (load (<32 x s8>) from `ptr addrspace(5) null`, addrspace 5)
+    %10:_(p0) = G_CONSTANT i20 0
+    %1:_(p5) = G_ADDRSPACE_CAST %10:_(p0)
+    %2:_(p6) = G_ADDRSPACE_CAST %1:_(p5)
+    %0:_(<32 x s8>) = G_LOAD %2:_(p6) :: (load (<32 x s8>) from `ptr addrspace(5) null`, addrspace 5)
```
gbossu, line 71: Nit: I think `ptr addrspace(6) null, addrspace 6` would be better in the MMO, as we are loading using `p6`

clang/test/CodeGen/aie/aie2/aie2-bank-annotation.cpp
```diff
@@ -0,0 +1,161 @@
...
+// CHECK-NEXT:    store i32 [[MUL]], ptr [[MEM]], align 4, !tbaa [[TBAA2]]
+// CHECK-NEXT:    ret i32 [[MUL]]
+//
+int square(int __aie_dm_resource_b *num, int *mem) {
+    *mem = (*num) * (*num);
+    return *mem;
+}
+
+
+// CHECK-LABEL: define dso_local noundef i32 @_Z6squarePU3AS7i(
```
gbossu, line 56: I think those "simple" tests below should be in a separate file where we really just test the translation of memory banks into address spaces. That file can then be added as a test in the first commit.

The tests above should remain here in the second commit, as they test that implicit address space casts are allowed.

gbossu (approved):
LGTM! I think the PR is in great shape now :) Please look at the remaining comments, but most of them are nits. I think the only thing I'd really like addressed is https://github.com/Xilinx/llvm-aie/pull/93#discussion_r1663707709

clang/test/CodeGen/aie/address_space/address-space-ctor.cpp
```diff
@@ -0,0 +1,30 @@
...
+//
+//  Part of the LLVM Project, under the Apache License v2.0 with LLVM Exceptions.
+//  See https://llvm.org/LICENSE.txt for license information.
+//  SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+//
+// Modifications (c) Copyright 2024 Advanced Micro Devices, Inc. or its affiliates
+//
+//===----------------------------------------------------------------------===//
+
+// RUN: %clang_cc1 -triple aie %s -std=c++14 -triple=spir -verify -fsyntax-only
```
andcarminati, line 11: Triple is twice here (spir).

llvm/lib/Target/AIE/AIEAddressSpaceFlattening.cpp
```diff
@@ -0,0 +1,124 @@
...
+private:
+  bool processMI(MachineInstr &MI, MachineRegisterInfo &MRI,
+                 MachineIRBuilder &MIB, GISelObserverWrapper &Observer) {
+    bool Changed = false;
+    // Lower all pointers to pointers with default address space
+    for (llvm::MachineOperand &MO : MI.operands()) {
+      if (!MO.isReg())
+        continue;
+      const Register Reg = MO.getReg();
+      if (MRI.getType(Reg).isPointer() && Reg.isVirtual()) {
```
andcarminati, line 89: I think you don't need  to check for `Reg.isVirtual()` because this is done in `MRI` anyway.

llvm/lib/Target/AIE/AIEAddressSpaceFlattening.cpp
andcarminati, line 91: Maybe something like:
```
    for (llvm::MachineOperand &MO : MI.operands()) {
      if (!MO.isReg())
        continue;
      const Register Reg = MO.getReg();
      const LLT RegType = MRI.getType(Reg);
      if (RegType.isPointer()) {
        Observer.changingInstr(MI);
        MRI.setType(Reg, LLT::pointer(0, RegType.getSizeInBits()));
        Observer.changedInstr(MI);
        Changed = true;
      }
    }
```

clang/test/CodeGen/aie/address_space/address-space-conversion.cpp
```diff
@@ -0,0 +1,207 @@
...
+
+  // Well-formed cast to/from void
+  (void)static_cast<void_ptr>(ap);
+  (void)static_cast<void_ptr_1>(ap1);
+  (void)static_cast<void_ptr_2>(ap2);
+  (void)static_cast<A_ptr>(vp);
+  (void)static_cast<A_ptr_1>(vp1);
+  (void)static_cast<A_ptr_2>(vp2);
+  
+  // Ill-formed upcasts
```
andcarminati, line 81: Maybe some reordering or some comment change will be nice here. When I see `Ill-formed upcasts` and the first sentence says `// Supported by AIE target`  I feel it suspicious &#128579;.

clang/test/CodeGen/aie/address_space/address-space-conversion.cpp
krishnamtibrewala: These are upstream tests and have been copied here to keep the expected working cases tracked here so once we make the change target specific we can re-enable the tests, I want to avoid changing any file in `clang/test/CodeGen/aie/address_space/`

krishnamtibrewala:
After moving files as suggested in the review, an incorrect git command led to closer of this PR and git is no longer allowing me to reopen the same PR so have opened a new PR #111 with required changes.

PR #95, opened 2024-06-25
[AIEX] Actively avoid bank conflicts in the scheduler

krishnamtibrewala:
Use bank annotation information in Hazard recognizer to avoid scheduling multiple memory operation to a bank in same cycle.

In the current approach
1. Two VLD will only be issued together if we have bank annotation for one or both.


llvm/lib/Target/AIE/AIE2Subtarget.cpp
```diff
@@ -76,3 +76,45 @@ const RegisterBankInfo *AIE2Subtarget::getRegBankInfo() const {
 InstructionSelector *AIE2Subtarget::getInstructionSelector() const {
   return InstSelector.get();
 }
+
+std::bitset<8>
+AIE2Subtarget::getMemoryBanksFromAddressSpace(unsigned AddrSpace) const {
+  std::bitset<8> MemoryBanks;
+
+  switch (AddrSpace) {
+  case AIE2AS::a:
```
stephenneuendorffer: What about unimplemented cases?  should they return something with all flags set?  or none?

llvm/lib/Target/AIE/AIEHazardRecognizer.cpp
```diff
@@ -62,6 +62,9 @@ void FuncUnitWrapper::dump() const {
   dbgs() << " Slots : ";
   for (int J = 9; J >= 0; J--)
     dbgs() << ((Slots & (1ULL << J)) ? Digits[J] : '-');
+  dbgs() << " Memorybanks : ";
+  for (int J = 9; J >= 0; J--)
+    dbgs() << ((MemoryBanks & (1ULL << J)) ? Digits[J] : '-');
```
stephenneuendorffer: good place for a lambda?

llvm/lib/Target/AIE/AIE2Subtarget.cpp
krishnamtibrewala: Have updated to set as none (assume no conflict)

llvm/lib/Target/AIE/AIEHazardRecognizer.cpp
krishnamtibrewala: Yes , have updated to use lambdas.

llvm/lib/Target/AIE/AIE2Subtarget.cpp
```diff
@@ -80,6 +80,7 @@ InstructionSelector *AIE2Subtarget::getInstructionSelector() const {
 std::bitset<8>
 AIE2Subtarget::getMemoryBanksFromAddressSpace(unsigned AddrSpace) const {
   std::bitset<8> MemoryBanks;
+  MemoryBanks.reset();
```
gbossu: Please move to the previous commit. But that means that by default a `std::bitset` is uninitialized?

llvm/lib/Target/AIE/AIEHazardRecognizer.h
```diff
@@ -128,9 +136,11 @@ class AIEHazardRecognizer : public ScheduleHazardRecognizer {
   ///        use from the pre-RA scheduler, where detailed resource modelling
   ///        doesn't pay off.
   void emitInScoreboard(ResourceScoreboard<FuncUnitWrapper> &Scoreboard,
-                        const MCInstrDesc &Desc, int DeltaCycles) const;
+                        const MCInstrDesc &Desc, int DeltaCycles,
```
gbossu: Nit: I'd prefer `DeltaCycles` to remain the last attribute, I believe `Desc` and `MemoryBanks` make sense to be kept next to each other.

llvm/lib/Target/AIE/AIEHazardRecognizer.cpp
```diff
@@ -392,28 +404,30 @@ auto toHazardType(bool Conflict) {
...
     const ResourceScoreboard<FuncUnitWrapper> &Scoreboard,
     const InstrItineraryData *ItinData, unsigned SchedClass, SlotBits SlotSet,
-    int DeltaCycles, std::optional<int> FUDepthLimit) {
+    MemoryBankBits MemoryBank, int DeltaCycles,
+    std::optional<int> FUDepthLimit) {
   assert(Scoreboard.isValidDelta(DeltaCycles));
 
   // Verify format hazards
-  FuncUnitWrapper EmissionCycle(/*Req=*/0, /*Res=*/0, SlotSet);
+  FuncUnitWrapper EmissionCycle(/*Req=*/0, /*Res=*/0, SlotSet, MemoryBank);
```
gbossu: I think this is a bit more complicated. Our memory instructions don't all happen in the same cycle relativ to the emission.  E.g. VST stores in E5, but VST.SRS stores in E7. I think you need to add the MemoryBank bits in `ScoreBoard[DeltaCycles + MemCycle]`. You can find the memory cycle using e.g. `AIEBaseInstrInfo::getFirstMemoryCycle(SchedClass)`

Note you also need to update `getHazardType` in a similar fashion.

krishnamtibrewala: Thank you for pointing out this, 
Assuming `DeltaCycles` represent E1, so I think I should be adding the MemoryBank bits in `ScoreBoard[DeltaCycles + MemCycle - 1]`

llvm/lib/Target/AIE/AIE2Subtarget.cpp
krishnamtibrewala: Have removed this , by default `bitset` is initialized to 0.

llvm/lib/Target/AIE/AIE2Subtarget.cpp
```diff
@@ -76,3 +76,47 @@ const RegisterBankInfo *AIE2Subtarget::getRegBankInfo() const {
...
   return InstSelector.get();
 }
+
+std::bitset<8>
+AIE2Subtarget::getMemoryBanksFromAddressSpace(unsigned AddrSpace) const {
+  std::bitset<8> MemoryBanks;
+
+  switch (AddrSpace) {
+  case AIE2AS::a:
+    MemoryBanks.set(0);
```
andcarminati: Hi, instead of constants, can we use another enum for clearness?

llvm/lib/Target/AIE/AIE2AddrSpace.h
```diff
@@ -0,0 +1,47 @@
...
+// This file declares the AIEngine V2 Address Space and DM banks
+//
+//===----------------------------------------------------------------------===//
+
+#ifndef LLVM_SUPPORT_AIE2ADDRSPACE_H
+#define LLVM_SUPPORT_AIE2ADDRSPACE_H
+
+namespace llvm {
+
+namespace AIE2AS {
```
gbossu: Nit: I'd prefer
```
namespace AIE2 {

enum class AddressSpaces { ... }

enum class DMBanks { ... }

}
```

This is more in line with what we are doing when writing AIE-version-specific code. Also note I'm using c++ enum classes, not C enums.

llvm/lib/Target/AIE/AIE2AddrSpace.h
```diff
@@ -0,0 +1,47 @@
...
+//
+// This file is licensed under the Apache License v2.0 with LLVM Exceptions.
+// See https://llvm.org/LICENSE.txt for license information.
+// SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+//
+// (c) Copyright 2023-2024 Advanced Micro Devices, Inc. or its affiliates
+//
+//===----------------------------------------------------------------------===//
+//
+// This file declares the AIEngine V2 Address Space and DM banks
```
gbossu, line 11: Nit: Please check with @abnikant or @konstantinschwarz , but I think we are starting to move version-specific files into separate sub-folders. Maybe we should move that new file into a `aie2/` folder already.

llvm/lib/Target/AIE/AIE2Subtarget.cpp
```diff
@@ -76,3 +76,46 @@ const RegisterBankInfo *AIE2Subtarget::getRegBankInfo() const {
 InstructionSelector *AIE2Subtarget::getInstructionSelector() const {
   return InstSelector.get();
 }
+
+std::bitset<32>
+AIE2Subtarget::getMemoryBanksFromAddressSpace(unsigned AddrSpace) const {
```
gbossu, line 91: Maybe we could move that to AIE2InstrInfo instead? Then we can make that a virtual in `AIEBaseInstrInfo`, which is a real base class for all AIE `InstrInfo` classes.

llvm/lib/Target/AIE/AIEHazardRecognizer.cpp
```diff
@@ -392,31 +404,45 @@ auto toHazardType(bool Conflict) {
...
 
 bool AIEHazardRecognizer::checkConflict(
     const ResourceScoreboard<FuncUnitWrapper> &Scoreboard,
     const InstrItineraryData *ItinData, unsigned SchedClass, SlotBits SlotSet,
+    MemoryBankBits MemoryBanks, std::optional<int> MemoryAccessCycle,
     int DeltaCycles, std::optional<int> FUDepthLimit) {
   assert(Scoreboard.isValidDelta(DeltaCycles));
 
-  // Verify format hazards
+  // Verify slot hazards
```
gbossu: I think format hazards was better. We don't just check that the slots are free, we also check that a format can accommodate all the new slots.

llvm/lib/Target/AIE/AIEHazardRecognizer.cpp
```diff
@@ -447,32 +473,44 @@ bool AIEHazardRecognizer::checkConflict(
...
     const InstrItineraryData *ItinData, unsigned SchedClass, SlotBits SlotSet,
+    MemoryBankBits MemoryBanks, std::optional<int> MemoryAccessCycle,
     int DeltaCycles, std::optional<int> FUDepthLimit) {
   assert(Scoreboard.isValidDelta(DeltaCycles));
 
   // Append slot usage
   FuncUnitWrapper EmissionCycle(/*Req=*/0, /*Res=*/0, SlotSet);
   Scoreboard[DeltaCycles] |= EmissionCycle;
 
+  // Append memory bank usage
```
gbossu, line 518: I would be happy with an assert like `assert(MemoryBanks.emptry() || MemoryAccessCycle.has_value())`

llvm/lib/Target/AIE/AIE2Subtarget.cpp
krishnamtibrewala: As discussed earlier we would like this to use with AA pass based on AddressSpace, and thus decided to keep it here moving there can make things challenging.

llvm/lib/Target/AIE/AIE2AddrSpace.h
krishnamtibrewala: have updated the enum to c++ enum and they way suggested above.

llvm/lib/Target/AIE/AIE2Subtarget.cpp
```diff
@@ -76,3 +76,53 @@ const RegisterBankInfo *AIE2Subtarget::getRegBankInfo() const {
...
+  using namespace AIE2;
+  std::bitset<32> MemoryBanks;
+
+  switch (static_cast<AddressSpaces>(AddrSpace)) {
+  case AddressSpaces::a:
+    MemoryBanks.set(static_cast<int>(DMBanks::A));
```
martien-de-jong: There's no way to write this as `return {DMBanks::A, DMBanks::B};` ?

llvm/lib/Target/AIE/AIEBaseSubtarget.cpp
```diff
@@ -133,6 +133,13 @@ void AIEBaseSubtarget::adjustSchedDependency(
   }
 }
 
+std::bitset<32>
+AIEBaseSubtarget::getMemoryBanksFromAddressSpace(unsigned AddrSpace) const {
+  // By default assume there are no conflicts.
+  std::bitset<32> MemoryBanks;
+  return MemoryBanks;
```
martien-de-jong: `return {};` ?

llvm/lib/Target/AIE/AIEHazardRecognizer.cpp
```diff
@@ -563,3 +602,14 @@ AIEHazardRecognizer::getSelectedAltOpcode(MachineInstr *MI) const {
     return It->second;
   return std::nullopt;
 }
+
+unsigned AIEHazardRecognizer::getMemoryBanks(MachineInstr *MI) const {
```
martien-de-jong: unsigned -> MemoryBankBits?

llvm/lib/Target/AIE/AIEHazardRecognizer.cpp
```diff
@@ -563,3 +602,14 @@ AIEHazardRecognizer::getSelectedAltOpcode(MachineInstr *MI) const {
...
+  unsigned MemoryBankUsed = 0;
```
martien-de-jong: This is a hot function in the scheduler. Perhaps we could first check mayLoad/mayStore for an early return.

llvm/lib/Target/AIE/AIEHazardRecognizer.cpp
```diff
@@ -454,32 +480,45 @@ bool AIEHazardRecognizer::checkConflict(
 }
 
 void AIEHazardRecognizer::emitInScoreboard(const MCInstrDesc &Desc,
+                                           MemoryBankBits MemoryBanks,
                                            int DeltaCycles) {
   enterResources(
       Scoreboard, ItinData, Desc.getSchedClass(),
       getSlotSet(Desc, *TII->getFormatInterface(), IgnoreUnknownSlotSets),
-      DeltaCycles, FUDepthLimit);
+      MemoryBanks, TII->getFirstMemoryCycle(Desc.getSchedClass()), DeltaCycles,
```
martien-de-jong: A- What about anything up to and including LastMemoryCycle?
B- I think a load and a store don't conflict in the same bank.
(B) can be fixed by only registering memory bank use by mayLoad instructions
for (A) we may need to have a more precise description of memory access, however currently I think our only problematic case is part-word stores, that have two memory cycles, the first of which is the load.

llvm/lib/Target/AIE/AIE2Subtarget.cpp
```diff
@@ -76,3 +76,53 @@ const RegisterBankInfo *AIE2Subtarget::getRegBankInfo() const {
...
+    break;
+  case AddressSpaces::b:
+    MemoryBanks.set(static_cast<int>(DMBanks::B));
```
andcarminati: I recommend to cast to `unsigned` here, because the parameter is `std::size_t` and we have the same definition as in `C`:

> C Standard Section 4.1.5 defines size_t as an unsigned integral type of the result of the sizeof operator

llvm/lib/Target/AIE/AIE2AddrSpace.h
```diff
@@ -0,0 +1,46 @@
...
+  c,     // Address space for DM Bank C
+  d,     // Address space for DM Bank D
+  ab,    // Address space for DM Bank A and B
+  ac,    // Address space for DM Bank A and C
+  ad,    // Address space for DM Bank A and D
+  bc,    // Address space for DM Bank B and C
+  bd,    // Address space for DM Bank B and D
+  cd,    // Address space for DM Bank C and D
+  TM     // Address space for TM (Tile Memory)
+};
```
martien-de-jong, line 39: I would like to have a clarifying comment. AFAICS, we assume conflict between e.g. AB and BD, which I guess is safe. I just don't know what the programmer may assume and what the compiler may try to optimize when declaring an object in e.g. BD.
I could imagine that using analysis the compiler could try to specialize a BD object to D in order not to make it conflict with AB. It would then need to retype all derived pointers to D. 
Alternatively, a BD object somehow straddles both B and D, which means it cannot be specialize to D. 
'for DM Bank B and D' sounds ambiguous to me.

llvm/lib/Target/AIE/AIEHazardRecognizer.h
```diff
@@ -26,6 +26,7 @@
 #include "llvm/MC/MCInstrItineraries.h"
 
 namespace llvm {
+using MemoryBankBits = uint64_t;
```
andcarminati: I think that, as everything is handled as 32bit (bitmask, `AIEHazardRecognizer::getMemoryBanks`), for consistency we can use `uint32_t`, or just `unsigned`, as `getMemoryBanks`.

llvm/lib/Target/AIE/AIE2Subtarget.cpp
```diff
@@ -76,3 +76,53 @@ const RegisterBankInfo *AIE2Subtarget::getRegBankInfo() const {
...
+  case AddressSpaces::bd:
+    MemoryBanks.set(static_cast<int>(DMBanks::B))
+        .set(static_cast<int>(DMBanks::D));
+    break;
+  case AddressSpaces::cd:
+    MemoryBanks.set(static_cast<int>(DMBanks::C))
+        .set(static_cast<int>(DMBanks::D));
+    break;
+  default:
+    // For unimplemented cases assume none
```
andcarminati: I think we need to think about this point! `none` here means empty bitmask and no collision inside the slot. Do we need a pessimistic or optimistic treatment? What do you think @martien-de-jong and @gbossu?

martien-de-jong: I think this is inconsistent. DMBanks::none should conflict with every other bank, and I would like to be complete, so not allow unimplemented cases here.
There was a suggestion to have an option to be optimistic or pessimistic, but I think that may also be partly hardwired in the 'BD conflicts with everything containing B or D' implementation decision.

llvm/lib/Target/AIE/AIEHazardRecognizer.cpp
krishnamtibrewala: I will think more about part (A) and how to handle that, 
w.r.t to part (B) we **do** have conflict in the same bank for load & store.

llvm/lib/Target/AIE/AIE2Subtarget.cpp
krishnamtibrewala: I think this is inconsistent. DMBanks::none should conflict with every other bank
--> There are two way of looking at how to handle DMBanks::none 
Case 1. If we consider it as it conflicts with every other bank then due to lack of info we end up forcing 2 vector loads from different bank in two different bundle leaving HW no choice.
Case 2. If we consider it as conflicting with none and like previous case if the 2 vector loads are in different banks then HW will send them and no conflict but if they do conflict HW will add a stall and handle it.
Given these two cases and HW support to handle conflict for case of DMBanks::none I would prefer we let it not conflict with other

I would like to be complete, so not allow unimplemented cases here.
--> The only unimplemented case here is DMBanks::stack, currently we cannot provide exact information of which bank the stack belongs so and is also not annotated by us or the user. Also user might use addresspace(AS) by themself and have some AS which is outside the scope.`

Note: A/B/C/D/AB/.... are virtual banks and have nothing to do with the physical allocation.

llvm/lib/Target/AIE/AIEHazardRecognizer.cpp
gbossu: Correct me if I'm wrong: Realistically, we only care about bank conflicts for loads. We have only one store unit, so we do not need to track parallel write accesses. If we have a parallel load and store to the same bank, I would assume this is fine as well, as we have distinct read and write ports. This is only problematic if those touch the same location, but if that happens, we have bigger problems and our AA does not work.

So I think it's fine to only care about `TII->getFirstMemoryCycle()` because all the loads happen in the same cycle (at least for AIE2). We can add a comment in the code to explain this, and maybe an assert like `assert(TII->getFirstMemoryCycle(...) == TII->getMinFirstMemoryCycle());`

Please share your thoughts @martien-de-jong @krishnamtibrewala :)

llvm/lib/Target/AIE/AIE2Subtarget.cpp
gbossu: I would vouch for a consistent and conservative approach for now, i.e.

- AddressSpaces::a conflicts with `DMBanks::A`
- AddressSpaces::ac conflicts with both `DMBanks::A` and `DMBanks::C`
- Default conflicts with every bank
- etc.

If that's too conservative, we can adjust in a follow-up PR.

llvm/lib/Target/AIE/AIEHazardRecognizer.cpp
```diff
@@ -563,3 +602,16 @@ AIEHazardRecognizer::getSelectedAltOpcode(MachineInstr *MI) const {
...
+  unsigned MemoryBankUsed = 0;
+  if (MI->mayLoad() || MI->mayStore()) {
+    const AIEBaseSubtarget &STI = AIEBaseSubtarget::get(*MI->getMF());
+    for (auto &MMO : MI->memoperands()) {
+      unsigned MemoryBank =
+          (STI.getMemoryBanksFromAddressSpace(MMO->getAddrSpace())).to_ulong();
+      MemoryBankUsed |= MemoryBank;
+    }
+  }
+  return MemoryBankUsed;
```
gbossu, line 625: Nit: To limit indentation, I'd prefer an early return like
```
if (!MI->mayLoad())
  return {};
```

I also think we shouldn't care about stores, as mentioned in https://github.com/Xilinx/llvm-aie/pull/95/files#r1668455116

llvm/unittests/Target/AIE/HazardRecognizerTest.cpp
```diff
@@ -375,3 +379,41 @@ TEST(HazardRecognizer, composeConflicting) {
     EXPECT_EQ(HR.hazard(9, C), C == -1 || C == 0);
   }
 }
+
+/// Check scoreboard conflicts from bank utilization
+TEST(HazardRecognizer, bankConflictHazard) {
+  AIE2InstrInfo InstrInfo;
+  MockHR HR(InstrInfo);
+
+  HR.emit(1, -2, /*SlotSet=*/0b0, /*MemoryBanks=*/0b1010,
```
gbossu: Nit: You emit everything with a delta of `-2`, I think we could just use `0`, which is a nicer number for reasonning :D

llvm/lib/Target/AIE/AIEHazardRecognizer.cpp
krishnamtibrewala: We cannot do LOAD & STORE to same bank that will lead to bank conflict and will trigger HW stall, have confirmed this with ARCH team and personally verified it by understanding the simulation model of AIE2.

llvm/lib/Target/AIE/AIE2Subtarget.cpp
krishnamtibrewala: Ok sounds good to me! 
Can we go other way around just for `Default conflicts with every bank` ?

The reason being today a vector load & store without any bank annotation could be packed in a same bundle but for default/none case if we consider conflict will all the bank then this wont happen effecting lot of aie llvm tests & QoR tests.

krishnamtibrewala: On second thought, I do not think the idea of `Default conflicts with every bank` aligns with the AIE-ML arch. If we assume this was the case then at first place the HW would not have to support Core-Stall on bank conflict

llvm/include/llvm/Target/AIETarget.td
```diff
@@ -20,15 +20,18 @@ class PreSchedInstExpansion<Instruction Target, Instruction Barrier> {
 }
 
 // Scheduling information for instructions that touch memory.
-class MemoryCycles<int First, int Last = -1> {
+class MemoryCycles<list<int> MemCycles> {
+  // Cycles in which memory is accessed 
+  list<int> MemCyclesList = MemCycles;
```
krishnamtibrewala, line 25: Changes done as part of this commit captures all the cycles in which memory is accessed as a list. The idea of firstMemCycle & LastMemCycle is still kept intact.

We will be using this list of cycles to reserve bank in those cycles.  

This approach is future proof in event we have instruction accessing memory in more than two cycles.

andcarminati: I think that you can complement the comments with this rationale.

llvm/lib/Target/AIE/AIE2Subtarget.cpp
gbossu: If we do track conflicts between loads and stores to the same bank, then yes, this is probably too conservative to assume unknown banks will conflict. That worked nicely if we only cared about loads, as we would prevent using LDA/LDB in parallel when not sure. Preventing LD/ST in parallel is a bigger problem.

llvm/lib/Target/AIE/AIEBaseInstrInfo.cpp
```diff
@@ -707,6 +707,11 @@ int AIEBaseInstrInfo::getMaxLastMemoryCycle() const {
   return std::numeric_limits<unsigned>().min();
 }
 
+std::optional<std::vector<int>>
+AIEBaseInstrInfo::getMemoryCycles(unsigned SchedClass) const {
+  return std::nullopt;
```
andcarminati: nit: As this is a default basic implementation for a base class method, it does not hurt to be directly implemented in the .h file.

llvm/test/CodeGen/AIE/aie2/schedule/resource/memory_bank.mir
```diff
@@ -0,0 +1,1083 @@
+# NOTE: Assertions have been autogenerated by utils/update_mir_test_checks.py
+#
+# This file is licensed under the Apache License v2.0 with LLVM Exceptions.
+# See https://llvm.org/LICENSE.txt for license information.
+# SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+#
+# (c) Copyright 2023-2024 Advanced Micro Devices, Inc. or its affiliates
```
andcarminati: nit: just 2024.

llvm/lib/Target/AIE/AIEHazardRecognizer.cpp
```diff
@@ -49,51 +49,63 @@ void FuncUnitWrapper::setFormatInterface(const AIEBaseMCFormats *Formats) {
...
 bool FuncUnitWrapper::isEmpty() const {
-  return Required == 0 && Reserved == 0 && Slots == 0;
+  return Required == 0 && Reserved == 0 && Slots == 0 && MemoryBanks == 0;
 }
 
 void FuncUnitWrapper::blockResources() {
   Required = ~0;
   Reserved = ~0;
   Slots = ~0;
+  MemoryBanks = ~0;
```
krishnamtibrewala: TODO : Based on debug & discussion, we plan to not block Memory Bank Resources.

llvm/lib/Target/AIE/AIEHazardRecognizer.h
```diff
@@ -67,8 +72,11 @@ class FuncUnitWrapper {
   FuncUnitWrapper() = default;
   FuncUnitWrapper(InstrStage::FuncUnits Req) : Required(Req), Reserved(0) {}
   FuncUnitWrapper(InstrStage::FuncUnits Req, InstrStage::FuncUnits Res,
-                  SlotBits Slots = 0)
+                  SlotBits Slots)
```
gbossu: Do we need that constructor? Can we remove it and only have the new one below?

llvm/lib/Target/AIE/AIEHazardRecognizer.cpp
```diff
@@ -49,51 +49,65 @@ void FuncUnitWrapper::setFormatInterface(const AIEBaseMCFormats *Formats) {
...
+  return Required == 0 && Reserved == 0 && Slots == 0 && MemoryBanks == 0;
 }
 
 void FuncUnitWrapper::blockResources() {
   Required = ~0;
   Reserved = ~0;
   Slots = ~0;
+  // Since the HW stalls in the event of memory bank conflicts, we don't need to
+  // block the resource. It is overly conservative if we block all memory banks.
+  MemoryBanks = 0;
```
gbossu: That is resetting the bank bits. If there were some already set, I think we should keep them.

llvm/lib/Target/AIE/AIE2Subtarget.cpp
martien-de-jong: I think that if bd means b or d, then none means a or b or c or d, but we implement bd as b and d, blocking both b and d when it occurs. I'm still waiting for a clarifying comment as to what the user expects. Since none means 'no information', my feeling is that it should conflict with everything, so have 0xf as a mask.

llvm/lib/Target/AIE/AIEHazardRecognizer.cpp
```diff
@@ -49,51 +49,64 @@ void FuncUnitWrapper::setFormatInterface(const AIEBaseMCFormats *Formats) {
...
-  return Required == 0 && Reserved == 0 && Slots == 0;
+  return Required == 0 && Reserved == 0 && Slots == 0 && MemoryBanks == 0;
 }
 
 void FuncUnitWrapper::blockResources() {
   Required = ~0;
   Reserved = ~0;
   Slots = ~0;
+  // Since the HW stalls in the event of memory bank conflicts, we don't need to
+  // block the resource. It is overly conservative if we block all memory banks.
```
martien-de-jong, line 96: The comment is superfluous. It makes more sense to have MemoryBanks = ~0. The function is used to inform the scheduler that not a single resource should land there, and there's no use in wrecking your brain to find out whether MemoryBanks should be exempt. If there's a regression connected to this I would be interested, since I think it's hard to place a memory bank resource there without hitting any of the other blocks.

llvm/lib/Target/AIE/AIEHazardRecognizer.cpp
```diff
@@ -563,3 +607,23 @@ AIEHazardRecognizer::getSelectedAltOpcode(MachineInstr *MI) const {
...
+  const AIEBaseSubtarget &STI = AIEBaseSubtarget::get(*MI->getMF());
+  for (auto &MMO : MI->memoperands()) {
+    unsigned MemoryBank =
+        (STI.getMemoryBanksFromAddressSpace(MMO->getAddrSpace())).to_ulong();
+    MemoryBankUsed |= MemoryBank;
+  }
+  if (MI->mayStore() && MemoryBankUsed == 1) {
+    // If its a Store instruction and it does not have Bank info then we allow
+    // it to go with a load instruction
+    MemoryBankUsed = 0;
```
martien-de-jong: What does the magic literal '1' mean here? And is there a good reason to treat store bank conflicts different from load bank conflicts? It looks as if supplying more information will lead to more scheduling restrictions here.

llvm/lib/Target/AIE/AIE2Subtarget.cpp
martien-de-jong: Assuming equal likelihoods, AB and BC only conflict in one {BB} in four {AB, AC, BB, BC} cases, yet we assume a conflict, and may force an extra cycle. Taking this further, we could maintain a probability for each bank, and literally add those, detecting a conflict if one of them exceeds a threshold. Since we don't need high precision on those probabilities, we can space them out in 32 bit not to be bothered by inter-element carry using a normal 32 bit add and use a mask to detect >= 50% in any of the elements.

llvm/lib/Target/AIE/AIEHazardRecognizer.cpp
```diff
@@ -563,3 +607,23 @@ AIEHazardRecognizer::getSelectedAltOpcode(MachineInstr *MI) const {
...
+MemoryBankBits AIEHazardRecognizer::getMemoryBanks(MachineInstr *MI) const {
+  MemoryBankBits MemoryBankUsed = 0;
+
+  if (!(MI->mayLoad() || MI->mayStore()))
+    return MemoryBankUsed;
+
+  const AIEBaseSubtarget &STI = AIEBaseSubtarget::get(*MI->getMF());
+  for (auto &MMO : MI->memoperands()) {
+    unsigned MemoryBank =
+        (STI.getMemoryBanksFromAddressSpace(MMO->getAddrSpace())).to_ulong();
```
gbossu: Nit: Do we need the extra parenthesis?

llvm/lib/Target/AIE/AIEHazardRecognizer.cpp
gbossu: Also, do we really need a magic bit to signify `unknown`? Why not just have `~0`, i.e. all the banks?

I would also slightly change the rationale for special-casing stores. It's mainly to maintain the status-quo, but also because in the kernels we deal with, the input and output buffers are usually in separate banks. So it is generally fine to be optimistic. Once the kernels are more specific with their bank usage, we can drop that special-casing and be fully-pessimistic when it comes to "unknown banks"

llvm/lib/Target/AIE/AIE2Subtarget.cpp
krishnamtibrewala: Hi @martien-de-jong ,
Kindly refer to the communication done in https://github.com/Xilinx/llvm-aie/pull/95#discussion_r1663956837

llvm/lib/Target/AIE/AIEHazardRecognizer.cpp
krishnamtibrewala: We had a discussion regarding the same and had decided to be optimistically handle memory bank bits while we do scheduling across basic block. Today we have case where we force all the resources to be blocked while we do bottom up scheduling which is extremely pessimistic when it comes to bank resources.

In the following case when we schedule VST_CONV (which has an AS of 6) which translate into Memorybanks bit 1 to be used in the 7th cycle and thus we need to reserve Memory Bank in 7th cycle relative to emitting of the same, the 7th cycle now lies in the bot-scoreboard which is fully reserved thus leading to -6 delta cycle compared to -2 without AS. This behavior is caused due to blocking. 

image

Your (@martien-de-jong) opinions when we discussed this was 
` I agree, memory bank contention could be handled optimistically. I'm not sure how easy it is to separate recording the backs from recording the functional units. That should probably be accommodated in the top-level HazardRecognizer interfaces, perhaps using a state member of it.`

We should be completely optimistic here since the worst we can do is create a bank conflict and the HW support of stall mitigates it which is not possible for FU/Slots.

llvm/lib/Target/AIE/AIEHazardRecognizer.cpp
krishnamtibrewala: As noted [here ](https://github.com/Xilinx/llvm-aie/pull/95/commits/a4baa6d9da1fc16d277c324a0b2c0a3c72ecf0de#diff-7b59f2472b3801ded7e5587449e12c8791893c5dc1a3a0ae58eb9f62279b6426:~:text=//%20Bit%200%20is%20for%20MI%20with%20no%20bank%20information%2C) bit 0 represent no/unknown bank annotation.  

We need to have this information to make sure we do not allow VLDA & VLDB in event we do not have bank information available.
Also "STORE" are handled differently because we would want to allow one VLD with VST, even if we do not have a bank annotation.

llvm/lib/Target/AIE/AIE2AddrSpace.h
krishnamtibrewala: Based on my understanding of how buffers are allocated in Data Memory of AIE Tile, they are in Ping-Pong fashion where in Ping is in a different bank Pong is in different bank. That is where AB, BD , etc comes in play.  In case say the IFM is on bank 0 & bank 1 as per mapper constrain the annotation would be AB. 

At least from the IFM, WFM, OFM buffer we cannot optimize the buffer using the compiler to a single bank. While one bank say Ping is being read by core , usually Pong is either read out/in by the DMA.

llvm/lib/Target/AIE/AIEBaseSubtarget.cpp
```diff
@@ -133,6 +135,21 @@ void AIEBaseSubtarget::adjustSchedDependency(
   }
 }
 
+const AIEBaseSubtarget &AIEBaseSubtarget::get(const MachineFunction &MF) {
+  if (MF.getTarget().getTargetTriple().isAIE1())
+    return static_cast<const AIEBaseSubtarget &>(
+        MF.getSubtarget<AIESubtarget>());
+  else if (MF.getTarget().getTargetTriple().isAIE2())
+    return static_cast<const AIEBaseSubtarget &>(
+        MF.getSubtarget<AIE2Subtarget>());
```
andcarminati, line 144: I suggest include this line:

```
llvm_unreachable("Unknown subtarget");
```

llvm/lib/Target/AIE/AIEHazardRecognizer.cpp
```diff
@@ -454,32 +483,47 @@ bool AIEHazardRecognizer::checkConflict(
 }
 
 void AIEHazardRecognizer::emitInScoreboard(const MCInstrDesc &Desc,
```
andcarminati, line 491: I think it is time to refactor this function, like:

```
void AIEHazardRecognizer::emitInScoreboard(const MCInstrDesc &Desc,
                                           MemoryBankBits MemoryBanks,
                                           int DeltaCycles) {

   emitInScoreboard(Scoreboard, Desc, MemoryBanks, DeltaCycles);
}

```

llvm/lib/Target/AIE/AIEHazardRecognizer.cpp
krishnamtibrewala: Based on todays(31st July 2024) discussion we have concluded on not to treat STORES differently.

llvm/lib/Target/AIE/AIEHazardRecognizer.cpp
krishnamtibrewala: Based on todays(31st July 2024) discussion we have concluded on not to reserve MemoryBank resources when doing inter basic block scheduling.

llvm/lib/Target/AIE/AIEHazardRecognizer.cpp
krishnamtibrewala: I checked the QoR without special consideration of STORE, there are NO regression.

llvm/lib/Target/AIE/AIE2Subtarget.cpp
martien-de-jong: I think the bottom line is we never bothered about memory banks, since the hardware would correct it. That means the only rationale to do this is to be more precise, and give the user handles to tell what the situation is. If we say that unannotated banks don't interfere with others, we are undermining our own rationale. It would be different if 'no annotation' would be distinct from an 'assume no conflict' annotation.

llvm/lib/Target/AIE/AIEHazardRecognizer.cpp
```diff
@@ -399,31 +412,46 @@ auto toHazardType(bool Conflict) {
...
 AIEHazardRecognizer::getHazardType(const MCInstrDesc &Desc,
+                                   MemoryBankBits MemoryBanks,
                                    const int DeltaCycles) {
-  return toHazardType(checkConflict(
-      Scoreboard, ItinData, Desc.getSchedClass(),
+  return getHazardType(
+      Desc.getSchedClass(),
       getSlotSet(Desc, *TII->getFormatInterface(), IgnoreUnknownSlotSets),
-      DeltaCycles, FUDepthLimit));
+      MemoryBanks, DeltaCycles);
```
martien-de-jong, line 427: Where did FUDepthLimit go?

llvm/lib/Target/AIE/AIEHazardRecognizer.cpp
```diff
@@ -399,31 +412,46 @@ auto toHazardType(bool Conflict) {
...
 
   // Verify format hazards
   FuncUnitWrapper EmissionCycle(/*Req=*/0, /*Res=*/0, SlotSet);
   if (EmissionCycle.conflict(Scoreboard[DeltaCycles]))
     return true;
 
+  // Verify memory bank hazards
+  if (MemoryAccessCycles.has_value()) {
+    for (auto Cycles : MemoryAccessCycles.value()) {
+      FuncUnitWrapper MemoryBankAccessCycle(/*Req=*/0, /*Res=*/0, /*SlotSet=*/0,
```
martien-de-jong: I think this can be moved out of the loop.

llvm/lib/Target/AIE/AIEHazardRecognizer.cpp
```diff
@@ -454,32 +482,43 @@ bool AIEHazardRecognizer::checkConflict(
...
   // Append slot usage
   FuncUnitWrapper EmissionCycle(/*Req=*/0, /*Res=*/0, SlotSet);
   Scoreboard[DeltaCycles] |= EmissionCycle;
 
+  // Append memory bank usage
+  assert((MemoryBanks == 0) || MemoryAccessCycles.has_value());
+  if (MemoryAccessCycles.has_value()) {
+    for (auto Cycles : MemoryAccessCycles.value()) {
+      FuncUnitWrapper MemoryBankAccessCycle(/*Req=*/0, /*Res=*/0, /*SlotSet=*/0,
+                                            MemoryBanks);
```
martien-de-jong: Loop invariant here too.

llvm/lib/Target/AIE/AIEHazardRecognizer.cpp
```diff
@@ -399,31 +412,46 @@ auto toHazardType(bool Conflict) {
...
   FuncUnitWrapper EmissionCycle(/*Req=*/0, /*Res=*/0, SlotSet);
   if (EmissionCycle.conflict(Scoreboard[DeltaCycles]))
     return true;
 
+  // Verify memory bank hazards
+  if (MemoryAccessCycles.has_value()) {
+    for (auto Cycles : MemoryAccessCycles.value()) {
+      FuncUnitWrapper MemoryBankAccessCycle(/*Req=*/0, /*Res=*/0, /*SlotSet=*/0,
+                                            MemoryBanks);
+      if (MemoryBankAccessCycle.conflict(Scoreboard[DeltaCycles + Cycles - 1]))
```
martien-de-jong, line 457: Add a comment that MemoryAccessCycles regrettably starts counting at 1?

llvm/lib/Target/AIE/AIEHazardRecognizer.cpp
```diff
@@ -563,3 +602,18 @@ AIEHazardRecognizer::getSelectedAltOpcode(MachineInstr *MI) const {
...
+  MemoryBankBits MemoryBankUsed = 0;
+
+  if (!(MI->mayLoad() || MI->mayStore()))
+    return MemoryBankUsed;
+
+  const AIEBaseSubtarget &STI = AIEBaseSubtarget::get(*MI->getMF());
+  for (auto &MMO : MI->memoperands()) {
+    unsigned MemoryBank =
+        STI.getMemoryBanksFromAddressSpace(MMO->getAddrSpace()).to_ulong();
+    MemoryBankUsed |= MemoryBank;
```
martien-de-jong: I see a bitset, an unsigned long, an unsigned (int) and a MemoryBankBits, all designating the same type within two lines. I propose to have a single STI.getMemoryBankBitsFromAddressSpace() to hide the implementation details.

llvm/lib/Target/AIE/AIEHazardRecognizer.cpp
krishnamtibrewala: With the above change we will call following function, and thus `FUDepthLimit` is abstracted there.  

```
ScheduleHazardRecognizer::HazardType
AIEHazardRecognizer::getHazardType(unsigned SchedClass, SlotBits SlotSet,
                                   MemoryBankBits MemoryBanks,
                                   int DeltaCycles) {
  return toHazardType(checkConflict(
      Scoreboard, ItinData, SchedClass, SlotSet, MemoryBanks,
      TII->getMemoryCycles(SchedClass), DeltaCycles, FUDepthLimit));
}
```

llvm/test/TableGen/aie-memory-cycles.td
```diff
@@ -75,3 +75,13 @@ let Itinerary = II_OTHER in
 // CHECK-NEXT:  return 5;
 // CHECK-LABEL: int TestAIEInstrInfo::getMaxLastMemoryCycle() const {
 // CHECK-NEXT:  return 11;
+
+// CHECK:      std::optional<std::vector<int>>
```
gbossu: That's a rather expensive type to return. How about a `SmallVector<int, 2>`? I don't think we need the optional, for instructions that do not access memory, we can just return an empty vector.

llvm/lib/Target/AIE/AIEHazardRecognizer.cpp
```diff
@@ -399,31 +412,47 @@ auto toHazardType(bool Conflict) {
...
+    std::optional<int> FUDepthLimit) {
   assert(Scoreboard.isValidDelta(DeltaCycles));
 
   // Verify format hazards
   FuncUnitWrapper EmissionCycle(/*Req=*/0, /*Res=*/0, SlotSet);
   if (EmissionCycle.conflict(Scoreboard[DeltaCycles]))
     return true;
 
+  // Verify memory bank hazards
+  if (MemoryAccessCycles.has_value()) {
```
gbossu: See comment before, I think we should pass `const SmallVector<int, 2> &MemoryAccessCycles` here, and just iterate over the elements.

llvm/lib/Target/AIE/AIEHazardRecognizer.cpp
gbossu: We could change that if it's confusing. So far it maps to the stage numbers in the ISA, but I agree counting from 1 is weird territory for programmers 😄

llvm/lib/Target/AIE/AIEHazardRecognizer.cpp
```diff
@@ -454,32 +483,43 @@ bool AIEHazardRecognizer::checkConflict(
 }
 
 void AIEHazardRecognizer::emitInScoreboard(const MCInstrDesc &Desc,
+                                           MemoryBankBits MemoryBanks,
                                            int DeltaCycles) {
-  enterResources(
-      Scoreboard, ItinData, Desc.getSchedClass(),
-      getSlotSet(Desc, *TII->getFormatInterface(), IgnoreUnknownSlotSets),
-      DeltaCycles, FUDepthLimit);
+  emitInScoreboard(Scoreboard, Desc, MemoryBanks, DeltaCycles);
```
gbossu, line 494: Yay, more shared code 💪

llvm/lib/Target/AIE/AIEHazardRecognizer.cpp
```diff
@@ -563,3 +603,18 @@ AIEHazardRecognizer::getSelectedAltOpcode(MachineInstr *MI) const {
...
+  MemoryBankBits MemoryBankUsed = 0;
+
+  if (!(MI->mayLoad() || MI->mayStore()))
+    return MemoryBankUsed;
+
+  const AIEBaseSubtarget &STI = AIEBaseSubtarget::get(*MI->getMF());
+  for (auto &MMO : MI->memoperands()) {
+    MemoryBankBits MemoryBank =
+        STI.getMemoryBanksFromAddressSpace(MMO->getAddrSpace());
+    MemoryBankUsed |= MemoryBank;
```
gbossu: What happens for TM accesses? For those, I think we have two MMOs: one from the IR value, and a custom one to disambiguate DM from TM. If the IR value does not have the right address space, I guess we end up blocking all banks, even though we know from the second MMO that it's in the TM bank?

gbossu: We could maybe actually start with `MemoryBankUsed = ~0;`, and let every MMO refine the banks with `MemoryBankUsed &= MemoryBank` instead of `|=`. 

In the first commit, I think we should define `STI.getMemoryBanksFromPseudoSourceValue` as well. E.g.
```
getMemoryBanksFromPseudoSourceValue(const PseudoSourceValue &PSV) {
  switch (PSV.getKind()) {
  case TileMemory: return Bank::TileMemory
  default: return Bank::all()
  }
}
```

Then in the loop here, we can use getMemoryBanksFromPseudoSourceValue if `MMO.getPseudoValue()` returns a valid pointer.

Edit: I didn't know that a generic `getAddressSpaceForPseudoSourceKind` existed. We should obviously use that instead of defining a new AIE-specific `getMemoryBanksFromPseudoSourceValue`

llvm/test/CodeGen/AIE/aie2/schedule/implicit_ops.mir
```diff
@@ -57,10 +57,8 @@ alignment:       16
 body:             |
   bb.0.entry:
     ; CHECK-LABEL: name: implicit_WAW
-    ; CHECK: BUNDLE implicit-def $wl0, implicit-def $x0, implicit-def $wh0, implicit killed $p0, implicit killed $p1 {
-    ; CHECK-NEXT:   $wl0 = VLDA_dmw_lda_w_ag_idx_imm killed $p0, 0, implicit-def $x0 :: (load (<8 x s32>) from stack - 64)
-    ; CHECK-NEXT:   $wh0 = VLDB_dmw_ldb_ag_idx_imm killed $p1, 0 :: (load (<8 x s32>) from stack - 32)
-    ; CHECK-NEXT: }
+    ; CHECK: $wl0 = VLDA_dmw_lda_w_ag_idx_imm killed $p0, 0, implicit-def $x0 :: (load (<8 x s32>) from stack - 64)
+    ; CHECK-NEXT: $wh0 = VLDB_dmw_ldb_ag_idx_imm killed $p1, 0 :: (load (<8 x s32>) from stack - 32)
```
gbossu: That test now changed meaning, can you add addresspace info to the MMOs so both loads can actually inssue in parallel?

llvm/test/CodeGen/AIE/aie2/schedule/vlda_vldb.mir
```diff
@@ -51,3 +52,42 @@ body:             |
...
+    ; CHECK-NEXT:   $wl1 = VLDB_dmw_ldb_ag_idx_imm killed $p0, 0
+    ; CHECK-NEXT: }
+    ; CHECK-NEXT: NOP
+    ; CHECK-NEXT: NOP
+    ; CHECK-NEXT: NOP
+    ; CHECK-NEXT: NOP
+    ; CHECK-NEXT: NOP
+    ; CHECK-NEXT: NOP
+    $wl1 = VLDB_dmw_ldb_ag_idx_imm $p0, 0
+    $wl2 = VLDA_dmw_lda_w_ag_idx_imm $p0, 0
```
gbossu, line 90: I think this is confusing, in absence of MMO, we should block all the banks. This goes in the direction of my comment here: https://github.com/Xilinx/llvm-aie/pull/95/commits/b2d99ac48855c1ebf1cb1574df7dbc86239be7af#r1711136944

llvm/lib/Target/AIE/AIE2Subtarget.cpp
```diff
@@ -119,6 +124,9 @@ AIE2Subtarget::getMemoryBanksFromAddressSpace(unsigned AddrSpace) const {
     MemoryBanks.set(static_cast<unsigned>(DMBanks::C))
         .set(static_cast<unsigned>(DMBanks::D));
     break;
+  case AddressSpaces::stack:
```
gbossu: It seems strange for me to have a stack address space. I'd like to keep them purely for banks, and instead move the decision making to `getAddressSpaceForPseudoSourceKind()`. There we can decide which address space a `PSV` maps to.

So I think we should rename the option to e.g. `--aie-stack-addrspace` and set it to `0` by default. Only in our QoR test suite we can say it actually maps to `AddressSpaces::a`. We can add a command line option there, but we should keep the compiler generic here.

llvm/test/CodeGen/AIE/aie2/GlobalISel/inst-select-indexed-vlda_ups.mir
```diff
@@ -717,36 +717,36 @@ body: |
     ; CHECK-NEXT: [[COPY2:%[0-9]+]]:er = COPY $r1
     ; CHECK-NEXT: [[PADD_mod_pseudo:%[0-9]+]]:ep = PADD_mod_pseudo [[COPY]], [[COPY1]]
     ; CHECK-NEXT: [[COPY3:%[0-9]+]]:mss = COPY [[COPY2]]
-    ; CHECK-NEXT: [[VLDA_UPS_S64_D32_ag_idx_imm:%[0-9]+]]:acc512 = VLDA_UPS_S64_D32_ag_idx_imm [[COPY3]], [[PADD_mod_pseudo]], 32, implicit-def $srups_of, implicit $crsat, implicit $crupssign :: (load (<8 x s32>) from stack - 32, basealign 64)
+    ; CHECK-NEXT: [[VLDA_UPS_S64_D32_ag_idx_imm:%[0-9]+]]:acc512 = VLDA_UPS_S64_D32_ag_idx_imm [[COPY3]], [[PADD_mod_pseudo]], 32, implicit-def $srups_of, implicit $crsat, implicit $crupssign :: (load (<8 x s32>) from stack - 32, basealign 64, addrspace 4)
```
gbossu: As stated above, without additional info, these should map `addrspace 0`

llvm/lib/Target/AIE/AIE2Subtarget.cpp
gbossu: As a side bonus, this should remove the ref updates for the tests below, and I would add a new small test with e.g. `--aie-stack-addrspace=4` to verify it works.

gbossu (commented):
Lots of great improvements, thanks for also splitting the code in nice commits! Please look through my remaining commits, we are almost there 💪

llvm/test/CodeGen/AIE/aie2/schedule/implicit_ops.mir
krishnamtibrewala: Yes.

llvm/lib/Target/AIE/AIE2AddrSpace.h
```diff
@@ -0,0 +1,46 @@
...
+namespace llvm {
+
+namespace AIE2 {
+
+enum class AddressSpaces {
+  none, // Default address space
+  PM,   // Address space for Program Memory (PM)
+  DM,   // Address space for Data Memory(DM) includes Bank A, B, C, D
+  DM_test,
+  stack, // Address space for stack
```
gbossu, line 27: Are `DM_test` and `stack` actually used somewhere?

llvm/lib/Target/AIE/AIE2Subtarget.cpp
```diff
@@ -76,3 +76,54 @@ const RegisterBankInfo *AIE2Subtarget::getRegBankInfo() const {
...
+    MemoryBanks.set(static_cast<unsigned>(DMBanks::C))
+        .set(static_cast<unsigned>(DMBanks::D));
+    break;
+  default:
+    // For unimplemented cases assume all
+    MemoryBanks.set();
+    break;
+  }
+
+  return MemoryBanks.to_ulong();
```
gbossu, line 140: Why not return as a bitset actually?

llvm/lib/Target/AIE/AIEHazardRecognizer.h
```diff
@@ -67,8 +74,8 @@ class FuncUnitWrapper {
   FuncUnitWrapper() = default;
   FuncUnitWrapper(InstrStage::FuncUnits Req) : Required(Req), Reserved(0) {}
   FuncUnitWrapper(InstrStage::FuncUnits Req, InstrStage::FuncUnits Res,
-                  SlotBits Slots = 0)
-      : Required(Req), Reserved(Res), Slots(Slots) {}
+                  SlotBits Slots, MemoryBankBits MemoryBanks = 0)
```
gbossu: Nit: Why not keep `Slots = 0` as well?

llvm/lib/Target/AIE/AIE2Subtarget.cpp
gbossu: At least maybe use `MemoryBankBits` as return type?

llvm/lib/Target/AIE/AIEHazardRecognizer.h
```diff
@@ -26,6 +26,9 @@
 #include "llvm/MC/MCInstrItineraries.h"
 
 namespace llvm {
+// Bit 0-3 if for bank A, B, C, D respectively
+// Bit 4 if for Tile Memory
```
gbossu: Nit: This comment will get out of date, do we need it?

llvm/test/CodeGen/AIE/aie2/end-to-end/Mul2D.ll
```diff
@@ -85,17 +85,17 @@ define void @mul2d(ptr noalias %in_ptr0, ptr noalias %in_ptr1, ptr noalias %out_
...
+; CHECK-NEXT:    jnz r1, #.LBB0_2
 ; CHECK-NEXT:    nop // Delay Slot 5
-; CHECK-NEXT:    nop // Delay Slot 4
-; CHECK-NEXT:    nop // Delay Slot 3
-; CHECK-NEXT:    vst.srs.d8.s32 cm0, s0, [p2], #32 // Delay Slot 2
-; CHECK-NEXT:    vst.srs.d8.s32 cm1, s0, [p2], #32 // Delay Slot 1
+; CHECK-NEXT:    vst.srs.d8.s32 cm0, s0, [p2], #32 // Delay Slot 4
+; CHECK-NEXT:    vst.srs.d8.s32 cm1, s0, [p2], #32 // Delay Slot 3
+; CHECK-NEXT:    nop // Delay Slot 2
+; CHECK-NEXT:    nop // Delay Slot 1
```
gbossu: It seems that the loop-aware scheduling stopped working, we should not have NOPs here. Is this because the vst.srs have a bank conflict with the vlda of the next iteration?

gbossu: If it's not too complicated, I'd be happy if you could add `addrspace` info to some pointers in the IR below.

llvm/lib/Target/AIE/AIEHazardRecognizer.cpp
```diff
@@ -563,3 +600,21 @@ AIEHazardRecognizer::getSelectedAltOpcode(MachineInstr *MI) const {
     return It->second;
   return std::nullopt;
 }
+
+MemoryBankBits AIEHazardRecognizer::getMemoryBanks(MachineInstr *MI) const {
+  MemoryBankBits MemoryBankUsed = 0;
+
+  if (!(MI->mayLoad() || MI->mayStore()))
+    return MemoryBankUsed;
```
gbossu: Nit: return `0` directly, and `~0` below. We can then move `MemoryBankBits MemoryBankUsed = 0;` down, closer to where it is used.

llvm/test/CodeGen/AIE/aie2/schedule/resource/memory_bank.mir
```diff
@@ -0,0 +1,1119 @@
...
+    ; CHECK-NEXT: NOP
+    ; CHECK-NEXT: NOP
+    ; CHECK-NEXT: NOP
+    ; CHECK-NEXT: NOP
+    ; CHECK-NEXT: NOP
+    $wl1 = VLDB_dmw_ldb_ag_idx_imm $p0, 0 :: (load (<8 x s32>), addrspace 5)
+    $wl2 = VLDA_dmw_lda_w_ag_idx_imm $p0, 0 :: (load (<8 x s32>), addrspace 6)
+    $wh1 = VLDB_dmw_ldb_ag_idx_imm $p0, 0 :: (load (<8 x s32>), addrspace 7)
+    $wh2 = VLDA_dmw_lda_w_ag_idx_imm $p0, 0 :: (load (<8 x s32>), addrspace 8)
+...
```
gbossu, line 1119: Can you add some test cases for TM loads? E.g.
 - One VLDB without MMO and one LD.TM with TileMemory MMO
 - One VLDB with addrspace MMO and one LD.TM with TileMemory MMO

For both cases I believe we should be able to schedule the two loads in the same cycle.

llvm/lib/Target/AIE/AIEHazardRecognizer.cpp
gbossu: See my comment here regarding extra tests: https://github.com/Xilinx/llvm-aie/pull/95/commits/87b1999cc7a112eb783f16e5bfc1047b6845836a#r1713401640

I think we should use some initialization like `MemoryBankUsed = STI.getDefaultMemoryBanks()`, where this returns `A|B|C|D` for AIE2. I.e. excluding the TM bank. Then inside the loop, every MMO refines the banks with `MemoryBankUsed &= MemoryBank` instead of `|=`.

This should give us better disambiguation between DM and TM memories.

llvm/lib/Target/AIE/AIEHazardRecognizer.cpp
gbossu, line 96: Could you add a unit test in `HazardRecognizerTest` to show that `blockResources` does not touch MemoryBanks?

llvm/lib/Target/AIE/AIE2TargetMachine.cpp
```diff
@@ -259,3 +264,14 @@ bool AIE2PassConfig::addInstSelector() {
...
 }
+
+unsigned
+AIE2TargetMachine::getAddressSpaceForPseudoSourceKind(unsigned Kind) const {
+  switch (Kind) {
+  case PseudoSourceValue::Stack:
+  case PseudoSourceValue::FixedStack:
+    return StackAddrSpace;
+  default:
+    return static_cast<unsigned>(AIE2::AddressSpaces::none);
```
gbossu, line 278: Based on my comment https://github.com/Xilinx/llvm-aie/pull/95#discussion_r1713409406 I think we should return `AIE2::AddressSpaces::all` in the default case, ~and handle `TileMemory` Kind here as well.~

llvm/lib/Target/AIE/AIE2TargetMachine.cpp
```diff
@@ -271,6 +271,8 @@ AIE2TargetMachine::getAddressSpaceForPseudoSourceKind(unsigned Kind) const {
   case PseudoSourceValue::Stack:
   case PseudoSourceValue::FixedStack:
     return StackAddrSpace;
+  case PseudoSourceValue::TargetCustom:
```
gbossu: Please use `AIETargetPSV::AIETileMem` instead of `TargetCustom`

llvm/test/CodeGen/AIE/aie2/schedule/memory_access_DM_TM.mir
```diff
@@ -0,0 +1,89 @@
...
+    ; CHECK-NEXT: }
+    ; CHECK-NEXT: NOP
+    ; CHECK-NEXT: NOP
+    ; CHECK-NEXT: NOP
+    ; CHECK-NEXT: NOP
+    ; CHECK-NEXT: NOP
+    ; CHECK-NEXT: NOP
+    $r0 = LDA_TM $p0 :: (load (s32) from custom "TileMemory")
+    $wl2 = VLDB_dmw_ldb_ag_idx_imm $p0, 0 :: (load (<8 x s32>), addrspace 8)
+...
```
gbossu, line 110: See https://github.com/Xilinx/llvm-aie/pull/95#discussion_r1713401640
Could you add a test where VLDB has no MMO? I think it should also get bundled. Maybe you can do all the changes suggested [here](https://github.com/Xilinx/llvm-aie/pull/95#discussion_r1713409406) in this commit actually.

gbossu (commented):
Please have a look at the comments I left, mostly to better disambiguate between TM and DM memories. We are almost there 💪

llvm/lib/Target/AIE/AIE2AddrSpace.h
krishnamtibrewala: `stack` could be used as per AIE-API, Not sure about DM_test but we need to have the same `enum` number for flows to work in future. We can discuss this in more detail.

llvm/lib/Target/AIE/AIEHazardRecognizer.h
krishnamtibrewala: Will remove it, kept it for improving readability. You point is valid it might get out of date.

llvm/lib/Target/AIE/AIE2TargetMachine.cpp
krishnamtibrewala: Here we need to return the default address space that is 0 which is AIE2::AddressSpaces::none.
This will translate into reserving all the bankbits.

llvm/test/CodeGen/AIE/aie2/schedule/resource/memory_bank.mir
krishnamtibrewala: These tests are added in  `llvm/test/CodeGen/AIE/aie2/schedule/memory_access_DM_TM.mir `

llvm/lib/Target/AIE/AIEHazardRecognizer.cpp
krishnamtibrewala: Updated as suggested.

llvm/lib/Target/AIE/AIE2Subtarget.cpp
krishnamtibrewala: Updated as suggested.

llvm/test/CodeGen/AIE/aie2/end-to-end/Mul2D.ll
krishnamtibrewala: Updated as suggested. The test expected result are not same as before the introduction of AS

llvm/lib/Target/AIE/AIE2AddrSpace.h
```diff
@@ -39,6 +39,7 @@ enum class AddressSpaces {
 };
 
 enum class DMBanks { A, B, C, D };
+enum class TileMemory { TileMemory = 4 };
```
gbossu: Nit: I think there should be a single enum, e.g. `enum class AIEBanks { A, B, C, D, Tile };`

gbossu (dismissed):
LGTM, great work :)

llvm/lib/Target/AIE/AIE2AddrSpace.h
krishnamtibrewala: Update DMBanks/TileMemory to AIEBanks

gbossu (approved):

PR #97, opened 2024-06-26
[AIE2][ValueTracking] Add support for AIE2 Intrinsics

andcarminati:
With this change, BasicAA can handle AIE2 intrinsics, applying all existent algorithms (aliasPHI, for example).

### DIFF
--- a/llvm/lib/Analysis/ValueTracking.cpp
+++ b/llvm/lib/Analysis/ValueTracking.cpp
@@ -9,6 +9,9 @@
 // This file contains routines that help analyze properties that chains of
 // computations have.
 //
+// Modifications (c) Copyright 2023-2024 Advanced Micro Devices, Inc. or its
+// affiliates
+//
 //===----------------------------------------------------------------------===//
 
 #include "llvm/Analysis/ValueTracking.h"
@@ -56,6 +59,7 @@
 #include "llvm/IR/IntrinsicInst.h"
 #include "llvm/IR/Intrinsics.h"
 #include "llvm/IR/IntrinsicsAArch64.h"
+#include "llvm/IR/IntrinsicsAIE2.h"
 #include "llvm/IR/IntrinsicsAMDGPU.h"
 #include "llvm/IR/IntrinsicsRISCV.h"
 #include "llvm/IR/IntrinsicsX86.h"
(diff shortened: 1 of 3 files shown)


llvm/lib/Analysis/ValueTracking.cpp
```diff
@@ -5939,6 +5956,9 @@ const Value *llvm::getUnderlyingObject(const Value *V, unsigned MaxLookup) {
           V = RP;
           continue;
         }
+      } else if (auto *AIEObject = getUnderlyingObjectAIEIntrinsic(V)) {
```
gbossu, line 5959: Could you summarize how that differs from our previous approach of having a separate AA pass for AIE intrinsics?

andcarminati: Hi @gbossu, basically, our pass can handle just linear code, in this way, we can decompose only in upwards direction. There is no simple way to, for example, switch between different analysis implementation when we don't understand a specific node (PHI, for example). Once we are running the BasicAA, and we reach a PHI node, for example, we cannot easily decide which AA in the chain can handle all the incoming values from that node. 

One approach could be the propagation of the PHI handling inside our AA, however we will definitely handle PHI chains twice (AIEAA an BasicAA)  for the cases containing our specific intrinsics.

andcarminati: I feel that it is better to have one heavy AA that can handle everything than two heavy AAs. If the first fail to prove NoAlias, we need to run everything again (next AA).

gbossu: I think I agree. Let's discuss with @konstantinschwarz but I think we can then add a second commit that removes our AIE-specific AA pass.

gbossu: Let's delay the removal of our custom AA wrapper pass. We might anyway need it for address space disambiguation.

gbossu (approved):
LGTM :)

martien-de-jong (approved):
I wonder whether we could get this functionality in a generic hook next to the address space  ones, and include both items in BasicAA upstream.

andcarminati:
> I wonder whether we could get this functionality in a generic hook next to the address space ones, and include both items in BasicAA upstream.

I think we can address this in the future, as a follow-up work.

PR #98, opened 2024-06-27
[AIEX] Fix memory leak error

krishnamtibrewala:
Changes in this PR fixed following error from address sanitizer 

image


andcarminati (approved):
LGTM. I think this is better than creating another version of `::create` with TargetMachine param just for us.

PR #100, opened 2024-06-27
[AIE2] Check all divisions for shufflevector extraction

ValentijnvdBeek:
Optimizes the additional cases of the extraction intrinsics. These are the cases where the destination vector fits more than twice in the target vector.

### DIFF
--- a/llvm/lib/CodeGen/GlobalISel/CombinerHelper.cpp
+++ b/llvm/lib/CodeGen/GlobalISel/CombinerHelper.cpp
@@ -524,36 +524,24 @@ bool CombinerHelper::tryCombineShuffleVector(MachineInstr &MI) {
   }
 
   // {1, 2, ..., |DstVector|} -> G_UNMERGE_VALUES
-  // Extracts the first chunk of the same size of the destination vector from
-  // the source
-  GeneratorType FirstQuarter = adderGenerator(0, DstNumElts - 1, 1);
-  if (matchCombineShuffleVector(MI, FirstQuarter, DstNumElts - 1)) {
-    // This optimization does not work if the target type is not a power of two,
-    // this can happen in some backends that support uneven vector types. We
-    // also need to make sure that the vector can be split into two.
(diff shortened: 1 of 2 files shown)


llvm/lib/CodeGen/GlobalISel/CombinerHelper.cpp
```diff
@@ -436,36 +436,24 @@ bool CombinerHelper::tryCombineShuffleVector(MachineInstr &MI) {
...
+  // Extracts the chunks of the same size of the destination vector from the
+  // source
+  for (uint8_t Current = 0, Total = SrcNumElts; Current < Total; Current++) {
+    uint32_t Start = Current * DstNumElts, End = Start + DstNumElts - 1;
+    GeneratorType Generator = adderGenerator(Start, End, 1);
+    if (matchCombineShuffleVector(MI, Generator, DstNumElts - 1)) {
+      // This optimization does not work if the target type is not a power of
+      // two, this can happen in some backends that support uneven vector types.
+      // We also need to make sure that the vector can be split into two.
+      if (SrcTy == DstTy || ((SrcNumElts / 2) % 2) != 0 ||
```
martien-de-jong, line 536: I don't understand the SrcNumElts condition. 5 will get rounded down by the division to an even number. Also, the comment says 'power of two', but I see no test for that. Perhaps you mean 'multiple of two' ?

ValentijnvdBeek: Oh yeah, I forgot that in C/C++ it is binary division. The reason why this still works is that the broken case is AMDGPU which uses 3 which does become uneven, but that could be done.

The wording is confusing, here we just need to check if the given type is even. I don't know how to word it precisely, but my intention is to communicate is that we assume that the architecture targets a power of two. This optimization breaks at the point whenever we get an uneven number of elements of our vector. The call itself can have uneven numbers provided it is able to split at least once cleanly.

PR #101, opened 2024-06-27
[AIEX] Form correct bundles in postmisched

gbossu:
This is very important as the scheduling DAG construction uses the standard query type for attributes like `mayLoad` or `hasSideEffects`. Querying properties for `ADD` in the bundle below can give unexpected/wrong edges.
```
r0 = ADD r0, r1 {
  ST r3, p0
}
```

Now such bundles are always properly formed:
```
BUNDLE {
  r0 = ADD r0, r1 
  ST r3, p0
}
```

More details in commits


llvm/test/CodeGen/AIE/aie2/aie2bundle.mir
```diff
@@ -16,7 +16,7 @@ body:             |
     ; CHECK: BUNDLE implicit-def $r0, implicit $r0, implicit $r1 {
     ; CHECK-NEXT:   $r0 = OR $r0, $r1
     ; CHECK-NEXT: }
-    ; CHECK-NEXT: BUNDLE implicit-def $r3, implicit $r1, implicit $r2 {
```
martien-de-jong, line 19: This looks like a functional change. I'm not aware that we need bundle operands, but I don't understand what changed.

llvm/test/CodeGen/AIE/aie2/schedule/loopaware/fixpoint.mir
```diff
@@ -30,7 +30,8 @@ body:             |
   ; CHECK-NEXT: {{  $}}
   ; CHECK-NEXT:   $r10 = OR $r0, $r1
   ; CHECK-NEXT:   $r3 = OR killed $r10, $r2
-  ; CHECK-NEXT:   $cm3 = VMOV_mv_cm $cm1 {
+  ; CHECK-NEXT:   BUNDLE implicit-def $cm3, implicit-def $bml3, implicit-def $amll3, implicit-def $amlh3, implicit-def $bmh3, implicit-def $amhl3, implicit-def $amhh3, implicit-def $cm2, implicit-def $bml2, implicit-def $amll2, implicit-def $amlh2, implicit-def $bmh2, implicit-def $amhl2, implicit-def $amhh2, implicit $cm1, implicit $cm0, implicit killed $r3 {
+  ; CHECK-NEXT:     $cm3 = VMOV_mv_cm $cm1
```
martien-de-jong, line 34: Where do all these implicits come from? Oh, it expands the subregisters?

llvm/test/CodeGen/AIE/schedule/bundled-me-mutation.mir
```diff
@@ -17,40 +17,54 @@
...
-    ; CHECK-NEXT: DelayedSchedBarrier implicit killed $r0
-    $r0 = LDA_dms_lda_idx_imm $p0, 0
-    $r1 = nsw ADD_add_r_ri $r1, -1, implicit-def $srcarry
-    $p0 = PADD_imm_pseudo $p0, 4 {
-      ST_dms_sts_idx_imm $r0, $p1, 0
-    }
-    RET implicit $lr
-    DelayedSchedBarrier implicit $r0
+    liveins: $p0, $cm0, $s0, $r1
+    VST_SRS_D8_S32_ag_idx_imm $p0, 0, $cm0, $s0, implicit-def $srsrs_of, implicit $crsat, implicit $crrnd, implicit $crsrssign
```
martien-de-jong, line 61: This seems to defeat the purpose of the test; it no longer has a bundle as input.

martien-de-jong (dismissed):
Some nitty comments, but looks good overall.

llvm/lib/Target/AIE/AIEBaseSubtarget.cpp
```diff
@@ -224,12 +224,7 @@ class RegionEndEdges : public ScheduleDAGMutation {
 
       SDep ExitDep(&SU, SDep::Artificial);
 
-      // By using IgnoreBundle, we can safely apply this mutation to already
-      // bundled instructions without causing misclassification of instructions
-      // that are bundled with control flow ones. Otherwise, the assertion
-      // below can be triggered for correct cases.
-      unsigned DelaySlots =
-          TII->getNumDelaySlots(MI, MachineInstr::QueryType::IgnoreBundle);
+      unsigned DelaySlots = TII->getNumDelaySlots(MI);
```
andcarminati, line 227: nit: `const`

andcarminati (dismissed):
LGTM. I think now it is be much better without those `QueryType`s.

llvm/test/CodeGen/AIE/aie2/aie2bundle.mir
gbossu: The implicit operands are added when calling `finalizeBundles`. In the finalize pass, we only call it on standalone instructions. For already-bundled instructions, the BUNDLE is left as is.

llvm/test/CodeGen/AIE/schedule/bundled-me-mutation.mir
gbossu: I cannot create BUNDLE instructions as inputs, because we don't have itineraries/sched info for them. Instead, I'm triggering the creation of bundles in the loop, and then those bundles will be used in AIEInterblockScheduling when connecting the loop region with the epilogue region.

andcarminati (approved):

PR #103, opened 2024-06-28
[Clang][BasicAA][AIE2] Enable full PHI AA for AIE

andcarminati:
Also increase the max limit of the search depth in DecomposeGEPExpression. With this change we can allow LICM to hoist more instructions.

### DIFF
--- a/clang/lib/Driver/ToolChains/AIE.cpp
+++ b/clang/lib/Driver/ToolChains/AIE.cpp
@@ -113,6 +113,12 @@ void AIEToolChain::addClangTargetOptions(
   // Make sure to perform most optimizations before mandatory inlinings,
   // otherwise noalias attributes can get lost and hurt AA results.
   CC1Args.append({"-mllvm", "-mandatory-inlining-before-opt=false"});
+
+  // Perform complete AA analysis on phi nodes.
+  CC1Args.append({"-mllvm", "-basic-aa-full-phi-analysis=true"});
+
+  // Extend the max limit of the search depth in BasicAA
+  CC1Args.append({"-mllvm", "-basic-aa-max-lookup-search-depth=10"});
 }
 
 // Avoid using newer dwarf versions, as the simulator doesn't understand newer
--- a/llvm/lib/Analysis/BasicAliasAnalysis.cpp
+++ b/llvm/lib/Analysis/BasicAliasAnalysis.cpp
@@ -4,6 +4,9 @@
 // See https://llvm.org/LICENSE.txt for license information.
 // SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
 //
+// Modifications (c) Copyright 2024 Advanced Micro Devices, Inc. or its
+// affiliates
+//
 //===----------------------------------------------------------------------===//
 //
 // This file defines the primary stateless implementation of the
@@ -68,8 +71,18 @@ using namespace llvm;
 static cl::opt<bool> EnableRecPhiAnalysis("basic-aa-recphi", cl::Hidden,
                                           cl::init(true));
 
+/// Enable full analysis of PHI nodes.
+static cl::opt<bool> EnableFullPHIAnalysis("basic-aa-full-phi-analysis",
+                                           cl::Hidden, cl::init(false));
+
 static cl::opt<bool> EnableSeparateStorageAnalysis("basic-aa-separate-storage",
                                                    cl::Hidden, cl::init(true));
+
+// The max limit of the search depth in DecomposeGEPExpression() and
+// getUnderlyingObject().
+static cl::opt<unsigned>
+    MaxLookupSearchDepth("basic-aa-max-lookup-search-depth", cl::Hidden,
+                         cl::init(6));
 
 /// SearchLimitReached / SearchTimes shows how often the limit of
 /// to decompose GEPs is reached. It will affect the precision
@@ -77,10 +90,6 @@ static cl::opt<bool> EnableSeparateStorageAnalysis("basic-aa-separate-storage",
 STATISTIC(SearchLimitReached, "Number of times the limit to "
                               "decompose GEPs is reached");
 STATISTIC(SearchTimes, "Number of times a GEP is decomposed");
-
-// The max limit of the search depth in DecomposeGEPExpression() and
-// getUnderlyingObject().
-static const unsigned MaxLookupSearchDepth = 6;
 
 bool BasicAAResult::invalidate(Function &Fn, const PreservedAnalyses &PA,
                                FunctionAnalysisManager::Invalidator &Inv) {
@@ -1389,7 +1398,7 @@ AliasResult BasicAAResult::aliasPHI(const PHINode *PN, LocationSize PNSize,
     if (PV1 == PN)
       continue;
 
-    if (isa<PHINode>(PV1)) {
+    if (!(EnableRecPhiAnalysis && EnableFullPHIAnalysis) && isa<PHINode>(PV1)) {
       if (OnePhi && OnePhi != PV1) {
         // To control potential compile time explosion, we choose to be
         // conserviate when we have more than one Phi input.  It is important
--- a/llvm/test/Analysis/BasicAA/phi-full-aa-diamond.ll
+++ b/llvm/test/Analysis/BasicAA/phi-full-aa-diamond.ll
@@ -1,0 +1,126 @@
+;
+; This file is licensed under the Apache License v2.0 with LLVM Exceptions.
+; See https://llvm.org/LICENSE.txt for license information.
+; SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+;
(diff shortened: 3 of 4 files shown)


llvm/lib/Analysis/BasicAliasAnalysis.cpp
```diff
@@ -68,6 +68,10 @@ using namespace llvm;
 static cl::opt<bool> EnableRecPhiAnalysis("basic-aa-recphi", cl::Hidden,
                                           cl::init(true));
 
+/// Enable full analysis of PHI nodes.
+static cl::opt<bool> EnableFullPHIAnalysis("basic-aa-full-phi-analysis",
+                                           cl::Hidden, cl::init(true));
```
martien-de-jong: nit: We initialize it to true by default and pass it from the driver as true.

llvm/lib/Analysis/BasicAliasAnalysis.cpp
```diff
@@ -80,7 +84,7 @@ STATISTIC(SearchTimes, "Number of times a GEP is decomposed");
 
 // The max limit of the search depth in DecomposeGEPExpression() and
 // getUnderlyingObject().
-static const unsigned MaxLookupSearchDepth = 6;
+static const unsigned MaxLookupSearchDepth = 10;
```
martien-de-jong: I think you won't get this upstreamed. Make it a command-line option and pass it from the driver?

llvm/lib/Analysis/BasicAliasAnalysis.cpp
```diff
@@ -1389,7 +1393,7 @@ AliasResult BasicAAResult::aliasPHI(const PHINode *PN, LocationSize PNSize,
     if (PV1 == PN)
       continue;
 
-    if (isa<PHINode>(PV1)) {
+    if (!EnableFullPHIAnalysis && isa<PHINode>(PV1)) {
```
martien-de-jong: I feel ill at ease with not setting OnePhi, since this is used below. Would it make sense to just control the conservative return here?

llvm/lib/Analysis/BasicAliasAnalysis.cpp
andcarminati: Oh, my mistake, fixed.

llvm/lib/Analysis/BasicAliasAnalysis.cpp
andcarminati: Done!

llvm/lib/Analysis/BasicAliasAnalysis.cpp
andcarminati: By not setting OnePhi, we will give the possibility to run some more pessimistic step further.

See:

```
  if (OnePhi && UniqueSrc.size() > 1)
    // Out of an abundance of caution, allow only the trivial lcssa and
    // recursive phi cases.
    return AliasResult::MayAlias;
```

If I just control the return, I loose some opportunities....

gbossu: Not sure I understand the surrounding code properly, but does that mean that for a diamond cfg, the phi node would not be followed because it has two distinct (non recursive) sources? Would it be possible to add a small `llc` test where we pass the `--basic-aa-full-phi-analysis` option?

andcarminati: Hi @gbossu, I extended the existing test to accommodate this change.

andcarminati: Hi @gbossu, I discover that a more complete phi analysis support was dropped in the past: https://reviews.llvm.org/D139719

gbossu (dismissed):
I looked through `llvm/lib/Analysis/BasicAliasAnalysis.cpp` to understand your change within more context. This LGTM! I'm sure it was quite challenging to trace back our problem to here 💪 

If possible, I'd be even happier with a simpler additional test case. Maybe a diamond cfg with one non-recursive phi node that has two incoming values, one for each side of the diamond. Would that be enough to show that `--basic-aa-full-phi-analysis=false` is not enough?

andcarminati:
> I looked through `llvm/lib/Analysis/BasicAliasAnalysis.cpp` to understand your change within more context. This LGTM! I'm sure it was quite challenging to trace back our problem to here 💪
> 
> If possible, I'd be even happier with a simpler additional test case. Maybe a diamond cfg with one non-recursive phi node that has two incoming values, one for each side of the diamond. Would that be enough to show that `--basic-aa-full-phi-analysis=false` is not enough?

Hi @gbossu, I included an additional test. I think now we can clearly see the difference regarding the behavior.

llvm/lib/Analysis/BasicAliasAnalysis.cpp
```diff
@@ -4,6 +4,9 @@
 // See https://llvm.org/LICENSE.txt for license information.
 // SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
 //
+// Modifications (c) Copyright 2023-2024 Advanced Micro Devices, Inc. or its
```
gbossu: Nit: Only 2024

llvm/test/Analysis/BasicAA/phi-full-aa-diamond.ll
```diff
@@ -0,0 +1,68 @@
...
+; The basic shape of the test is the following (simplified without phis):
+; if (...) {
+;    if (...) {
+;      %P = getelementptr @X ...
+;    } else {
+;      %P = getelementptr @X ...
+;    }    
+; } else {
+;   %P = getelementptr @X ...
+; }
```
gbossu, line 29: Thanks for the pseudo code! :)

gbossu (dismissed):

gbossu (dismissed):

clang/lib/Driver/ToolChains/AIE.cpp
```diff
@@ -113,6 +113,12 @@ void AIEToolChain::addClangTargetOptions(
   // Make sure to perform most optimizations before mandatory inlinings,
   // otherwise noalias attributes can get lost and hurt AA results.
   CC1Args.append({"-mllvm", "-mandatory-inlining-before-opt=false"});
+
+  // Perform complete AA analysis on phi nodes.
+  CC1Args.append({"-mllvm", "-basic-aa-full-phi-analysis=true"});
+
+  // Extend the max limit of the search depth in BasicAA
+  CC1Args.append({"-mllvm", "-basic-aa-max-lookup-search-depth=10"});
```
martien-de-jong, line 121: I think we (ie in the average AIE application) wouldn't really notice it if we make it 20, (or 50?)

llvm/lib/Analysis/BasicAliasAnalysis.cpp
```diff
@@ -68,20 +71,26 @@ using namespace llvm;
 static cl::opt<bool> EnableRecPhiAnalysis("basic-aa-recphi", cl::Hidden,
                                           cl::init(true));
 
+/// Enable full analysis of PHI nodes.
+static cl::opt<bool> EnableFullPHIAnalysis("basic-aa-full-phi-analysis",
+                                           cl::Hidden, cl::init(false));
+
```
martien-de-jong, line 77: I am worried that EnableFullPhiAnalysis=true relies on EnableRecPhiAnalysis=true for correctness.
For instance, there is a conservative return at line 1450 that is controlled by isRecursive, which is controlled by EnableRecPhiAnalysis.

martien-de-jong: Discussed offline: if we use the && of these two conditions it looks mych safer.

andcarminati: Done!

clang/lib/Driver/ToolChains/AIE.cpp
andcarminati: By looking to the statistics I saw this limit being exhausted for one benchmark.

martien-de-jong: So 20 would be better?

martien-de-jong (approved):

gbossu (approved):
LGTM!, Are you looking still looking into https://github.com/Xilinx/llvm-aie/pull/103#discussion_r1681194413 ?

clang/lib/Driver/ToolChains/AIE.cpp
andcarminati: Hi @martien-de-jong, I believe that 10 is enough for now, considering our benchmarks that are quite complex.

andcarminati:
> LGTM!, Are you looking still looking into [#103 (comment)](https://github.com/Xilinx/llvm-aie/pull/103#discussion_r1681194413) ?

I gave a resolution, I think we can keep as 10.

PR #104, opened 2024-06-28
[AIE2] Combine some vector shifts by zero

andcarminati:
Something like:
```
From : %1:_(<64 x s8>) = G_INTRINSIC intrinsic(@llvm.aie2.v64int8/v32int16)
        %2:_(<16 x s32>) = G_BITCAST %1:_(<64 x s8>)
        %3:_(s32) = G_CONSTANT i32 0
        %4:_(<16 x s32>) = G_INTRINSIC
        intrinsic(@llvm.aie2.vshift.I512.I512), %X:_(<16 x s32>), %2:_(<16 x
        s32>), %3:_(s32), %3:_(s32)
To :    %4:_(<16 x s32>) = COPY %X
Or even:
From : %1:_(<64 x s8>) = G_INTRINSIC intrinsic(@llvm.aie2.v16int32)
        %2:_(s32) = G_CONSTANT i32 0
        %3:_(<16 x s32>) = G_INTRINSIC
        intrinsic(@llvm.aie2.vshift.I512.I512), %X:_(<16 x s32>), %1:_(<16 x
        s32>), %2:_(s32), %2:_(s32)
 To :   %3:_(<16 x s32>) = COPY %X
```


llvm/lib/Target/AIE/AIE2PreLegalizerCombiner.cpp
```diff
@@ -89,6 +97,83 @@ AIE2PreLegalizerCombinerImpl::AIE2PreLegalizerCombinerImpl(
...
+      return false;
+    }
+  }
+
+  auto IsConstantZeroReg = [&](const Register Reg) {
+    auto Cst = getIConstantVRegValWithLookThrough(Reg, MRI);
+    return Cst && (Cst->Value.getZExtValue() == 0);
+  };
+
+  if (!IsConstantZeroReg(ThirdSrcReg) || !IsConstantZeroReg(ShiftAmtSrcReg))
```
martien-de-jong, line 129: I expect this to occur frequently. Can we move the condition up and avoid some work?

llvm/lib/Target/AIE/AIE2PreLegalizerCombiner.cpp
```diff
@@ -89,6 +97,83 @@ AIE2PreLegalizerCombinerImpl::AIE2PreLegalizerCombinerImpl(
...
+    return false;
+
+  MI.removeOperand(5); // remove shift amount.
+  MI.removeOperand(3); // remove other vec.
+  MI.removeOperand(3);
```
martien-de-jong: I am confused. I guess the numbering changes, and we remove first 3, and then the new 3, so the old 4? 
Can't we do 5, 4, 3 then?

llvm/lib/Target/AIE/AIE2PreLegalizerCombiner.cpp
```diff
@@ -89,6 +97,83 @@ AIE2PreLegalizerCombinerImpl::AIE2PreLegalizerCombinerImpl(
...
+
+  if (const GIntrinsic *Intrinisc = dyn_cast<GIntrinsic>(SecondDefInst)) {
+    if (!IsIn512BitsUndef(Intrinisc->getIntrinsicID())) {
+      return false;
+    }
+  }
+
+  auto IsConstantZeroReg = [&](const Register Reg) {
+    auto Cst = getIConstantVRegValWithLookThrough(Reg, MRI);
+    return Cst && (Cst->Value.getZExtValue() == 0);
```
martien-de-jong: use isZero() ?

llvm/lib/Target/AIE/AIE2PreLegalizerCombiner.cpp
andcarminati: Changed, now it is linear.

gbossu:
Completely different idea: I wonder if we could recognize `xxxx = VSHIFT xxxx, 0` as a copy in `AIE2InstrInfo::isCopyInstrImpl`? Then I would assume it would be simplified by the reg coalescer?

andcarminati:
> Completely different idea: I wonder if we could recognize `xxxx = VSHIFT xxxx, 0` as a copy in `AIE2InstrInfo::isCopyInstrImpl`? Then I would assume it would be simplified by the reg coalescer?

Hi @gbossu, probably it can work as well. However we need to recognize more instructions together, like the move that give us the shift amount. If this move is hoisted in some other place it can be difficult to prove that it really contains the zero. From the legalization perspective we just have constants here and SSA to help. Do you think we can have some benefit?

llvm/lib/Target/AIE/AIE2PreLegalizerCombiner.cpp
```diff
@@ -89,6 +97,83 @@ AIE2PreLegalizerCombinerImpl::AIE2PreLegalizerCombinerImpl(
...
+      return true;
+    default:
+      break;
+    }
+    return false;
+  };
+
+  if (const GIntrinsic *Intrinisc = dyn_cast<GIntrinsic>(SecondDefInst)) {
+    if (!IsIn512BitsUndef(Intrinisc->getIntrinsicID())) {
+      return false;
```
gbossu: Do we need to check this? I think that if we shift by 0, then the second operand does not matter.

llvm/lib/Target/AIE/AIE2PreLegalizerCombiner.cpp
```diff
@@ -89,6 +97,83 @@ AIE2PreLegalizerCombinerImpl::AIE2PreLegalizerCombinerImpl(
...
+    }
+  }
+
+  MI.removeOperand(5); // remove shift amount.
+  MI.removeOperand(4); // remove step.
+  MI.removeOperand(3); // remove other vector.
+  MI.removeOperand(1); // remove intrinsic id.
+  Helper.replaceOpcodeWith(MI, TargetOpcode::COPY);
```
gbossu: Maybe it is easier to just remove the previous instruction with `eraseFromParent` and add a new one? I think it's actually the "recommended" way to do this in the backend due to implicit ops and MMOs and things...

gbossu (approved):
LGTM :)

PR #105, opened 2024-07-01
Add min/max iteration count pragmas

martien-de-jong:
This provides the information in the form of pragmas, not as attributes. This is less of a break with the way other loop pragmas are handled; there is no precedent for adding attribute syntax, even if the pragmas are represented internally by attributes.

The min and max itercount pragmas are added as top-level entries in the LoopID metadata.

LoopUnroll and LoopPeeling update the attributes after transformations.


clang/test/CodeGen/aie/looppragma.c
```diff
@@ -0,0 +1,51 @@
+// RUN: %clang --target=aie2 -S -emit-llvm %s -o - | FileCheck %s
```
andcarminati, line 10: I think we can consider this as a target-independent test.

martien-de-jong: true.

llvm/lib/Target/AIE/AIEBasePipelinerLoopInfo.cpp
```diff
@@ -36,7 +36,45 @@ cl::opt<bool> TrackRegPressure(
...
+      continue;
+
+    MDString *S = dyn_cast<MDString>(MD->getOperand(0));
+    if (S == nullptr)
+      continue;
+
+    if (S->getString() == "llvm.loop.itercount") {
+      assert(
+          (MD->getNumOperands() >= 2 && MD->getNumOperands() <= 3) &&
+          "Iteration count hint metadata should have two or three operands.");
```
andcarminati: I feel that this message can be a bit confusing. Maybe: "one or two numerical operands".

clang/lib/CodeGen/CGLoopInfo.h
```diff
@@ -82,6 +82,12 @@ struct LoopAttributes {
   /// Value for 'llvm.loop.align' metadata.
   unsigned CodeAlign;
 
+  /// Argument values of llvm.loop.itercount metadata;
+  struct {
+    std::optional<int64_t> Lwb;
+    std::optional<int64_t> Upb;
```
andcarminati, line 91: Just a silly question from my side: why do we care about Upb if take the information after from Lwb?

llvm/lib/Target/AIE/AIEBasePipelinerLoopInfo.cpp
andcarminati: `MinTripCount` is  `int64_t`, is this way, I think it is better to use `getSExtValue`, insted.

llvm/lib/Target/AIE/AIEBasePipelinerLoopInfo.cpp
```diff
@@ -36,7 +36,45 @@ cl::opt<bool> TrackRegPressure(
...
+      MinTripCount =
+          mdconst::extract<ConstantInt>(MD->getOperand(1))->getZExtValue();
+      assert(MinTripCount >= 1 && "Range lwb must be positive.");
+      LLVM_DEBUG(dbgs() << "PLI: MinTripcount from pragma =  " << MinTripCount
```
andcarminati: nit: `MinTripCount`

llvm/lib/Target/AIE/AIEBasePipelinerLoopInfo.cpp
martien-de-jong: There's no way negative numbers make sense as an iteration count. 0 is fine as a "don't know" placeholder.

clang/lib/CodeGen/CGLoopInfo.h
martien-de-jong: We are trying to implement LOOP_RANGE, which has an upper bound. Later we may not bother to create a three stage pipeline if max_iter == 2. I don't know.

llvm/test/CodeGen/AIE/aie2/schedule/swp/min_iter_loopinfo.ll
```diff
@@ -0,0 +1,43 @@
...
+}
+
+!0 = !{i32 1, !"wchar_size", i32 4}
+!1 = !{!"clang version 17.0.0 (git@gitenterprise.xilinx.com:XRLabs/llvm-aie.git c07f38f221ca4d3736152230e12f5cfc637a1d47)"}
+!2 = !{!3, !3, i64 0}
+!3 = !{!"int", !4, i64 0}
+!4 = !{!"omnipotent char", !5, i64 0}
+!5 = !{!"Simple C/C++ TBAA"}
+!6 = distinct !{!6, !7}
+!7 = !{!"llvm.loop.itercount", i64 3}
```
gbossu: I find it a bit confusing to name the MD `llvm.loop.itercount`. This makes it sound like an exact iteration count, while this really is a minimum.

clang/test/CodeGen/aie/looppragma.c
```diff
@@ -0,0 +1,60 @@
+// --- looppragma.c ---------------------------------------------------------///
+// NOTE: Assertions have been autogenerated by utils/update_llc_test_checks.py
+//
+// This file is licensed under the Apache License v2.0 with LLVM Exceptions.
+// See https://llvm.org/LICENSE.txt for license information.
+// SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+//
+// (c) Copyright 2024 Advanced Micro Devices, Inc. or its affiliates
+// --------------------------------------------------------------------------///
+// RUN: %clang --target=aie2 -S -emit-llvm %s -o - | FileCheck %s
```
gbossu, line 10: Maybe also run that test at -O2 to verify the metadata is preserved? I guess for the last test it will not, at least without the next commit.

llvm/test/CodeGen/AIE/aie2/schedule/swp/min_iter_loopinfo.ll
martien-de-jong: It describes a range. This form with one numeric operand means [3, ?)

clang/test/CodeGen/aie/looppragma.c
martien-de-jong: I had intended this to be a unit test for the frontend

llvm/lib/Target/AIE/AIEBasePipelinerLoopInfo.cpp
martien-de-jong: oh, actually everything is signed. Sign extended as requested.

clang/test/CodeGen/aie/looppragma.c
gbossu: Yes let's keep it that way, I see you added tests for -O2 in the next commit.

gbossu (dismissed):
LGTM! Up to you if you want to rename the MD to something like `llvm.loop.itercount.range`

llvm/test/CodeGen/AIE/aie2/schedule/swp/min_iter_loopinfo.ll
```diff
@@ -1,3 +1,4 @@
+;
```
gbossu, line 1: I think that commit can be dropped?

martien-de-jong: [...]

gbossu (approved):

clang/test/CodeGen/aie/peel-itercount.c
```diff
@@ -0,0 +1,35 @@
...
+// NOTE: Assertions have been autogenerated by utils/update_llc_test_checks.py
+//
+// This file is licensed under the Apache License v2.0 with LLVM Exceptions.
+// See https://llvm.org/LICENSE.txt for license information.
+// SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+//
+// (c) Copyright 2024 Advanced Micro Devices, Inc. or its affiliates
+// --------------------------------------------------------------------------///
+
+// RUN: %clang --target=aie2 -O2 -mllvm --unroll-force-peel-count=2 -S -emit-llvm %s -o - | FileCheck %s
```
konstantinschwarz, line 11: Would be good to have some generic tests (ideally `opt`) in llvm itself, that don't depend on AIE.
Maybe in llvm/test/Transforms/LoopUnroll?

PR #106, opened 2024-07-01
[AIE2] Lowering of natively unsupported SEXT_INREGs

andcarminati:
We can decide during the legalization if one SEXT_INREG can be lowered to shifts, this allow existing combine rules to kick in and remove unnecessary shifts.

Now, we moved the implementation to the legalizer.
~~As we select SEXT_INREG using shifts for size > 16, we can have some shift optimization opportunities.~~

~~* Actually, this PR considers the selection, but we can also perform the same task AIEPostSelectOptimize. However, this case is related specifically to SEXT_INREG (other shift optimizations are performed in InstrCombine* or even GISel combiners) in the sense that we don't need to look to all shifts, just those surrounding SEXT_INREG.~~


llvm/test/CodeGen/AIE/aie2/GlobalISel/inst-select-combine-shift-sext.mir
```diff
@@ -0,0 +1,132 @@
...
+    ; CHECK-NEXT: [[MOV_RLC_imm10_pseudo1:%[0-9]+]]:er = MOV_RLC_imm10_pseudo -16
+    ; CHECK-NEXT: [[ASHL:%[0-9]+]]:er = ASHL [[LSHL]], [[MOV_RLC_imm10_pseudo1]]
+    ; CHECK-NEXT: $r0 = COPY [[ASHL]]
+    ; CHECK-NEXT: PseudoRET implicit $lr, implicit $r0
+    %0:ptrregbank(p0) = COPY $p0
+    %1:gprregbank(s32) = G_LOAD %0 :: (load (s32), align 4)
+    %2:gprregbank(s32) = G_CONSTANT i32 4
+    %3:gprregbank(s32) = G_SHL %1:gprregbank(s32), %2:gprregbank(s32)
+    %4:gprregbank(s32) = G_SEXT_INREG %3:gprregbank(s32), 20
+    %5:gprregbank(s32) = G_ASHR %4:gprregbank(s32), %2:gprregbank(s32)
```
konstantinschwarz: Is this combine really specific to AIE?
Could this be written as a generic pre/post-legalization combiner instead?

andcarminati: Hi @konstantinschwarz, thank you for the idea. I completely changed the implementation. I Solved the problem in legalization phase, in this way, all sifts will be combined through already available combiners.

llvm/lib/CodeGen/GlobalISel/LegalizerHelper.cpp
```diff
@@ -7122,6 +7110,23 @@ LegalizerHelper::lowerUnmergeValues(MachineInstr &MI) {
...
+  Register TmpRes = MRI.createGenericVirtualRegister(DstTy);
+
+  auto MIBSz =
+      MIRBuilder.buildConstant(DstTy, DstTy.getScalarSizeInBits() - SizeInBits);
+  MIRBuilder.buildShl(TmpRes, SrcReg, MIBSz->getOperand(0));
+  MIRBuilder.buildAShr(DstReg, TmpRes, MIBSz->getOperand(0));
+  MI.eraseFromParent();
+  return Legalized;
+}
+
```
martien-de-jong, line 7129: CHECK: we factor out this function so that we can selectively use it in our legalizer.

llvm/lib/Target/AIE/AIELegalizerInfo.cpp
```diff
@@ -757,6 +759,23 @@ bool AIELegalizerInfo::legalizeG_UNMERGE_VALUES(LegalizerHelper &Helper,
...
 
+bool AIELegalizerInfo::legalizeG_SEXT_INREG(LegalizerHelper &Helper,
+                                            MachineInstr &MI) const {
+
+  MachineIRBuilder &MIRBuilder = Helper.MIRBuilder;
+  MachineRegisterInfo &MRI = *MIRBuilder.getMRI();
+
+  const Register DestReg = MI.getOperand(0).getReg();
+  const LLT DestRegTy = MRI.getType(DestReg);
+  const LLT S32 = LLT::scalar(32);
```
martien-de-jong, line 770: I keep wondering why we don't have these standard types as constant members in AIELegalizerInfo

llvm/lib/Target/AIE/AIELegalizerInfo.cpp
```diff
@@ -757,6 +759,23 @@ bool AIELegalizerInfo::legalizeG_UNMERGE_VALUES(LegalizerHelper &Helper,
...
+
+  const int64_t Imm = MI.getOperand(2).getImm();
+  if ((Imm != 8 && Imm != 16) || DestRegTy != S32)
+    Helper.lowerSextInreg(MI);
+
+  return true;
+}
```
martien-de-jong, line 777: CHECK: if DestRegTy > S32, the lowering will depend on existing legalization of the 64 bit shifts that are produced.
DestRegTy < S32  does not occur since it has been widened earlier?

llvm/lib/Target/AIE/AIELegalizerInfo.cpp
andcarminati: Yes, makes sense to to refactor this out. Every custom legalization need to create those types again...

llvm/lib/Target/AIE/AIELegalizerInfo.cpp
andcarminati: For the first case, this is covered by the test https://github.com/Xilinx/llvm-aie/blob/aeb9129715bd6baebf12b6b5da5893cb38811fe5/llvm/test/CodeGen/AIE/GlobalISel/legalize-sext-inreg.mir#L102C26-L102C47 . For the second case I believe yes, we did not clamp before, now the behavior is the same.

gbossu:
I think it looks good :) Do you have maybe a small reproducer where we extra instructions that not get combined away?

andcarminati:
> I think it looks good :) Do you have maybe a small reproducer where we extra instructions that not get combined away?

Hi @gbossu , I included a new test, as you suggested.

martien-de-jong (approved):

gbossu (approved):

PR #108, opened 2024-07-01
[AIE2] Offload Vector Loads to SlotB

krishnamtibrewala:
We define MultiSlot Pseudo instructions for Vector Load operation which can be materialized to Slot A or B.

Specific care is taken to make sure requirement of slot are meet. 
--> For example Slot B has lower immediate offset range then slot B, if immediate range is over slot B, Instr. select will use VLDA instr and not the MultiSlot Pseudo instructions
--> In case of 512 Vector Load , the instr is broken in two 256 bit VLOAD and we can have a case where the lower offset is in range of Slot A & B and the higher offset is only in range of Slot A in such case we use MultiSlot Pseudo instructions for lower load and VLDA for higher load.


llvm/test/CodeGen/AIE/aie2/schedule/resource/memory_bank_vld_multi_slot.mir
```diff
@@ -0,0 +1,1083 @@
...
+#
+# This file is licensed under the Apache License v2.0 with LLVM Exceptions.
+# See https://llvm.org/LICENSE.txt for license information.
+# SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+#
+# (c) Copyright 2023-2024 Advanced Micro Devices, Inc. or its affiliates
+# RUN: llc -march=aie2 -run-pass=postmisched %s -o - | FileCheck %s
+
+# This test checks that we are scheduling mulislot loads from differnt banks in on VLIW bundle.
+
```
krishnamtibrewala, line 11: This test will pass only when #95 is merged

llvm/test/CodeGen/AIE/aie2/end-to-end/Conv2D-red.ll
```diff
@@ -130,17 +130,17 @@ define dso_local void @conv2d.loop.nest(ptr %add.ptr6.i51, ptr %add.ptr5, ptr %c
 ; ASM-NEXT:  .LBB0_2: // %inner.loop
 ; ASM-NEXT:    // Parent Loop BB0_1 Depth=1
 ; ASM-NEXT:    // => This Inner Loop Header: Depth=2
-; ASM-NEXT:    vlda wl6, [p0], m6; nopx
-; ASM-NEXT:    vlda wh6, [p0], m6
-; ASM-NEXT:    vlda wl8, [p0], m6
+; ASM-NEXT:    vldb wl6, [p0], m6; nopx
+; ASM-NEXT:    vldb wh6, [p0], m6
+; ASM-NEXT:    vldb wl8, [p0], m6
 ; ASM-NEXT:    vlda.3d wh8, [p0], d0
```
gbossu: Don't we also have 3D loads for `VLDB`?

llvm/lib/Target/AIE/AIE2MultiSlotPseudoInstrInfo.td
```diff
@@ -51,3 +53,25 @@ let isMoveImm = 1, isReMaterializable = 1, isAsCheapAsAMove = 1, Itinerary = II_
...
+      "vld_pstm_pseudo", "$dst, [$ptr], $mod", [VLDB_dmw_ldb_ag_pstm_nrm, VLDA_dmw_lda_w_ag_pstm_nrm]>;
+  def VLD_pstm_imm_4x32_pseudo : MultiSlot_Pseudo< (outs mWa:$dst, eP:$ptr_out) , (ins eP:$ptr, imm4x32:$imm),
+      "vld_pstm_imm_4x32_pseudo", "$dst, [$ptr], $imm", [VLDB_dmw_ldb_ag_pstm_nrm_imm, VLDA_dmw_lda_w_ag_pstm_nrm_imm]>;
+  }
+  let Itinerary = II_VLDA_2D_W in
+  let Constraints = "$ptr_out = $ptr" in {
+  def VLD_2D_pseudo : MultiSlot_Pseudo< (outs mWa:$dst, eP:$ptr_out, eDC:$count_out) , (ins eP:$ptr, eD:$mod),
+      "vld.2d_pseudo", "$dst, [$ptr], $mod", [VLDB_2D, VLDA_2D_dmw_lda_w]>;
+  }
+}
```
gbossu: Missing newline

llvm/lib/Target/AIE/AIE2MultiSlotPseudoInstrInfo.td
```diff
@@ -12,6 +12,8 @@
 // The last parameter to the MulitSlot_Pseudo is instruction into which the Multi-Slot Pseudo could be materialize to.
 // Priority to Slot/Instruction is based on the sequence in which the real instructions are passed
 
+// Note : Any of the Itinerary could be selected, but you need to make sure the operand latency is same for all the possible instr.
```
martien-de-jong: I think it's the other way. We can only use a multislot pseudo for instructions that have equal latencies.

llvm/lib/Target/AIE/AIE2MultiSlotPseudoInstrInfo.td
```diff
@@ -51,3 +53,25 @@ let isMoveImm = 1, isReMaterializable = 1, isAsCheapAsAMove = 1, Itinerary = II_
...
       "mov_PD_imm11_pseudo ", "$dst, $i", [MOVA_lda_cg, MOVXM_lng_cg]>;
 }
+
+// Pseudo VLD
+let hasSideEffects = false, mayLoad = true, mayStore = false in {
+  let Itinerary = II_VLDA_W in {
+  def VLD_idx_pseudo : MultiSlot_Pseudo< (outs mWa:$dst), (ins eP:$ptr, eDJ:$dj),
+      "vld_idx_pseudo", "$dst, [$ptr, $dj]", [VLDB_dmw_ldb_ag_idx, VLDA_dmw_lda_w_ag_idx]>;
+  def VLD_idx_imm_3x32_pseudo :MultiSlot_Pseudo< (outs mWa:$dst) , (ins eP:$ptr, imm3x32:$imm),
+      "vld_idx_imm_imm3x32_pseudo", "$dst, [$ptr, $imm]", [VLDB_dmw_ldb_ag_idx_imm, VLDA_dmw_lda_w_ag_idx_imm]>;
```
martien-de-jong: I would be happy if we also wrap .td files in 80 columns

llvm/lib/Target/AIE/AIE2MultiSlotPseudoInstrInfo.td
```diff
@@ -51,3 +53,25 @@ let isMoveImm = 1, isReMaterializable = 1, isAsCheapAsAMove = 1, Itinerary = II_
   def MOV_PD_imm11_pseudo : MultiSlot_Pseudo< (outs ePmDm:$dst) , (ins simm11:$i),
       "mov_PD_imm11_pseudo ", "$dst, $i", [MOVA_lda_cg, MOVXM_lng_cg]>;
 }
+
+// Pseudo VLD
+let hasSideEffects = false, mayLoad = true, mayStore = false in {
+  let Itinerary = II_VLDA_W in {
+  def VLD_idx_pseudo : MultiSlot_Pseudo< (outs mWa:$dst), (ins eP:$ptr, eDJ:$dj),
+      "vld_idx_pseudo", "$dst, [$ptr, $dj]", [VLDB_dmw_ldb_ag_idx, VLDA_dmw_lda_w_ag_idx]>;
+  def VLD_idx_imm_3x32_pseudo :MultiSlot_Pseudo< (outs mWa:$dst) , (ins eP:$ptr, imm3x32:$imm),
```
martien-de-jong: nit: spacing ` : ` and `$dst), `

llvm/lib/Target/AIE/AIE2InstructionSelector.cpp
```diff
@@ -3570,10 +3570,29 @@ LoadStoreOpcodes AIE2InstructionSelector::getLoadStoreOpcode(
...
+        } else if (checkImmediateRange<8, 32>(Offset)) {
+          // When Offset is positive and one of the offset is in range of SlotB
+          ISelOpcode = AIE2::VLD_idx_imm_3x32_pseudo;
+          OffsetOpcode = AIE2::VLDA_dmw_lda_w_ag_idx_imm;
+          FitsImmediateRange = true;
+        } else if (Offset.has_value() && (*Offset).isNegative() &&
+                   checkImmediateRange<8, 32>((*Offset) + 32)) {
+          // When Offset is negative and one of the offset is in range of SlotB
+          ISelOpcode = AIE2::VLDA_dmw_lda_w_ag_idx_imm;
+          OffsetOpcode = AIE2::VLD_idx_imm_3x32_pseudo;
```
martien-de-jong, line 3576: I think 3x32 is describing the immediate, where 3 is a number of bits and  32 is a value range. The offset check above is checking 8 and 32, which seems to use range in both cases.
The 'x' connector is appropriate for the range notation; number of bits would ask for '+'.  Could we use imm8x32, to group this name element and to show the connection to the range check?

llvm/lib/Target/AIE/AIE2InstructionSelector.cpp
```diff
@@ -3278,12 +3278,12 @@ LoadStoreOpcodes AIE2InstructionSelector::getLoadStoreOpcode(
       unsigned RBID = deriveRegBankID(I.getOperand(0).getReg(), MRI, RBI);
       if (RBID == AIE2::AccRegBankID) {
         return {/*ISelOpcode=*/AIE2::VST_dmw_sts_am_ag_idx_imm,
-                /*FitsImmediateRange=*/true,
+                AlwaysFitsImmediateRange,
```
martien-de-jong, line 3269: This is an existing constant?

llvm/lib/Target/AIE/AIE2MultiSlotPseudoInstrInfo.td
krishnamtibrewala: Correct me if I am wrong my interpretation of `same operand latency` is same as `equal latencies`

llvm/lib/Target/AIE/AIE2InstructionSelector.cpp
krishnamtibrewala: Yes.

llvm/lib/Target/AIE/AIE2InstructionSelector.cpp
krishnamtibrewala: 3x32 bit here is total 8 bits 
3 + 5 bits, where the 5 bits are always 0 does the offset can be in steps of 32 only

In the multi slot pseudo name I have tried to align that with the input operand type imm3x32, which was does for previously defined  multi slot pseudo

llvm/lib/Target/AIE/AIE2InstrInfo.td
```diff
@@ -584,8 +585,8 @@ class Split2DInstr<Instruction RealInst, int opidx> : SplitPseudo<RealInst,
 foreach instr = [VLDA_2D_dmw_lda_w, VLDA_2D_dmw_lda_am, VLDA_2D_CONV_FP32_BF16,
                  VLDB_2D, VLDB_2D_128, LDA_2D_dmv_lda_q, VLDB_2D_UNPACK_S8_S4,
                  VLDB_2D_UNPACK_S16_S8, VLDB_2D_UNPACK_D8_D4, VLDB_2D_UNPACK_D16_D8,
-                 VST_2D_dmw_sts_w, VST_2D_dmw_sts_am, VST_2D_128, ST_2D_dmv_sts_q,
-                 LDA_2D_dms_lda, LDA_2D_S8_dmhb_lda, LDA_2D_U8_dmhb_lda,
+                 VLD_2D_pseudo, VST_2D_dmw_sts_w, VST_2D_dmw_sts_am, VST_2D_128,
```
krishnamtibrewala, line 588: Note : This is critical for the `aie-superreg-rewrite` to work as expected. This helps in avoiding spills of the `D` type regs.

llvm/lib/Target/AIE/AIE2MultiSlotPseudoInstrInfo.td
martien-de-jong: I'm not denying that. I would just state that the itineraries of all instructions covered by a multi-slot pseudo should be equivalent. It's not part of selecting the itinerary, it's part of defining the MSPseudo.

llvm/lib/Target/AIE/AIE2InstructionSelector.cpp
martien-de-jong: If you explain it as 3 + 5 bits, I think you agree that 3x32 is wrong?

martien-de-jong (dismissed):

llvm/lib/Target/AIE/AIE2InstrPatterns.td
```diff
@@ -242,11 +242,8 @@ def : PartWordStPat<truncstorei16, ST_S16_ag_idx_imm, Ty, eR>;
 }
 
 // 3.2 VLD - Vector Load
-class VectorLdPat<ValueType ResType, AIE2Inst Inst = VLDA_dmw_lda_w_ag_idx_imm>
-    : Pat<(ResType (load eP:$rs1)), (Inst eP:$rs1, (i20 0))>;
```
gbossu, line 246: That was unused?

llvm/lib/Target/AIE/AIE2InstructionSelector.cpp
gbossu: I think `checkImmediateRange<8, 32>` is correct based on what that function does, because `8 = 3 + ln2(32)`.

In a follow-up PR, we could change `bool checkImmediateRange<unsigned MaxPow2, unsigned Step>` into something like `bool checkImmediateRange<unsigned NumEncodingBits, unsigned Step>`, and `MaxPow2` can be derived from both `NumEncodingBits` and `Step` That would make the code a whole lot easier to read, because `checkImmediateRange<NumEncodingBits=3, Step=32>` will nicely map to `3x32`.

llvm/lib/Target/AIE/AIE2MultiSlotPseudoInstrInfo.td
```diff
@@ -31,23 +37,68 @@ let isMoveImm = 1, isReMaterializable = 1, isAsCheapAsAMove = 1, Itinerary = II_
     hasSideEffects = false, mayLoad = false, mayStore = false in {
 
   // To move 10bit imm to R/lc type reg.
-  def MOV_RLC_imm10_pseudo : MultiSlot_Pseudo< (outs eRLC:$dst) , (ins simm10:$i),
-      "mov_rlc_imm10_pseudo ", "$dst, $i", [MOVA_lda_cg, MOVX_alu_cg, MOV_mv_cg, MOVXM_lng_cg]>;
+  def MOV_RLC_imm10_pseudo : MultiSlot_Pseudo< (outs eRLC:$dst), (ins simm10:$i),
+      "mov_rlc_imm10_pseudo ", "$dst, $i",
+      [MOVA_lda_cg, MOVX_alu_cg, MOV_mv_cg, MOVXM_lng_cg]>;
```
gbossu, line 42: Nit: Please avoid committing unnecessary diff lines.

llvm/lib/Target/AIE/AIE2MultiSlotPseudoInstrInfo.td
```diff
@@ -31,23 +37,68 @@ let isMoveImm = 1, isReMaterializable = 1, isAsCheapAsAMove = 1, Itinerary = II_
...
+  def VLD_2D_pseudo : MultiSlot_Pseudo< (outs mWa:$dst, eP:$ptr_out, eDC:$count_out), 
+                                        (ins eP:$ptr, eD:$mod),
+      "vld.2d_pseudo", "$dst, [$ptr], $mod", 
+      [VLDB_2D, VLDA_2D_dmw_lda_w]>;
+  }
+  let Itinerary = II_VLDA_3D_W in
+  let Constraints = "$ptr_out = $ptr" in {
+  def VLD_3D_pseudo : MultiSlot_Pseudo< (outs mWa:$dst, eP:$ptr_out, eDC:$count_lo_out, eDC:$count_hi_out), 
+                                        (ins eP:$ptr, eDS:$mod),
+      "vlda.3d_pseudo", "$dst, [$ptr], $mod",
```
gbossu: `vld.3d`, not `vlda`

llvm/lib/Target/AIE/AIEHazardRecognizer.cpp
```diff
@@ -121,6 +121,9 @@ bool AIEResourceCycle::canReserveResources(MachineInstr &MI) {
   if (!AlternateInsts)
     return Bundle.canAdd(&MI);
 
+  if (MI.mayLoad())
```
gbossu, line 138: Can you add a comment to explain that change? So far we want the change for VLD pseudos to be NFC for the SW pipeliner

llvm/lib/Target/AIE/AIE2InstrPatterns.td
krishnamtibrewala: yes, there was no pattern using `VectorLdPat`

llvm/lib/Target/AIE/AIE2MultiSlotPseudoInstrInfo.td
krishnamtibrewala: Sure, was suggested to have lines within  80 char. So fixed the whole file.

llvm/lib/Target/AIE/AIE2InstructionSelector.cpp
krishnamtibrewala: @gbossu have update the function as suggested above as a separate NFC comment.

llvm/lib/Target/AIE/AIE2InstructionSelector.cpp
```diff
@@ -3557,10 +3557,29 @@ LoadStoreOpcodes AIE2InstructionSelector::getLoadStoreOpcode(
...
-                FitsImmediateRange,
-                /*OffsetOpcode=*/AIE2::VLDA_dmw_lda_w_ag_idx_imm};
+        unsigned OffsetOpcode;
+        // First try if the Instruction can be selected as multi-slot offset
+        // load
+        if (checkImmediateRangeSplitting<8, 32, 32>(Offset)) {
+          FitsImmediateRange = true;
+          ISelOpcode = OffsetOpcode = AIE2::VLD_idx_imm_3x32_pseudo;
+        } else if (checkImmediateRange<8, 32>(Offset)) {
+          // When Offset is positive and one of the offset is in range of SlotB
```
gbossu, line 3568: Nit: I found the comments about positive/negative a bit confusing. I think we are just trying to find out which of `Offset` or `Offset+32` fits for VLDB. And we then assume the one that does not fit will fit for VLDA. Why do we need to check below for `(*Offset).isNegative()`?

llvm/lib/Target/AIE/AIE2InstructionSelector.cpp
```diff
@@ -3557,10 +3557,29 @@ LoadStoreOpcodes AIE2InstructionSelector::getLoadStoreOpcode(
...
+          OffsetOpcode = AIE2::VLDA_dmw_lda_w_ag_idx_imm;
+          FitsImmediateRange = true;
+        } else if (Offset.has_value() && (*Offset).isNegative() &&
+                   checkImmediateRange<8, 32>((*Offset) + 32)) {
+          // When Offset is negative and one of the offset is in range of SlotB
+          ISelOpcode = AIE2::VLDA_dmw_lda_w_ag_idx_imm;
+          OffsetOpcode = AIE2::VLD_idx_imm_3x32_pseudo;
+          FitsImmediateRange = true;
+        } else {
+          FitsImmediateRange = checkImmediateRangeSplitting<11, 32, 32>(Offset);
```
gbossu: Nit: add a comment like `neither Offset nor Offset+32 fit for VLDB, try both for VLDA`

llvm/test/CodeGen/AIE/aie2/GlobalISel/inst-select-indexed-load-store.mir
```diff
@@ -179,6 +259,116 @@ body:             |
...
+    ; CHECK-NEXT: [[REG_SEQUENCE:%[0-9]+]]:vec512 = REG_SEQUENCE [[VLDA_dmw_lda_w_ag_idx_imm1]], %subreg.sub_256_lo, [[VLDA_dmw_lda_w_ag_idx_imm]], %subreg.sub_256_hi
+    ; CHECK-NEXT: $x0 = COPY [[REG_SEQUENCE]]
+    %1:ptrregbank(p0) = COPY $p0
+    %2:modregbank(s20) = G_CONSTANT i20 128
+    %0:vregbank(<16 x s32>) = G_AIE_OFFSET_LOAD %1:ptrregbank(p0), %2:modregbank(s20) :: (load (<16 x s32>))
+    $x0 = COPY %0:vregbank(<16 x s32>)
+...
+
+---
+name:            VEC512_LOAD_overMinOffset_slotB
```
gbossu: Nit: overMinOffset confuses me, maybe BelowMinOffset?

llvm/test/CodeGen/AIE/aie2/GlobalISel/inst-select-pre-post-increment.mir
```diff
@@ -23,13 +23,13 @@ body: |
...
+    ; CHECK-NEXT: [[VLD_pstm_pseudo:%[0-9]+]]:vec256, [[VLD_pstm_pseudo1:%[0-9]+]]:ep = VLD_pstm_pseudo [[COPY]], [[COPY2]] :: (load (<16 x s16>))
+    ; CHECK-NEXT: [[VLD_pstm_imm_4x32_pseudo:%[0-9]+]]:vec256, [[VLD_pstm_imm_4x32_pseudo1:%[0-9]+]]:ep = VLD_pstm_imm_4x32_pseudo [[VLD_pstm_pseudo1]], 0 :: (load (<16 x s16>))
+    ; CHECK-NEXT: [[VLD_pstm_pseudo2:%[0-9]+]]:vec256, [[VLD_pstm_pseudo3:%[0-9]+]]:ep = VLD_pstm_pseudo [[VLD_pstm_imm_4x32_pseudo1]], [[MOV_PD_imm10_pseudo]] :: (load (<16 x s16>))
+    ; CHECK-NEXT: [[VLDA_dmw_lda_w_ag_pstm_nrm_imm:%[0-9]+]]:vec256, [[VLDA_dmw_lda_w_ag_pstm_nrm_imm1:%[0-9]+]]:ep = VLDA_dmw_lda_w_ag_pstm_nrm_imm [[VLD_pstm_pseudo3]], -2048 :: (load (<16 x s16>))
+    ; CHECK-NEXT: [[VLDA_dmw_lda_w_ag_pstm_nrm_imm2:%[0-9]+]]:vec256, [[VLDA_dmw_lda_w_ag_pstm_nrm_imm3:%[0-9]+]]:ep = VLDA_dmw_lda_w_ag_pstm_nrm_imm [[VLDA_dmw_lda_w_ag_pstm_nrm_imm1]], 2016 :: (load (<16 x s16>))
+    ; CHECK-NEXT: [[VLD_pstm_pseudo4:%[0-9]+]]:vec256, [[VLD_pstm_pseudo5:%[0-9]+]]:ep = VLD_pstm_pseudo [[VLDA_dmw_lda_w_ag_pstm_nrm_imm3]], [[MOVXM_lng_cg]] :: (load (<16 x s16>))
+    ; CHECK-NEXT: PseudoRET implicit $lr, implicit [[VLD_pstm_pseudo4]]
```
gbossu, line 32: Can you also extend the test to show the limits of VLDB.postinc?

gbossu (dismissed):
LGTM, before merging, I would just like you to add some tests fro the ISel of `VLD_pstm_pseudo`. I don't think we properly check its immediate range.

llvm/lib/Target/AIE/AIE2InstructionSelector.cpp
krishnamtibrewala: When we are loading 512 bit , we convert that in two 256 bit load i.e from `offset & offset+32`

When offset is negative we will come to that check only when Offset is **negative** and over the min range , but if number is negative then `Offset + 32` might be in range for Multi-Slot Pseudo. To be on safer side I added a check of `isNegative` and I felt the code was more reliable.

llvm/test/CodeGen/AIE/aie2/GlobalISel/inst-select-pre-post-increment.mir
krishnamtibrewala: Have added test to the show the limits of `VLD_pstm_pseudo`

andcarminati (approved):
LGTM. Thank you for accurately considering all the comments.

PR #115, opened 2024-07-08
[AIE2] Fix SIGSEGV for VLDA_CONV.

SagarMaheshwari99:
We should not try and combine VCONV with anything but G_LOAD or the like.

`VCONV` instruction expects a `VLDA` to combine to `VLDA.CONV`. Currently, we only check the `mayLoad` attribute for any instruction to classify it as load and combine to `VLDA.CONV`, which is not right. We should check the correct opcodes that lead to `VLDA`. 
This PR resolves that and prohibits, for ex., `G_INTRINSIC_W_SIDE_EFFECTS intrinsic(@llvm.aie2.v16accfloat.to.v16bf16)`, which `mayLoad`, from combining to `VCONV.fp32.bf16` (and avoid a segfault from accessing non-existent MMOs).

https://jira.xilinx.com/browse/AIECC-676

### DIFF
--- a/llvm/lib/Target/AIE/AIE2InstructionSelector.cpp
+++ b/llvm/lib/Target/AIE/AIE2InstructionSelector.cpp
@@ -4427,51 +4427,57 @@ bool AIE2InstructionSelector::select512BitG_AIE_LOAD_STORE(
   }
 }
 
+static bool getVLDA_CONVOpcode(const MachineInstr &MemOp,
+                               const std::optional<APInt> Immediate,
+                               unsigned &ISelOpcode, bool &FitsImmediateRange) {
+  if (!MemOp.mayLoad())
+    return false;
+
+  switch (MemOp.getOpcode()) {
+  case AIE2::G_LOAD:
+    ISelOpcode = AIE2::VLDA_CONV_FP32_BF16_ag_idx_imm;
+    FitsImmediateRange = true;
+    return true;
+  case AIE2::G_AIE_OFFSET_LOAD:
+    FitsImmediateRange = checkImmediateRange<8, 32>(Immediate);
+    ISelOpcode = FitsImmediateRange ? AIE2::VLDA_CONV_FP32_BF16_ag_idx_imm
+                                    : AIE2::VLDA_CONV_FP32_BF16_ag_idx;
+    return true;
+  case AIE2::G_AIE_POSTINC_LOAD:
+    FitsImmediateRange = checkImmediateRange<9, 32>(Immediate);
+    ISelOpcode = FitsImmediateRange ? AIE2::VLDA_CONV_FP32_BF16_pstm_nrm_imm
+                                    : AIE2::VLDA_CONV_FP32_BF16_pstm_nrm;
+    return true;
+  case AIE2::G_AIE_POSTINC_2D_LOAD:
+    ISelOpcode = AIE2::VLDA_2D_CONV_FP32_BF16;
+    FitsImmediateRange = false;
+    return true;
+  case AIE2::G_AIE_POSTINC_3D_LOAD:
+    ISelOpcode = AIE2::VLDA_3D_CONV_FP32_BF16;
+    FitsImmediateRange = false;
+    return true;
+  }
+  return false;
+}
+
 std::optional<LoadStoreOpcodes>
 getCombinedOpcodeCONVLoad(const MachineInstr &MemOp, const MachineInstr &CombOp,
-                          std::optional<APInt> Immediate) {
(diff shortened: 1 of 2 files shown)


gbossu:
Could you give a bit more context in the PR description? Why is it that we cannot combine?

SagarMaheshwari99:
> Could you give a bit more context in the PR description? Why is it that we cannot combine?

Done.. 

`VCONV` instruction expects a `VLDA` to combine to `VLDA.CONV`. Currently, we only check the `mayLoad` attribute for any instruction to classify it as load and combine to `VLDA.CONV`, which is not right. We should check the correct opcodes that lead to `VLDA`. 
This PR resolves that and prohibits, for ex., `G_INTRINSIC_W_SIDE_EFFECTS intrinsic(@llvm.aie2.v16accfloat.to.v16bf16)`, which `mayLoad`, from combining to `VCONV.fp32.bf16` (and avoid a segfault from accessing non-existent MMOs).

llvm/test/CodeGen/AIE/aie2/GlobalISel/inst-select-vlda_conv.mir
```diff
@@ -315,3 +315,26 @@ body: |
...
+  bb.0:
+    liveins: $wl0, $bml0
+    ; CHECK-LABEL: name: no_combine
+    ; CHECK: liveins: $wl0, $bml0
+    ; CHECK-NEXT: {{  $}}
+    ; CHECK-NEXT: [[COPY:%[0-9]+]]:acc512 = COPY $bml0
+    ; CHECK-NEXT: [[VCONV_BF16_FP32_:%[0-9]+]]:vec256 = VCONV_BF16_FP32 [[COPY]], implicit-def dead $srf2fflags, implicit $crf2fmask, implicit $crrnd
+    ; CHECK-NEXT: [[VCONV_FP32_BF16_:%[0-9]+]]:acc512 = VCONV_FP32_BF16 [[VCONV_BF16_FP32_]]
+    ; CHECK-NEXT: PseudoRET implicit $lr, implicit [[VCONV_FP32_BF16_]]
+    %0:vregbank(<16 x s16>) = COPY $wl0
```
andcarminati: nit: this first copy dead.

llvm/lib/Target/AIE/AIE2InstructionSelector.cpp
```diff
@@ -4427,51 +4427,57 @@ bool AIE2InstructionSelector::select512BitG_AIE_LOAD_STORE(
...
-          Intrinsic::aie2_v16bf16_to_v16accfloat)
-    return {};
-
+bool getVLDA_CONVOpcode(const MachineInstr &MemOp,
+                        const std::optional<APInt> Immediate,
+                        unsigned &ISelOpcode, bool &FitsImmediateRange) {
   if (!MemOp.mayLoad())
-    return {};
-
-  assert(getLoadStoreSize(MemOp) == 256 && "Unexpected VLDA.CONV size");
```
andcarminati, line 4444: As I understood, this assert is the origin of the problem. If we move it outside this logic everything goes well. `getLoadStoreSize` expects LD/ST opcodes, anything else will crash. Your PR basically moved this assert to a safe place. Looks good to me.

SagarMaheshwari99: Right..

llvm/test/CodeGen/AIE/aie2/GlobalISel/inst-select-vlda_conv.mir
SagarMaheshwari99: ah yes..

llvm/lib/Target/AIE/AIE2InstructionSelector.cpp
```diff
@@ -4427,51 +4427,57 @@ bool AIE2InstructionSelector::select512BitG_AIE_LOAD_STORE(
...
-                          std::optional<APInt> Immediate) {
-
-  // const bool AlwaysFitsImmediateRange = true;
-  const bool NoImmediate = false;
-  if (CombOp.getOpcode() != AIE2::G_INTRINSIC ||
-      cast<GIntrinsic>(CombOp).getIntrinsicID() !=
-          Intrinsic::aie2_v16bf16_to_v16accfloat)
-    return {};
-
+bool getVLDA_CONVOpcode(const MachineInstr &MemOp,
```
konstantinschwarz: nit: make static?

konstantinschwarz (dismissed):

konstantinschwarz (approved):

PR #119, opened 2024-07-09
[AIEX] Fix CR register assignments simplification pass

gbossu:
The first commit adds two examples where we are too optimistic and simplify too much. This is then fixed in the next two commits.

I rewrote the pass to rely on a more standard "available value" analysis. This introduces a copy tracker and is a bit similar to the `MachineCopyPropagation` pass, but simplified because we have a different use case. Mainly, we have a mix of vregs and phys regs, and the vregs are in ssa form.

I'd be happy with reviews from @martien-de-jong @andcarminati @abhinay-anubola 

FYI @konstantinschwarz I think you also noticed some bugs here.

### DIFF
--- a/llvm/lib/Target/AIE/AIEPostSelectOptimize.cpp
+++ b/llvm/lib/Target/AIE/AIEPostSelectOptimize.cpp
@@ -57,6 +57,61 @@ static cl::opt<bool>
                              cl::desc("Enable post select optimize."));
 
 namespace {
+
+/// Information about a COPY that can be tracked by PhysRegCopyTracker.
+struct TrackableCopyOperands {
+  Register VirtReg;
+  MCRegister PhysReg;
+  bool IsPhysRegDef; // Whether PhysReg is being redefined.
+};
+
+/// Track SSA copies of simple physical registers.
+/// Simple means they don't have sub or super registers.
+class PhysRegCopyTracker {
+  DenseMap<MCRegister, SmallSet<Register, 4>> Copies;
+  const TargetRegisterInfo &TRI;
+
+public:
+  PhysRegCopyTracker(const MachineRegisterInfo &MRI)
+      : TRI(*MRI.getTargetRegisterInfo()) {
+    assert(MRI.isSSA() && "PhysRegCopyTracker can only track SSA copies");
+  }
+
+  /// Invalidate the copies of \p Reg.
+  void invalidateCopies(MCRegister Reg) { Copies.erase(Reg); }
+
+  /// Invalidate copies for any reg defined by \p MI
+  void invalidateDefCopies(const MachineInstr &MI) {
+    for (const MachineOperand &MO : MI.all_defs()) {
+      if (MO.getReg().isPhysical()) {
+        invalidateCopies(MO.getReg());
+      }
+    }
+  }
+
+  /// Track that \p VRegCopy is a copy of \p PhysReg
+  void trackCopy(MCRegister PhysReg, Register VRegCopy) {
+    assert(VRegCopy.isVirtual());
+    assert(range_size(TRI.regunits(PhysReg)) == 1 && "Phys reg has aliases.");
+    Copies[PhysReg].insert(VRegCopy);
+  }
+
+  /// Return true if \p VReg is a copy of \p PhysReg
+  bool isCopy(MCRegister PhysReg, Register VReg) const {
+    if (auto It = Copies.find(PhysReg); It != Copies.end())
+      return It->second.contains(VReg);
+    return false;
+  }
+
+  /// Returns true if \p MI is a COPY instruction that can be eliminated.
+  bool isRedundantCopy(const MachineInstr &MI) {
+    if (!MI.isCopy())
+      return false;
+    auto [DstReg, SrcReg] = MI.getFirst2Regs();
+    return DstReg.isPhysical() && isCopy(DstReg, SrcReg);
+  }
+};
+
 class AIEPostSelectOptimize : public MachineFunctionPass {
 public:
   static char ID;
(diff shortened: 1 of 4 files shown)


llvm/test/CodeGen/AIE/aie2/dyn-stackalloc.ll
```diff
@@ -53,27 +53,29 @@ define void @test_loop_dyn_alloca(i32 noundef %n) {
 ; CHECK-LABEL: test_loop_dyn_alloca:
 ; CHECK:         .p2align 4
 ; CHECK-NEXT:  // %bb.0: // %entry
-; CHECK-NEXT:    paddb [sp], #32; nopa ; nops ; nopxm ; nopv
```
gbossu, line 56: Somehow we used to simplify copies to/from `SP`, I'd rather not touch it for now.

llvm/lib/Target/AIE/AIEPostSelectOptimize.cpp
```diff
@@ -57,197 +57,139 @@ static cl::opt<bool>
...
+  Register VirtReg;
+  MCRegister PhysReg;
+  bool InvalidatePhysReg; // Whether PhysReg is being redefined.
+};
 
-  /// Checks if the \p Op is an unallocatable phys reg.
-  bool isUnallocatablePhysReg(const MachineOperand &Op);
+/// Track SSA copies of simple physical registers.
+/// Simple means they don't have sub or super registers.
+class PhysRegCopyTracker {
```
andcarminati, line 70: I think this is more than a Tracker.

llvm/lib/Target/AIE/AIEPostSelectOptimize.cpp
```diff
@@ -57,197 +57,139 @@ static cl::opt<bool>
...
+  bool trySimplifyCopy(MachineInstr &MI) {
+    if (!MI.isCopy())
+      return false;
+    auto [DstReg, SrcReg] = MI.getFirst2Regs();
+    if (DstReg.isPhysical() && isCopy(DstReg, SrcReg)) {
+      // Assigning to DstReg a copy of itself. Just remove MI.
+      MI.eraseFromParent();
+      return true;
     }
+    if (SrcReg.isPhysical() && hasVirtualCopy(SrcReg)) {
```
andcarminati: Hi @gbossu. Do you think that this part is responsible for the additional spilling? I mean, we are extending the usage of a virt reg here.

gbossu: I don't think so, the extra vreg would get coalesced anyway before RA. It's just that I want to get rid of it early to allow future passes like our custom LICM to have a simple MIR to deal with.

llvm/lib/Target/AIE/AIEPostSelectOptimize.cpp
gbossu: Can't argue :D I can maybe make `trySimplifyCopy` a free function, then it would be just a tracker. What do you think?

andcarminati: > Can't argue :D I can maybe make `trySimplifyCopy` a free function, then it would be just a tracker. What do you think?

I think it would be nice!

llvm/lib/Target/AIE/AIEPostSelectOptimize.cpp
```diff
@@ -57,197 +57,139 @@ static cl::opt<bool>
...
-  void storeFirstIdentityCopy(MachineInstr &MI);
+  /// Invalidate the copies of \p Reg.
+  void invalidateCopies(MCRegister Reg) { Copies.erase(Reg); }
 
-  /// Erase \param MI if it is similar to \param LastVisitedDef.
-  bool eraseSimilarClobber(MachineInstr &MI, MachineInstr &LastVisitedDef);
-};
-} // end anonymous namespace
+  /// Invalidate copies for any reg defined by \p MI
+  void invalidateDefCopies(const MachineInstr &MI) {
```
andcarminati, line 84: Here you can safely use (all_defs guarantee a Register for us):
```
  /// Invalidate copies for any reg defined by \p MI
  void invalidateDefCopies(const MachineInstr &MI) {
    for (const MachineOperand &MO : MI.all_defs()) {
      auto Reg = MO.getReg();
      if (Reg.isPhysical()) {
        invalidateCopies(MO.getReg());
      }
    }
  }
```

gbossu:
Note: Out of 300+ benchmarks, I see 4 regressions (2 of which are 10% extra cycles), but those are due to our previous optimizations being way too optimistic/incorrect.

andcarminati (dismissed):
LGTM.

abhinay-anubola (dismissed):
LGTM!!

llvm/lib/Target/AIE/AIEPostSelectOptimize.cpp
```diff
@@ -57,197 +57,121 @@ static cl::opt<bool>
...
-/// If there is an intervening clobber but copy propagation suggests it
-/// comes from the same PhysReg, it can be removed.
-bool AIEPostSelectOptimize::eraseSimilarClobber(MachineInstr &MI,
-                                                MachineInstr &LastVisitedDef) {
-  if (!isCopyFromPhysReg(LastVisitedDef, MI.getOperand(0).getReg()) ||
-      !isCopyFromPhysReg(MI, LastVisitedDef.getOperand(0).getReg()))
+  /// Return true if \p VReg is a copy of \p PhysReg
+  bool isCopy(MCRegister PhysReg, Register VReg) const {
+    if (auto It = Copies.find(PhysReg); It != Copies.end())
+      return It->second.contains(VReg);
```
martien-de-jong, line 102: nit: These two interfaces record and check VReg = COPY(PhysReg); I would swap the parameters

llvm/lib/Target/AIE/AIEPostSelectOptimize.cpp
```diff
@@ -57,197 +57,121 @@ static cl::opt<bool>
...
-  /// PhysReg as the destination register of the \p MI.
-  bool isCopyFromPhysReg(const MachineInstr &MI, const Register PhysReg);
-
-  /// Store the first %virtreg to $physreg copy corresponding to \param MI.
-  void storeFirstIdentityCopy(MachineInstr &MI);
-
-  /// Erase \param MI if it is similar to \param LastVisitedDef.
-  bool eraseSimilarClobber(MachineInstr &MI, MachineInstr &LastVisitedDef);
+/// Information about a COPY that can be tracked by PhysRegCopyTracker.
+struct TrackableCopyOps {
```
martien-de-jong: This looks plural

llvm/lib/Target/AIE/AIEPostSelectOptimize.cpp
```diff
@@ -57,197 +57,121 @@ static cl::opt<bool>
...
+  Register VirtReg;
+  MCRegister PhysReg;
+  bool InvalidatePhysReg; // Whether PhysReg is being redefined.
```
martien-de-jong: The comment doesn't really explain. 'is being' sounds transient. If so, describe its scope of validity

martien-de-jong: Ok. After reading the rest: it is transient, and there's no need to store it in the map. Basically, both Phys = Virt and Virt = Phys set up an equivalence between the two which holds until it is killed. I think it would read if you would recognise both cases by their own predicates and take the appropriate actions in the main traversal. In that way, you don't need this flag at all.

llvm/test/CodeGen/AIE/aie2/GlobalISel/legalize-dyn-stackalloc.ll
```diff
@@ -92,23 +94,24 @@ define void @test_loop_dyn_alloca(i32 noundef %n) {
 ; CHECK-NEXT:    nop // Delay Slot 4
 ; CHECK-NEXT:    nop // Delay Slot 3
 ; CHECK-NEXT:    nop // Delay Slot 2
-; CHECK-NEXT:    nop // Delay Slot 1
+; CHECK-NEXT:    mov sp, p6 // Delay Slot 1
```
martien-de-jong, line 97: interesting. That was pretty wrong, was it not?

llvm/lib/Target/AIE/AIEPostSelectOptimize.cpp
```diff
@@ -147,6 +147,26 @@ getTrackableCopyOps(const MachineInstr &MI, const TargetRegisterInfo &TRI) {
...
+    // Assigning to DstReg a copy of itself. Just remove MI.
+    MI.eraseFromParent();
+    return true;
+  }
+  if (std::optional<Register> EquivVirtSrc;
+      SrcReg.isPhysical() && (EquivVirtSrc = CT.getVirtualCopy(SrcReg))) {
+    // Avoid defining a new VReg if another available one has the same value.
+    MI.getMF()->getRegInfo().replaceRegWith(DstReg, *EquivVirtSrc);
+    MI.eraseFromParent();
+    return true;
```
martien-de-jong, line 166: Can we now guarantee we never have more than one VirtReg equivalent to a particular PhysReg?

martien-de-jong: Could we be mixing reg classes here and get in trouble similar to the grief we've seen in the singleton regclass for SELECT? (Not now, but in the (distant) future)

llvm/lib/Target/AIE/AIEPostSelectOptimize.cpp
gbossu: My point is to gather information about a COPY instruction in a single place, then it's easy to track in in the CopyTracker. I didn't want to overload the traversal and keep it at three cases:
 - Either it can be simplified
 - Or we can track the operands being copied
 - Or we can't and need to invalidate copies for any re-defined registers

Do you feel I should expose more there to make it easier to udnerstand?

llvm/lib/Target/AIE/AIEPostSelectOptimize.cpp
gbossu: I don't think so, the phys registers we look at here are not allocatable, only the copies are, but they always are in classes that are allocatable, and never directly used by instructions.

gbossu: > Can we now guarantee we never have more than one VirtReg equivalent to a particular PhysReg

I think we can actually, I just wanted to keep the tracker generic if we later extend to an interblock analysis.

llvm/lib/Target/AIE/AIEPostSelectOptimize.cpp
gbossu: It is, because it holds two operands. Do you mean I should make it singular?

llvm/test/CodeGen/AIE/aie2/GlobalISel/legalize-dyn-stackalloc.ll
gbossu: Looks like it yes! Then again, I'm not sure I understand this test fully.

llvm/lib/Target/AIE/AIEPostSelectOptimize.cpp
gbossu: Discussed offline, I renamed `InvalidatePhysReg` to `IsPhysRegDef`

llvm/lib/Target/AIE/AIEPostSelectOptimize.cpp
martien-de-jong: You mean it's CopyOperands, not CopyOperations?

gbossu: I'll rename that as well :D

andcarminati (approved):

PR #120, opened 2024-07-09
Add missing float intrinsics

konstantinschwarz:


clang/lib/Headers/aiev2_scl2vec.h
```diff
@@ -51,11 +51,10 @@ INTRINSIC(v16accfloat)
 shiftx(v16accfloat a, v16accfloat b, int step, int shift) {
   return __builtin_aiev2_vshift_I512_I512(a, b, step, shift);
 }
-#if 0
```
gbossu, line 54: Curious: What was blocking the definition of those intrinsics before?

clang/test/CodeGen/aie/aie2/aie2-upd-ext-intrinsic.cpp
```diff
@@ -1835,3 +1836,139 @@ v8bfloat16 test_extract_v8bfloat16_512(v32bfloat16 a, int idx) {
 v8bfloat16 test_extract_v8bfloat16_256(v16bfloat16 a, int idx) {
    return extract_v8bfloat16(a,idx);
 }
+
+// CHECK-LABEL: @_Z16test_set_v8floatiDv4_f(
```
gbossu, line 1839: Nit: Would be nice to only push the diff for those new tests. THe lines changed above are just artifacts of the test updater script.

clang/test/CodeGen/aie/aie2/aiev2-vadd-intrinsics.cpp
```diff
@@ -296,11 +296,11 @@ v64uint8 test_bneg_v64uint8(v64uint8 a) { return bneg(a); }
 // CHECK-NEXT:  entry:
 // CHECK-NEXT:    [[TMP0:%.*]] = tail call { <64 x i8>, <2 x i32> } @llvm.aie2.vbneg.ltz8(<64 x i8> [[B:%.*]])
 // CHECK-NEXT:    [[TMP1:%.*]] = extractvalue { <64 x i8>, <2 x i32> } [[TMP0]], 0
-// CHECK-NEXT:    [[AND_I7_I:%.*]] = and <64 x i8> [[TMP1]], [[A:%.*]]
+// CHECK-NEXT:    [[AND_I_I:%.*]] = and <64 x i8> [[TMP1]], [[A:%.*]]
 // CHECK-NEXT:    [[TMP2:%.*]] = tail call { <64 x i8>, <2 x i32> } @llvm.aie2.vbneg.ltz8(<64 x i8> [[A]])
 // CHECK-NEXT:    [[TMP3:%.*]] = extractvalue { <64 x i8>, <2 x i32> } [[TMP2]], 0
-// CHECK-NEXT:    [[AND_I_I:%.*]] = and <64 x i8> [[TMP3]], [[B]]
-// CHECK-NEXT:    [[OR_I_I:%.*]] = or <64 x i8> [[AND_I_I]], [[AND_I7_I]]
```
gbossu: Nit: Same here, would be nice not to push those noisy diffs

clang/lib/Headers/aiev2_vadd.h
```diff
@@ -1291,6 +1291,77 @@ INTRINSIC(v16int32) bfloat16_to_int(v16bfloat16 a, int shft) {
   return __builtin_aiev2_v16bf16_to_v16i32(a, shft);
 }
 
+INTRINSIC(unsigned int) lt(v16float v1, v16float v2) {
+  v32bfloat16 a = broadcast_zero_to_v32bfloat16();
+  v32bfloat16 b = broadcast_zero_to_v32bfloat16();
+  a = insert(a, 0, to_v16bfloat16(sub((v16accfloat)v1, (v16accfloat)v2)));
+  return lt(a, b);
```
gbossu, line 1298: That must generate lots of instructions, right?
If there a reason we don't do `return lt(to_v16bfloat16(v1), to_v16bfloat16(v2)`? Is that because of precision?

gbossu (dismissed):

clang/lib/Headers/aiev2_core.h
```diff
@@ -1495,14 +1525,19 @@ inline v8caccfloat msc_elem_8(v8cfloat v1, v8cfloat v2,v8caccfloat acc) { return
...
+}
 inline v16accfloat negmul_4x8_8x4( v32float v1, v32float v2                                     ) { return negmul_4x8_8x4_accuracy_fast( v1, v2             );}
 inline v16accfloat mac_4x8_8x4(    v32float v1, v32float v2, v16accfloat acc                    ) { return mac_4x8_8x4_accuracy_fast(    v1, v2, acc        );}
 inline v16accfloat msc_4x8_8x4(    v32float v1, v32float v2, v16accfloat acc                    ) { return msc_4x8_8x4_accuracy_fast(    v1, v2, acc        );}
 inline v16accfloat addmac_4x8_8x4( v32float v1, v32float v2, v16accfloat acc1, v16accfloat acc2 ) { return addmac_4x8_8x4_accuracy_fast( v1, v2, acc1, acc2 );}
-inline v16accfloat addmsc_4x8_8x4( v32float v1, v32float v2, v16accfloat acc1, v16accfloat acc2 ) { return addmsc_4x8_8x4_accuracy_fast( v1, v2, acc1, acc2 );}
-#endif
+inline v16accfloat addmsc_4x8_8x4(v32float v1, v32float v2, v16accfloat acc1,
+                                  v16accfloat acc2) {
+  return addmsc_4x8_8x4_accuracy_fast(v1, v2, acc1, acc2);
```
SagarMaheshwari99, line 1539: Are there tests for these intrinsics?

clang/lib/Headers/aiev2_enums.h
```diff
@@ -0,0 +1,95 @@
...
+#define INTLV_lo_512o1024 T512_1x2_lo
+#define INTLV_hi_8o16 T8_2x64_hi
+#define INTLV_hi_16o32 T16_2x32_hi
+#define INTLV_hi_32o64 T32_2x16_hi
+#define INTLV_hi_64o128 T64_2x8_hi
+#define INTLV_hi_128o256 T128_2x4_hi
+#define INTLV_hi_256o512 T256_2x2_hi
+#define INTLV_hi_512o1024 T512_1x2_hi
+
+#endif // __AIEV2_ENUMS_H__
```
SagarMaheshwari99, line 95: Since you are adding this, can we have the entire me_enums.h in this file?

clang/lib/Headers/aiev2_scl2vec.h
konstantinschwarz: I think initially we didn't define all these floating point vector types. Since those were added a while back, we could easily enable them now

clang/lib/Headers/aiev2_core.h
konstantinschwarz: No there were no tests added for these emulation intrinsics.
One reason is that they only use already existing intrinsics, which are all covered by tests.
It would be better to have e2e execution tests for these intrinsics, but those should go somewhere else

clang/lib/Headers/aiev2_enums.h
konstantinschwarz: I couldn't find any user of the additional defines in that file, and its not clear to me what part is really part of the API. We can add additional defines if those are required

clang/lib/Headers/aiev2_vadd.h
konstantinschwarz: Yes this is due to precision. If you look at the test, I think the total number of instructions isn't too bad either

clang/test/CodeGen/aie/aie2/aie2-upd-ext-intrinsic.cpp
konstantinschwarz: yeah these changes are really annoying, and happen with almost every major LLVM version bump..
I'll revert these changes to reduce the noise

SagarMaheshwari99 (approved):
Lgtm

PR #111, opened 2024-07-03
[AIE] Memory bank modelling in the frontend/backend

krishnamtibrewala:
Reopening #93 for review comments kindly refer #93

PR #121, opened 2024-07-10
[AIE2] Fix regress caused by adding address space in IR

krishnamtibrewala:
Fixed the drop in QoR after merging of #111, The PR does following things 
1. Consider address space case as Noop operation
2. Enable infer address space pass : This cleans up up a lot of things from IR
3. Introduce a pass that eliminated identical PHIs : Identical PHIs can be generated after `aie-addrspace-flattening` pass, in the previous flow these are handled very late during register allocation by  llvm PHIEliminater but by then these PHI had already caused ripples of effect starting from instruction select


llvm/lib/Target/AIE/AIE2TargetMachine.cpp
```diff
@@ -300,3 +300,10 @@ void AIE2TargetMachine::registerPassBuilderCallbacks(
     });
   }
 }
+
+bool AIE2TargetMachine::isNoopAddrSpaceCast(unsigned SrcAS,
```
gbossu: how about adding this to `AIEBaseTargetMachine`?

krishnamtibrewala: Done

llvm/lib/Target/AIE/AIEEliminateDuplicatePHI.cpp
```diff
@@ -0,0 +1,161 @@
...
+      LLVM_DEBUG(dbgs() << "Identical PHI nodes found\n");
+      for (auto &[BasePHI, DuplicatePHI] : IdenticalPHI) {
+        LLVM_DEBUG(dbgs() << "Base PHI: " << *BasePHI << "\n");
+        LLVM_DEBUG(dbgs() << "Duplicate PHI: " << *DuplicatePHI << "\n");
+        MIB.setInstr(*DuplicatePHI);
+        MIB.buildCopy(DuplicatePHI->getOperand(0).getReg(),
+                      BasePHI->getOperand(0).getReg());
+        Observer.changingInstr(*DuplicatePHI);
+        Observer.changedInstr(*DuplicatePHI);
+        DuplicatePHI->eraseFromParent();
```
krishnamtibrewala: Test to follow soon.

andcarminati:
Nice work. I see regressions disappearing.

llvm/lib/Target/AIE/AIEEliminateDuplicatePHI.cpp
```diff
@@ -0,0 +1,161 @@
...
+#define DEBUG_TYPE "aie-duplicate-PHI-elimination"
+
+using namespace llvm;
+
+static const char AIE_ELIMINATE_DUPLICATE_PHI[] =
+    "AIE Eliminate Duplicate PHI Pass";
+
+namespace {
+
+class AIEEliminateDuplicatePHI : public MachineFunctionPass {
```
gbossu, line 79: Nit: We could maybe have that as part of the addrspace retyping pass, because it's really there that the duplicates are introduced. Not sure what's better. I'm fine to keep it as a separate pass.

llvm/lib/Target/AIE/AIEEliminateDuplicatePHI.cpp
```diff
@@ -0,0 +1,161 @@
...
+    }
+
+    bool Changed = false;
+    for (MachineBasicBlock &MBB : MF) {
+      // If a BB has PHI nodes, the very first instruction will be PHI
+      if (MBB.empty() ||
+          MBB.begin()->getOpcode() != llvm::TargetOpcode::G_PHI) {
+        continue;
+      }
+      Changed = processPHI(MBB, MIB, Observer);
```
gbossu: Nit: function is called `processPHI`, but it gets a `MBB` as input.

llvm/lib/Target/AIE/AIEEliminateDuplicatePHI.cpp
krishnamtibrewala: After the address space retype pass, PHIs are not in state where they can be classified duplicate. After our pre-legalizer combiner passes they come into state where they are easy to classify as duplicate.

I have tried to add early checks to and use unorded_map to minimize the impact of the pass.

llvm/lib/Target/AIE/AIEEliminateDuplicatePHI.cpp
```diff
@@ -0,0 +1,161 @@
...
+        for (const auto &[BBNumber, VRegNumber] : s) {
+          hash ^= std::hash<unsigned>()(BBNumber) ^
+                  std::hash<unsigned>()(VRegNumber);
+        }
+        return hash;
+      }
+    };
+    std::unordered_map<std::set<BBVRegPair>, MachineInstr *, BBVRegPairSetHash>
+        PHIList;
+    std::set<std::pair<MachineInstr *, MachineInstr *>> IdenticalPHI;
```
gbossu: That seems quite complicated. I'd rather use `MachineInstr::isIdenticalTo()` to be really safe, and maybe have an easier algorithm which is easy to verify.

E.g. For each `G_PHI` in a BB, verify if it `isIdenticalTo` a `G_PHI` that precedes it. That's quadratic, but we can skip phis that's aren't of `p0` type, so honestly I don't think it matters much.

gbossu: We can then immediately eliminate redundant phis, no need to keep state and do the elimination in a second step.

llvm/lib/Target/AIE/AIE2TargetMachine.cpp
```diff
@@ -139,6 +139,10 @@ void AIE2PassConfig::addIRPasses() {
           }));
     }
   }
+
+  if (TM->getOptLevel() > CodeGenOptLevel::None)
+    addPass(createInferAddressSpacesPass());
```
gbossu: Do you know if running the pass helped?

llvm/lib/Target/AIE/AIEEliminateDuplicatePHI.cpp
krishnamtibrewala: The approach has been inspired from a PHI elimination pass which works at IR level and not on MIR

1. Use of `MachineInstr::isIdenticalTo()` is not safe since it does not check for commutative property.  Kindly refer to `llvm/test/CodeGen/AIE/GlobalISel/eliminate-duplicate-phi.mir`
2. Modifying/Adding instruction while iterating leads to segfaults.

Based on the above review have update the pass to consider G_PHI with pointers having default address space.

llvm/lib/Target/AIE/AIE2TargetMachine.cpp
krishnamtibrewala: Yes, it helps have added a test that shows how it cleans up the IR at early stage `llvm/test/CodeGen/AIE/aie2/GlobalISel/infer-address-space.ll`.

Without this pass we are still meeting the current best for the inner most loop but it helps in the code around.

There were many target hooks, as shared by you that are used by this pass. Currently, we only implement one that aids in avoid regression in the QoR

llvm/lib/Target/AIE/AIEEliminateDuplicatePHI.cpp
gbossu: 1. Do we need to know about commutative operands? I would expect all of the PHI operands to be in the same order, that is probably derived from the order of the BB predecessors.
2. I think we just need `make_early_inc_range`, this is how most passes deal with deleting instructions safely.

llvm/lib/Target/AIE/AIE2TargetMachine.cpp
gbossu: Sounds good :)

llvm/lib/Target/AIE/AIEEliminateDuplicatePHI.cpp
krishnamtibrewala: The need to handle commutative property for PHI nodes come an issue that was fixed by https://github.com/llvm/llvm-project/pull/75765

The idea of using hash map come from https://github.com/llvm/llvm-project/commit/aadf55d1cea24a4e5384ab8546c3d794cb1ec724 and how it affects the compile time https://llvm-compile-time-tracker.com/compare.php?from=5f4e9bf6416e45eba483a4e5e263749989fdb3b3&to=4739e6e4eb54d3736e6457249c0919b30f6c855a&stat=instructions  given we are striving for minimal compile, I prefer using the hash map.

gbossu: Creating the hash maps can also be expensive, I'm not sure we'll see any gain for our use case. Looking at the commit you mention, they only use the set approach when there are more than 32 PHI nodes. That won't be the case for us if we only look at `p0` PHI nodes. Again, I'd rather have simple and concise code that's easy to maintain unless we can really motivate the compile time gains. 

(I think we can shrink the number of lines by a factor of two if we use the simpler approach, and I'd consider this a huge win for readability)

llvm/lib/Target/AIE/AIEEliminateDuplicatePHI.cpp
```diff
@@ -0,0 +1,169 @@
...
+      std::set<std::pair<MachineInstr *, MachineInstr *>> &IdenticalPHI,
+      MachineIRBuilder &MIB, GISelObserverWrapper &Observer) {
+    bool Changed = false;
+    if (!IdenticalPHI.empty()) {
+      LLVM_DEBUG(dbgs() << "Identical PHI nodes found\n");
+      for (auto &[BasePHI, DuplicatePHI] : IdenticalPHI) {
+        LLVM_DEBUG(dbgs() << "Base PHI: " << *BasePHI << "\n");
+        LLVM_DEBUG(dbgs() << "Duplicate PHI: " << *DuplicatePHI << "\n");
+        MIB.setInstr(*DuplicatePHI);
+        MIB.buildCopy(DuplicatePHI->getOperand(0).getReg(),
```
gbossu: Can't we change the uses instead of creating a COPY?

llvm/lib/Target/AIE/AIEEliminateDuplicatePHI.cpp
krishnamtibrewala: have updated the code to remove duplicate PHI on use, and not using Hash maps any more.

llvm/lib/Target/AIE/AIEEliminateDuplicatePHI.cpp
krishnamtibrewala: Not creating any new instructions and replacing the use of duplicate PHI

llvm/lib/Target/AIE/AIEEliminateDuplicatePHI.cpp
```diff
@@ -0,0 +1,148 @@
...
+        if (MO.isReg() && MO.getReg() == DuplicateReg) {
+          Observer.changingInstr(UseInstr);
+          MO.setReg(BaseReg);
+          Observer.changedInstr(UseInstr);
+        }
+      }
+    }
+  }
+
+  bool isPointerOperandPHI(MachineRegisterInfo &MRI, const MachineInstr &MI) {
```
gbossu: Nit: That can be a free function, no need to make it a class method.

llvm/lib/Target/AIE/AIEEliminateDuplicatePHI.cpp
```diff
@@ -0,0 +1,148 @@
...
+  }
+
+  StringRef getPassName() const override { return AIE_ELIMINATE_DUPLICATE_PHI; }
+
+private:
+  void eliminateDuplicatePHIUses(MachineRegisterInfo &MRI,
+                                 GISelObserverWrapper &Observer,
+                                 Register BaseReg, Register DuplicateReg) {
+    const auto UseInstIter =
+        make_early_inc_range(MRI.use_nodbg_instructions(DuplicateReg));
```
gbossu: I think it would also make sense to rewrite debug users

llvm/lib/Target/AIE/AIEEliminateDuplicatePHI.cpp
```diff
@@ -0,0 +1,148 @@
...
+    AU.addRequired<MachineModuleInfoWrapperPass>();
+    AU.addRequired<GISelCSEAnalysisWrapperPass>();
+    AU.addRequired<TargetPassConfig>();
+    AU.setPreservesAll();
+  }
+
+  StringRef getPassName() const override { return AIE_ELIMINATE_DUPLICATE_PHI; }
+
+private:
+  void eliminateDuplicatePHIUses(MachineRegisterInfo &MRI,
```
gbossu: Nit: Same here, if you pass `Observer` and `MRI` I think that can be made a free function.

llvm/lib/Target/AIE/AIEEliminateDuplicatePHI.cpp
```diff
@@ -0,0 +1,148 @@
...
+               make_range(std::next(MachineBasicBlock::instr_iterator(MI)),
+                          MachineBasicBlock::instr_iterator(*LastPHI)))) {
+        if (!isPointerOperandPHI(MRI, PHI))
+          continue;
+        if (MI.isIdenticalTo(PHI, MachineInstr::IgnoreDefs)) {
+          LLVM_DEBUG(dbgs() << "Identical PHI nodes found\n");
+          LLVM_DEBUG(dbgs() << "  Base PHI: " << MI);
+          LLVM_DEBUG(dbgs() << "  Duplicate PHI: " << PHI);
+          eliminateDuplicatePHIUses(MRI, Observer, MI.getOperand(0).getReg(),
+                                    PHI.getOperand(0).getReg());
```
gbossu, line 137: Checking: I guess you can't really eliminate `PHI`, because it might be the next instruction that the outer loop iterates on?

llvm/test/CodeGen/AIE/GlobalISel/eliminate-duplicate-phi.mir
```diff
@@ -0,0 +1,132 @@
...
+  bb.2:
+    successors: %bb.2(0x80000000); %bb.2(100.00%)
+    %12:_(p0) = G_PHI %5:_(p0), %bb.1, %15:_(p0), %bb.2
+    %14:_(p0) = G_PHI %5:_(p0), %bb.1, %15:_(p0), %bb.2
+    %44:_(p0) = G_PHI %5:_(p0), %bb.1, %16:_(p0), %bb.2
+    %34:_(p0) = G_PHI %5:_(p0), %bb.1, %15:_(p0), %bb.2
+    %0:_(<32 x s8>) = G_LOAD %12:_(p0) :: (load (<32 x s8>) from `ptr addrspace(5) null`, addrspace 5)
+    %1:_(<32 x s8>) = G_LOAD %14:_(p0) :: (load (<32 x s8>) from `ptr addrspace(5) null`, addrspace 5)
+    %2:_(<32 x s8>) = G_LOAD %34:_(p0) :: (load (<32 x s8>) from `ptr addrspace(5) null`, addrspace 5)
+    %3:_(<32 x s8>) = G_LOAD %44:_(p0) :: (load (<32 x s8>) from `ptr addrspace(5) null`, addrspace 5)
```
gbossu, line 130: Can you order the PHI nodes and the loads in the same order? Otherwise it's a bit confusing when reading :D

llvm/lib/Target/AIE/AIEEliminateDuplicatePHI.cpp
```diff
@@ -0,0 +1,148 @@
...
+//
+// Part of the LLVM Project, under the Apache License v2.0 with LLVM Exceptions.
+// See https://llvm.org/LICENSE.txt for license information.
+// SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+//
+// (c) Copyright 2024 Advanced Micro Devices, Inc. or its affiliates
+//
+//===----------------------------------------------------------------------===//
+// Identify and remove duplicate PHI nodes with pointer type in a basic block
+// Duplicate PHI nodes are converted in to copy instructions.
```
gbossu: Comment out of date. We don't remove anything or insert copies, but just make redundant phi nodes dead.

llvm/lib/Target/AIE/AIEEliminateDuplicatePHI.cpp
krishnamtibrewala: Can we provide an example I am not very sure of the case ?

gbossu: Reading the name, I was expecting that the function deletes `PHI` from the basic block. But it "only" retargets its uses to `MI`, which I think is safer.

llvm/lib/Target/AIE/AIEEliminateDuplicatePHI.cpp
```diff
@@ -0,0 +1,160 @@
...
+///
+/// \param MRI The MachineRegisterInfo object for the current function.
+/// \param Observer GISelObserverWrapper object to observe and track changes.
+/// \param BaseReg The register to replace DuplicateReg with.
+/// \param DuplicateReg The register to be replaced.
+static void eliminateDuplicatePHIUses(MachineRegisterInfo &MRI,
+                                      GISelObserverWrapper &Observer,
+                                      Register BaseReg, Register DuplicateReg) {
+  const auto UseInstIter =
+      make_early_inc_range(MRI.use_instructions(DuplicateReg));
```
gbossu: Nit: I don't think we need `make_early_inc_range` here

llvm/lib/Target/AIE/AIEEliminateDuplicatePHI.cpp
```diff
@@ -0,0 +1,160 @@
...
+        Observer.changedInstr(UseInstr);
+      }
+    }
+  }
+}
+
+/// Returns true if the instruction is a PHI node with a pointer type operand.
+/// \param MRI The MachineRegisterInfo object for the current function.
+/// \param MI The instruction to check.
+static bool isPointerOperandPHI(MachineRegisterInfo &MRI,
```
gbossu: Nit: I would just rename to `isPointerPHI` or `isPointerTypePHI`

llvm/lib/Target/AIE/AIE2TargetMachine.cpp
martien-de-jong: I wouldn't mind always adding this. -O0 is not a useful optimization level anyway, since we always run out of stack.

llvm/lib/Target/AIE/AIEEliminateDuplicatePHI.cpp
martien-de-jong: Since the result type and operand types should match anyway.

gbossu (dismissed):

llvm/lib/Target/AIE/AIEEliminateDuplicatePHI.cpp
```diff
@@ -0,0 +1,160 @@
...
+
+  StringRef getPassName() const override { return AIE_ELIMINATE_DUPLICATE_PHI; }
+
+private:
+  bool processMBBWithPHI(MachineBasicBlock &MBB, MachineRegisterInfo &MRI,
+                         MachineIRBuilder &MIB,
+                         GISelObserverWrapper &Observer) {
+    bool Changed = false;
+    auto LastPHI = MBB.getFirstNonPHI();
+    for (auto &MI : make_early_inc_range(make_range(MBB.begin(), LastPHI))) {
```
martien-de-jong: nit: I think this can be expressed using `MBB.phis()`

llvm/lib/Target/AIE/AIEEliminateDuplicatePHI.cpp
```diff
@@ -0,0 +1,160 @@
...
+      MIB.setChangeObserver(Observer);
+    }
+
+    bool Changed = false;
+    for (MachineBasicBlock &MBB : MF) {
+      // If a BB has PHI nodes, the very first instruction will be PHI
+      if (MBB.empty() ||
+          MBB.begin()->getOpcode() != llvm::TargetOpcode::G_PHI) {
+        continue;
+      }
```
martien-de-jong: nit: I would just call processMBBWithPHI unconditionally. It will finish quickly if there are no PHIs.

llvm/lib/Target/AIE/AIEEliminateDuplicatePHI.cpp
```diff
@@ -0,0 +1,160 @@
...
+      Changed = processMBBWithPHI(MBB, MRI, MIB, Observer);
```
martien-de-jong: This redefines Changed, so ignores the result from earlier blocks.

llvm/lib/Target/AIE/AIEEliminateDuplicatePHI.cpp
```diff
@@ -0,0 +1,160 @@
...
+        getAnalysis<GISelCSEAnalysisWrapperPass>().getCSEWrapper();
+    auto *CSEInfo = &Wrapper.get(TPC.getCSEConfig());
+    std::unique_ptr<MachineIRBuilder> Builder =
+        CSEInfo ? std::make_unique<CSEMIRBuilder>()
+                : std::make_unique<MachineIRBuilder>();
+    Builder->setMF(MF);
+    MachineIRBuilder &MIB = *Builder;
+    // Set Observer to inform change of register type or change in op-code
+    GISelObserverWrapper Observer;
+    if (CSEInfo) {
```
martien-de-jong, line 98: Aren't we implementing a dedicated CSE pass here? So could we merge this with another run of a CSE pass after address space lowering?

krishnamtibrewala: This basically reuse the very first CSE pass created here.  For now that is in address space flattening.

llvm/lib/Target/AIE/AIEEliminateDuplicatePHI.cpp
krishnamtibrewala: Based on two passes that does PHI elimination, they took this approach so following the same.

llvm/lib/Target/AIE/AIEEliminateDuplicatePHI.cpp
krishnamtibrewala: We need it, else the iteration below it does not work for all the uses. Without `make_early_inc_range` it will work for only the very first use.

martien-de-jong: Trying to understand, the only thing I see that could invalidate the iterator is the Observer?

llvm/lib/Target/AIE/AIEEliminateDuplicatePHI.cpp
gbossu: I'm fine without this change, but I agree with Martien. Some parts of the LLVM codebase aren't great examples for code readability, so I wouldn't always assume that the way similar passes do is the best way. Even if it very often is ;)

llvm/lib/Target/AIE/AIEEliminateDuplicatePHI.cpp
gbossu: Good catch!

llvm/lib/Target/AIE/AIEEliminateDuplicatePHI.cpp
krishnamtibrewala: Directly calling `processMBB`

llvm/lib/Target/AIE/AIEEliminateDuplicatePHI.cpp
krishnamtibrewala: The reason is `MO.setReg` this calls `removeRegOperandFromUseList` which updates `VRegInfo` and we are using `MRI.use_instructions` which also relies in `VRegInfo`

gbossu: But I think the `MI` is still valid and not re-allocated. Did you try without `make_early_inc_range`? I would really expect it to work

krishnamtibrewala: I did try without `make_early_inc_range` it only updates the first use MI.

krishnamtibrewala: As discussed today, have updated the code. Caching the use MI and iteration on those.

martien-de-jong (dismissed):

llvm/lib/Target/AIE/AIEEliminateDuplicatePHI.cpp
```diff
@@ -0,0 +1,160 @@
+//===-- AIEEliminateDuplicatePHI.cpp - Identify and remove duplicate PHI --===//
+//
+// Part of the LLVM Project, under the Apache License v2.0 with LLVM Exceptions.
```
konstantinschwarz: Nit: files that are solely added by us should have, see other files in AIE target folder
```
// This file is licensed under the Apache License v2.0 with LLVM Exceptions.
```

gbossu (approved):

konstantinschwarz (approved):

PR #125, opened 2024-07-16
[AIE2] Add support for ZOL loops for SWP

andcarminati:
Some refactoring was also included.


llvm/lib/Target/AIE/AIEBasePipelinerLoopInfo.cpp
```diff
@@ -32,6 +35,123 @@ cl::opt<bool> TrackRegPressure(
     "aie-pipeliner-track-regpressure",
     cl::desc("Refuse SWP schedules likely to run into register spills"),
     cl::init(true), cl::Hidden);
+cl::opt<bool> HoistZOLAdjust(
```
andcarminati, line 44: FYI: If we hoist we can have get a better performance (no clue why). If we don't hoist, we can change the LoopsStart expansion to reuse this non-hoisted add. I already have this change, but I saw no direct gains using this approach.

llvm/lib/Target/AIE/AIEBasePipelinerLoopInfo.cpp
```diff
@@ -21,7 +24,7 @@
 #include "llvm/Support/raw_ostream.h"
 #include <memory>
 
-#define DEBUG_TYPE "pipeliner"
+#define DEBUG_TYPE "aie-pipeliner"
```
andcarminati, line 29: This was just to filter a bit the output.

llvm/lib/Target/AIE/AIEBasePipelinerLoopInfo.cpp
```diff
@@ -627,18 +659,151 @@ void DownCountLoop::startExpand() {
...
 
+/// A Zero overhead loop enters the software pipeliner with
+/// a PseudoLoopEnd as single loop terminator and
+/// a LoopStart instruction somewhere (usually close to the end)
+/// in the preheader block
+///
+class ZeroOverheadLoop : public AIEBasePipelinerLoopInfo {
+  // The instruction initializing the counter.
+  MachineInstr *Init;
+  // The instruction defining the value used the to initialize
```
andcarminati: ~~the~~  to

llvm/lib/Target/AIE/AIEBasePipelinerLoopInfo.cpp
```diff
@@ -211,6 +331,45 @@ AIEBasePipelinerLoopInfo::checkLoopControl(MachineInstr *EndLoop) {
...
+
+bool AIEBasePipelinerLoopInfo::shouldUseSchedule(SwingSchedulerDAG &SSD,
+                                                 SMSchedule &SMS) {
+  const int PrologCount = SMS.getMaxStageCount();
+  const int StageCount = PrologCount + 1;
+  LLVM_DEBUG(dbgs() << "StageCount = " << StageCount << "\n");
+  if (StageCount <= 1) {
+    LLVM_DEBUG(dbgs() << "PLI: Rejected schedule (StageCount <= 1)\n");
+    return false;
+  }
```
martien-de-jong, line 228: I am a bit surprised that this isn't a standard reason for rejection in MachinePipeliner. StageCount = 1 basically says it's not software-pipelined.

llvm/lib/Target/AIE/AIEBasePipelinerLoopInfo.cpp
```diff
@@ -381,125 +538,13 @@ DownCountLoop::Assessment DownCountLoop::accept(MachineInstr *EndLoop) {
...
-}
-
 bool DownCountLoop::shouldUseSchedule(SwingSchedulerDAG &SSD, SMSchedule &SMS) {
   const int PrologCount = SMS.getMaxStageCount();
   const int StageCount = PrologCount + 1;
-  LLVM_DEBUG(dbgs() << "StageCount = " << StageCount << "\n");
-  if (StageCount <= 1) {
-    LLVM_DEBUG(dbgs() << "PLI: Rejected schedule (StageCount <= 1)\n");
+
+  // If AIEBasePipelinerLoopInfo refuses it, let's conservativelly
```
martien-de-jong: nit: conservatively

llvm/lib/Target/AIE/AIEBasePipelinerLoopInfo.cpp
```diff
@@ -627,18 +659,151 @@ void DownCountLoop::startExpand() {
...
+  if (!LoopStartBlock) {
+    LLVM_DEBUG(dbgs() << "Could not locate loop start block!\n");
+    return false;
+  }
+
+  auto FindLoopStart = [&](MachineBasicBlock &Block) {
+    MachineInstr *LoopStart = nullptr;
+    for (auto &MI : reverse(Block)) {
+      if (TII.isHardwareLoopStart(MI.getOpcode())) {
+        LoopStart = &MI;
```
martien-de-jong: Eliminate LoopStart and just return &MI here and nullptr at the end? perhaps rename MI to Candidate or something like that.

llvm/lib/Target/AIE/AIEBasePipelinerLoopInfo.cpp
```diff
@@ -627,18 +659,151 @@ void DownCountLoop::startExpand() {
...
+    Register Reg = DefTripCount->getOperand(0).getReg();
+    Register NewReg = MRI.createVirtualRegister(MRI.getRegClass(Reg));
+    MachineBasicBlock::iterator Insert(Init);
+    MachineInstr *AdjacentInstr = Init;
+
+    if (HoistZOLAdjust) {
+      // Insert the adjustment just after the instruction that defines it.
+      // Probably it will be hoisted.
+      AdjacentInstr = DefTripCount;
+      Insert = {DefTripCount};
```
martien-de-jong: Perhaps use a more explicit constructor? I can suggest 
MachineBasicBlock::iterator AfterDef(DefTripCount);
AfterDef++;
Insert = AfterDef;

alternatively, isn't there a DefTripCount.getIterator() ?

llvm/lib/Target/AIE/AIEBasePipelinerLoopInfo.cpp
martien-de-jong: I would put this in a code comment. The rationale of reuse in LoopStart expansion is utterly hidden otherwise.

llvm/lib/Target/AIE/AIEBasePipelinerLoopInfo.cpp
```diff
@@ -627,18 +659,151 @@ void DownCountLoop::startExpand() {
...
+
+  std::optional<bool> createTripCountGreaterCondition(
+      int TC, MachineBasicBlock &MBB,
+      SmallVectorImpl<MachineOperand> &Cond) override;
+};
+
+bool ZeroOverheadLoop::accept(MachineInstr *EndLoop) {
+  // We are using LoopMinTripCount below just for testing purposes.
+  // For MIR test cases without IR, we can't encode loop-related metadata.
+  if (!MinTripCount && !LoopMinTripCount) {
```
andcarminati: This should be:

```
  if (!MinTripCount && LoopMinTripCount <= 0) {
    LLVM_DEBUG(dbgs() << "Unbounded loop detected!\n");
    return false;
  }
```

llvm/lib/Target/AIE/AIEBasePipelinerLoopInfo.cpp
```diff
@@ -32,6 +35,123 @@ cl::opt<bool> TrackRegPressure(
...
+    cl::desc("Host the trip count adjustment for ZOL (when possible)"),
+    cl::init(true), cl::Hidden);
+
+/// Get an instruction sequence from an \p SMS schedule that is estimated
+/// to have a similar register pressure to what would come out of the
+/// premisched.
+///
+/// Currently, this uses "reverse stage extraction", i.e. the stages are laid
+/// down in reverse order and their instructions are not interleaved.
+std::vector<MachineInstr *> getInstrSequence(SMSchedule &Sched) {
```
martien-de-jong: CHECK: this is just factored out of DownCountLoop.

llvm/lib/Target/AIE/AIEBasePipelinerLoopInfo.cpp
```diff
@@ -627,18 +659,151 @@ void DownCountLoop::startExpand() {
...
+      : AIEBasePipelinerLoopInfo(EndLoop, TII) {}
+  void adjustTripCount(int TripCountAdjust) override;
+  bool accept(MachineInstr *EndLoop);
+
+  std::optional<bool> createTripCountGreaterCondition(
+      int TC, MachineBasicBlock &MBB,
+      SmallVectorImpl<MachineOperand> &Cond) override;
+};
+
+bool ZeroOverheadLoop::accept(MachineInstr *EndLoop) {
```
martien-de-jong: the bool return here was a bit lazy on my part. We could consider to lift the acceptance enum to AIEBase level

llvm/lib/Target/AIE/AIEBasePipelinerLoopInfo.cpp
andcarminati: Indeed, this was already checked, I just refactored to the base class. Do you think it is dead code?

martien-de-jong: This may be a relic from the time I didn't know too many details of the MachinePipeliner. Perhaps it's a speed hack to skip some otherwise redundant processing. Or, I wanted the rejection message to be included in my "grep PLI" summary.

martien-de-jong (commented):
Since we can't generate a createTripCountGreater branch, I would like to know what would happen if we have MinTripCount=2 and SWP creates a 3 or more stagecount schedule. I think we should pick this up early, so that we can try a bigger II with a lower stage count.
Testing that we really take this path may be tricky. Perhaps check the logging. Otherwise perhaps invest in a unittest with a mock SWP driver?

andcarminati:
> Since we can't generate a createTripCountGreater branch, I would like to know what would happen if we have MinTripCount=2 and SWP creates a 3 or more stagecount schedule. I think we should pick this up early, so that we can try a bigger II with a lower stage count. Testing that we really take this path may be tricky. Perhaps check the logging. Otherwise perhaps invest in a unittest with a mock SWP driver?

I extended a bit to cover this case, I also include some tests for this particular situations. I also created a filter to not apply HL to loops with MinTripCount=1, according to our internal discussion.

llvm/lib/Target/AIE/Utils/AIE2LoopUtils.cpp
```diff
@@ -0,0 +1,48 @@
+//===- AIE2LoppUtils.cpp --------------------------------------------------===//
```
martien-de-jong: nit: LoopUtils

llvm/lib/Target/AIE/Utils/AIE2LoopUtils.cpp
```diff
@@ -0,0 +1,48 @@
...
+
+namespace AIE2LoopUtils {
+
+std::optional<int64_t> getMinTripCount(const MDNode *LoopID) {
+
+  assert(LoopID->getNumOperands() > 0 && "requires at least one operand");
+  assert(LoopID->getOperand(0) == LoopID && "invalid loop");
+
+  int64_t MinTripCount = 0;
+  for (unsigned i = 1, e = LoopID->getNumOperands(); i < e; ++i) {
```
martien-de-jong: nit: Uppercase variables?

llvm/lib/Target/AIE/Utils/AIE2LoopUtils.cpp
```diff
@@ -0,0 +1,48 @@
...
+#include "llvm/IR/Constants.h"
+
+#define DEBUG_TYPE "aie2looputils"
+
+namespace llvm {
+
+namespace AIE2LoopUtils {
+
+std::optional<int64_t> getMinTripCount(const MDNode *LoopID) {
+
```
martien-de-jong: I guess you could put the nullptr check on LoopID here as well. no LoopID -> nullopt.

llvm/lib/Target/AIE/Utils/AIE2LoopUtils.h
```diff
@@ -0,0 +1,29 @@
...
+//
+// This file contains helper functions to handle loops and its metadata.
+//
+//===----------------------------------------------------------------------===//
+#ifndef LLVM_LIB_TARGET_AIE_UTILS_AIE2LOOPUTILS_H
+#define LLVM_LIB_TARGET_AIE_UTILS_AIE2LOOPUTILS_H
+
+#include "llvm/Analysis/LoopInfo.h"
+
+namespace llvm {
```
martien-de-jong: nit `namespace llvm::AIE2LoopUitls {` ?

llvm/lib/Target/AIE/Utils/CMakeLists.txt
```diff
@@ -6,6 +6,7 @@
 # (c) Copyright 2023-2024 Advanced Micro Devices, Inc. or its affiliates
 add_llvm_component_library(LLVMAIEUtils
   AIEBaseInfo.cpp
+  AIE2LoopUtils.cpp
```
martien-de-jong: Very much appreciated ;-) Although I think it should just be AIELoopUtils. There's nothing AIE2 specific in it.

llvm/lib/Target/AIE/AIE2TargetTransformInfo.cpp
```diff
@@ -27,6 +28,10 @@ static cl::opt<bool>
                 cl::desc("Enable true zero overhead hardware loops on AIE"),
                 cl::init(false), cl::Hidden);
 
+static cl::opt<int> MinIterCountHLReject(
+    "aie-hardware-loops-minitercount", cl::Hidden, cl::init(1),
+    cl::desc("Minimum trip count threshold for HL rejection"));
```
martien-de-jong, line 33: Future: perhaps we can tie this with the "prepare-for-software-pipelining" pragma

llvm/test/CodeGen/AIE/aie2/hardware-loops/zol-md-accept-reject.ll
```diff
@@ -0,0 +1,79 @@
...
+
+define void @simple_loop_accept_no_md(i32 noundef %n, ptr nocapture readonly %in, ptr nocapture writeonly %out) {
+entry:
+  %cmp4 = icmp sgt i32 %n, 0
+  br i1 %cmp4, label %for.body, label %for.cond.cleanup
+
+for.cond.cleanup:                                 ; preds = %for.body, %entry
+  ret void
+
+; CHECK: [[DEC:%[0-9]+]] = call i1 @llvm.loop.decrement.i32(i32 1)
```
martien-de-jong: Shouldn't `no metadata` be counted as `mintripcount == 0`?
Perhaps try to find a static bound.

llvm/lib/Target/AIE/AIE2TargetTransformInfo.cpp
andcarminati: Hi @martien-de-jong , I think this is a good idea.

llvm/test/CodeGen/AIE/aie2/hardware-loops/zol-md-accept-reject.ll
andcarminati: Maybe it is not necessary. With no metadata we will accept as HL but will reject for SWP. I think this type of loop can be part of the initialization, for example. In this case we can let the compiler to and apply HL freely, otherwise we will need to annotate every loop. What do you think?

andcarminati: Looking again and benchmarking again I see this as a good think, it is included.

andcarminati:
Hi @konstantinschwarz , @martien-de-jong and @gbossu ,

This PR is in a state without any additional failures and no significant regression. We have an average regression of 0.1% in Instruction count and an average increase of 0,97% code size.

For GEMM (negative means Total Insn Count decrease) and code size:
```
GEMM_bf16_0	-2,67%  (Insn Count decrease) 1,15% (Code increase)
GEMM_int8_0	-2,44%  (Insn Count decrease) 2,05%  (Code increase)
GEMM_int8_1	-2,72%  (Insn Count decrease) 1,78%  (Code increase)
GEMM_bf16_1	-3,06%  (Insn Count decrease) 1,01%  (Code increase)
```

For CONV2D_0, we keep 10 cycles in the innermost loop with a slightly instruction count increase (outside the innermost loop):
```
Conv2D_0         0,10%  (Insn Count increase) 2,51%  (Code increase)
```

Biggest outlier:
```
Conv2D_Transpose_bf16_AIE2_1  3,15% (Insn Count increase) 2,27% (Code increase)

```

llvm/test/CodeGen/AIE/aie2/hardware-loops/zol-md-accept-reject.ll
martien-de-jong: I'm not quite sure I know what you mean. DownCountLoop would detect a static tripcount if it was given as a constant, even if no metadata was present. I don't think that's much more difficult for ZOL, without 'annotating' it, if with that you mean providing min-itercount metadata from the value of the constant.
So, in zol construction,  if you have a constant tripcount, I would allow it as a hardware loop, trusting that SWP would find the same tripcount from the LoopStart constant.

andcarminati: Hi @martien-de-jong, indeed this is quite simple and we already detect this case with constant trip count, because the update of this trip count is can be done more efficiently.  I found some performance issues in some benchmarks when I release the breaks for HL:

1 - By allowing unconditionally, we can fall in some cases where the loop is quite simple and outside the kernel function.  I saw some loops with small headers and just 1 iteration, is this case the setup of HL was costly, with performance degradation.

2 - Talking about kernels, by allowing freely cases where we don't know much about the loop (in TTI - no metadata), we give to Pipeliner something that we are not sure if it will be accepted. Once we create ZOL, we can reject much more easily than DCL (the loops or the scheduler).

In this moment, by analyzing the numbers, I directed the PR to a state were we allow ZOL for the cases were we more confident: annotated loops, ie cases where plan to apply pipeline with a less rejection rate. I could suggest an follow-up work that can try to overcome this problem by converting some rejected ZOL loops in DCL on the fly (MPLoopInfo). We can also try to improve TTI WRT loops.

Sometimes I feel that HL transformation kicks in too early, where we can't analyze tome tradeoffs....

martien-de-jong (dismissed):

llvm/lib/Target/AIE/AIEBasePipelinerLoopInfo.cpp
```diff
@@ -381,125 +532,13 @@ DownCountLoop::Assessment DownCountLoop::accept(MachineInstr *EndLoop) {
...
-    }
-  }
-
-  return Seq;
-}
-
-/// Replay the instructions in \p Seq and collect all the livein registers.
-std::vector<RegisterMaskPair>
-collectLiveInRegs(const std::vector<MachineInstr *> &Seq,
-                  const MachineFunction &MF) {
```
gbossu: Nit: I'd really prefer not moving the code around too much, otherwise that creates lots of "irrelevant" diff lines and conflicts. Could we maybe keep the functions there? Either by keeping their users below, or pre-declaring them?

llvm/lib/Target/AIE/Utils/AIELoopUtils.cpp
```diff
@@ -0,0 +1,48 @@
...
+//
+//===----------------------------------------------------------------------===//
+#include "AIELoopUtils.h"
+#include "llvm/IR/Constants.h"
+
+#define DEBUG_TYPE "aielooputils"
+
+namespace llvm::AIELoopUtils {
+
+std::optional<int64_t> getMinTripCount(const MDNode *LoopID) {
```
gbossu, line 17: Thanks for outlining that piece of code :)

llvm/lib/Target/AIE/AIEBasePipelinerLoopInfo.cpp
andcarminati: Done! It looks better now.

gbossu (approved):
LGTM, great work looking into all the benchmarks and fixing regressions!
I also verified this branch with my prototype for tweaking the SW pipeliner, and get even better results now 🎉

PR #127, opened 2024-07-18
Update aie-public to upstream commit 6e41d60a717132fadac74abe61ac6a9b1ca98778

konstantinschwarz:


stephenneuendorffer (dismissed):

libc/cmake/modules/LLVMLibCCompileOptionRules.cmake
```diff
@@ -1,3 +1,10 @@
+#
+# Part of the LLVM Project, under the Apache License v2.0 with LLVM Exceptions.
+# See https://llvm.org/LICENSE.txt for license information.
+# SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+#
+# Modifications (c) Copyright 2023-2024 Advanced Micro Devices, Inc. or its affiliates
```
gbossu: Nit: just 2024?

llvm/lib/Target/AIE/AIEMachineScheduler.h
```diff
@@ -114,8 +109,6 @@ class AIEPostRASchedStrategy : public PostGenericScheduler {
   /// cycle of instructions to be scheduled.
   int getMaxDeltaCycles(const SchedBoundary &Zone) const;
 
-  SchedBoundary Bot;
```
gbossu: Yay!

llvm/lib/CodeGen/SelectionDAG/SelectionDAGBuilder.cpp
```diff
@@ -10446,8 +10446,8 @@ TargetLowering::LowerCallTo(TargetLowering::CallLoweringInfo &CLI) const {
     SmallVector<EVT, 4> ValueVTs;
     SmallVector<Type *, 4> IRTypes;
     TypeSize Offset = TypeSize::get(0, Args[i].Ty->isScalableTy());
-    ComputeValueVTs(*this, DL, Args[i].Ty, ValueVTs, /*Offsets=*/nullptr,
-                    Offset, &IRTypes);
+    ComputeValueVTs(*this, DL, Args[i].Ty, ValueVTs, /*MemVTs*/ nullptr,
+                    /*Offsets=*/nullptr, Offset, &IRTypes);
```
gbossu: Note: If we drop AIE1 support, those changes we made here can go. I think the `IRTypes` parameter is something I added a long time ago to identify pointer types.

llvm/test/CodeGen/AIE/aie2/hoist_s_reg_copy.mir
```diff
@@ -84,7 +84,8 @@ body:             |
   ; CHECK-NEXT: {{  $}}
   ; CHECK-NEXT:   [[PHI:%[0-9]+]]:er = PHI [[MOV_RLC_imm10_pseudo]], %bb.0, %7, %bb.1
   ; CHECK-NEXT:   [[PHI1:%[0-9]+]]:ep = PHI [[COPY]], %bb.0, %9, %bb.1
-  ; CHECK-NEXT:   [[VLDA_UPS_S64_D32_ag_idx_imm:%[0-9]+]]:acc512 = VLDA_UPS_S64_D32_ag_idx_imm [[COPY2]], [[PHI1]], 0, implicit-def $srups_of, implicit $crsat, implicit $crupssign :: (load (<8 x s32>))
+  ; CHECK-NEXT:   [[COPY5:%[0-9]+]]:mss = COPY [[COPY1]]
+  ; CHECK-NEXT:   [[VLDA_UPS_S64_D32_ag_idx_imm:%[0-9]+]]:acc512 = VLDA_UPS_S64_D32_ag_idx_imm [[COPY5]], [[PHI1]], 0, implicit-def $srups_of, implicit $crsat, implicit $crupssign :: (load (<8 x s32>))
```
gbossu: So this LLVM bump seems to restore all of the expectations for this test as it was initially written. Looks good.

.github/workflows/amd-aie-tests.yml
```diff
@@ -29,5 +29,5 @@ jobs:
       build_target: check-llvm-aie check-clang-aie check-lld-aie
       projects: clang;lld
       cache-key: amd-aie
-      extra_cmake_args: '-DLLVM_USE_LINKER=gold -C ../clang/cmake/caches/Peano-AIE.cmake'
+      extra_cmake_args: '-DLLVM_USE_LINKER=gold -C clang/cmake/caches/Peano-AIE.cmake'
```
gbossu, line 32: What exactly changed here?

gbossu (dismissed):

llvm/lib/CodeGen/SelectionDAG/SelectionDAGBuilder.cpp
konstantinschwarz: Yes, GlobalISel CallLowering tracks pointer type/address spaces separately nowadays, so this change is only needed for SelectionDAG

.github/workflows/amd-aie-tests.yml
konstantinschwarz: Upstream CI made a change to make use of reusable workflow files, and we use those same workflows under the hood.
The cmake config invocation now happens outside of the build directory, i.e. inside the main source repository.
Hence we need to update the relative path to <clang>

PR #128, opened 2024-07-19
[RFC][AIE2] Add an IR Phi reordering pass

andcarminati:
This pass helps Greedy to avoid spills.

This pass is indented to solve spill problems related to `-fno-unroll-loops`. No specific tests yet, just for evaluation and comments.

### DIFF
--- a/llvm/lib/Target/AIE/AIEPHISorterPass.cpp
+++ b/llvm/lib/Target/AIE/AIEPHISorterPass.cpp
@@ -1,0 +1,101 @@
+//==========-- AIEPHISorterPass.cpp - Sort PHI nodes in loops --==============//
+//
+// This file is licensed under the Apache License v2.0 with LLVM Exceptions.
+// See https://llvm.org/LICENSE.txt for license information.
+// SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+//
+// (c) Copyright 2024 Advanced Micro Devices, Inc. or its affiliates
+//
+//===----------------------------------------------------------------------===//
+//
+// This Pass reorders PHI nodes in a more friendly way to Greedy Allocator
+//
+//===----------------------------------------------------------------------===//
+
+#include "AIE2.h"
+#include "llvm/ADT/StringRef.h"
+#include "llvm/Analysis/LoopInfo.h"
+#include "llvm/IR/Constants.h"
+#include "llvm/IR/IRBuilder.h"
+#include "llvm/IR/Instructions.h"
+#include "llvm/IR/IntrinsicInst.h"
+#include "llvm/IR/IntrinsicsAArch64.h"
+#include "llvm/IR/LLVMContext.h"
+#include "llvm/InitializePasses.h"
+#include "llvm/Support/Debug.h"
+#include "llvm/Transforms/Utils/Cloning.h"
+
+using namespace llvm;
+
+#define DEBUG_TYPE "aie-phi-sorter"
+
+namespace {
+struct AIEPHISorter : public FunctionPass {
+  static char ID; // Pass identification, replacement for typeid
+  AIEPHISorter() : FunctionPass(ID) {
+    initializeAIEPHISorterPass(*PassRegistry::getPassRegistry());
+  }
+
+  bool runOnFunction(Function &F) override;
+
+  void getAnalysisUsage(AnalysisUsage &AU) const override {
+    AU.addRequired<LoopInfoWrapperPass>();
+    AU.setPreservesAll();
+  }
+};
+} // end anonymous namespace
+
+char AIEPHISorter::ID = 0;
+static const char *name = "AIE PHI Sorter Pass";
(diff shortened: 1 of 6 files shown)


llvm/lib/Target/AIE/AIEPHISorterPass.cpp
```diff
@@ -0,0 +1,101 @@
...
+//
+// (c) Copyright 2024 Advanced Micro Devices, Inc. or its affiliates
+//
+//===----------------------------------------------------------------------===//
+//
+// This Pass reorders PHI nodes in a more friendly way to Greedy Allocator
+//
+//===----------------------------------------------------------------------===//
+
+#include "AIE2.h"
```
martien-de-jong, line 15: If it is AIE2 specific, consider renaming to AIE2PHISorterPass. Otherwise, try to avoid including AIE2.h

llvm/lib/Target/AIE/AIEPHISorterPass.cpp
```diff
@@ -0,0 +1,101 @@
...
+        for (BasicBlock *BB : PHI.blocks()) {
+          // Check basically if this loop block
+          // is part of the body.
+          if (BB == &B)
+            WorkList.push_back(&PHI);
+        }
+      }
+
+      if (WorkList.empty())
+        continue;
```
martien-de-jong, line 76: I guess InstrOrder could be declared after this continue.

llvm/lib/Target/AIE/AIEPHISorterPass.cpp
```diff
@@ -0,0 +1,101 @@
...
+  for (BasicBlock &B : F) {
+    if (LI.getLoopFor(&B)) {
+      SmallVector<PHINode *> WorkList;
+      std::map<const Instruction *, unsigned> InstrOrder;
+      // Collect all interesting recursive PHI nodes.
+      for (auto &PHI : B.phis()) {
+        for (BasicBlock *BB : PHI.blocks()) {
+          // Check basically if this loop block
+          // is part of the body.
+          if (BB == &B)
```
martien-de-jong, line 70: I'm wondering why you use LoopInfo. If this condition holds, you have a single block loop. If getLoopFor holds, it's a block somewhere in a multi-block loop. Therefore, this inner condition is stronger than the outer one.
If you only want to process 'relevant' blocks, consider traversing the loop tree rather than the basic block list. In particular, you could limit yourself to the header blocks of the leaf loops, ie. the ones without children.

llvm/lib/Target/AIE/AIEPHISorterPass.cpp
```diff
@@ -0,0 +1,101 @@
...
+      }
+
+      if (WorkList.empty())
+        continue;
+
+      Changed = true;
+      unsigned PositionCounter = 0;
+
+      // Collect instruction ordering.
+      for (auto &I : B)
```
martien-de-jong, line 82: How about making PositionCounter local-scoped by declaring it here as `const unsigned PositionCounter = InstrOrder.size();`

llvm/lib/Target/AIE/AIEPHISorterPass.cpp
```diff
@@ -0,0 +1,101 @@
...
+#include "llvm/Analysis/LoopInfo.h"
+#include "llvm/IR/Constants.h"
+#include "llvm/IR/IRBuilder.h"
+#include "llvm/IR/Instructions.h"
+#include "llvm/IR/IntrinsicInst.h"
+#include "llvm/IR/IntrinsicsAArch64.h"
+#include "llvm/IR/LLVMContext.h"
+#include "llvm/InitializePasses.h"
+#include "llvm/Support/Debug.h"
+#include "llvm/Transforms/Utils/Cloning.h"
```
martien-de-jong, line 26: I think we only need a fraction of these includes. And as remarked, I think it could be independent on LoopInfo. If you want to generalize later, I would already make the 'is part of the body' test based on comparing getLoopFor of both blocks for equality.

qcolombet:
I feel that this patch is about pushing the coalescing (or reg alloc) towards a particular direction.
But this is done indirectly.
What is missing to be able to drive the related passes directly?

andcarminati:
> I feel that this patch is about pushing the coalescing (or reg alloc) towards a particular direction. But this is done indirectly. What is missing to be able to drive the related passes directly?

Hi @qcolombet, indeed it is. This PR was more for discussions, because the effect is very hard to identify. We had one benchmark where the `fno-unroll-loops` induced a different ordering in the phi nodes. The result after GSelect was also the same (except phi ordering) and this ended in side-effects after Machine Pipeliner. In Both cases we had sw-pipeline, but  with `fno-unroll-loops` we had also spills.

qcolombet:
Thanks @andcarminati for the confirmation!
Could you file an issue, capture and upload the outputs of `-debug-only regalloc` with and without the phi ordering change that results in spill/non-spill?
I can take a look and maybe identify what heuristics could be improvement.

PR #129, opened 2024-07-19
[Generic][AIE2] Combiner for shufflevectors that use build vector

ValentijnvdBeek:
Transforms a shufflevector that uses a build vector or undefined into just a build vector. This can be done is because a shuffle vector lowering is an unmerge and then merge. Since build is a merge, the merge and unmerge cancel each other out and we can just merge the vector directly.

Example:
```
    %1:_(s32) = COPY $r0
    %3:_(<8 x s32>) = G_IMPLICIT_DEF
    %5:_(s32) = G_IMPLICIT_DEF
    %2:_(<8 x s32>) = G_BUILD_VECTOR %1(s32), %5(s32), %5(s32), %5(s32), %5(s32), %5(s32), %5(s32), %5(s32)
    %0:_(<8 x s32>) = G_SHUFFLE_VECTOR %2(<8 x s32>), %3, shufflemask(0, 0, 0, 0, 0, 0, 0, 0)
    ===>
    %2:_(<8 x s32>) = G_BUILD_VECTOR %1(s32), %1(s32), %1(s32), %1(s32), %1(s32), %1(s32), %1(s32), %1(s32)
```


llvm/include/llvm/CodeGen/GlobalISel/GenericMachineInstrs.h
```diff
@@ -240,13 +240,22 @@ class GMergeLikeInstr : public GenericMachineInstr {
...
+    case TargetOpcode::G_SHUFFLE_VECTOR:
       return true;
     default:
       return false;
     }
   }
 };
 
+/// Represents a G_SHUFFLE_VECTOR
+class GShuffleVector : public GMergeLikeInstr {
```
konstantinschwarz: G_SHUFFLE_VECTOR takes the shufflemask as the last operand, which shouldn't be part of the `getNumSources` I think?
To me that's what makes G_SHUFFLE_VECTOR different compared to the other "merge like" instructions.
I'd rather not make it part of that group, as this can have surprising effects to other existing combines

llvm/lib/CodeGen/GlobalISel/CombinerHelper.cpp
```diff
@@ -2147,6 +2149,77 @@ static Register peekThroughBitcast(Register Reg,
...
+  // Check if the Source registers are either merges or implicit definitions
+  auto *SrcInstr1 = getOpcodeDef<GBuildVector>(SrcReg1, MRI);
+  auto *SrcInstr2 = getOpcodeDef<GBuildVector>(SrcReg2, MRI);
+  auto *IsUndef1 = getOpcodeDef(TargetOpcode::G_IMPLICIT_DEF, SrcReg1, MRI);
+  auto *IsUndef2 = getOpcodeDef(TargetOpcode::G_IMPLICIT_DEF, SrcReg2, MRI);
+
+  // Our inputs need to be either be build vectors or undefined, register inputs
+  // break this optimization. You could maybe do something clever were you
+  // concatenate vectors to save half a build vector.
+  if ((SrcInstr1 == 0 && IsUndef1 == 0) || (SrcInstr2 == 0 && IsUndef2 == 0))
```
konstantinschwarz: These are pointers, just use `(!SrcInstr1 && !IsUndef1) || ...)`

llvm/test/CodeGen/AIE/aie2/GlobalISel/prelegalizercombiner-shufflevector-buildvector.mir
```diff
@@ -0,0 +1,322 @@
...
+---
+name:            shufflevector_undef_undef
+legalized:       false
+body:             |
+  bb.1.entry:
+    ; CHECK-LABEL: name: shufflevector_undef_undef
+    ; CHECK: [[DEF:%[0-9]+]]:_(<16 x s32>) = G_IMPLICIT_DEF
+    ; CHECK-NEXT: $x0 = COPY [[DEF]](<16 x s32>)
+    ; CHECK-NEXT: PseudoRET implicit $lr, implicit $x0
+    %0:_(s32) = G_CONSTANT i32 28
```
konstantinschwarz: Nit: unused G_CONSTANTs here and in the tests below

llvm/test/CodeGen/AIE/aie2/GlobalISel/prelegalizercombiner-shufflevector-buildvector.mir
```diff
@@ -0,0 +1,322 @@
...
+    ; CHECK-NEXT: {{  $}}
+    ; CHECK-NEXT: [[COPY:%[0-9]+]]:_(s32) = COPY $r0
+    ; CHECK-NEXT: [[BUILD_VECTOR:%[0-9]+]]:_(<8 x s32>) = G_BUILD_VECTOR [[COPY]](s32), [[COPY]](s32), [[COPY]](s32), [[COPY]](s32), [[COPY]](s32), [[COPY]](s32), [[COPY]](s32), [[COPY]](s32)
+    ; CHECK-NEXT: PseudoRET implicit $lr, implicit [[BUILD_VECTOR]](<8 x s32>)
+    %1:_(s32) = COPY $r0
+    %3:_(<8 x s32>) = G_IMPLICIT_DEF
+    %5:_(s32) = G_IMPLICIT_DEF
+    %2:_(<8 x s32>) = G_BUILD_VECTOR %1(s32), %5(s32), %5(s32), %5(s32), %5(s32), %5(s32), %5(s32), %5(s32)
+    %0:_(<8 x s32>) = G_SHUFFLE_VECTOR %2(<8 x s32>), %3, shufflemask(0, 0, 0, 0, 0, 0, 0, 0)
+    PseudoRET implicit $lr, implicit %0
```
konstantinschwarz, line 419: Looks like this is missing the closing `...` at the end, and then I think this would fail, because this is further combined into `G_AIE_BROADCAST`?

llvm/test/CodeGen/AIE/aie2/GlobalISel/prelegalizercombiner-shufflevector-buildvector.mir
```diff
@@ -0,0 +1,322 @@
+# NOTE: Assertions have been autogenerated by utils/update_mir_test_checks.py
+#
+# This file is licensed under the Apache License v2.0 with LLVM Exceptions.
+# See https://llvm.org/LICENSE.txt for license information.
+# SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+#
+# (c) Copyright 2023-2024 Advanced Micro Devices, Inc. or its affiliates
```
konstantinschwarz: nit: just 2024

konstantinschwarz (commented):
Mostly nits, combiner looks good to me

llvm/lib/CodeGen/GlobalISel/CombinerHelper.cpp
```diff
@@ -2147,6 +2149,77 @@ static Register peekThroughBitcast(Register Reg,
...
+    return false;
+
```
niwinanto, line 2293: Nit: If both are IsUndef you could return early.
```
if (IsUndef1|| IsUndef2)
  return true;
```

llvm/include/llvm/CodeGen/GlobalISel/GenericMachineInstrs.h
ValentijnvdBeek: I can see that argument. I was looking at it from the perspective of what the result of the function would be. I'll separate them, thanks.

llvm/lib/CodeGen/GlobalISel/CombinerHelper.cpp
ValentijnvdBeek: Sure, will do. I personally prefer explicit NULL pointer checking for clarity (which, to be fair, this isn't yet), but doing it like that is fine with me.

llvm/test/CodeGen/AIE/aie2/GlobalISel/prelegalizercombiner-shufflevector-buildvector.mir
ValentijnvdBeek: Fair, thanks Konstantin.

llvm/test/CodeGen/AIE/aie2/GlobalISel/prelegalizercombiner-shufflevector-buildvector.mir
ValentijnvdBeek: The shufflevector branch has become really divergent in the past month or so, so that hasn't landed yet. Let's try to get all of that merged as quickly as we can, I'll manually merge them to make sure that spectests are updated before enter the main branch.

Missing the closing `...` is deliberate, trailing ellipses have caused errors with the unit tests scripts for me before.

ValentijnvdBeek: Ran into it again just now. If you have a trailing ellipsis, you need a trailing newline, or otherwise the update mir script will error out. If you skip it, then it doesn't matter whether you have a newline or not. That is why I tend to commit mine without the trailing ellipsis.

llvm/lib/Target/AArch64/AArch64Combine.td
```diff
@@ -295,5 +295,5 @@ def AArch64PostLegalizerCombiner
                         ptr_add_immed_chain, overlapping_and,
                         split_store_zero_128, undef_combines,
                         select_to_minmax, or_to_bsp, combine_concat_vector,
-                        commute_constant_to_rhs]> {
+                        commute_constant_to_rhs, shufflevector_merge]> {
 }
```
ValentijnvdBeek, line 299: This is needed since some ARM64 tests relies on the legalizer changing the inputs of the shufflevector and if you don't run them afterwards you get worse code.

llvm/test/CodeGen/AArch64/GlobalISel/legalize-shuffle-vector-widen-crash.ll
```diff
@@ -8,13 +8,18 @@ declare i32 @llvm.aarch64.neon.uaddv.i32.v4i32(<4 x i32>) #0
...
 ; CHECK-LABEL: bar:
 ; CHECK:       ; %bb.0: ; %bb
-; CHECK-NEXT:    movi.2d v0, #0000000000000000
-; CHECK-NEXT:    mov b1, v0[1]
-; CHECK-NEXT:    mov b2, v0[2]
-; CHECK-NEXT:    mov b3, v0[3]
-; CHECK-NEXT:    mov.h v0[1], v1[0]
+; CHECK-NEXT:    mov w8, #0 ; =0x0
+; CHECK-NEXT:    fmov s0, w8
+; CHECK-NEXT:    mov.16b v1, v0
```
ValentijnvdBeek, line 13: Add copyright header!

llvm/include/llvm/Target/GlobalISel/Combine.td
```diff
@@ -765,6 +765,16 @@ def unmerge_merge : GICombineRule<
   (apply [{ Helper.applyCombineUnmergeMergeToPlainValues(*${d}, ${info}); }])
 >;
 
+// Fold (unmerge(merge x, y, z)) -> z, y, z.
+def shufflevector_merge_matchinfo : GIDefMatchData<"SmallVector<Register, 8>">;
+def shufflevector_merge : GICombineRule<
+  (defs root:$d, shufflevector_merge_matchinfo:$info),
+  (match (wip_match_opcode G_SHUFFLE_VECTOR): $d,
```
ValentijnvdBeek, line 772: A recent move in LLVM is that you shouldn't use `wip_match_opcode` anymore since it slows down compilation. 

https://llvm.org/docs/GlobalISel/MIRPatterns.html#gallery

martien-de-jong:
Overtaken by other combiners

PR #133, opened 2024-07-22
[AIEX] Basic LICM pass for control registers

gbossu:
This is reviving some older work I made to hoist assignments to CR registers out of loops. This can both hoist to the pre-header, or sink to the exit block. This is needed to clean up our loops so they can later be SW pipelined. QoR shows minor improvements, and one regression: Add2D. What happens is that we now SW-pipeline the inner loop, but spill vector registers.  In a future PR, I'll address those spills by teaching the pre-RA scheduler to try even harder to limit register pressure in cases where it is especially high.

E.g.
```
loop:
  $cr = MOV ...
  USE $cr
  $cr = MOV 0
  JNZ ... loop
```
into:
```
  $cr = MOV ...
loop:
  USE $cr
  JNZ ... loop
exit:
  $cr = MOV 0
```

Limitations:
 - Only supports simple one-bb loops.


llvm/lib/Target/AIE/ReservedRegsLICM.cpp
```diff
@@ -0,0 +1,330 @@
...
+#include "llvm/CodeGen/LivePhysRegs.h"
+#include "llvm/CodeGen/MachineFunctionPass.h"
+#include "llvm/CodeGen/MachineLoopInfo.h"
+#include "llvm/CodeGen/MachineRegisterInfo.h"
+#include "llvm/InitializePasses.h"
+#include "llvm/Support/Debug.h"
+
+using namespace llvm;
+
+#define DEBUG_TYPE "reserved-reg-licm"
```
andcarminati, line 29: nit: `aie-reserved-reg-licm` because this is also the name of the pass.

llvm/lib/Target/AIE/ReservedRegsLICM.cpp
```diff
@@ -0,0 +1,330 @@
...
+    runOnLoop(*L);
+  }
+
+  return Changed;
+}
+
+static MCRegister getSinglePhysRegDef(const MachineInstr &MI) {
+  if (MI.getNumDefs() == 1 && MI.getNumExplicitDefs() == 1) {
+    const MachineOperand &MO = MI.getOperand(0);
+    return MO.isReg() && MO.getReg().isPhysical() ? MO.getReg().asMCReg()
```
andcarminati: Is there any possibility that MO is not a register? I think your first If guarantees this condition.

I think it that it can be simpler:
```
if (MO.getReg().isPhysical()) 
    return MO.getReg().asMCReg();
```

llvm/lib/Target/AIE/ReservedRegsLICM.cpp
```diff
@@ -0,0 +1,330 @@
...
+
+  // Conservativaly assume reserved reserved regs are all liveouts
+  for (MCPhysReg PhysReg : MRI->getReservedRegs().set_bits()) {
+    if (MRI->canSimplifyPhysReg(PhysReg)) {
+      LiveRegs.addReg(PhysReg);
+    }
+  }
+
+  // Traverse instructions to remove defs
+  for (const MachineInstr &MI : reverse(Header))
```
andcarminati, line 222: ~~I was thinking about one case that one register is defined by a block that dominates the Header. I think this case is not captured here, this register will not be in the live in set.~~

~~Can we just trust `LoopBody->liveins()`? set~~

Unserstood now: `Note we do not need to extend the liveins of the loops BBs with the
  // newly-hoisted def, because reserved regs are always considered live.`

llvm/lib/Target/AIE/ReservedRegsLICM.cpp
```diff
@@ -0,0 +1,330 @@
...
+
+  /// Go through \p L to look for instructions to hoist in the preheader or
+  /// exit block.
+  void runOnLoop(MachineLoop &L);
+
+  /// Find instructions to hoist to the exit block
+  void processForExitHoist(MachineLoop &L, const BitVector &ReservedLiveins);
+
+  /// Hoist \p Cand to \p L's exit block it it is safe to do so.
+  bool tryHoistToExitBlock(const CandidateInfo &Cand, MachineLoop &L);
```
andcarminati: Maybe `Sink` instead of `Hoist`

llvm/lib/Target/AIE/ReservedRegsLICM.cpp
```diff
@@ -0,0 +1,330 @@
...
+    MCPhysReg DefinedReg = MCRegister::NoRegister;
+    MachineInstr *HoistCandidate = nullptr;
+  };
+  DenseMap<MCPhysReg, CandidateInfo> Candidates;
+  CandidateInfo *getCandidateInfo(const MachineInstr &MI);
+
+  /// Collect all simplifiable reserved registers that all livein for \p L.
+  BitVector collectLoopReservedLiveins(const MachineLoop &L);
+
+  /// Go through \p L to look for instructions to hoist in the preheader or
```
martien-de-jong: nit: hoist into the preheader or sink into the exit block

llvm/lib/Target/AIE/ReservedRegsLICM.cpp
```diff
@@ -0,0 +1,330 @@
...
+  /// exit block.
+  void runOnLoop(MachineLoop &L);
+
+  /// Find instructions to hoist to the exit block
+  void processForExitHoist(MachineLoop &L, const BitVector &ReservedLiveins);
+
+  /// Hoist \p Cand to \p L's exit block it it is safe to do so.
```
martien-de-jong: nit *if* it is

llvm/lib/Target/AIE/ReservedRegsLICM.cpp
```diff
@@ -0,0 +1,330 @@
...
+  bool tryHoistToExitBlock(const CandidateInfo &Cand, MachineLoop &L);
+
+  /// Hoist \p Cand to \p L's preheader it it is safe to do so.
```
martien-de-jong: And one more

llvm/lib/Target/AIE/ReservedRegsLICM.cpp
```diff
@@ -0,0 +1,330 @@
...
+BitVector ReservedRegsLICM::collectLoopReservedLiveins(const MachineLoop &L) {
+  const MachineBasicBlock &Header = *L.getHeader();
+  LivePhysRegs LiveRegs(*TRI);
+
+  // Conservativaly assume reserved reserved regs are all liveouts
+  for (MCPhysReg PhysReg : MRI->getReservedRegs().set_bits()) {
+    if (MRI->canSimplifyPhysReg(PhysReg)) {
+      LiveRegs.addReg(PhysReg);
+    }
+  }
```
martien-de-jong, line 219: I would make this function take a block, not a loop, to make it easier to reuse in multi-block loops. And perhaps split off livereg initialization, which would be different for exiting blocks and internal blocks.

llvm/lib/Target/AIE/ReservedRegsLICM.cpp
```diff
@@ -0,0 +1,330 @@
...
+  TRI = ST.getRegisterInfo();
+  MRI = &MF.getRegInfo();
+  assert(MRI->isSSA() && "PhysRegCopyTracker can only track SSA copies");
+
+  LLVM_DEBUG(dbgs() << "******** Reserved register LICM: " << MF.getName()
+                    << " ********\n");
+
+  MachineLoopInfo &MLI = getAnalysis<MachineLoopInfo>();
+
+  for (MachineLoop *L : MLI.getBase()) {
```
martien-de-jong: I don't think this will visit nested loops. Normally runOnLoop recurses

llvm/test/CodeGen/AIE/aie2/GlobalISel/post-select-licm-reserved.mir
```diff
@@ -0,0 +1,209 @@
...
+    PseudoRET implicit $lr
+
+  bb.2:
+    successors: %bb.2(0x7c000000), %bb.1(0x04000000)
+    liveins: $p0, $r0, $r1, $r2, $x0
+    VST_PACK_D8_D16_ag_idx_imm %1, 0, %3, implicit $crsat, implicit $crpacksign
+    $crpacksign = MOV_scalar_imm10_pseudo 0
+    PseudoJNZ %0, %bb.2
+    PseudoJ_jump_imm %bb.1
+...
```
martien-de-jong, line 387: perhaps add an example with a nested loop

llvm/lib/Target/AIE/ReservedRegsLICM.cpp
gbossu: I remember in my initial work I extended the pass to support multi-BB loops. Typically outer loops. It was enough to only look in the header, and be conservative if a reserved register was not met there. If I need something more complicated, I'd probably scrap this altogether and extend`AIELiveRegs.cpp`

llvm/lib/Target/AIE/ReservedRegsLICM.cpp
gbossu: Now fixed!

llvm/test/CodeGen/AIE/aie2/GlobalISel/post-select-licm-reserved.mir
gbossu: Added, great catch :)

llvm/lib/Target/AIE/ReservedRegsLICM.cpp
```diff
@@ -0,0 +1,381 @@
...
+  BitVector ReservedLiveins(TRI->getNumRegs());
+  for (MCRegister Reg : LiveRegs) {
+    if (MRI->isReserved(Reg)) {
+      ReservedLiveins.set(Reg);
+    }
+  }
+  return ReservedLiveins;
+}
+
+/// Walk the specified region of the CFG and hoist loop invariants out to the
```
andcarminati, line 234: nit: update this comment wit hoist and sink.

llvm/lib/Target/AIE/ReservedRegsLICM.cpp
```diff
@@ -0,0 +1,381 @@
...
+
+  void getAnalysisUsage(AnalysisUsage &AU) const override {
+    AU.addRequired<MachineLoopInfo>();
+    AU.addPreserved<MachineLoopInfo>();
+    MachineFunctionPass::getAnalysisUsage(AU);
+  }
+
+private:
+  /// Keep track of information about hoisting candidates.
+  struct CandidateInfo {
```
andcarminati: nit: you can encapsulate all Candidate-related stuff in class. Something like this:

```
  struct Candidates {
    struct CandidateInfo {
      CandidateInfo(MCPhysReg DefinedReg) : DefinedReg(DefinedReg) {}
      MCPhysReg DefinedReg = MCRegister::NoRegister;
      MachineInstr *HoistOrSinkCandidate = nullptr;
    };
    const MachineRegisterInfo* MRI;
    DenseMap<MCPhysReg, CandidateInfo> AllCandidates;
    Candidates(const MachineRegisterInfo* MRI) : MRI(MRI) {}
    CandidateInfo *getCandidateInfo(const MachineInstr &MI);
  };
  using CandidateInfo = Candidates::CandidateInfo;
```
Also, you can use this as a local object in Hoist/Sink methods, not as an attribute that you need to clean before the second use....

llvm/lib/Target/AIE/ReservedRegsLICM.cpp
```diff
@@ -0,0 +1,381 @@
...
+      ReservedLiveins.set(Reg);
+    }
+  }
+  return ReservedLiveins;
+}
+
+/// Walk the specified region of the CFG and hoist loop invariants out to the
+/// preheader.
+void ReservedRegsLICM::runOnLoop(MachineLoop &L) {
+  MachineBasicBlock *Preheader = L.getLoopPreheader();
```
andcarminati: nit: `const MachineBasicBlock.....`

llvm/lib/Target/AIE/ReservedRegsLICM.cpp
```diff
@@ -0,0 +1,381 @@
...
+    Changed |= trySinkToExitBlock(CandInfo, L);
+  }
+}
+
+bool ReservedRegsLICM::trySinkToExitBlock(const CandidateInfo &Cand,
+                                          MachineLoop &L) {
+  if (isLoopInvariantInst(Cand, L)) {
+    MachineBasicBlock *InsertMBB = L.getExitBlock();
+    assert(InsertMBB);
+    hoist(Cand, InsertMBB->getFirstNonPHI(), *InsertMBB);
```
andcarminati: As we are in `trySinkToExitBlock` it is a good idea to rename `hoist` to a different name, like `hoistOrSink` os just `moveInstruction`. You ca also pass directly the instruction instead of Cand.

llvm/lib/Target/AIE/ReservedRegsLICM.cpp
```diff
@@ -0,0 +1,381 @@
...
+  return false;
+}
+
+/// When an instruction is found to only use loop invariant operands that is
+/// safe to hoist, this instruction is called to do the dirty work.
+void ReservedRegsLICM::hoist(const CandidateInfo &Cand,
+                             MachineBasicBlock::iterator InsertBefore,
+                             MachineBasicBlock &InsertMBB) {
+  MachineInstr &MI = *Cand.HoistCandidate;
+  LLVM_DEBUG(dbgs() << "Hoisting to " << printMBBReference(InsertMBB)
```
andcarminati: nit: "Moving to"

llvm/lib/Target/AIE/ReservedRegsLICM.cpp
gbossu: Agreed, that is always better to avoid keeping too much state ;)

gbossu:
I think I have addressed all the comments now :)

llvm/lib/Target/AIE/ReservedRegsLICM.cpp
```diff
@@ -0,0 +1,387 @@
...
+    }
+  }
+  return ReservedLiveins;
+}
+
+/// Walk the specified region of the CFG and hoist loop invariants out to the
+/// preheader.
+void ReservedRegsLICM::runOnLoop(MachineLoop &L) {
+  const MachineBasicBlock *Preheader = L.getLoopPreheader();
+  const MachineBasicBlock *ExitBlock = L.getExitBlock();
```
martien-de-jong, line 238: Shouldn't you check that it only has one predecessor?

llvm/lib/Target/AIE/ReservedRegsLICM.cpp
```diff
@@ -0,0 +1,387 @@
...
+  MRI = &MF.getRegInfo();
+  assert(MRI->isSSA() && "ReservedRegsLICM can only deal with SSA vregs");
+
+  LLVM_DEBUG(dbgs() << "******** Reserved register LICM: " << MF.getName()
+                    << " ********\n");
+
+  MachineLoopInfo &MLI = getAnalysis<MachineLoopInfo>();
+
+  SmallVector<MachineLoop *, 4> Loops = MLI.getBase().getLoopsInPreorder();
+  for (MachineLoop *L : reverse(Loops)) {
```
martien-de-jong, line 203: nit: we only do leaf loops. reverse is not really necessary.

martien-de-jong (dismissed):
I'm not sure whether you check for dedicated exit to hoist into. For the rest it looks fine.

llvm/lib/Target/AIE/ReservedRegsLICM.cpp
gbossu: We do check all loops now, `getLoopsInPreorder` will find nested ones as well.

llvm/lib/Target/AIE/ReservedRegsLICM.cpp
gbossu: Definitely... 🙈 I have a local fix that will create a dedicated exit when required. Checking QoR now.

martien-de-jong (approved):

andcarminati (approved):
LGTM. Nice work!

PR #136, opened 2024-07-26
[AIE2] Simplify Intrinsics with demanded bits VBCST

abhinay-anubola:
Eliminating unnecessary sign-extensions for intrinsics that access sub-words of a 32-bit register

### DIFF
--- a/clang/test/CodeGen/aie/aie2/aie2-scl2vec-intrinsic.cpp
+++ b/clang/test/CodeGen/aie/aie2/aie2-scl2vec-intrinsic.cpp
@@ -435,7 +435,7 @@ v32uint16 test_insert(v32uint16 v, int idx, unsigned long long b) {
 
 // CHECK-LABEL: @_Z18test_broadcast_s16s(
 // CHECK-NEXT:  entry:
-// CHECK-NEXT:    [[CONV_I:%.*]] = sext i16 [[B:%.*]] to i32
+// CHECK-NEXT:    [[CONV_I:%.*]] = zext i16 [[B:%.*]] to i32
 // CHECK-NEXT:    [[TMP0:%.*]] = tail call noundef <32 x i16> @llvm.aie2.vbroadcast16.I512(i32 [[CONV_I]])
 // CHECK-NEXT:    ret <32 x i16> [[TMP0]]
 //
@@ -473,7 +473,7 @@ v16int32 test_broadcast_v2s32 (v2int32 b) {
 
 // CHECK-LABEL: @_Z26test_broadcast_to_v128int4DB8_(
 // CHECK-NEXT:  entry:
-// CHECK-NEXT:    [[CONV_I:%.*]] = sext i8 [[B:%.*]] to i32
+// CHECK-NEXT:    [[CONV_I:%.*]] = zext i8 [[B:%.*]] to i32
 // CHECK-NEXT:    [[TMP0:%.*]] = tail call noundef <64 x i8> @llvm.aie2.vbroadcast8.I512(i32 [[CONV_I]])
 // CHECK-NEXT:    ret <64 x i8> [[TMP0]]
 //
@@ -484,7 +484,7 @@ v128int4 test_broadcast_to_v128int4 (v2int4 b) {
 // CHECK-LABEL: @_Z25test_broadcast_to_v64int8Dv2_a(
 // CHECK-NEXT:  entry:
 // CHECK-NEXT:    [[TMP0:%.*]] = bitcast <2 x i8> [[B:%.*]] to i16
-// CHECK-NEXT:    [[CONV_I:%.*]] = sext i16 [[TMP0]] to i32
+// CHECK-NEXT:    [[CONV_I:%.*]] = zext i16 [[TMP0]] to i32
 // CHECK-NEXT:    [[TMP1:%.*]] = tail call <32 x i16> @llvm.aie2.vbroadcast16.I512(i32 [[CONV_I]])
 // CHECK-NEXT:    [[TMP2:%.*]] = bitcast <32 x i16> [[TMP1]] to <64 x i8>
 // CHECK-NEXT:    ret <64 x i8> [[TMP2]]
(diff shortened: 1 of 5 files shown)


clang/test/CodeGen/aie/aie2/aie2-scl2vec-intrinsic.cpp
```diff
@@ -435,7 +435,7 @@ v32uint16 test_insert(v32uint16 v, int idx, unsigned long long b) {
 
 // CHECK-LABEL: @_Z18test_broadcast_s16s(
 // CHECK-NEXT:  entry:
-// CHECK-NEXT:    [[CONV_I:%.*]] = sext i16 [[B:%.*]] to i32
+// CHECK-NEXT:    [[CONV_I:%.*]] = zext i16 [[B:%.*]] to i32
```
konstantinschwarz, line 438: Do you know why this changed to `zext`?
Since the parameter is a signed type, those are already sign extended (ABI requirement).
This change requires an explicit zero extension, although our intrinsic only reads the lower 16 bits.
Changing this to `anyext` would be ok, but zext seems wrong to me.

abhinay-anubola: In `SimplifyDemandedUseBits` a comment says that "if the NewBits are not demanded convert this into a zero extension." here "this" mean SExt. And also gaetan said "anyext does not exist in the middle end". So, maybe we need to let the SExt be, as this gets optimized later.

martien-de-jong: zext seems to be preferred by demandedBits, probably because it yields a definite zero value for the upper bits.
Since the conversion is necessary to make the types sane, I think the conversion should be swallowed by instruction selection.

andcarminati: Hi, with the change:
```
	nopb	;		nopa	;		nops	;		ret lr	;		nopm	;		nopv	
	nopx	                                //  Delay Slot 5
	nop	                                //  Delay Slot 4
	extend.u16	r0, r0                  //  Delay Slot 3
	vbcst.16	x0, r0                          //  Delay Slot 2
```

Without:
```
	ret lr	
	nop	                                //  Delay Slot 5
	nop	                                //  Delay Slot 4
	nop	                                //  Delay Slot 3
	vbcst.16	x0, r0                          //  Delay Slot 2
```

It seems that without we get rid of sext anyway.....

abhinay-anubola: Added a pattern for broadcast, now zext is getting eliminated too.

andcarminati: Do you have some QoR results for this change?

llvm/lib/Target/AIE/AIE2InstrPatterns.td
```diff
@@ -779,7 +779,9 @@ def : VEQZ_INSTR<int_aie2_veqz32, VEQZ_32>;
 
 // VBCST
 def : Pat<(int_aie2_vbroadcast8_I512 eR:$s0),  (VBCST_8 eR:$s0)>;
+def : Pat<(int_aie2_vbroadcast8_I512 (and eR:$rs1, 255)), (VBCST_8 eR:$rs1)>;
 def : Pat<(int_aie2_vbroadcast16_I512 eR:$s0), (VBCST_16 eR:$s0)>;
+def : Pat<(int_aie2_vbroadcast16_I512 (and eR:$rs1, 65535)), (VBCST_16 eR:$rs1)>;
```
gbossu, line 781: Do we still need the DemandedBits then? It does not bother me to keep the `zext`, but I get @martien-de-jong and @konstantinschwarz preference for keeping the semantics visible.

clang/test/CodeGen/aie/aie2/aie2-scl2vec-intrinsic.cpp
```diff
@@ -556,9 +556,7 @@ v16float test_broadcast_float(float b) {
 
 // CHECK-LABEL: @_Z16test_shiftl_elemDv64_ai(
 // CHECK-NEXT:  entry:
-// CHECK-NEXT:    [[SEXT:%.*]] = shl i32 [[S:%.*]], 24
-// CHECK-NEXT:    [[CONV_I1:%.*]] = ashr exact i32 [[SEXT]], 24
-// CHECK-NEXT:    [[TMP0:%.*]] = tail call noundef <64 x i8> @llvm.aie2.vbroadcast8.I512(i32 [[CONV_I1]])
+// CHECK-NEXT:    [[TMP0:%.*]] = tail call noundef <64 x i8> @llvm.aie2.vbroadcast8.I512(i32 [[S:%.*]])
 // CHECK-NEXT:    [[TMP1:%.*]] = bitcast <64 x i8> [[V:%.*]] to <16 x i32>
```
abhinay-anubola, line 560: @gbossu Without DemandedBits these instructions wont be eliminated

gbossu: Let's see with @konstantinschwarz if he has some concerns with `zext`, to me the code looks correct.

llvm/test/CodeGen/AIE/aie2/GlobalISel/inst-select-vbcst.mir
```diff
@@ -23,7 +23,8 @@ body:             |
     ; CHECK-NEXT: $x0 = COPY [[VBCST_8_]]
     ; CHECK-NEXT: PseudoRET implicit $lr, implicit $x0
       %2:gprregbank(s32) = COPY $r0
-      %3:gprregbank(s32) = G_ASSERT_SEXT %2:gprregbank, 8
+      %4:gprregbank(s32) = G_CONSTANT i32 255
+      %3:gprregbank(s32) = G_AND %2, %4
```
konstantinschwarz, line 49: We should keep these tests and add two additional ones for the new patterns added by your PR

clang/test/CodeGen/aie/aie2/aie2-scl2vec-intrinsic.cpp
konstantinschwarz: I think its fine. Having cleaner LLVM IR potentially allows other optimizations to kick in.
The additional instruction selection patterns don't hurt and cover the cases where some extensions still survive the IR optimizers.

llvm/lib/Target/AIE/AIE2InstrPatterns.td
konstantinschwarz: Strictly speaking no, but cleaner IR might unlock other optimizations.
However these patterns only cover zero extensions. Maybe add patterns for sign-extensions as well?

konstantinschwarz (approved):

PR #137, opened 2024-07-26
Update aie-public to upstream commit 821935bb55449f708f822d0e381164ebbc33483c

konstantinschwarz:


jsetoain:
@konstantinschwarz Is there a reason why the version bump is so old? It looks like that commit landed upstream in April.

konstantinschwarz:
> @konstantinschwarz Is there a reason why the version bump is so old? It looks like that commit landed upstream in April.

I'm doing an incremental upgrade to avoid having too many conflicts and fixup commits at once.
Previous bump brought us from end of December to end of of February, this one takes us to end of April.
I plan to do 1-2 more bumps to get us to tip, then hopefully update ~every two weeks

jsetoain:
> > @konstantinschwarz Is there a reason why the version bump is so old? It looks like that commit landed upstream in April.
> 
> I'm doing an incremental upgrade to avoid having too many conflicts and fixup commits at once. Previous bump brought us from end of December to end of of February, this one takes us to end of April. I plan to do 1-2 more bumps to get us to tip, then hopefully update ~every two weeks

I thought it might be something like this, I just wanted to make sure there wasn't any massively breaking change upstream that prevented us from going beyond this particular commit 😄 Thanks for the clarification!

PR #139, opened 2024-07-30
[RFC][AIE2][AA] Disambiguate pointers updated by intrinsics

andcarminati:
This is RFC. Soon we will need more AA support, and this PR tries to clear a bit the way.

Now we can disambiguate some cases where aie2_add_2d and aie2_add_3d are used to update pointers.

No QoR failures. 
Gains for Reduce* family up to 50% (instruction count). 
Average gain is ~2% (~350 benchmarks). 

Updated to include post-RA AA support.

### DIFF
--- a/llvm/lib/Target/AIE/AIEBaseAliasAnalysis.cpp
+++ b/llvm/lib/Target/AIE/AIEBaseAliasAnalysis.cpp
@@ -18,6 +18,7 @@
 #include "llvm/Analysis/AliasAnalysis.h"
 #include "llvm/Analysis/MemoryLocation.h"
 #include "llvm/Analysis/ValueTracking.h"
+#include "llvm/CodeGen/MachineInstr.h"
 #include "llvm/IR/Constants.h"
 #include "llvm/IR/Instructions.h"
 #include "llvm/IR/IntrinsicsAIE2.h"
@@ -241,6 +242,10 @@ struct IntrinsicChainInfo {
 
     return false;
   }
+
+  void bumpIteration(const unsigned VirtIteration) {
+    IntrinsicCalls += VirtIteration * TotalIntrinsicCalls;
+  }
 };
 
 static IntrinsicArgsInfo collectIntrinsicOperands(const Value *V) {
@@ -369,7 +374,9 @@ trackIntrinsicChain(const Value *LastUpdate, const Value *TargetPointer) {
 }
 
 static AliasResult aliasAIEIntrinsic(const Value *ValueA, const Value *ValueB,
-                                     const Value *BaseA, const Value *BaseB) {
+                                     const Value *BaseA, const Value *BaseB,
+                                     const unsigned VirtUnrollLevelA,
+                                     const unsigned VirtUnrollLevelB) {
 
   const PHINode *PhiA = dyn_cast<PHINode>(BaseA);
   const PHINode *PhiB = dyn_cast<PHINode>(BaseB);
@@ -449,6 +456,14 @@ static AliasResult aliasAIEIntrinsic(const Value *ValueA, const Value *ValueB,
   if (!BTracking) // Non uniform address update accross the loop.
     return AliasResult::MayAlias;
 
+  // If we are in PostRA SWP, we need to increase the count
+  // by the number of unrolls x complete number of pointer
+  // updates for every completed iteration. In this way,
+  // we will have the notion of distance between different
+  // iterations and associated instructions.
+  ATracking->bumpIteration(VirtUnrollLevelA);
+  BTracking->bumpIteration(VirtUnrollLevelB);
+
   // If there is no overlap, NoAlias
   if (ATracking->mayOverlap(*BTracking))
     return AliasResult::MayAlias;
@@ -464,8 +479,8 @@ AliasResult AIEBaseAAResult::alias(const MemoryLocation &LocA,
   const Value *BaseB = getUnderlyingObjectAIE(LocB.Ptr);
 
   if (DisambiguateAccessSameOriginPointers &&
-      aliasAIEIntrinsic(LocA.Ptr, LocB.Ptr, BaseA, BaseB) ==
-          AliasResult::NoAlias)
+      aliasAIEIntrinsic(LocA.Ptr, LocB.Ptr, BaseA, BaseB, 0, 0
+                        /*No virtually unrolled*/) == AliasResult::NoAlias)
     return AliasResult::NoAlias;
 
   // Our look-through didn't reveal a different object, pass on to the next
@@ -484,4 +499,47 @@ AliasResult AIEBaseAAResult::alias(const MemoryLocation &LocA,
   // NoAlias (based on different base object). In particular, we cannot infer
   // MustAlias, and need to be conservative by returning MayAlias
   return AliasResult::MayAlias;
+}
+
+AliasResult AIE::aliasAcrossVirtualUnrolls(const MachineInstr *MIA,
+                                           const MachineInstr *MIB,
+                                           unsigned UnrollLevelMIA,
+                                           unsigned UnrollLevelMIB) {
+
+  auto GetValueBasePair = [](const MachineMemOperand *MMO)
+      -> std::optional<std::pair<const Value *, const Value *>> {
+    const Value *Val = MMO->getValue();
+    // We know nothing about the MMO.
+    if (!Val)
+      return std::nullopt;
+
+    const Value *Base = getUnderlyingObjectAIE(Val);
+    if (!Base)
+      return std::nullopt;
+
+    return std::make_pair(Val, Base);
+  };
+
+  // Check each pair of memory operands from both instructions.
+  for (auto *MMOA : MIA->memoperands()) {
+
+    auto ValueBasePairA = GetValueBasePair(MMOA);
+    if (!ValueBasePairA)
+      return AliasResult::MayAlias;
+
+    for (auto *MMOB : MIB->memoperands()) {
+
+      auto ValueBasePairB = GetValueBasePair(MMOB);
+      if (!ValueBasePairB)
+        return AliasResult::MayAlias;
+
+      if (aliasAIEIntrinsic(ValueBasePairA->first, ValueBasePairB->first,
+                            ValueBasePairA->second, ValueBasePairB->second,
+                            UnrollLevelMIA,
+                            UnrollLevelMIB) == AliasResult::MayAlias) {
+        return AliasResult::MayAlias;
+      }
+    }
+  }
+  return AliasResult::NoAlias;
 }
--- a/llvm/lib/Target/AIE/AIEBaseAliasAnalysis.cpp
+++ b/llvm/lib/Target/AIE/AIEBaseAliasAnalysis.cpp
@@ -24,6 +24,7 @@
 #include "llvm/Support/Casting.h"
 #include "llvm/Support/CommandLine.h"
 #include "llvm/Support/Debug.h"
+#include <optional>
 
 using namespace llvm;
 
@@ -31,6 +32,11 @@ static cl::opt<unsigned> BaseObjectSearchLimit(
     "aie-alias-analysis-search-limit",
     cl::desc("Search depth limit for base objects in AIE alias analysis"),
     cl::init(100), cl::Hidden);
+
+static cl::opt<bool> DisambiguateAccessSameOriginPointers(
+    "aie-alias-analysis-disambiguation",
+    cl::desc("Disambiguate pointers derived from the same origin"),
+    cl::init(true), cl::Hidden);
 
 #define DEBUG_TYPE "aie-aa"
 
@@ -77,6 +83,36 @@ static bool isAIEPtrAddIntrinsic(Intrinsic::ID ID, unsigned &InPtrIdx) {
   return false;
 }
 
+/// This gives all indexes for read only operands ie.
+/// excluding counters.
+static SmallVector<unsigned, 5> getAddIntrinsicOps(Intrinsic::ID ID) {
+  switch (ID) {
+  case Intrinsic::aie2_add_2d:
+    return {1, 2, 3 /*,count*/};
+  case Intrinsic::aie2_add_3d:
+    return {1, 2, 3, 4, /*,count1*/ 6,
+            /*,count2*/};
+  default:
+    llvm_unreachable("Unknown intrinsic");
+  }
+
+  return {};
+}
+
+/// This gives all indexes counter operands.
+static SmallVector<unsigned, 5> getAddIntrinsicCounterOps(Intrinsic::ID ID) {
+  switch (ID) {
+  case Intrinsic::aie2_add_2d:
+    return {4};
+  case Intrinsic::aie2_add_3d:
+    return {5, 7};
+  default:
+    llvm_unreachable("Unknown intrinsic");
+  }
+
+  return {};
+}
+
 static const Value *lookThroughAIEAddressingModes(const Value *V) {
   if (const auto *Extract = dyn_cast<ExtractValueInst>(V)) {
     if (const auto *MaybeIntrinsicCall =
@@ -117,12 +153,320 @@ static const Value *getUnderlyingObjectAIE(const Value *V) {
   return V;
 }
 
+static const Value *skipCast(const Value *V) {
+  if (const Instruction *Instr = dyn_cast<const Instruction>(V)) {
+    if (Instr->isCast())
+      return Instr->getOperand(0);
+  }
+  return V;
+}
+
+struct IntrinsicArgsInfo {
+  SmallVector<const Value *, 5> ConstantArgs;
+  SmallVector<const Value *, 5> CounterArgs;
+};
+
+struct IntrinsicChainInfo {
+  // Calls up to a specific point.
+  unsigned IntrinsicCalls = 0;
+  // Total calls across the loop.
+  unsigned TotalIntrinsicCalls = 0;
+  // All parameters.
+  IntrinsicArgsInfo IntrinsicParameters;
+
+  // Same number of call until the point of interest.
+  bool hasSameNumberOfCalls(const IntrinsicChainInfo &Other) const {
+    return IntrinsicCalls == Other.IntrinsicCalls;
+  }
+
+  // Same number of calls accross the loop.
+  bool hasSameTotalNumberOfCalls(const IntrinsicChainInfo &Other) const {
+    return TotalIntrinsicCalls == Other.TotalIntrinsicCalls;
+  }
+
+  // Check for the same constant parameters.
+  bool hasSameParameters(const IntrinsicChainInfo &Other) const {
+    return IntrinsicParameters.ConstantArgs ==
+           Other.IntrinsicParameters.ConstantArgs;
+  }
+
+  // Check for the same initial values for the counters.
+  bool hasSameStartingCounters(const IntrinsicChainInfo &Other) const {
+
+    auto GetExternalIndex = [](const PHINode *Phi) {
+      return Phi->getIncomingBlock(0) == Phi->getParent() ? 1 : 0;
+    };
+
+    for (const Value *CountA : IntrinsicParameters.CounterArgs) {
+      // Some truncation can be expected.
+      CountA = skipCast(CountA);
+      if (auto *PhiA = dyn_cast<PHINode>(CountA)) {
+        const Value *ExtPhiValA =
+            PhiA->getIncomingValue(GetExternalIndex(PhiA));
+        for (const Value *CountB : Other.IntrinsicParameters.CounterArgs) {
+          CountB = skipCast(CountB);
+          if (auto *PhiB = dyn_cast<PHINode>(CountB)) {
+            const Value *ExtPhiValB =
+                PhiB->getIncomingValue(GetExternalIndex(PhiB));
+            if (ExtPhiValA != ExtPhiValB)
+              return false;
+          } else {
+            return false; // CountB is not PHI.
+          }
+        }
+      } else {
+        return false; // CountA is not PHI.
+      }
+    }
+    return true;
+  }
+
+  bool mayOverlap(const IntrinsicChainInfo &Other) const {
+    // Different number of updates in the chain.
+    // * but maybe with consistent updates (same steps)
+    if (!hasSameTotalNumberOfCalls(Other))
+      return true;
+
+    // Different parameters.
+    if (!hasSameParameters(Other))
+      return true;
+
+    // Every counter must start equally.
+    if (!hasSameStartingCounters(Other))
+      return true;
+
+    // Check for different steps of calculation.
+    if (hasSameNumberOfCalls(Other))
+      return true;
+
+    return false;
+  }
+};
+
+static IntrinsicArgsInfo collectIntrinsicOperands(const Value *V) {
+  const auto *Extract = dyn_cast<ExtractValueInst>(V);
+  assert(Extract && "Not an Extract");
+  const auto *IntrinsicCall =
+      dyn_cast<CallBase>(Extract->getAggregateOperand());
+  assert(IntrinsicCall && "Not an Intrinsic");
+  Intrinsic::ID ID = IntrinsicCall->getIntrinsicID();
+  IntrinsicArgsInfo Args;
+  // Necessary operands to disambiguate.
+  // Constants.
+  for (unsigned Op : getAddIntrinsicOps(ID)) {
+    Args.ConstantArgs.push_back(IntrinsicCall->getArgOperand(Op));
+  }
+  // Counters.
+  for (unsigned Op : getAddIntrinsicCounterOps(ID)) {
+    Args.CounterArgs.push_back(IntrinsicCall->getArgOperand(Op));
+  }
+  return Args;
+}
+
+// We can reach a pointer definition starting from a counter.
+static bool
+isAddressingIntrinsicOutput(ArrayRef<const Value *> AdressingOutputs,
+                            const Value *Intrinsic) {
+
+  for (const Value *Counter : AdressingOutputs) {
+    if (Intrinsic != lookThroughAIEAddressingModes(Counter))
+      return false;
+  }
+
+  return true;
+}
+
+static bool allInSameBB(SmallVector<const Value *> Values) {
+
+  for (unsigned I = 0; I < Values.size() - 1; I++) {
+    const Instruction *InstrA = dyn_cast<Instruction>(Values[I]);
+    const Instruction *InstrB = dyn_cast<Instruction>(Values[I + 1]);
+
+    if (!InstrA || !InstrB)
+      return false;
+
+    if (InstrA->getParent() != InstrB->getParent())
+      return false;
+  }
+  return true;
+}
+
+/// This function returns some tracking information related to the upward path
+/// that starts from a Value *TargetPointer to the Phi node that firstly defines
+/// a pointer value to the same ambiguous memory. We start the search from the
+/// last update of the pointer.
+std::optional<IntrinsicChainInfo>
+trackIntrinsicChain(const Value *LastUpdate, const Value *TargetPointer) {
+  IntrinsicChainInfo ChainInfo;
+  // We count backwards.
+  std::optional<unsigned> LastCalls = std::nullopt;
+  // Let's expose the real pointer.
+  TargetPointer = skipCast(TargetPointer);
+  for (unsigned Count = 0; Count < BaseObjectSearchLimit; ++Count) {
+    if (!allInSameBB({LastUpdate, TargetPointer}))
+      return std::nullopt;
+
+    LastUpdate = skipCast(LastUpdate);
+
+    // This information is crucial to calculate the number
+    // of pointer updates before TargetPointer.
+    if (LastUpdate == TargetPointer)
+      LastCalls = ChainInfo.TotalIntrinsicCalls;
+
+    if (!LastUpdate->getType()->isPointerTy())
+      return std::nullopt; // not a pointer.
+
+    if (auto *GEP = dyn_cast<GEPOperator>(LastUpdate)) {
+      if (!GEP->hasAllZeroIndices())
+        return std::nullopt; // No pointer arithmetic is allowed.
+      // The result pointer and the first operand have
+      // the same value
+      LastUpdate = GEP->getPointerOperand();
+    } else if (auto *PHI = dyn_cast<PHINode>(LastUpdate)) {
+      // Reached final destination.
+      if (LastCalls) {
+        ChainInfo.IntrinsicCalls = ChainInfo.TotalIntrinsicCalls - *LastCalls;
+        return ChainInfo;
+      }
+      return std::nullopt; // The object was not found.
+
+    } else {
+      const Value *AIEObject = lookThroughAIEAddressingModes(LastUpdate);
+
+      // lookThroughAIEAddressingModes cannot decode LastUpdate.
+      if (LastUpdate == AIEObject) {
+        return std::nullopt;
+      }
+
+      IntrinsicArgsInfo NewArgs = collectIntrinsicOperands(LastUpdate);
+
+      // Initialize counters.
+      if (!ChainInfo.IntrinsicParameters.CounterArgs.empty() &&
+          !isAddressingIntrinsicOutput(
+              ArrayRef(ChainInfo.IntrinsicParameters.CounterArgs), AIEObject)) {
+        // if we already have counters, we need to reach the same
+        // AIEObject from the counters. Otherwise we can have some undesired
+        // counter update in between.
+        return std::nullopt; // not reachable.
+      }
+
+      if (!ChainInfo.IntrinsicParameters.ConstantArgs.empty() &&
+          ChainInfo.IntrinsicParameters.ConstantArgs != NewArgs.ConstantArgs) {
+        // Constant iterator information mismatch between sequence of
+        // intrinsic calls. We stop here.
+        return std::nullopt;
+      }
+
+      // Update params for the next round.
+      ChainInfo.IntrinsicParameters.CounterArgs = NewArgs.CounterArgs;
+      ChainInfo.IntrinsicParameters.ConstantArgs = NewArgs.ConstantArgs;
+
+      LastUpdate = AIEObject;
+      ChainInfo.TotalIntrinsicCalls++;
+    }
+  }
+  return std::nullopt;
+}
+
+static AliasResult aliasAIEIntrinsic(const Value *ValueA, const Value *ValueB,
+                                     const Value *BaseA, const Value *BaseB) {
+
+  const PHINode *PhiA = dyn_cast<PHINode>(BaseA);
+  const PHINode *PhiB = dyn_cast<PHINode>(BaseB);
+
+  // Both bases must be PHI nodes.
+  if (!PhiA || !PhiB)
+    return AliasResult::MayAlias;
+
+  // Skip non-pointer values (cast from pointers?).
+  if (!PhiA->getType()->isPointerTy() || !PhiB->getType()->isPointerTy())
+    return AliasResult::MayAlias;
+
+  // Our target is loop-recurrent PHI nodes.
+  if (PhiA->getNumIncomingValues() != 2 || PhiB->getNumIncomingValues() != 2)
+    return AliasResult::MayAlias;
+
+  // Both PHI nodes must be in the same MBB.
+  if (PhiA->getParent() != PhiB->getParent())
+    return AliasResult::MayAlias;
+
+  auto IsPHIRecurrent = [](const PHINode *Phi) {
+    return Phi->getIncomingBlock(0) == Phi->getParent() ||
+           Phi->getIncomingBlock(1) == Phi->getParent();
+  };
+
+  // Both PHI nodes are recurrent.
+  if (!IsPHIRecurrent(PhiA) || !IsPHIRecurrent(PhiA)) {
+    return AliasResult::MayAlias;
+  }
+
+  auto GetExternalIndex = [](const PHINode *Phi) {
+    return Phi->getIncomingBlock(0) == Phi->getParent() ? 1 : 0;
+  };
+
+  auto GetInternalIndex = [](const PHINode *Phi) {
+    return Phi->getIncomingBlock(0) == Phi->getParent() ? 0 : 1;
+  };
+
+  const Value *ExtPhiValA = PhiA->getIncomingValue(GetExternalIndex(PhiA));
+  const Value *ExtPhiValB = PhiB->getIncomingValue(GetExternalIndex(PhiB));
+
+  // Both Phis must have the same value as an external incoming operand,
+  // so we guarantee that we have two new pointers rooted in the same
+  // original pointer.
+  //
+  // NOTE: We don't care if PhiA and PhiB are the same or not, so we can also
+  // disambiguate different steps starting from from the same pointer (Phi).
+  // We only care if the external value is the same.
+  // For example, we can have the following case:
+  // a =  phi ptr [ %an, %this_mbb ], [ %a_ext, %pred_mbb ]
+  // x1 = load (a)
+  // a1 = add_2d(a, /*fixed parameters + chained counters*/)
+  // x2 = load (a1)
+  // an = add_2d(a1, /*fixed parameters + chained counters*/)
+  // Where we can prove that those two loads are not aliasing, even though
+  // they are sharing the same phi node as an underlying object.
+  if (ExtPhiValA != ExtPhiValB)
+    return AliasResult::MayAlias;
+
+  // Most of the remaining code aims to prove that the two pointers
+  // are updated with same steps.
+
+  const Value *IntPhiValA = PhiA->getIncomingValue(GetInternalIndex(PhiA));
+  const Value *IntPhiValB = PhiB->getIncomingValue(GetInternalIndex(PhiB));
+
+  // Assumption: same basic block.
+  if (!allInSameBB({ValueA, ValueB, BaseA, BaseB, IntPhiValA, IntPhiValB}))
+    return AliasResult::MayAlias;
+
+  // Now we know that the pointers are updated consistently across
+  // loop iterations, let's look to the values.
+  auto ATracking = trackIntrinsicChain(IntPhiValA, ValueA);
+  if (!ATracking) // Non uniform address update accross the loop.
+    return AliasResult::MayAlias;
+
+  auto BTracking = trackIntrinsicChain(IntPhiValB, ValueB);
+  if (!BTracking) // Non uniform address update accross the loop.
+    return AliasResult::MayAlias;
+
+  // If there is no overlap, NoAlias
+  if (ATracking->mayOverlap(*BTracking))
+    return AliasResult::MayAlias;
+
+  return AliasResult::NoAlias;
+}
+
 AliasResult AIEBaseAAResult::alias(const MemoryLocation &LocA,
                                    const MemoryLocation &LocB,
                                    AAQueryInfo &AAQI, const Instruction *) {
 
   const Value *BaseA = getUnderlyingObjectAIE(LocA.Ptr);
   const Value *BaseB = getUnderlyingObjectAIE(LocB.Ptr);
+
+  if (DisambiguateAccessSameOriginPointers &&
+      aliasAIEIntrinsic(LocA.Ptr, LocB.Ptr, BaseA, BaseB) ==
+          AliasResult::NoAlias)
+    return AliasResult::NoAlias;
 
   // Our look-through didn't reveal a different object, pass on to the next
   // alias analysis
--- a/llvm/unittests/Target/AIE/VirtUnrollAliasAnalysisTest.cpp
+++ b/llvm/unittests/Target/AIE/VirtUnrollAliasAnalysisTest.cpp
@@ -1,0 +1,254 @@
+//===- VirtUnrollAliasAnalysisTest.cpp ------------------------------------===//
+//
+// This file is licensed under the Apache License v2.0 with LLVM Exceptions.
+// See https://llvm.org/LICENSE.txt for license information.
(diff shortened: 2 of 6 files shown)


llvm/lib/Target/AIE/AIEBaseAliasAnalysis.cpp
```diff
@@ -117,13 +135,149 @@ static const Value *getUnderlyingObjectAIE(const Value *V) {
...
+  assert(Extract && "Not an Extract");
+  const auto *IntrinsicCall =
+      dyn_cast<CallBase>(Extract->getAggregateOperand());
+  assert(IntrinsicCall && "Not an Intrinsic");
+  Intrinsic::ID ID = IntrinsicCall->getIntrinsicID();
+  // Intrinsic ID is part of the hash.
+  hash_code Hash = hash_value(ID);
+  unsigned InPtrIdx;
+  isAIEPtrAddIntrinsic(ID, InPtrIdx);
+  // Necessary operands to desambiguate.
```
andcarminati: nit: dis

llvm/lib/Target/AIE/AIEBaseAliasAnalysis.cpp
```diff
@@ -117,13 +135,170 @@ static const Value *getUnderlyingObjectAIE(const Value *V) {
...
+  for (unsigned Op = InPtrIdx + 1; Op < hashableAddIntrinsicOps(ID); Op++) {
+    Hash = hash_combine(Hash, hash_value(IntrinsicCall->getArgOperand(Op)));
+  }
+  return Hash;
+}
+
+static std::pair<int, short> trackIntrinsicChain(const Value *V) {
```
martien-de-jong: Could you add some more comment here? Especially what the returned tuple represents.

llvm/lib/Target/AIE/AIEBaseAliasAnalysis.cpp
```diff
@@ -117,13 +135,170 @@ static const Value *getUnderlyingObjectAIE(const Value *V) {
...
+  if (IntrinsicCallsPA != IntrinsicCallsPB)
+    return AliasResult::MayAlias;
+
+  // Different iterators in the chain.
+  // * same number of updated but.
+  // * different steps.
+  if (ChainHashPA != ChainHashPB)
+    return AliasResult::MayAlias;
+
+  // Now we now that the pointers are updated consistently across
```
martien-de-jong: nit: *know*

llvm/lib/Target/AIE/AIEBaseAliasAnalysis.cpp
```diff
@@ -117,13 +135,170 @@ static const Value *getUnderlyingObjectAIE(const Value *V) {
...
+
+  const Value *ExtPhiValA = PhiA->getIncomingValue(GetExternalIndex(PhiA));
+  const Value *ExtPhiValB = PhiB->getIncomingValue(GetExternalIndex(PhiB));
+
+  // Both Phis must have the same value as an external incoming operand,
+  // so we guarantee that we have two new pointers rooted in the same
+  // original pointer.
+  //
+  // NOTE: We don't care if A and B are the same or not, so we can also
+  // desambiguate different steps starting from from the same pointer (Phi).
```
martien-de-jong: nit: *dis*ambiiguate

llvm/test/CodeGen/AIE/alias-analysis-pointer-desambiguation.ll
```diff
@@ -0,0 +1,160 @@
+;
+; This file is licensed under the Apache License v2.0 with LLVM Exceptions.
+; See https://llvm.org/LICENSE.txt for license information.
+; SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+;
+; (c) Copyright 2024 Advanced Micro Devices, Inc. or its affiliates
+; RUN: opt -mtriple=aie2 -passes=aa-eval -print-all-alias-modref-info -disable-output < %s 2>&1 | FileCheck %s --check-prefix=ENABLED-AIE-AA-DES
+; RUN: opt -mtriple=aie2 -passes=aa-eval -print-all-alias-modref-info -disable-output \
+; RUN:    --aie-alias-analysis-disambiguation=false < %s 2>&1 | FileCheck %s --check-prefix=DISABLED-AIE-AA-DES
```
martien-de-jong: DIgital Encryption Standard?

llvm/lib/Target/AIE/AIEBaseAliasAnalysis.cpp
```diff
@@ -117,13 +135,170 @@ static const Value *getUnderlyingObjectAIE(const Value *V) {
...
+    } else {
+      if (auto *PHI = dyn_cast<PHINode>(V)) {
+        // Reached final destination;
+        return {CountIntrinsicCall, Hash};
+      } else if (auto *AIEObject = lookThroughAIEAddressingModes(V)) {
+        hash_code NewHash = hashIntrinsic(V);
+        // Initialize or compare hash?
+        if (Hash == hash_code(0)) {
+          Hash = NewHash;
+        } else if (Hash != NewHash) {
```
martien-de-jong: It's not clear to me whether this hash compare could give false positives

llvm/lib/Target/AIE/AIEBaseAliasAnalysis.cpp
andcarminati: Oh, this is always a confusion in my head because of my first language ;-)

llvm/test/CodeGen/AIE/alias-analysis-pointer-desambiguation.ll
andcarminati: Maybe I tried some AA improvement and got a new encryption alg instead.

llvm/lib/Target/AIE/AIEBaseAliasAnalysis.cpp
andcarminati: Sure! 16 bits can be quite small to be safe! I will replace by a full set of operands and then we can have an exact comparison!

andcarminati: Done!

llvm/lib/Target/AIE/AIEBaseAliasAnalysis.cpp
```diff
@@ -117,13 +138,181 @@ static const Value *getUnderlyingObjectAIE(const Value *V) {
...
+
+  const Value *ExtPhiValA = PhiA->getIncomingValue(GetExternalIndex(PhiA));
+  const Value *ExtPhiValB = PhiB->getIncomingValue(GetExternalIndex(PhiB));
+
+  // Both Phis must have the same value as an external incoming operand,
+  // so we guarantee that we have two new pointers rooted in the same
+  // original pointer.
+  //
+  // NOTE: We don't care if A and B are the same or not, so we can also
+  // disambiguate different steps starting from from the same pointer (Phi).
```
gbossu, line 426: Nit: I feel having some pseudo code example would facilitate understanding

llvm/lib/Target/AIE/AIEBaseAliasAnalysis.cpp
```diff
@@ -117,13 +138,181 @@ static const Value *getUnderlyingObjectAIE(const Value *V) {
...
+        } else if (ArgSet != NewArgSet) {
+          // Iterator information mismatch between sequence of intrinsic
+          // calls. We stop here.
+          return {-1, {}};
+        }
+        V = AIEObject;
+        CountIntrinsicCall++;
+        continue;
+      } else {
+        break; // something that we can't understand.
```
gbossu, line 204: Shouldn't we return then?

llvm/lib/Target/AIE/AIEBaseAliasAnalysis.cpp
```diff
@@ -117,13 +138,181 @@ static const Value *getUnderlyingObjectAIE(const Value *V) {
...
+      V = GEP->getPointerOperand();
+    } else if (Operator::getOpcode(V) == Instruction::BitCast ||
+               Operator::getOpcode(V) == Instruction::AddrSpaceCast) {
+      V = cast<Operator>(V)->getOperand(0);
+      if (!V->getType()->isPointerTy())
+        break;
+    } else {
+      if (auto *PHI = dyn_cast<PHINode>(V)) {
+        // Reached final destination.
+        return {CountIntrinsicCall, ArgSet};
```
gbossu: Should we check that this PHI node corresponds to `PhiA`/`PhiB` in `aliasAIEIntrinsic`?

llvm/lib/Target/AIE/AIEBaseAliasAnalysis.cpp
```diff
@@ -117,13 +138,181 @@ static const Value *getUnderlyingObjectAIE(const Value *V) {
   return V;
 }
 
+SmallPtrSet<Value *, 5> collectIntrinsicOperands(const Value *V) {
```
gbossu: I think the ordering matters, should we return a SmallVector?

llvm/lib/Target/AIE/AIEBaseAliasAnalysis.cpp
```diff
@@ -117,13 +138,181 @@ static const Value *getUnderlyingObjectAIE(const Value *V) {
...
+/// * int: the number of times this pointer is updated through a chain of
+///   add_2d/3d intrinsics calls. If these the pointer is updated in a non
+///   uniform way, ie. using a different addressing information (offset,
+///   (increment and size) this value will be -1. We also return {0 and {}}
+///   int case we track a value that represents the external Phi incoming
+///   value.
+/// * SmallPtrSet<Value*>: represent the constant (across updates) values
+///   that are used to update the pointer across different updates.
+static std::pair<int, SmallPtrSet<Value *, 5>>
+trackIntrinsicChain(const Value *V) {
```
gbossu: I think the function would gain in readability if we had an interface like
```
struct IntrinsicChainInfo {
  const Value *BasePtr = nullptr;
  unsigned IntrinsicCalls = 0;
  SmallVector<const Value *, 5> IntrinsicParameters;
}

std::optional<IntrinsicChainInfo> trackIntrinsicChain(const Value *V);
```
It would return std::nullopt when the chain of intrinsics could not be safely followed. I.e. we replace `return {-1, {}}` by `return {}`;

We can also then check `Result->BasePtr` against `PhiA`/`PhiB`.

llvm/test/CodeGen/AIE/alias-analysis-pointer-desambiguation.ll
```diff
@@ -0,0 +1,160 @@
...
+  %14 = extractvalue { ptr, i20 } %12, 0
+  store i8 %b, ptr %14
+
+  %inc = add nuw nsw i32 %i.08, 1
+  %exitcond.not = icmp eq i32 %inc, %c
+  br i1 %exitcond.not, label %for.cond.cleanup, label %for.body
+
+for.cond.cleanup:                                 ; preds = %for.body
+  ret i32 0
+}
```
gbossu: Nit: Missing new line.

llvm/lib/Target/AIE/AIEBaseAliasAnalysis.cpp
andcarminati: Basically, we get this `PhiA` and `PhiB` by calling the calling getUnderlyingObjectAIE(A/B). Later on, we reconstruct  the tracking information from A/B to the underlying object (we redo the getUnderlyingObjectAIE collecting the information along the way). In this way,  BasePtr will always be PhiA or PhiB.

llvm/lib/Target/AIE/AIEBaseAliasAnalysis.cpp
```diff
@@ -117,13 +138,213 @@ static const Value *getUnderlyingObjectAIE(const Value *V) {
...
+///   that are used to update the pointer across different updates.
+std::optional<IntrinsicChainInfo> trackIntrinsicChain(const Value *V) {
+  IntrinsicChainInfo ChainInfo;
+  SmallVector<Value *, 5> ArgSet;
+  if (!V->getType()->isPointerTy())
+    return std::nullopt;
+  for (unsigned Count = 0; Count < BaseObjectSearchLimit; ++Count) {
+
+    if (auto *GEP = dyn_cast<GEPOperator>(V)) {
+      V = GEP->getPointerOperand();
```
gbossu: Can we really allow non-zero geps? `ptr.add.2d(a,b,c) + offset0` might alias with  `ptr.add.2d(a,b,c) + offset1`

llvm/lib/Target/AIE/AIEBaseAliasAnalysis.cpp
```diff
@@ -117,13 +138,213 @@ static const Value *getUnderlyingObjectAIE(const Value *V) {
...
+        if (ChainInfo.IntrinsicParameters.empty()) {
+          ChainInfo.IntrinsicParameters = NewArgSet;
+        } else if (ChainInfo.IntrinsicParameters != NewArgSet) {
+          // Iterator information mismatch between sequence of intrinsic
+          // calls. We stop here.
+          return std::nullopt;
+        }
+        V = AIEObject;
+        ChainInfo.IntrinsicCalls++;
+        continue;
```
gbossu: Nit: `continue` not needed?

llvm/lib/Target/AIE/AIEBaseAliasAnalysis.cpp
```diff
@@ -117,13 +138,213 @@ static const Value *getUnderlyingObjectAIE(const Value *V) {
...
+    } else if (Operator::getOpcode(V) == Instruction::BitCast ||
+               Operator::getOpcode(V) == Instruction::AddrSpaceCast) {
+      V = cast<Operator>(V)->getOperand(0);
+      if (!V->getType()->isPointerTy())
+        break;
+    } else {
+      if (auto *PHI = dyn_cast<PHINode>(V)) {
+        // Reached final destination.
+        return ChainInfo;
+      } else if (auto *AIEObject = lookThroughAIEAddressingModes(V)) {
```
gbossu: Nit: `else if` -> `if` due to the return above. This will also bring the indentation level down

llvm/lib/Target/AIE/AIEBaseAliasAnalysis.cpp
```diff
@@ -117,13 +138,213 @@ static const Value *getUnderlyingObjectAIE(const Value *V) {
...
+        } else if (ChainInfo.IntrinsicParameters != NewArgSet) {
+          // Iterator information mismatch between sequence of intrinsic
+          // calls. We stop here.
+          return std::nullopt;
+        }
+        V = AIEObject;
+        ChainInfo.IntrinsicCalls++;
+        continue;
+      } else {
+        break; // something that we can't understand.
```
gbossu: Nit: `return std::nullopt` to be consistent

llvm/lib/Target/AIE/AIEBaseAliasAnalysis.cpp
```diff
@@ -117,13 +138,213 @@ static const Value *getUnderlyingObjectAIE(const Value *V) {
...
+  // Necessary operands to disambiguate.
+  for (unsigned Op : getAddIntrinsicOps(ID)) {
+    ArgSet.push_back(IntrinsicCall->getArgOperand(Op));
+  }
+  return ArgSet;
+}
+
+struct IntrinsicChainInfo {
+  unsigned IntrinsicCalls = 0;
+  SmallVector<const Value *, 5> IntrinsicParameters;
```
gbossu: I think we should also track the initial values for the non read-only parameters, i.e. the counters. Maybe add `SmallVector<const Value *, 2> IntrinsicInitialCounters`?

gbossu: Maybe add a test where the counters are different?

llvm/lib/Target/AIE/AIEBaseAliasAnalysis.cpp
gbossu: Maybe add a test with some non-trivial geps?

llvm/lib/Target/AIE/AIEBaseAliasAnalysis.cpp
andcarminati: We can assume the same as the comment below

llvm/lib/Target/AIE/AIEBaseAliasAnalysis.cpp
andcarminati: Hi, now we are fully tracking the counters to make sure they form a chain, without side updates.  The related test is: @diff_initial_counters.

llvm/lib/Target/AIE/AIEBaseAliasAnalysis.cpp
andcarminati: Good point! Now we consider save only zero-indexed geps. We have now 2 tests:

- @test_with_gep_zero
- @test_with_gep_non_zero

llvm/lib/Target/AIE/AIEBaseAliasAnalysis.cpp
```diff
@@ -117,13 +152,304 @@ static const Value *getUnderlyingObjectAIE(const Value *V) {
...
+          // AIEObject from the counters. Otherwise we can have some undesired
+          // counter update in between.
+          if (!isDirectReachable(ChainInfo.IntrinsicParameters.CounterArgSet,
+                                 AIEObject)) {
+            return std::nullopt; // not reachable.
+          }
+        }
+
+        // Update counters for the next round.
+        ChainInfo.IntrinsicParameters.CounterArgSet = NewArgSet.CounterArgSet;
```
gbossu: Isn't that done in the `if` above already?

gbossu: Maybe we could reverse the checks to limit the indentation, e.g.
```
        if (!ChainInfo.IntrinsicParameters.CounterArgSet.empty() &&
            !isDirectReachable(ChainInfo.IntrinsicParameters.CounterArgSet,
                                           AIEObject)) {
          // if we already have counters, we need to reach the same
          // AIEObject from the counters. Otherwise we can have some undesired
          // counter update in between.
          return std::nullopt; // not reachable.
        }
        ChainInfo.IntrinsicParameters.CounterArgSet = NewArgSet.CounterArgSet;
```

llvm/lib/Target/AIE/AIEBaseAliasAnalysis.cpp
```diff
@@ -117,13 +152,304 @@ static const Value *getUnderlyingObjectAIE(const Value *V) {
...
+        return std::nullopt; // something that we can't understand.
+      }
+    }
+    assert(V->getType()->isPointerTy() && "Unexpected operand type!");
+  }
+  return std::nullopt;
+}
+
+static AliasResult aliasAIEIntrinsic(const Value *ValueA, const Value *ValueB,
+                                     const Value *BaseA, const Value *BaseB) {
```
gbossu: I'm getting a bit confused with the different traversals we are doing: one for Value, one for Base, and then we also use `getUnderlyingObjectAIE` to compute `Base`. I would expect that here we do a more specialized traversal, so only `ValueA` and `ValueB` are needed.

I believe we could strart the traversal at both values using `trackIntrinsicChain()` and then compare the results:
 - Both traversals need to end on PHI nodes in the same BB
 - Both PHI nodes need the same initial pointer
 - The chain parameters must be the same: same number of calls, same initial counters, same parameters to the intrinsic calls

I think this is what you're doing here, but I'm wondering if it could not be siomplified into something like below in the end:
```
static AliasResult aliasAIEIntrinsic(const Value *ValueA, const Value *ValueB, unsigned LoopIterationDelta = 0) {
  std::optional<IntrinsicChainInfo> PATracking = trackIntrinsicChain(ValueA);
  std::optional<IntrinsicChainInfo> PBTracking = trackIntrinsicChain(ValueB);

  if (!PATracking || !PATracking)
    return MayAlias;

  return PATracking.samePointer(PBTracking, LoopIterationDelta) ? MustAlias : NoAlias;
}
```

Do you think something like this would be possible? Note we probably need to store a bit more in `IntrinsicChainInfo`,  like the base value we ended up at after the traversal in `trackIntrinsicChain`

llvm/unittests/Target/AIE/VirtUnrollAliasAnalysisTest.cpp
```diff
@@ -0,0 +1,293 @@
...
+  savePoint:       ''
+  restorePoint:    ''
+fixedStack:      []
+stack:           []
+entry_values:    []
+callSites:       []
+debugValueSubstitutions: []
+constants:       []
+machineFunctionInfo:
+  varArgsFrameIndex: 0
```
gbossu: Nit: Do we need all those attributes?

llvm/unittests/Target/AIE/VirtUnrollAliasAnalysisTest.cpp
```diff
@@ -0,0 +1,293 @@
...
+
+  void getInstructions(const MachineBasicBlock *MBB,
+                       std::vector<const MachineInstr *> &Instr) {
+    for (const MachineInstr &MI : *MBB) {
+      Instr.push_back(&MI);
+    }
+  }
+};
+
+TEST_F(VistUnrollAliasAnalysisTest, TestSameIterators) {
```
gbossu: I'm really happy that we are able to unit-test the "virtual unroll" logic 🤩

llvm/lib/Target/AIE/AIEBaseAliasAnalysis.cpp
andcarminati: Hi @gbossu, indeed it looks a bit complicated. The base analysis is to prove the two pointers are updated in the same way (in a cyclic way). I will try a different approach that injects more information  in IntrinsicChainInfo and to not require 4 calls to trackIntrinsicChain.

llvm/lib/Target/AIE/AIEBaseAliasAnalysis.cpp
```diff
@@ -117,13 +153,300 @@ static const Value *getUnderlyingObjectAIE(const Value *V) {
...
+
+  // Most of the remaining code aims to prove that the two pointers
+  // are updated with same steps.
+
+  const Value *IntPhiValA = PhiA->getIncomingValue(GetInternalIndex(PhiA));
+  const Value *IntPhiValB = PhiB->getIncomingValue(GetInternalIndex(PhiB));
+
+  // Now we know that the pointers are updated consistently across
+  // loop iterations, let's look to the values.
+  auto ATracking = trackIntrinsicChain(IntPhiValA, ValueA);
```
andcarminati, line 451: Hi @gbossu, another simplification round, according to your comments.

Now I have just 2 calls to `trackIntrinsicChain`. I already need to have the phi nodes and the incoming values. The external value is used to fast skip some obvious cases. 

In this way, we track all updates between the "Internal value" to the direction of the phi node. During this traversal, we will find the pointer of interest and get some information to precisely calculate the number of updated between phi and this node. We need to start from the last update (internal value) to the phi node to know precisely the number of updates across the loop (this information is also required to virtual unroll).

Also, all verifications are encapsulated in `IntrinsicChainInfo` except the initial phi validation. 

Thank you for all suggestions so far.

llvm/lib/Target/AIE/AIEBaseAliasAnalysis.cpp
```diff
@@ -117,13 +153,300 @@ static const Value *getUnderlyingObjectAIE(const Value *V) {
...
+    Args.CounterArgSet.push_back(IntrinsicCall->getArgOperand(Op));
+  }
+  return Args;
+}
+
+// We can reach a pointer definition starting from a counter.
+static bool isDirectReachable(SmallVector<const Value *, 5> CounterArgSet,
+                              const Value *V) {
+
+  for (auto Counter : CounterArgSet) {
```
gbossu: Nit: Can you expand the `auto`. I guess that is a `Value *`?

llvm/lib/Target/AIE/AIEBaseAliasAnalysis.cpp
```diff
@@ -117,13 +152,286 @@ static const Value *getUnderlyingObjectAIE(const Value *V) {
...
+  // Counters.
+  for (unsigned Op : getAddIntrinsicCounterOps(ID)) {
+    Args.CounterArgSet.push_back(IntrinsicCall->getArgOperand(Op));
+  }
+  return Args;
+}
+
+// We can reach a pointer definition starting from a counter.
+static bool isDirectReachable(SmallVector<const Value *, 5> CounterArgSet,
+                              const Value *V) {
```
gbossu: Nit: maybe
```
static bool isAddressingIntrinsicOutput(ArrayRef<const Value*> AddressingOutputs, const Value *InitialAddress);
```

llvm/lib/Target/AIE/AIEBaseAliasAnalysis.cpp
```diff
@@ -117,13 +152,286 @@ static const Value *getUnderlyingObjectAIE(const Value *V) {
...
+  for (unsigned Count = 0; Count < BaseObjectSearchLimit; ++Count) {
+    if (auto *GEP = dyn_cast<GEPOperator>(LastUpdate)) {
+      if (GEP->hasAllZeroIndices()) {
+        // The result pointer and the first operand have the same value
+        LastUpdate = GEP->getPointerOperand();
+      } else {
+        return std::nullopt; // No pointer arithmetic is allowed.
+      }
+    } else if (Operator::getOpcode(LastUpdate) == Instruction::BitCast ||
+               Operator::getOpcode(LastUpdate) == Instruction::AddrSpaceCast) {
```
gbossu: SuperNit: That seems a bit redundant with `skipCast`. Maybe we could call `LastUpdate = skipCast(LastUpdate)` at every iteration? Then I guess we only need `if (!LastUpdate->getType()->isPointerTy()) { ... }`

llvm/lib/Target/AIE/AIEBaseAliasAnalysis.cpp
```diff
@@ -117,13 +152,286 @@ static const Value *getUnderlyingObjectAIE(const Value *V) {
...
+          return std::nullopt; // not reachable.
+        }
+
+        // Update counters for the next round.
+        ChainInfo.IntrinsicParameters.CounterArgSet = NewArgSet.CounterArgSet;
+
+        // Initialize or compare other operands.
+        if (ChainInfo.IntrinsicParameters.ArgSet.empty()) {
+          ChainInfo.IntrinsicParameters.ArgSet = NewArgSet.ArgSet;
+        } else if (ChainInfo.IntrinsicParameters.ArgSet != NewArgSet.ArgSet) {
```
gbossu: Super-nit:
```
        if (!ChainInfo.IntrinsicParameters.ArgSet.empty() &&
            ChainInfo.IntrinsicParameters.ArgSet != NewArgSet.ArgSet) {
          // Constant iterator information mismatch between sequence of
          // intrinsic calls. We stop here.
          return std::nullopt;
        }

        // Update counters for the next round.
        ChainInfo.IntrinsicParameters.CounterArgSet = NewArgSet.CounterArgSet;
        ChainInfo.IntrinsicParameters.ArgSet = NewArgSet.ArgSet;
```

To be consistent with above.

llvm/lib/Target/AIE/AIEBaseAliasAnalysis.cpp
gbossu: I think it looks great now :)

llvm/lib/Target/AIE/AIEBaseAliasAnalysis.cpp
andcarminati: Thanks, I think now it is a bit more readable!

llvm/lib/Target/AIE/AIEBaseAliasAnalysis.cpp
```diff
@@ -117,13 +152,281 @@ static const Value *getUnderlyingObjectAIE(const Value *V) {
...
+      ChainInfo.IntrinsicParameters.ArgSet = NewArgSet.ArgSet;
+
+      LastUpdate = AIEObject;
+      ChainInfo.TotalIntrinsicCalls++;
+    } else {
+      return std::nullopt; // something that we can't understand.
+    }
+    // This information is crucial to calculate the number
+    // of pointer updates before TargetPointer.
+    if (LastUpdate == TargetPointer)
```
gbossu: Should we actually check `skipCast(LastUpdate) == TargetPointer`?

gbossu: Could you maybe add a test where there is a pointer cast right before the addressing intrinsic?

llvm/lib/Target/AIE/AIEBaseAliasAnalysis.cpp
```diff
@@ -117,13 +152,281 @@ static const Value *getUnderlyingObjectAIE(const Value *V) {
...
+          }
+        }
+      } else {
+        return false; // CountA is not PHI.
+      }
+    }
+    return true;
+  }
+
+  bool notOverlap(const IntrinsicChainInfo &Other) const {
```
gbossu: Super-nit: I'd find `overlaps()` clearer, i.e. avoiding double negations.

llvm/lib/Target/AIE/AIEBaseAliasAnalysis.cpp
```diff
@@ -77,6 +82,36 @@ static bool isAIEPtrAddIntrinsic(Intrinsic::ID ID, unsigned &InPtrIdx) {
   return false;
 }
 
+/// This gives all indexes for read only operands ie.
+/// excluding counters.
+static SmallVector<unsigned, 5> getAddIntrinsicOps(Intrinsic::ID ID) {
```
martien-de-jong, line 89: I think these should be virtual in AIEBaseInstrInfo

llvm/lib/Target/AIE/AIEBaseAliasAnalysis.cpp
```diff
@@ -117,13 +152,281 @@ static const Value *getUnderlyingObjectAIE(const Value *V) {
...
+  return V;
+}
+
+struct IntrinsicArgsInfo {
+  SmallVector<const Value *, 5> ArgSet;
+  SmallVector<const Value *, 5> CounterArgSet;
+};
+
+struct IntrinsicChainInfo {
+  // Calls to an specific point.
```
martien-de-jong: nit: *upto a specific*

llvm/lib/Target/AIE/AIEBaseAliasAnalysis.cpp
```diff
@@ -117,13 +152,281 @@ static const Value *getUnderlyingObjectAIE(const Value *V) {
...
+  unsigned IntrinsicCalls = 0;
+  // Total calls accross the loop.
```
martien-de-jong: nit: *across*

llvm/lib/Target/AIE/AIEBaseAliasAnalysis.cpp
```diff
@@ -117,13 +152,281 @@ static const Value *getUnderlyingObjectAIE(const Value *V) {
...
+  if (const Instruction *Instr = dyn_cast<const Instruction>(V)) {
+    if (Instr->isCast())
+      return Instr->getOperand(0);
+  }
+  return V;
+}
+
+struct IntrinsicArgsInfo {
+  SmallVector<const Value *, 5> ArgSet;
+  SmallVector<const Value *, 5> CounterArgSet;
```
martien-de-jong: I think they are ordered. 'Set' suggests they are not.

llvm/lib/Target/AIE/AIEBaseAliasAnalysis.cpp
martien-de-jong: I guess that would have to be `mayOverlap()`

llvm/lib/Target/AIE/AIEBaseAliasAnalysis.cpp
```diff
@@ -117,13 +152,281 @@ static const Value *getUnderlyingObjectAIE(const Value *V) {
...
+static IntrinsicArgsInfo collectIntrinsicOperands(const Value *V) {
+  const auto *Extract = dyn_cast<ExtractValueInst>(V);
+  assert(Extract && "Not an Extract");
+  const auto *IntrinsicCall =
+      dyn_cast<CallBase>(Extract->getAggregateOperand());
+  assert(IntrinsicCall && "Not an Intrinsic");
+  Intrinsic::ID ID = IntrinsicCall->getIntrinsicID();
+  IntrinsicArgsInfo Args;
+  unsigned InPtrIdx;
+  isAIEPtrAddIntrinsic(ID, InPtrIdx);
```
martien-de-jong: This looks unused?

llvm/lib/Target/AIE/AIEBaseAliasAnalysis.cpp
```diff
@@ -117,13 +152,281 @@ static const Value *getUnderlyingObjectAIE(const Value *V) {
...
+
+    // Check for different steps of calculation.
+    if (hasSameNumberOfCalls(Other))
+      return false;
+
+    return true;
+  }
+};
+
+static IntrinsicArgsInfo collectIntrinsicOperands(const Value *V) {
```
martien-de-jong, line 251: Use anonymous namespace instead of static?

llvm/lib/Target/AIE/AIEBaseAliasAnalysis.cpp
```diff
@@ -117,13 +152,281 @@ static const Value *getUnderlyingObjectAIE(const Value *V) {
...
+    LastUpdate = skipCast(LastUpdate);
+    if (!LastUpdate->getType()->isPointerTy())
+      return std::nullopt; // not a pointer.
+
+    if (auto *GEP = dyn_cast<GEPOperator>(LastUpdate)) {
+      if (!GEP->hasAllZeroIndices())
+        return std::nullopt; // No pointer arithmetic is allowed.
+      // The result pointer and the first operand have
+      // the same value
+      LastUpdate = GEP->getPointerOperand();
```
martien-de-jong, line 329: Could we wrap this case in skipCast() ?

llvm/lib/Target/AIE/AIEBaseAliasAnalysis.cpp
```diff
@@ -117,13 +152,281 @@ static const Value *getUnderlyingObjectAIE(const Value *V) {
...
+    } else if (auto *PHI = dyn_cast<PHINode>(LastUpdate)) {
+      // Reached final destination.
```
martien-de-jong, line 331: I would make this the first case in the if-then-elseif ladder.

llvm/lib/Target/AIE/AIEBaseAliasAnalysis.cpp
```diff
@@ -117,13 +152,281 @@ static const Value *getUnderlyingObjectAIE(const Value *V) {
...
+
+  // Both bases must be PHI nodes.
+  if (!PhiA || !PhiB)
+    return AliasResult::MayAlias;
+
+  // Skip non-pointer values (cast from pointers?).
+  if (!PhiA->getType()->isPointerTy() || !PhiB->getType()->isPointerTy())
+    return AliasResult::MayAlias;
+
+  // Our target is loop-recurrent PHI nodes.
```
martien-de-jong, line 392: I'm not sure whether the condition guarantees that they are loop-recurrent. I'm not sure whether you rely on it. In particular, GetExternalIndex and GetInternalIndex seem to assume that the phi node resides in a loop header.
It is not clear to me whether ValueA and ValueB are supposed to be in the same block as BaseA and BaseB.

llvm/lib/Target/AIE/AIEBaseAliasAnalysis.cpp
```diff
@@ -444,3 +459,42 @@ AliasResult AIEBaseAAResult::alias(const MemoryLocation &LocA,
...
+
+  // Check each pair of memory operands from both instructions.
+  for (auto *MMOA : MIA->memoperands()) {
+    const Value *ValA = MMOA->getValue();
+
+    // We know nothing about the MMO.
+    if (!ValA)
+      return AliasResult::MayAlias;
+
+    const Value *BaseA = getUnderlyingObjectAIE(ValA);
```
martien-de-jong: factor into `std::optional<ValueBasePair>` lambda?

llvm/lib/Target/AIE/AIEBaseAliasAnalysis.cpp
```diff
@@ -117,13 +152,281 @@ static const Value *getUnderlyingObjectAIE(const Value *V) {
...
+
+  // Now we know that the pointers are updated consistently across
+  // loop iterations, let's look to the values.
+  auto ATracking = trackIntrinsicChain(IntPhiValA, ValueA);
+  auto BTracking = trackIntrinsicChain(IntPhiValB, ValueB);
+
+  // Non uniform address update accross the loop.
+  // * one can have more updates or,
+  // * non-uniform updates were found.
+  if (!ATracking || !BTracking)
```
martien-de-jong: I would check ATracking before trying to compute BTracking.

llvm/lib/Target/AIE/AIEBaseAliasAnalysis.cpp
andcarminati: It could be, however, as in the standard value tracking and basic AA, specific machine instructions are not used, so there is no conflicting opcodes (all instrinsics share the same namespace). Also, AA is handled mostly in IR level, so the injection of InstrInfo dependency is not trivial here.

llvm/lib/Target/AIE/AIEBaseAliasAnalysis.cpp
andcarminati: I think the purpose is different. For example, skipCast always return return a non null pointer, and if the GEP is not a cast (non-zero indexes) we need to handle this case in a different way, like an optional return or a null ckeck after the call. I feel that it can complicate the readability.

llvm/lib/Target/AIE/AIEBaseAliasAnalysis.cpp
andcarminati: The existing code is already using static. What do you think about a follow up PR moving everything to anonymous namespace? I feel that changing everything now will create an even bigger change set, and changing just the introduced code will create an "asymmetric" code style in the same file.

llvm/lib/Target/AIE/AIEBaseAliasAnalysis.cpp
andcarminati: I changed a bit the code and the current position matters, because now the intrinsic part is in the else block.

llvm/lib/Target/AIE/AIEBaseAliasAnalysis.cpp
andcarminati: Done! `@with_addrspace_cast`.

llvm/test/CodeGen/AIE/alias-analysis-pointer-desambiguation.ll
```diff
@@ -0,0 +1,557 @@
...
+  %ptrcast = addrspacecast ptr %7 to ptr addrspace(3)
+
+  store i8 %b, ptr addrspace(3) %ptrcast
+
+  %8 = tail call { ptr, i20 } @llvm.aie2.add.2d(ptr %3, i20 0, i20 1, i20 2, i20 %2)
+  %9 = extractvalue { ptr, i20 } %8, 1
+  %10 = extractvalue { ptr, i20 } %8, 0
+  load i8, ptr %10
+
+  %12 = tail call { ptr, i20 } @llvm.aie2.add.2d(ptr %7, i20 0, i20 1, i20 2, i20 %6)
```
gbossu: Is there any way to get the alias info for `%7`? Maybe by inserting a store to `%7` on top of the one for `%ptrcast`?

I'm suspecting that the alias info will be wrong for `%7`, because in the code below, I think that `LastUpdate` will point to `%ptrcast` after visiting the instrinsic, which is different from `%7`, so the update of `LastCalls` will not trigger.
```
%ptrcast = addrspacecast ptr %7 to ptr addrspace(3)
%12 = tail call { ptr, i20 } @llvm.aie2.add.2d(ptr %7, i20 0, i20 1, i20 2, i20 %6)
```

andcarminati: The cast situation is handled by the `@with_addrspace_cast_phi` test.

llvm/lib/Target/AIE/AIEBaseAliasAnalysis.cpp
gbossu: I think we should then also check that the incoming BBs are the same, and one of them is block itself where the PHI nodes are?

llvm/unittests/Target/AIE/VirtUnrollAliasAnalysisTest.cpp
```diff
@@ -0,0 +1,254 @@
...
+  ASSERT_TRUE(AIE::aliasAcrossVirtualUnrolls(Load2, Store2, 0, 0) ==
+              AliasResult::MayAlias);
+  // Store and load with diff virt unroll, same pointer step.
+  ASSERT_TRUE(AIE::aliasAcrossVirtualUnrolls(Load3, Store3, 4, 0) ==
+              AliasResult::NoAlias);
+  // Store and load with diff virt unroll, diff pointer step.
+  // This case is interesting because the first store from the 3rd iteration
+  // will alias with the 3rd load of the second iteration since we are not
+  // updating the pointer after the last load and store (post. inc.).
+  ASSERT_TRUE(AIE::aliasAcrossVirtualUnrolls(Store1, Load3, 3, 2) ==
```
gbossu: Nit: `EXPECT_TRUE` is better because it doesn't stop the test execution at the first failure.

gbossu (dismissed):
LGTM, great work!

llvm/unittests/Target/AIE/VirtUnrollAliasAnalysisTest.cpp
andcarminati: Interesting, I was not aware! I think I will leave as is this time to not interrupt CI again ;-).

llvm/lib/Target/AIE/AIEBaseAliasAnalysis.cpp
martien-de-jong: I think the purpose is to separate the updates from the no-ops. Conceptually, a zero index GEP *is* a pointer cast.

llvm/lib/Target/AIE/AIEBaseAliasAnalysis.cpp
```diff
@@ -117,13 +153,319 @@ static const Value *getUnderlyingObjectAIE(const Value *V) {
...
+        return std::nullopt; // No pointer arithmetic is allowed.
+      // The result pointer and the first operand have
+      // the same value
+      LastUpdate = GEP->getPointerOperand();
+    } else if (auto *PHI = dyn_cast<PHINode>(LastUpdate)) {
+      // Reached final destination.
+      if (LastCalls) {
+        ChainInfo.IntrinsicCalls = ChainInfo.TotalIntrinsicCalls - *LastCalls;
+        return ChainInfo;
+      } else {
```
martien-de-jong: no else after return?

llvm/lib/Target/AIE/AIEBaseAliasAnalysis.cpp
```diff
@@ -117,13 +153,319 @@ static const Value *getUnderlyingObjectAIE(const Value *V) {
...
+
+  // Most of the remaining code aims to prove that the two pointers
+  // are updated with same steps.
+
+  const Value *IntPhiValA = PhiA->getIncomingValue(GetInternalIndex(PhiA));
+  const Value *IntPhiValB = PhiB->getIncomingValue(GetInternalIndex(PhiB));
+
+  // Assumption: same basic block.
+  if (!allInSameBB({ValueA, ValueB, BaseA, BaseB, IntPhiValA, IntPhiValB}))
+    return AliasResult::MayAlias;
```
martien-de-jong, line 447: nit: we've done quite some work until this point, and then bail out on a simple criterion. We know the basic block from the start, and can check on every update we include.

llvm/unittests/Target/AIE/VirtUnrollAliasAnalysisTest.cpp
andcarminati: I considered this! Already updated.

llvm/lib/Target/AIE/AIEBaseAliasAnalysis.cpp
andcarminati: I included another check inside the decomposition loop.

gbossu (approved):

llvm/lib/Target/AIE/AIEBaseAliasAnalysis.cpp
andcarminati: Sure, but in this case we will need to handle the GEP twice, considering the null (skipCast)  and non-null (tracking function). As is, we cast just once to look to the offset.

llvm/lib/Target/AIE/AIEBaseAliasAnalysis.cpp
```diff
@@ -77,6 +83,36 @@ static bool isAIEPtrAddIntrinsic(Intrinsic::ID ID, unsigned &InPtrIdx) {
...
+            /*,count2*/};
+  default:
+    llvm_unreachable("Unknown intrinsic");
+  }
+
+  return {};
+}
+
+/// This gives all indexes counter operands.
+static SmallVector<unsigned, 5> getAddIntrinsicCounterOps(Intrinsic::ID ID) {
```
martien-de-jong, line 104: nit: it looks as if `getAddIntriscOps` is complementary to `getAddIntrinsicCounterOps`, so can be computed as `allInputOperands - getAddIntrinsicCounterOps`

martien-de-jong (approved):

PR #140, opened 2024-07-31
[AIE] Add missing Copyright info

andcarminati:


llvm/lib/Target/AIE/AIEInterBlockScheduling.cpp
```diff
@@ -4,6 +4,8 @@
 // See https://llvm.org/LICENSE.txt for license information.
 // SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
 //
+// (c) Copyright 2023-2024 Advanced Micro Devices, Inc. or its affiliates
+//
```
ValentijnvdBeek: These files were committed three months ago, so you should just put 2024 and not 2023-2024.

gbossu (dismissed):

ValentijnvdBeek (changes requested):

llvm/lib/Target/AIE/AIEInterBlockScheduling.cpp
andcarminati: Done! Thank you for pointing this detail.

ValentijnvdBeek (approved):

gbossu (approved):

PR #142, opened 2024-08-02
[AIE2] Enable libc++ support

konstantinschwarz:
This PR does three things:

1. Enable libc++ and libc++abi builds (headers-only for now)
2. Enable libc/libc++ standard header includes in AIE clang driver
3. Add a compatibility header to support kernels written in AIE_API


makslevental:
Cool :)

clang/lib/Driver/ToolChains/AIE.cpp
```diff
@@ -80,6 +80,50 @@ void AIEToolChain::AddClangSystemIncludeArgs(const ArgList &DriverArgs,
   // Don't pull in system headers from /usr/include or /usr/local/include.
   // All of the basic headers that we need come from the compiler.
   CC1Args.push_back("-nostdsysteminc");
+
+  if (DriverArgs.hasArg(options::OPT_nostdlibinc))
+    return;
+
+  const Driver &D = getDriver();
+  std::string Target = getTripleString();
```
SagarMaheshwari99, line 88: nit: `const` std::string , below as well..

clang/lib/Headers/aiev2_aie_api_compat.h
```diff
@@ -0,0 +1,1461 @@
...
+typedef cint32_t cint32_w64;
+
+struct v2cacc64 {};
+struct v4cacc64 {
+  v4cacc64(v8cint32);
+  operator v8cint32();
+};
+struct v8cacc64;
+struct v8cacc64 {};
+
```
SagarMaheshwari99: nit: eliminate either one?

clang/lib/Headers/aiev2_aie_api_compat.h
```diff
@@ -0,0 +1,1461 @@
...
+struct v1cfloat {};
+struct v2cfloat {};
+struct v4cfloat {};
+struct v8cfloat {
+  v8cfloat(v8cint32);
+};
+struct v16cfloat {};
+
+struct cbfloat16 {};
```
SagarMaheshwari99, line 57: I think this won't work.. There is a conflicting definition in aie-api.. 
But this is required for the intrinsics in this file.. 
This can be changed to `struct cbfloat16;`

clang/lib/Driver/ToolChains/AIE.cpp
```diff
@@ -80,6 +80,50 @@ void AIEToolChain::AddClangSystemIncludeArgs(const ArgList &DriverArgs,
...
+    const llvm::opt::ArgList &DriverArgs,
+    llvm::opt::ArgStringList &CC1Args) const {
+
+  if (DriverArgs.hasArg(options::OPT_nostdinc, options::OPT_nostdlibinc,
+                        options::OPT_nostdincxx))
+    return;
+
+  // We don't ship libc/libcxx for AIE1
+  if (getTriple().isAIE1())
+    return;
```
stephenneuendorffer, line 105: Does this actually fail, or have you just not tried it?

clang/test/CodeGen/aie/aie2/acc32.cpp
```diff
@@ -8,7 +8,7 @@
 // (c) Copyright 2023-2024 Advanced Micro Devices, Inc. or its affiliates
 //
 //===----------------------------------------------------------------------===//
-// RUN: %clang --target=aie2 -S -emit-llvm %s -o - | FileCheck %s
+// RUN: %clang --target=aie2 -nostdlibinc -S -emit-llvm %s -o - | FileCheck %s
```
stephenneuendorffer, line 11: I'm a bit concerned at having to sprinkle this everywhere.  Is this just because these tests might need to run without the runtimes having been built?

clang/lib/Headers/aiev2_aie_api_compat.h
konstantinschwarz: That definition is guarded through the `__AIE_API_CBF16_SUPPORT__` macro, which is to true for our definition of the core model version macro (`10500`)

clang/lib/Driver/ToolChains/AIE.cpp
konstantinschwarz: I haven't tried it

clang/test/CodeGen/aie/aie2/acc32.cpp
konstantinschwarz: Yes this is needed because we call the `%clang` driver here instead of `-cc1`. That expects the runtimes to be present, which we cannot rely on in clang codegen tests.
Applications will not have to set this option

jgmelber:
Tested this using mlir-aie and a [PR](https://github.com/Xilinx/mlir-aie/pull/1717) from @stephenneuendorffer. I verified on hardware using this [passthrough kernel](https://github.com/Xilinx/mlir-aie/blob/main/aie_kernels/generic/passThrough.cc) (after removing unnecessary `includes` and `printfs`) in [this design](https://github.com/Xilinx/mlir-aie/tree/main/programming_examples/basic/passthrough_kernel).

jgmelber (approved):

stephenneuendorffer (approved):

PR #144, opened 2024-08-06
[AIEX] Add support to select Itinerary based of RegClass of Operand 

krishnamtibrewala:
NFC, allowing to define instruction with variable itinerary based on regClass.
By default we assume operand 0 for an instruction but the newly added support allows to provide an operand no.


llvm/utils/TableGen/Common/CodeGenSchedule.cpp
```diff
@@ -1175,7 +1180,7 @@ void CodeGenSchedModels::collectProcItins() {
     // Insert each itinerary data record in the correct position within
     // the processor model's ItinDefList.
     for (Record *ItinData : ItinRecords) {
-      const Record *ItinDef = ItinData->getValueAsDef("TheClass");
+      Record *ItinDef = ItinData->getValueAsDef("TheClass");
```
gbossu: Do we really have to drop `const`?

llvm/test/TableGen/aie-variable-instr-itinerary.td
```diff
@@ -0,0 +1,97 @@
+//===- aie-memory-cycles.td --------------------------------*- tablegen -*-===//
```
gbossu: Nit: Wrong name

llvm/utils/TableGen/Common/CodeGenSchedule.cpp
krishnamtibrewala: the `addSchedClass` expects a non const and from everywhere it is used inside it again expects a `non const` .
I tried making everything `const` but the use tree keeps on growing where it expects a `non const` version.

llvm/lib/Target/AIE/AIE2InstrInfo.cpp
```diff
@@ -52,6 +52,20 @@ using namespace llvm;
...
+static bool regClassMatches(const TargetRegisterClass &TRC,
+                            const TargetRegisterClass *RC, unsigned Reg) {
+  if (RC && TRC.hasSubClassEq(RC))
+    return true;
+  if (TRC.contains(Reg))
+    return true;
+  return false;
+}
+
+#include "AIE2GenVarInstructionItin.inc"
```
krishnamtibrewala, line 54: using already defined `regClassMatches` in `AIE2GenVarInstructionItin.inc`

gbossu: That is a weird dependency. Maybe we can make `regClassMatches` into a static method of `AIEBaseRegisterInfo`? Then `AIE2GenVarInstructionItin.inc` can just use `AIEBaseRegisterInfo::regClassMatches(...)`? This is also better for code sharing between AIE versions

llvm/utils/TableGen/Common/CodeGenSchedule.cpp
```diff
@@ -1190,6 +1195,14 @@ void CodeGenSchedModels::collectProcItins() {
...
                           << " missing class for itinerary "
                           << ItinDef->getName() << '\n');
+        if (EmitUnusedItinerary) {
+          IdxVec Writes, Reads;
+          unsigned SCIdx =
+              addSchedClass(ItinDef, Writes, Reads, /*ProcIndices*/ {0});
+          NumInstrSchedClasses++;
+          ProcModel.ItinDefList.resize(NumInstrSchedClasses);
+          ProcModel.ItinDefList[SCIdx] = ItinData;
+        }
```
gbossu: How about we instead extend `CodeGenSchedModels::collectSchedClasses` so it also reads `ItineraryRegPairs` and not just `Itinerary`? I think we can then have properly defined `Writes, Reads` variables as well if we call `addSchedClass` there. Essentially, we would call `addSchedClass` for `Inst->TheDef->getValueAsDef("Itinerary")` and then for every itinerary in `ItineraryRegPairs` as well.

llvm/utils/TableGen/AIEVariableInstrItineraryEmitter.cpp
```diff
@@ -0,0 +1,176 @@
...
+
+  OS << "  switch (Desc.getOpcode()) {\n";
+  OS << "  default:\n";
+  OS << "    return Desc.getSchedClass();\n";
+  for (const auto &[RegItineraryPairs, Instrs] : UniqueRegItineraryPairs) {
+    for (const auto &Instr : Instrs)
+      OS << "  case " << CurrentNamespace << "::" << Instr << ":\n";
+    OS << "  {\n";
+    for (const auto &[RegClass, Itinerary] : RegItineraryPairs) {
+      OS << "    if (regClassMatches(" << CurrentNamespace << "::" << RegClass
```
gbossu: As mentioned before, please use something like `AIEBaseRegisterInfo::regClassMatches` to make it obvious where that function comes from.

llvm/utils/TableGen/Common/CodeGenSchedule.cpp
krishnamtibrewala: This does sounds like a better idea, 
I believe this way we can make `ItineraryRegPairs` part of standard llvm `Instruction` class.

krishnamtibrewala: Have updated the CodeGenSchedule.cpp as suggested.

llvm/lib/Target/AIE/AIE2InstrInfo.cpp
```diff
@@ -498,32 +485,41 @@ void AIE2InstrInfo::storeRegToStackSlot(MachineBasicBlock &MBB,
   unsigned Opcode;
   LLVM_DEBUG(dbgs() << "Attempting to Store: " << SrcReg << " To " << FI
                     << "\n");
-  if (regClassMatches(AIE2::mSclStRegClass, RC, SrcReg)) {
+  if (AIEBaseInstrInfo::regClassMatches(AIE2::mSclStRegClass, RC, SrcReg)) {
```
gbossu: I don't think you need the `AIEBaseInstrInfo` qualifier here?

llvm/lib/Target/AIE/AIEHazardRecognizer.h
```diff
@@ -178,12 +178,8 @@ class AIEHazardRecognizer : public ScheduleHazardRecognizer {
...
 
 protected:
-  ScheduleHazardRecognizer::HazardType getHazardType(const MCInstrDesc &Desc,
-                                                     MemoryBankBits MemoryBanks,
-                                                     int DeltaCycles);
-  ScheduleHazardRecognizer::HazardType getHazardType(unsigned SchedClass,
-                                                     SlotBits SlotSet,
-                                                     MemoryBankBits MemoryBanks,
+  ScheduleHazardRecognizer::HazardType getHazardType(const MachineInstr *MI,
+                                                     const MCInstrDesc &Desc,
```
gbossu: Please only have `const MachineInstr &MI` and `int DeltaCycles` as parameters. We can easily retrieve `Desc` from `MI`.

Also note I changed the pointer to a reference.

gbossu: Oh I guess you do that because you want to allow a `Desc` that does not correspond to `MI.getDesc()`. I think we discussed that it would be nice not to bring a whole `MachineInstr` as parameter. We want to keep the scheduler open and flexible. In some cases, we won't have an MI to verify conflicts.

It seems you only need `MI` to compute the sched class. Could you then keep the old prototype and just add a `MIOperands` parameter? E.g.

```
ScheduleHazardRecognizer::HazardType getHazardType(
  const MCInstrDesc &Desc, MemoryBankBits MemoryBanks,
  iterator_range<const MachineOperand *> MIOperands, int DeltaCycles);
```

llvm/lib/Target/AIE/AIEHazardRecognizer.h
```diff
@@ -178,12 +178,8 @@ class AIEHazardRecognizer : public ScheduleHazardRecognizer {
   unsigned computeScoreboardDepth() const;
 
 protected:
-  ScheduleHazardRecognizer::HazardType getHazardType(const MCInstrDesc &Desc,
-                                                     MemoryBankBits MemoryBanks,
-                                                     int DeltaCycles);
-  ScheduleHazardRecognizer::HazardType getHazardType(unsigned SchedClass,
```
gbossu, line 184: I agree, let's remove that variant of getHazardType, it's only used by the one above.

llvm/lib/Target/AIE/AIEHazardRecognizer.h
krishnamtibrewala: As you mentioned i bring MI & Desc because I do not want to relate `Desc` & `MI.getDesc()`. Based on our last conversation regarding the reason of not bring `MachineInstr` was it would impact unit test, but that is not the case since we use `checkConflict` and `enterResources`.
 
With respect to possibility of not having `MI`, currently every where `MI` is available.
Have also added an interface as suggested which will work without `MI`. I believe we should have interface with MI because its very simple to use and will avoid possible mistake.

llvm/lib/Target/AIE/AIEBaseInstrInfo.cpp
```diff
@@ -528,14 +528,11 @@ static bool isPreRA(const MachineFunction &MF) {
...
 
 // Reminder: this is used to compute the latency of RAW edges only.
-std::optional<unsigned>
-AIEBaseInstrInfo::getOperandLatency(const InstrItineraryData *ItinData,
-                                        const MachineInstr &DefMI,
-                                        unsigned DefIdx,
-                                        const MachineInstr &UseMI,
-                                        unsigned UseIdx) const {
-  std::optional<int> Lat = getSignedOperandLatency(
-          ItinData, DefMI, DefIdx, UseMI, UseIdx, SDep::Data);
```
gbossu, line 538: Is there a way not to commit these un-necessary diff lines?

llvm/lib/Target/AIE/AIEHazardRecognizer.h
gbossu: The scheduler is evolving quite a bit, and in future cases, I can assume we won't have a full `MI` to pass around. This is also un-doing a previous decision of only passing what is necessary to `AIEHazardRecognizer` internal calls. In that regard, I believe we should only have the interface below, and get rid of any `MachineInstr`:
```
ScheduleHazardRecognizer::HazardType getHazardType(
  const MCInstrDesc &Desc, MemoryBankBits MemoryBanks,
  iterator_range<const MachineOperand *> MIOperands, int DeltaCycles);
```

krishnamtibrewala: Have updated to one interface as suggested above.

llvm/lib/Target/AIE/AIEBaseInstrInfo.cpp
krishnamtibrewala: Yes, have cleaned up.

llvm/test/TableGen/aie-variable-instr-itinerary.td
```diff
@@ -0,0 +1,132 @@
...
+  def MOV_instr_GPR_1 : MOVInstruction<>;
+  def MOV_instr_GPR_2 : MOVInstruction<>;  
+}
+
+let Itinerary = II_MOV, ItineraryRegPairs = [ItinRegClassPair<II_MOV_GPR,GPR, 1>, ItinRegClassPair<II_MOV_PTR,PTR, 1>]  in {
+  def MOV_instr_GPR_PTR_1 : MOVInstruction<>;
+  def MOV_instr_GPR_PTR_2 : MOVInstruction<>;
+}
+
+// CHECK:      static std::optional<unsigned> getOperandIdx(const unsigned Opcode) {
```
gbossu: Nit: This is a very generic function name and it might create name conflicts when including this `.inc` file. Maybe we could name it `getOperandIdxForSchedClassDisambiguation()`?

llvm/test/TableGen/aie-variable-instr-itinerary.td
```diff
@@ -0,0 +1,132 @@
...
+// CHECK:      unsigned TestAIEInstrInfo::getSchedClass(
+// CHECK-NEXT:     const MCInstrDesc &Desc,
+// CHECK-NEXT:     iterator_range<const MachineOperand *> Operands) const {
+// CHECK-NEXT:   std::optional<unsigned> OperandIdx = getOperandIdx(Desc.getOpcode());
+// CHECK-NEXT:   if (!OperandIdx.has_value())
+// CHECK-NEXT:     return Desc.getSchedClass();
+// CHECK-EMPTY: 
+// CHECK-NEXT:   const MachineOperand &MO = *std::next(Operands.begin(), OperandIdx.value());
+// CHECK-NEXT:   const Register Reg = MO.getReg();
+// CHECK-NEXT:   assert(MO.isReg() && "Operand must be of Reg type");
```
gbossu: That assert should be one line higher. `MO.getReg();` will already fail if `MO` is not a register.

llvm/test/TableGen/aie-variable-instr-itinerary.td
```diff
@@ -0,0 +1,132 @@
...
+// CHECK-EMPTY:  
+// CHECK-NEXT:   const MachineInstr *MI = MO.getParent();
+// CHECK-NEXT:   const MachineRegisterInfo &MRI =
+// CHECK-NEXT:       MI->getParent()->getParent()->getRegInfo();
```
gbossu: This is a very complicated way to access `MRI`. This also defeats the purpose of passing only the operands to this function. Can you just add a `const MachineRegisterInfo &MRI` parameter?

Note this means we'll also need to add it to `AIEHazardRecognizer::getHazardType` in a later commit. It seems maybe a bit too much, but at least it's clear from the prototype of `getSchedClass` what its inputs are: a descriptor, a list of register operands, and the register class assignments of those operands.

llvm/test/TableGen/aie-variable-instr-itinerary.td
```diff
@@ -0,0 +1,132 @@
...
+// CHECK-NEXT:       Reg.isVirtual() ? MRI.getRegClass(Reg) : nullptr;
+// CHECK-EMPTY: 
+// CHECK-NEXT:   switch (Desc.getOpcode()) {
+// CHECK-NEXT:   default:
+// CHECK-NEXT:     return Desc.getSchedClass();
+// CHECK-NEXT:   case TestAIE::MOV_instr_GPR:
+// CHECK-NEXT:   case TestAIE::MOV_instr_GPR_1:
+// CHECK-NEXT:   case TestAIE::MOV_instr_GPR_2:
+// CHECK-NEXT:   {
+// CHECK-NEXT:     if (AIEBaseInstrInfo::regClassMatches(TestAIE::GPRRegClass, RC, Reg))
```
gbossu: We could also just write the operand index here for each opcode. E.g. here we would write `checkRCForOperand(TestAIE::GPRRegClass, 0)`, and for `MOV_instr_GPR_PTR_2` we would write
`checkRCForOperand(TestAIE::GPRRegClass, 1)`.

Note I would define `checkRCForOperand()` as a simple lambda within this function, e.g.
```
auto checkRCForOperand= [&Operands, &MRI] (const TargetRegisterClass &ExpectedRC, unsigned OpIdx) {
  Register Reg = std::next(Operands.begin(), OpIdx)->getReg();
  const TargetRegisterClass *RC = Reg.isVirtual() ? MRI.getRegClass(Reg) : nullptr;
  return AIEBaseInstrInfo::regClassMatches(TestAIE::GPRRegClass, RC, Reg);
};
```

TLDR: I would be happy to have a single switch statement instead of two.

llvm/test/TableGen/aie-variable-instr-itinerary.td
krishnamtibrewala: Have updated everything as you suggested.
With respect to passing MRI can you suggest a way. Places from where `getSchedClass(..)` is called from I do not find a neat way of extracting MRI. Currently the best I can find is to get it out from `MachineOperand`

llvm/lib/Target/AIE/AIEHazardRecognizer.cpp
```diff
@@ -332,7 +333,8 @@ AIEHazardRecognizer::getHazardType(SUnit *SU, int DeltaCycles) {
     return NoopHazard;
   }
 
-  return getHazardType(MI->getDesc(), getMemoryBanks(MI), DeltaCycles);
+  return getHazardType(MI->getDesc(), getMemoryBanks(MI), MI->operands(),
+                       DeltaCycles);
```
gbossu: Here I believe you can pass `MI->getMF()->getRegInfo()`

llvm/test/TableGen/aie-variable-instr-itinerary.td
krishnamtibrewala: Have added `const MachineRegisterInfo &MRI` parameter

gbossu (approved):

PR #146, opened 2024-08-07
Post RA software pipeliner

martien-de-jong:
First workable version of the post-RegAlloc pipeliner.
It creates a DAG that simulates an unrolled loop, and it check whether all latencies can be met by scheduling every copy in the same cycle modulo II.
Since the DAG considers the physical register deps, there's no register renaming to be done, and it uses negative latencies in the same way as the regular post-scheduler.
The results look promising for underfull loops, typically load-operate-store, where the load latency should span quite a few loop iterations to get the resources saturated.

There are still some rough edges, and more tests need to be added, especially to cover the cases where SWP fails.


llvm/lib/Target/AIE/AIEInterBlockScheduling.cpp
```diff
@@ -81,10 +81,19 @@ void emitBundlesInScoreboardDelta(
...
-MachineBasicBlock *getSinglePredecessor(const MachineBasicBlock &MBB) {
-  assert(MBB.pred_size() == 1 && "MBB contains more than 1 predecessor");
-  MachineBasicBlock *SinglePredMBB = *MBB.predecessors().begin();
-  return SinglePredMBB;
+MachineBasicBlock *getLoopPredecessor(const MachineBasicBlock &MBB) {
+  if (MBB.pred_size() == 1) {
+    // if we have only one, it must be the loop
+    return *MBB.predecessors().begin();
+  }
+  // Otherwise, the loop is the fallthrough predecessor by construction
```
gbossu: I'm not sure I understand, are we looking for a predecessor of the loop, or for the block which the loop is the predecessor of? I assume the latter given the comment you wrote. Maybe clarify the function name?

Also, how can we guarantee the loop is the fall-through pred, can't we have a branch if MachineBlockPlacement does a bad job?

llvm/lib/Target/AIE/AIEInterBlockScheduling.cpp
```diff
@@ -422,39 +431,79 @@ void InterBlockScheduling::enterRegion(MachineBasicBlock *BB,
     BS.addRegion(BB, RegionBegin, RegionEnd);
   }
 }
+
+// Create a block, insert it before Succ, and route the control flow edge
+// between Pred and Succ through it.
+// Since we don't add any control flow instructions, the edge should be a
+// fallthrough edge; it will be replaced with two fallthrough edges and a block
```
gbossu: Where do we verify that the loop falls through to its exit block?

llvm/lib/Target/AIE/AIEInterBlockScheduling.cpp
martien-de-jong: We don't classify something as a loop if the epilogue is not a dedicated exit, unless that exit is fallthrough from the loop. And that classification runs in the same pass. We are looking for the loop, which is a predecessor of the epilogue, and if there are more, it can only be the fallthrough one

llvm/lib/Target/AIE/AIEInterBlockScheduling.cpp
martien-de-jong: In BlockState::classify(), using the layout numbering. If we still have a jump on a fallthrough, we should get rid of that first before thinking about post-pipelining.

llvm/lib/Target/AIE/AIE2InstrInfo.cpp
```diff
@@ -1174,12 +1174,21 @@ bool AIE2InstrInfo::isHardwareLoopEnd(unsigned Opcode) const {
...
 }
 
+bool AIE2InstrInfo::isZOLTripCountDef(const MachineInstr &MI) const {
+  return MI.getOpcode() == AIE2::ADD_NC &&
+         MI.getOperand(0).getReg() == AIE2::LC;
+}
+
+void AIE2InstrInfo::adjustTripCount(MachineInstr &MI, int Adjustment) const {
+  assert(MI.getOpcode() == AIE2::ADD_NC);
+  auto &Imm = MI.getOperand(2);
```
andcarminati, line 1197: Should we assert also for the lc register? Or even more general:
`assert(isZOLTrip....`

llvm/lib/Target/AIE/AIE2TargetTransformInfo.h
```diff
@@ -54,10 +54,10 @@ class AIE2TTIImpl : public BasicTTIImplBase<AIE2TTIImpl> {
   void getUnrollingPreferences(Loop *L, ScalarEvolution &SE,
                                TTI::UnrollingPreferences &UP,
                                OptimizationRemarkEmitter *ORE) {
-    UP.Partial = UP.Runtime = true;
-    UP.AllowExpensiveTripCount = true;
     UP.Threshold = 200;
     BaseT::getUnrollingPreferences(L, SE, UP, ORE);
+    UP.Partial = UP.Runtime = false;
```
andcarminati: This is nice! Now our decisions are not changed by BaseT implementation! Maybe those small changes can be merged in small PRs.

llvm/lib/Target/AIE/AIEInterBlockScheduling.cpp
```diff
@@ -79,10 +86,19 @@ void emitBundlesInScoreboardDelta(
   }
 }
 
-MachineBasicBlock *getSinglePredecessor(const MachineBasicBlock &MBB) {
-  assert(MBB.pred_size() == 1 && "MBB contains more than 1 predecessor");
-  MachineBasicBlock *SinglePredMBB = *MBB.predecessors().begin();
-  return SinglePredMBB;
+MachineBasicBlock *getLoopPredecessor(const MachineBasicBlock &MBB) {
```
andcarminati: nit: good candidate for LoopUtils.

llvm/lib/Target/AIE/AIEInterBlockScheduling.cpp
```diff
@@ -79,10 +86,19 @@ void emitBundlesInScoreboardDelta(
...
+  if (MBB.pred_size() == 1) {
+    // if we have only one, it must be the loop
+    return *MBB.predecessors().begin();
+  }
+  // Otherwise, the loop is the fallthrough predecessor by construction
+  int FallThroughNumber = MBB.getNumber() - 1;
+  for (auto *Pred : MBB.predecessors()) {
+    if (Pred->getNumber() == FallThroughNumber) {
```
andcarminati: Can we use `isLayoutSuccessor` instead of the comparison of getNumbers?

llvm/lib/Target/AIE/AIEBaseHardwareLoops.cpp
```diff
@@ -297,9 +297,12 @@ void AIEBaseHardwareLoops::expandLoopStart(LowOverheadLoop &LoLoop) {
   MachineInstr *Start = LoLoop.Start;
   MachineBasicBlock *MBB = Start->getParent();
   LLVM_DEBUG(dbgs() << "AIE Loops: ZOL loop. Expanding LoopStart.\n");
-  BuildMI(*MBB, Start, Start->getDebugLoc(), TII->get(AIE2::MOV_mv_scl),
-          AIE2::LC)
-      .addReg(Start->getOperand(0).getReg());
+
+  // We use ADD_NC, which allows PostPipeliner to tweak it by modifying the
+  // immediate value.
+  BuildMI(*MBB, Start, Start->getDebugLoc(), TII->get(AIE2::ADD_NC), AIE2::LC)
```
andcarminati, line 359: We merged e33bbe18b8ed6b608cf77a9c5677d9819ae58563 as a bug fix for the trip count adjustment. As this change also overcomes the same problem, we can replace later (revert the first fix, for example).

llvm/lib/Target/AIE/AIEPostPipeliner.cpp
```diff
@@ -0,0 +1,278 @@
...
+  if (!TripCountDef) {
+    return false;
+  }
+
+  // 3a. If the addend immediate is non zero, we have been pipelined before
+  // and we can't trust min itercount metadata. Return on investment is probably
+  // low anyway.
+  // FIXME: Strictly speaking, this is AIE2 specific and should get some
+  // TII support. On the other hand, it will probably remain compatible.
+  if (TripCountDef->getNumOperands() != 3 ||
```
andcarminati: Maybe something like:
```
TII->isNonPipelinedZOLTripCountDef(MI)
```
We can vanish some code here...

llvm/lib/Target/AIE/AIEInterBlockScheduling.cpp
```diff
@@ -165,33 +170,109 @@ void InterBlockScheduling::leaveFunction() {
...
+  CurrentBlockState->resetRegion();
   DEBUG_BLOCKS(dbgs() << "  >> enterBlock " << BB->getNumber() << " "
-                      << CurrentBlock->kindAsString() << " FixPointIter="
-                      << CurrentBlock->FixPoint.NumIters << "\n");
-}
+                      << CurrentBlockState->kindAsString() << " FixPointIter="
+                      << CurrentBlockState->FixPoint.NumIters
+                      << " II=" << CurrentBlockState->FixPoint.II);
+}
+namespace {
```
andcarminati, line 229: As this file is growing, maybe we can move this class to AIEPostPipeliner* files.

llvm/lib/Target/AIE/AIEInterBlockScheduling.cpp
```diff
@@ -150,33 +170,109 @@ void InterBlockScheduling::leaveFunction() {
...
+class PipelineExtractor : public PipelineScheduleVisitor {
+  BlockState &Loop;
+  BlockState *Prologue = nullptr;
+  BlockState *Epilogue = nullptr;
+  MachineBundle CurrentBundle;
+  std::vector<MachineBundle> TimedRegion;
+  bool InLoop = false;
+
+  void startPrologue() override {
+    // Nothing at this time, but let's keep the override around
```
andcarminati, line 244: Do we have some future plan for this method?

andcarminati:
Hi @martien-de-jong, some commits from this PR were merged:

https://github.com/Xilinx/llvm-aie/pull/152
https://github.com/Xilinx/llvm-aie/pull/162

llvm/lib/Target/AIE/AIEInterBlockScheduling.cpp
martien-de-jong: Only vaguely. Currently we insert an opaque region which is not seen by the normal scheduler. In future we may need to do some boundary signaling.

llvm/lib/Target/AIE/AIEInterBlockScheduling.cpp
martien-de-jong: I'm hesitant, since it only applies to epilogues that have been selected by the classification of loop-aware candidates here.

llvm/lib/Target/AIE/AIEPostPipeliner.cpp
martien-de-jong: Done. Included 'Pristine' parameter in isZOLTripCountDef

llvm/lib/Target/AIE/AIEInterBlockScheduling.cpp
martien-de-jong: Perhaps not the best criterion to do this. It implements an interface to the pipeliner that is dedicated to the interblock data structures.  All that knowledge would need to be exported to the postpipeliner, which would clutter it.

llvm/include/llvm/CodeGen/ResourceScoreboard.h
```diff
@@ -63,17 +63,21 @@ template <typename RC> class ResourceScoreboard {
...
 
-  void reset(int D = 1) {
-    // Implementation relies on masking to wrap-around, so round up
-    // to a power of two.
-    int Pow2 = 1;
-    while (Pow2 < D) {
-      Pow2 += Pow2;
-    }
-    if (Cycles.empty()) {
+  void reset(int D = 0) {
```
gbossu: Nit: How about just introducing a `clear()` method?

llvm/lib/Target/AIE/AIEInterBlockScheduling.cpp
```diff
@@ -4,14 +4,19 @@
 // See https://llvm.org/LICENSE.txt for license information.
 // SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
 //
-// (c) Copyright 2024 Advanced Micro Devices, Inc. or its affiliates
+// (c) Copyright 2023-2024 Advanced Micro Devices, Inc. or its affiliates
```
gbossu: I don't think this is as old as 2023

llvm/lib/Target/AIE/AIEHazardRecognizer.h
```diff
@@ -179,13 +181,20 @@ class AIEHazardRecognizer : public ScheduleHazardRecognizer {
   /// For efficiency, this size is rounded up to a power of two.
   unsigned computeScoreboardDepth() const;
 
+  bool checkConflict(const ResourceScoreboard<FuncUnitWrapper> &Scoreboard,
+                     const MCInstrDesc &Desc, const int DeltaCycles) const;
```
gbossu: Do we need even more of those wrappers? This one seems somewhat error-prone as it ignores register-based itineraries and memory banks.

llvm/lib/Target/AIE/AIEInterBlockScheduling.h
```diff
@@ -314,10 +315,17 @@ class InterBlockScheduling {
...
   std::optional<int> getBlockedResourceCap(MachineBasicBlock *BB) const;
 
-  /// Return the number of nops that must be inserted before this block.
-  /// This is typical for epilogue blocks, which supply the safety margin of
-  /// their loops.
-  int getNumEntryNops(const BlockState &BS) const;
+  /// Return the number of nops that must be inserted before the epilogue
+  /// of the loop represented by this block.
+  int getSafetyMargin(MachineBasicBlock *Loop,
+                      MachineBasicBlock *Epilogue) const;
```
gbossu, line 378: Why change references to pointers?

llvm/lib/Target/AIE/AIEInterBlockScheduling.cpp
```diff
@@ -207,36 +215,111 @@ void InterBlockScheduling::leaveFunction() {
...
+public:
+  PipelineExtractor(InterBlockScheduling &InterBlock, BlockState &BS,
+                    const AIEBaseInstrInfo &TII)
+      : Loop(BS), CurrentBundle(TII.getFormatInterface()) {
+    MachineBasicBlock *LoopBlock = Loop.TheBlock;
+    for (auto *P : LoopBlock->predecessors()) {
+      if (P == LoopBlock) {
+        continue;
+      }
+      Prologue = &InterBlock.getBlockState(P);
```
gbossu, line 279: Checking: We never re-allocate the vector of `BlockStates`?

llvm/lib/Target/AIE/AIEInterBlockScheduling.cpp
```diff
@@ -381,12 +464,39 @@ bool InterBlockScheduling::updateFixPoint(BlockState &BS) {
...
+    // We cut off at the load latency because the main benefit from
+    // post-pipelining comes from wrapping latencies over multiple stages.
+    // This will only work if the stages are small enough. If we miss
+    // opportunities because of this, we should try to get the Prepipeliner
+    // to accept it.
+    if (II >= 7) {
+      // Fall back to the loop schedule. Note that we can only have II != 0
+      // after the loop schedule has stabilized.
+      BS.setPipeliningFailed();
+      return true;
```
gbossu: Doesn't `true` mean that we have converged and do not need to iterate on `MBB` again? In that case don't we want to return `false` to try the "normal" loop scheduling?

gbossu: Hmm looking around, it seems the meaning of the return value is quite overloaded and tied to `isPipelined()` / `isScheduled()`. Maybe `updateFixPoint` could return a `enum State { Failed, TryLoopScheduling, TryModuloScheduling, Done }`?

llvm/lib/Target/AIE/AIEInterBlockScheduling.cpp
```diff
@@ -381,12 +465,39 @@ bool InterBlockScheduling::updateFixPoint(BlockState &BS) {
...
   DEBUG_LOOPAWARE(dbgs() << "Converged,"
                          << " LatencyExtent=" << BS.FixPoint.MaxLatencyExtent
                          << " ResourceExtent=" << BS.FixPoint.MaxResourceExtent
                          << "\n");
 
+  // The loop schedule has converged, so we could declare our work done.
+  // But first try SWP
+  if (BS.getRegions().size() == 1 && BS.getPostSWP().accept(*BS.TheBlock)) {
+    II = 1;
+    return false;
```
gbossu: Oh so we are first trying the normal "loop iterative scheduling", and once we found a decent schedule, we try to chop it with different II. Do we ever change the "linear schedule" once we reach the `isScheduled()` state? Or do we jus try to take it and "modulo-place" it?

llvm/lib/Target/AIE/AIEMachineScheduler.cpp
```diff
@@ -595,9 +604,12 @@ void AIEPostRASchedStrategy::enterRegion(MachineBasicBlock *BB,
 
 void AIEPostRASchedStrategy::leaveRegion(const SUnit &ExitSU) {
   LLVM_DEBUG(dbgs() << "    << leaveRegion\n");
-  materializeMultiOpcodeInstrs();
 
   auto &BS = InterBlock.getBlockState(CurMBB);
+  if (BS.FixPoint.II) {
+    return;
+  }
```
gbossu, line 611: Checking: We have an early exit, because when we try to modulo schedule, we do not ever amend the "linear" schedule that is taken as base.

I think somethink like below would really be clearer:
```
if (BS.getState() == TryModuloScheduling)
  return;
```

I don't even think we need a new attribute in BS, it can maybe just be implemented like below.
```
ScheduleState BlocState::getState() {
  if (isPipelined() || (isScheduled() && FixPoint.II > MaxII)
    return Done;
  if (FixPoint.II)
    return TryModuloScheduling;
  if (!isScheduled() && FixPoint.Iters < MaxIters)
    return TryLinearSchedule;
  return Fail;
}
```

This is just making it more explicit that we do have a state machine.  (Note my understanding of that state machine is still quite partial, so my implementation is probably wrong 😄)

llvm/lib/Target/AIE/AIEMachineScheduler.cpp
```diff
@@ -1151,11 +1163,41 @@ void llvm::AIEPostRASchedStrategy::buildGraph(ScheduleDAGMI &DAG, AAResults *AA,
...
   }
   DAG.ExitSU.setInstr(Region.getExitInstr());
   DAG.makeMaps();
   DAG.buildEdges(Context->AA);
   static_cast<AIEScheduleDAGMI &>(DAG).recordDbgInstrs(Region);
 }
+
+void AIEScheduleDAGMI::schedule() {
+  BlockState &BS = getSchedImpl()->getInterBlock().getBlockState(getBB());
+  if (BS.Kind == BlockType::Loop && BS.FixPoint.II) {
```
gbossu: Again, here I think `BS.getState() == TryModuleScheduling` would be clearer

llvm/lib/Target/AIE/AIEPostPipeliner.cpp
```diff
@@ -0,0 +1,320 @@
...
+  if (Next < NTotalInstrs) {
+    Info[Next].Earliest = std::max(Info[Next].Earliest, Cycle + II);
+  }
+}
+
+// Check resources. We only insert at the position modulo II. Since we insert
+// all iterations separately, the resources that wrap around accumulate in the
+// overflow area, causing conflicts when inserting future iterations
+int PostPipeliner::fit(MachineInstr *MI, int Earliest, int NTries, int II) {
+  for (int C = Earliest; C < Earliest + NTries; C++) {
```
gbossu, line 140: So we are delaying an instruction until it has no resource conflict? Can we guarantee it is safe given that the linuar schedule has already been critically timed?

llvm/lib/Target/AIE/AIEPostPipeliner.cpp
```diff
@@ -0,0 +1,320 @@
...
+    int Mod = C % II;
+    LLVM_DEBUG(dbgs() << "   at " << C << " (" << Mod << ")\n");
+    if (!HR.checkConflict(Scoreboard, *MI, -Depth + Mod)) {
+      LLVM_DEBUG(dbgs() << "    Success\n");
+      return C;
+    }
+  }
+  LLVM_DEBUG(dbgs() << "    Fail\n");
+
+  return -1;
```
gbossu, line 150: `std::optional` :)

-1 could be interpreted as a relative cycle

llvm/lib/Target/AIE/AIEPostPipeliner.cpp
```diff
@@ -0,0 +1,320 @@
...
+}
+
+bool PostPipeliner::scheduleFirstIteration() {
```
gbossu, line 253: I'm still a bit confused as to what "first iteration" means here. I assume it's not the first iteration of the non-pipelined loop (which would be e.g. S0+S1+S2), but rather the first iteration of the steady state? Which would be e.g. S2(iter=0) + S1(iter=1) + S0(iter=2)

llvm/lib/Target/AIE/AIEInterBlockScheduling.cpp
martien-de-jong: No, only on first creation. That's part and parcel of the function-level design of interblock

llvm/include/llvm/CodeGen/ResourceScoreboard.h
martien-de-jong: Yes, sure. Probably an unsupported hunch of many call sites to fix.

llvm/lib/Target/AIE/AIEHazardRecognizer.h
martien-de-jong: Agreed. This is a level zero resolution of a conflict between Krishnam's PR and mine. I was hesitant to bring in MachineInstr internals here, but I would really like to have a concise conflict checker with an external scoreboard.

llvm/lib/Target/AIE/AIEInterBlockScheduling.h
martien-de-jong: I don't understand 'change'. This is a new interface. I can make them references. It could be that my first version could have a nullptr Epilogue or Loop to overload the loop and normal interblock case.

llvm/lib/Target/AIE/AIEInterBlockScheduling.cpp
martien-de-jong: Yeah, makes sense. And that enumeration should replace the different flags in blockstate. I'm still hinging on running the scheduling algorithm as one big fixed-point iteration of all blocks, and then that state determines what to do on enterBlock etc.

llvm/lib/Target/AIE/AIEInterBlockScheduling.cpp
martien-de-jong: Once isScheduled is reached, the regular schedule is fixed, it can only be undone by a succesful postpipelining, which rebuilds the MachineInstr order from scratch. Currently we need one thing from the linear schedule, which is its length, since that determines the number of copies when creating the unrolled DAG. I've been doubting that, since it holds the assumption that this is on par with what the postpipeliner will find. 
The cleanest way would be to generate the nromal single copy DAG, then run scheduleFirstIteration, then use the length from there as stagecount to add the necessary copies to the DAG, then run scheduleOtherIterations.
At this point that would be quite invasive.

llvm/lib/Target/AIE/AIEMachineScheduler.cpp
martien-de-jong: That state machine should be absolutely clear, since it is interpreted by a disparate set of call backs, and potentially represents the state of every basic block in the function.
I think that IsScheduled, IsPipelined, etc could all be represented by a single enumeration per block, which tell the different scheduler call backs tell what to do.  My gut feeling is that 'BuildRegionStructure' could be added as an initial non-scheduling state, which would probably tidy up some of the awkward locations of doing region management.
enterRegion would just administrate the regions, and enterBasicBlock could decide to iterate over the regions by itself.

llvm/lib/Target/AIE/AIEPostPipeliner.cpp
martien-de-jong: The intention is not to be dependent on the linear schedule. scheduleFirstIteration is in fact generating a new linear schedule, that is meeting resource and latency. scheduleOtherIterations calls fit() with NTries = 1, which means that it can not delay instructions.
As remarked elsewhere, the linear schedule's length determines the upperbound on the number of stages and the number of copies in the DAG.  I guess we could eliminate any mismatch there by not accepting a schedule with more stages, or rework PostPipeliner to use scheduleFirstIteration for the upperbound of NStages.

llvm/lib/Target/AIE/AIEPostPipeliner.cpp
martien-de-jong: And how would you interpret -1 relative to Earliest? :-D

llvm/lib/Target/AIE/AIEPostPipeliner.cpp
martien-de-jong: It is actually the first iteration of the non-pipelined loop, and it will probably span all stages of the schedule, though not have instructions in all stages. If we take a simple load+store loop, the load will be in stage 0, the store in stage 7.  With an II of 1, the second iteration will be in stage 1 and stage 8, the third in stage 2 and stage 9, etc
Each next iteration will be shifted by exactly II by the constraints we apply in scheduleOtherIterations.

llvm/lib/Target/AIE/AIEHazardRecognizer.h
martien-de-jong: In fact, it is dead after that rebase.

llvm/lib/Target/AIE/AIEInterBlockScheduling.h
```diff
@@ -4,7 +4,7 @@
 // See https://llvm.org/LICENSE.txt for license information.
 // SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
 //
-// (c) Copyright 2024 Advanced Micro Devices, Inc. or its affiliates
+// (c) Copyright 2023-2024 Advanced Micro Devices, Inc. or its affiliates
```
gbossu: Same: please leave as 2024

llvm/lib/Target/AIE/AIEHazardRecognizer.h
```diff
@@ -186,6 +186,8 @@ class AIEHazardRecognizer : public ScheduleHazardRecognizer {
                 const MachineRegisterInfo &MRI, int DeltaCycles) const;
 
 protected:
+  ScheduleHazardRecognizer::HazardType getHazardType(const MCInstrDesc &Desc,
+                                                     int DeltaCycles);
```
gbossu, line 194: I don't see that new wrapper being defined. Do we need it?

llvm/lib/Target/AIE/AIEInterBlockScheduling.h
gbossu: Correct, `change` does not make sense :D

Still, if you expect the pointers to be non-null, I think a reference is better

llvm/lib/Target/AIE/AIEMachineScheduler.cpp
```diff
@@ -545,15 +545,12 @@ void AIEPostRASchedStrategy::enterMBB(MachineBasicBlock *MBB) {
...
-  MachineBasicBlock::iterator It = BB->begin();
-  const TargetInstrInfo *TII = getTII(BB);
 
-  // Emit the top safety margin
-  int Margin = InterBlock.getNumEntryNops(BS);
-  while (Margin--) {
-    TII->insertNoop(*BB, It);
-  }
+  // Safety margin, swp epilogue
+  InterBlock.emitInterBlockTop(BS);
```
gbossu, line 553: We emit epilogue-specific code, shouldn't that be in `emitInterBlockBottom`?

llvm/lib/Target/AIE/AIEInterBlockScheduling.h
```diff
@@ -314,10 +315,17 @@ class InterBlockScheduling {
   /// This represents the resource usage we assume for an unscheduled successor.
   std::optional<int> getBlockedResourceCap(MachineBasicBlock *BB) const;
 
-  /// Return the number of nops that must be inserted before this block.
-  /// This is typical for epilogue blocks, which supply the safety margin of
-  /// their loops.
-  int getNumEntryNops(const BlockState &BS) const;
+  /// Return the number of nops that must be inserted before the epilogue
+  /// of the loop represented by this block.
+  int getSafetyMargin(MachineBasicBlock *Loop,
```
gbossu, line 377: Nit: safety-margin is very generic, and up to this day, we mostly used it to represent the distance to `ExitSU`. Maybe rename to `getEpilogueSafetyMargin`?

llvm/test/CodeGen/AIE/aie2/schedule/postpipeliner/bitwisexor.mir
```diff
@@ -0,0 +1,163 @@
...
+
+    $lc = ADD_NC $r0, 0
+    $ls = MOVXM_lng_cg %bb.2
+    $le = MOVXM_lng_cg <mcsymbol .L_LEnd0>
+
+  bb.2.for.body (align 16):
+    successors: %bb.2(0x7c000000), %bb.3(0x04000000)
+    liveins: $p0, $p1, $p2
+
+    $wh0 = VLD_idx_imm_3x32_pseudo $p1, 32 :: (load (<32 x s8>) from %ir.in2.addr.014 + 32, !tbaa !2, addrspace 6)
```
gbossu, line 159: Nit: I'd prefer to use "fake" MMOs like `load from stack - 32` if it means we can get rid of the IR and debloat the file :)

llvm/test/CodeGen/AIE/aie2/schedule/postpipeliner/load-add-store-renamed.prepiplined.mir
```diff
@@ -18,21 +18,21 @@
...
-  ; CHECK-NEXT:    jnz r1, #.LBB0_4
+  ; CHECK-NEXT:    jnz r1, #.LBB0_3
   ; CHECK-NEXT:    nop // Delay Slot 5
   ; CHECK-NEXT:    nop // Delay Slot 4
   ; CHECK-NEXT:    nop // Delay Slot 3
   ; CHECK-NEXT:    nop // Delay Slot 2
   ; CHECK-NEXT:    nop // Delay Slot 1
   ; CHECK-NEXT:  // %bb.1: // %for.body.preheader
-  ; CHECK-NEXT:    add.nc lc, r0, #-3
+  ; CHECK-NEXT:    add.nc lc, r0, #-1
```
gbossu, line 28: What happened here? I don't see any code change that could trigger this in this commit

llvm/lib/Target/AIE/AIEBaseHardwareLoops.cpp
```diff
@@ -173,6 +173,54 @@ char AIEBaseHardwareLoops::ID = 0;
 INITIALIZE_PASS(AIEBaseHardwareLoops, DEBUG_TYPE, AIE_HARDWARE_LOOPS_NAME,
                 false, false)
 
+namespace {
+// Everything becomes much simpler if ZOLs exit to their layout successor
+// We can achieve that by splitting a non-fallthrough exit jump off into
+// a new fallthrough block.
```
gbossu, line 179: This isn't a requirement somewhere in our block placement?

llvm/lib/Target/AIE/AIEBaseHardwareLoops.cpp
```diff
@@ -173,6 +173,54 @@ char AIEBaseHardwareLoops::ID = 0;
...
+  auto Terminator = MBB.getFirstInstrTerminator();
+  if (Terminator == MBB.end() ||
+      !TII->isHardwareLoopEnd((*Terminator).getOpcode())) {
+    return false;
+  }
+  SmallVector<MachineOperand, 4> Cond;
+  MachineBasicBlock *TBB, *FBB;
+  const bool AllowModify = false;
+  TII->analyzeBranch(MBB, TBB, FBB, Cond, AllowModify);
+  if (!FBB) {
```
gbossu, line 196: I don't remember how `analyzeBranch` works (who does?), but shouldn't we only trust `TBB` or `FBB` when the branch was successfully analyzed?

llvm/lib/Target/AIE/AIEHazardRecognizer.cpp
```diff
@@ -441,6 +442,19 @@ ScheduleHazardRecognizer::HazardType AIEHazardRecognizer::getHazardType(
...
+    int DeltaCycles) const {
+  const MCInstrDesc &Desc = MI.getDesc();
+  const unsigned SchedClass =
+      TII->getSchedClass(Desc, MI.operands(), MI.getMF()->getRegInfo());
+  const MemoryBankBits MemoryBanks = getMemoryBanks(&MI);
+  return checkConflict(
+      Scoreboard, ItinData, SchedClass,
+      getSlotSet(Desc, *TII->getFormatInterface(), IgnoreUnknownSlotSets),
+      MemoryBanks, TII->getMemoryCycles(SchedClass), DeltaCycles, std::nullopt);
+}
```
gbossu, line 456: I'm probably being pedantic, but do we really need one more wrapper if it's called a single time within the new pipeliner code?

llvm/lib/Target/AIE/AIEInterBlockScheduling.cpp
```diff
@@ -210,36 +215,100 @@ void InterBlockScheduling::leaveFunction() {
...
   DEBUG_BLOCKS(dbgs() << "  >> enterBlock " << BB->getNumber() << " "
-                      << CurrentBlock->kindAsString() << " FixPointIter="
-                      << CurrentBlock->FixPoint.NumIters << "\n");
-}
+                      << CurrentBlockState->kindAsString() << " FixPointIter="
+                      << CurrentBlockState->FixPoint.NumIters
+                      << " II=" << CurrentBlockState->FixPoint.II << "\n");
+}
+namespace {
+class PipelineExtractor : public PipelineScheduleVisitor {
```
gbossu, line 235: Nit: Maybe add a high-level description?

I think we use this as a helper to insert instructions (or their clones) at the right places (prologue, steady state, epilogue)

llvm/lib/Target/AIE/AIEInterBlockScheduling.cpp
```diff
@@ -210,36 +215,100 @@ void InterBlockScheduling::leaveFunction() {
...
+  BlockState &Loop;
+  BlockState *Prologue = nullptr;
+  BlockState *Epilogue = nullptr;
+  MachineBundle CurrentBundle;
+  std::vector<MachineBundle> TimedRegion;
```
gbossu, line 240: Maybe note that `TimedRegion` can represent the epilogue, steady state, or epilogue depending on how far we are in the modulo scheduling.

llvm/lib/Target/AIE/AIEPostPipeliner.cpp
```diff
@@ -0,0 +1,352 @@
...
+void PostPipeliner::visitPipelineSchedule(
+    PipelineScheduleVisitor &Visitor) const {
+
+  // This runs StageCount times across the original body instructions and
+  // calls the bundle emission callbacks according to the Include filter.
+  // It provide the stage and the modulo cycle in that stage
+  // (both starting at zero) to the filter
+  auto ExtractSection =
+      [&](int StageCount,
+          std::function<bool(const NodeInfo &Node, int Stage, int M)> Include) {
```
gbossu: Super-nit: `Filter` instead of `Include` seems more conventional.

llvm/lib/Target/AIE/AIEInterBlockScheduling.cpp
```diff
@@ -210,36 +215,100 @@ void InterBlockScheduling::leaveFunction() {
...
+  }
+  void startCycle() override { CurrentBundle.clear(); }
+  void emit(MachineInstr *MI) override {
+    // We re-emit the original instructions into the loop body.
+    // Prologue and epilogue obtain copies.
+    MachineInstr *ToBeEmitted =
+        InLoop ? MI : Loop.TheBlock->getParent()->CloneMachineInstr(MI);
+    CurrentBundle.add(ToBeEmitted);
+  }
+  void endCycle() override { TimedRegion.emplace_back(CurrentBundle); }
```
gbossu: Super-nit: `Cycle` was confusing to me, because I interpreted it as a CFG cycle, i.e. the steady state. I'd find the interfaces clearer if we had `appendInBundle` and `emitBundle` instead of `startCycle` + `emit` + `endCycle`.

But well, that's probably personal taste at this point :)

llvm/lib/Target/AIE/AIEPostPipeliner.cpp
gbossu: Oh you can't (or shouldn't), but an optional value is better than using sentinels I'd say

llvm/lib/Target/AIE/AIEInterBlockScheduling.h
```diff
@@ -106,14 +107,25 @@ class InterBlockEdges {
...
 // handling.
 enum class BlockType { Regular, Loop, Epilogue };
+enum class SchedulingStage {
+  Scheduling,
+  SchedulingNotConverged,
+  SchedulingDone,
+  Pipelining,
+  PipeliningDone,
+  PipeliningFailed
+};
```
gbossu, line 134: Maybe give more details about the states and the legal transitions? In particular:
- Is `SchedulingNotConverged` an error state, or do we still have a way to schedule conservatively? In this case, is it equivalent to `SchedulingDone`?
- Similar questions for `PipeliningFailed`
- Is `SchedulingDone` equivalent to `Pipelining`, or is `SchedulingDone` a final state?

llvm/lib/Target/AIE/AIEMachineScheduler.cpp
```diff
@@ -549,24 +552,33 @@ void AIEPostRASchedStrategy::commitBlockSchedule(MachineBasicBlock *BB) {
...
+        }
+        It = std::next(Bundle.getInstrs().back()->getIterator());
       }
-      It = std::next(Bundle.getInstrs().back()->getIterator());
+      AIEHazardRecognizer::applyBundles(Region.Bundles, BS.TheBlock);
     }
-    AIEHazardRecognizer::applyBundles(Region.Bundles, BS.TheBlock);
   }
+  // swp prologue
+  InterBlock.emitInterBlockBottom(BS);
```
gbossu, line 581: Similar comment here: I'm wondering why this is called `emitInterBlockBottom` if we insert things at the top of the block?

llvm/lib/Target/AIE/AIEMachineScheduler.cpp
```diff
@@ -592,9 +604,12 @@ void AIEPostRASchedStrategy::enterRegion(MachineBasicBlock *BB,
 
 void AIEPostRASchedStrategy::leaveRegion(const SUnit &ExitSU) {
   LLVM_DEBUG(dbgs() << "    << leaveRegion\n");
-  materializeMultiOpcodeInstrs();
 
   auto &BS = InterBlock.getBlockState(CurMBB);
+  if (BS.FixPoint.II) {
```
gbossu: I think I still don't understand that early exit. Is it because when trying pipelining, there is no region conflicts to solve? Could that line be changed to `BS.getStage() == Pipelining`?

llvm/test/CodeGen/AIE/aie2/schedule/postpipeliner/bitwisexor.mir
martien-de-jong: Ooohh, clever!

llvm/test/CodeGen/AIE/aie2/schedule/postpipeliner/load-add-store-renamed.prepiplined.mir
martien-de-jong: This is a refactor glitch. I rerun update_tests on all commits and will push shortly.

llvm/lib/Target/AIE/AIEPostPipeliner.cpp
```diff
@@ -0,0 +1,352 @@
...
+}
+
+// Check resources. We only insert at the position modulo II. Since we insert
+// all iterations separately, the resources that wrap around accumulate in the
+// overflow area, causing conflicts when inserting future iterations
+int PostPipeliner::fit(MachineInstr *MI, int Earliest, int NTries, int II) {
+  for (int C = Earliest; C < Earliest + NTries; C++) {
+    int Mod = C % II;
+    LLVM_DEBUG(dbgs() << "   at " << C << " (" << Mod << ")\n");
+    if (!HR.checkConflict(Scoreboard, *MI, -Depth + Mod)) {
```
gbossu, line 143: Is that really enough to verify conflicts at `-Depth + Mod + II`, `-Depth + Mod + 2*II`, etc.? Or do we have a special scoreboard here that has a size of `II` and wraps around?

llvm/lib/Target/AIE/AIEBaseHardwareLoops.cpp
martien-de-jong: There's no 'our block placement'. The pass that's called like that only sets block alignment. The standard block placement is complicated, and choses to let a guard fall through to the exit block. Ultimately, two loops may have the same exit block and one of the two will lose.
Note that the jump would start a new region, which would act like a basic block boundary anyway.

llvm/lib/Target/AIE/AIEBaseHardwareLoops.cpp
martien-de-jong: Yes, that would be better. I will change this.

llvm/lib/Target/AIE/AIEInterBlockScheduling.cpp
```diff
@@ -343,54 +412,94 @@ MachineInstr *InterBlockScheduling::latencyConverged(BlockState &BS) const {
...
-    auto Res = BS.FixPoint.PerMILatencyMargin.try_emplace(MINeedsHigherCap, 0);
-    // Increase the latency margin per instruction, unless we already iterated
-    // more than MaxExpensiveIterations without converging.
-    if (BS.FixPoint.NumIters <= MaxExpensiveIterations) {
-      ++Res.first->second;
-    } else {
-      BS.FixPoint.LatencyMargin++;
+  BS.FixPoint.NumIters++;
+  int &II = BS.FixPoint.II;
+  if (!II) {
```
gbossu: That's now a big function with lots of `if`s. How about introducing two functions for checking the fixpoint state? E.g.

```
BS.FixPoint.NumIters++;

if (State== Scheduling)
  return checkLoopAwareScheduling();
return checkPipelining()
```

gbossu: (That would also help limit the noise in diff lines)

llvm/lib/Target/AIE/AIEInterBlockScheduling.cpp
```diff
@@ -343,54 +412,94 @@ MachineInstr *InterBlockScheduling::latencyConverged(BlockState &BS) const {
...
+        ++Res.first->second;
+      } else {
+        BS.FixPoint.LatencyMargin++;
+      }
+      DEBUG_LOOPAWARE(dbgs() << "  not converged: latency RM="
+                             << BS.FixPoint.ResourceMargin
+                             << " LM=" << BS.FixPoint.LatencyMargin
+                             << " MIM=" << Res.first->second << "\n");
+      // Iterate on CurMBB
+      return BS.FixPoint.Stage = SchedulingStage::Scheduling;
```
gbossu: Are you voluntarily re-assigning `BS.FixPoint.Stage`?

llvm/lib/Target/AIE/AIEHazardRecognizer.cpp
martien-de-jong: We could talk about this. I would vouch for dropping the less generic wrappers. I guess that public interfaces using Descr could be dropped in favour of those using MachineInstr. Descr used to be sufficient, but since we use operandinfo to compute memory banks and iteneraries, MachineInstr is more appropriate.

llvm/lib/Target/AIE/AIEInterBlockScheduling.cpp
gbossu: Shouldn't we already be at `SchedulingStage::Scheduling` when starting the loop-aware scheduling?

llvm/lib/Target/AIE/AIEInterBlockScheduling.cpp
```diff
@@ -343,54 +412,94 @@ MachineInstr *InterBlockScheduling::latencyConverged(BlockState &BS) const {
...
+  }
 
-  return true;
+  // Otherwise try a larger II.
+  // We cut off at the load latency because the main benefit from
+  // post-pipelining comes from wrapping latencies over multiple stages.
+  // This will only work if the stages are small enough. If we miss
+  // opportunities because of this, we should try to get the Prepipeliner
+  // to accept it.
+  if (++II <= 7) {
```
gbossu: Make that magic number a CL option?

llvm/lib/Target/AIE/AIEPostPipeliner.cpp
martien-de-jong: Hmm. I think I have explained in a number of places that the +II +2*II etc conflicts are accumulating by scheduling all iterations into steady state.

llvm/lib/Target/AIE/AIEInterBlockScheduling.cpp
martien-de-jong: Oh hell.  C++ should require an explicit operator to use an assignment as a value.
But it's not as ugly as I thought, we return the stage, not some boolean.  All returns are of this form to guarantee that the return value is the same as the stage that we save.

llvm/lib/Target/AIE/AIEInterBlockScheduling.cpp
martien-de-jong: Already did that.

llvm/lib/Target/AIE/AIEInterBlockScheduling.cpp
```diff
@@ -523,6 +633,29 @@ int InterBlockScheduling::getSafetyMargin(MachineBasicBlock *Loop,
...
+    if (Bundle.empty()) {
+      TII->insertNoop(*BB, Before);
+      continue;
+    }
+    bool InBundle = false;
+    for (auto *MI : Bundle.getInstrs()) {
+      if (Move) {
+        BB->remove_instr(MI);
+      }
+      BB->insert(Before, MI);
```
gbossu, line 659: Is there an equivalent `insert_instr`? Or is it on purpose so that we insert inside an existing bundle?

llvm/lib/Target/AIE/AIEInterBlockScheduling.cpp
```diff
@@ -763,8 +919,13 @@ void BlockState::classify() {
   // construction.
 }
 
-void BlockState::initInterBlock(const MachineSchedContext &Context) {
+void BlockState::initInterBlock(const MachineSchedContext &Context,
+                                const AIEHazardRecognizer &HR) {
   BoundaryEdges = std::make_unique<InterBlockEdges>(Context);
+  if (Regions.size() == 1) {
+    // Don't worry, this just constructs a mostly empty container class
```
gbossu, line 942: I like this note for the anxious programmer 😄

llvm/lib/Target/AIE/AIEMachineScheduler.cpp
martien-de-jong: Yep that's an oversight.

llvm/lib/Target/AIE/AIEPostPipeliner.cpp
```diff
@@ -0,0 +1,352 @@
...
+// that are still blocked from earlier copies are free.
+// The resource state is maintained in a ResourceScoreboard that is advances
+// by II for each copy.
+// The latency state is maintained in an 'Earliest' entry for each SUnit,
+// which is updated whenvever we schedule a predecessor of that SUnit.
+
+PostPipeliner::PostPipeliner(const AIEHazardRecognizer &HR, int NInstr)
+    : HR(HR), NInstr(NInstr) {}
+
+bool PostPipeliner::accept(MachineBasicBlock &LoopBlock) {
```
gbossu: Nit: `canAccept`?

llvm/lib/Target/AIE/AIEMachineScheduler.cpp
martien-de-jong: The epilogue is pushed into the top of the exit block.

llvm/lib/Target/AIE/AIEPostPipeliner.cpp
```diff
@@ -0,0 +1,352 @@
...
+bool PostPipeliner::scheduleFirstIteration() {
+  // Set up the basic schedule from the original instructions
+  for (int K = 0; K < NInstr; K++) {
+    const int N = K;
+    LLVM_DEBUG(dbgs() << "  Trying " << N << "\n");
+    SUnit &SU = DAG->SUnits[N];
+    MachineInstr *MI = SU.getInstr();
+    const int Earliest = Info[N].Earliest;
+    // The first iteration should fit entirely within II
+    const int Actual = fit(MI, Earliest, II, II);
```
gbossu, line 262: Shouldn't the third parameter be `II - Earliest` if you want to fit within `II`?

gbossu: Oh, I took that comment too literally, obviously we do not want the whole thing to fit within II, otherwise that would be a single stage. I don't think I understand the comment then. What are we protecting ourselves against here?

llvm/lib/Target/AIE/AIEPostPipeliner.cpp
gbossu: Yes I just read the rest of the code now. Coming back here, I can understand it is not needed. There's just a lot of changes to process at once :)

llvm/lib/Target/AIE/AIEPostPipeliner.cpp
```diff
@@ -0,0 +1,352 @@
...
+  // Let's not skimp on size here. This allows us to insert any instruction
+  // in the unrolled dag.
+  NCopies = NTotalInstrs / NInstr;
+  Depth = NCopies * II + HR.getPipelineDepth();
+  Scoreboard.reset(Depth);
+
+  Info.resize(NTotalInstrs);
+  LLVM_DEBUG(for (int I = 0; I < NInstr;
+                  I++) { dbgs() << I << " " << *DAG->SUnits[I].getInstr(); });
+  computeLoopCarriedEarliest();
```
gbossu: super-nit: The MachineScheduler does something similar and calls this `bias`. Maybe we could call this function `biasEarliesCyclesForLCDs`?

llvm/test/CodeGen/AIE/aie2/hardware-loops/split-zol-jmp.mir
```diff
@@ -0,0 +1,56 @@
...
+
+  bb.1:
+    successors: %bb.1, %bb.2
+    liveins: $r0
+    $r0 = ADD_NC $r0, 0
+    PseudoLoopEnd <mcsymbol .L_1120>, %bb.1
+    PseudoJ_jump_imm  %bb.2
+
+  bb.2 (align 16):
+    PseudoRET implicit $lr
```
gbossu, line 54: Nit: if the exit is already dedicated, I guess we do not need a new fall-through? But I guess that is not important because we later optimize the CFG anyway?

llvm/test/CodeGen/AIE/aie2/schedule/postpipeliner/add-store.mir
```diff
@@ -0,0 +1,130 @@
...
+  ; CHECK-NEXT:    nopb ; nopa ; nops ; nopxm ; nopv
+  ; CHECK-NEXT:    nopb ; nopa ; nops ; nopxm ; nopv
+  ; CHECK-NEXT:    nopb ; nopa ; nops ; nopxm ; nopv
+  ; CHECK-NEXT:    nopa ; nopb ; movx r1, #0; nopm
+  ; CHECK-NEXT:    add r1, r1, #1
+  ; CHECK-NEXT:    .p2align 4
+  ; CHECK-NEXT:  .LBB0_2: // %for.body
+  ; CHECK-NEXT:    // =>This Inner Loop Header: Depth=1
+  ; CHECK-NEXT:  .L_LEnd0:
+  ; CHECK-NEXT:    nopb ; nopa ; st r1, [p0], #4; add r1, r1, #1; nopm ; nopv
```
gbossu, line 40: 🎉

gbossu, line 40: 🎉

llvm/test/CodeGen/AIE/aie2/schedule/postpipeliner/bitwisenot.mir
```diff
@@ -0,0 +1,123 @@
...
+  ; CHECK-NEXT:    vldb wl0, [p0], #64
+  ; CHECK-NEXT:    vldb wh0, [p0, #32]
+  ; CHECK-NEXT:    vldb wl0, [p0], #64
+  ; CHECK-NEXT:    vldb wh0, [p0, #32]; vbneg_ltz.s16 x1, r21, x0
+  ; CHECK-NEXT:    vldb wl0, [p0], #64
+  ; CHECK-NEXT:  .LBB0_1: // %for.body
+  ; CHECK-NEXT:    // =>This Inner Loop Header: Depth=1
+  ; CHECK-NEXT:    vldb wh0, [p0, #32]; nopa ; vst wh1, [p1, #32]; nopx ; vbneg_ltz.s16 x1, r21, x0; nopv
+  ; CHECK-NEXT:  .L_LEnd0:
+  ; CHECK-NEXT:    vldb wl0, [p0], #64; nopa ; vst wl1, [p1], #64; nopxm ; nopv
```
gbossu, line 33: Very nice 🎉 🎉

llvm/lib/Target/AIE/AIEPostPipeliner.cpp
martien-de-jong: Ah. This comment has some history. fit() actually tries to fit in Earliest % II. If that wraps around, it effectively moves one stage further. So NTries = II just says "Try it in every position in the modulo schedule". More than that is useless.

llvm/lib/Target/AIE/AIEPostPipeliner.cpp
martien-de-jong: I already renamed this, because it does more.

llvm/lib/Target/AIE/AIEBasePipelinerLoopInfo.cpp
```diff
@@ -664,6 +669,8 @@ class ZeroOverheadLoop : public AIEBasePipelinerLoopInfo {
       SmallVectorImpl<MachineOperand> &Cond) override;
 
   bool canAcceptII(SMSchedule &SMS) override;
+
+  bool shouldUseSchedule(SwingSchedulerDAG &SSD, SMSchedule &SMS) override;
```
gbossu, line 676: We should probably update `canAcceptII` as well. Typically, this will increase the II until we have a low enough stage count. So I think that if we find a schedule with a low II and high stage count, we should immediately refuse it sop the post-pipeliner can pick it; and not increase the II.

gbossu (commented):
I think I'm at a point where I understand all individual pieces, and now I need to go over the code a final time to make sure I understand how it's all connected. But I couldn't find any blocker so far, great work!

llvm/test/CodeGen/AIE/aie2/hardware-loops/split-zol-jmp.mir
martien-de-jong: Well, the fallthrough also makes the body single-regioned, Programming around that is sordid. Also note that hwloop lowering sits just before postmisched.

llvm/lib/Target/AIE/AIEBasePipelinerLoopInfo.cpp
martien-de-jong: That's an interface change I think? it should be able to say yes, higherII and stop. That would definitely save some time.

gbossu: I guess we would accept the II if we are confident we can pick up the loop in the post-pipeliner. And for those loops, `shouldUseSchedule` would return `false`.

martien-de-jong: Somehow that makes sense, except for the name of canAcceptII. I will add a fat comment.

gbossu: I mean, something can accept it, just not the pre-pipeliner 😄

llvm/lib/Target/AIE/AIEInterBlockScheduling.cpp
martien-de-jong: I can never get my head around these issues. My assumption is that at this point we have no bundles, and we insert instructions before other instructions, leaving BundledWithPred markers to really create the bundles afterwards.

llvm/lib/Target/AIE/AIEInterBlockScheduling.cpp
gbossu: Did you fix this in a future commit? The diff is still introduced in 46a9ac2bde005b976eee70c59f5f8a98acd0e5db

llvm/lib/Target/AIE/AIEMachineScheduler.cpp
gbossu: I get that, but reading the function name really sounds like "we emit the top of the interblock schedule". I think emitInterblockPrologue and emitInterblockEpilogue is clearer. But nvm, I won't block for that.

llvm/lib/Target/AIE/AIEInterBlockScheduling.cpp
```diff
@@ -343,13 +421,32 @@ MachineInstr *InterBlockScheduling::latencyConverged(BlockState &BS) const {
   return nullptr;
 }
 
-bool InterBlockScheduling::updateFixPoint(BlockState &BS) {
+SchedulingStage InterBlockScheduling::updateFixPoint(BlockState &BS) {
+  if (BS.Kind != BlockType::Loop) {
+    return BS.FixPoint.Stage = SchedulingStage::SchedulingDone;
```
gbossu, line 426: I still find it confusing to return an assignment. The underlying value of the enum could change (and that will be an annoying thing to debug). How about 
```
BS.FixPoint.Stage = SchedulingStage::SchedulingDone;
return true;
```

Same for other cases when you assign and return at the same time.

gbossu: Clarified, I thought we were implicitly converting the enum to a `bool`, but the return type changed. All good.

llvm/lib/Target/AIE/AIEInterBlockScheduling.cpp
```diff
@@ -210,36 +219,105 @@ void InterBlockScheduling::leaveFunction() {
...
-                      << CurrentBlock->TheBlock->getNumber() << "\n");
+                      << CurrentBlockState->TheBlock->getNumber() << "\n");
   // After scheduling a basic block, check convergence to determine which block
   // to schedule next and with what parameters
-  auto &BS = *CurrentBlock;
-  if (BS.Kind == BlockType::Loop && !updateFixPoint(BS)) {
-    BS.FixPoint.NumIters++;
+  auto &BS = *CurrentBlockState;
+  switch (updateFixPoint(BS)) {
+  case SchedulingStage::SchedulingNotConverged:
```
gbossu, line 298: I guess this is just for completeness, but we will have crashed before checking `SchedulingStage::SchedulingNotConverged` here?

llvm/lib/Target/AIE/AIEMachineScheduler.cpp
```diff
@@ -1148,11 +1163,41 @@ void llvm::AIEPostRASchedStrategy::buildGraph(ScheduleDAGMI &DAG, AAResults *AA,
...
-  for (auto *I : Region) {
-    DAG.initSUnit(*I);
+  int NCopies = 1;
+  if (int II = BS.FixPoint.II) {
+    assert(BS.Kind == BlockType::Loop);
+    assert(BS.getRegions().size() == 1);
+    // Try to wrap the linear schedule within II.
+    // We virtually unroll the body by the stagecount, computed from rounding
+    // up the length divided by II.
+    NCopies = (BS.getScheduleLength() + II - 1) / II;
```
gbossu, line 1173: The "standard" schedule has been achieved using different scheduling techniques. Can we guarantee there will be enough copies if the linear schedule used for pipelining is different?

llvm/lib/Target/AIE/AIEPostPipeliner.cpp
```diff
@@ -0,0 +1,441 @@
...
+      LLVM_DEBUG(dbgs() << "Out of resources\n");
+      return false;
+    }
+    const int LocalCycle = Actual % II;
+    const MemoryBankBits MemoryBanks = HR.getMemoryBanks(MI);
+    LLVM_DEBUG(dbgs() << "  Emit in " << -Depth + LocalCycle << "\n");
+    int Cycle = -Depth + LocalCycle;
+    LLVM_DEBUG(dbgs() << "  Emit in " << Cycle << "\n");
+    HR.emitInScoreboard(Scoreboard, MI->getDesc(), MemoryBanks, MI->operands(),
+                        MI->getMF()->getRegInfo(), Cycle);
```
gbossu, line 274: I'm not debating for correctness, but I wonder if we should not immediately emit the resources at C+II, C+2II, etc. Otherwise, we will only see conflicts due to resources that "wrap around the II" when scheduling the next iterations. At that point, it's too late and we will try a larger II. Maybe if we could anticipate those resources, we could find a better linear schedule that has less conflicts when scheduling all copies.

martien-de-jong: I agree, but that can't be done with the current HazardRecognizer/Scoreboard, since I would need to push the modulo limit into the scoreboard handling. I will create a follow up to do that.

llvm/lib/Target/AIE/AIEInterBlockScheduling.cpp
martien-de-jong: Correct. Hence that I wrote 'may not be observable.` at the enum definition.

llvm/lib/Target/AIE/AIEPostPipeliner.cpp
gbossu: I mean, you know the II and NCopies, you could just have a `for` loop here instead of a single call to `HR.emitInScoreboard`? (Not in this PR, agreed)

llvm/lib/Target/AIE/AIEMachineScheduler.cpp
martien-de-jong: Good point. I could make it safe (for now) by rejecting NStages > NCopies. What do you think?

gbossu: Fine to me, I don't think that case will happen very often

llvm/lib/Target/AIE/AIEPostPipeliner.cpp
martien-de-jong: Ah well, it could be done by having enough depth in the scoreboard. Anyway, given that it doesn't impact correctness, I would push that to a follow up.

gbossu (dismissed):
I'm happy! This is massive change, but it all makes sense and this is a solid base for future improvements. I want it :)

llvm/lib/Target/AIE/AIEHazardRecognizer.cpp
gbossu: We'll fix interfaces in a follow-up. Closing this discussion

gbossu (approved):

llvm/lib/Target/AIE/AIEBaseHardwareLoops.cpp
```diff
@@ -173,6 +173,57 @@ char AIEBaseHardwareLoops::ID = 0;
...
+  // 4. add back the jump to the new block
+  // 5. fix successors
+  auto *MF = MBB.getParent();
+  MachineBasicBlock *NewBB = MF->CreateMachineBasicBlock(FBB->getBasicBlock());
+  MF->insert(++MBB.getIterator(), NewBB);
+  TII->removeBranch(MBB);
+  DebugLoc DL;
+  TII->insertBranch(MBB, TBB, nullptr, Cond, DL);
+  Cond.clear();
+  TII->insertBranch(*NewBB, FBB, nullptr, Cond, DL);
```
qcolombet, line 214: I think it would be easier to read if you use `TII->insertUnconditionalBranch(NewBB, FBB, DL)`.

llvm/lib/Target/AIE/AIEInterBlockScheduling.cpp
```diff
@@ -207,36 +219,105 @@ void InterBlockScheduling::leaveFunction() {
...
+public:
+  PipelineExtractor(InterBlockScheduling &InterBlock, BlockState &BS,
+                    const AIEBaseInstrInfo &TII)
+      : Loop(BS), CurrentBundle(TII.getFormatInterface()) {
+    MachineBasicBlock *LoopBlock = Loop.TheBlock;
+    for (auto *P : LoopBlock->predecessors()) {
+      if (P == LoopBlock) {
+        continue;
+      }
+      Prologue = &InterBlock.getBlockState(P);
```
qcolombet, line 279: Is the expectation that we have at most one non-`LoopBlock` predecessor/successor?
If so, could we assert that `Prologue/Epilogue == nullptr` before the assignment?

llvm/lib/Target/AIE/AIEPostPipeliner.h
```diff
@@ -0,0 +1,155 @@
...
+  MachineBasicBlock *Preheader = nullptr;
+
+  // The instruction defining the tripcount
+  MachineInstr *TripCountDef = nullptr;
+
+  // Basic modulo scheduling parameters
+  int NInstr;
+  int NCopies;
+  int II = 1;
+  int NStages = 0;
```
qcolombet, line 97: Is there a reason that this is `int` and not `unsigned`?

llvm/lib/Target/AIE/AIEPostPipeliner.h
```diff
@@ -0,0 +1,155 @@
...
+
+  /// Place SU in cycle Cycle; update Earliest of dependent instructions
+  void scheduleNode(SUnit &SU, int Cycle);
+
+  /// Compute the stage in which each instruction runs
```
qcolombet, line 102: Nit: To be able to upstream such code, all sentences must end with proper punctuatioin. Flagging this to make upstreaming code in general easier :). https://llvm.org/docs/CodingStandards.html#commenting

llvm/lib/Target/AIE/AIEInterBlockScheduling.cpp
```diff
@@ -340,13 +421,32 @@ MachineInstr *InterBlockScheduling::latencyConverged(BlockState &BS) const {
   return nullptr;
 }
 
-bool InterBlockScheduling::updateFixPoint(BlockState &BS) {
+SchedulingStage InterBlockScheduling::updateFixPoint(BlockState &BS) {
+  if (BS.Kind != BlockType::Loop) {
+    return BS.FixPoint.Stage = SchedulingStage::SchedulingDone;
```
qcolombet, line 426: To add on @gbossu returning an assignment is not a common pattern and I too read this as an `==` statement.

My suggestion would be to use the more idiomatic:
```
BS.FixPoint.Stage = SchedulingStage::SchedulingDone;
return BS.FixPoint.Stage;
```

Going beyond (I haven't checked if what I am about to suggest is correct), you could refactor the code to be:
```
if (...) {
  // update BS.FixPoint.Stage
} else if {
  // update BS.FixPoint.Stage differently
} else {
  ...
}
return BS.FixPoint.Stage;
```

I.e., do the updates like you do, but have a single return statement.

Then you can have methods like `updateScheduling` to either return void, or take a `const &` and return the new scheduling state to be assigned to the BlockState.

llvm/lib/Target/AIE/AIEInterBlockScheduling.cpp
```diff
@@ -207,36 +219,105 @@ void InterBlockScheduling::leaveFunction() {
...
+  }
+  case SchedulingStage::SchedulingDone:
+  case SchedulingStage::PipeliningFailed:
+    break;
   }
 
-  BS.setScheduled();
-  CurrentBlock = nullptr;
+  CurrentBlockState = nullptr;
   return true;
```
qcolombet, line 321: The `return true` for me is confusing because I thought it meant success (and false failure) but we return `true` when the scheduling stage is `PipeliningFailed`. Therefore, from a pure API standpoint it is unclear to me what the users of this API are supposed to do with the result.

I would at least update the comment on leaveBlock (in the header file) to clarify this.

llvm/lib/Target/AIE/AIEInterBlockScheduling.cpp
```diff
@@ -340,13 +421,32 @@ MachineInstr *InterBlockScheduling::latencyConverged(BlockState &BS) const {
...
+  if (BS.FixPoint.Stage == SchedulingStage::Scheduling) {
+    return updateScheduling(BS);
+  }
+
+  return updatePipelining(BS);
+}
+
+SchedulingStage InterBlockScheduling::updateScheduling(BlockState &BS) {
+  if (BS.FixPoint.NumIters >
+      MaxExpensiveIterations + 2 * HR->getConflictHorizon()) {
```
qcolombet, line 447: I feel it would be useful to keep the previous comment around (assuming, this is still accurate):
```
    // We will first try to increase the latency margin for one instruction at
    // a time, before increasing that margin for all instructions at once.
    // If we are very unlucky, we may step both the latency margin and
    // the resource margin to the max. Any more indicates failure to converge,
    // and we abort to prevent an infinite loop.
```

qcolombet (commented):
Late to the party and LGTM anyway :).

Just a few comments.

llvm/lib/Target/AIE/AIEInterBlockScheduling.cpp
martien-de-jong: Yes, we check/create these for Loop blocks. The assert would make sense.

llvm/lib/Target/AIE/AIEPostPipeliner.h
martien-de-jong: Not really. It is a personal preference for the signed integer datatype. For one, you can subtract 'normal' values without underflow.

llvm/lib/Target/AIE/AIEInterBlockScheduling.cpp
martien-de-jong: Right. It actually means 'fixpoint loop for this block is ready'. Pipelining is the last thing we try, and if we fail, we fall back to the schedule prepared earlier.

PR #148, opened 2024-08-09
Fix for some problems related to loops

andcarminati:
With this PR we remove 9 from 11 failures from QoR when using `-fno-unroll-loops`.

### DIFF
--- a/llvm/lib/Target/AIE/AIEBasePipelinerLoopInfo.cpp
+++ b/llvm/lib/Target/AIE/AIEBasePipelinerLoopInfo.cpp
@@ -726,7 +726,8 @@ std::optional<bool> ZeroOverheadLoop::createTripCountGreaterCondition(
 
 void ZeroOverheadLoop::adjustTripCount(int TripCountAdjust) {
   LLVM_DEBUG(dbgs() << "TripCountAdjust =  " << TripCountAdjust << "\n");
-  if (DefTripCount->getOperand(1).isImm()) {
+  if (DefTripCount->getOperand(1).isImm() &&
+      MRI.hasOneUse(DefTripCount->getOperand(0).getReg())) {
     // If we have a constant here, just update the value.
     const int64_t InitVal = DefTripCount->getOperand(1).getImm();
     DefTripCount->getOperand(1).setImm(InitVal + TripCountAdjust);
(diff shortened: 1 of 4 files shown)


llvm/lib/Target/AIE/AIEInterBlockScheduling.cpp
```diff
@@ -485,9 +485,11 @@ int InterBlockScheduling::getCyclesToRespectTiming(
     for (auto &Bundle : R.Bundles) {
       for (MachineInstr *MI : Bundle.getInstrs()) {
         DistancesFromLoopEntry[MI] = DistFromLoopEntry;
-        Edges.addNode(MI);
       }
-      ++DistFromLoopEntry;
```
konstantinschwarz, line 490: curious: any reason to change this from pre-increment to post-increment?

andcarminati: Hi, good eyes. Probably I deleted this increment and retyped it again, ! I will change it to be the same as before.

llvm/lib/Target/AIE/AIEInterBlockScheduling.cpp
```diff
@@ -485,9 +485,11 @@ int InterBlockScheduling::getCyclesToRespectTiming(
...
+      DistFromLoopEntry++;
+    }
+    for (auto MI = R.begin(); MI != R.end(); MI++) {
```
gbossu: Nit: Add a comment to mention that this is iterating in semantic order, not scheduled order.
Nit2: Can we use a loop like `for (const MachineInstr* MI : R) { ... }` instead?

gbossu (dismissed):
LGTM, great job finding those bugs!

llvm/lib/Target/AIE/AIEBasePipelinerLoopInfo.cpp
```diff
@@ -726,7 +726,8 @@ std::optional<bool> ZeroOverheadLoop::createTripCountGreaterCondition(
 
 void ZeroOverheadLoop::adjustTripCount(int TripCountAdjust) {
   LLVM_DEBUG(dbgs() << "TripCountAdjust =  " << TripCountAdjust << "\n");
-  if (DefTripCount->getOperand(1).isImm()) {
+  if (DefTripCount->getOperand(1).isImm() &&
```
andcarminati, line 729: This conflicts with https://github.com/Xilinx/llvm-aie/pull/146. As agreed with @martien-de-jong  his solution will cover this case, so this is just a temporary fix.

gbossu (approved):

PR #151, opened 2024-08-12
[AIE] Cast operand in AIELoopUtils assert

andcarminati:
We need to cast to the correct type.


andcarminati:
This PR is required to have a successful Windows build.

gbossu (approved):

PR #152, opened 2024-08-14
Loop-aware scheduling for loops with no dedicated exit

gbossu:
This cherry-picks two commits for loop-scheduling from #146

We can now do "loop-aware" scheduling for loops without dedicated exit block as well.

Only thing I changed from the original commits is to expand `auto CanFixLoopSchedule = [L = TheBlock, &IsLoop](auto *S)` into `auto CanFixLoopSchedule = [L = TheBlock, &IsLoop](const MachineBasicBlock *S)`

QoR effect is good, 0.5% improvement in isolation, more than 1% on my prototype enabling more software pipelining.


llvm/lib/Target/AIE/AIEInterBlockScheduling.cpp
```diff
@@ -83,10 +83,19 @@ void emitBundlesInScoreboardDelta(
...
-  MachineBasicBlock *SinglePredMBB = *MBB.predecessors().begin();
-  return SinglePredMBB;
+MachineBasicBlock *getLoopPredecessor(const MachineBasicBlock &MBB) {
+  if (MBB.pred_size() == 1) {
+    // if we have only one, it must be the loop
+    return *MBB.predecessors().begin();
+  }
+  // Otherwise, the loop is the fallthrough predecessor by construction
+  int FallThroughNumber = MBB.getNumber() - 1;
+  for (auto *Pred : MBB.predecessors()) {
```
andcarminati, line 94: nit: I think we can use `isLayoutSuccessor` instead of numbers.

andcarminati:
I left just one small comment, however it does not affect the correctness of this PR. For me it is good as is, so I will approve.

andcarminati (dismissed):

andcarminati (approved):

PR #154, opened 2024-08-15
[GISel][Combiner] Better insertion combiner WIP

ValentijnvdBeek:
A work in progress combiner that allows for the replacement of the insertion intrinsic for the more complicated patterns where there are multiple sources. They are tricky since we don't know what the size is of the vector we insert. This PR is a bit messy, I will fix it up later.


martien-de-jong:
Overtaken by other combiners

PR #156, opened 2024-08-19
Combine two VUPS_w2b into VUPS_x2c

abhinay-anubola:
Replaced v32acc32 sups and v16acc64 lups logic with builtins.

### DIFF
--- a/clang/lib/Headers/aiev2_ups.h
+++ b/clang/lib/Headers/aiev2_ups.h
@@ -94,36 +94,28 @@ inline v8cacc64 lups(v8cint16 a, int shft) { uint6_t sft = (uint6_t)(shft);
 #endif
 
 inline v16acc64 lups(v16int32 a, int shft, int sign) {
-  v16acc64 result = set_v16acc64(0, lups(extract_v8int32(a, 0), shft, sign));
-  result = insert(result, 1, lups(extract_v8int32(a, 1), shft, sign));
-  return result;
+  return __builtin_aiev2_acc64_v16_I512_ups(a, shft, sign);
 }
 inline v16acc64 lups(v16int32 a, int shft) {
   return lups(a, shft, __SIGN_SIGNED);
 }
 
 inline v16acc64 lups(v16uint32 a, int shft, int sign) {
-  v16acc64 result = set_v16acc64(0, lups(extract_v8uint32(a, 0), shft, sign));
-  result = insert(result, 1, lups(extract_v8uint32(a, 1), shft, sign));
-  return result;
+  return __builtin_aiev2_acc64_v16_I512_ups(a, shft, sign);
 }
 inline v16acc64 lups(v16uint32 a, int shft) {
   return lups(a, shft, __SIGN_UNSIGNED);
 }
 
 inline v32acc32 sups(v32int16 a, int shft, int sign) {
-  v32acc32 result = set_v32acc32(0, sups(extract_v16int16(a, 0), shft, sign));
-  result = insert(result, 1, sups(extract_v16int16(a, 1), shft, sign));
-  return result;
+  return __builtin_aiev2_acc32_v32_I512_ups(a, shft, sign);
 }
 inline v32acc32 sups(v32int16 a, int shft) {
   return sups(a, shft, __SIGN_SIGNED);
 }
 
 inline v32acc32 sups(v32uint16 a, int shft, int sign) {
-  v32acc32 result = set_v32acc32(0, sups(extract_v16uint16(a, 0), shft, sign));
-  result = insert(result, 1, sups(extract_v16uint16(a, 1), shft, sign));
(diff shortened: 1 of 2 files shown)


llvm/lib/Target/AIE/AIEPostSelectOptimize.cpp
```diff
@@ -266,6 +267,22 @@ bool combineINSERT_SUBREG(MachineBasicBlock &MBB) {
...
 }
 
+bool combineVUPS(MachineBasicBlock &MBB) {
+  MachineRegisterInfo &MRI = MBB.getParent()->getRegInfo();
+  const TargetInstrInfo &TII = *MBB.getParent()->getSubtarget().getInstrInfo();
+  auto *AIETII = static_cast<const AIEBaseInstrInfo *>(&TII);
+  bool Changed = false;
+  for (MachineInstr &MI : make_early_inc_range(reverse(MBB))) {
+    if (MI.getOpcode() != TargetOpcode::REG_SEQUENCE ||
+        MI.getNumOperands() != 5)
```
gbossu: Can you fold `MI.getNumOperands() != 5` into the target specific implementation? I think the only requirement for `combineLoHiMI` should be that MI is a REG_SEQUENCE

llvm/lib/Target/AIE/AIE2InstrInfo.cpp
```diff
@@ -1305,3 +1305,57 @@ bool AIE2InstrInfo::canHoistCheapInst(const MachineInstr &MI) const {
...
+      break;
+    case AIE2::VUPS_S64_S32_mv_ups_w2b:
+      CombinedOpCode = AIE2::VUPS_S64_S32_mv_ups_x2c;
+      break;
+    default:
+      return false;
+    }
+    Register Dst = MI.getOperand(0).getReg();
+    Register Src = MRI.createVirtualRegister(&AIE2::VEC512RegClass);
+    Register Shift = HiMI->getOperand(2).getReg();
```
gbossu: We should also verify that the Shift amount is the same

llvm/lib/Target/AIE/AIE2InstrInfo.cpp
```diff
@@ -1305,3 +1305,57 @@ bool AIE2InstrInfo::canHoistCheapInst(const MachineInstr &MI) const {
          AIE2::eSRegClass.hasSubClassEq(
              MRI.getRegClass(MI.getOperand(0).getReg()));
 }
+
+bool AIE2InstrInfo::combineLoHiMI(MachineInstr &MI,
+                                  MachineRegisterInfo &MRI) const {
```
gbossu: I feel a lot of the machinery here could be share between AIE versions. Could you maybe move the MI creation/manipulation to the PostSelectOptimize pass? Then I would suggest only implementing an interface like below for each AIEInstrInfo version. E.g. for AIE2:

```
struct LoHiCombineInfo {
  unsigned CombinedOpcode;
  unsigned OpIdxToAggregate;
};

std::optional<LoHiCombineInfo> AIE2InstrInfo::getCombinedLoHiOpcode(LoMI, HiMI) {
  unsigned LoOpcode = LoMI.getOpcode();
  unsigned HIOpcode = HiMI.getOpcode();
  if (LoOpcode != HiOpcode)
    return {};

  switch (LoOpcode) {
  case AIE2::VUPS_S32_D16_mv_ups_w2b:
    if (LoMI.getOperand(2) == HiMI.getOperand(2))
      return LoHiCombineInfo{VUPS_S32_D16_mv_ups_x2c, /*OpIdxToAggregate=*/1}
    break;
  case ...
  }
  return {}
}

andcarminati: Maybe `LoHiCombineInfo` could also include register classes for input and output.....

konstantinschwarz:
Do we see this missing combine in real world code?
I think we discovered a missing ,combine in aiev2_ups.h: https://github.com/Xilinx/llvm-aie/blob/aie-public/clang/lib/Headers/aiev2_ups.h#L115
Might be easier to just use the proper builtin there?
@abnikant has more details

gbossu:
> Do we see this missing combine in real world code? I think we discovered a missing ,combine in aiev2_ups.h: https://github.com/Xilinx/llvm-aie/blob/aie-public/clang/lib/Headers/aiev2_ups.h#L115 Might be easier to just use the proper builtin there? @abnikant has more details

I spotted that missing combine in `SigmoidTemplated_int8`, maybe other benchmarks as well. If we can solve it at the api level, I agree this is a good enough fix.

gbossu (commented):
LGTM, but maybe you could sqush the two commits together? I think it's always better to have the change and the corresponding tests in the same commit.

gbossu (approved):

PR #159, opened 2024-08-20
RFC: Reorganize AIELegalizerInfo

konstantinschwarz:
This factors out the custom legalization actions to be re-usable by AIE specific LegalizerInfo objects.
Moreover, this splits this AIELegalizerInfo object into AIE1LegalizerInfo and AIE2LegalizerInfo objects to get rid of numerous ST.isAIE2() checks, which make the code hard to read and difficult to port to a new architecture version.

There will be some duplication in the declarative legalization action specification for each architecture.


llvm/lib/Target/AIE/AIEBaseInstrInfo.h
```diff
@@ -84,6 +84,23 @@ struct AIEBaseInstrInfo : public TargetInstrInfo {
   virtual unsigned getCycleSeparatorOpcode() const {
     llvm_unreachable("Target didn't implement getCycleSeparatorOpcode");
   }
+  /// Return the opcode to be used for pushing a vector element at the LSB
+  /// position in a vector
+  virtual unsigned getGenericAddVectorEltOpcode() const { return -1; }
```
gbossu: How about `llvm_unreachable()` to be consistent with other functions?

llvm/lib/Target/AIE/AIEBaseInstrInfo.h
```diff
@@ -84,6 +84,23 @@ struct AIEBaseInstrInfo : public TargetInstrInfo {
...
+  /// \param signExt True if the extracted element shall be sign extended
+  virtual unsigned getGenericExtractVectorEltOpcode(bool SignExt) const {
+    return -1;
+  }
+  /// Return the opcode to be used for padding undefined values in the high bits
+  /// of a vector
+  virtual unsigned getGenericPadVectorOpcode() const { return -1; }
+  /// Return the opcode to be used for extracting a smaller vector by ignoring
+  /// the high bits
+  virtual unsigned getGenericUnpadVectorOpcode() const { return -1; }
```
gbossu: Nit: I feel we are really overloading InstrInfo with all those `getXXXXopcode()` function. At least for generic instrucitons, I'm wondering if we should not create e.g.
```
class AIEGenericInstructionsInterface {
  virtual unsigned getInsertVectorEltOpcode() const {...}
  virtual unsigned getExtractVectorEltOpcode() const {...}
  ...
}
```

I guess sometimes we will only really need to pass `AIEGenericInstructionsInterface`, and not the whole `AIEBaseInstrInfo`. This will also make unit testing easier, because we could just define a test interface for what is required to be tested.

What do you think?

konstantinschwarz: I don't really see an issue with adding more of these getter functions to InstrInfo, its a single place where we collect all architecture specific opcode information.

Now for AIE specific generic opcodes we could also consider them to be always available on every AIE architecture, and requiring them to be defined as the first opcodes after `GENERIC_OP_END`.
Then it doesn't matter which AIE namespace we use to access those opcodes (`AIE::` or `AIE2::` at this point).
We would have to be extremely careful not to introduce an opcode in between GENERIC_OP_END and the shared opcodes only for a single AIE version. Not sure that's any better

llvm/lib/Target/AIE/AIELegalizerHelper.h
```diff
@@ -0,0 +1,70 @@
+//===- AIELegalizerHelper ---------------------------------------*- C++ -*-===//
+//
+// This file is licensed under the Apache License v2.0 with LLVM Exceptions.
+// See https://llvm.org/LICENSE.txt for license information.
+// SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+//
+// (c) Copyright 2023-2024 Advanced Micro Devices, Inc. or its affiliates
```
andcarminati: nit: AIELegalizerHelper.cpp states just `2024`

llvm/lib/Target/AIE/AIELegalizerHelper.h
```diff
@@ -0,0 +1,70 @@
+//===- AIELegalizerHelper ---------------------------------------*- C++ -*-===//
```
andcarminati: nit: `.h`

llvm/lib/Target/AIE/AIELegalizerHelper.cpp
```diff
@@ -951,8 +957,9 @@ bool AIELegalizerHelper::legalizeG_FADDSUB(LegalizerHelper &Helper,
 
   const Register ExtEltDstReg = MRI.createGenericVirtualRegister(S32);
   const Register ExtDstReg = MRI.createGenericVirtualRegister(S32);
-  MIRBuilder.buildInstr(AIE2::G_AIE_SEXT_EXTRACT_VECTOR_ELT, {ExtEltDstReg},
-                        {Conv, IdxReg});
+  unsigned ExtractEltOpc =
```
andcarminati: nit: `const unsigned`

llvm/lib/Target/AIE/AIELegalizerHelper.cpp
```diff
@@ -927,14 +932,15 @@ bool AIELegalizerHelper::legalizeG_FADDSUB(LegalizerHelper &Helper,
...
 
-  Register NewSrc1 = MIRBuilder
-                         .buildInstr(AIE2::G_AIE_INSERT_VECTOR_ELT, {V16FP32},
-                                     {Src1Vec, NewSrcReg1, IdxReg})
-                         .getReg(0);
-  Register NewSrc2 = MIRBuilder
-                         .buildInstr(AIE2::G_AIE_INSERT_VECTOR_ELT, {V16FP32},
-                                     {Src2Vec, NewSrcReg2, IdxReg})
-                         .getReg(0);
+  unsigned InsertEltOpc = ST.getInstrInfo()->getGenericInsertVectorEltOpcode();
```
andcarminati: nit: `const unsigned`

andcarminati:
I think `Reorganization` can be used instead of `Refactor` because we are doing something like the opposite here. Also, I like the idea of splitting the code because it makes our life easier when we write aie2 legalization rules (we don't need to think about the disjoint behavior of the two versions).

niwinanto (dismissed):
LGTM!

llvm/lib/Target/AIE/AIELegalizerHelper.h
ValentijnvdBeek: 2023-2024 is in this correct, I think, you are moving AMD code from 2023 into this file, so it should keep the same date range.

llvm/lib/Target/AIE/AIELegalizerHelper.h
```diff
@@ -0,0 +1,70 @@
...
+class MachineInstr;
+
+class AIELegalizerHelper {
+  const AIEBaseSubtarget &ST;
+
+public:
+  AIELegalizerHelper(const AIEBaseSubtarget &ST);
+  const AIEBaseInstrInfo *getInstrInfo();
+
+  bool legalizeG_VASTART(LegalizerHelper &Helper, MachineInstr &MI) const;
```
ValentijnvdBeek, line 33: Should the legalizations be public? I don't think that they should be outside of the classes that inherit from it. Maybe make them protected instead?

llvm/lib/Target/AIE/AIELegalizerInfo.cpp
```diff
@@ -99,7 +86,7 @@ negatePredicate(const std::function<bool(const LegalityQuery &)> &Func) {
   return [=](const LegalityQuery &Query) { return !Func(Query); };
 }
 
-AIELegalizerInfo::AIELegalizerInfo(const AIEBaseSubtarget &ST) {
+AIELegalizerInfo::AIELegalizerInfo(const AIEBaseSubtarget &ST) : AIEHelper(ST) {
```
ValentijnvdBeek: I think this breaks the principle of least surprise, other LLVM passes don't directly inherit from their Helper. Maybe changing the name to AIELegalizerPass or AIEBaseLegalizer would make it clearer that they are directly related rather than it being a utility class?

llvm/lib/Target/AIE/AIEBaseInstrInfo.h
ValentijnvdBeek: :+1: for making it easier, I know that I made that mistake a couple of times and it hard to catch. An alternative design would be to define `getOpcode` that takes an opcode in the `TargetOpcode` space and turns it into `AIE1`, `AIE2` or whatever.

llvm/lib/Target/AIE/AIE1LegalizerInfo.cpp
```diff
@@ -0,0 +1,299 @@
...
+
+  // 1024-bit vectors
+  const LLT V32S32 = LLT::fixed_vector(32, 32);
+  const LLT V64S16 = LLT::fixed_vector(64, 16);
+  const LLT V128S8 = LLT::fixed_vector(128, 8);
+
+  // 1024-bit accumulators
+  const LLT ACC1024 = LLT::fixed_vector(16, 64);
+
+  const LLT S128 = LLT::scalar(128);
```
ValentijnvdBeek, line 67: Nit: maybe put these in a separate header with if guards per generation? It would nice for these to be global in the file and these definitions will probably overlap quite a bit and you'd want to include it more parts of the instruction selection stack.

llvm/lib/Target/AIE/AIE1LegalizerInfo.cpp
```diff
@@ -0,0 +1,299 @@
...
+
+  auto &IMPLICIT = getActionDefinitionsBuilder({G_IMPLICIT_DEF, G_FREEZE})
+                       .legalFor({S20, S32, P0, S128});
+
+  IMPLICIT.widenScalarToNextPow2(0).clampScalar(0, S32, S32);
+
+  getActionDefinitionsBuilder(G_CONSTANT)
```
ValentijnvdBeek, line 74: These is a decent overlap of these definitions, which is a shame since it adds duplication. Maybe we can reuse the trick you used for the opcodes and define a base class with `createXXXActionDefinitions` calls which is overloaded by the architecture when needed. Then you only get the divergence when you need it.

Disclaimer: didn't test this yet

llvm/lib/Target/AIE/AIE1LegalizerInfo.h
```diff
@@ -0,0 +1,37 @@
+//===- AIELegalizerInfo -----------------------------------------*- C++ -*-===//
```
andcarminati: nit: `.h`

ValentijnvdBeek (commented):
Hi Konstantin, 

Thanks for doing this work, it was something that was on my todo list that I didn't get time to do. Since I did spent a decent chunk of time spelunking in the legalizer, I thought it might be useful for me to chime in.

It is a good reorganization, I am happy to see the code being split up and the if statements being removed. There is some ground to be gained in terms of duplication, since this move will cause a decent amount of copying back and forth. As shown in #160, the legalizer isn't necessarily feature complete for each opcode yet, which means that both legalizers would need to be updated to deal with changes like this. So I am unsure if the DRY-complexity tradeoff here is optimal.

As to fix it, there are two ideas that I have toyed with and also put here as review comments:

1. Move the type definitions into a separate header that either imports the right subheader or does a few if guards. Basic types will likely be shared and this has the extra benefit of just being able to import `aieX_types.h` through the GISel stack. Now these definitions are given ad-hoc when they are needed.

2. Reuse the same trick you use for the opcodes and use that for the legalizer statements themselves. This is the [template pattern](https://refactoring.guru/design-patterns/template-method), where each opcode is a function call that the subclass can override when it is needed. This gets rid of both the `isAIE` calls, but allows you define "default" rules.

But, besides that, great work!

TL;DR This is a good refactor that solves a decent of the problems, but I think that a bit more can be won by separating the LLT variables and applying the template pattern

llvm/lib/Target/AIE/AIELegalizerHelper.h
konstantinschwarz: They need to be public so they can be called

llvm/lib/Target/AIE/AIELegalizerInfo.cpp
konstantinschwarz: This is just initializing the class members in C++ (member initializer lists)

We inherit from LegalizerInfo as usual, see here: https://github.com/Xilinx/llvm-aie/pull/159/files#diff-6e5726c037cb4f260ca385b0b7b2416bfd69e0478c713239dbf134f7e0b72c13R28

llvm/lib/Target/AIE/AIE1LegalizerInfo.cpp
konstantinschwarz: I did consider this, e.g. through the common [CRTP](https://en.cppreference.com/w/cpp/language/crtp) pattern.
I felt the added complexity isn't really worth it, much of the legalization is straight forward. Being able to share the custom legalization code seemed like a good compromise to me

llvm/lib/Target/AIE/AIEBaseInstrInfo.h
konstantinschwarz: The problem is that these "target specific generic machine instructions" do not exist in the `TargetOpcode` namespace, so we cannot translate them into the target namespace.

`TargetOpcode::` only contains the instructions up to `GENERIC_OP_END`, which also marks the beginning of the target specific (generic) machine instructions

andcarminati (approved):
LGTM.

gbossu (approved):
LGTM! Thanks for cleaning things up and preparing the future 🚀

llvm/lib/Target/AIE/AIELegalizerInfo.cpp
ValentijnvdBeek: Yeah, that makes sense. Those two thins look alike, my bad, sorry. That also deals with setting public/protected

llvm/lib/Target/AIE/AIEBaseInstrInfo.h
ValentijnvdBeek: Yeah, but the idea could have a generic namespace that is an amalgamation of the different architectures. That has different drawbacks, like keeping it up to date and it becoming very large, but it is an alternative. Both are good to me and I happy with his as well

llvm/lib/Target/AIE/AIE1LegalizerInfo.cpp
ValentijnvdBeek: Yeah, I can see that. It is very much a DRY-code complexity trade off. My worry is mainly in having to deal with the duplication if the number of variants becomes very large. Then changes to the legalizer might require changes to 10 different files which need to be kept into sync. But that is a worry for the long term and something that can be dealt with then.

PR #161, opened 2024-08-21
[AIEX] AIEEliminateDuplicatePHI pass - Add a check for valid next MI.

SagarMaheshwari99:


llvm/lib/Target/AIE/AIEEliminateDuplicatePHI.cpp
```diff
@@ -124,9 +124,14 @@ class AIEEliminateDuplicatePHI : public MachineFunctionPass {
...
       if (!isPointerTypePHI(MRI, MI))
         continue;
-      for (auto &PHI : make_early_inc_range(
-               make_range(std::next(MachineBasicBlock::instr_iterator(MI)),
-                          MachineBasicBlock::instr_iterator(*LastPHI)))) {
+
+      // Ensure that next MI is valid
+      auto NextMI = std::next(MachineBasicBlock::instr_iterator(MI));
+      if (NextMI == MBB.instr_end() || NextMI == LastPHI)
+        continue;
```
andcarminati: nit: If we reached the the end, should we simply break here?

SagarMaheshwari99: yeah probably.. I'll change it.

andcarminati (approved):
LGTM.

PR #163, opened 2024-08-21
[AIEX] Loop-aware scheduler: finer convergence for resource conflicts

gbossu:
There are a couple of "preparation commits", before the real change in the final commit.

FYI @krishnamtibrewala this will conflict with #144 

This brings tiny QoR improvements: around 0.05%. But on my prototype branch which allows more pipelining, this brings an extra `0.22%` improvement, no regression.


llvm/lib/Target/AIE/AIEInterBlockScheduling.cpp
```diff
@@ -85,6 +85,64 @@ void emitBundlesInScoreboardDelta(
...
+  for (const MachineBundle &B : reverse(MinBundles)) {
+    for (MachineInstr *MI : B.getInstrs())
+      HR.emitInScoreboard(Scoreboard, MI->getDesc(), HR.getMemoryBanks(MI), 0);
+    Scoreboard.recede();
+  }
+  return Scoreboard;
+}
+
+/// Replay the bundles in \p SuccBundles bottom-up, followed by those in
+/// \p PredBundles. If that causes a resource conflict, return the instruction
```
gbossu: I need to adjust the comment into something like `Replay the \p PredBundles bottom-up into \p ScoreBoard.`

llvm/lib/Target/AIE/AIEInterBlockScheduling.cpp
```diff
@@ -85,6 +85,64 @@ void emitBundlesInScoreboardDelta(
...
+      Bundles.begin(), Bundles.begin() + std::min(NumBundles, RequiredCycles));
+  for (const MachineBundle &B : reverse(MinBundles)) {
+    for (MachineInstr *MI : B.getInstrs())
+      HR.emitInScoreboard(Scoreboard, MI->getDesc(), HR.getMemoryBanks(MI), 0);
+    Scoreboard.recede();
+  }
+  return Scoreboard;
+}
+
+/// Replay the bundles in \p SuccBundles bottom-up, followed by those in
```
andcarminati: ~~I was trying to follow this comment in the code, but it seems that we are not replaying~~ `SuccBundles`, just `PredBundles`.

Edit: understood.

llvm/test/CodeGen/AIE/aie2/schedule/loopaware/fixpoint.mir
```diff
@@ -29,10 +29,8 @@ body:             |
   ; CHECK-NEXT: {{  $}}
   ; CHECK-NEXT:   $r10 = OR $r0, $r1
   ; CHECK-NEXT:   $r3 = OR killed $r10, $r2
-  ; CHECK-NEXT:   BUNDLE implicit-def $cm3, implicit-def $bml3, implicit-def $amll3, implicit-def $amlh3, implicit-def $bmh3, implicit-def $amhl3, implicit-def $amhh3, implicit-def $cm2, implicit-def $bml2, implicit-def $amll2, implicit-def $amlh2, implicit-def $bmh2, implicit-def $amhl2, implicit-def $amhh2, implicit $cm1, implicit $cm0, implicit killed $r3 {
-  ; CHECK-NEXT:     $cm3 = VMOV_mv_cm $cm1
-  ; CHECK-NEXT:     $cm2 = VADD $cm0, $cm1, killed $r3
-  ; CHECK-NEXT:   }
+  ; CHECK-NEXT:   $cm2 = VADD $cm0, $cm1, killed $r3
```
andcarminati, line 32: This is interesting! It looks like we had a CM_WM port conflict in the second iteration before your change.

gbossu: Yes were were not checking for long instructions whose FuncUnits might cross more than one iteration

gbossu:
FYI @andcarminati  I have rebased on `aie-public` to fix the conflicts from #144. I have also re-ordered the commits, `[AIEX] New baseline tests for loop-aware scheduling` is now one of the first commits to make the scheduling changes of `[AIEX] Update resource convergence check for loop-aware scheduling` clearer.

llvm/lib/Target/AIE/AIEHazardRecognizer.h
```diff
@@ -179,12 +179,13 @@ class AIEHazardRecognizer : public ScheduleHazardRecognizer {
   /// For efficiency, this size is rounded up to a power of two.
   unsigned computeScoreboardDepth() const;
 
-protected:
   ScheduleHazardRecognizer::HazardType
```
andcarminati, line 182: Hi, is there any reason to change the protected scope?

gbossu: Yes, I'm accessing this from `AIEInterblockScheduling`. This is now consistent with `emitInScoreboard()` which also supports an external scoreboard, and is also public.

llvm/test/CodeGen/AIE/aie2/schedule/loopaware/short-hwloop-resources.mir
```diff
@@ -0,0 +1,191 @@
...
+  ; NO-LOOPAWARE-SCOREBOARD-NEXT:   successors: %bb.1(0x40000000), %bb.2(0x40000000)
+  ; NO-LOOPAWARE-SCOREBOARD-NEXT:   liveins: $s1, $p2
+  ; NO-LOOPAWARE-SCOREBOARD-NEXT: {{  $}}
+  ; NO-LOOPAWARE-SCOREBOARD-NEXT:   $cm2, $p2 = VLDA_UPS_S32_S8_ag_pstm_nrm_imm $s1, killed $p2, 32, implicit-def $srups_of, implicit $crsat
+  ; NO-LOOPAWARE-SCOREBOARD-NEXT:   PseudoLoopEnd %bb.2, %bb.1
+  ; NO-LOOPAWARE-SCOREBOARD-NEXT: {{  $}}
+  ; NO-LOOPAWARE-SCOREBOARD-NEXT: bb.2:
+  ; NO-LOOPAWARE-SCOREBOARD-NEXT:   liveins: $p2, $s2
+  ; NO-LOOPAWARE-SCOREBOARD-NEXT: {{  $}}
+  ; NO-LOOPAWARE-SCOREBOARD-NEXT:   $amlh5, $p2 = VLDA_dmw_lda_am_ag_pstm_nrm_imm killed $p2, 2016
```
andcarminati, line 48: Just for clarification, we booked some cycles to avoid loop-caried resource conflicts to the epilogue, right? In this case the comment mention two nops here.  I think we have more than enough nops here....

gbossu: Not exactly, there should be two NOPs before `VLDA_dmw_lda_am_ag_pstm_nrm_imm` to make sure the epilogue cannot have resource conflicts with **any** of the previous iterations of the loop. They are not there because in epilogue scheduling we only account for the **last** iteration of the loop. That is safe to do in the normal case because the loop scheduling itself ensures that scoreboard resources never stick out of more than one iteration. it's just that the scoreboard is disabled for the `NO-LOOPAWARE-SCOREBOARD` label, just to show how epilogue scheduling can get complicated :)

andcarminati: Understood! thank you very much!

llvm/lib/Target/AIE/AIEInterBlockScheduling.cpp
```diff
@@ -68,23 +68,64 @@ void emitBundlesInScoreboard(const std::vector<MachineBundle> &Bundles,
...
+  ResourceScoreboard<FuncUnitWrapper> Scoreboard;
+  Scoreboard.reset(HR.getMaxLookAhead());
 
-    for (MachineInstr *MI : Bundle.getInstrs())
-      HR->emitInScoreboard(Scoreboard, MI->getDesc(), HR->getMemoryBanks(MI),
-                           MI->operands(), MI->getMF()->getRegInfo(), Delta);
+  // We have less known bundles than the minimum number required for
+  // correctness. Conservatively block some cycles.
+  if (NumBundles < RequiredCycles) {
+    unsigned CyclesToBlock = RequiredCycles - Bundles.size();
```
andcarminati, line 83: nit: subtract directly `NumBundles` instead of `Bundles.size()`.

llvm/lib/Target/AIE/AIEInterBlockScheduling.cpp
```diff
@@ -353,6 +366,22 @@ bool InterBlockScheduling::updateFixPoint(BlockState &BS) {
     return false;
   }
 
+  if (MachineInstr *MINeedsHigherCap = resourcesConverged(BS);
+      InterBlockScoreboard && MINeedsHigherCap) {
+    auto Res = BS.FixPoint.PerMILatencyMargin.try_emplace(MINeedsHigherCap, 0);
```
andcarminati, line 371: nit: Can we use another name for this `PerMILatencyMargin`? In feel it to tied to latency. Now it encloses a bigger scope.

andcarminati (approved):
LGTM. Also including nice refactoring ;-). Just some optional nits.

llvm/lib/Target/AIE/AIEInterBlockScheduling.cpp
gbossu: I got the same feeling, but in the end we "push" instructions up by increasing their latency to `ExitSU` so I kept the name. Do you have some name suggestion? Maybe something like `MILatencyToExitSU` would be clearer?

andcarminati: I think it is fine to leave as is, or maybe something neutral, like `MIDistanceToExitSU`.

PR #164, opened 2024-08-21
[AIEX] Update AIE base frame lowering class to accept stack alignment

niwinanto:
It would be nice if child classes of `AIEBaseFrameLowering` can set the stack alignment. It helps the targets with stack alignment requirement other than 32.


llvm/lib/Target/AIE/AIE2FrameLowering.h
```diff
@@ -23,7 +23,10 @@ class AIE2Subtarget;
 class AIE2FrameLowering : public AIEBaseFrameLowering {
 public:
   explicit AIE2FrameLowering(const AIEBaseSubtarget &STI)
-      : AIEBaseFrameLowering(STI) {}
+      : AIEBaseFrameLowering(
+            STI,
+            // A guess, based on code generated by the target compiler
```
andcarminati: I think this comment is a bit confusing. What do you mean by target compiler?

niwinanto: True, I just copy pasted from `AIEBaseFrameLowering`, removed it.

abnikant (approved):

PR #172, opened 2024-08-28
[AIE2] Fix missing VST.SRS combine

abhinay-anubola:
Combined into VST.SRS, if two 256-bit SRS are input to 512-bit STORE through CONCAT

### DIFF
--- a/llvm/lib/Target/AIE/AIE2InstructionSelector.cpp
+++ b/llvm/lib/Target/AIE/AIE2InstructionSelector.cpp
@@ -82,6 +82,16 @@ struct LoadStoreOpcodes {
   std::optional<unsigned> OffsetOpcode;
 };
 
+/// Hold the information needed to select 512-bit instructions that can be
+/// combined into VST.SRS
+struct StoreSRS512BitInfo {
+  // Hold the information of Store
+  LoadStoreOpcodes LSO;
+  // Registers required to select VST.SRS instruction
+  Register SrcReg, ShftReg, SignReg;
+  bool ConstantSign;
+};
+
 /// Create a memory operand representing Tile Memory
 /// \param I Instruction to build it for
 /// \param Mode Store or Load
@@ -188,10 +198,14 @@ public:
                                     AddressingModeInfo &AMI,
                                     MachineRegisterInfo &MRI);
   bool selectG_AIE_STORE_SRS(MachineInstr &StoreI, MachineRegisterInfo &MRI);
-  bool select512BitG_AIE_STORE_SRS(LoadStoreOpcodes &LSO,
-                                   AddressingModeInfo &AMI, Register SrcReg,
-                                   Register ShftReg, Register SignReg,
-                                   bool ConstantSign, MachineRegisterInfo &MRI);
+  bool selectG_AIE_STORE_CONCAT_SRS(MachineInstr &StoreI,
+                                    MachineRegisterInfo &MRI);
+  std::tuple<Register, Register> split512BitReg(Register SrcReg,
+                                                MachineRegisterInfo &MRI);
+  bool select512BitG_AIE_STORE_SRS(StoreSRS512BitInfo SSILo,
+                                   StoreSRS512BitInfo SSIHi,
+                                   AddressingModeInfo &AMI,
+                                   MachineRegisterInfo &MRI);
   bool selectG_AIE_STORE_CONV(MachineInstr &StoreI, MachineRegisterInfo &MRI);
   bool selectG_AIE_LOAD_CONV(MachineInstr &StoreI, MachineRegisterInfo &MRI);
   bool selectG_AIE_STORE_PACK(MachineInstr &StoreI, MachineRegisterInfo &MRI);
@@ -1712,6 +1726,17 @@ getCombinedOpcodeSRSUPS(const MachineInstr &MemOp, const MachineInstr &CombOp,
               /*ISelOpcode=*/AIE2::VST_SRS_S32_S64_ag_idx_imm,
               AlwaysFitsImmediateRange,
               /*OffsetOpcode=*/AIE2::VST_SRS_S32_S64_ag_idx_imm};
+        // If two 256-bit SRS are inputs to 512-bit G_STORE through CONCAT
+        case Intrinsic::aie2_I256_v32_acc32_srs:
+          return LoadStoreOpcodes{
+              /*ISelOpcode=*/AIE2::VST_SRS_S8_S32_ag_idx_imm,
+              AlwaysFitsImmediateRange,
+              /*OffsetOpcode=*/AIE2::VST_SRS_S8_S32_ag_idx_imm};
+        case Intrinsic::aie2_I256_v8_acc64_srs:
+          return LoadStoreOpcodes{
+              /*ISelOpcode=*/AIE2::VST_SRS_S32_S64_ag_idx_imm,
+              AlwaysFitsImmediateRange,
+              /*OffsetOpcode=*/AIE2::VST_SRS_S32_S64_ag_idx_imm};
         }
       }
       if (getLoadStoreSize(MemOp) == 256) {
@@ -1751,6 +1776,20 @@ getCombinedOpcodeSRSUPS(const MachineInstr &MemOp, const MachineInstr &CombOp,
           return LoadStoreOpcodes{
               /*ISelOpcode=*/AIE2::VST_SRS_S32_S64_ag_idx_imm,
               FitsImmediateRange,
+              /*OffsetOpcode=*/AIE2::VST_SRS_S32_S64_ag_idx_imm};
+        // If two 256-bit SRS are inputs to 512-bit G_AIE_OFFSET_STORE through
+        // CONCAT
+        case Intrinsic::aie2_I256_v32_acc32_srs:
+          FitsImmediateRange =
+              checkImmediateRangeSplitting<3, 32, 32>(Immediate);
+          return LoadStoreOpcodes{
+              AIE2::VST_SRS_S8_S32_ag_idx_imm, FitsImmediateRange,
+              /*OffsetOpcode=*/AIE2::VST_SRS_S8_S32_ag_idx_imm};
+        case Intrinsic::aie2_I256_v8_acc64_srs:
+          FitsImmediateRange =
+              checkImmediateRangeSplitting<3, 32, 32>(Immediate);
+          return LoadStoreOpcodes{
+              AIE2::VST_SRS_S32_S64_ag_idx_imm, FitsImmediateRange,
               /*OffsetOpcode=*/AIE2::VST_SRS_S32_S64_ag_idx_imm};
         }
       }
@@ -1795,6 +1834,23 @@ getCombinedOpcodeSRSUPS(const MachineInstr &MemOp, const MachineInstr &CombOp,
               ISelOpcode, FitsImmediateRange,
               /*OffsetOpcode=*/AIE2::VST_SRS_S16_S32_ag_idx_imm};
         case Intrinsic::aie2_I512_v16_acc64_srs:
+          FitsImmediateRange = checkImmediateRange<4, 32>(Immediate);
+          ISelOpcode = FitsImmediateRange
+                           ? AIE2::VST_SRS_S32_S64_ag_pstm_nrm_imm
+                           : AIE2::VST_SRS_S32_S64_ag_pstm_nrm;
+          return LoadStoreOpcodes{
+              ISelOpcode, FitsImmediateRange,
+              /*OffsetOpcode=*/AIE2::VST_SRS_S32_S64_ag_idx_imm};
+        // If two 256-bit SRS are inputs to 512-bit G_AIE_POSTINC_STORE through
+        // CONCAT
+        case Intrinsic::aie2_I256_v32_acc32_srs:
+          FitsImmediateRange = checkImmediateRange<4, 32>(Immediate);
+          ISelOpcode = FitsImmediateRange ? AIE2::VST_SRS_S8_S32_ag_pstm_nrm_imm
+                                          : AIE2::VST_SRS_S8_S32_ag_pstm_nrm;
+          return LoadStoreOpcodes{
+              ISelOpcode, FitsImmediateRange,
+              /*OffsetOpcode=*/AIE2::VST_SRS_S8_S32_ag_idx_imm};
+        case Intrinsic::aie2_I256_v8_acc64_srs:
           FitsImmediateRange = checkImmediateRange<4, 32>(Immediate);
           ISelOpcode = FitsImmediateRange
                            ? AIE2::VST_SRS_S32_S64_ag_pstm_nrm_imm
@@ -1847,6 +1903,16 @@ getCombinedOpcodeSRSUPS(const MachineInstr &MemOp, const MachineInstr &CombOp,
           return LoadStoreOpcodes{
               /*ISelOpcode=*/AIE2::VST_2D_SRS_S32_S64, NoImmediate,
               /*OffsetOpcode=*/AIE2::VST_SRS_S32_S64_ag_idx_imm};
+        // If two 256-bit SRS are inputs to 512-bit G_AIE_POSTINC_2D_STORE
+        // through CONCAT
+        case Intrinsic::aie2_I256_v32_acc32_srs:
+          return LoadStoreOpcodes{
+              /*ISelOpcode=*/AIE2::VST_2D_SRS_S8_S32, NoImmediate,
+              /*OffsetOpcode=*/AIE2::VST_SRS_S8_S32_ag_idx_imm};
+        case Intrinsic::aie2_I256_v8_acc64_srs:
+          return LoadStoreOpcodes{
+              /*ISelOpcode=*/AIE2::VST_2D_SRS_S32_S64, NoImmediate,
+              /*OffsetOpcode=*/AIE2::VST_SRS_S32_S64_ag_idx_imm};
         }
       }
       if (getLoadStoreSize(MemOp) == 256) {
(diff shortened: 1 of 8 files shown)


llvm/lib/Target/AIE/AIE2InstructionSelector.cpp
```diff
@@ -4140,6 +4266,165 @@ bool AIE2InstructionSelector::select512BitG_AIE_STORE_SRS(
...
+    return false;
+
+  std::optional<AddressingModeInfo> AMI =
+      getOrDefineAddressingRegister(StoreI, MRI);
+  if (!AMI)
+    return false;
+
+  Register SrcRegLo = SrsOpLo->getOperand(2).getReg();
+  Register ShftRegLo = SrsOpLo->getOperand(3).getReg();
+  Register SignRegLo = SrsOpLo->getOperand(4).getReg();
```
gbossu: Nit: You could use something like `auto [Res, CarryOut, LHS, RHS, CarryIn] = MI.getFirst5Regs();`

llvm/lib/Target/AIE/AIE2InstructionSelector.cpp
```diff
@@ -4140,6 +4266,165 @@ bool AIE2InstructionSelector::select512BitG_AIE_STORE_SRS(
...
+  case AIE2::G_AIE_POSTINC_STORE:
+  case AIE2::G_AIE_POSTINC_2D_STORE:
+  case AIE2::G_AIE_POSTINC_3D_STORE: {
+    auto StoreHigher = MIB.buildInstr(*LSOHi->OffsetOpcode)
+                           .addReg(AMI->PtrOp.getReg())
+                           .addImm(32) // Offset
+                           .addReg(SrcRegHi)
+                           .addReg(ShftRegHi);
+
+    auto StoreLower = MIB.buildInstr(LSOLo->ISelOpcode);
```
gbossu: I guess we are building the `vst.srs` instructions at the position of the `G_STORE` instruction. As a consequence, this means we will read the `crSat` and `crRnd` registers at that point, instead of where the `vsrs` instructions used to be. If those registers are changed in between (e.g. with `set_sat()`), this means `vst.srs` would be configured differently compared to the initial `vsrs`. Do you know if there is anything protecting us against that? Maybe we check for sideEffect when trying to move vsrs down?

llvm/test/CodeGen/AIE/aie2/GlobalISel/inst-select-vst_srs.mir
```diff
@@ -481,3 +481,213 @@ body: |
     %6:vregbank(<32 x s16>) = G_INTRINSIC_W_SIDE_EFFECTS intrinsic(@llvm.aie2.I512.v32.acc32.srs), %0:accregbank(<16 x s64>), %1:gprregbank(s32), %7:gprregbank(s32)
     G_STORE %6:vregbank(<32 x s16>), %3:ptrregbank(p0) :: (store (<32 x s16>))
 ...
+
+---
```
gbossu, line 489: Could you squash that commit with the previous one?

llvm/test/CodeGen/AIE/aie2/GlobalISel/inst-select-vst_srs.mir
```diff
@@ -481,3 +481,213 @@ body: |
...
+    ; CHECK-NEXT: $crsrssign = MOV_scalar_imm10_pseudo 0
+    %0:accregbank(<8 x s64>) = COPY $bml0
+    %1:accregbank(<8 x s64>) = COPY $bml1
+    %2:ptrregbank(p0) = COPY $p0
+    %3:gprregbank(s32) = COPY $r0
+    %5:gprregbank(s32) = G_CONSTANT i32 0
+    %7:gprregbank(s32) = G_CONSTANT i32 1
+    %4:vregbank(<8 x s32>) = G_INTRINSIC_W_SIDE_EFFECTS intrinsic(@llvm.aie2.I256.v8.acc64.srs), %0(<8 x s64>), %5(s32), %3(s32)
+    %6:vregbank(<8 x s32>) = G_INTRINSIC_W_SIDE_EFFECTS intrinsic(@llvm.aie2.I256.v8.acc64.srs), %1(<8 x s64>), %5(s32), %7(s32)
+    %8:vregbank(<16 x s32>) = G_INTRINSIC intrinsic(@llvm.aie2.concat.I512.I256), %4(<8 x s32>), %6(<8 x s32>)
```
gbossu, line 695: Could you add a test with a `set_sat` intrinsic call in between `llvm.aie2.concat` and `G_STORE`?

llvm/lib/Target/AIE/AIE2InstructionSelector.cpp
```diff
@@ -4140,6 +4266,165 @@ bool AIE2InstructionSelector::select512BitG_AIE_STORE_SRS(
...
+
+  Register ConcatResult = (StoreI.uses().begin())->getReg();
+  MachineInstr *ConcatOp = MRI.getUniqueVRegDef(ConcatResult);
+
+  assert(ConcatOp && "Expected SSA.");
+
+  if (getLoadStoreSize(StoreI) != 512 ||
+      ConcatOp->getOpcode() != AIE2::G_INTRINSIC ||
+      cast<GIntrinsic>(*ConcatOp).getIntrinsicID() !=
+          Intrinsic::aie2_concat_I512_I256)
```
martien-de-jong, line 4312: I'm wondering whether we would want to normalize this to G_CONCAT_VECTORS. Any way, having common pattern recognizer classes for these common glue patterns (e.g. CONCAT(vsz, vsz) -> vszx2)  would really help readability, and be good candidates to have tablegen support for pattern matching.

llvm/lib/Target/AIE/AIE2InstructionSelector.cpp
abhinay-anubola: Unable to use `getFirst5Regs` as second operand is not reg(intrinsic)

llvm/lib/Target/AIE/AIE2InstructionSelector.cpp
andcarminati: I agree, if we can solve this in PostLegalizer combiner I think the solution can be simpler. Something like:

```
X = CONCAT(SRS(A), SRS(B))
```
To something like:

```
X = SRS2(CONCAT(A, B))
```
And let Selector intact.

gbossu: That indeed seems to be a better way, but I'd rather not split the target-specific handling of intrinsics in one more place, and keep it solely here. I think we all agree the logic for selecting opcodes in this file isn't great, but hopefully we can table-gen-erate most of the logic at some point. Hopefully :)

llvm/lib/Target/AIE/AIE2InstructionSelector.cpp
```diff
@@ -4140,6 +4266,168 @@ bool AIE2InstructionSelector::select512BitG_AIE_STORE_SRS(
...
+bool AIE2InstructionSelector::selectG_AIE_STORE_CONCAT_SRS(
+    MachineInstr &StoreI, MachineRegisterInfo &MRI) {
+
+  Register ConcatResult = (StoreI.uses().begin())->getReg();
+  MachineInstr *ConcatOp = MRI.getUniqueVRegDef(ConcatResult);
+
+  assert(ConcatOp && "Expected SSA.");
+
+  if (!canDelayMemOp(*ConcatOp, StoreI, MRI))
+    return false;
```
gbossu, line 4307: Can you add a comment to explain what this is checking? I think we are verifying that it is safe to delay the VSRS operation. In particular, we verify that the control registers aren't clobbered in between the vsrs and the store.

llvm/lib/Target/AIE/AIE2InstructionSelector.cpp
```diff
@@ -4140,6 +4266,168 @@ bool AIE2InstructionSelector::select512BitG_AIE_STORE_SRS(
...
+                                    SignVal.value().Value == 0x1);
+    ConstantSignHi = true;
+  } else {
+    LSOHi =
+        getCombinedOpcodeSRSUPS(StoreI, *SrsOpHi, AMI->ImmediateOffset, false);
+  }
+
+  assert(LSOHi && "Unexpected VST.SRS combine failure");
+
+  switch (AMI->MemI.getOpcode()) {
```
gbossu: That insertion logic here is very similar to `select512BitG_AIE_STORE_SRS`. Is there any way to share code? The difference seems mainly that the function takes a 512-bit src register and splits it, while in your case you already have two 256-bit regs (the two inputs to concat). 

How about changing the prototype of `select512BitG_AIE_STORE_SRS(..., Register SrcReg ...)` into `select512BitG_AIE_STORE_SRS(..., Register SrcRegLo, Register SrcRegHi ...)`?

The old code could be updated through something like
```
    auto [SrcRegLo, SrcRegHi] = split512BitReg(SrcReg);
    return select512BitG_AIE_STORE_SRS(*LSO, *AMI, SrcRegLo, SrcRegHi, ShftReg, SignReg, ConstantSign, MRI);
```

llvm/lib/Target/AIE/AIE2InstructionSelector.cpp
```diff
@@ -188,10 +195,14 @@ class AIE2InstructionSelector : public InstructionSelector {
...
   bool selectG_AIE_STORE_SRS(MachineInstr &StoreI, MachineRegisterInfo &MRI);
-  bool select512BitG_AIE_STORE_SRS(LoadStoreOpcodes &LSO,
-                                   AddressingModeInfo &AMI, Register SrcReg,
-                                   Register ShftReg, Register SignReg,
-                                   bool ConstantSign, MachineRegisterInfo &MRI);
+  bool selectG_AIE_STORE_CONCAT_SRS(MachineInstr &StoreI,
+                                    MachineRegisterInfo &MRI);
+  std::tuple<Register, Register> split512BitReg(Register SrcReg,
+                                                MachineRegisterInfo &MRI);
+  bool select512BitG_AIE_STORE_SRS(StoreSRS512BitInfo SSI,
```
andcarminati: Hi @abhinay-anubola, I think you can just have two StoreSRS512BitInfo parameters here SSILo and SSIHi, in this way you can simplify your logic a a lot. You can pass twice even if they are the same.

llvm/lib/Target/AIE/AIE2InstructionSelector.cpp
```diff
@@ -4044,92 +4179,120 @@ bool AIE2InstructionSelector::selectG_AIE_STORE_PACK(MachineInstr &StoreI,
...
+  return {LowerBits.getReg(0), HigherBits.getReg(0)};
+}
+
+bool AIE2InstructionSelector::select512BitG_AIE_STORE_SRS(
+    StoreSRS512BitInfo SSI, AddressingModeInfo &AMI, Register SrcRegLo,
+    Register SrcRegHi, MachineRegisterInfo &MRI,
+    std::optional<StoreSRS512BitInfo> SSIHi) {
+  LoadStoreOpcodes LSOHi;
+  Register ShftRegHi, SignRegHi;
+  if (SSIHi) {
```
andcarminati: With the simplification, you can avoid this logic.

llvm/lib/Target/AIE/AIE2InstructionSelector.cpp
```diff
@@ -4044,92 +4179,120 @@ bool AIE2InstructionSelector::selectG_AIE_STORE_PACK(MachineInstr &StoreI,
...
 
-    StoreLower.addReg(LowerBits.getReg(0)).addReg(ShftReg);
+    StoreLower.addReg(SrcRegLo).addReg(SSI.ShftReg);
 
     addSplitMemOperands(AMI.MemI, StoreHigher, StoreLower, 0, 2);
 
-    if (!ConstantSign)
-      setUnsetCtrlRegister(*StoreHigher, *StoreLower, MRI, AIE2::crSRSSign,
-                           SignReg);
+    if (SSIHi) {
```
andcarminati: Also this part can be simplified.

abhinay-anubola: If we remove this logic, we have to use `setUnsetCtrlRegister` twice, which is adding extra instructions in existing tests.

andcarminati: I think there is no problem, we already have logic to remove/hoist such duplicated moves.

andcarminati:
Hi @abhinay-anubola,

I feel that we need to think in a common approach here. I identified similar case in DivAttributeBroadcasting_aie2_bf16_0:

```
    %198:vregbank(<16 x s16>) = G_INTRINSIC_W_SIDE_EFFECTS intrinsic(@llvm.aie2.v16accfloat.to.v16bf16), %197(<8 x s64>)
    %199:accregbank(<8 x s64>) = G_INTRINSIC intrinsic(@llvm.aie2.ext.512.1024.acc), %196(<16 x s64>), %74(s32)
    %200:vregbank(<16 x s16>) = G_INTRINSIC_W_SIDE_EFFECTS intrinsic(@llvm.aie2.v16accfloat.to.v16bf16), %199(<8 x s64>)
    %201:vregbank(<32 x s16>) = G_INTRINSIC intrinsic(@llvm.aie2.concat.bf512.bf256), %198(<16 x s16>), %200(<16 x s16>)
    %224:modregbank(s20) = G_CONSTANT i20 64
    %202:ptrregbank(p0) = G_AIE_POSTINC_STORE %138(<32 x s16>), %38, %224(s20) :: (store (<32 x s16>) into %ir.p_out.0, align 32, !tbaa !4, addrspace 7)
    %225:modregbank(s20) = G_CONSTANT i20 64
    %204:ptrregbank(p0) = G_AIE_POSTINC_STORE %201(<32 x s16>), %202, %225(s20) :: (store (<32 x s16>) into %ir.arrayidx.i.i.i.i.i.i.i.i.i.i.i12, align 32, !tbaa !4, addrspace 7)
    %206:gprregbank(s32) = nsw G_ADD %35, %205
    %211:gprregbank(s32) = G_ICMP intpred(eq), %206(s32), %56
    G_BRCOND %211(s32), %bb.5
    G_BR %bb.6
```

I think the best here is to have a transparent way to handle those concat/regseq/bitcast among all operations. What do you think?

llvm/lib/Target/AIE/AIE2InstructionSelector.cpp
```diff
@@ -4140,6 +4303,92 @@ bool AIE2InstructionSelector::select512BitG_AIE_STORE_SRS(
...
+
+  Register ConcatResult = (StoreI.uses().begin())->getReg();
+  MachineInstr *ConcatOp = MRI.getUniqueVRegDef(ConcatResult);
+
+  assert(ConcatOp && "Expected SSA.");
+
+  // Do not try to combine if one of the concat's defs is used by another
+  // instruction between the concat and the store or if there is a other store
+  // between the concat and the store.
+  if (!canDelayMemOp(*ConcatOp, StoreI, MRI))
```
andcarminati: Maybe we need a different name here, or different function. ConcatOp is not a mem op and also there is no problem about crossing stores.

andcarminati: This checking is blocking some optimizations. See: 

```
---
name:            A
alignment:       16
legalized:       true
regBankSelected: true
body:             |
  bb.1.entry:
    liveins: $bml0, $bml1, $p0, $r0
    %0:accregbank(<8 x s64>) = COPY $bml0
    %1:accregbank(<8 x s64>) = COPY $bml1
    %2:ptrregbank(p0) = COPY $p0
    %3:gprregbank(s32) = G_CONSTANT i32 1
    %5:gprregbank(s32) = G_CONSTANT i32 0
    %4:vregbank(<8 x s32>) = G_INTRINSIC_W_SIDE_EFFECTS intrinsic(@llvm.aie2.I256.v8.acc64.srs), %0(<8 x s64>), %5(s32), %3(s32)
    %6:vregbank(<8 x s32>) = G_INTRINSIC_W_SIDE_EFFECTS intrinsic(@llvm.aie2.I256.v8.acc64.srs), %1(<8 x s64>), %5(s32), %3(s32)
    %7:vregbank(<16 x s32>) = G_INTRINSIC intrinsic(@llvm.aie2.concat.I512.I256), %4(<8 x s32>), %6(<8 x s32>)
    %8:gprregbank(s32) = G_CONSTANT i32 9
    %9:modregbank(s20) = G_CONSTANT i20 -128
    G_INTRINSIC_W_SIDE_EFFECTS intrinsic(@llvm.aie2.set.ctrl.reg), %8:gprregbank(s32), %3:gprregbank(s32)
    %10:ptrregbank(p0) = G_AIE_POSTINC_STORE %7(<16 x s32>), %2(p0), %9 :: (store (<16 x s32>))
...

```

I tested both cases and they were not combined.

I got:
```
    %0:acc512 = COPY $bml0
    %1:acc512 = COPY $bml1
    %2:ep = COPY $p0
    %4:er = MOV_RLC_imm10_pseudo 0
    %14:mss = COPY %4
    %5:ewl = VSRS_S32_S64_mv_w_srs %0, %14, implicit-def dead $srsrs_of, implicit $crsat, implicit $crrnd
    %13:mss = COPY %4
    %6:ewh = VSRS_S32_S64_mv_w_srs %1, %13, implicit-def dead $srsrs_of, implicit $crsat, implicit $crrnd
    %7:vec512 = REG_SEQUENCE %5, %subreg.sub_256_lo, %6, %subreg.sub_256_hi
    $crsat = MOV_scalar_imm10_pseudo 1
    %11:vec256 = COPY %7.sub_256_lo
    %12:vec256 = COPY %7.sub_256_hi
    VST_dmw_sts_w_ag_idx_imm %12, %2, 32 :: (store (<8 x s32>) into unknown-address + 32)
    %10:ep = VST_dmw_sts_w_ag_pstm_nrm_imm %11, %2, -128 :: (store (<8 x s32>), align 64)
```

Instead of:
```
    %0:acc512 = COPY $bml0
    %1:acc512 = COPY $bml1
    %2:ep = COPY $p0
    %4:er = MOV_RLC_imm10_pseudo 0
    $crsat = MOV_scalar_imm10_pseudo 1
    %15:mss = COPY %4
    VST_SRS_S32_S64_ag_idx_imm %2, 32, %1, %15, implicit-def $srsrs_of, implicit $crsat, implicit $crrnd :: (store (<8 x s32>) into unknown-address + 32)
    %14:mss = COPY %4
    %10:ep = VST_SRS_S32_S64_ag_pstm_nrm_imm %2, -128, %0, %14, implicit-def $srsrs_of, implicit $crsat, implicit $crrnd :: (store (<8 x s32>), align 64)
```

The same applies if I use a G_STORE instead of POSTINC.

abhinay-anubola:
This was solved by: https://github.com/Xilinx/llvm-aie/pull/195

PR #182, opened 2024-09-06
[AIEX] Delay metalizing of multi-slot until iterative scheduling is converged

krishnamtibrewala:
This PR allows Multi-Slot Instr. to be used during iterative scheduling of loop.
Before this PR we were materializing Multi-Slot Instr. to selected OpCode/Slot after first iteration of iterative scheduling. 
Now we wait until PostRA scheduling is converged to an acceptable schedule and we have decided to commit the schedule and move to next MBB. 

- When is the materialization triggered now? : When we leave a MBB.
- Does this change depending on the region type? : The changes are agnostic to region type.
- Could you think of the case where the materialization is not triggered before moving on to another MBB? : None that I can think of

Note : Given the information of what Alternate opcode/desc was selected is stored in Hazard Recognizer for a region.
And by the time we come to the end of MBB ( i.e leaveMBB() ) we do not have access to the instance of those Hazard Recognizer, therefore we need to make the Alternate opcode/desc part of the `BlockState`


krishnamtibrewala:
@martien-de-jong , @andcarminati.
Kindly review and provide an early feedback toward the approach

llvm/lib/Target/AIE/AIEHazardRecognizer.h
```diff
@@ -153,10 +153,19 @@ class AIEHazardRecognizer : public ScheduleHazardRecognizer {
...
 
+  /// For instructions with multiple "alternative opcodes", this will return
+  /// the MCInstrDesc pointer selected during scheduling.
+  std::optional<const MCInstrDesc *> getSelectedAltDesc(MachineInstr *MI) const;
+
   /// For instructions with multiple "alternative opcodes", this will return
   /// the opcode selected during scheduling.
   std::optional<unsigned> getSelectedAltOpcode(MachineInstr *MI) const;
 
+  /// Function to return a pointer to the SelectedAltOpcodes map
```
andcarminati: nit: maybe `SelectedAltDesc map`.

llvm/lib/Target/AIE/AIEHazardRecognizer.h
```diff
@@ -202,7 +211,7 @@ class AIEHazardRecognizer : public ScheduleHazardRecognizer {
 
 private:
   ResourceScoreboard<FuncUnitWrapper> Scoreboard;
-  std::map<MachineInstr *, unsigned> SelectedAltOpcodes;
+  std::map<MachineInstr *, const MCInstrDesc *> SelectedAltOpcodes;
```
andcarminati: nit: `SelectedAltDescs`

llvm/lib/Target/AIE/AIEInterBlockScheduling.cpp
```diff
@@ -91,9 +97,13 @@ createBottomUpScoreboard(ArrayRef<MachineBundle> Bundles,
   ArrayRef<MachineBundle> MinBundles(
       Bundles.begin(), Bundles.begin() + std::min(NumBundles, RequiredCycles));
   for (const MachineBundle &B : reverse(MinBundles)) {
-    for (MachineInstr *MI : B.getInstrs())
-      HR.emitInScoreboard(Scoreboard, MI->getDesc(), HR.getMemoryBanks(MI),
+    for (MachineInstr *MI : B.getInstrs()) {
```
andcarminati: Lines 64-68 and 100-104 are pretty much the same, can we refactor some code?

andcarminati:
Hi @krishnamtibrewala, nice work! Do you have some results for the PixelShuffle*/PixelUnshuffle* benchmarks? If I remember correctly, we have some suboptimal mov desc assignments (movx should be selected instead of mova to not shift loads ups).

llvm/lib/Target/AIE/AIEMachineScheduler.cpp
```diff
@@ -613,34 +624,17 @@ void AIEPostRASchedStrategy::leaveRegion(const SUnit &ExitSU) {
   assert(BS.getCurrentRegion().Bundles.empty());
   BS.addBundles(TopBundles);
   BS.addBundles(BotBundles);
+  BS.addSelectedAltOpcodes(
+      *getAIEHazardRecognizer(Top)->getSelectedAltOpcodes());
+  BS.addSelectedAltOpcodes(
+      *getAIEHazardRecognizer(Bot)->getSelectedAltOpcodes());
```
martien-de-jong: I'm wondering whether we could pass a handle to this map to AIEHazardRecognizer, so that we don't have to make a copy. I guess AIEHazardRecognizer commits the selected opcode in the scoreboard and doesn't use the map in any other way?

llvm/lib/Target/AIE/AIEInterBlockScheduling.h
```diff
@@ -151,6 +151,8 @@ class Region {
   MachineInstr *getExitInstr() const { return ExitInstr; }
 
   std::vector<MachineBundle> Bundles;
+  // Muli-Opcode Instr and their selected opcode
+  std::map<MachineInstr *, const MCInstrDesc *> SelectedAltOpcodes;
```
martien-de-jong: I wouldn't mind using just one map per BB, rather than per region.

llvm/lib/Target/AIE/AIEInterBlockScheduling.h
```diff
@@ -151,6 +151,8 @@ class Region {
   MachineInstr *getExitInstr() const { return ExitInstr; }
 
   std::vector<MachineBundle> Bundles;
+  // Muli-Opcode Instr and their selected opcode
```
martien-de-jong: nit: *Multi*

martien-de-jong (commented):
Could we have a test example where we see improvement?

llvm/lib/Target/AIE/AIEMachineScheduler.cpp
krishnamtibrewala: As i understand the instances of Top & Bottom HR are re-initialized when we start a new region. And we do convergence check at the end while leaving the MBB.

krishnamtibrewala:
Do you have some results for the PixelShuffle*/PixelUnshuffle* benchmarks? If I remember correctly, we have some suboptimal mov desc assignments (movx should be selected instead of mova to not shift loads ups).
>> @andcarminati I tried but I did not see any change, still investigating why.

Could we have a test example where we see improvement?
>> @martien-de-jong still figuring out.

Based on discussion with @gbossu we were expecting some impact but with current implementation QoR have no change.

llvm/test/CodeGen/AIE/aie2/schedule/loopaware/loop-multiSlot.mir
```diff
@@ -0,0 +1,300 @@
...
+    $wl10, $p0 = VLD_pstm_pseudo $p0, $m0 :: (load (<8 x s32>), align 32, addrspace 5)
+    $wl1, $p1 = VLD_pstm_pseudo $p1, $m1 :: (load (<8 x s32>), align 32, addrspace 6)
+    $r3 = ADD_add_r_ri $r3, -4, implicit-def $srcarry
+    $wh3 = VMOV_mv_w $wl0
+    $wh7 = VMOV_mv_w $wl0
+    $wh2 = VMOV_mv_w $wl0
+    $wh10 = VMOV_mv_w $wl0
+    JNZ $r3, %bb.1
+    DelayedSchedBarrier
+  bb.2:
```
krishnamtibrewala, line 293: Attached is a diff, I am still trying to figure out how things are interacting.
Also will try to come up with a smaller test case.  

image

llvm/lib/Target/AIE/AIEInterBlockScheduling.cpp
```diff
@@ -51,26 +51,32 @@ void dumpInterBlock(const InterBlockEdges &Edges) {
...
 
   const int TotalBundles = Bundles.size();
   const int AmountToEmit = std::min(TotalBundles, HR->getConflictHorizon());
   // Do not emit more than the specified by the conflict horizon. More
   // then this will not cause conflicts.
   for (int i = TotalBundles - AmountToEmit; i < TotalBundles; i++) {
-    for (MachineInstr *MI : Bundles[i].getInstrs())
-      HR->emitInScoreboard(Scoreboard, MI->getDesc(), HR->getMemoryBanks(MI),
+    for (MachineInstr *MI : Bundles[i].getInstrs()) {
+      auto Desc = SelectedAltDesc.find(MI) != SelectedAltDesc.end()
```
andcarminati: I think this whole AltDesc handling should be in a different class/file. This file is a bit bloated and we share this manipulation logic across here and HazardRecognizer.

andcarminati:
As a general advice, I think we should have a class to manage the description handling. We can encapsulate and use it in HazardRecognizer and InterBlockScheduling.

llvm/lib/Target/AIE/AIEInterBlockScheduling.cpp
krishnamtibrewala: Thanks for the guidance have accordingly created #193

llvm/lib/Target/AIE/AIEInterBlockScheduling.cpp
```diff
@@ -263,14 +266,22 @@ class PipelineExtractor : public PipelineScheduleVisitor {
     // Prologue and epilogue obtain copies.
     MachineInstr *ToBeEmitted =
         InLoop ? MI : Loop.TheBlock->getParent()->CloneMachineInstr(MI);
```
krishnamtibrewala, line 337: I think we should prefer not using `CloneMachineInstr`, since this is not added to MBB and other structure like `BlockState`

martien-de-jong: However, using CloneMachineInstr is essentially what the post pipeliner needs. When we call this extraction, everything in the schedule has been frozen. We can materialize the pseudos just before calling the extractor.
The original instructions reside in the loop block's region. We materialize them first, then cloning them will see the correct descriptor.

martien-de-jong: If you see a better way to emit an exact copy of an instruction into another basic block, please let me know.

gbossu: We had discussed that offline with Krishnam. The post-pipeliner should only deal with "finalized" instructions when materializing the schedule, i.e. no multi-slot opcodes anywhere. So cloning is fine at this stage, because state like `AlternateOpcodes` is outdated (and using it should be forbidden with asserts if possible).

llvm/lib/Target/AIE/AIEAlternateDescriptors.h
```diff
@@ -31,6 +31,10 @@ class AIEAlternateDescriptors {
   AIEAlternateDescriptors() = default;
   ~AIEAlternateDescriptors() = default;
 
+  const MIAltDescsMap &getAlternateDescriptors() const {
+    return AlternateDescs;
+  }
```
gbossu: Nit: I'd rather avoid exposing too much about the internal data structure unless really needed. After all, `AIEAlternateDescriptors` is meant as a container, so I think we should just define:
```
MIAltDescsMap::const_iterator begin() const { return AlternateDescs.begin(); }
MIAltDescsMap::const_iterator end() const { return AlternateDescs.end(); }
```

llvm/lib/Target/AIE/AIEInterBlockScheduling.cpp
```diff
@@ -233,6 +235,7 @@ namespace {
 /// into the appropriate blockstate region.
 /// TimedRegion is built one bundle at the time
 class PipelineExtractor : public PipelineScheduleVisitor {
+  InterBlockScheduling &InterBlock;
```
gbossu: We seem to only use `InterBlock.getSelectedAltDescs()`. Maybe just have a `AlternateDescriptors` member?

llvm/lib/Target/AIE/AIEMachineScheduler.cpp
```diff
@@ -535,7 +539,20 @@ void AIEPostRASchedStrategy::enterMBB(MachineBasicBlock *MBB) {
   IsBottomRegion = true;
 }
 
+void AIEPostRASchedStrategy::materializeMultiOpcodeInstrs() {
+  for (auto [MI, Desc] : make_range(InterBlock.getSelectedAltDescs().begin(),
```
gbossu: Maybe use `auto &[MI, Desc]`?

llvm/lib/Target/AIE/AIEMachineScheduler.cpp
```diff
@@ -535,7 +539,20 @@ void AIEPostRASchedStrategy::enterMBB(MachineBasicBlock *MBB) {
...
+    MI->setDesc(*Desc);
+  }
+
+  InterBlock.getSelectedAltDescs().clear();
+}
+
 void AIEPostRASchedStrategy::commitBlockSchedule(MachineBasicBlock *BB) {
+
+  if (ReAssignMultiSlotInstr)
+    materializeMultiOpcodeInstrs();
```
andcarminati: MultiOpcode looks a bit strange... Maybe MultiSlot is what you mean here too.

llvm/lib/Target/AIE/AIEMachineScheduler.cpp
```diff
@@ -616,34 +632,15 @@ void AIEPostRASchedStrategy::leaveRegion(const SUnit &ExitSU) {
   assert(BS.getCurrentRegion().Bundles.empty());
   BS.addBundles(TopBundles);
   BS.addBundles(BotBundles);
+  if (!ReAssignMultiSlotInstr)
+    materializeMultiOpcodeInstrs();
```
gbossu: Any particular reason to "move" the lines down here? What if we do not enter `if (IsBottomRegion) { ... }`?

gbossu:
Could you summarize what this PR does? Maybe in the PR description. I'm particularly interested in:
 - When is the materialization triggered now?
 - Does this change depending on the region type?
 - Could you think of the case where the materialization is not triggered before moving on to another `MBB`?

llvm/lib/Target/AIE/AIEMachineScheduler.cpp
krishnamtibrewala: The following code block is outside the `if (IsBottomRegion) { ... }`
```
  if (!ReAssignMultiSlotInstr)
    materializeMultiOpcodeInstrs();
```

llvm/lib/Target/AIE/AIEMachineScheduler.cpp
```diff
@@ -616,34 +632,15 @@ void AIEPostRASchedStrategy::leaveRegion(const SUnit &ExitSU) {
   assert(BS.getCurrentRegion().Bundles.empty());
   BS.addBundles(TopBundles);
   BS.addBundles(BotBundles);
+  if (!ReAssignMultiSlotInstr)
+    materializeMultiSlotInstrs();
```
gbossu: Why not keep the materialization call where it used to be? Is there any reason why you now do it after computing the Bundles?

llvm/lib/Target/AIE/AIEInterBlockScheduling.cpp
```diff
@@ -330,14 +331,19 @@ class PipelineExtractor : public PipelineScheduleVisitor {
     // Prologue and epilogue obtain copies.
     MachineInstr *ToBeEmitted =
         InLoop ? MI : Loop.TheBlock->getParent()->CloneMachineInstr(MI);
-    CurrentBundle.add(ToBeEmitted);
+    if (auto AltDesc = AlternateDesc.getSelectedDescriptor(MI);
+        AltDesc.has_value())
```
gbossu: Nit: just `if (auto SelectedDesc = AlternateDesc.getSelectedDescriptor(MI))`?

llvm/lib/Target/AIE/AIEInterBlockScheduling.cpp
```diff
@@ -330,14 +331,19 @@ class PipelineExtractor : public PipelineScheduleVisitor {
     // Prologue and epilogue obtain copies.
     MachineInstr *ToBeEmitted =
         InLoop ? MI : Loop.TheBlock->getParent()->CloneMachineInstr(MI);
```
gbossu, line 337: Maybe leave a note that so far we force the prologue/epilogue copies to have the same descriptor as the "original" instruction in the steady state.

llvm/lib/Target/AIE/AIEMachineScheduler.cpp
```diff
@@ -269,7 +273,8 @@ void AIEPostRASchedStrategy::initializeBotScoreBoard(ScoreboardTrust Trust) {
   /// by starting in the earliest possible cycle, -Depth
   auto InsertInCycle = [=](MachineInstr &MI, int Cycle) {
     BotHazardRec->emitInScoreboard(
-        MI.getDesc(), BotHazardRec->getMemoryBanks(&MI), MI.operands(),
+        *BotHazardRec->getSelectedAltDescs().getDesc(&MI),
```
gbossu: This `BotHazardRec->emitInScoreboard(...)` will be called with instructions that have already been materialized. So here I would actually assert that `MI` has no "selected descriptor"

llvm/lib/Target/AIE/AIEMachineScheduler.cpp
krishnamtibrewala: There was no specific reason, have reverted it to where it used to be.

llvm/lib/Target/AIE/AIEInterBlockScheduling.cpp
krishnamtibrewala: Updated!

llvm/lib/Target/AIE/AIEInterBlockScheduling.cpp
krishnamtibrewala: Thanks , a comment will be helpful in future. Have added one.

llvm/lib/Target/AIE/AIEMachineScheduler.cpp
krishnamtibrewala: Yes , have reverted back to MI.getDesc() and added an assert to make sure instruction is already materialized

llvm/test/CodeGen/AIE/aie2/schedule/loopaware/loop-multiSlot.mir
```diff
@@ -0,0 +1,301 @@
...
+    liveins: $r1, $r2
+    successors: %bb.3
+    $r2 = OR $r2, $r1
+  bb.3:
```
gbossu, line 297: That test is unfortunately very hard to read. Could you think of something smaller that shows a diff? I'd also suggest avoiding two labels like on/off and stick to unique CHECK lines unless the test is very concise.

krishnamtibrewala:
QoR Results are as followed, there are few regression reset all benchmarks have same QoR

Select_aie2_bf16      	409	440	        REGR(+7.58%)	
BitwiseXor_aie2_int8	731	776	        REGR(+6.16%)	
BilinearInterpolation_2	996	1018	REGR(+2.21%)	
BilinearInterpolation_3	996	1018	REGR(+2.21%)	
BilinearInterpolation_4	996	1018	REGR(+2.21%)	
BilinearInterpolation_0	780	794	        REGR(+1.79%)	
Conv2D_bf16_2	19089	19281	REGR(+1.01%)	
Conv2D_bf16_5	19089	19281     	REGR(+1.01%)	
Conv2D_bf16_8	20607	20799	REGR(+0.93%)	
BilinearInterpolation_1	474	478	        REGR(+0.84%)	
Conv2D_bf16_59	6253	6301	REGR(+0.77%)

gbossu:
> QoR Results are as followed, there are few regression reset all benchmarks have same QoR
> 
> Select_aie2_bf16 409 440 REGR(+7.58%) BitwiseXor_aie2_int8 731 776 REGR(+6.16%) BilinearInterpolation_2 996 1018 REGR(+2.21%) BilinearInterpolation_3 996 1018 REGR(+2.21%) BilinearInterpolation_4 996 1018 REGR(+2.21%) BilinearInterpolation_0 780 794 REGR(+1.79%) Conv2D_bf16_2 19089 19281 REGR(+1.01%) Conv2D_bf16_5 19089 19281 REGR(+1.01%) Conv2D_bf16_8 20607 20799 REGR(+0.93%) BilinearInterpolation_1 474 478 REGR(+0.84%) Conv2D_bf16_59 6253 6301 REGR(+0.77%)

I'm quite surprised about the results. Did you have time to investigate the regressions?

krishnamtibrewala:
> > QoR Results are as followed, there are few regression reset all benchmarks have same QoR
> > Select_aie2_bf16 409 440 REGR(+7.58%) BitwiseXor_aie2_int8 731 776 REGR(+6.16%) BilinearInterpolation_2 996 1018 REGR(+2.21%) BilinearInterpolation_3 996 1018 REGR(+2.21%) BilinearInterpolation_4 996 1018 REGR(+2.21%) BilinearInterpolation_0 780 794 REGR(+1.79%) Conv2D_bf16_2 19089 19281 REGR(+1.01%) Conv2D_bf16_5 19089 19281 REGR(+1.01%) Conv2D_bf16_8 20607 20799 REGR(+0.93%) BilinearInterpolation_1 474 478 REGR(+0.84%) Conv2D_bf16_59 6253 6301 REGR(+0.77%)
> 
> I'm quite surprised about the results. Did you have time to investigate the regressions?

With the latest changes + 2nd Commit ([[AIE2] Update VLD Multi-Slot Itinerary](https://github.com/Xilinx/llvm-aie/pull/182/commits/506ed745c4adf73676c247ba7649f14759011929)) of the PR there are no regressions.
The issue was how Multi-slot latency Vector Load instruction latency was treated.

llvm/lib/Target/AIE/AIE2MultiSlotPseudoInstrInfo.td
```diff
@@ -73,7 +73,7 @@ let isMoveImm = 1, isReMaterializable = 1, isAsCheapAsAMove = 1, Itinerary = II_
 
 // Pseudo VLD
 let hasSideEffects = false, mayLoad = true, mayStore = false in {
-  let Itinerary = II_VLDA_W in {
+  let Itinerary = II_VLDB in {
```
gbossu, line 76: Can you give more details about this change? Why does this impact the postmisched? It should not be impacted by the resources of the multi-slot itinerary, because when checking or adding to the scoreboard, the final itinerary should be used.

PR #185, opened 2024-09-11
Martien.kissfix

martien-de-jong:
dup phi elimination crashed in a supertest on a block with just two phi nodes

### DIFF
--- a/llvm/lib/Target/AIE/AIEEliminateDuplicatePHI.cpp
+++ b/llvm/lib/Target/AIE/AIEEliminateDuplicatePHI.cpp
@@ -120,26 +120,26 @@ private:
   bool processMBB(MachineBasicBlock &MBB, MachineRegisterInfo &MRI,
                   MachineIRBuilder &MIB, GISelObserverWrapper &Observer) {
     bool Changed = false;
-    auto LastPHI = MBB.getFirstNonPHI();
-    for (auto &MI : make_early_inc_range(MBB.phis())) {
-      if (!isPointerTypePHI(MRI, MI))
-        continue;
-
-      // Ensure that next MI is valid
-      auto NextMI = std::next(MachineBasicBlock::instr_iterator(MI));
-      if (NextMI == MBB.instr_end() || NextMI == LastPHI)
+    const auto End = MBB.end();
+    for (auto PhiItA = MBB.begin(); PhiItA != End; ++PhiItA) {
+      auto &PhiA = *PhiItA;
+      if (PhiA.getOpcode() != TargetOpcode::G_PHI)
         break;
-
-      for (auto &PHI : make_early_inc_range(make_range(
-               NextMI, MachineBasicBlock::instr_iterator(*LastPHI)))) {
-        if (!isPointerTypePHI(MRI, PHI))
+      if (!isPointerTypePHI(MRI, PhiA))
+        continue;
+      auto PhiItB = PhiItA;
+      for (++PhiItB; PhiItB != End; ++PhiItB) {
+        auto &PhiB = *PhiItB;
+        if (PhiB.getOpcode() != TargetOpcode::G_PHI)
+          break;
+        if (!isPointerTypePHI(MRI, PhiB))
           continue;
-        if (MI.isIdenticalTo(PHI, MachineInstr::IgnoreDefs)) {
+        if (PhiA.isIdenticalTo(PhiB, MachineInstr::IgnoreDefs)) {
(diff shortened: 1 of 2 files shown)


llvm/lib/Target/AIE/AIEEliminateDuplicatePHI.cpp
```diff
@@ -120,26 +120,26 @@ class AIEEliminateDuplicatePHI : public MachineFunctionPass {
...
     bool Changed = false;
-    auto LastPHI = MBB.getFirstNonPHI();
-    for (auto &MI : make_early_inc_range(MBB.phis())) {
-      if (!isPointerTypePHI(MRI, MI))
-        continue;
-
-      // Ensure that next MI is valid
-      auto NextMI = std::next(MachineBasicBlock::instr_iterator(MI));
-      if (NextMI == MBB.instr_end() || NextMI == LastPHI)
+    const auto End = MBB.end();
```
martien-de-jong, line 123: We are comparing every pointer PHI node in the block to any other pointer PHI node in that block.  We know that all PHI nodes are consecutive at the start of the block. The transformation only changes operands in other instructions, so iterators stay intact. By writing the loops using basic iterator idiom reveals the symmetry between the two loops. It also avoids using constructs like .phis() and .getFirstNonPhi() that both perform exactly the same iteration as the main loops.

llvm/lib/Target/AIE/AIEEliminateDuplicatePHI.cpp
```diff
@@ -120,26 +120,26 @@ class AIEEliminateDuplicatePHI : public MachineFunctionPass {
...
+    for (auto PhiItA = MBB.begin(); PhiItA != End; ++PhiItA) {
+      auto &PhiA = *PhiItA;
+      if (PhiA.getOpcode() != TargetOpcode::G_PHI)
         break;
-
-      for (auto &PHI : make_early_inc_range(make_range(
-               NextMI, MachineBasicBlock::instr_iterator(*LastPHI)))) {
```
martien-de-jong, line 134: The error was here. LastPHI was at .end(), which means you cannot dereference it to create an instr_iterator. Note that in that case the validity check reads `if (NextMI == end || NextMI == end)`

llvm/lib/Target/AIE/AIEEliminateDuplicatePHI.cpp
```diff
@@ -120,26 +120,26 @@ class AIEEliminateDuplicatePHI : public MachineFunctionPass {
   bool processMBB(MachineBasicBlock &MBB, MachineRegisterInfo &MRI,
                   MachineIRBuilder &MIB, GISelObserverWrapper &Observer) {
     bool Changed = false;
-    auto LastPHI = MBB.getFirstNonPHI();
```
martien-de-jong, line 123: I would normally expect an iterator with this name to point to the last phi, not the the first non-phi.

llvm/test/CodeGen/AIE/GlobalISel/eliminate-duplicate-phi.mir
```diff
@@ -232,3 +232,51 @@ body:             |
...
+
+  bb.1:
+    G_BR %bb.3
+
+  bb.2:
+    G_BR %bb.3
+
+  bb.3:
+    %3:_(p0) = G_PHI %2:_(p0), %bb.1, %3:_(p0), %bb.2
+    %4:_(p0) = G_PHI %2:_(p0), %bb.1, %3:_(p0), %bb.2
```
martien-de-jong, line 284: This is just triggering the crash. These are equivalent phis, and the test can be strengthened by using %4 in a downstream block.

llvm/lib/Target/AIE/AIEEliminateDuplicatePHI.cpp
```diff
@@ -120,26 +120,26 @@ class AIEEliminateDuplicatePHI : public MachineFunctionPass {
...
+      if (PhiA.getOpcode() != TargetOpcode::G_PHI)
         break;
-
-      for (auto &PHI : make_early_inc_range(make_range(
-               NextMI, MachineBasicBlock::instr_iterator(*LastPHI)))) {
-        if (!isPointerTypePHI(MRI, PHI))
+      if (!isPointerTypePHI(MRI, PhiA))
+        continue;
+      auto PhiItB = PhiItA;
+      for (++PhiItB++; PhiItB != End; ++PhiItB) {
```
andcarminati: Looks interesting: `++PhiItB++`

martien-de-jong:
I approved the competing PR

PR #187, opened 2024-09-17
[AIE2] Simplify sequence of VInserts to G_BUILD_VECTOR

abhinay-anubola:
Simplify sequence of VInserts by legalizing G_BUILD_VECTOR_TRUNC and combining VInsert sequence in PreLegalizerCombiner into G_BUILD_VECTOR_TRUNC


llvm/lib/Target/AIE/AIE2PreLegalizerCombiner.cpp
```diff
@@ -136,13 +138,35 @@ bool AIE2PreLegalizerCombinerImpl::tryToCombineVectorShiftsByZero(
   return true;
 }
 
+bool AIE2PreLegalizerCombinerImpl::tryToCombineSetExtract(
+    MachineInstr &MI) const {
+  const Register DstReg = MI.getOperand(0).getReg();
```
martien-de-jong, line 152: I guess operand(1) holds the intrinsic ID?

llvm/lib/Target/AIE/AIE2PreLegalizerCombiner.cpp
```diff
@@ -157,6 +159,55 @@ bool AIE2PreLegalizerCombinerImpl::tryToCombineSetExtract(
...
+    assert(SrcMI && "Expected SSA.");
+    if (IsSet(SrcMI) && !tryToCombineSetExtract(*SrcMI))
+      break;
+
+    SrcMI = getDefIgnoringCopies(SrcReg, MRI);
+    if (!IsVInsert(SrcMI))
+      break;
+    CurMI = SrcMI;
+  }
+  if (!++NumInserts)
```
martien-de-jong: This is not a very clear condition. I think it corresponds to the break from the loop above. Why not return false immediately instead of breaking?

llvm/lib/Target/AIE/AIE2PreLegalizerCombiner.cpp
```diff
@@ -157,6 +159,55 @@ bool AIE2PreLegalizerCombinerImpl::tryToCombineSetExtract(
...
 
+bool AIE2PreLegalizerCombinerImpl::tryToCombineVectorInserts(
+    MachineInstr &MI) const {
+  const Register DstReg = MI.getOperand(0).getReg();
+  unsigned DstRegLen = MRI.getType(DstReg).getElementCount().getKnownMinValue();
+  SmallVector<Register, 16> Regs;
+  MachineInstr *CurMI = &MI;
+  int NumInserts = DstRegLen;
+
+  auto IsSet = [&](const MachineInstr *MI) {
```
martien-de-jong: nit: I think you can use an empty capture list.

llvm/lib/Target/AIE/AIE2PreLegalizerCombiner.cpp
```diff
@@ -157,6 +159,55 @@ bool AIE2PreLegalizerCombinerImpl::tryToCombineSetExtract(
...
+               Intrinsic::aie2_vinsert8_I512;
+  };
+
+  while (NumInserts--) {
+    const Register SrcReg = CurMI->getOperand(2).getReg();
+    const Register IdxReg = CurMI->getOperand(3).getReg();
+    const Register SclSrcReg = CurMI->getOperand(4).getReg();
+
+    auto Cst = getIConstantVRegValWithLookThrough(IdxReg, MRI);
+    if (!Cst || Cst->Value.getSExtValue() != NumInserts)
```
martien-de-jong: This expects the lane indices to be in a specific order. Instead, we could collect them and check whether we are complete afterwards.

llvm/lib/Target/AIE/AIE2PreLegalizerCombiner.cpp
```diff
@@ -136,13 +140,95 @@ bool AIE2PreLegalizerCombinerImpl::tryToCombineVectorShiftsByZero(
   return true;
 }
 
+bool AIE2PreLegalizerCombinerImpl::tryToCombineSetExtract(
+    MachineInstr &MI) const {
+  const Register DstReg = MI.getOperand(0).getReg();
+  MachineInstr *ExtOp = MRI.getUniqueVRegDef(MI.getOperand(2).getReg());
```
andcarminati: I think it is better to use getDefIgnoringCopies. We are in SSA here, so no need to worry about uniqueness.

llvm/lib/Target/AIE/AIE2PreLegalizerCombiner.cpp
```diff
@@ -136,13 +140,95 @@ bool AIE2PreLegalizerCombinerImpl::tryToCombineVectorShiftsByZero(
...
+
+  assert(ExtOp && "Expected SSA.");
+  if (ExtOp->getOpcode() != AIE2::G_INTRINSIC ||
+      cast<GIntrinsic>(*ExtOp).getIntrinsicID() !=
+          Intrinsic::aie2_extract_I128_I512)
+    return false;
+
+  MachineIRBuilder MIRBuilder(MI);
+  MIRBuilder.buildCopy(DstReg, ExtOp->getOperand(2).getReg());
+  ExtOp->eraseFromParent();
```
andcarminati: Probably here we may corrupt the iterator., do we need to erase ExtOp? DCE should handle this case.

llvm/lib/Target/AIE/AIE2PreLegalizerCombiner.cpp
```diff
@@ -136,13 +140,95 @@ bool AIE2PreLegalizerCombinerImpl::tryToCombineVectorShiftsByZero(
...
+    MachineInstr *SrcMI = MRI.getUniqueVRegDef(SrcReg);
+    assert(SrcMI && "Expected SSA.");
+    if (IsSet(SrcMI) && !tryToCombineSetExtract(*SrcMI))
+      break;
+    CurMI = getDefIgnoringCopies(SrcReg, MRI);
+  }
+
+  MachineIRBuilder MIRBuilder(MI);
+  // Collect registers in order for G_BUILD_VECTOR
+  for (int i = 0; i < DstRegLen; i++) {
```
andcarminati: nit: i ->I.

llvm/lib/Target/AIE/AIE2PreLegalizerCombiner.cpp
```diff
@@ -136,13 +140,95 @@ bool AIE2PreLegalizerCombinerImpl::tryToCombineVectorShiftsByZero(
...
+    MachineInstr &MI, int SclSrcBits) const {
+  const Register DstReg = MI.getOperand(0).getReg();
+  int DstRegLen = MRI.getType(DstReg).getElementCount().getKnownMinValue();
+  SmallVector<Register, 16> Regs;
+  std::map<int, Register> RegMap;
+  MachineInstr *CurMI = &MI;
+  int NumInserts = DstRegLen;
+
+  auto IsSet = [](const MachineInstr *MI) {
+    return MI->getOpcode() == AIE2::G_INTRINSIC &&
```
andcarminati: You can use `isa<GIntrinsic>(*MI)` to see it MI is an intrinsic.

llvm/lib/Target/AIE/AIELegalizerHelper.cpp
```diff
@@ -182,9 +185,32 @@ bool AIELegalizerHelper::legalizeG_BUILD_VECTOR(LegalizerHelper &Helper,
...
+    Register ShiftConstant =
+        MIRBuilder.buildConstant(LLT::scalar(32), 48).getReg(0);
+
+    Register NewSrc1 =
+        EltSize == 32 ? Src : MIRBuilder.buildBitcast(V16S32, Src).getReg(0);
+
+    // Shift the result to the lower 128 bits of a 512-bit vector, as the
+    // result is currently stored in the upper 128 bits of the vector.
+    MIRBuilder
+        .buildIntrinsic(Intrinsic::aie2_vshift_I512_I512, Vec512Reg, false,
```
gbossu: I don't think we should bring an AIE2-specific intrinsic here, this code should stay version agnostic I believe

llvm/test/CodeGen/AIE/aie2/GlobalISel/inst-select-aie-add-vector-elt-left.mir
```diff
@@ -21,8 +21,8 @@ body:             |
     ; CHECK-LABEL: name: vpush_int32_512
     ; CHECK: [[MOV_RLC_imm10_pseudo:%[0-9]+]]:er = MOV_RLC_imm10_pseudo 12
     ; CHECK-NEXT: [[DEF:%[0-9]+]]:vec512 = IMPLICIT_DEF
-    ; CHECK-NEXT: [[VPUSH_LO_32_:%[0-9]+]]:vec512 = VPUSH_LO_32 [[MOV_RLC_imm10_pseudo]], [[DEF]]
-    ; CHECK-NEXT: PseudoRET implicit $lr, implicit [[VPUSH_LO_32_]]
+    ; CHECK-NEXT: [[VPUSH_HI_32_:%[0-9]+]]:vec512 = VPUSH_HI_32 [[DEF]], [[MOV_RLC_imm10_pseudo]]
+    ; CHECK-NEXT: PseudoRET implicit $lr, implicit [[VPUSH_HI_32_]]
```
gbossu, line 25: I don't think there are changes to the instruction selection in this commit. Maybe these tests belong in another commit?

llvm/lib/Target/AIE/AIE2LegalizerInfo.cpp
```diff
@@ -461,6 +461,17 @@ AIE2LegalizerInfo::AIE2LegalizerInfo(const AIE2Subtarget &ST) : AIEHelper(ST) {
       .clampMaxNumElements(0, S32, 16)
       .custom();
 
+  getActionDefinitionsBuilder(G_BUILD_VECTOR_TRUNC)
+      .unsupportedIf(IsNotValidDestinationVector)
+      // We clamp the high values and not the low ones, sice the former
+      // splits the values but the latter keeps the same G_BUILD_VECTOR_TRUNC in
+      // the output instructions which causes an infinite loop since it
+      // can't reach our custom legalization code.
```
gbossu, line 469: I'm not sure I understand the comment. Do you mean we don't set a minimum element count because that would somehow trigger an infinite loop?

llvm/test/CodeGen/AIE/aie2/GlobalISel/legalize-build-vector-trunc.mir
```diff
@@ -0,0 +1,180 @@
+# NOTE: Assertions have been autogenerated by utils/update_mir_test_checks.py
+#
+# This file is licensed under the Apache License v2.0 with LLVM Exceptions.
+# See https://llvm.org/LICENSE.txt for license information.
+# SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+#
+# (c) Copyright 2023-2024 Advanced Micro Devices, Inc. or its affiliates
```
gbossu: Only 2024

llvm/test/CodeGen/AIE/aie2/GlobalISel/legalize-build-vector-trunc.mir
```diff
@@ -0,0 +1,180 @@
...
+  - {id: 0, name: "", type: default, offset: 0, size: 32, alignment: 32}
+body:             |
+  bb.1.entry:
+    %2:_(s32) = G_CONSTANT i32 9
+    %3:_(s32) = G_CONSTANT i32 12
+    %4:_(s32) = G_CONSTANT i32 13
+    %5:_(s32) = G_CONSTANT i32 0
+    %1:_(<16 x s16>) = G_BUILD_VECTOR_TRUNC %2(s32), %3(s32), %4(s32), %5(s32), %5(s32), %5(s32), %5(s32), %5(s32), %2(s32), %3(s32), %4(s32), %5(s32), %5(s32), %5(s32), %5(s32), %5(s32)
+    %0:_(p0) = G_FRAME_INDEX %stack.0
+    G_STORE %1(<16 x s16>), %0(p0) :: (store (<16 x s16>))
```
gbossu, line 150: There are no check lines?

llvm/lib/Target/AIE/AIELegalizerHelper.cpp
```diff
@@ -142,7 +142,7 @@ bool AIELegalizerHelper::legalizeG_BUILD_VECTOR(LegalizerHelper &Helper,
   assert(DstVecSize < 1024 && "vadd takes a 512-bit argument");
 
   // If our vector is 32-bit we can store it as packed integer vector
-  if (DstVecSize == 32)
+  if (DstVecSize == 32 && MI.getOpcode() == TargetOpcode::G_BUILD_VECTOR)
```
gbossu: Why do we need that extra check? Is the handling of 32-bit vectors different for `G_BUILD_VECTOR_TRUNC`?

llvm/lib/Target/AIE/AIE2PreLegalizerCombiner.cpp
```diff
@@ -136,13 +138,34 @@ bool AIE2PreLegalizerCombinerImpl::tryToCombineVectorShiftsByZero(
   return true;
 }
 
+bool AIE2PreLegalizerCombinerImpl::tryToCombineSetExtract(
```
gbossu, line 150: Can you add tests?

gbossu: FYI @andcarminati maybe you're also introducing something similar?

llvm/lib/Target/AIE/AIE2PreLegalizerCombiner.cpp
```diff
@@ -141,13 +143,26 @@ bool AIE2PreLegalizerCombinerImpl::tryToCombineVectorShiftsByZero(
...
+  }
+  case Intrinsic::aie2_set_I512_I256: {
+    if (cast<GIntrinsic>(*ExtOp).getIntrinsicID() !=
+        Intrinsic::aie2_ext_I256_I512)
+      return false;
+    break;
+  }
+  default:
+    return false;
+  }
```
gbossu, line 180: You seem to be "fixing" what was just defined in the previous commit. Maybe have all the changes for `tryToCombineSetExtract` in a single commit?

gbossu: (And add relevant test cases for the newly supported `aie2_ext_I256_I512`)

llvm/test/CodeGen/AIE/aie2/GlobalISel/combine-vinsert-sequence-prelegalizer.mir
```diff
@@ -0,0 +1,170 @@
+# NOTE: Assertions have been autogenerated by utils/update_mir_test_checks.py
+#
+# This file is licensed under the Apache License v2.0 with LLVM Exceptions.
+# See https://llvm.org/LICENSE.txt for license information.
+# SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+#
+# (c) Copyright 2024 Advanced Micro Devices, Inc. or its affiliates
+# RUN: llc -mtriple aie2 -run-pass=aie2-prelegalizer-combiner %s -verify-machineinstrs -o - | FileCheck %s
```
gbossu, line 8: Could you add more complicated "negative" test cases where e.g.
 - The is one lane index that does not have have a VINSERT
 - A lane is inserted multiple times

llvm/lib/Target/AIE/AIE2LegalizerInfo.cpp
abhinay-anubola: Low value clamping can cause an infinite loop because each split operation still uses G_BUILD_VECTOR_TRUNC, causing the compiler to repeatedly try to legalize the same instruction (for smaller vectors).

llvm/lib/Target/AIE/AIE2PreLegalizerCombiner.cpp
```diff
@@ -171,6 +173,65 @@ bool AIE2PreLegalizerCombinerImpl::tryToCombineSetExtract(
   return true;
 }
 
+/// Look for VINSERT sequence that can be rewritten as G_BUILD_VECTOR
+bool AIE2PreLegalizerCombinerImpl::tryToCombineVectorInserts(
+    MachineInstr &MI, int SclSrcBits) const {
+  const Register DstReg = MI.getOperand(0).getReg();
+  LLT DstRegTy = MRI.getType(DstReg);
+  const LLT DstVecEltTy = DstRegTy.getElementType();
+  SmallVector<Register, 16> Regs;
```
gbossu: Could you move that definition closer to its place of use? I.e. around `// Collect registers in order for G_BUILD_VECTOR`

llvm/lib/Target/AIE/AIE2PreLegalizerCombiner.cpp
```diff
@@ -171,6 +173,65 @@ bool AIE2PreLegalizerCombinerImpl::tryToCombineSetExtract(
...
+    // Collecting registers and their indices
+    auto Cst = getIConstantVRegValWithLookThrough(IdxReg, MRI);
+    if (!Cst ||
+        !RegMap.try_emplace(Cst->Value.getSExtValue(), SclSrcReg).second)
+      return false;
+    CurMI = getDefIgnoringCopies(SrcReg, MRI);
+  }
+
+  if (!IsSet(CurMI))
+    return false;
```
gbossu, line 248: Does this really matter? If we insert every lane, the source vector is irrelevant I believe. I wouldn't check what `CurMI` is, or can you think of a case where this is needed?

llvm/lib/Target/AIE/AIE2PreLegalizerCombiner.cpp
```diff
@@ -171,6 +173,65 @@ bool AIE2PreLegalizerCombinerImpl::tryToCombineSetExtract(
...
+    const Register IdxReg = CurMI->getOperand(3).getReg();
+    const Register SclSrcReg = CurMI->getOperand(4).getReg();
+
+    // Collecting registers and their indices
+    auto Cst = getIConstantVRegValWithLookThrough(IdxReg, MRI);
+    if (!Cst ||
+        !RegMap.try_emplace(Cst->Value.getSExtValue(), SclSrcReg).second)
+      return false;
+    CurMI = getDefIgnoringCopies(SrcReg, MRI);
+  }
```
gbossu, line 221: How about wrapping that in a function call? E.g.
```
std::map<int, Register> InsertIndices = getVectorInsertIndices(CurMI);
```

llvm/test/CodeGen/AIE/aie2/GlobalISel/combine-vinsert-sequence-prelegalizer.mir
```diff
@@ -0,0 +1,330 @@
...
+    ; CHECK-NEXT: [[C8:%[0-9]+]]:_(s32) = G_CONSTANT i32 8
+    ; CHECK-NEXT: [[C9:%[0-9]+]]:_(s32) = G_CONSTANT i32 9
+    ; CHECK-NEXT: [[C10:%[0-9]+]]:_(s32) = G_CONSTANT i32 10
+    ; CHECK-NEXT: [[C11:%[0-9]+]]:_(s32) = G_CONSTANT i32 11
+    ; CHECK-NEXT: [[C12:%[0-9]+]]:_(s32) = G_CONSTANT i32 12
+    ; CHECK-NEXT: [[C13:%[0-9]+]]:_(s32) = G_CONSTANT i32 13
+    ; CHECK-NEXT: [[C14:%[0-9]+]]:_(s32) = G_CONSTANT i32 14
+    ; CHECK-NEXT: [[C15:%[0-9]+]]:_(s32) = G_CONSTANT i32 15
+    ; CHECK-NEXT: [[BUILD_VECTOR_TRUNC:%[0-9]+]]:_(<16 x s8>) = G_BUILD_VECTOR_TRUNC [[C15]](s32), [[C13]](s32), [[C11]](s32), [[C9]](s32), [[C7]](s32), [[C5]](s32), [[C3]](s32), [[C1]](s32), [[C]](s32), [[C2]](s32), [[C4]](s32), [[C6]](s32), [[C8]](s32), [[C10]](s32), [[C12]](s32), [[C14]](s32)
+    ; CHECK-NEXT: [[BITCAST:%[0-9]+]]:_(<4 x s32>) = G_BITCAST [[BUILD_VECTOR_TRUNC]](<16 x s8>)
```
gbossu: Do we really need the bitcast? Can't `G_AIE_PAD_VECTOR_UNDEF` handle the original type?

llvm/lib/Target/AIE/AIE2PreLegalizerCombiner.cpp
```diff
@@ -171,6 +173,65 @@ bool AIE2PreLegalizerCombinerImpl::tryToCombineSetExtract(
...
+bool AIE2PreLegalizerCombinerImpl::tryToCombineVectorInserts(
+    MachineInstr &MI, int SclSrcBits) const {
+  const Register DstReg = MI.getOperand(0).getReg();
+  LLT DstRegTy = MRI.getType(DstReg);
+  const LLT DstVecEltTy = DstRegTy.getElementType();
+  SmallVector<Register, 16> Regs;
+  std::map<int, Register> RegMap;
+  MachineInstr *CurMI = &MI;
+
+  auto IsVInsert = [](const MachineInstr *MI) {
```
gbossu: Given we only check `aie2_vinsert8_I512` so far, I would name that lambda `Is8BitVInsert `

llvm/lib/Target/AIE/AIE2PreLegalizerCombiner.cpp
gbossu: Oh I get it, for small 128-bit vectors, we do not actually define all lanes. But I think we should look a bit further then and detect that `CurMI` actually represent an `IMPLICIT_DEF`. We could maybe even write another combiner for that. 

If our source MI is detected as an implicit def, or all lanes are inserted, then the transformation is safe.

llvm/lib/Target/AIE/AIELegalizerHelper.cpp
```diff
@@ -182,9 +194,30 @@ bool AIELegalizerHelper::legalizeG_BUILD_VECTOR(LegalizerHelper &Helper,
   //    https://github.com/llvm/llvm-project/pull/84538
   if (DstVecSize == 256) {
     const Register UnusedSubReg = MRI.createGenericVirtualRegister(DstVecTy);
-    MIRBuilder.buildUnmerge({DstReg, UnusedSubReg}, Src);
+    MIRBuilder.buildUnmerge({UnusedSubReg, DstReg}, Src);
```
gbossu, line 196: Why did this change?

abhinay-anubola: Because now we need higher 256 bits (As we are using VPUSH.hi now)

llvm/lib/Target/AIE/AIELegalizerHelper.cpp
```diff
@@ -53,8 +53,9 @@ bool AIELegalizerHelper::pack32BitVector(LegalizerHelper &Helper,
   const LLT RegTy = MRI.getType(DstReg);
   while (Operand != OperandEnd) {
     Register DestinationOperand = Operand->getReg();
+    const LLT DstOpTy = MRI.getType(DestinationOperand);
 
-    if (RegTy.getScalarSizeInBits() != 32) {
+    if (RegTy.getScalarSizeInBits() != 32 && DstOpTy.getSizeInBits() != 32) {
```
gbossu: I still don't get why we have to verify this now? What changed in the code?

llvm/lib/Target/AIE/AIELegalizerHelper.cpp
```diff
@@ -153,17 +162,20 @@ bool AIELegalizerHelper::legalizeG_BUILD_VECTOR(LegalizerHelper &Helper,
...
-    if (DstVecSize == 512 && Operand == OperandBegin + 1)
+    if (DstVecSize == 512 && Operand == OperandBegin + 1) {
       Dst = DstReg;
+    }
 
     // vpush takes 32-bit operands so we sign extend the input variable. This is
     // required here since we don't have 16 or 32-bit registers.
-    if (DstVecEltTy.getSizeInBits() != 32) {
+    if (DstVecEltTy.getSizeInBits() != 32 &&
+        MI.getOpcode() == TargetOpcode::G_BUILD_VECTOR) {
```
gbossu: Does that mean `G_BUILD_VECTOR_TRUNC` can accept `s8` and  `s16` types? I think we should always extend to 32-bits, because this is the size of our GPRs.

llvm/lib/Target/AIE/AIELegalizerHelper.cpp
gbossu: Maybe add that as a comment? We are using `getGenericAddVectorEltOpcode()` which adds elements from the hi bits, so the 256-bit result will be in the hi part of the 512-bit vector.

llvm/lib/Target/AIE/AIELegalizerHelper.cpp
```diff
@@ -153,17 +162,20 @@ bool AIELegalizerHelper::legalizeG_BUILD_VECTOR(LegalizerHelper &Helper,
 
   const AIEBaseInstrInfo *II = ST.getInstrInfo();
   MachineOperand *OperandBegin = MI.operands_begin(),
-                 *Operand = MI.operands_end() - 1;
-  while (Operand != OperandBegin) {
+                 *Operand = MI.operands_begin() + 1,
+                 *OperandEnd = MI.operands_end();
+  while (Operand != OperandEnd) {
```
gbossu: Any particular reason why we are reversing the iteration order?

llvm/test/CodeGen/AIE/aie2/GlobalISel/legalize-build-vector-trunc.mir
```diff
@@ -0,0 +1,354 @@
...
+    ; CHECK-NEXT: [[BITCAST1:%[0-9]+]]:_(<64 x s8>) = G_BITCAST [[INT]](<16 x s32>)
+    ; CHECK-NEXT: [[AIE_UNPAD_VECTOR:%[0-9]+]]:_(<16 x s8>) = G_AIE_UNPAD_VECTOR [[BITCAST1]](<64 x s8>)
+    ; CHECK-NEXT: [[FRAME_INDEX:%[0-9]+]]:_(p0) = G_FRAME_INDEX %stack.0
+    ; CHECK-NEXT: G_STORE [[AIE_UNPAD_VECTOR]](<16 x s8>), [[FRAME_INDEX]](p0) :: (store (<16 x s8>))
+    ; CHECK-NEXT: PseudoRET implicit $lr
+    %2:_(s32) = G_CONSTANT i32 42
+    %3:_(s32) = G_CONSTANT i32 19
+    %4:_(s32) = G_CONSTANT i32 32
+    %5:_(s32) = G_CONSTANT i32 12
+    %1:_(<16 x s8>) = G_BUILD_VECTOR_TRUNC %2(s32), %3(s32), %4(s32), %5(s32), %2(s32), %3(s32), %4(s32), %5(s32), %2(s32), %3(s32), %4(s32), %5(s32), %2(s32), %3(s32), %4(s32), %5(s32)
```
gbossu, line 51: Is it possible to get a `G_BUILD_VECTOR_TRUNC` with non-`s32` inputs?

llvm/lib/Target/AIE/AIE2LegalizerInfo.cpp
```diff
@@ -461,6 +461,17 @@ AIE2LegalizerInfo::AIE2LegalizerInfo(const AIE2Subtarget &ST) : AIEHelper(ST) {
       .clampMaxNumElements(0, S32, 16)
       .custom();
 
+  getActionDefinitionsBuilder(G_BUILD_VECTOR_TRUNC)
```
gbossu, line 464: Do you know how those `G_BUILD_VECTOR_TRUNC` get generated and why we need to support them only now?

llvm/lib/Target/AIE/AIELegalizerHelper.cpp
abhinay-anubola: As we changed corresponding instruction of `G_AIE_ADD_VECTOR_ELT_LEFT` from vpush.lo to vpush.hi.

llvm/lib/Target/AIE/AIE2LegalizerInfo.cpp
abhinay-anubola: We need `G_BUILD_VECTOR_TRUNC` because input element for VInsert.8 is `s32`, if we use existing `G_BUILD_VECTOR` we need to build truncations that truncates `s32` into `s8`.  So, we are generating `G_BUILD_VECTOR_TRUNC` from combining those vinserts.

llvm/lib/Target/AIE/AIELegalizerHelper.cpp
abhinay-anubola: Earlier this `pack32BitVector` used by G_BUILD_VECTOR where inputs are `s8` or `s16`, here they were zero extended. But for G_BUILD_VECTOR_TRUNC where the inputs are `s32`, there is no necessity for ZEXT

llvm/test/CodeGen/AIE/aie2/GlobalISel/legalize-build-vector-trunc.mir
abhinay-anubola: As documentation suggests
**The _TRUNC version truncates the larger operand types to fit the destination vector elt type.**
It is possible, but for now we are using only `s32` input type.

llvm/lib/Target/AIE/AIELegalizerHelper.cpp
abhinay-anubola: Maybe I will add a check for input type `s32`

llvm/lib/Target/AIE/AIE2LegalizerInfo.cpp
gbossu: Makes sense, better avoid extra tunc/ext :)

llvm/lib/Target/AIE/AIELegalizerHelper.cpp
gbossu: Could we remove `RegTy.getScalarSizeInBits() != 32` then? This seems redundant, as we check a "safer" predicate now.

llvm/lib/Target/AIE/AIELegalizerHelper.cpp
gbossu: My point is: we should always have `s32` inputs, because this is the size of our scalar operands. So I think regardless of the opcode we should extend the operands to `s32`, unless they already are of that type ofc ;)

llvm/lib/Target/AIE/AIELegalizerHelper.cpp
gbossu: Oh so actually it's the previous order that was "un-natural", now it all makes sense.

llvm/lib/Target/AIE/AIELegalizerHelper.cpp
```diff
@@ -153,17 +162,20 @@ bool AIELegalizerHelper::legalizeG_BUILD_VECTOR(LegalizerHelper &Helper,
 
   const AIEBaseInstrInfo *II = ST.getInstrInfo();
   MachineOperand *OperandBegin = MI.operands_begin(),
```
gbossu: Nit: Why not have `MachineOperand *OperandBegin = MI.operands_begin() + 1`, given that's it's actually the first operand we care about?

llvm/lib/Target/AIE/AIE2InstructionSelector.cpp
```diff
@@ -814,25 +818,23 @@ bool AIE2InstructionSelector::selectG_AIE_ADD_VECTOR_ELT_LEFT(
   unsigned Opcode;
   switch (VecEltDstTySize) {
   case 8:
-    Opcode = AIE2::VPUSH_LO_8;
+    Opcode = AIE2::VPUSH_HI_8;
```
andcarminati, line 821: Now I think I am a bit confused.  Did we change G_AIE_ADD_VECTOR_ELT_LEFT to add to the right? Should we have GInst here?

gbossu: I would actually vouch for renaming this `G_AIE_ADD_VECTOR_ELT_HI`. "left" is too ambiguous

andcarminati: [X] `G_AIE_ADD_VECTOR_ELT_HI`

llvm/lib/Target/AIE/AIE2PreLegalizerCombiner.cpp
abhinay-anubola: `%64:_(<16 x s8>) = G_INTRINSIC intrinsic(@llvm.aie2.v16int8)                                                
 %65:_(<4 x s32>) = G_BITCAST %64(<16 x s8>)                                                                 
 %66:_(<16 x s32>) = G_INTRINSIC intrinsic(@llvm.aie2.set.I512.I128), %65(<4 x s32>)`
 There is no `IMPLICIT_DEF` further, these are the instructions above the topmost vinsert.

llvm/lib/Target/AIE/AIE2LegalizerInfo.cpp
```diff
@@ -461,6 +461,17 @@ AIE2LegalizerInfo::AIE2LegalizerInfo(const AIE2Subtarget &ST) : AIEHelper(ST) {
...
+      .unsupportedIf(IsNotValidDestinationVector)
+      // We clamp the high values and not the low ones, sice the former
```
andcarminati: nit: sice

llvm/lib/Target/AIE/AIE2InstructionSelector.cpp
```diff
@@ -5016,6 +5017,29 @@ bool AIE2InstructionSelector::selectSetI128(MachineInstr &I,
   return true;
 }
 
+bool AIE2InstructionSelector::selectG_AIE_PAD_VECTOR_UNDEF(
+    MachineInstr &I, MachineOperand &DstReg, MachineOperand &SrcReg,
+    MachineRegisterInfo &MRI) {
+  LLT SrcTy = MRI.getType(SrcReg.getReg());
```
andcarminati: nit: `const LLT`

llvm/lib/Target/AIE/AIE2PreLegalizerCombiner.cpp
andcarminati: We have some new ones on Post Custom Combiner now, more related to vector updates, so it is different ;-).

llvm/lib/Target/AIE/AIE2PreLegalizerCombiner.cpp
```diff
@@ -136,13 +140,115 @@ bool AIE2PreLegalizerCombinerImpl::tryToCombineVectorShiftsByZero(
...
+    if (cast<GIntrinsic>(*ExtOp).getIntrinsicID() !=
+        Intrinsic::aie2_extract_I128_I512)
+      return false;
+    break;
+  }
+  case Intrinsic::aie2_set_I512_I256: {
+    const Register IdxReg = MI.getOperand(3).getReg();
+    if (cast<GIntrinsic>(*ExtOp).getIntrinsicID() !=
+            Intrinsic::aie2_ext_I256_I512 ||
+        IdxReg != ExtOp->getOperand(3).getReg())
```
andcarminati: I suggest to include one more case here: registers are different but they trace back to the equal G_CONSTANT. Maybe this can be the only rule here. I saw cases where we have an exclusive register for each intrinsic, so it can be more general.

llvm/test/CodeGen/AIE/aie2/GlobalISel/combine-set-extract-prelegalizer.mir
```diff
@@ -0,0 +1,66 @@
...
+legalized:       false
+body:             |
+  bb.1.entry:
+    liveins: $x0
+    ; CHECK-LABEL: name: set-extract-I256_different_idx
+    ; CHECK: liveins: $x0
+    ; CHECK-NEXT: {{  $}}
+    ; CHECK-NEXT: [[COPY:%[0-9]+]]:_(<16 x s32>) = COPY $x0
+    ; CHECK-NEXT: [[C:%[0-9]+]]:_(s32) = G_CONSTANT i32 0
+    ; CHECK-NEXT: [[C1:%[0-9]+]]:_(s32) = G_CONSTANT i32 1
```
andcarminati, line 75: Even if the constant are the same, we cannot match this case, i think. I suggest an additional test with:

```
    %1:_(s32) = G_CONSTANT i32 0
    %10:_(s32) = G_CONSTANT i32 0
```

llvm/lib/Target/AIE/AIE2PreLegalizerCombiner.cpp
```diff
@@ -136,13 +140,115 @@ bool AIE2PreLegalizerCombinerImpl::tryToCombineVectorShiftsByZero(
...
+  }
+
+  MachineIRBuilder MIRBuilder(MI);
+  MIRBuilder.buildCopy(DstReg, ExtOp->getOperand(2).getReg());
+  MI.eraseFromParent();
+
+  return true;
+}
+
+std::optional<std::pair<std::map<int, Register>, int>>
```
andcarminati: Can we have some comment about this complex return type?

llvm/lib/Target/AIE/AIE2PreLegalizerCombiner.cpp
```diff
@@ -136,13 +140,115 @@ bool AIE2PreLegalizerCombinerImpl::tryToCombineVectorShiftsByZero(
...
+  while (Is8BitVInsert(CurMI)) {
+    // In this case of G_INTRINSIC operand 1 is target intrinsic
+    const Register SrcReg = CurMI->getOperand(2).getReg();
+    const Register IdxReg = CurMI->getOperand(3).getReg();
+    const Register SclSrcReg = CurMI->getOperand(4).getReg();
+
+    // Collecting registers and their indices
+    auto Cst = getIConstantVRegValWithLookThrough(IdxReg, MRI);
+    if (!Cst ||
+        !RegMap.try_emplace(Cst->Value.getSExtValue(), SclSrcReg).second)
```
andcarminati: Can we have negative index? What about ZExt? I suggest to use unsigned for index and size.

llvm/lib/Target/AIE/AIE2PreLegalizerCombiner.cpp
```diff
@@ -136,13 +140,115 @@ bool AIE2PreLegalizerCombinerImpl::tryToCombineVectorShiftsByZero(
...
+  auto InsertIndicesInfo = getVectorInsertIndices(CurMI, MRI);
+  if (!InsertIndicesInfo)
+    return false;
+  std::map<int, Register> InsertIndices = InsertIndicesInfo->first;
+  int DstRegLen = InsertIndicesInfo->second / SclSrcBits;
+
+  MachineIRBuilder MIRBuilder(MI);
+  SmallVector<Register, 16> Regs;
+  // Collect registers in order for G_BUILD_VECTOR
+  for (int I = 0; I < DstRegLen; I++) {
```
andcarminati: nit: `unsigned`

llvm/lib/Target/AIE/AIE2PreLegalizerCombiner.cpp
```diff
@@ -136,13 +140,115 @@ bool AIE2PreLegalizerCombinerImpl::tryToCombineVectorShiftsByZero(
...
+    CurMI = getDefIgnoringCopies(SrcReg, MRI);
+  }
+
+  if (!IsSet(CurMI))
+    return std::nullopt;
+  int DstRegBits = MRI.getType(CurMI->getOperand(2).getReg()).getSizeInBits();
+  return std::make_pair(RegMap, DstRegBits);
+}
+
+/// Look for VINSERT sequence that can be rewritten as G_BUILD_VECTOR
```
andcarminati: nit: `*_TRUNC`

llvm/lib/Target/AIE/AIE2PreLegalizerCombiner.cpp
```diff
@@ -136,13 +140,115 @@ bool AIE2PreLegalizerCombinerImpl::tryToCombineVectorShiftsByZero(
...
+bool AIE2PreLegalizerCombinerImpl::tryToCombineVectorInserts(
+    MachineInstr &MI, int SclSrcBits) const {
+  std::map<int, Register> RegMap;
+  MachineInstr *CurMI = &MI;
+
+  auto InsertIndicesInfo = getVectorInsertIndices(CurMI, MRI);
+  if (!InsertIndicesInfo)
+    return false;
+  std::map<int, Register> InsertIndices = InsertIndicesInfo->first;
+  int DstRegLen = InsertIndicesInfo->second / SclSrcBits;
```
andcarminati: nit: `unsigned`.

llvm/lib/Target/AIE/AIE2PreLegalizerCombiner.cpp
gbossu: I think when I suggested outlining that code, I imagined a `std::map<int, Register>` return type. Is there anything that prevents using this simpler return type @abhinay-anubola ?

abhinay-anubola: While building `G_BUILD_VETCOR_TRUNC` we need the size of vector, size is calculated in this function and returned.

gbossu: Here's what I would do: I think this function should only return something when it could safely follow the chain of VINSERT. This means:
 - We ended up on a `set` intrinsic (so far I think it's sufficient to consider it the only safe "source")
 - We collected the right amount of lanes to cover the size of the input vector to the `set` intrinsic

In that case, we return a map with all the lane mappings, and otherwise we just return an empty map.

We then do not need to return the "source vector" size, because then in `tryToCombineVectorInserts` we can just iterate on every lane from `0` to `InsertIndices.size()`

andcarminati:
Hi @abhinay-anubola, this PR is getting very nice, but with MIR simplifications we can expect side effects as well. Do you have some updated QoR numbers to share?

abhinay-anubola:
I ran QoR with the updated changes, but the results remain unchanged compared to the previous run.

llvm/lib/Target/AIE/AIE2PreLegalizerCombiner.cpp
```diff
@@ -136,13 +140,130 @@ bool AIE2PreLegalizerCombinerImpl::tryToCombineVectorShiftsByZero(
...
+
+  MachineIRBuilder MIRBuilder(MI);
+  MIRBuilder.buildCopy(DstReg, ExtOp->getOperand(2).getReg());
+  MI.eraseFromParent();
+
+  return true;
+}
+
+// Returns a map with InsertIndices and registers holding the insert values.
+std::optional<std::map<int, Register>>
```
andcarminati: nit: this can be unsigned, because the map is accessed with unsigned.

llvm/lib/Target/AIE/AIELegalizerHelper.cpp
```diff
@@ -124,6 +125,14 @@ bool AIELegalizerHelper::unpack32BitVector(LegalizerHelper &Helper,
   return true;
 }
 
+/// @brief Get the AIE intrinsic corresponding to the VSHIFT.
+static unsigned getVshiftIntrID(const AIEBaseSubtarget &ST) {
```
andcarminati: nit: `getVShiftIntrID`

llvm/lib/Target/AIE/AIE2InstrInfo.cpp
```diff
@@ -737,7 +737,7 @@ unsigned AIE2InstrInfo::getCycleSeparatorOpcode() const {
 }
 
 unsigned AIE2InstrInfo::getGenericAddVectorEltOpcode() const {
-  return AIE2::G_AIE_ADD_VECTOR_ELT_LEFT;
```
gbossu, line 740: Please update the doc in the base class:
```
  /// Return the opcode to be used for pushing a vector element at the LSB
  /// position in a vector
  virtual unsigned getGenericAddVectorEltOpcode() const {
    llvm_unreachable("Target didn't implement getGenericAddVectorEltOpcode");
  }
```

gbossu:
LGTM, I just think it is important to fix https://github.com/Xilinx/llvm-aie/pull/187#discussion_r1810350896

gbossu:
I looked at QoR and this seems to cause regressions more than improvements, I'd like to understand them before merging

llvm/lib/Target/AIE/AIELegalizerHelper.cpp
```diff
@@ -152,28 +161,31 @@ bool AIELegalizerHelper::legalizeG_BUILD_VECTOR(LegalizerHelper &Helper,
...
 
   const AIEBaseInstrInfo *II = ST.getInstrInfo();
-  MachineOperand *OperandBegin = MI.operands_begin(),
-                 *Operand = MI.operands_end() - 1;
-  while (Operand != OperandBegin) {
-    Register Reg = Operand->getReg();
+  MachineOperand *OperandBegin = MI.operands_begin() + 1,
+                 *Operand = MI.operands_begin() + 1,
+                 *OperandEnd = MI.operands_end();
+  while (Operand != OperandEnd) {
```
martien-de-jong: If there's any way to write this as a range-based for, I would be happier. Aren't we traversing all uses of MI here?

llvm/lib/Target/AIE/AIELegalizerHelper.cpp
```diff
@@ -152,28 +161,31 @@ bool AIELegalizerHelper::legalizeG_BUILD_VECTOR(LegalizerHelper &Helper,
...
+    Register EltReg = Operand->getReg();
+    LLT EltRegTy = MRI.getType(EltReg);
     Register Dst = MRI.createGenericVirtualRegister(VecTy);
 
-    if (DstVecSize == 512 && Operand == OperandBegin + 1)
+    if (DstVecSize == 512 && Operand == OperandBegin) {
       Dst = DstReg;
```
martien-de-jong, line 171: nit: we've created an unused virtual register.

llvm/lib/Target/AIE/AIE2PreLegalizerCombiner.cpp
```diff
@@ -136,13 +138,55 @@ bool AIE2PreLegalizerCombinerImpl::tryToCombineVectorShiftsByZero(
...
+  if (!isa<GIntrinsic>(MI) || !isa<GIntrinsic>(*ExtOp))
+    return false;
+  switch (cast<GIntrinsic>(MI).getIntrinsicID()) {
+  case Intrinsic::aie2_set_I512_I128: {
+    if (cast<GIntrinsic>(*ExtOp).getIntrinsicID() !=
+        Intrinsic::aie2_extract_I128_I512)
+      return false;
+    break;
+  }
+  case Intrinsic::aie2_set_I512_I256: {
```
martien-de-jong, line 164: I don't see a call to this function that would land here?

llvm/lib/Target/AIE/AIE2PreLegalizerCombiner.cpp
```diff
@@ -136,13 +138,55 @@ bool AIE2PreLegalizerCombinerImpl::tryToCombineVectorShiftsByZero(
...
+    if (cast<GIntrinsic>(*ExtOp).getIntrinsicID() !=
+        Intrinsic::aie2_ext_I256_I512)
+      return false;
+    const Register SetOpIdxReg = MI.getOperand(3).getReg();
+    const Register ExtOpIdxReg = ExtOp->getOperand(3).getReg();
+    auto SetOpCst = getIConstantVRegValWithLookThrough(SetOpIdxReg, MRI);
+    auto ExtOpCst = getIConstantVRegValWithLookThrough(ExtOpIdxReg, MRI);
+    if (SetOpIdxReg != ExtOpIdxReg &&
```
martien-de-jong, line 172: nit: if they are the same, we only need one `getIConstantVRegValWithLookThrough` and if we don't have SetOpCst, we don't need to find ExtOpCst.

llvm/lib/Target/AIE/AIE2PreLegalizerCombiner.cpp
abhinay-anubola: There is call from `tryToCombineVectorInserts` that will land here.

llvm/lib/Target/AIE/AIELegalizerHelper.cpp
abhinay-anubola: Is this about the register `Dst`, we have used later.

llvm/lib/Target/AIE/AIE2PreLegalizerCombiner.cpp
abhinay-anubola: In that case we will end up having two if blocks.

llvm/lib/Target/AIE/AIE2PreLegalizerCombiner.cpp
```diff
@@ -136,13 +138,55 @@ bool AIE2PreLegalizerCombinerImpl::tryToCombineVectorShiftsByZero(
...
 bool AIE2PreLegalizerCombinerImpl::tryToCombineIntrinsic(
     MachineInstr &MI) const {
 
   switch (cast<GIntrinsic>(MI).getIntrinsicID()) {
   case Intrinsic::aie2_vshift_I512_I512: {
     return CombineVecShiftByZero && tryToCombineVectorShiftsByZero(MI);
   }
+  case Intrinsic::aie2_set_I512_I128: {
+    return tryToCombineSetExtract(MI);
+  }
```
gbossu, line 282: `tryToCombineSetExtract` also supports 256bit vectors, I find a bit sad to not make use of it at all. I think we should either:
 - Delete that code to avoid confusion
 - Or guard the combiner with something like
    ```
    case Intrinsic::aie2_set_I512_I256: {
      return Combine256To512SetExtract && tryToCombineSetExtract(MI);
    }
    ```
    (and add/re-add tests of course by passing `--combine-256-to-512-set-extract` specifically in the .mir test)

llvm/lib/Target/AIE/AIE2PreLegalizerCombiner.cpp
```diff
@@ -177,6 +182,81 @@ bool AIE2PreLegalizerCombinerImpl::tryToCombineSetExtract(
...
+    // In this case of G_INTRINSIC operand 1 is target intrinsic
+    const Register SrcReg = CurMI->getOperand(2).getReg();
+    const Register IdxReg = CurMI->getOperand(3).getReg();
+    const Register SclSrcReg = CurMI->getOperand(4).getReg();
+
+    // Collecting registers and their indices
+    auto Cst = getIConstantVRegValWithLookThrough(IdxReg, MRI);
+    if (!Cst ||
+        !RegMap.try_emplace(Cst->Value.getZExtValue(), SclSrcReg).second)
+      break;
```
gbossu: Why not just `return {};`?

llvm/lib/Target/AIE/AIE2PreLegalizerCombiner.cpp
```diff
@@ -177,6 +182,81 @@ bool AIE2PreLegalizerCombinerImpl::tryToCombineSetExtract(
   return true;
 }
 
+// Returns a map with InsertIndices and registers holding the insert values.
+std::optional<std::map<unsigned, Register>>
```
gbossu: I feel we could simplify that return type. How about just `std::map<unsigned, Register>`? The returned map would be empty if we did not manage to gather all indices.

llvm/lib/Target/AIE/AIE2PreLegalizerCombiner.cpp
```diff
@@ -177,6 +182,81 @@ bool AIE2PreLegalizerCombinerImpl::tryToCombineSetExtract(
...
+    // Collecting registers and their indices
+    auto Cst = getIConstantVRegValWithLookThrough(IdxReg, MRI);
+    if (!Cst ||
+        !RegMap.try_emplace(Cst->Value.getZExtValue(), SclSrcReg).second)
+      break;
+    CurMI = getDefIgnoringCopies(SrcReg, MRI);
+
+    // Combining Set and Extract to fetch next VInsert
+    if (IsSet(CurMI) && tryToCombineSetExtract(*CurMI))
+      CurMI = getDefIgnoringCopies(SrcReg, MRI);
```
gbossu, line 220: Isn't `tryToCombineSetExtract` potentially deleting `CurMI`? Is it always safe to call `CurMI = getDefIgnoringCopies(SrcReg, MRI);`?

llvm/test/CodeGen/AIE/aie2/GlobalISel/combine-vinsert-sequence-prelegalizer.mir
```diff
@@ -0,0 +1,328 @@
...
+    ; CHECK-NEXT: [[C8:%[0-9]+]]:_(s32) = G_CONSTANT i32 8
+    ; CHECK-NEXT: [[C9:%[0-9]+]]:_(s32) = G_CONSTANT i32 9
+    ; CHECK-NEXT: [[C10:%[0-9]+]]:_(s32) = G_CONSTANT i32 10
+    ; CHECK-NEXT: [[C11:%[0-9]+]]:_(s32) = G_CONSTANT i32 11
+    ; CHECK-NEXT: [[C12:%[0-9]+]]:_(s32) = G_CONSTANT i32 12
+    ; CHECK-NEXT: [[C13:%[0-9]+]]:_(s32) = G_CONSTANT i32 13
+    ; CHECK-NEXT: [[C14:%[0-9]+]]:_(s32) = G_CONSTANT i32 14
+    ; CHECK-NEXT: [[C15:%[0-9]+]]:_(s32) = G_CONSTANT i32 15
+    ; CHECK-NEXT: [[BUILD_VECTOR_TRUNC:%[0-9]+]]:_(<16 x s8>) = G_BUILD_VECTOR_TRUNC [[C15]](s32), [[C13]](s32), [[C11]](s32), [[C9]](s32), [[C7]](s32), [[C5]](s32), [[C3]](s32), [[C1]](s32), [[C]](s32), [[C2]](s32), [[C4]](s32), [[C6]](s32), [[C8]](s32), [[C10]](s32), [[C12]](s32), [[C14]](s32)
+    ; CHECK-NEXT: [[AIE_PAD_VECTOR_UNDEF:%[0-9]+]]:_(<16 x s32>) = G_AIE_PAD_VECTOR_UNDEF [[BUILD_VECTOR_TRUNC]](<16 x s8>)
```
gbossu: I find the `<16 x s32>` type a bit confusing. Shouldn't we have `<64 x s8>`? We are just adding `undef` lanes, not changing the size of the lanes.

llvm/lib/Target/AIE/AIE2PreLegalizerCombiner.cpp
```diff
@@ -177,6 +182,81 @@ bool AIE2PreLegalizerCombinerImpl::tryToCombineSetExtract(
...
+  SmallVector<Register, 16> Regs;
+  // Collect registers in order for G_BUILD_VECTOR_TRUNC
+  for (unsigned I = 0; I < DstRegLen; I++) {
+    auto It = InsertIndices->find(I);
+    if (It == InsertIndices->end())
+      return false;
+    Regs.push_back(It->second);
+  }
+  Register DstRegTrunc = MRI.createGenericVirtualRegister(
+      LLT::fixed_vector(DstRegLen, SclSrcBits));
```
gbossu, line 260: Could we re-use the type from `MI.getOperand(0)` instead of trying to reconstruct it with `LLT::fixed_vector(DstRegLen, SclSrcBits)`?

llvm/test/CodeGen/AIE/aie2/GlobalISel/combine-vinsert-sequence-prelegalizer.mir
```diff
@@ -0,0 +1,328 @@
+# NOTE: Assertions have been autogenerated by utils/update_mir_test_checks.py
+#
+# This file is licensed under the Apache License v2.0 with LLVM Exceptions.
+# See https://llvm.org/LICENSE.txt for license information.
+# SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+#
+# (c) Copyright 2024 Advanced Micro Devices, Inc. or its affiliates
+# RUN: llc -mtriple aie2 -run-pass=aie2-prelegalizer-combiner %s -o - | FileCheck %s
```
gbossu: Can you run with `-varify-machineinstrs`?

llvm/lib/Target/AIE/AIE2PreLegalizerCombiner.cpp
abhinay-anubola: We can't do that, as they both are not same. In a case `<16 x s32>` is type of `MI.getOperand(0)` and we need `<16 x s8>` for `buildBuildVectorTrunc`

llvm/test/CodeGen/AIE/aie2/GlobalISel/combine-vinsert-sequence-prelegalizer.mir
abhinay-anubola: If we do that, we need to add extra bitcast from `<64 x s8>` to `<16 x s32>` to match the type of `MI.getOperand(0)`

llvm/lib/Target/AIE/AIE2PreLegalizerCombiner.cpp
gbossu: Why is it like this actually, do our VINSERT.S8 intrinsics take a `<16 x s32>` type?

abhinay-anubola: yes, they take `<16 x s32>` type

gbossu (approved):
LGTM, nice work :)

llvm/test/CodeGen/AIE/aie2/GlobalISel/legalize-build-vector.mir
```diff
@@ -105,40 +105,40 @@ body:             |
...
+    ; CHECK-NEXT: [[AIE_ADD_VECTOR_ELT_HI25:%[0-9]+]]:_(<16 x s32>) = G_AIE_ADD_VECTOR_ELT_HI [[AIE_ADD_VECTOR_ELT_HI24]], [[C3]](s32)
+    ; CHECK-NEXT: [[AIE_ADD_VECTOR_ELT_HI26:%[0-9]+]]:_(<16 x s32>) = G_AIE_ADD_VECTOR_ELT_HI [[AIE_ADD_VECTOR_ELT_HI25]], [[C3]](s32)
+    ; CHECK-NEXT: [[AIE_ADD_VECTOR_ELT_HI27:%[0-9]+]]:_(<16 x s32>) = G_AIE_ADD_VECTOR_ELT_HI [[AIE_ADD_VECTOR_ELT_HI26]], [[C3]](s32)
+    ; CHECK-NEXT: [[AIE_ADD_VECTOR_ELT_HI28:%[0-9]+]]:_(<16 x s32>) = G_AIE_ADD_VECTOR_ELT_HI [[AIE_ADD_VECTOR_ELT_HI27]], [[C3]](s32)
+    ; CHECK-NEXT: [[AIE_ADD_VECTOR_ELT_HI29:%[0-9]+]]:_(<16 x s32>) = G_AIE_ADD_VECTOR_ELT_HI [[AIE_ADD_VECTOR_ELT_HI28]], [[C3]](s32)
+    ; CHECK-NEXT: [[AIE_ADD_VECTOR_ELT_HI30:%[0-9]+]]:_(<16 x s32>) = G_AIE_ADD_VECTOR_ELT_HI [[AIE_ADD_VECTOR_ELT_HI29]], [[C3]](s32)
+    ; CHECK-NEXT: [[AIE_ADD_VECTOR_ELT_HI31:%[0-9]+]]:_(<16 x s32>) = G_AIE_ADD_VECTOR_ELT_HI [[AIE_ADD_VECTOR_ELT_HI30]], [[C3]](s32)
+    ; CHECK-NEXT: [[CONCAT_VECTORS:%[0-9]+]]:_(<32 x s32>) = G_CONCAT_VECTORS [[AIE_ADD_VECTOR_ELT_HI]](<16 x s32>), [[AIE_ADD_VECTOR_ELT_HI16]](<16 x s32>)
```
konstantinschwarz, line 141: This is wrong code : we need to concatenate `AIE_ADD_VECTOR_ELT_HI15` and `AIE_ADD_VECTOR_ELT_HI31`.
As it is now, all of the following `G_AIE_ADD_VECTOR_ELT_HI ` are dead and will be DCE'd.
cc @abhinay-anubola @gbossu

llvm/test/CodeGen/AIE/aie2/GlobalISel/legalize-build-vector.mir
```diff
@@ -66,24 +66,24 @@ body:             |
...
+    ; CHECK-NEXT: [[AIE_ADD_VECTOR_ELT_HI10:%[0-9]+]]:_(<16 x s32>) = G_AIE_ADD_VECTOR_ELT_HI [[AIE_ADD_VECTOR_ELT_HI9]], [[C3]](s32)
+    ; CHECK-NEXT: [[AIE_ADD_VECTOR_ELT_HI11:%[0-9]+]]:_(<16 x s32>) = G_AIE_ADD_VECTOR_ELT_HI [[AIE_ADD_VECTOR_ELT_HI10]], [[C3]](s32)
+    ; CHECK-NEXT: [[AIE_ADD_VECTOR_ELT_HI12:%[0-9]+]]:_(<16 x s32>) = G_AIE_ADD_VECTOR_ELT_HI [[AIE_ADD_VECTOR_ELT_HI11]], [[C3]](s32)
+    ; CHECK-NEXT: [[AIE_ADD_VECTOR_ELT_HI13:%[0-9]+]]:_(<16 x s32>) = G_AIE_ADD_VECTOR_ELT_HI [[AIE_ADD_VECTOR_ELT_HI12]], [[C3]](s32)
+    ; CHECK-NEXT: [[AIE_ADD_VECTOR_ELT_HI14:%[0-9]+]]:_(<16 x s32>) = G_AIE_ADD_VECTOR_ELT_HI [[AIE_ADD_VECTOR_ELT_HI13]], [[C3]](s32)
+    ; CHECK-NEXT: [[AIE_ADD_VECTOR_ELT_HI15:%[0-9]+]]:_(<16 x s32>) = G_AIE_ADD_VECTOR_ELT_HI [[AIE_ADD_VECTOR_ELT_HI14]], [[C3]](s32)
     ; CHECK-NEXT: [[FRAME_INDEX:%[0-9]+]]:_(p0) = G_FRAME_INDEX %stack.0
-    ; CHECK-NEXT: G_STORE [[AIE_ADD_VECTOR_ELT_LEFT15]](<16 x s32>), [[FRAME_INDEX]](p0) :: (store (<16 x s32>), align 32)
+    ; CHECK-NEXT: G_STORE [[AIE_ADD_VECTOR_ELT_HI]](<16 x s32>), [[FRAME_INDEX]](p0) :: (store (<16 x s32>), align 32)
```
konstantinschwarz, line 86: Note: this is also wrong, we need to store the *last* update to the vector, i.e. `AIE_ADD_VECTOR_ELT_HI15`

PR #173, opened 2024-08-28
[AIE2] Alias Analysis using AddrSpace Info

krishnamtibrewala:
The work is guarded under an option `--aie-alias-analysis-addrspace`.

**Note : There are no QoR gain with this work!!**

PR #188, opened 2024-09-17
[AIE2] NFC : Refactor Address Space Code

krishnamtibrewala:
The PR would allow to use a single architecture agnostic AA pass which can do AA using addresspace info.  
Kindly refer #173 to see the use case. 

Note : AA using addresspace info is guarded by a CMD line option.

### DIFF
--- a/llvm/lib/Target/AIE/AIE2AddrSpace.h
+++ b/llvm/lib/Target/AIE/AIE2AddrSpace.h
@@ -4,13 +4,16 @@
 // See https://llvm.org/LICENSE.txt for license information.
 // SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
 //
-// (c) Copyright 2023-2024 Advanced Micro Devices, Inc. or its affiliates
+// (c) Copyright 2024 Advanced Micro Devices, Inc. or its affiliates
 //
 //===----------------------------------------------------------------------===//
 //
 // This file declares the AIEngine V2 Address Space and DM banks
 //
 //===----------------------------------------------------------------------===//
+
+#include "AIEBaseAddrSpaceInfo.h"
+#include <bitset>
 
 #ifndef LLVM_SUPPORT_AIE2ADDRSPACE_H
 #define LLVM_SUPPORT_AIE2ADDRSPACE_H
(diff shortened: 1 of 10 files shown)


llvm/lib/Target/AIE/AIE2AddrSpace.h
```diff
@@ -41,6 +44,73 @@ enum class AddressSpaces {
 enum class AIEBanks { A, B, C, D, TileMemory };
 
 } // end namespace AIE2
+
+class AIE2AddrSpaceInfo : public AIEBaseAddrSpaceInfo {
```
andcarminati: In this a final class?

andcarminati:
This refactoring looks nice!

llvm/lib/Target/AIE/AIE2AddrSpace.h
krishnamtibrewala: yes, have added the "final" keyword there.

gbossu:
Can you add a rationale in the PR description as to why this change is needed?

llvm/lib/Target/AIE/AIEBaseAddrSpaceInfo.h
```diff
@@ -0,0 +1,44 @@
...
+    // By default assume conflicts.
+    return ~0;
+  }
+  virtual MemoryBankBits
+  getMemoryBanksFromAddressSpace(unsigned AddrSpace) const {
+    // By default assume conflicts.
+    return ~0;
+  }
+
+  virtual bool addrspacesMayAlias(unsigned AS1, unsigned AS2) {
```
gbossu: I guess you will use that for `bool TargetTransformInfo::addrspacesMayAlias(unsigned AS0, unsigned AS1) const;`? How do you intend to get the right `AIEAddrSpaceInfo` for the current target?

krishnamtibrewala: This is how I plan to get the AddrSpaceInfo for the correct target 

```
bool AIE2TTIImpl::addrspacesMayAlias(unsigned AS0, unsigned AS1) const {
  const AIEBaseAddrSpaceInfo &ASI = ST->getAddrSpaceInfo();
  return ASI.getMemoryBanksFromAddressSpace(AS0) &&
         ASI.getMemoryBanksFromAddressSpace(AS1);
}
```

`getAddrSpaceInfo` is defined in `llvm/lib/Target/AIE/AIE2Subtarget.cpp`

gbossu: Then could we remove `AIEBaseAddrSpaceInfo::addrspacesMayAlias`? Or do you plan to use it elsewhere? It seems to do the same as `AIE2TTIImpl::addrspacesMayAlias`.

Also, could `AIE2TTIImpl::addrspacesMayAlias` actually be `AIEBaseTTIImple::addrspacesMayAlias`? (if that even exists)

krishnamtibrewala: 1. Have removed `AIEBaseAddrSpaceInfo::addrspacesMayAlias`
2. No since we do not have `AIEBaseTTIImpl`, also based on discussion with Andreu the idea was scrapped https://github.com/Xilinx/llvm-aie/pull/173#discussion_r1763406439

llvm/lib/Target/AIE/AIE2Subtarget.cpp
```diff
@@ -46,6 +46,7 @@ AIE2Subtarget::AIE2Subtarget(const Triple &TT, StringRef CPU, StringRef TuneCPU,
                              StringRef FS, StringRef ABIName,
                              const TargetMachine &TM)
     : AIE2GenSubtargetInfo(TT, CPU, TuneCPU, FS), AIEBaseSubtarget(TT),
+      AddrSpaceInfo(),
```
gbossu: I don't think we need this line, `AddrSpaceInfo` will already be default constructed.

llvm/lib/Target/AIE/AIESubtarget.cpp
```diff
@@ -45,6 +45,7 @@ AIESubtarget::AIESubtarget(const Triple &TT, StringRef CPU, StringRef TuneCPU,
                            StringRef FS, StringRef ABIName,
                            const TargetMachine &TM)
     : AIEGenSubtargetInfo(TT, CPU, TuneCPU, FS), AIEBaseSubtarget(TT),
+      AddrSpaceInfo(),
```
gbossu: Same, I don't think we need an explicit call

gbossu (approved):
LGTM

PR #191, opened 2024-09-20
[AIE2] Skip copies and bitcasts when combining loads and stores

andcarminati:
This helps to improve combination of loads and stores with other vector operations.

### DIFF
--- a/llvm/lib/Target/AIE/AIECombinerHelper.cpp
+++ b/llvm/lib/Target/AIE/AIECombinerHelper.cpp
@@ -76,6 +76,32 @@ bool isUseOf(const MachineInstr &MI, const MachineInstr &Use) {
   return false;
 }
 
+///  Check for dead \a InBetweenMI MI and copy-like instructions that can be
+///  coalesced once \a MemI and \a Dest are combined.
+bool isNonCoalesceableUseOf(const MachineInstr &MemI,
+                            const MachineInstr &InBetweenMI,
+                            const MachineInstr &Dest,
+                            const MachineRegisterInfo &MRI) {
+
+  if (isTriviallyDead(InBetweenMI, MRI))
+    return false;
+
+  // We can delay an instruction after a copy, if the copy just
+  // connects MemI and Dest. After combining, this copy will be dead.
+  if (InBetweenMI.isCopy() &&
+      MRI.hasOneNonDBGUse(InBetweenMI.getOperand(1).getReg()) &&
+      MRI.hasOneNonDBGUse(InBetweenMI.getOperand(0).getReg())) {
+    const MachineInstr *CopyOrignMI =
+        MRI.getVRegDef(InBetweenMI.getOperand(1).getReg());
+    const MachineInstr *CopyDestMI =
+        &*MRI.use_instr_nodbg_begin(InBetweenMI.getOperand(0).getReg());
+    if (CopyOrignMI == &MemI && CopyDestMI == &Dest)
+      return false;
+  }
+
+  return isUseOf(InBetweenMI, MemI);
+}
+
 /// \return true if \a MemI can be moved just before \a Dest in order to allow
 /// post-increment combining
 bool llvm::canDelayMemOp(MachineInstr &MemI, MachineInstr &Dest,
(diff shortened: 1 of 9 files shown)


gbossu (dismissed):
LGTM, and I think this is very needed given how our intrinsics are defined. Do you remember if that had an impact on QoR?

andcarminati:
> LGTM, and I think this is very needed given how our intrinsics are defined. Do you remember if that had an impact on QoR?

Hi @gbossu, as is, there is no impact on QoR! However, it will impact on the upcoming changes. I will update this PR for the loads as well, because the following pattern will appear when we split loads before selection:

```
    %190:vregbank(<16 x s16>), %37:ptrregbank(p0) = G_AIE_POSTINC_LOAD %32, %184(s20) :: (load (<16 x s16>) from %ir.p_in.0.i, !tbaa !4, !alias.scope !1658, !noalias !1661, addrspace 5)
    %38:vregbank(<16 x s16>) = COPY %190(<16 x s16>)
    %40:accregbank(<8 x s64>) = G_INTRINSIC intrinsic(@llvm.aie2.v16bf16.to.v16accfloat), %38(<16 x s16>)
```

In this case, we cannot combine because we cannot move the load after the copy in the current implementation. I will handle this case as well.

llvm/lib/Target/AIE/AIECombinerHelper.cpp
```diff
@@ -87,12 +87,43 @@ bool llvm::canDelayMemOp(MachineInstr &MemI, MachineInstr &Dest,
...
+MachineInstr *
+llvm::getDefIgnoringCopiesAndBitcasts(Register Reg,
+                                      const MachineRegisterInfo &MRI) {
+
+  MachineInstr *DefInstr = getDefIgnoringCopies(Reg, MRI);
+
+  while (
+      DefInstr && DefInstr->getOpcode() == TargetOpcode::G_BITCAST &&
+      MRI.hasOneNonDBGUse(
+          DefInstr->getOperand(0).getReg()) /*No other use for this bitcast*/) {
```
gbossu: Super-nit: I would take the `/*No other use for this bitcast*/` comment out of the loop condition to make it easier to read.

llvm/lib/Target/AIE/AIECombinerHelper.cpp
```diff
@@ -87,12 +87,43 @@ bool llvm::canDelayMemOp(MachineInstr &MemI, MachineInstr &Dest,
   auto InstrRange = make_range(MII, MIE);
   bool SawStore = MemI.mayStore();
   auto UnsafeToMovePast = [&](const MachineInstr &MI) {
-    return (isUseOf(MI, MemI) && !isTriviallyDead(MI, MRI)) ||
+    bool IsCopyBetween = false;
+    // We can delay an instruction after a copy, if the copy just
+    // connects MemI and Dest. After combining, this copy will be dead.
+    if (MI.isCopy() && MRI.hasOneNonDBGUse(MI.getOperand(1).getReg()) &&
```
gbossu: Should we also look at bitcasts here as well?

llvm/lib/Target/AIE/AIECombinerHelper.cpp
```diff
@@ -87,12 +87,43 @@ bool llvm::canDelayMemOp(MachineInstr &MemI, MachineInstr &Dest,
...
+      MachineInstr *CopyOrignMI = MRI.getVRegDef(MI.getOperand(1).getReg());
+      MachineInstr *CopyDestMI =
+          &*MRI.use_instr_nodbg_begin(MI.getOperand(0).getReg());
+      if (CopyOrignMI == &MemI && CopyDestMI == &Dest) {
+        IsCopyBetween = true;
+      }
+    }
+    return ((isUseOf(MI, MemI) && !IsCopyBetween) &&
+            !isTriviallyDead(MI, MRI)) ||
            !MI.isSafeToMove(nullptr, SawStore);
```
gbossu, line 117: That condition is getting really complex. How about something like:
```
return isNonCoalesceableUseOf(MI, MemI) || !MI.isSafeToMove(nullptr, SawStore)
```

In `isNonCoalesceableUseOf` we could check for dead MI and copy-like instructions that can be coalesced once MemI and Dst are combined.

llvm/lib/Target/AIE/AIECombinerHelper.cpp
andcarminati: Hi @gbossu, basically I saw copys for loads and bitcasts+copy for stores, and we are using this function to move loads down. I think we can keep bitcast out for now and handle it just in case it is really necessary.

llvm/lib/Target/AIE/AIECombinerHelper.cpp
```diff
@@ -76,6 +76,34 @@ bool isUseOf(const MachineInstr &MI, const MachineInstr &Use) {
...
+
+  // We can delay an instruction after a copy, if the copy just
+  // connects MemI and Dest. After combining, this copy will be dead.
+  if (InBetweenMI.isCopy() &&
+      MRI.hasOneNonDBGUse(InBetweenMI.getOperand(1).getReg()) &&
+      MRI.hasOneNonDBGUse(InBetweenMI.getOperand(0).getReg())) {
+    const MachineInstr *CopyOrignMI =
+        MRI.getVRegDef(InBetweenMI.getOperand(1).getReg());
+    const MachineInstr *CopyDestMI =
+        &*MRI.use_instr_nodbg_begin(InBetweenMI.getOperand(0).getReg());
```
gbossu, line 97: We verify that the bitcast has a single use, but not the intermediate copies. Maybe we should just handle `TargetOpcode::COPY` specifically as well?

andcarminati: It makes sense. I update the PR and included a test for copy with no single use.

llvm/lib/Target/AIE/AIECombinerHelper.cpp
```diff
@@ -76,6 +76,34 @@ bool isUseOf(const MachineInstr &MI, const MachineInstr &Use) {
...
+    if (CopyOrignMI == &MemI && CopyDestMI == &Dest) {
+      IsCopyBetween = true;
```
gbossu: Just return `false` here? Then the final return below just becomes an easy `return isUseOf(InBetweenMI, MemI);`

gbossu (approved):
LGTM

PR #193, opened 2024-09-23
[AIEX] NFC: Refactor Alternate Descriptor codebase

krishnamtibrewala:


llvm/lib/Target/AIE/AIEAlternateDescriptor.h
```diff
@@ -0,0 +1,63 @@
...
+#define LLVM_SUPPORT_AIEALTERNATEDESCRIPTOR_H
+
+#include "AIEBaseSubtarget.h"
+#include "MCTargetDesc/AIEMCFormats.h"
+
+#include <unordered_map>
+
+namespace llvm {
+
+typedef std::unordered_map<MachineInstr *, const MCInstrDesc *> MIAltDescsMap;
```
andcarminati: It is better to use:

`using MIAltDescsMap = std::unordered_map<MachineInstr *, const MCInstrDesc *>;`

llvm/lib/Target/AIE/AIEHazardRecognizer.cpp
```diff
@@ -385,7 +386,8 @@ void AIEHazardRecognizer::EmitInstruction(SUnit *SU, int DeltaCycles) {
 
   // If the instruction has multiple options, find the opcode that was selected
   // and use the latter to update the scoreboard.
-  unsigned SelectedOpcode = getSelectedAltOpcode(MI).value_or(MI->getOpcode());
+  unsigned SelectedOpcode =
+      SelectedAltDescs.getAlternateOpcode(MI).value_or(MI->getOpcode());
```
andcarminati: Can we hide this `value_or` inside the new class? I  think we should just call `getAlternateOpcode` and receive an opcode.

krishnamtibrewala: I did that but found a case where, based on std::nullopt we decide and IF cond.
I will try to factor that code to make what you shared possible.

llvm/lib/Target/AIE/AIEHazardRecognizer.cpp
```diff
@@ -324,7 +324,8 @@ AIEHazardRecognizer::getHazardType(SUnit *SU, int DeltaCycles) {
       // Check if there is NoHazard, If there is a Hazard or NoopHazard check
       // for the next possible Opcode.
       if (Haz == NoHazard) {
-        SelectedAltOpcodes[MI] = AltInstOpcode;
+        SelectedAltDescs.setAlternateDescriptor(
+            MI, const_cast<llvm::MCInstrDesc *>(&TII->get(AltInstOpcode)));
```
andcarminati: I think we can have a clean call here, like:

```
SelectedAltDescs.setAlternateDescriptor(MI, AltInstOpcode);
```

In this way, TII can be a field in `AIEAlternateDescriptor`.

We can also remove the cast.

krishnamtibrewala: Have updates as suggested. Thank you!

llvm/lib/Target/AIE/AIEAlternateDescriptor.h
```diff
@@ -0,0 +1,74 @@
...
+#include "AIEBaseSubtarget.h"
+#include "MCTargetDesc/AIEMCFormats.h"
+
+#include <unordered_map>
+
+namespace llvm {
+
+using MIAltDescsMap = std::unordered_map<MachineInstr *, const MCInstrDesc *>;
+
+class AIEAlternateDescriptor {
```
gbossu: I think the name makes it sound like this data structure stores a single descriptor. Maybe rename to `AIEAlternateDescriptors` or `AIEAlternateDescriptorMap`?

llvm/lib/Target/AIE/AIEAlternateDescriptor.h
```diff
@@ -0,0 +1,74 @@
...
+      return It->second;
+    return std::nullopt;
+  }
+
+  const MCInstrDesc *getDesc(MachineInstr *MI) {
+    return getAlternateDescriptor(MI).value_or(&MI->getDesc());
+  }
+
+  // Return the alternate opcode for the given multi-opcode instruction.
+  std::optional<unsigned> getAlternateOpcode(MachineInstr *MI) const {
```
gbossu: I would prefer a name like `getSelectedOpcode` or `getAssignedOpcode`, because it's not just any alternate opcode, it's the one the scheduler selected.

gbossu: Same for `getAlternateDescriptor`

gbossu: (Or to be fully explicit, maybe even `getSelectedAlternateXXXX`, up to you 😄)

gbossu:
I'm starting to wonder how far we should go with adding more layers in the schedule. I'd say we need to go even further. Our interfaces to the HazardRecognizer are becoming painful to work with. I'd like group together what is really needed to make a correct scheduling decision. I'd say at this point we need something like:

```
struct DynamicMIDescriptor {
  ArrayRef<InstrStage> ItinStages;
  SlotBits SlotSet,
  MemoryBankBits MemoryBanks,
  SmallVector<int, 2> MemoryAccessCycles
};
```

Then we could just pass that dynamic descriptor to the checkHazard/enterResources methods of the HazardRecognizer. Within the scheduler, we could easily retrieve that dynamic descriptor by doing something like `AlternateDescrMap->getDynamicDescr(MI)`, and that would find the right itinerary, slotset, etc. all based on the selected opcode (or the standard one).

 I don't think we should do that change just now, and I can also work on this, but I'd like to start the discussion. What do you think @martien-de-jong @krishnamtibrewala @andcarminati ?

llvm/lib/Target/AIE/AIEHazardRecognizer.h
```diff
@@ -107,6 +108,8 @@ class AIEHazardRecognizer : public ScheduleHazardRecognizer {
 
   ~AIEHazardRecognizer() override {}
 
+  AIEAlternateDescriptors SelectedAltDescs;
```
gbossu: Do we want this to be public? Why not just replace the old private attribute `std::map<MachineInstr *, unsigned> SelectedAltOpcodes;` with `AIEAlternateDescriptors SelectedAltDescs`?

martien-de-jong:
> I'm starting to wonder how far we should go with adding more layers in the schedule. I'd say we need to go even further. Our interfaces to the HazardRecognizer are becoming painful to work with. I'd like group together what is really needed to make a correct scheduling decision. I'd say at this point we need something like:
> 
> ```
> struct DynamicMIDescriptor {
>   ArrayRef<InstrStage> ItinStages;
>   SlotBits SlotSet,
>   MemoryBankBits MemoryBanks,
>   SmallVector<int, 2> MemoryAccessCycles
> };
> ```
> 
> Then we could just pass that dynamic descriptor to the checkHazard/enterResources methods of the HazardRecognizer. Within the scheduler, we could easily retrieve that dynamic descriptor by doing something like `AlternateDescrMap->getDynamicDescr(MI)`, and that would find the right itinerary, slotset, etc. all based on the selected opcode (or the standard one).
> 
> I don't think we should do that change just now, and I can also work on this, but I'd like to start the discussion. What do you think @martien-de-jong @krishnamtibrewala @andcarminati ?

That looks good. I would then also like a constructor for this that takes a MachineInstr. I can also imagine that we grab the static part of this and let tablegen generate it.

gbossu:
> That looks good, except that all of this is static, not dynamic ;-)

Fair enough 😆 I thought of dynamic as in "the scheduling descriptor is determined based on the selected opcode, and operands". Those things can change throughout the compilation process. Maybe you can think of a better qualifier than "dynamic"? Maybe "FullyQualifiedDescr"?

martien-de-jong:
> > That looks good, except that all of this is static, not dynamic ;-)
> 
> Fair enough 😆 I thought of dynamic as in "the scheduling descriptor is determined based on the selected opcode, and operands". Those things can change throughout the compilation process. Maybe you can think of a better qualifier than "dynamic"? Maybe "FullyQualifiedDescr"?

I guess that MemoryBanks is dynamic

llvm/lib/Target/AIE/AIEHazardRecognizer.h
krishnamtibrewala: It is public because of the way I am using it in AIEMachineScheduler.cpp

gbossu: Why not just keep the existing `getSelectedAltOpcode()` method, and just change the implementation underneath? I wouldn't mind adding `AIEAlternateDescriptors SelectedAltDescs` as a public member to BlockState/RegionState, but I feel it's out of place within the HazardRecognizer.

krishnamtibrewala: I would like to abstract getting Opcode or Descriptor for multiSlot or sigleSlot all under a same object.
In the follow-up work I fund have an object of `AIEAlternateDescriptors` the best way. (in #182 )

gbossu: I agree it's better to encapsulate that in a new object, I just think we shouldn't refactor and stop halfway. I would be fine with either:
 - keeping the existing `getSelectedAltOpcode()` method, and just change the implementation underneath
 - or make `AIEAlternateDescriptors SelectedAltDescs` a member of BlockState/RegionState, because this is really state that is updated by the scheduler, and if we want interblock to know about that state, it cannot stay in the HazardRecognizer.

krishnamtibrewala: Agreed, I prefer going with the second approach of making AIEAlternateDescriptors  a member of  BlockState/RegionState. To a certain extend that is done as part of #182 

I need to think through if we can completely get it rid from HazardRecognizer.

krishnamtibrewala: I have made `AIEAlternateDescriptors SelectedAltDescs` a private member of the HazardRecognizer class, and provided a member function to get a const pointer to same.

We cannot remove  `AIEAlternateDescriptors SelectedAltDescs` from hazardRecognizer without affecting the functionality of multi-slot instruction.

gbossu: Discussed offline: we'll make `AIEAlternateDescriptors` part of the `BlockState` and pass a reference to it when constructing the `HazardRecognizer`

llvm/lib/Target/AIE/AIEHazardRecognizer.h
```diff
@@ -203,9 +205,9 @@ class AIEHazardRecognizer : public ScheduleHazardRecognizer {
 
 private:
   ResourceScoreboard<FuncUnitWrapper> Scoreboard;
-  std::map<MachineInstr *, unsigned> SelectedAltOpcodes;
   const AIEBaseInstrInfo *TII;
   const InstrItineraryData *ItinData;
+  AIEAlternateDescriptors *SelectedAltDescs;
```
gbossu: I think we always expect this to be non-null. Could you make that a reference? Same for `AIEAlternateDescriptors *getSelectedAltDescs()`

llvm/lib/Target/AIE/AIEHazardRecognizerPRAS.cpp
```diff
@@ -24,7 +24,8 @@ using namespace llvm;
 
 AIEHazardRecognizerPRAS::AIEHazardRecognizerPRAS(const AIEBaseInstrInfo *TII,
                                                  const InstrItineraryData *II)
-    : AIEHazardRecognizer(TII, II, /*IsPreRA=*/false),
+    : AIEHazardRecognizer(TII, II, new AIEAlternateDescriptors(),
```
gbossu: Will that ever be freed?

llvm/lib/Target/AIE/AIEInterBlockScheduling.h
```diff
@@ -318,6 +319,8 @@ class InterBlockScheduling {
   /// This is typical for epilogue blocks, which supply the safety margin of
   /// their loops.
   int getNumEntryNops(const BlockState &BS) const;
+
+  AIEAlternateDescriptors *getSelectedAltDescs() { return &SelectedAltDesc; }
```
gbossu: If we know it's not null, please return a reference

llvm/lib/Target/AIE/AIEMachineScheduler.h
```diff
@@ -190,6 +196,8 @@ class AIEPreRASchedStrategy : public GenericScheduler {
   std::vector<unsigned> SUDelayerMap;
 
   std::vector<unsigned> PSetThresholds;
+
+  AIEAlternateDescriptors SelectedAltDescs;
```
gbossu, line 202: We should probably clear that on `enterRegion`, otherwise the map will grow un-necessarily.

llvm/lib/Target/AIE/AIEHazardRecognizerPRAS.cpp
krishnamtibrewala: Will free in the destructor of  `AIEHazardRecognizerPRAS`

llvm/lib/Target/AIE/AIEMachineScheduler.h
krishnamtibrewala: I have done this in `leaveRegion`

llvm/lib/Target/AIE/AIEHazardRecognizerPRAS.cpp
```diff
@@ -24,7 +24,8 @@ using namespace llvm;
 
 AIEHazardRecognizerPRAS::AIEHazardRecognizerPRAS(const AIEBaseInstrInfo *TII,
                                                  const InstrItineraryData *II)
-    : AIEHazardRecognizer(TII, II, /*IsPreRA=*/false),
+    : AIEHazardRecognizer(TII, II, *(new AIEAlternateDescriptors()),
+                          /*IsPreRA=*/false),
```
gbossu, line 28: Maybe we could just add a `AIEAlternateDescriptors PRASAlternateDescriptors` attribute in `AIEHazardRecognizerPRAS`? Dynamic allocation is more error-prone than anything.

andcarminati:
> I'm starting to wonder how far we should go with adding more layers in the schedule. I'd say we need to go even further. Our interfaces to the HazardRecognizer are becoming painful to work with. I'd like group together what is really needed to make a correct scheduling decision. I'd say at this point we need something like:
> 
> ```
> struct DynamicMIDescriptor {
>   ArrayRef<InstrStage> ItinStages;
>   SlotBits SlotSet,
>   MemoryBankBits MemoryBanks,
>   SmallVector<int, 2> MemoryAccessCycles
> };
> ```
> 
> Then we could just pass that dynamic descriptor to the checkHazard/enterResources methods of the HazardRecognizer. Within the scheduler, we could easily retrieve that dynamic descriptor by doing something like `AlternateDescrMap->getDynamicDescr(MI)`, and that would find the right itinerary, slotset, etc. all based on the selected opcode (or the standard one).
> 
> I don't think we should do that change just now, and I can also work on this, but I'd like to start the discussion. What do you think @martien-de-jong @krishnamtibrewala @andcarminati ?

I agree that HR is handling several different interfaces now and this redesign will help to concentrate all the information that is related to one specific instruction. We can put this on our backlog for sure.

llvm/unittests/Target/AIE/AIEScheduleDAGMITest.cpp
```diff
@@ -29,9 +29,10 @@ namespace {
 class DummyAIEHazardRecognizer : public AIEHazardRecognizer {
 public:
   DummyAIEHazardRecognizer(const ScheduleDAG *DAG)
-      : AIEHazardRecognizer(nullptr, nullptr, DAG, /*ScoreboardDepth=*/0) {}
+      : AIEHazardRecognizer(nullptr, nullptr, *(new AIEAlternateDescriptors()),
+                            DAG, /*ScoreboardDepth=*/0) {}
 
-  ~DummyAIEHazardRecognizer() override {}
+  ~DummyAIEHazardRecognizer() override { delete &getSelectedAltDescs(); }
```
gbossu: Can you do the same as in PRASAieHazardRecognizer to avoid dynamic allocation?

krishnamtibrewala: Thank you, have updated accordingly!

gbossu (approved):
LGTM

PR #194, opened 2024-09-23
[AIE] Phi nodes need not be dominated

martien-de-jong:
The original safety test for a PTRADD LD/ST postinc combine was overly conservative.

### DIFF
--- a/llvm/lib/Target/AIE/AIECombinerHelper.cpp
+++ b/llvm/lib/Target/AIE/AIECombinerHelper.cpp
@@ -212,6 +212,20 @@ findConstantOffsetsToMove(MachineInstr &PtrAdd, MachineInstr &PtrAddInsertLoc,
   return GConstsToMove;
 }
 
+// Check that MI is after First and not after Last
+bool isBetween(MachineInstr &MI, MachineInstr &First, MachineInstr &Last,
+               CombinerHelper &Helper) {
+  assert(First.getParent() == Last.getParent());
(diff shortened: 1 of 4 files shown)


andcarminati (dismissed):
LGTM.

llvm/lib/Target/AIE/AIECombinerHelper.cpp
```diff
@@ -239,12 +239,19 @@ MachineInstr *findPostIncMatch(MachineInstr &MemI, MachineRegisterInfo &MRI,
...
-            return &MemI != &PtrAddUse && Helper.dominates(MemI, PtrAddUse);
-          });
+      bool MemOpDominatesAddrUses =
+          all_of(MRI.use_nodbg_instructions(AddrUse.getOperand(0).getReg()),
+                 [&](MachineInstr &PtrAddUse) {
+                   return &MemI != &PtrAddUse &&
+                          (PtrAddUse.getParent() != AddrUse.getParent() ||
+                           Helper.dominates(MemI, PtrAddUse) ||
+                           Helper.dominates(PtrAddUse, AddrUse));
+                 });
```
gbossu: Discussed offline, ultimately, what we want to check is that there is no use of the address in between the two instructions, because those will get combined in a single one. Maybe define a `IsBetweenOps` lambda and simplify the long comment above?

gbossu (approved):

PR #197, opened 2024-09-24
[AIE2] Enhance VCONV_FP32_BF16 instr itinerary

krishnamtibrewala:


llvm/test/CodeGen/AIE/aie2/schedule/negative_latencies/bypass.mir
```diff
@@ -97,6 +97,36 @@ body:             |
...
+    $bmh0 = VCONV_FP32_BF16 $wl0
+    $x0 = VBAND $x2, $x4
+    VST_dmw_sts_w_ag_idx_imm $wh0, $p0, 0
+...
+# CHECK-SU-LABEL: antidep_minus_1_plus_wl_bypass_VCONV_FP32_BF16:%bb.0 entry
+# CHECK-SU: SU(0):   $bmh0 = VCONV_FP32_BF16 $wl0
+# CHECK-SU:  Successors:
+# CHECK-SU:    SU(1): Anti Latency=0{{$}}
+# CHECK-SU: SU(1):   $x0 = VBAND $x2, $x4
+---
```
gbossu: Can you remove the extra `---`? I think that might confuse some MIR parsers

llvm/lib/Target/AIE/AIE2GenFixupInstrInfo.td
```diff
@@ -958,7 +958,7 @@ let Defs = [srUPS_of] in {
 }
 
 // 7.5 vconv.fp32.bf16 – bfloat16 to float Conversion instructions
-let Itinerary = II_VCONVfp32bf16 in {
+let Itinerary = II_VCONVfp32bf16, ItineraryRegPairs = [ItinRegClassPair<II_VCONVfp32bf16_WL,[OperandRegClass<1, eWL>]>] in {
```
gbossu, line 961: Just to be 100% sure, could you also add tests in `llvm/test/CodeGen/AIE/aie2/schedule/mov_bypass.mir` to show when the bypass is taken, and not taken for VCONV when the source comes from `VMOV_mv_w`?

We already have `bypass_wl` and `bypass_wh` tests, but I'd like to add two more where the first VMOV is changed into a `VCONV_FP32_BF16` for both cases. Bypasses are hard, so I'd like to be completely sure that our itinerary match the expected behaviour :)

gbossu (approved):
LGTM

PR #203, opened 2024-10-02
[AIEX] Fix `reserved-reg-licm` pass

mludevid:
This PR fixes an assertion being triggered in the `reserved-reg-licm` pass.

The pass was not correctly checking for reserved registers and in some call paths it tries moving instructions involving physical registers that are not reserved. I was able to trigger this bug by calling a function that takes a 512-bit register as an input from inside a loop. The ABI introduces a copy of the 512-bit register into a physical register before the jump and link and the pass was trying to move that precise copy. The actual assert being triggered is a sanity check that makes sure that the physical register involved in the moved instruction only contains one register unit, which is the case for all reserved registers but not for the 512-bit vector register.
TLDR: a pre-condition was not being checked properly

### DIFF
--- a/llvm/test/CodeGen/AIE/aie2/GlobalISel/post-select-licm-reserved.mir
+++ b/llvm/test/CodeGen/AIE/aie2/GlobalISel/post-select-licm-reserved.mir
@@ -385,3 +385,44 @@ body:             |
     PseudoJNZ %0, %bb.2
     PseudoJ_jump_imm %bb.1
 ...
+
+# Verify that non-reserved physical registers aren't considered for LICM
+---
+name:            test_is_simplifiable_reserved_reg
+tracksRegLiveness: true
+body:             |
+  ; CHECK-LABEL: name: test_is_simplifiable_reserved_reg
+  ; CHECK: bb.0:
+  ; CHECK-NEXT:   successors: %bb.2(0x80000000)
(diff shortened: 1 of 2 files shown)


konstantinschwarz:
Could you describe what bug this fixes?
If possible, it would be great if you could pre-commit your test first that exposes the bug, and then update the CHECK-lines with the actual fix.

mludevid:
> Could you describe what bug this fixes? If possible, it would be great if you could pre-commit your test first that exposes the bug, and then update the CHECK-lines with the actual fix.

I updated the PR description to explain the bug in detail. I also updated the test with a simpler one. Since the bug was leading to an assert i cannot pre-commit the test and then update the CHECK-lines with the fix.

llvm/test/CodeGen/AIE/aie2/GlobalISel/post-select-licm-reserved.mir
```diff
@@ -385,3 +385,42 @@ body:             |
     PseudoJNZ %0, %bb.2
     PseudoJ_jump_imm %bb.1
 ...
+---
```
gbossu, line 390: Nit: Maybe add a small description, e.g. `Verify that non-reserved physical registers aren't considered for LICM`

gbossu (dismissed):

gbossu (approved):

PR #204, opened 2024-10-05
[AIEX] Reschedule multi-slot instruction for better packing/schedule

krishnamtibrewala:


llvm/lib/Target/AIE/AIEHazardRecognizer.cpp
```diff
@@ -464,11 +481,34 @@ bool AIEHazardRecognizer::checkConflict(
     int DeltaCycles, std::optional<int> FUDepthLimit) {
   assert(Scoreboard.isValidDelta(DeltaCycles));
 
+  if (checkFormatConflict(Scoreboard, DeltaCycles, SlotSet))
```
andcarminati, line 494: I was wondering if we could specify conflict types as bits in a bit sequence/unsigned, instead of boolean. I say it because we almost reimplement this method entirely in `onlyFormatHazard`. I was thinking in something like:

```
if (Conflicts::Format == checkConflict(...) ) {

}

```

In this way we can query any individual conflict without a custom implementation...

andcarminati: Or `checkConflict` can be a wrapper around this function that can give all conflicts as a bitmask....

gbossu: I think splitting into multiple `checkXXXXConclict` is great! From what I see in the code it seems to be enough to be able to query specific conflicts, and nobody really checks the bitset. Could we keep the original `checkConflict` as returning a `bool`, or am I missing something?

krishnamtibrewala: The specific use is done in following place https://github.com/Xilinx/llvm-aie/pull/204/files#diff-65f8659bb554df4fc3fa735e811e4be1770c9b019e060dea78a9d9537d5d1367:~:text=(HR.,%3A%3AFormat)))

Here we want to make sure the conflict region is only Format/Slot hazard.

llvm/lib/Target/AIE/AIEHazardRecognizer.h
```diff
@@ -99,6 +101,13 @@ class AIEHazardRecognizer : public ScheduleHazardRecognizer {
   void computeMaxLatency();
 
 public:
```
andcarminati, line 103: Enum class is a better option:

```
enum class ConflictType: std::uint32_t {
  NoConflict = 0x0,
  Format = 0x1,
  MemoryBank,
  FuncUnit
};
```

krishnamtibrewala:
A better version of the work is created PR #217.
Hence closing this PR

PR #206, opened 2024-10-07
Add missing intrinsics.

SagarMaheshwari99:


abnikant (approved):
I guess you need this to compile newer mllib (kernel) ?

SagarMaheshwari99:
> I guess you need this to compile newer mllib (kernel) ?

No, this is to bypass some new v16cbfloat16 overloads which started appearing in AIE API.

PR #207, opened 2024-10-08
[AIE2] refactor load and store instruction selection

martien-de-jong:
Move the do simple combine conditions to the front


gbossu:
I think it makes a lot of sense. Let's wait for the AIE2p-related refactorings to to come up, but I would like to integrate this change

PR #211, opened 2024-10-14
[AIE2] Fix pathological case where greedy reloads directly from a CR register

gbossu:
It turns out greedy does quite a lot of "operation folding" to get more compact spill code.

It will turn
```
%0:ercr = LDA_spill ...
$crSat = COPY %0
```
into 
```
$crSat = LDA_spill ...
```

This is problematic because we cannot directly reload into a `$crSat` register. I actually expected such folding to be illegal, because after all, `$crSat` is a reserved register. And usually, reserved registers just mean "don't you dare optimize me". But it's also true that this register is part of `eRCR` (which it needs to be in for some of our copy-propagation extensions to kick in), so I guess what greedy does is fair game.

I could change `const TargetRegisterClass *canFoldCopy(...)` to actually never fold into reserved registers, but that causes one X86 test to fail. At this point, I'd rather just constrain `eRCR` to `eR` before greedy, because it's anyway what we want: the classes are equivalent in terms of allocatable registers, and `eR` is easy to spill.

No QoR impact, which is expected, but you never know.


llvm/lib/Target/AIE/AIERegClassConstrainer.cpp
```diff
@@ -0,0 +1,135 @@
...
+
+private:
+  /// Find the largest subclass of \p RC that is suitable for register
+  /// allocation. That is, it does not contain reserved registers, as those
+  /// are harder to spill/reload.
+  const TargetRegisterClass &getSuitableRCForRA(const TargetRegisterClass &RC,
+                                                const MachineRegisterInfo &MRI);
+
+  /// Cache for \p getSuitableRCForRA
+  std::map<unsigned, const TargetRegisterClass *> SuitableRCForRA;
```
krishnamtibrewala, line 56: nit : we should use std::unordered_map because of its avg insert/search time complexity.

gbossu: I agree, in the general case it is faster. But I think for small maps it doesn't really have an impact, and it might sometimes even be faster because there's no need for calling a hashing function.
I'm also generally wary of un-ordered data structures and avoid them if possible because they can lead to indeterministic behaviour (not here though), or just complicate debugging. Again, this doesn't really have an impact here, so I would keep it if you're fine with it :)

llvm/lib/Target/AIE/AIERegClassConstrainer.cpp
```diff
@@ -0,0 +1,135 @@
...
+
+const TargetRegisterClass &
+getSuitableRCForRAImpl(const TargetRegisterClass &RC,
+                       const MachineRegisterInfo &MRI) {
+  const TargetRegisterInfo &TRI = *MRI.getTargetRegisterInfo();
+
+  auto ContainsReservedReg = [&MRI](const TargetRegisterClass &RC) {
+    return any_of(RC, [&MRI](MCRegister Reg) { return MRI.isReserved(Reg); });
+  };
+
```
andcarminati, line 102: If `const TargetRegisterClass &RC` does not contain reserved regs at all, do we need to iterate over all SubRegClass set?

andcarminati: I think that if the class simply is free of reserved registers, then the loop will find the same class.

llvm/lib/Target/AIE/AIERegClassConstrainer.cpp
```diff
@@ -0,0 +1,135 @@
...
+};
+
+bool AIERegClassConstrainer::runOnMachineFunction(MachineFunction &MF) {
+  LLVM_DEBUG(llvm::dbgs() << "Constraining RCs for RA: " << MF.getName()
+                          << "\n");
+  MachineRegisterInfo &MRI = MF.getRegInfo();
+  const TargetRegisterInfo &TRI = *MRI.getTargetRegisterInfo();
+
+  // Clear previous cache, reserved registers can change for each function
+  SuitableRCForRA.clear();
```
andcarminati: I can suggest that we don't need to clean this cache always (from the compile time perspective). You can encapsulate this map with a BitVector representing the current set or reserved regs in a standalone class/struct (you can access the reserved registers with `BitVector getReservedRegs`  from TRI).   

We can have, in the end something like:

`SuitableRCForRA.clearIfDiverges(TRI)` -> just clean if we have different sets here.

llvm/lib/Target/AIE/AIERegClassConstrainer.cpp
gbossu: Correct, I can add an early exit

gbossu: Done

llvm/lib/Target/AIE/AIERegClassConstrainer.cpp
gbossu: Done

andcarminati (approved):
LGTM.

PR #213, opened 2024-10-15
Martien.pre2post

martien-de-jong:
Some tweaking to get more leverage from the PostPipeliner:
- Pre-allocate resources of future operations to take more heed of possible conflicts
- Avoid unrolling for very high MinTripCount loops
- PrePipeliner doesn't increase II if MaxStages is exceeded, instead it stops searching and rejects the schedule in favour of pipelining
- Allow larger II's in PostPipeliner; compute ResMII to take away some compile time


llvm/lib/Target/AIE/AIEPostPipeliner.cpp
```diff
@@ -270,8 +270,15 @@ bool PostPipeliner::scheduleFirstIteration() {
...
-                        MI->getMF()->getRegInfo(), Cycle);
+    for (int N = 0; N < NCopies; N++) {
+      if (N > 0 && HR.checkConflict(Scoreboard, *MI, Cycle)) {
+        return false;
+      }
+
+      HR.emitInScoreboard(Scoreboard, MI->getDesc(), MemoryBanks,
+                          MI->operands(), MI->getMF()->getRegInfo(), Cycle);
+      Cycle += II;
+    }
```
gbossu, line 305: Noice!

llvm/lib/CodeGen/MachinePipeliner.cpp
```diff
@@ -141,7 +141,7 @@ static cl::opt<int> SwpForceII("pipeliner-force-ii",
 static cl::opt<int>
     SwpMaxStages("pipeliner-max-stages",
                  cl::desc("Maximum stages allowed in the generated scheduled."),
-                 cl::Hidden, cl::init(3));
+                 cl::Hidden, cl::init(-1));
```
gbossu, line 144: We might have trouble upstreaming that

llvm/lib/Target/AIE/AIEMachineScheduler.cpp
```diff
@@ -1226,6 +1227,21 @@ void llvm::AIEPostRASchedStrategy::buildGraph(ScheduleDAGMI &DAG, AAResults *AA,
...
+    int II = BS.FixPoint.II;
+    int IterA = SUa->NodeNum / II;
+    int IterB = SUb->NodeNum / II;
+    if (aliasAcrossVirtualUnrolls(SUa->getInstr(), SUb->getInstr(), IterA,
+                                  IterB) == AliasResult::NoAlias) {
+      return false;
+    }
+  }
+
+  return SUa->getInstr()->mayAlias(AAForDep, *SUb->getInstr(), UseTBAA);
```
gbossu: How about `return ScheduleDAGMI::mayAlias(SUa, SUb, UseTBAA)`

llvm/lib/CodeGen/MachinePipeliner.cpp
andcarminati: In the end, this option seems strange. This decision should always delegated to the target.

andcarminati: On the other hand, we can push this option to -1 from the frontend.

llvm/include/llvm/CodeGen/ScheduleDAGInstrs.h
```diff
@@ -272,6 +272,9 @@ namespace llvm {
     MachineFunction *CurrFunc;
     MachineFunction::iterator NextMBB;
 
+    // Supply specific memory disambiguation
```
andcarminati, line 275: I like this flexibility!

llvm/lib/Target/AIE/AIEBasePipelinerLoopInfo.cpp
```diff
@@ -763,11 +768,23 @@ void ZeroOverheadLoop::adjustTripCount(int TripCountAdjust) {
 bool ZeroOverheadLoop::preferPostPipeliner(SMSchedule &SMS) {
   // Zero overhead loops are candidates for PostPipeliner, which does a better
   // job on multi-stage live-ranges without spilling or moving.
+
+  // PostPipeliner can do nothing without tripcount > 1
+  if (MinTripCount <= 1) {
+    return false;
+  }
```
martien-de-jong, line 775: Hmm. It's a ZOL, Prepipeliner can't do anything either.

llvm/lib/Target/AIE/AIEInterBlockScheduling.cpp
```diff
@@ -55,7 +55,7 @@ static cl::opt<bool>
                        "iterative loop scheduling"));
 
 static cl::opt<int> PostPipelinerMaxII(
-    "aie-postpipeliner-maxii", cl::init(10),
+    "aie-postpipeliner-maxii", cl::init(40),
```
martien-de-jong, line 58: Probably too high. Maybe revisit when we hit noticable compile time.

llvm/test/CodeGen/AIE/aie2/schedule/swp/swp-zoloop-accept-reject.mir
```diff
@@ -29,18 +29,183 @@ tracksRegLiveness: true
...
-
-  ; CHECK-TC-3: PLI: Rejected schedule MaxStageCount=3 MinTripCount=3)
-  ; CHECK-TC-3: PLI: Accepting II=4
-
-  ; CHECK-TC-2: PLI: Rejected schedule MaxStageCount=3 MinTripCount=2)
-  ; CHECK-TC-2: PLI: Rejected schedule MaxStageCount=2 MinTripCount=2)
-  ; CHECK-TC-2: PLI: Accepting II=5
-
-  ; CHECK-TC-1: Not interesting MinTripCount (<=1)!
-  ; CHECK-TC-1-NEXT: Loop rejected
```
martien-de-jong: Questionable auto-update.

llvm/lib/CodeGen/MachinePipeliner.cpp
martien-de-jong: I guess this predates shouldUseSchedule. I agree with Andreu that it should be decided there. 
When we upstream, I'll repair machine-pipeliner users, e.g. by implementing it in the base class of PipelinerLoopInfo rather than in the pipeliner itself.

llvm/lib/Target/AIE/AIEMachineScheduler.cpp
martien-de-jong: > How about `return ScheduleDAGMI::mayAlias(SUa, SUb, UseTBAA)`

Yep, that would be better

llvm/lib/Target/AIE/AIEHazardRecognizer.cpp
```diff
@@ -634,11 +638,14 @@ AIEHazardRecognizer::getMemoryBanks(const MachineInstr *MI) const {
...
+  MemoryBankBits MemoryBankUsed = 0;
   for (auto &MMO : MI->memoperands()) {
+    unsigned AddrSpace = MMO->getAddrSpace();
     MemoryBankBits MemoryBank =
-        ASI.getMemoryBanksFromAddressSpace(MMO->getAddrSpace());
-    MemoryBankUsed &= MemoryBank;
+        AddrSpace                ? ASI.getMemoryBanksFromAddressSpace(AddrSpace)
+        : AddressSpaceNoneIsSafe ? 0
+                                 : ASI.getDefaultMemoryBank();
+    MemoryBankUsed |= MemoryBank;
```
martien-de-jong, line 648: Note that this is more than an added option. We start at none, and collect the memory banks from all MMO's. We used to start at 'all banks' and intersect. The latter is incompatible with multi-memory ops like VLDB.4x

llvm/lib/Target/AIE/AIE2TargetTransformInfo.cpp
```diff
@@ -54,6 +54,9 @@ static cl::opt<int>
 static cl::opt<unsigned>
     MaxUnrollCost("aie-unroll-max-cost", cl::Hidden, cl::init(200),
                   cl::desc("Maximum partial unroll cost for loops"));
+static cl::opt<unsigned> PreferSwpOverUnroll(
+    "aie-prefer-swp-over-unroll", cl::Hidden, cl::init(9),
+    cl::desc("Aim for pipelining if MinIterCount is at least this value."));
```
martien-de-jong, line 59: The default is inspired by the fact that all ReduceABC's regress for 8 and lower.

llvm/lib/Target/AIE/AIEInterBlockScheduling.cpp
```diff
@@ -592,9 +592,12 @@ SchedulingStage InterBlockScheduling::updateScheduling(BlockState &BS) {
 
   // The loop schedule has converged, so we could declare our work done.
   // But first try SWP
-  if (BS.getRegions().size() == 1 && BS.getPostSWP().canAccept(*BS.TheBlock)) {
-    BS.FixPoint.II = 1;
-    return BS.FixPoint.Stage = SchedulingStage::Pipelining;
+  if (BS.getRegions().size() == 1) {
+    auto &PostSWP = BS.getPostSWP();
+    if (PostSWP.canAccept(*BS.TheBlock)) {
+      BS.FixPoint.II = PostSWP.getResMII(*BS.TheBlock);
```
martien-de-jong, line 598: This saves a lot of work. II divides the linear schedule length to determine the number of copies in the DAG, and DAG construction is expensive and worse than linear in the number of nodes. Something like Conv2D immediately hits the target II, so we would also save 7 iterations of the main loop.

llvm/lib/Target/AIE/AIEPostPipeliner.cpp
```diff
@@ -188,15 +188,35 @@ void PostPipeliner::computeLoopCarriedParameters() {
...
+        int P = Dep.getSUnit()->NodeNum;
+        int NewLatest = Latest - Dep.getSignedLatency();
+        if (NewLatest < Info[P].Latest) {
+          Info[P].Latest = NewLatest;
+          Changed = true;
+        }
+      }
+    }
+  }
+#if 0
```
martien-de-jong: I think this can go. It is the old version of Latest, and it isn't used anyway.

llvm/lib/Target/AIE/AIEPostPipeliner.cpp
gbossu: Could you think of maybe adding a small test that benefits from "pre-booking modulo resources"?

llvm/lib/Target/AIE/AIEBasePipelinerLoopInfo.cpp
```diff
@@ -52,9 +52,14 @@ cl::opt<bool> LoopWholeLoopGuard(
...
 cl::opt<int> PostPipelinerCandidateLimit(
     "aie-postpipeliner-limit",
-    cl::desc("II below which postpipeliner preference kicks in"), cl::init(2),
+    cl::desc("II below which postpipeliner preference kicks in"), cl::init(4),
     cl::Hidden);
 
+cl::opt<int> PostPipelinerCutoff(
+    "aie-postpipeliner-cutoff",
+    cl::desc("II below which high stage count is left to the postpipeliner"),
+    cl::init(11), cl::Hidden);
```
gbossu, line 61: Checking:
 - PostPipelinerCandidateLimit means we will leave the loop intact if a schedule with `II<=PostPipelinerCandidateLimit` was found. No questions asked
 - PostPipelinerCutoff means we will leave the loop intact if a schedule was found below that II, but only if the stage count is high. (I guess greater than `LoopMaxStageCount`?)

llvm/lib/Target/AIE/AIEBasePipelinerLoopInfo.cpp
gbossu: Why not? We can add guards I think

llvm/lib/Target/AIE/AIEInterBlockScheduling.cpp
gbossu: It's okay, if makes me feel better about `"aie-loop-aware-expensive-iterations", cl::init(35)` 😆

llvm/test/CodeGen/AIE/aie2/schedule/swp/swp-zoloop-accept-reject.mir
gbossu: Yes I think the `# NOTE: Assertions have been autogenerated by utils/update_mir_test_checks.py UTC_ARGS: --version 4` header should be removed. Could you maybe manually update the test?

llvm/lib/Target/AIE/AIEPostPipeliner.cpp
```diff
@@ -106,6 +106,27 @@ bool PostPipeliner::canAccept(MachineBasicBlock &LoopBlock) {
   return true;
 }
 
+int PostPipeliner::getResMII(MachineBasicBlock &LoopBlock) {
+  // For each instruction, find the first cycle in which it fits and collect the
+  // maximum
+  std::vector<uint64_t> Scoreboard(NInstr, 0);
+  int MII = 1;
+  for (auto &MI : LoopBlock) {
+    auto *SlotInfo = TII->getSlotInfo(TII->getSlotKind(MI.getOpcode()));
```
gbossu, line 115: Might want to start using `SelectedAltDescs` to prepare for the future?

llvm/lib/Target/AIE/AIEPostPipeliner.cpp
gbossu: The whole commit?

llvm/lib/Target/AIE/AIEBasePipelinerLoopInfo.cpp
martien-de-jong: There's no actual difference, but I would phrase the second one as 'when a high stagecount is found but only if the II is below cutoff.'

llvm/lib/Target/AIE/AIEPostPipeliner.cpp
martien-de-jong: Since this is just a lower bound, I think I want an optimistic value, not values established in a different schedule. And I think the future is to run the pipeliner first. I guess not counting unknown slots at all would be the proper thing to do.

llvm/lib/Target/AIE/AIEPostPipeliner.cpp
martien-de-jong: I'll see whether this tipped the conv2d balamce.

llvm/lib/CodeGen/MachinePipeliner.cpp
martien-de-jong: In fact I will do that in this PR.

llvm/lib/Target/AIE/AIEBasePipelinerLoopInfo.cpp
andcarminati: I feel that is hard to add guards in post, it will change the CFG and will require a new register for the constant...

llvm/lib/Target/AIE/AIEPostPipeliner.cpp
```diff
@@ -106,11 +106,35 @@ bool PostPipeliner::canAccept(MachineBasicBlock &LoopBlock) {
...
+    SlotBits Slots = SlotInfo ? SlotInfo->getSlotSet() : 0;
+
+    int C = 0;
+    while (C < NInstr && (Scoreboard[C] & Slots)) {
+      C++;
+    }
+    if (C >= NInstr) {
```
andcarminati, line 122: nit: `C == NInstr`. with C > NInstr means that we need k+1 slots for k instructions.

llvm/lib/Target/AIE/AIEPostPipeliner.cpp
```diff
@@ -106,11 +106,35 @@ bool PostPipeliner::canAccept(MachineBasicBlock &LoopBlock) {
...
+      MII = NInstr;
```
andcarminati, line 123: When this happens, do we try to run the pipeliner anyway? Should we return `PostPipelinerMaxII` (provided we share this option)?

llvm/lib/Target/AIE/AIEPostPipeliner.cpp
```diff
@@ -270,8 +300,15 @@ bool PostPipeliner::scheduleFirstIteration() {
     LLVM_DEBUG(dbgs() << "  Emit in " << -Depth + LocalCycle << "\n");
     int Cycle = -Depth + LocalCycle;
     LLVM_DEBUG(dbgs() << "  Emit in " << Cycle << "\n");
-    HR.emitInScoreboard(Scoreboard, MI->getDesc(), MemoryBanks, MI->operands(),
-                        MI->getMF()->getRegInfo(), Cycle);
+    for (int N = 0; N < NCopies; N++) {
```
andcarminati, line 297: nit: I would use unsigned here ;-).

llvm/lib/CodeGen/MachinePipeliner.cpp
martien-de-jong: Hmm. It's more complicated than that, because there's a MaxStages test in the main II loop. That means it could be moved to canAcceptII, but that's not upstreamed either. I propose to upstream both at the same time, moving the MaxStages check to the default implementation of canAcceptII

llvm/lib/Target/AIE/AIEBasePipelinerLoopInfo.cpp
martien-de-jong: Without a loop counter GPR?

llvm/lib/Target/AIE/AIEPostPipeliner.cpp
martien-de-jong: I'm sorry, I solemnly pledged only to use unsigned for bitmasks (and perhaps for very small types like char for lookup tables)

llvm/lib/Target/AIE/AIEPostPipeliner.cpp
martien-de-jong: nitter: This is the reverse of the first part of the loop condition, without having to know how C is updated.

llvm/lib/Target/AIE/AIEPostPipeliner.cpp
martien-de-jong: This computes ResMII, not how it it used. I don't want to tie too many strings to it.
As we compute it now, it will never exceed NInstr. but even if II == NInstr, the latency path through those NInstrs can allow a 4 stage pipeline to be useful. And if the DAG is a straight chain of dependences, pipelining can still be cheap.

llvm/lib/Target/AIE/AIEPostPipeliner.cpp
```diff
@@ -190,29 +190,35 @@ void PostPipeliner::computeLoopCarriedParameters() {
...
+    for (int K = NInstr - 1; K >= 0; K--) {
+      SUnit &SU = DAG->SUnits[K];
+      const int Latest = Info[K].Latest;
+      for (auto &Dep : SU.Preds) {
+        int P = Dep.getSUnit()->NodeNum;
+        int NewLatest = Latest - Dep.getSignedLatency();
+        if (NewLatest < Info[P].Latest) {
+          Info[P].Latest = NewLatest;
+          Changed = true;
+        }
```
gbossu: Can you give an overview of what this commit is doing?

llvm/test/CodeGen/AIE/aie2/schedule/postpipeliner/conv2d.mir
```diff
@@ -0,0 +1,784 @@
+# NOTE: Assertions have been autogenerated by utils/update_llc_test_checks.py UTC_ARGS: --version 4
```
gbossu, line 1: I expected that to be generated with update_mir_test_checks?

gbossu (approved):
LGTM 🎉

PR #171, opened 2024-08-27
[AIE2] Enable VEXTRACT & ADD_NC to use optimal itinerary

krishnamtibrewala:
Please review starting from : ad99c411b07ca44254468615de7c60803e78ba6a

PR #215, opened 2024-10-18
[AIE2] Return to conservative itinerary for ADD_NC/VEXTRACT

krishnamtibrewala:
The itinerary are inline with docs but are causing failures, which were missed.


gbossu:
I think we potentially need to revert the whole #171

PR #216, opened 2024-10-22
[AIE2] Fix itinerary for M slot

krishnamtibrewala:


llvm/lib/Target/AIE/AIE2Schedule.td
```diff
@@ -600,11 +600,11 @@ InstrItinData<II_LT, [InstrStage<1,  [R_WX_PORT]>], [1,1,1,1]>,
...
-InstrItinData<II_MOV_DC, [SimpleCycle<DC_WM_PORT>], [1,1]>,
-InstrItinData<II_MOV_DJ, [SimpleCycle<DJ_WM_PORT>], [1,1]>,
-InstrItinData<II_MOV_DN, [SimpleCycle<DN_WM_PORT>], [1,1]>,
-InstrItinData<II_MOV_M, [SimpleCycle<M_WM_PORT>], [1,1]>,
-InstrItinData<II_MOV_P, [SimpleCycle<P_WM_PORT>], [1,1]>,
+InstrItinData<II_MOV_DC, [PrefixCycle<RS_WM_PORT>, SimpleCycle<DC_WM_PORT>], [1,1]>,
+InstrItinData<II_MOV_DJ, [PrefixCycle<RS_WM_PORT>, SimpleCycle<DJ_WM_PORT>], [1,1]>,
+InstrItinData<II_MOV_DN, [PrefixCycle<RS_WM_PORT>, SimpleCycle<DN_WM_PORT>], [1,1]>,
+InstrItinData<II_MOV_M, [PrefixCycle<RS_WM_PORT>, SimpleCycle<M_WM_PORT>], [1,1]>,
+InstrItinData<II_MOV_P, [PrefixCycle<RS_WM_PORT>, SimpleCycle<P_WM_PORT>], [1,1]>,
```
martien-de-jong, line 607: Is there any use of the {DC,DJ,DN,M,P}_WM_PORT that doesn't also use the RS_WM_PORT?

gbossu: Also what I'm wondering, do we need all those itineraries then?

krishnamtibrewala: Yes we have use of {DC,DJ,DN,M,P}_WM_PORT from few instructions from slot A like LDA*, MOVA that does not use RS_WM_PORT

gbossu: Right, so it is okay to have
```
LDA p0
NOP
NOP
NOP
NOP
NOP
MOV m0
```

but not
```
VEXTRACT p0
MOV m0
```
because they both need to book `RS_WM_PORT` at the same time.

llvm/lib/Target/AIE/AIE2Schedule.td
```diff
@@ -790,15 +790,15 @@ InstrItinData<II_VEXTBCST, [EmptyCycles<1>, InstrStage<1,  [W_WM_PORT]>],
...
+InstrItinData<II_VEXTRACT_DJ, [EmptyCycles<1>, PrefixCycle<RS_WM_PORT>, SimpleCycle<DJ_WM_PORT>],
               [2,1,1,/*crVaddSign*/1], [NoBypass,MOV_Bypass,NoBypass]>,
-InstrItinData<II_VEXTRACT_DN, [EmptyCycles<1>, SimpleCycle<DN_WM_PORT>],
+InstrItinData<II_VEXTRACT_DN, [EmptyCycles<1>, PrefixCycle<RS_WM_PORT>, SimpleCycle<DN_WM_PORT>],
               [2,1,1,/*crVaddSign*/1], [NoBypass,MOV_Bypass,NoBypass]>,
-InstrItinData<II_VEXTRACT_M, [EmptyCycles<1>, SimpleCycle<M_WM_PORT>],
+InstrItinData<II_VEXTRACT_M, [EmptyCycles<1>, PrefixCycle<RS_WM_PORT>, SimpleCycle<M_WM_PORT>],
               [2,1,1,/*crVaddSign*/1], [NoBypass,MOV_Bypass,NoBypass]>,
-InstrItinData<II_VEXTRACT_P, [EmptyCycles<1>, SimpleCycle<P_WM_PORT>],
+InstrItinData<II_VEXTRACT_P, [EmptyCycles<1>, PrefixCycle<RS_WM_PORT>, SimpleCycle<P_WM_PORT>],
```
gbossu, line 801: I'm surprised such a change isn't requiring test updates. Can you add some to show that MOV and VEXTRACT can conflict, even though they seemingly write to different register files?

krishnamtibrewala: I did add few new tests, in r_wm.mir

gbossu (approved):

PR #222, opened 2024-10-24
[AIE] add StickyWAW mutator to postscheduler/postpipeliner

martien-de-jong:
Not checked on QoR yet, but I do see a significant improvement in one of the postpipeliner unit tests

### DIFF
--- a/llvm/lib/Target/AIE/AIEMachineScheduler.cpp
+++ b/llvm/lib/Target/AIE/AIEMachineScheduler.cpp
@@ -1212,8 +1212,9 @@ void llvm::AIEPostRASchedStrategy::buildGraph(ScheduleDAGMI &DAG, AAResults *AA,
     assert(BS.getRegions().size() == 1);
     // Try to wrap the linear schedule within II.
     // We virtually unroll the body by the stagecount, computed from rounding
-    // up the length divided by II.
-    NCopies = (BS.getScheduleLength() + II - 1) / II;
+    // up the length divided by II, adding one more stage to account for
+    // the added resource contention
+    NCopies = (BS.getScheduleLength() + II - 1) / II + 1;
   }
   DEBUG_BLOCKS(dbgs() << "    buildGraph, NCopies=" << NCopies << "\n");
   for (int S = 0; S < NCopies; S++) {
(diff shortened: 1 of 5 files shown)


andcarminati:
Hi @martien-de-jong, in the last update of this mutator, we left an orphan command line option: 

```
static cl::opt<unsigned> WAWStickyRegistersMemOpsThreshold(
    "aie-waw-sticky-register-mem-threshold", cl::Hidden, cl::init(4),
    cl::desc("Number of memory instructions to enable the register exclusion "
             "heuristic in WAW sticky registers dep. removal"));
```

Is it possible to include this removal in this PR?

andcarminati (dismissed):
LGTM.

gbossu:
I'd be happy to see QoR numbers, especially for `Neg` :)

llvm/lib/Target/AIE/AIEMachineScheduler.cpp
```diff
@@ -1254,6 +1255,8 @@ void AIEScheduleDAGMI::schedule() {
     // If it succeeds, we need to implement it, if we fail we fall back on the
     // normal loop schedule
     SchedImpl->buildGraph(*this, AA);
+    postProcessDAG();
```
krishnamtibrewala, line 1258: CHECK : Why do we need the call to "postProcessDAG" ?

gbossu: Some of the DAG mutations are required for correctness, e.g. giving a correct latency to memory edges

gbossu:
I'd rather not merge this now, running the DAG mutators for pipelining seems to have a nasty side effect on the final loop schedule when pipelining fails. In the post-processing loop of softmax, which essentially does `VLD -> VMUL -> VST.CONV`, we now have two NOPs at the end, probably because we see a memory dep between `VST.CONV` and the `VLD` of the next iteration.

gbossu (changes requested):
Marking as "request changes" to make sure we don't inadvertently merge this before looking into https://github.com/Xilinx/llvm-aie/pull/222#issuecomment-2441546532

andcarminati:
Now that this change was integrated through other PR, we can close this one!

PR #224, opened 2024-10-25
[GlobalISel] Match G_SHUFFLE_VECTORs representing sub-vector extracts

konstantinschwarz:


llvm/lib/CodeGen/GlobalISel/CombinerHelper.cpp
```diff
@@ -7254,6 +7255,60 @@ bool CombinerHelper::matchAddOverflow(MachineInstr &MI, BuildFnTy &MatchInfo) {
...
+  assert(MI.getOpcode() == TargetOpcode::G_SHUFFLE_VECTOR);
+  const Register DstReg = MI.getOperand(0).getReg();
+  const Register Src1Reg = MI.getOperand(1).getReg();
+  ArrayRef<int> Mask = MI.getOperand(3).getShuffleMask();
+
+  const LLT DstTy = MRI.getType(DstReg);
+  const LLT Src1Ty = MRI.getType(Src1Reg);
+
+  if (!DstTy.isVector() || !Src1Ty.isVector())
+    return false;
```
khallouh, line 7270: Is this needed? `G_SHUFFLE_VECTOR` only operates on vectors.

llvm/lib/CodeGen/GlobalISel/CombinerHelper.cpp
```diff
@@ -7254,6 +7255,60 @@ bool CombinerHelper::matchAddOverflow(MachineInstr &MI, BuildFnTy &MatchInfo) {
   return false;
 }
 
+bool CombinerHelper::matchShuffleToExtractSubvector(MachineInstr &MI,
+                                                    BuildFnTy &MatchInfo) {
+
+  assert(MI.getOpcode() == TargetOpcode::G_SHUFFLE_VECTOR);
+  const Register DstReg = MI.getOperand(0).getReg();
+  const Register Src1Reg = MI.getOperand(1).getReg();
+  ArrayRef<int> Mask = MI.getOperand(3).getShuffleMask();
```
khallouh, line 7264: Nit: `ShuffleMask`?

khallouh (commented):
The change looks good. Just a few nits. No comments on the tests yet because of the draft status. I have 2 questions: 
- Is the goal to match only half vector extract patterns or Sub-Vector extracts in general? (quarter extracts being also relevant for AIE's API)
- How about the missing canonicalization combine for shufflevector in GlobalISel? I think this is needed if we want to support extracts from the second vector input with the current combine implementation in this PR.

llvm/include/llvm/Target/GlobalISel/Combine.td
```diff
@@ -1527,6 +1527,16 @@ def combine_shuffle_concat : GICombineRule<
...
 
+// Combines Shuffles representing vector extracts into Unmerges
+// res = G_SHUFFLE_VECTORS a, b, <0, 1>
+// ===>
+// res, undef = G_UNMERGE_VALUES a
+def combine_shuffle_to_extract_vector : GICombineRule<
+  (defs root:$root, build_fn_matchinfo:$matchinfo),
+  (match (wip_match_opcode G_SHUFFLE_VECTOR):$root,
+        [{ return Helper.matchShuffleToExtractSubvector(*${root}, ${matchinfo}); }]),
+  (apply [{ Helper.applyBuildFn(*${root}, ${matchinfo}); }])>;
```
ValentijnvdBeek, line 1538: Avoid using `wip_match_opcode`, upstream really doesn't like this since it causes imprecise patterns and slows down the compilation. In this case, it should be very little difference compared to just matching SHUFFLE_VECTOR directly. 

https://llvm.org/docs/GlobalISel/MIRPatterns.html#gallery
See MR linked in: https://github.com/Xilinx/llvm-aie/pull/41#issuecomment-2429277423

llvm/lib/CodeGen/GlobalISel/CombinerHelper.cpp
ValentijnvdBeek: Actually, `G_SHUFFLE_VECTOR` can have scalar inputs. The opcode is a bit weird.

https://github.com/Xilinx/llvm-aie/blob/aie-public/llvm/test/CodeGen/AArch64/GlobalISel/legalize-shuffle-vector.mir#L119

llvm/include/llvm/Target/GlobalISel/Combine.td
```diff
@@ -1642,7 +1652,7 @@ def all_combines : GICombineGroup<[trivial_combines, vector_ops_combines,
     sub_add_reg, select_to_minmax, redundant_binop_in_equality,
     fsub_to_fneg, commute_constant_to_rhs, match_ands, match_ors,
     combine_concat_vector, double_icmp_zero_and_or_combine, match_addos,
-    combine_shuffle_concat]>;
+    combine_shuffle_concat, combine_shuffle_to_extract_vector]>;
```
ValentijnvdBeek, line 1655: Note: this is a bit of an expensive combiner to run since it iterates through a decent chunk of elements twice. Previously, this was an explicitly opt in combiner that a target enabled by calling a specific function. This will now run it on all targets, this is probably worth it. Also, upstream will probably be running those checks now anyways, so it doesn't matter as much. 

https://github.com/Xilinx/llvm-aie/blob/5a53fd639e17afd269114856ce8e34113fd0d573/llvm/lib/Target/AIE/AIE2PreLegalizerCombiner.cpp#L172
See MR linked in: https://github.com/Xilinx/llvm-aie/pull/41#issuecomment-2429277423

llvm/lib/CodeGen/GlobalISel/CombinerHelper.cpp
ValentijnvdBeek: I think that this misses the following edge case:
```
%0:_(<2 x s32>) = COPY $x0
%1:_(<2 x s32>) = UNDEF
%2:_(s32) = G_SHUFFLE_VECTOR %0, %1, shufflemask(0)
```

In this case, you could unmerge the first array into the destination. Since the target is not a vector, this would skip it. How useful this is? Who knows.

llvm/lib/CodeGen/GlobalISel/CombinerHelper.cpp
```diff
@@ -7254,6 +7255,60 @@ bool CombinerHelper::matchAddOverflow(MachineInstr &MI, BuildFnTy &MatchInfo) {
...
+    for (unsigned I = 0; I < NumDstElems; I++) {
+      if (Mask[I] == -1)
+        continue;
+
+      if (Mask[I] != ExtractMask[I])
+        return false;
+    }
+
+    return true;
+  };
```
ValentijnvdBeek, line 7290: I like this is a lot, this is great.

ValentijnvdBeek (commented):
Hi Konstantin, thanks for taking my design and generally unjunior/internifying. This looks very clean, great job. I am not able to run it a.t.m., but I will look at that whenever the PR is complete.

There are two small worries: 

This check is really expensive, this replaces the basic design but makes it hard to add the optimizations that might be needed to merge upstream. 

Having separate optimizer causes it to be missed by the `G_EXTRACT->G_INSERT->G_BUILD` sequences that might be valid representations. Upstream doesn't, for sensible reasons, canonize those into shufflevectors so you need to make sure that this optimizer is on those sequences as well.

See MR linked in: https://github.com/Xilinx/llvm-aie/pull/41#issuecomment-2429277423

llvm/lib/CodeGen/GlobalISel/CombinerHelper.cpp
khallouh: I think that case could even be matched to a `G_EXTRACT_VECTOR_ELT`. It might not be very useful for AIE's API since element extracts would be implemented using `extractelement` instead of `shufflevector`.

llvm/include/llvm/Target/GlobalISel/Combine.td
khallouh: What upstream checks are you referring to? I think the current `tryCombineShuffleVector` only matches merge-like patterns, but we could definitely move our `combine_shuffle_to_extract_vector` combine there, would that make it less expensive?

llvm/lib/CodeGen/GlobalISel/CombinerHelper.cpp
ValentijnvdBeek: Yeah, it is a bit obscure and the overhead it would cause is minimal.

llvm/include/llvm/Target/GlobalISel/Combine.td
ValentijnvdBeek: No, checking elements to the shuffle vector is just an O(n²) operations unless you use a hashmap. At the moment, this is done in `tryCombineShuffleVector` which a backend _can_ enable. This is now run always in all backends. I don't know exactly why someone made the decision to have that combiner be run explicitly rather than using tablegen like the others

llvm/lib/CodeGen/GlobalISel/CombinerHelper.cpp
```diff
@@ -7254,6 +7255,60 @@ bool CombinerHelper::matchAddOverflow(MachineInstr &MI, BuildFnTy &MatchInfo) {
...
+  if (!DstTy.isVector() || !Src1Ty.isVector())
+    return false;
+
+  const unsigned NumDstElems = DstTy.getNumElements();
+  const unsigned NumSrc1Elems = Src1Ty.getNumElements();
+
+  if (NumDstElems * 2 != NumSrc1Elems)
+    return false;
+
+  auto CheckExtractMask = [=](unsigned Start, unsigned NumElems) -> bool {
```
andcarminati, line 7278: nit: we can use by-reference capture and automatic return type deduction.

konstantinschwarz:
We added a lot of helper functions for matching sub-vector extraction operations recently. Those will be extended to cover the combine in this PR.

PR #225, opened 2024-10-28
[loop] moved loop iteration utils to LoopUtils, so that Middle end pa…

F-Stuckmann:
moved loop iteration utils to LoopUtils, so that Middle end passes also have access

### DIFF
--- a/llvm/include/llvm/Transforms/Utils/LoopUtils.h
+++ b/llvm/include/llvm/Transforms/Utils/LoopUtils.h
@@ -283,6 +283,12 @@ MDNode *updateIterCounts(LLVMContext &Context, MDNode *LoopID,
                          std::function<int64_t(int64_t)> FixMin,
                          std::function<int64_t(int64_t)> FixMax);
 
+/// Get Minimum Trip Count of the LoopID
+std::optional<int64_t> getMinTripCount(const MDNode *LoopID);
+
+/// Get Minimum Trip Count of the Loop
+std::optional<int64_t> getMinTripCount(const Loop *L);
+
 /// Look for the loop attribute that disables all transformation heuristic.
 bool hasDisableAllTransformsHint(const Loop *L);
 
--- a/llvm/lib/Transforms/Utils/LoopUtils.cpp
+++ b/llvm/lib/Transforms/Utils/LoopUtils.cpp
@@ -56,6 +56,7 @@ using namespace llvm::PatternMatch;
 
 static const char *LLVMLoopDisableNonforced = "llvm.loop.disable_nonforced";
 static const char *LLVMLoopDisableLICM = "llvm.licm.disable";
+static const char *LLVMLoopIterCount = "llvm.loop.itercount.range";
 
 bool llvm::formDedicatedExitBlocks(Loop *L, DominatorTree *DT, LoopInfo *LI,
                                    MemorySSAUpdater *MSSAU,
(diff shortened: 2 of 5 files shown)


llvm/include/llvm/Transforms/Utils/LoopUtils.h
```diff
@@ -283,6 +284,22 @@ MDNode *updateIterCounts(LLVMContext &Context, MDNode *LoopID,
...
+
+/// Get the LoopID from a single block loop or nullptr
+const MDNode *getLoopID(const MachineBasicBlock &LoopBlock);
+
+/// Get LoopID for Latch
+const MDNode *getLoopID(const BasicBlock *LoopBB);
+
+/// LoopBlock should hold the backedge. Since we generally call it for
+/// single block loops, that's automatically true.
+std::optional<int64_t> getMinTripCount(const MachineBasicBlock &LoopBlock);
```
gbossu: Could you leave helper functions that deal with `MachineBasicBlock` out of this header? I believe it's better to keep LoopInfo about LLVM IR, not MIR.

llvm/lib/Transforms/Utils/LoopUtils.cpp
```diff
@@ -415,6 +417,87 @@ MDNode *llvm::updateIterCounts(LLVMContext &Context, MDNode *LoopID,
...
+const MDNode *llvm::getLoopID(const BasicBlock *LoopBB) {
+  if (!LoopBB)
+    return nullptr;
+  const Instruction *TI = LoopBB->getTerminator();
+  if (!TI)
+    return nullptr;
+
+  const MDNode *LoopID = TI->getMetadata(LLVMContext::MD_loop);
+  return LoopID;
+}
```
gbossu: This seems to be a new function. Is that different from `LoopT *LoopInfoBase::getLoopFor(const BlockT *BB)`? I think we should avoid adding such low-level helpers if there's already a similar one.

llvm/include/llvm/Transforms/Utils/LoopUtils.h
```diff
@@ -283,6 +284,22 @@ MDNode *updateIterCounts(LLVMContext &Context, MDNode *LoopID,
...
+
```
andcarminati: I think we can handle everything here using `class BasicBlock`. In the backend we can have a wrapper function that uses *[getBasicBlock](https://llvm.org/doxygen/classllvm_1_1MachineBasicBlock.html#a4874816314c3308be0bf1e71de2078d8)(). I agree that this is the place for IR-only code.

llvm/include/llvm/Transforms/Utils/LoopUtils.h
```diff
@@ -283,6 +284,22 @@ MDNode *updateIterCounts(LLVMContext &Context, MDNode *LoopID,
...
+/// Loop for the attribute that provides minimum iteration counts
```
martien-de-jong: *Look*

llvm/lib/Transforms/Utils/LoopUtils.cpp
```diff
@@ -415,6 +417,87 @@ MDNode *llvm::updateIterCounts(LLVMContext &Context, MDNode *LoopID,
...
+  assert(dyn_cast<MDString>(LoopID->getOperand(0))->getString() ==
+         LLVMLoopIterCount);
+  if (LoopID) {
+    assert((LoopID->getNumOperands() >= 2 && LoopID->getNumOperands() <= 3) &&
+           "Iteration count hint should have one or two numeric operands.");
+    int64_t MinTripCount =
+        mdconst::extract<ConstantInt>(LoopID->getOperand(1))->getSExtValue();
+    assert(MinTripCount >= 0 && "Range lwb should not be negative.");
+    // LLVM_DEBUG(dbgs() << "AIELoopUtils: MinTripCount from pragma =  "
+    //                   << MinTripCount << "\n");
```
martien-de-jong: I'd say remove this LLVM_DEBUG, but at least get rid of AIE

llvm/include/llvm/Transforms/Utils/LoopUtils.h
F-Stuckmann: fixed it

llvm/lib/Transforms/Utils/LoopUtils.cpp
F-Stuckmann: fixed it

llvm/include/llvm/Transforms/Utils/LoopUtils.h
F-Stuckmann: removed the wrapper in `LoopUtils.h` and moved it to `AIELoopUtils.h`

llvm/lib/Transforms/Utils/LoopUtils.cpp
F-Stuckmann: I left this function alone in `AIELoopUtils.cpp`

llvm/include/llvm/Transforms/Utils/LoopUtils.h
F-Stuckmann: removed the `MachineBasicBlock `from `LoopUtils`

gbossu (approved):
LGTM

PR #228, opened 2024-10-29
Combine into VPUSH for s16 and s32 as well

abhinay-anubola:
Add support for 16-bit and 32-bit VInserts in PreLegalizerCombiner

### DIFF
--- a/llvm/lib/Target/AIE/AIE2PreLegalizerCombiner.cpp
+++ b/llvm/lib/Target/AIE/AIE2PreLegalizerCombiner.cpp
@@ -186,14 +186,28 @@ bool AIE2PreLegalizerCombinerImpl::tryToCombineSetExtract(
   return true;
 }
 
+unsigned getVInsertScalarSize(unsigned IntrinsicID) {
+  switch (IntrinsicID) {
+  case Intrinsic::aie2_vinsert8_I512:
+    return 8;
+  case Intrinsic::aie2_vinsert16_I512:
+    return 16;
+  case Intrinsic::aie2_vinsert32_I512:
+    return 32;
+  default:
+    return 0;
+  }
+}
+
 // Returns a map with InsertIndices and registers holding the insert values.
 std::map<unsigned, Register>
 AIE2PreLegalizerCombinerImpl::getVectorInsertIndices(
     MachineInstr *CurMI, unsigned SclSrcBits, MachineRegisterInfo &MRI) const {
   std::map<unsigned, Register> RegMap;
-  auto Is8BitVInsert = [](const MachineInstr *MI) {
-    return isa<GIntrinsic>(MI) && cast<GIntrinsic>(*MI).getIntrinsicID() ==
-                                      Intrinsic::aie2_vinsert8_I512;
+  auto IsVInsert = [](const MachineInstr *MI, unsigned SclSrcBits) {
+    return isa<GIntrinsic>(MI) &&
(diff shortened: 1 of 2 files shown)


llvm/lib/Target/AIE/AIE2PreLegalizerCombiner.cpp
```diff
@@ -202,7 +206,7 @@ AIE2PreLegalizerCombinerImpl::getVectorInsertIndices(
                                        Intrinsic::aie2_set_I512_I256);
   };
 
-  while (Is8BitVInsert(CurMI)) {
+  while (IsVInsert(CurMI)) {
```
gbossu: Does this mean we are allowing a mix in VINSERT instructions?

llvm/lib/Target/AIE/AIE2PreLegalizerCombiner.cpp
```diff
@@ -286,6 +300,12 @@ bool AIE2PreLegalizerCombinerImpl::tryToCombineIntrinsic(
   case Intrinsic::aie2_vinsert8_I512: {
```
andcarminati: I was wondering if we can refactor in some way this scalar size information. We use this information here and in `IsVInsert`. If we have something like `getVInsertScalarSize(...)` that returns zero if the intrinsic is not know would help. For example, in your previous lambda will be enough use `return getVInsertScalarSize(cast<GIntrinsic>(*MI).getIntrinsicID()) == SclSrcBits;` You can also drop your switch-case here and pass directly the value of `getVInsertScalarSize(...)` to `tryToCombineVectorInserts` if it is greater than 0.

andcarminati:
This PR looks good! Do you have some QoR numbers?

abhinay-anubola:
There is no change in QoR for now.

llvm/test/CodeGen/AIE/aie2/GlobalISel/combine-vinsert-sequence-prelegalizer.mir
```diff
@@ -328,3 +328,375 @@ body:             |
...
+    %28:_(<16 x s32>) = G_INTRINSIC intrinsic(@llvm.aie2.vinsert16.I512), %27(<16 x s32>), %12(s32), %7(s32)
+    %29:_(<16 x s32>) = G_INTRINSIC intrinsic(@llvm.aie2.vinsert16.I512), %28(<16 x s32>), %13(s32), %9(s32)
+    %30:_(<16 x s32>) = G_INTRINSIC intrinsic(@llvm.aie2.vinsert16.I512), %29(<16 x s32>), %14(s32), %11(s32)
+    %31:_(<16 x s32>) = G_INTRINSIC intrinsic(@llvm.aie2.vinsert16.I512), %30(<16 x s32>), %15(s32), %13(s32)
+    %32:_(<16 x s32>) = G_INTRINSIC intrinsic(@llvm.aie2.vinsert16.I512), %31(<16 x s32>), %16(s32), %15(s32)
+    PseudoRET implicit $lr, implicit %32
+...
+
+---
+name:            vinsert16-I128_idx_miss
```
martien-de-jong, line 440: Perhaps a very short description of why these cases do not yield a nice G_BUILD_VECTOR_TRUNC

martien-de-jong (dismissed):
Just one nit

martien-de-jong (approved):

PR #230, opened 2024-11-05
Martien.skippresched

martien-de-jong:
If we want to give the postpipeliner a chance on (bigger) loops that fall outside of the normal offloading  criteria, we can add a disable pragma to which the pre-pipeliner responds. In that case we also want to disable the scheduler since its actions clutter the view of the postpipeliner


llvm/test/CodeGen/AIE/aie2/schedule/swp/disable.ll
```diff
@@ -0,0 +1,119 @@
...
+  %exitcond.not = icmp eq i32 %inc, %n
+  br i1 %exitcond.not, label %for.cond.cleanup, label %for.body, !llvm.loop !6
+
+for.cond.cleanup:                                 ; preds = %for.body, %entry
+  ret i32 %or
+
+}
+
+!0 = !{i32 1, !"wchar_size", i32 4}
+!1 = !{!"clang version 17.0.0 (git@gitenterprise.xilinx.com:XRLabs/llvm-aie.git c07f38f221ca4d3736152230e12f5cfc637a1d47)"}
```
andcarminati: Can we change this version to something simple, like just clang version 19.0.0?

llvm/test/CodeGen/AIE/aie2/schedule/swp/disable.ll
```diff
@@ -0,0 +1,119 @@
+; NOTE: Assertions have been autogenerated by utils/update_mir_test_checks.py UTC_ARGS: --version 4
+;
+; This file is licensed under the Apache License v2.0 with LLVM Exceptions.
+; See https://llvm.org/LICENSE.txt for license information.
+; SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+;
+; (c) Copyright 2024 Advanced Micro Devices, Inc. or its affiliates
+
+; RUN: llc --mtriple=aie2 -O2 --stop-before=postmisched %s -o - | FileCheck %s
```
andcarminati: As we are disabling pre-scheduler for some blocks, I thing it is enough to stop after pre-scheduler itself. Also, we can run just the scheduler on a MIR (instead of IR) input to see that the sequence remains unchanged.

llvm/lib/Target/AIE/Utils/AIELoopUtils.cpp
```diff
@@ -32,6 +32,27 @@ std::optional<int64_t> getMinTripCount(const MachineBasicBlock &LoopBlock) {
   return getMinTripCount(getLoopID(LoopBlock));
 }
 
+std::optional<bool> getPipelinerDisabled(const MachineBasicBlock &LoopBlock) {
+  auto *LoopID = getLoopID(LoopBlock);
+  if (!LoopID) {
```
andcarminati, line 37: I think it is better to keep the same brace convention for all single-sentence ifs.

martien-de-jong: I agree. I will add braces everywhere. (The only way to have a universal brace convention...)

llvm/test/CodeGen/AIE/aie2/schedule/swp/disable.ll
martien-de-jong: I can just throw it out.

llvm/test/CodeGen/AIE/aie2/schedule/swp/disable.ll
martien-de-jong: Agree. I was lazy and copy-edited an existing test.

llvm/lib/Target/AIE/AIEMachineScheduler.cpp
```diff
@@ -852,6 +853,23 @@ void AIEPreRASchedStrategy::initialize(ScheduleDAGMI *DAG) {
   }
 }
 
+MachineBasicBlock *AIEPreRASchedStrategy::nextBlock() {
+  MachineBasicBlock *Next = nullptr;
+
+  // The pipeliner is usually disabled to give the postpipeliner a chance.
+  // The prescheduler also clutters the view of the postpipeliner, so we skip
+  // such blocks here.
+  auto Skip = [](MachineBasicBlock *Block) {
```
andcarminati, line 866: What do you think about a command line option to disable this?

martien-de-jong: We discussed that this is a user-pragma driven affair so that an option is not really necessary. Om second thought I perhaps add one anyway, so that we can easily assess the overall effect of prescheduling

andcarminati (approved):
LGTM.

PR #233, opened 2024-11-08
[AIE2] Support for allowing direct VEXTRACT to 20-bit registers

abhinay-anubola:
- This update introduces a new generic combiner that simplifies the sequence sext(trunc x) directly to x when applicable.
- Added VExtract combiner that enables above generic combiner, thus we have 20-bit vextract.
- The `MachineVerifier` has been updated to allow `G_AIE_SEXT_EXTRACT_VECTOR_ELT` and `G_AIE_ZEXT_EXTRACT_VECTOR_ELT` to accept 20-bit outputs. 
- Additionally, tests have been added and updated to reflect these functional changes.


llvm/lib/Target/AIE/AIECombinerHelper.cpp
```diff
@@ -901,6 +916,31 @@ bool modifyToS20(InstrNode Start, MachineRegisterInfo &MRI, MachineIRBuilder &B,
     Helper.tryCombineCopy(*StartNodeMI);
     return true;
   }
+  case TargetOpcode::G_INTRINSIC: {
+    if (isS20Intrinsic(*StartNodeMI, MRI)) {
```
andcarminati: I have a feeling that we need to refactor some subtarget-independent stuff in this file to TII. I think we can use and approach like `getVExtractOpInfo` in this same file. My goal is to generalize this CombinerHelper as much as possible.

llvm/lib/Target/AIE/AIECombinerHelper.cpp
```diff
@@ -567,6 +567,19 @@ void llvm::applyGlobalValOffset(MachineInstr &MI, MachineRegisterInfo &MRI,
       B.buildConstant(LLT::scalar(20), -static_cast<int64_t>(Offset)));
 }
 
+/// Check if the Intrinsic can produce S20
+static bool isS20Intrinsic(const MachineInstr &MI,
+                           const MachineRegisterInfo &MRI) {
+  const unsigned IntrinsicID = cast<GIntrinsic>(MI).getIntrinsicID();
+  if (IntrinsicID == Intrinsic::aie2_vextract_elem8_I512 ||
+      IntrinsicID == Intrinsic::aie2_vextract_elem16_I512 ||
+      IntrinsicID == Intrinsic::aie2_vextract_elem32_I512) {
```
krishnamtibrewala: nit : use switch case, it will make it more readable.

llvm/lib/Target/AIE/AIECombinerHelper.cpp
```diff
@@ -567,6 +567,19 @@ void llvm::applyGlobalValOffset(MachineInstr &MI, MachineRegisterInfo &MRI,
...
+    // Check if the sign value is constant
+    return getIConstantVRegSExtVal(MI.getOperand(4).getReg(), MRI).has_value();
```
krishnamtibrewala: define a new const Reg with a meaning full name.

llvm/lib/Target/AIE/AIECombinerHelper.cpp
krishnamtibrewala: +1.  
Luckily we do not have many hardcoded aie2_* intrinsic in the file.

llvm/lib/Target/AIE/AIECombinerHelper.cpp
```diff
@@ -901,6 +916,31 @@ bool modifyToS20(InstrNode Start, MachineRegisterInfo &MRI, MachineIRBuilder &B,
     Helper.tryCombineCopy(*StartNodeMI);
     return true;
   }
+  case TargetOpcode::G_INTRINSIC: {
+    if (isS20Intrinsic(*StartNodeMI, MRI)) {
+      Register ExtractDstReg = StartNodeMI->getOperand(0).getReg();
+      // Note: Operand 1 is the ID of the intrinsic
+      const Register SrcReg0 = StartNodeMI->getOperand(2).getReg();
+      const Register SrcReg1 = StartNodeMI->getOperand(3).getReg();
+      const Register SignReg = StartNodeMI->getOperand(4).getReg();
```
krishnamtibrewala: We should not do like this; this is very error prone. 
We might add a new intrinsic in inS20Intrinsic and this would then fail  if we do not have the required number of operands.

llvm/lib/Target/AIE/AIECombinerHelper.cpp
```diff
@@ -901,6 +916,31 @@ bool modifyToS20(InstrNode Start, MachineRegisterInfo &MRI, MachineIRBuilder &B,
...
+
+      auto SignVal = getIConstantVRegSExtVal(SignReg, MRI);
```
krishnamtibrewala: nit : const auto SignVal

llvm/lib/Target/AIE/AIECombinerHelper.cpp
```diff
@@ -901,6 +916,31 @@ bool modifyToS20(InstrNode Start, MachineRegisterInfo &MRI, MachineIRBuilder &B,
...
+      assert(SignVal.has_value() && "Expected SignVal to be constant");
+      // Erase the original instruction and set the type for the destination
+      // register
+      Helper.eraseInst(*StartNodeMI);
+      MRI.setType(ExtractDstReg, S20);
+      const unsigned Opcode = SignVal.value()
+                                  ? AIE2::G_AIE_SEXT_EXTRACT_VECTOR_ELT
+                                  : AIE2::G_AIE_ZEXT_EXTRACT_VECTOR_ELT;
+      StartNodeMI =
+          B.buildInstr(Opcode, {ExtractDstReg}, {SrcReg0, SrcReg1}).getInstr();
```
krishnamtibrewala: Please check if the Observer will be aware of the newly added instruction

llvm/lib/Target/AIE/AIECombinerHelper.cpp
```diff
@@ -901,6 +916,31 @@ bool modifyToS20(InstrNode Start, MachineRegisterInfo &MRI, MachineIRBuilder &B,
...
+    } else {
+      LLVM_DEBUG(dbgs() << "Node :" << *StartNodeMI);
+      llvm_unreachable("Unexpected OpCode, while modifying IR");
```
krishnamtibrewala: nit : "Unexpected G_INTRINSIC ..."

llvm/lib/Target/AIE/AIECombinerHelper.cpp
krishnamtibrewala: Check : How do we make sure that the newly created instruction is in right order. By order I mean is inserted between expected set of instruction

llvm/lib/Target/AIE/AIECombinerHelper.cpp
```diff
@@ -910,7 +950,9 @@ bool modifyToS20(InstrNode Start, MachineRegisterInfo &MRI, MachineIRBuilder &B,
   switch (StartNodeMI->getOpcode()) {
   case TargetOpcode::COPY:
   case TargetOpcode::G_LOAD:
-  case TargetOpcode::G_PHI: {
+  case TargetOpcode::G_PHI:
+  case AIE2::G_AIE_ZEXT_EXTRACT_VECTOR_ELT:
+  case AIE2::G_AIE_SEXT_EXTRACT_VECTOR_ELT: {
```
krishnamtibrewala: We should somehow abstract this to arch specific files. As @andcarminati  mentioned we want all our combiner passes to work generically.

llvm/test/CodeGen/AIE/aie2/GlobalISel/prelegalizercombiner-s20-narrowing.mir
```diff
@@ -781,3 +781,294 @@ body:             |
...
+  ; CHECK-NEXT:   liveins: $p0, $x0
+  ; CHECK-NEXT: {{  $}}
+  ; CHECK-NEXT:   [[COPY:%[0-9]+]]:_(p0) = COPY $p0
+  ; CHECK-NEXT:   [[C:%[0-9]+]]:_(s32) = G_CONSTANT i32 0
+  ; CHECK-NEXT:   [[COPY1:%[0-9]+]]:_(<64 x s8>) = COPY $x0
+  ; CHECK-NEXT: {{  $}}
+  ; CHECK-NEXT: bb.1:
+  ; CHECK-NEXT:   successors: %bb.1(0x80000000)
+  ; CHECK-NEXT: {{  $}}
+  ; CHECK-NEXT:   [[AIE_SEXT_EXTRACT_VECTOR_ELT:%[0-9]+]]:_(s20) = G_AIE_SEXT_EXTRACT_VECTOR_ELT [[COPY1]](<64 x s8>), [[C]](s32)
```
krishnamtibrewala: this is incorrect, we have moved vextract into a new basic block.

llvm/test/CodeGen/AIE/aie2/GlobalISel/prelegalizercombiner-s20-narrowing.mir
```diff
@@ -781,3 +781,294 @@ body:             |
...
+  ; CHECK-NEXT:   liveins: $p0, $x0
+  ; CHECK-NEXT: {{  $}}
+  ; CHECK-NEXT:   [[COPY:%[0-9]+]]:_(p0) = COPY $p0
+  ; CHECK-NEXT:   [[C:%[0-9]+]]:_(s32) = G_CONSTANT i32 0
+  ; CHECK-NEXT:   [[COPY1:%[0-9]+]]:_(<32 x s16>) = COPY $x0
+  ; CHECK-NEXT: {{  $}}
+  ; CHECK-NEXT: bb.1:
+  ; CHECK-NEXT:   successors: %bb.1(0x80000000)
+  ; CHECK-NEXT: {{  $}}
+  ; CHECK-NEXT:   [[AIE_SEXT_EXTRACT_VECTOR_ELT:%[0-9]+]]:_(s20) = G_AIE_SEXT_EXTRACT_VECTOR_ELT [[COPY1]](<32 x s16>), [[C]](s32)
```
krishnamtibrewala: same issue as mentioned in https://github.com/Xilinx/llvm-aie/pull/233/files#r1834826147

krishnamtibrewala:
Given that you mentioned there are no QoR gain, I would recommend you to re look at the instruction that consume S20 type reg.
Because for the optimization starts to trace back from an instruction that consumes S20 type which might not be captured in `isNativeS20Consumer` function.

llvm/lib/Target/AIE/AIECombinerHelper.cpp
andcarminati: The good thing: we already have this in `TII->getGenericExtractVectorEltOpcode`

llvm/test/CodeGen/AIE/aie2/GlobalISel/prelegalizercombiner-s20-narrowing.mir
andcarminati: As this patterns matches from G_TRUNC, actually is is ok because usually set the builder insertion point to the root of the matched chain (not a rule, but it is common).

krishnamtibrewala: In this specific case it will work but there can be a case where it won't. Say we had a use of virtual reg defined in the same BB.

llvm/lib/Target/AIE/AIECombinerHelper.cpp
abhinay-anubola: When I attempt to replace `AIE2::G_AIE_Z/SEXT_EXTRACT_VECTOR_ELT` with `TII->getGenericExtractVectorEltOpcode(SignExt)`, I encounter an issue because the latter returns a dynamic value, which cannot be used directly in a switch case label.

andcarminati: I suggest to concentrate those cases in the default case (with ifs), because is is important to have generic code here.

llvm/lib/Target/AIE/AIECombinerHelper.cpp
```diff
@@ -925,11 +959,27 @@ bool modifyToS20(InstrNode Start, MachineRegisterInfo &MRI, MachineIRBuilder &B,
...
     break;
   }
   default: {
-    LLVM_DEBUG(dbgs() << "Node :" << *StartNodeMI);
-    llvm_unreachable("Unexpected OpCode, while modifying IR");
+    const unsigned Opcode = StartNodeMI->getOpcode();
+    // Check if the opcode corresponds to a generic extract vector element
+    // operation. Since these opcodes are determined at runtime, we use an if
+    // statement to handle them.
+    if (Opcode == TII->getGenericExtractVectorEltOpcode(false) ||
```
andcarminati: I like in this way!

llvm/lib/Target/AIE/AIECombinerHelper.cpp
```diff
@@ -581,6 +581,11 @@ static bool canProduceS20(const MachineRegisterInfo &MRI,
   case TargetOpcode::G_CONSTANT:
   case TargetOpcode::G_IMPLICIT_DEF:
     return true;
+  case TargetOpcode::G_INTRINSIC: {
+    const AIEBaseSubtarget &STI = AIEBaseSubtarget::get(*MI.getMF());
+    const AIEBaseInstrInfo *TII = STI.getInstrInfo();
+    return TII->isS20Intrinsic(MI, MRI);
+  }
```
gbossu: I'd rather have a pre-combiner than transforms constant-sign vextract intrinsics into `G_AIE_ZEXT_EXTRACT_VECTOR_ELT` or `G_AIE_SEXT_EXTRACT_VECTOR_ELT`. Then we do not need to define `isS20Intrinsic`.

llvm/lib/Target/AIE/AIECombinerHelper.cpp
```diff
@@ -901,16 +908,43 @@ bool modifyToS20(InstrNode Start, MachineRegisterInfo &MRI, MachineIRBuilder &B,
     Helper.tryCombineCopy(*StartNodeMI);
     return true;
   }
+  case TargetOpcode::G_INTRINSIC: {
+    if (TII->isS20Intrinsic(*StartNodeMI, MRI)) {
```
gbossu: Here we insert a generic vector extract, regardless of what the intrinsic is, so I think `isS20Intrinsic` is not an accurate name and should rather be `isVExtractIntrinsic`. Still, I'd rather have generic opcodes like mentioned in https://github.com/Xilinx/llvm-aie/pull/233/commits/d5d7cf087b0a27d02cdb7a0bf1f68f3e618470a1#r1839776194 and not have to deal with intrinsics at all.

llvm/lib/Target/AIE/AIECombinerHelper.cpp
```diff
@@ -581,9 +593,14 @@ static bool canProduceS20(const MachineRegisterInfo &MRI,
   case TargetOpcode::G_CONSTANT:
   case TargetOpcode::G_IMPLICIT_DEF:
     return true;
-  default:
+  default: {
+    // Since generic extract vector element opcodes are determined at runtime,
+    // we call a function in if statement to handle them.
```
gbossu: Nit: they are not determined at run-time, it's just that they are different for each AIE version, so we need an extra layer to abstract them.

llvm/lib/Target/AIE/AIECombinerHelper.cpp
```diff
@@ -902,15 +919,21 @@ bool modifyToS20(InstrNode Start, MachineRegisterInfo &MRI, MachineIRBuilder &B,
...
+    }
   }
   }
 
-  switch (StartNodeMI->getOpcode()) {
-  case TargetOpcode::COPY:
-  case TargetOpcode::G_LOAD:
-  case TargetOpcode::G_PHI: {
+  // Function to handle the modification of instructions
+  auto modifyInstructionUses = [&](MachineInstr *StartNodeMI) {
```
gbossu: Nit: `ModifyInstructionUses `

llvm/lib/Target/AIE/AIE2PreLegalizerCombiner.cpp
```diff
@@ -288,6 +290,26 @@ bool AIE2PreLegalizerCombinerImpl::tryToCombineVectorInserts(
   return true;
 }
 
+bool AIE2PreLegalizerCombinerImpl::tryToCombineVExtractElt(
+    MachineInstr &MI) const {
```
krishnamtibrewala, line 342: Please add an assert to check the specific "opcode" we expected for the MI here. 
What I understand is this works for specific base opcode. This will improve readability for future and right away tell which Opcode this function should be called for. 

Similar this is done in multiple target in legalizer or instruction select.

llvm/lib/Target/AIE/AIECombinerHelper.cpp
```diff
@@ -902,15 +919,21 @@ bool modifyToS20(InstrNode Start, MachineRegisterInfo &MRI, MachineIRBuilder &B,
     return true;
   }
   default: {
-    LLVM_DEBUG(dbgs() << "Node :" << *StartNodeMI);
-    llvm_unreachable("Unexpected OpCode, while modifying IR");
+    // Since generic extract vector element opcodes are determined at runtime,
+    // we call a function in if statement to handle them.
+    if (IsGenericExtractVectorElt(*StartNodeMI)) {
```
krishnamtibrewala: can we do the following, rather than doing it default: ?

```
case TargetOpcode::G_INTRINSIC
    // Since generic extract vector element opcodes are determined at runtime,
    // we call a function in if statement to handle them.
    if (IsGenericExtractVectorElt(*StartNodeMI)) {
      Observer.changingInstr(*StartNodeMI);
      MRI.setType(StartNodeMI->getOperand(0).getReg(), S20);
      Observer.changedInstr(*StartNodeMI);
    } 
```

llvm/lib/Target/AIE/AIECombinerHelper.cpp
```diff
@@ -581,9 +593,14 @@ static bool canProduceS20(const MachineRegisterInfo &MRI,
   case TargetOpcode::G_CONSTANT:
   case TargetOpcode::G_IMPLICIT_DEF:
     return true;
-  default:
+  default: {
+    // Since generic extract vector element opcodes are determined at runtime,
+    // we call a function in if statement to handle them.
+    if (IsGenericExtractVectorElt(MI))
+      return true;
```
krishnamtibrewala: can we do the following, instead of putting directly in default ? 

```
case TargetOpcode::G_INTRINSIC
    if (IsGenericExtractVectorElt(MI))
      return true;
```

llvm/lib/Target/AIE/AIECombinerHelper.cpp
abhinay-anubola: Gaetan's suggestion to do in this way https://github.com/Xilinx/llvm-aie/pull/233#discussion_r1839779739

llvm/lib/Target/AIE/AIECombinerHelper.cpp
abhinay-anubola: Same as above

llvm/lib/Target/AIE/AIE2PreLegalizerCombiner.cpp
abhinay-anubola: Like other combines in this file, we are calling this function for multiple intrinsic IDs. However, I don't see the necessity for assert, as there are no assertions in the other combines either.
```
case Intrinsic::aie2_vextract_elem8_I512:
case Intrinsic::aie2_vextract_elem16_I512:
case Intrinsic::aie2_vextract_elem32_I512: {
  return tryToCombineVExtractElt(MI);
}

File: llvm/lib/Target/AIE/AIE2PreLegalizerCombiner.cpp [hunk 19]
```diff
@@ -288,6 +293,89 @@ bool AIE2PreLegalizerCombinerImpl::tryToCombineVectorInserts(
...
+///        %30:_(s20) = G_SEXT %20(s8)
+/// To :   %10:_(s32) = G_INTRINSIC intrinsic(@llvm.aie2.vextract.elem8.I512),
+/// %2(<64 x s8>), %0(s32), %1(s32)
+///        %30:_(s20) = G_TRUNC %10(s32)
+/// This also enables S20Narrowing for vextract
+bool AIE2PreLegalizerCombinerImpl::tryToCombineTruncExt(
+    Register DstReg, bool SignVal, unsigned SrcEltSize) const {
+  // Checks if a given register has non-debug user with a specific opcode and
+  // destination size, and return that user.
+  auto GetUseWithOpCode =
```
andcarminati: I suggest to simplify this lambda to capture MRI directly:

```
  auto GetUseWithOpCode =
      [&](const Register Reg, const unsigned OpcodeToCheck,
          const unsigned DstSize) -> std::optional<MachineInstr *> {
    for (auto &Use : MRI.use_nodbg_instructions(Reg)) {
      if (Use.getOpcode() == OpcodeToCheck) {
        const LLT DstRegTy = MRI.getType(Use.getOperand(0).getReg());
        if (DstRegTy.getSizeInBits() == DstSize)
          return &Use;
      }
    }
```

llvm/lib/Target/AIE/AIE2PreLegalizerCombiner.cpp
```diff
@@ -288,6 +293,89 @@ bool AIE2PreLegalizerCombinerImpl::tryToCombineVectorInserts(
...
+    const Register UseDstReg = TruncMI->getOperand(0).getReg();
+    // Ensure G_TRUNC has a single non-debug user before safely eliminating it.
+    if (!MRI.hasOneNonDBGUser(UseDstReg))
+      return false;
+    if (auto Ext = GetUseWithOpCode(MRI, UseDstReg, ExtOpcode, 20)) {
+      MachineInstr *ExtMI = Ext.value();
+      MachineIRBuilder MIRBuilder(*ExtMI);
+      MIRBuilder.buildInstr(TargetOpcode::G_TRUNC, {ExtMI->getOperand(0)},
+                            {DstReg});
+      ExtMI->eraseFromParent();
```
andcarminati: I suggest to not erase 2 instructions here, because we may corrupt the iterator. Let `ExtMI` to be removed by DCE.

llvm/lib/Target/AIE/AIE2PreLegalizerCombiner.cpp
```diff
@@ -288,6 +293,89 @@ bool AIE2PreLegalizerCombinerImpl::tryToCombineVectorInserts(
...
+  const auto SignVal = getIConstantVRegSExtVal(SignReg, MRI);
+  if (!SignVal.has_value())
+    return false;
+
+  const LLT SrcVecTy = MRI.getType(SrcReg);
+  const unsigned SrcEltSize = SrcVecTy.getScalarSizeInBits();
+  if (SrcEltSize == 8 || SrcEltSize == 16) {
+    tryToCombineTruncExt(DstReg, SignVal.value(), SrcEltSize);
+  }
+  const unsigned Opcode = SignVal.value() ? AIE2::G_AIE_SEXT_EXTRACT_VECTOR_ELT
```
andcarminati: We consider this as AIE2-specific, but we can always use the same source of knowledge: `TII.getGenericExtractVectorEltOpcode()`

llvm/lib/Target/AIE/AIE2PreLegalizerCombiner.cpp
```diff
@@ -288,6 +293,89 @@ bool AIE2PreLegalizerCombinerImpl::tryToCombineVectorInserts(
...
+bool AIE2PreLegalizerCombinerImpl::tryToCombineVExtractElt(
+    MachineInstr &MI) const {
+  const Register DstReg = MI.getOperand(0).getReg();
+  // In this case of G_INTRINSIC operand 1 is target intrinsic
+  const Register SrcReg = MI.getOperand(2).getReg();
+  const Register IdxReg = MI.getOperand(3).getReg();
+  const Register SignReg = MI.getOperand(4).getReg();
+
+  const auto SignVal = getIConstantVRegSExtVal(SignReg, MRI);
+  if (!SignVal.has_value())
```
andcarminati: nit: use just `if (!SignVal)`

llvm/lib/Target/AIE/AIE2PreLegalizerCombiner.cpp
andcarminati: Combiner engine may access freed memory:

```
    while (!WorkList.empty()) {
      MachineInstr *CurrInst = WorkList.pop_back_val();
      LLVM_DEBUG(dbgs() << "\nTry combining " << *CurrInst;);
      Changed |= tryCombineAll(*CurrInst);
      WLObserver->reportFullyCreatedInstrs();
    }
```

llvm/lib/Target/AIE/AIE2PreLegalizerCombiner.cpp
```diff
@@ -288,6 +293,89 @@ bool AIE2PreLegalizerCombinerImpl::tryToCombineVectorInserts(
...
+    return false;
+
+  const LLT SrcVecTy = MRI.getType(SrcReg);
+  const unsigned SrcEltSize = SrcVecTy.getScalarSizeInBits();
+  if (SrcEltSize == 8 || SrcEltSize == 16) {
+    tryToCombineTruncExt(DstReg, SignVal.value(), SrcEltSize);
```
andcarminati: I think we can trigger the same multiple erased instructions problem here.

llvm/lib/Target/AIE/AIE2PreLegalizerCombiner.cpp
andcarminati: Question is, should we refactor `void makeDeadMI(MachineInstr &MI, MachineRegisterInfo &MRI);`  as a common utility function?

abhinay-anubola: We cant erase just the TRUNC as it is input for EXT.

llvm/lib/Target/AIE/AIE2PreLegalizerCombiner.cpp
```diff
@@ -288,6 +293,89 @@ bool AIE2PreLegalizerCombinerImpl::tryToCombineVectorInserts(
...
+                            {DstReg});
+      ExtMI->eraseFromParent();
+      TruncMI->eraseFromParent();
+      return true;
+    }
+  }
+  return false;
+}
+
+bool AIE2PreLegalizerCombinerImpl::tryToCombineVExtractElt(
```
andcarminati, line 341: We can have a description of the combiner here.

llvm/lib/Target/AIE/AIE2PreLegalizerCombiner.cpp
andcarminati: You will also erase MI (intrinsic) in the caller function.

llvm/lib/Target/AIE/AIE2PreLegalizerCombiner.cpp
```diff
@@ -288,6 +293,89 @@ bool AIE2PreLegalizerCombinerImpl::tryToCombineVectorInserts(
...
+    MachineInstr &MI) const {
+  const Register DstReg = MI.getOperand(0).getReg();
+  // In this case of G_INTRINSIC operand 1 is target intrinsic
+  const Register SrcReg = MI.getOperand(2).getReg();
+  const Register IdxReg = MI.getOperand(3).getReg();
+  const Register SignReg = MI.getOperand(4).getReg();
+
+  const auto SignVal = getIConstantVRegSExtVal(SignReg, MRI);
+  if (!SignVal)
```
andcarminati, line 350: This looks like an assert. We always need constant signal to be able to handle this intrinsic.

llvm/lib/Target/AIE/AIE2PreLegalizerCombiner.cpp
```diff
@@ -288,6 +293,89 @@ bool AIE2PreLegalizerCombinerImpl::tryToCombineVectorInserts(
...
+    return false;
+  const LLT SrcVecTy = MRI.getType(SrcReg);
+  const unsigned SrcEltSize = SrcVecTy.getScalarSizeInBits();
+  if (SrcEltSize == 8 || SrcEltSize == 16) {
+    tryToCombineTruncExt(DstReg, SignVal.value(), SrcEltSize);
```
andcarminati: I understood that this `tryToCombineTruncExt` is only safe because we know the range of the output of the defining instruction. However, it would be good to have this as a separated combiner (may looking to the input reg. def.). It will help to solve the erasing problem and we also can have different tests and, as a gift, less coupled code.

llvm/lib/Target/AIE/AIE2PreLegalizerCombiner.cpp
```diff
@@ -290,6 +293,62 @@ bool AIE2PreLegalizerCombinerImpl::tryToCombineVectorInserts(
...
+/// Or even:
+/// From : %10:_(s32) = G_INTRINSIC intrinsic(@llvm.aie2.vextract.elem8.I512),
+/// %2(<64 x s8>), %0(s32), %1(s32)
+///        %20:_(s8) = G_TRUNC %10(s32)
+///        %30:_(s20) = G_SEXT %20(s8)
+/// To :   %10:_(s32) = G_INTRINSIC intrinsic(@llvm.aie2.vextract.elem8.I512),
+/// %2(<64 x s8>), %0(s32), %1(s32)
+///        %30:_(s20) = G_TRUNC %10(s32)
+/// This also enables S20Narrowing for vextract
+bool AIE2PreLegalizerCombinerImpl::tryToCombineTruncExt(
```
konstantinschwarz: Instead of doing this combine here, why don't we just use `G_ASSERT_ZEXT`/`G_ASSERT_SEXT`?
In your previous commit 16445fb7d1ae5ecec7584ac2780cbd3a1310da4c,
we should pre-select
```
%0:_(s32) = G_INTRINSIC intrinsic(@llvm.aie2.vextract.elem8.I512), ...
```
into
```
%new:_(s32) = G_AIE_ZEXT_EXTRACT_VECTOR_ELT
%0:_(s32) = G_ASSERT_ZEXT %new, 16
```

Then you should get the G_TRUNC + G_ZEXT combine for free.

konstantinschwarz (changes requested):

llvm/lib/Target/AIE/AIE2PreLegalizerCombiner.cpp
andcarminati: We should have something like:
```
    %0:_(s32) = G_CONSTANT i32 7
    %2:_(<32 x s16>) = COPY $x0
    %9:_(s20) = G_AIE_ZEXT_EXTRACT_VECTOR_ELT %2(<32 x s16>), %0(s32)
    %3:_(s20) = G_ASSERT_ZEXT %9, 16
    %4:_(s16) = G_TRUNC %3(s20)
    %5:_(s20) = G_ZEXT %4(s16)
    %6:_(p0) = G_CONSTANT i20 0
    %7:_(p0), %8:_(s20) = G_INTRINSIC intrinsic(@llvm.aie2.add.2d), %6(p0), %5(s20), %5(s20), %5(s20), %5(s20)
    PseudoRET implicit $lr, implicit %7(p0)
```

However, for SEXT case, there is no combine pattern AFAIK.

andcarminati: We can try to match this pattern explicitly looking to ASSERT, something like:

```
bool CombinerHelper::matchCombineSextTrunc(MachineInstr &MI, Register &Reg) {
  assert(MI.getOpcode() == TargetOpcode::G_SEXT && "Expected a G_SEXT");
  Register SrcReg = MI.getOperand(1).getReg();
  LLT DstTy = MRI.getType(DstReg);
  if (mi_match(SrcReg, MRI,
               m_GTrunc(m_all_of(m_Reg(Reg), m_SpecificType(DstTy))))) {
    unsigned DstSize = DstTy.getScalarSizeInBits();
    unsigned SrcSize = MRI.getType(SrcReg).getScalarSizeInBits();
    MachineInstr* MIDef = MRI.getVRegDef(Reg);
    if (MIDef->getOpcode() != TargetOpcode::G_ASSERT_SEXT)
      return false;
    unsigned ExtBits = MIDef->getOperand(2).getImm();
    return SrcSize == ExtBits;
  }
  return false;
}
```

And apply `replaceSingleDefInstWithReg`

konstantinschwarz: Instead of hardcoding the G_ASSERT_SEXT, I'd suggest to use the KnownBits analysis to check if we have known sign bits: ` KB->computeNumSignBits(Reg)`

abhinay-anubola: Above pattern is not matching in our case as m_SpecificType is s20 and DstTy is s32.
Even for ZEXT is not matching here, but combining is getting done somewhere later to pre combiner

andcarminati: Hi @abhinay-anubola , all you need to do is to translate the intrinsic to:

```
    %9:_(s20) = G_AIE_ZEXT_EXTRACT_VECTOR_ELT %2(<32 x s16>), %0(s32)
    %3:_(s20) = G_ASSERT_ZEXT %9, 16
    %10:_(s32) = G_ZEXT %3(s20)
```

Globally speaking you should have:

```
    %0:_(s32) = G_CONSTANT i32 7
    %2:_(<32 x s16>) = COPY $x0
    %9:_(s20) = G_AIE_ZEXT_EXTRACT_VECTOR_ELT %2(<32 x s16>), %0(s32)
    %3:_(s20) = G_ASSERT_ZEXT %9, 16
    %10:_(s32) = G_ZEXT %3(s20)
    %4:_(s16) = G_TRUNC %10(s32)
    %5:_(s20) = G_ZEXT %4(s16)
    %6:_(p0) = G_CONSTANT i20 0
    %7:_(p0), %8:_(s20) = G_INTRINSIC intrinsic(@llvm.aie2.add.2d), %6(p0), %5(s20), %5(s20), %5(s20), %5(s20)
    PseudoRET implicit $lr, implicit %7(p0)
```

That can be automatically combined to:

```
    %0:_(s32) = G_CONSTANT i32 7
    %2:_(<32 x s16>) = COPY $x0
    %9:_(s20) = G_AIE_ZEXT_EXTRACT_VECTOR_ELT %2(<32 x s16>), %0(s32)
    %3:_(s20) = G_ASSERT_ZEXT %9, 16
    %6:_(p0) = G_CONSTANT i20 0
    %7:_(p0), %8:_(s20) = G_INTRINSIC intrinsic(@llvm.aie2.add.2d), %6(p0), %3(s20), %3(s20), %3(s20), %3(s20)
    PseudoRET implicit $lr, implicit %7(p0)
```

For the SEXT case, we need a new combiner looking to  `computeNumSignBits`.

andcarminati: Hi @abhinay-anubola, just have in mind that we need to filter the cases: just consider cases with single consumer (s20).

llvm/lib/Target/AIE/AIE2PreLegalizerCombiner.cpp
gbossu: I thought so too, but actually we can have `VEXTRACT` with a dynamic sign coming from `$crVaddSign`

llvm/lib/Target/AIE/AIE2PreLegalizerCombiner.cpp
```diff
@@ -290,6 +293,62 @@ bool AIE2PreLegalizerCombinerImpl::tryToCombineVectorInserts(
   return true;
 }
 
+/// \returns true if it is possible to combine the below sequence of MIRs
+/// From : %10:_(s32) = G_INTRINSIC intrinsic(@llvm.aie2.vextract.elem16.I512),
+/// %2(<32 x s16>), %0(s32), %1(s32)
+///        %20:_(s16) = G_TRUNC %10(s32)
+///        %30:_(s20) = G_SEXT %20(s16)
+/// To :   %10:_(s32) = G_INTRINSIC intrinsic(@llvm.aie2.vextract.elem16.I512),
+/// %2(<32 x s16>), %0(s32), %1(s32)
```
gbossu: Nit: Formatting makes it hard to read

llvm/lib/CodeGen/GlobalISel/CombinerHelper.cpp
```diff
@@ -2465,6 +2465,20 @@ bool CombinerHelper::matchCombineZextTrunc(MachineInstr &MI, Register &Reg) {
   return false;
 }
 
+bool CombinerHelper::matchCombineSextTrunc(MachineInstr &MI, Register &Reg) {
```
andcarminati, line 2468: nit: `const` everywhere:

```
 bool CombinerHelper::matchCombineSextTrunc(MachineInstr &MI, Register &Reg) {
   assert(MI.getOpcode() == TargetOpcode::G_SEXT && "Expected a G_SEXT");
-  Register DstReg = MI.getOperand(0).getReg();
-  Register SrcReg = MI.getOperand(1).getReg();
-  LLT DstTy = MRI.getType(DstReg);
   if (mi_match(SrcReg, MRI,
                m_GTrunc(m_all_of(m_Reg(Reg), m_SpecificType(DstTy))))) {
-    unsigned DstSize = DstTy.getScalarSizeInBits();
-    unsigned SrcSize = MRI.getType(SrcReg).getScalarSizeInBits();
+    const unsigned DstSize = DstTy.getScalarSizeInBits();
+    const unsigned SrcSize = MRI.getType(SrcReg).getScalarSizeInBits();
     return KB->computeNumSignBits(Reg) >= (DstSize - SrcSize + 1);
   }
   return false;
```

llvm/lib/Target/AIE/AIE2PreLegalizerCombiner.cpp
```diff
@@ -288,6 +292,89 @@ bool AIE2PreLegalizerCombinerImpl::tryToCombineVectorInserts(
...
+  const LLT S20 = LLT::scalar(20);
+  Register DstReg20Bit = MRI.createGenericVirtualRegister(S20);
+  Register ExtReg20Bit = MRI.createGenericVirtualRegister(S20);
+  MachineIRBuilder MIRBuilder(MI);
+
+  MIRBuilder.buildInstr(Opcode, {DstReg20Bit}, {SrcReg, IdxReg});
+  MIRBuilder.buildAssertInstr(AssertExtOpcode, ExtReg20Bit, DstReg20Bit,
+                              SrcEltSize);
+  MIRBuilder.buildInstr(ExtOpcode, {DstReg}, {ExtReg20Bit});
+  MI.eraseFromParent();
```
andcarminati, line 375: Now we are safe ;-)

llvm/lib/Target/AIE/AIE2PreLegalizerCombiner.cpp
```diff
@@ -288,6 +292,89 @@ bool AIE2PreLegalizerCombinerImpl::tryToCombineVectorInserts(
   return true;
 }
 
+/// Determines if it is safe to combine vextract by checking the uses of DstReg,
+/// specfically for a pattern involving TRUNC followed by EXT.
+bool AIE2PreLegalizerCombinerImpl::canCombineVExtractElt(
+    Register DstReg, bool SignVal, unsigned SrcEltSize) const {
+  // Returns the single non-debug use of a register with a specific opcode
+  // and destination size.
+  auto GetOneUseWithOpCode =
```
andcarminati: nit: `GetOneUseWithOpcode`

llvm/lib/Target/AIE/AIE2PreLegalizerCombiner.cpp
```diff
@@ -288,6 +292,89 @@ bool AIE2PreLegalizerCombinerImpl::tryToCombineVectorInserts(
...
+    return std::nullopt;
+  };
+
+  if (auto Trunc =
+          GetOneUseWithOpCode(DstReg, TargetOpcode::G_TRUNC, SrcEltSize)) {
+    MachineInstr *TruncMI = Trunc.value();
+    const unsigned ExtOpcode =
+        SignVal ? TargetOpcode::G_SEXT : TargetOpcode::G_ZEXT;
+    const Register UseDstReg = TruncMI->getOperand(0).getReg();
+    if (auto Ext = GetOneUseWithOpCode(UseDstReg, ExtOpcode, 20)) {
```
andcarminati: nit: we can use:
```
    if (GetOneUseWithOpcode(UseDstReg, ExtOpcode, 20)) {
      return true;
    }
```

Because there is no further use for `auto Ext`.

Or even:

```
bool AIE2PreLegalizerCombinerImpl::canCombineVExtractElt(
    Register DstReg, bool SignVal, unsigned SrcEltSize) const {
  // Returns the single non-debug use of a register with a specific opcode
  // and destination size.
  auto GetOneUseWithOpcode =
      [&](const Register Reg, const unsigned OpcodeToCheck,
          const unsigned DstSize) -> std::optional<MachineInstr *> {
    if (MRI.hasOneNonDBGUser(Reg)) {
      MachineInstr &Use = *MRI.use_nodbg_instructions(Reg).begin();
      if (Use.getOpcode() == OpcodeToCheck) {
        const LLT DstRegTy = MRI.getType(Use.getOperand(0).getReg());
        if (DstRegTy.getSizeInBits() == DstSize)
          return &Use;
      }
    }
    return std::nullopt;
  };

  auto Trunc = GetOneUseWithOpcode(DstReg, TargetOpcode::G_TRUNC, SrcEltSize);

  if (!Trunc)
    return false;

  const MachineInstr *TruncMI = *Trunc;
  const unsigned ExtOpcode =
      SignVal ? TargetOpcode::G_SEXT : TargetOpcode::G_ZEXT;
  const Register UseDstReg = TruncMI->getOperand(0).getReg();

  return GetOneUseWithOpcode(UseDstReg, ExtOpcode, 20).has_value();
}
```

llvm/lib/Target/AIE/AIE2PreLegalizerCombiner.cpp
```diff
@@ -288,6 +292,86 @@ bool AIE2PreLegalizerCombinerImpl::tryToCombineVectorInserts(
   return true;
 }
 
+/// Determines if it is safe to combine vextract by checking the uses of DstReg,
+/// specfically for a pattern involving TRUNC followed by EXT.
```
andcarminati: nit: `specifically`

llvm/lib/Target/AIE/AIE2PreLegalizerCombiner.cpp
```diff
@@ -288,6 +292,86 @@ bool AIE2PreLegalizerCombinerImpl::tryToCombineVectorInserts(
...
+  const Register SignReg = MI.getOperand(4).getReg();
+
+  const auto SignVal = getIConstantVRegSExtVal(SignReg, MRI);
+  if (!SignVal)
+    return false;
+
+  const LLT SrcVecTy = MRI.getType(SrcReg);
+  const unsigned SrcEltSize = SrcVecTy.getScalarSizeInBits();
+  // Checks for the required pattern in uses of DstReg
+  if (!canCombineVExtractElt(DstReg, SignVal.value(), SrcEltSize))
```
andcarminati: I think we could change the name of this function. When I look to the name I expect something related to `vextract`, but in the end we are looking for an TRUC/EXT pattern related to a specific size. I think the name can be more generic and this function can be even used in other patterns. What do you think?

Something like:

```
isTruncExtToS20Sequence
```

llvm/test/CodeGen/RISCV/GlobalISel/jumptable.ll
```diff
@@ -192,7 +190,6 @@ define void @above_threshold(i32 signext %in, ptr %out) nounwind {
 ; RV64I-PIC-LABEL: above_threshold:
 ; RV64I-PIC:       # %bb.0: # %entry
 ; RV64I-PIC-NEXT:    li a2, 5
-; RV64I-PIC-NEXT:    sext.w a0, a0
```
gbossu, line 195: Nice! Did you check all targets?

llvm/lib/Target/AIE/AIE2PreLegalizerCombiner.cpp
```diff
@@ -288,6 +292,86 @@ bool AIE2PreLegalizerCombinerImpl::tryToCombineVectorInserts(
...
+/// \returns true if it is possible to combine the below sequence of MIRs
+/// From : %3:_(s32) = G_INTRINSIC
+///         intrinsic(@llvm.aie2.vextract.elem[8/16].I512), %2(<32 x s16>),
+///         %0(s32), %1(s32)
+///        %4:_(s16) = G_TRUNC %3(s32)
+///        %5:_(s20) = G_SEXT %4(s16)
+/// To :   %9:_(s20) = G_AIE_SEXT_EXTRACT_VECTOR_ELT %2(<32 x s16>), %0(s32)
+///        %10:_(s20) = G_ASSERT_[S/Z]EXT %9, 16
+///        %4:_(s16) = G_TRUNC %10(s20)
+///        %5:_(s20) = G_[S/Z]EXT %4(s16)
```
gbossu, line 339: Do we need to change the return types? I would expect that we only need to add a `%10:_(s32) = G_ASSERT_[S/Z]EXT %9, 16` and keep the rest intact thanks to the new `sext(trunc x)` combiner you added previously.

llvm/test/CodeGen/RISCV/GlobalISel/jumptable.ll
abhinay-anubola: Yes, I have checked all targets.

llvm/lib/Target/AIE/AIE2PreLegalizerCombiner.cpp
abhinay-anubola: Yes, we need to change the return types, because the pattern that is written in  new `sext(trunc x)` combiner will not match in this case as m_SpecificType is trying to match s20 but return type here is s32.
```mi_match(SrcReg, MRI, m_GTrunc(m_all_of(m_Reg(Reg), m_SpecificType(DstTy))))```

**Review by andcarminati — DISMISSED** [review]
LGTM.

File: llvm/test/CodeGen/AMDGPU/GlobalISel/combine-sext-trunc.mir [hunk 37]
```diff
@@ -0,0 +1,170 @@
...
+    ; GCN-NEXT: {{  $}}
+    ; GCN-NEXT: %var:_(s32) = COPY $vgpr0
+    ; GCN-NEXT: %assert:_(s32) = G_ASSERT_SEXT %var, 16
+    ; GCN-NEXT: %trunc:_(s16) = G_TRUNC %assert(s32)
+    ; GCN-NEXT: %sext:_(s64) = G_SEXT %trunc(s16)
+    ; GCN-NEXT: $vgpr0_vgpr1 = COPY %sext(s64)
+    %var:_(s32) = COPY $vgpr0
+    %assert:_(s32) = G_ASSERT_SEXT %var, 16
+    %trunc:_(s16) = G_TRUNC %assert(s32)
+    %sext:_(s64) = G_SEXT %trunc(s16)
```
konstantinschwarz, line 84: Nit: I guess we could still eliminate the `G_TRUNC`, and change `G_SEXT` to extend from `32 -> 64`?

llvm/lib/Target/AIE/aie2p/AIE2PInstrInfo.cpp
```diff
@@ -177,9 +177,10 @@ bool AIE2PInstrInfo::verifyGenericInstruction(const MachineInstr &MI,
         return false;
       }
     }
-    ErrInfo = "Expected 32/64bit scalar destination";
-    return MRI.getType(MI.getOperand(0).getReg()) == LLT::scalar(32) ||
-           MRI.getType(MI.getOperand(0).getReg()) == LLT::scalar(64);
+    ErrInfo = "Expected 20/32/64bit scalar destination";
```
konstantinschwarz: nit: we haven't changed anything for AIE2p, we still expect only 32-/64-bit destination types

llvm/lib/Target/AIE/AIE2PreLegalizerCombiner.cpp
```diff
@@ -243,6 +247,86 @@ bool AIE2PreLegalizerCombinerImpl::tryToCombineVectorInserts(
...
+      TII->getGenericExtractVectorEltOpcode(SignVal.value());
+  const unsigned AssertExtOpcode = SignVal.value()
+                                       ? TargetOpcode::G_ASSERT_SEXT
+                                       : TargetOpcode::G_ASSERT_ZEXT;
+  const unsigned ExtOpcode =
+      SignVal.value() ? TargetOpcode::G_SEXT : TargetOpcode::G_ZEXT;
+  const LLT S20 = LLT::scalar(20);
+  Register DstReg20Bit = MRI.createGenericVirtualRegister(S20);
+  Register ExtReg20Bit = MRI.createGenericVirtualRegister(S20);
+  MachineIRBuilder MIRBuilder(MI);
```
konstantinschwarz: The Combiner class already has a `MachineIRBuilder` member `B`, we should use it here to make us of potential CSE opportunities: initialize as `B.setInstrAndDebugLoc(MI);` and then use below

llvm/lib/Target/AIE/AIE2InstrInfo.cpp
```diff
@@ -159,8 +159,9 @@ bool AIE2InstrInfo::verifyGenericInstruction(const MachineInstr &MI,
         return false;
       }
     }
-    ErrInfo = "Expected 32bit scalar destination";
-    return MRI.getType(MI.getOperand(0).getReg()) == LLT::scalar(32);
+    ErrInfo = "Expected 20/32bit scalar destination";
```
konstantinschwarz, line 162: Can you please add instruction selection tests for `s20 = G_AIE_[SZ]EXT_EXTRACT_VECTOR_ELT` now that we made these new types legal?

llvm/test/CodeGen/AMDGPU/GlobalISel/combine-sext-trunc.mir
abhinay-anubola: This story requires a `(sext (trunc x)) → x` combiner; we can create a new ticket for the `G_TRUNC` elimination

llvm/lib/CodeGen/GlobalISel/CombinerHelper.cpp
```diff
@@ -2465,6 +2465,20 @@ bool CombinerHelper::matchCombineZextTrunc(MachineInstr &MI, Register &Reg) {
...
+  assert(MI.getOpcode() == TargetOpcode::G_SEXT && "Expected a G_SEXT");
+  const Register DstReg = MI.getOperand(0).getReg();
```
niwinanto, line 2470: nit: you could use `const auto [DstReg, DstTy, SrcReg, SrcTy] = MI.getFirst2RegLLTs()`

llvm/test/CodeGen/AMDGPU/GlobalISel/combine-sext-trunc.mir
konstantinschwarz: agreed 👍

konstantinschwarz (approved):

PR #235, opened 2024-11-13
fixup! [MachinePipeliner] No default SwpMaxStages

gbossu:
A tiny fixup because we forgot to update a PowerPC test when changing the default max stage count for MachinePipeliner

See https://github.com/Xilinx/llvm-aie/commit/806ee98491c5035cfd8a50a9982b8ee0f8978073


andcarminati (approved):
LGTM.

PR #236, opened 2024-11-14
Interleave loop pre-header with SWP prologue

gbossu:
Better review commit by commit, a lot of them are NFC and prepare for the final commit which enables interleaving.

QoR shows only improvements. Those will get bigger as we increase the scope of the post-pipeliner.

```
| Core_Compute_Cycle_Count       | ReduceProdAxis_1_aie2_bf16 | ReduceProdAxis_2_aie2_bf16 | ReduceProdAxis_4_aie2_bf16 | ReduceMeanAxis_2_aie2_bf16 | ReduceMeanAxis_4_aie2_bf16 | ReduceMeanAxis_1_aie2_bf16 | Sigmoid_bf16_0 | Requantize_0 | Softmax_bf16_0 | ReduceMeanAxis_3_aie2_bf16 | ReduceMeanAxis_6_aie2_bf16 | ReduceMeanAxis_5_aie2_bf16 | ReduceSumAxis_2_aie2_bf16 | Sigmoid_bf16_1 | Exp_bf16_0   | ReduceSumAxis_4_aie2_bf16 | ReduceSumAxis_1_aie2_bf16 | Log_bf16_0   | Neg_aie2_1   | EleMax_aie2_bfloat16 | EleMin_aie2_bfloat16 | ReduceMeanAxis_7_aie2_bf16 | Requantize_1 | MulAttributeBroadcasting_aie2_int8_0 | Tanh_1       | ReduceProdAxis_6_aie2_bf16 | ReduceSumAxis_5_aie2_bf16 | ReduceSumAxis_3_aie2_bf16 | ReduceSumAxis_6_aie2_bf16 | GELU_1       | HardSigmoid_bf16_0 | ReduceProdAxis_3_aie2_bf16 | ReduceProdAxis_5_aie2_bf16 | Exp_bf16_1   | ReduceProdAxis_7_aie2_bf16 | ReduceSumAxis_7_aie2_bf16 | Tanh_0       | GELU_0       | InstanceNormPart2_aie2_bf16_0 | MulBroadcasting_aie2_0 | HardSigmoid_bf16_1 | EleMax_aie2_int8 | EleMin_aie2_int8 | BatchNorm1d_aie2_bfloat16 | Mul_aie2_0   | HardSigmoid_int8_1 | HardSigmoid_int8_0 | BatchNorm1d_aie2_int8 | BitwiseNot_aie2_0 | BatchNorm2D_1 | GEMM_int8_0  | GEMM_int8_1  | Softmax_1    | BatchNorm2D_0 | Average diff |
| ------------------------------ | -------------------------- | -------------------------- | -------------------------- | -------------------------- | -------------------------- | -------------------------- | -------------- | ------------ | -------------- | -------------------------- | -------------------------- | -------------------------- | ------------------------- | -------------- | ------------ | ------------------------- | ------------------------- | ------------ | ------------ | -------------------- | -------------------- | -------------------------- | ------------ | ------------------------------------ | ------------ | -------------------------- | ------------------------- | ------------------------- | ------------------------- | ------------ | ------------------ | -------------------------- | -------------------------- | ------------ | -------------------------- | ------------------------- | ------------ | ------------ | ----------------------------- | ---------------------- | ------------------ | ---------------- | ---------------- | ------------------------- | ------------ | ------------------ | ------------------ | --------------------- | ----------------- | ------------- | ------------ | ------------ | ------------ | ------------- | ------------ |
| baseline                       | 35466                      | 17821                      | 35498                      | 13076                      | 13046                      | 13040                      | 2627           | 1421         | 8176           | 7233                       | 7219                       | 7212                       | 11900                     | 1727           | 7872         | 11920                     | 11884                     | 4149         | 455          | 227                  | 227                  | 6263                       | 781          | 517                                  | 2578         | 8709                       | 7055                      | 7052                      | 7038                      | 2811         | 937                | 8722                       | 8730                       | 1531         | 1794                       | 6212                      | 1970         | 2144         | 9536                          | 294                    | 649                | 164              | 164              | 390                       | 231          | 427                | 417                | 408                   | 135               | 416           | 2797         | 32931        | 570          | 308           | +0.00%       |
| this PR                        | 35405                      | 17789                      | 35433                      | 13051                      | 13016                      | 13010                      | 2620           | 1417         | 8144           | 7204                       | 7190                       | 7183                       | 11852                     | 1720           | 7840         | 11871                     | 11835                     | 4131         | 453          | 226                  | 226                  | 6231                       | 777          | 514                                  | 2560         | 8647                       | 7004                      | 7001                      | 6987                      | 2790         | 930                | 8656                       | 8662                       | 1519         | 1778                       | 6156                      | 1952         | 2123         | 9440                          | 291                    | 642                | 162              | 162              | 385                       | 228          | 421                | 411                | 402                   | 132               | 406           | 2701         | 31779        | 550          | 296           | -0.15%       |
| Total diff                     | IMPR(-0.17%)               | IMPR(-0.18%)               | IMPR(-0.18%)               | IMPR(-0.19%)               | IMPR(-0.23%)               | IMPR(-0.23%)               | IMPR(-0.27%)   | IMPR(-0.28%) | IMPR(-0.39%)   | IMPR(-0.40%)               | IMPR(-0.40%)               | IMPR(-0.40%)               | IMPR(-0.40%)              | IMPR(-0.41%)   | IMPR(-0.41%) | IMPR(-0.41%)              | IMPR(-0.41%)              | IMPR(-0.43%) | IMPR(-0.44%) | IMPR(-0.44%)         | IMPR(-0.44%)         | IMPR(-0.51%)               | IMPR(-0.51%) | IMPR(-0.58%)                         | IMPR(-0.70%) | IMPR(-0.71%)               | IMPR(-0.72%)              | IMPR(-0.72%)              | IMPR(-0.72%)              | IMPR(-0.75%) | IMPR(-0.75%)       | IMPR(-0.76%)               | IMPR(-0.78%)               | IMPR(-0.78%) | IMPR(-0.89%)               | IMPR(-0.90%)              | IMPR(-0.91%) | IMPR(-0.98%) | IMPR(-1.01%)                  | IMPR(-1.02%)           | IMPR(-1.08%)       | IMPR(-1.22%)     | IMPR(-1.22%)     | IMPR(-1.28%)              | IMPR(-1.30%) | IMPR(-1.41%)       | IMPR(-1.44%)       | IMPR(-1.47%)          | IMPR(-2.22%)      | IMPR(-2.40%)  | IMPR(-3.43%) | IMPR(-3.50%) | IMPR(-3.51%) | IMPR(-3.90%)  | -0.15%       |
```


llvm/lib/Target/AIE/Utils/AIELoopUtils.cpp
```diff
@@ -29,7 +37,11 @@ const MDNode *getLoopID(const MachineBasicBlock &LoopBlock) {
 }
 
 std::optional<int64_t> getMinTripCount(const MachineBasicBlock &LoopBlock) {
-  return getMinTripCount(getLoopID(LoopBlock));
+  std::optional<int64_t> MinTripCount = getMinTripCount(getLoopID(LoopBlock));
```
andcarminati, line 40: This makes a lot of sense ;-)

llvm/test/CodeGen/AIE/aie2/schedule/postpipeliner/interleave-prologue.mir
```diff
@@ -0,0 +1,422 @@
+# NOTE: Assertions have been autogenerated by utils/update_mir_test_checks.py
```
andcarminati, line 1: Clever! I like TDD. It also helps to see from where you start and how the work progresses with the sequence of commits.

llvm/lib/Target/AIE/AIEMaxLatencyFinder.cpp
```diff
@@ -221,17 +221,25 @@ unsigned MaxLatencyFinder::operator()(MachineInstr &MI) {
...
   for (ArrayRef<MachineBundle> TopBundles :
        getScheduledSuccessorInstructions(*CurBB, *Scheduler)) {
-    Earliest = findEarliestRef(MI, TopBundles, Earliest);
+    Earliest =
+        findEarliestRef(MI, TopBundles, Earliest + NumScheduledInstrsInRegion);
   }
 
   LLVM_DEBUG(dbgs() << "   Earliest=" << Earliest << "\n");
-  Latency = std::max(Latency - Earliest, 1);
+  Latency = std::max(Latency + NumScheduledInstrsInRegion - Earliest, 1);
```
gbossu: I'm not the biggest fan of this change in `AIEMaxLatencyFinder`. It would probably be better to extend the `EmitFixedSUnits` so that it also adds edges between "free" and "fixed" instructions.

llvm/lib/Target/AIE/AIEMaxLatencyFinder.cpp
```diff
@@ -131,6 +129,46 @@ int MaxLatencyFinder::findEarliestRef(
   return Cycle;
 }
 
+/// Whether the SUnits in the DAG have know successor instructions. Those
```
andcarminati: Know successor instructions of know scheduled successor instructions?

llvm/lib/Target/AIE/AIEMaxLatencyFinder.cpp
```diff
@@ -131,6 +129,46 @@ int MaxLatencyFinder::findEarliestRef(
...
+/// by the AIEPostPipeliner.
+bool hasScheduledSuccessorInstructions(MachineBasicBlock::iterator RegionEnd,
+                                       MachineBasicBlock &MBB,
+                                       const AIEPostRASchedStrategy &Sched) {
+  const BlockState &BS = Sched.getInterBlock().getBlockState(&MBB);
+  return isBottomRegion(RegionEnd, MBB) &&
+         (!BS.BottomInsert.empty() || Sched.successorsAreScheduled(&MBB));
+}
+
+/// Return the list of scheduled instructions that will follow those in the DAG.
```
andcarminati: Maybe DAG is out of context here (is a higher level artifact from the calling part).

llvm/lib/Target/AIE/AIEMaxLatencyFinder.cpp
```diff
@@ -131,6 +129,46 @@ int MaxLatencyFinder::findEarliestRef(
...
+  if (!BS.BottomInsert.empty()) {
+    return {BS.BottomInsert};
+  }
+
+  // Otherwise, look at the top region for the successor MBBs.
+  SmallVector<ArrayRef<AIE::MachineBundle>, 2> ScheduledRegs;
+  for (MachineBasicBlock *SuccBB : MBB.successors()) {
+    auto &SBS = Sched.getInterBlock().getBlockState(SuccBB);
+    assert(SBS.isScheduled());
+    if (SBS.getRegions().empty()) {
```
andcarminati: I am curious about how rare this can be. If this can be frequent, we can take the successor's successor to reduce the pessimism.

andcarminati:
Hi @gbossu, as we did before, we could integrate the 1st and 3rd commits before, as they are standalone changes.

gbossu:
> Hi @gbossu, as we did before, we could integrate the 1st and 3rd commits before, as they are standalone changes.

I'm actually tempted to drop the 3rd commit and do not have AIEMaxLatencyFinder/RegionEndEdges care about "timed"/"fixed" instructions. I think it makes sense if they keep focusing on successor MBBs instead. For dependencies between "fixed" and "free" instructions, I can add them in the `EmitFixedSUnits` mutator instead. Any opinion?

andcarminati:
> > Hi @gbossu, as we did before, we could integrate the 1st and 3rd commits before, as they are standalone changes.
> 
> I'm actually tempted to drop the 3rd commit and do not have AIEMaxLatencyFinder/RegionEndEdges care about "timed"/"fixed" instructions. I think it makes sense if they keep focusing on successor MBBs instead. For dependencies between "fixed" and "free" instructions, I can add them in the `EmitFixedSUnits` mutator instead. Any opinion?

Is this case, I think it makes sense, as we are introducing this new mutator! In this way we can also restrict a bit more the changes.

llvm/lib/Target/AIE/AIEInterBlockScheduling.cpp
```diff
@@ -712,10 +712,14 @@ void InterBlockScheduling::enterRegion(MachineBasicBlock *BB,
   auto &BS = getBlockState(BB);
   DEBUG_BLOCKS(dbgs() << "    >> enterRegion, Iter=" << BS.FixPoint.NumIters
                       << "\n");
+
+  // Only add regions of loops when in the GatheringRegions phase
   if (BS.Kind != BlockType::Loop ||
       BS.FixPoint.Stage == SchedulingStage::GatheringRegions) {
-    // Only add regions of loops when in the GatheringRegions phase
-    BS.addRegion(BB, RegionBegin, RegionEnd);
+    const int NumTopFixedInstructions = 0;
```
andcarminati: nit: the parameters are of type `size_t` (`unsigned`) and we are using `int` here.

llvm/lib/Target/AIE/AIEInterBlockScheduling.h
```diff
@@ -166,23 +166,61 @@ class Region {
...
+
+  /// Iterate over the instructions that are fixed at the top. Typically those
+  /// represent a SWP epilogue.
+  inline iterator_range<fixed_iterator> top_fixed_instrs() const {
+    fixed_iterator FixedEnd = std::next(BB->begin(), NumTopFixedInstructions);
+    return make_range(BB->begin(), FixedEnd);
+  }
+  size_t getNumTopFixedInstructions() const { return NumTopFixedInstructions; }
+
+  /// Iterate over the instructions that are fixed at the top. Typically those
```
andcarminati: nit: at the bot

llvm/lib/Target/AIE/AIEMachineScheduler.cpp
```diff
@@ -479,27 +480,72 @@ SUnit *AIEPostRASchedStrategy::pickNodeAndCycle(
...
+    return DAG->top().isValid() ? std::distance(DAG->begin(), DAG->top()) : 0;
+  return DAG->bottom().isValid() ? std::distance(DAG->bottom(), DAG->end()) : 0;
+}
+
+SUnit *AIEPostRASchedStrategy::getNextUnscheduledFixedInstr(
+    const SchedBoundary &Zone) const {
+  if (Zone.isTop())
+    return nullptr;
+  const Region &Reg = InterBlock.getBlockState(CurMBB).getCurrentRegion();
+  unsigned NumEmitted = getNumEmittedInstrs(DAG, /*IsTop=*/false);
```
andcarminati: nit: `const unsigned`

llvm/lib/CodeGen/ScheduleDAGInstrs.cpp
```diff
@@ -596,6 +596,10 @@ std::optional<unsigned> ScheduleDAGInstrs::initSUnit(MachineInstr &MI) {
   if (MI.isDebugOrPseudoInstr())
     return {};
 
+  // Ensure all instructions can be given a SUnit without re-allocation.
+  if (SUnits.empty())
+    SUnits.reserve(NumRegionInstrs);
```
martien-de-jong: Hmm. This proliferates NumRegionInstrs

llvm/lib/Target/AIE/AIEBaseSubtarget.cpp
```diff
@@ -543,11 +543,10 @@ class WAWEdges : public ScheduleDAGMutation {
     LiveRegs.init(*TRI);
     bool AddReservedRegs = true;
     if (Scheduler) {
+      assert(!Scheduler->doMBBSchedRegionsTopDown());
```
martien-de-jong, line 590: CHECK: this assert is specific for the prologue handling, and the code needs to be extended for epilogue handling.

llvm/lib/CodeGen/ScheduleDAGInstrs.cpp
```diff
@@ -596,6 +596,10 @@ std::optional<unsigned> ScheduleDAGInstrs::initSUnit(MachineInstr &MI) {
   if (MI.isDebugOrPseudoInstr())
     return {};
 
+  // Ensure all instructions can be given a SUnit without re-allocation.
```
andcarminati: I was thinking if we could do this in `ScheduleDAGInstrs::enterRegion` to remove this fixed cost from here. What do you think?

llvm/lib/Target/AIE/AIEBaseSubtarget.cpp
```diff
@@ -543,11 +543,10 @@ class WAWEdges : public ScheduleDAGMutation {
...
       MachineBasicBlock *MBB = DAG->getBB();
       const BlockState &BS = Scheduler->getInterBlock().getBlockState(MBB);
-      auto Region = BS.getCurrentRegion();
-      auto BottomRegion = BS.getBottom();
-      if (*Region.begin() == *BottomRegion.begin()) {
+      if (&BS.getCurrentRegion() == &BS.getBottom()) {
```
martien-de-jong, line 593: CHECK: this a relic of the pre-GatheringRegions era

llvm/lib/Target/AIE/AIEInterBlockScheduling.cpp
```diff
@@ -708,10 +712,13 @@ void InterBlockScheduling::enterRegion(MachineBasicBlock *BB,
...
   DEBUG_BLOCKS(dbgs() << "    >> enterRegion, Iter=" << BS.FixPoint.NumIters
                       << "\n");
+
+  // Only add regions of loops when in the GatheringRegions phase
   if (BS.Kind != BlockType::Loop ||
       BS.FixPoint.Stage == SchedulingStage::GatheringRegions) {
-    // Only add regions of loops when in the GatheringRegions phase
-    BS.addRegion(BB, RegionBegin, RegionEnd);
+    ArrayRef<MachineBundle> TopFixedBundles;
+    ArrayRef<MachineBundle> BotFixedBundles;
```
martien-de-jong: Uninitialized & dead?

llvm/lib/Target/AIE/AIEInterBlockScheduling.h
```diff
@@ -166,23 +166,77 @@ class Region {
   // The instruction that starts the next region, if any
   MachineInstr *ExitInstr = nullptr;
 
+  MachineBasicBlock *BB = nullptr;
+
+  /// Instructions that are already scheduled at the top.
```
martien-de-jong: add "e.g. an swp epilogue" ?

llvm/lib/Target/AIE/AIEInterBlockScheduling.h
```diff
@@ -166,23 +166,77 @@ class Region {
...
+      : BB(BB), TopFixedBundles(TopFixedBundles),
+        BotFixedBundles(BotFixedBundles) {
+    MachineBasicBlock::iterator FreeBegin =
+        std::next(Begin, TopFixedBundles.size());
+    MachineBasicBlock::iterator FreeEnd =
+        std::prev(End, BotFixedBundles.size());
+
+    // Verify that all fixed instructions are at the right place in the MBB
+    assert(TopFixedBundles.empty() || Begin == BB->begin());
+    assert(TopFixedBundles.empty() ||
```
martien-de-jong: I like it. && at top level in an assert should be split in different asserts.

llvm/lib/Target/AIE/AIEInterBlockScheduling.h
```diff
@@ -166,23 +166,77 @@ class Region {
...
+           }));
+    assert(BotFixedBundles.empty() || End == BB->end());
+    assert(BotFixedBundles.empty() ||
+           all_of(BotFixedBundles.front().Instrs,
+                  [FreeEnd](const MachineInstr *MI) {
+                    return getBundleStart(MI->getIterator()) == FreeEnd;
+                  }));
+
+    for (auto It = FreeBegin; It != FreeEnd; ++It) {
       SemanticOrder.push_back(&*It);
```
martien-de-jong: Maybe mention whether and where fixed instructions are added to the semantic order.

llvm/lib/Target/AIE/AIEHazardRecognizer.cpp
```diff
@@ -441,6 +461,12 @@ ScheduleHazardRecognizer::HazardType AIEHazardRecognizer::getHazardType(
     iterator_range<const MachineOperand *> MIOperands,
     const MachineRegisterInfo &MRI, int DeltaCycles) const {
   const unsigned SchedClass = TII->getSchedClass(Desc, MIOperands, MRI);
+  if (!IsPreRA && SchedClass == 0 &&
```
andcarminati, line 460: This verification could be refactored out to a helper function like `check...` or `validate...`, that can be used also below (same code).

llvm/lib/Target/AIE/AIEInterBlockScheduling.h
```diff
@@ -166,23 +166,77 @@ class Region {
...
+
+  /// Iterate over the instructions that are fixed at the top. Typically those
+  /// represent a SWP prologue.
+  inline iterator_range<fixed_iterator> bot_fixed_instrs() const {
+    fixed_iterator FixedBegin = std::prev(BB->end(), BotFixedBundles.size());
+    return make_range(FixedBegin, BB->end());
   }
-  size_t size() const { return SemanticOrder.size(); }
+  ArrayRef<MachineBundle> getBotFixedBundles() const { return BotFixedBundles; }
+  size_t getNumBotFixedInstructions() const { return BotFixedBundles.size(); }
```
martien-de-jong: nit: NumBotFixedBundles to be absolutely clear?

llvm/lib/Target/AIE/AIEMachineScheduler.cpp
```diff
@@ -1146,8 +1153,9 @@ void AIEScheduleDAGMI::recordDbgInstrs(const Region &CurrentRegion) {
   // We connect any Debug machine instruction to the instruction before it.
   // if there is no instruction before it, it is recorded in FirstDbgValue;
   MachineInstr *DbgMI = nullptr;
-  for (auto MII = CurrentRegion.end(), MIE = CurrentRegion.begin(); MII != MIE;
-       --MII) {
+  for (auto MII = CurrentRegion.free_instrs().end(),
+            MIE = CurrentRegion.free_instrs().begin();
+       MII != MIE; --MII) {
```
martien-de-jong: nit: for over reverse range?

llvm/lib/CodeGen/ScheduleDAGInstrs.cpp
gbossu: What do you mean? I'm simply moving the call from `initSUnits` to `initSUnit` so the latter is easier to use. I could also add a `reserverSUnits()` hook, but that seems a bit overkill.

llvm/lib/CodeGen/ScheduleDAGInstrs.cpp
gbossu: That would also be fine, the reason I kept it here is to avoid distributing the logic about creating `SUnits` in too many places. But I'm open to discussion :)

llvm/lib/Target/AIE/AIEHazardRecognizer.cpp
```diff
@@ -227,11 +228,14 @@ AIEHazardRecognizer::AIEHazardRecognizer(
...
 namespace llvm {
 void applyFormatOrdering(AIE::MachineBundle &Bundle, const VLIWFormat &Format,
-                         MachineInstr *BundleRoot,
                          MachineBasicBlock::iterator InsertPoint) {
+  assert(Bundle.SlotMap.size() == Bundle.Instrs.size() &&
+         "Bundle has instructions without slot");
   if (Bundle.empty())
     return;
 
+  MachineInstr *FirstMI = nullptr;
```
martien-de-jong: Nit: since this is a variant of the main loop move its declaration below the invariants?

llvm/lib/Target/AIE/AIEBaseSubtarget.cpp
gbossu: No, WAW handling could run on any region. Here I'm making sure that `BS.getBottom()` actually represents the bottom region.

llvm/lib/Target/AIE/AIEInterBlockScheduling.cpp
gbossu: Those represent empty arrays, once I want non-empty "fixed" regions, they will be referencing e.g. `BlockState::BottomInsert`

llvm/lib/Target/AIE/AIEMachineScheduler.cpp
gbossu: Yes I think I could do that in an NFC commit

llvm/lib/Target/AIE/AIEBaseSubtarget.cpp
```diff
@@ -318,6 +305,49 @@ class RegionEndEdges : public ScheduleDAGMutation {
   };
 };
 
+class EmitFixedSUnits : public ScheduleDAGMutation {
```
gbossu, line 315: I could probably add a top-level comment for that mutator

andcarminati, line 315: nit: for the future, we could have  a high-level description of the mutator.

llvm/lib/Target/AIE/AIEBaseSubtarget.cpp
```diff
@@ -318,6 +305,49 @@ class RegionEndEdges : public ScheduleDAGMutation {
...
+
+    // First, create SUnits for all "fixed" instructions
+    unsigned DistToExitSU = 0;
+    for (MachineInstr &MI : reverse(CurRegion.bot_fixed_instrs())) {
+      Scheduler->addFixedSUnit(MI, /*IsTop=*/false, DistToExitSU);
+      ++DistToExitSU;
+    }
+    DAG->makeMaps();
+
+    // Then, create create dependencies between "free" and "fixed" instructions
```
andcarminati: nit: `create create`.

llvm/lib/Target/AIE/AIEHazardRecognizer.cpp
```diff
@@ -243,21 +247,30 @@ void applyFormatOrdering(AIE::MachineBundle &Bundle, const VLIWFormat &Format,
...
-      MBB.insert(InsertPoint, Instr);
+    if (!Instr)
+      continue;
+
+    // If the instruction is bundled below a BUNDLE instruction, remove the
+    // latter as well. A new BUNDLE opcode will be created at the end.
+    if (auto BundleStart = getBundleStart(Instr->getIterator());
+        BundleStart->isBundle()) {
+      assert(BundleStart != Instr->getIterator());
+      BundleStart->eraseFromBundle();
```
martien-de-jong: I'm wondering when this occurs, and whether we should not remove  BUNDLEs as a precondition to calling this. Apart from that, I wonder whether there's anybody caring about the format order apart from the MClayer.

gbossu: - This happens when the prologue has VLIW instructions. They get materialized as `BUNDLE`. I could separate that "debundling" in a helper function if you want.
 - The ordering is mostly for stable ref updates at this point. Otherwise slight changes to the `tryCandidate` heuristics can be annoying, because they would change the order of instructions within bundles.

llvm/lib/Target/AIE/AIEInterBlockScheduling.cpp
```diff
@@ -767,16 +767,11 @@ void InterBlockScheduling::emitBundles(
...
-    bool InBundle = false;
     for (auto *MI : Bundle.getInstrs()) {
       if (Move) {
         BB->remove_instr(MI);
       }
       BB->insert(Before, MI);
-      if (InBundle) {
-        MI->bundleWithPred();
-      }
-      InBundle = true;
```
martien-de-jong, line 772: CHECK: bundling will now be done when processing the inserted region in its parent's block

llvm/lib/Target/AIE/AIEBaseSubtarget.cpp
```diff
@@ -664,6 +694,7 @@ AIEBaseSubtarget::getPostRAMutationsImpl(const Triple &TT) {
     Mutations.emplace_back(std::make_unique<MemoryEdges>());
     Mutations.emplace_back(std::make_unique<MachineSchedWAWEdges>());
     Mutations.emplace_back(std::make_unique<BiasDepth>());
+    Mutations.emplace_back(std::make_unique<EmitFixedSUnits>());
```
andcarminati, line 711: I was wondering if we can have some sort of missing deps here between the fixed and free  SUs, since we already executed the `MemoryEdges` mutator.

gbossu: We shouldn't, `AIE::maxLatency` should be very conservative with memory instructions. But I'll check :)

llvm/lib/Target/AIE/AIEMaxLatencyFinder.h
```diff
@@ -28,6 +28,16 @@ namespace llvm::AIE {
 int maxLatency(const MachineInstr *MI, const AIEBaseInstrInfo &InstrInfo,
                const InstrItineraryData &Itineraries, bool IncludeStages);
 
+struct InstrAndCycle {
+  MachineInstr *MI = nullptr;
+  int Cycle;
+};
+
+/// Find the first dependence on SrcMI in Bundles or Prune,
+/// whichever is less
```
martien-de-jong: But Prune isn't an InstrAndCycle,,,

llvm/lib/Target/AIE/AIEBaseSubtarget.cpp
gbossu: I checked, we correctly use `Latency = std::max(Latency, InstrInfo.getConservativeMemoryLatency(SrcClass))`

llvm/lib/Target/AIE/AIEMaxLatencyFinder.h
gbossu: Maybe:
```
/// Find the first dependence on SrcMI in Bundles[0,Prune)
/// \returns the Cycle in which the dependence happens or a conservative lower bound
///               and the instruction responsible for the dependency if it is found
```

llvm/lib/Target/AIE/AIEInterBlockScheduling.cpp
gbossu: This was just redundant. the `AIEHazardRecognizer::applyBundles()` call below does the bundling as well. (And finalizes them)

llvm/lib/Target/AIE/AIEInterBlockScheduling.h
gbossu: I'm actually thinking of deleting that function. One can just do `Region.getBotFixedBundles().size()`. Previously it was harder because I only had iterators, and `std::distance` isn't `O(1)`

llvm/lib/Target/AIE/AIEMachineScheduler.cpp
```diff
@@ -479,27 +480,78 @@ SUnit *AIEPostRASchedStrategy::pickNodeAndCycle(
...
+           SU.NodeNum < FirstBotFixedSU.value_or(DAG->SUnits.size());
+  }
+  return FirstBotFixedSU && SU.NodeNum >= *FirstBotFixedSU &&
+         SU.NodeNum <= LastBotFixedSU.value();
+}
+
+bool AIEPostRASchedStrategy::isFreeSU(const SUnit &SU) const {
+  const unsigned NumUpperBound = DAG->SUnits.size();
+  return SU.NodeNum < FirstTopFixedSU.value_or(NumUpperBound) &&
+         SU.NodeNum < FirstBotFixedSU.value_or(NumUpperBound);
```
martien-de-jong, line 532: Top and Bottom look too symmetric here. Is that intentional?

llvm/lib/CodeGen/ScheduleDAGInstrs.cpp
martien-de-jong: It was possible to add SUnits without any context, provided that you didn't use the MI to SU map. AFAIK, that was originally introduced to create the lcd edge DAG in loop aware, where the MI-> SU map can not be used because we have a one-to-many relation there. That may have NumRegionInstrs visible, but non sensibly initialized.

llvm/lib/Target/AIE/AIEMachineScheduler.cpp
```diff
@@ -1270,6 +1336,30 @@ void llvm::AIEPostRASchedStrategy::buildGraph(ScheduleDAGMI &DAG, AAResults *AA,
   static_cast<AIEScheduleDAGMI &>(DAG).recordDbgInstrs(Region);
 }
 
+void AIEPostRASchedStrategy::addFixedSUnit(MachineInstr &MI, bool IsTop,
+                                           unsigned Dist) {
+  assert(!(IsTop && FirstBotFixedSU) && "Top-fixed SUnits must be added first");
+  assert(DAG->SUnits.size() < DAG->SUnits.capacity() &&
+         "SUnits need to be re-allocated.");
+  unsigned SUNum = DAG->initSUnit(MI).value();
```
martien-de-jong, line 1347: nit: ignoring has_value() here.

llvm/lib/Target/AIE/AIEMachineScheduler.cpp
gbossu: Yes, all "fixed" SUnits are created after "free" SUnits, regardless of Top/Bot. See the header:
```
  /// SU numbers for fixed instructions.
  /// "top" fixed SUnits belong in [FirstTopFixedSU,FirstBotFixedSU)
  /// "bot" fixed SUnits belong in [FirstBotFixedSU,LastBotFixedSU]
  std::optional<unsigned> FirstTopFixedSU;
  std::optional<unsigned> FirstBotFixedSU;
  std::optional<unsigned> LastBotFixedSU;
```

llvm/lib/CodeGen/ScheduleDAGInstrs.cpp
gbossu: @martien-de-jong Would you find this better if I do the initialization in `ScheduleDAGInstrs::enterRegion` then?

llvm/lib/Target/AIE/AIEMachineScheduler.cpp
```diff
@@ -476,27 +480,78 @@ SUnit *AIEPostRASchedStrategy::pickNodeAndCycle(
...
+  return nullptr;
+}
+
+bool AIEPostRASchedStrategy::isFixedSU(const SUnit &SU, bool IsTop) const {
+  if (IsTop) {
+    return FirstTopFixedSU && SU.NodeNum >= *FirstTopFixedSU &&
+           SU.NodeNum < FirstBotFixedSU.value_or(DAG->SUnits.size());
+  }
+  return FirstBotFixedSU && SU.NodeNum >= *FirstBotFixedSU &&
+         SU.NodeNum <= LastBotFixedSU.value();
```
andcarminati, line 526: nit: `SU.NodeNum <= *LastBotFixedSU` (to be symmetric with the previous case) or `FirstBotFixedSU.value()`.

gbossu: This is on purpose to have dynamic checks on `LastBotFixedSU` (and have a runtime failure if there is no value). For `FirstBotFixedSU`, I previously checked `has_value()`, so there is no need for checking again.

llvm/lib/Target/AIE/AIEMachineScheduler.cpp
gbossu: That would be an error if the value is missing

llvm/lib/CodeGen/ScheduleDAGInstrs.cpp
gbossu: Discussed offline, I will reserve capacity in `enterRegion`, see https://github.com/Xilinx/llvm-aie/pull/236#discussion_r1846399685

llvm/lib/Target/AIE/AIEBaseSubtarget.cpp
gbossu: Not really, the DAGMutator can run for all types of regions. For non-loop blocks, we do not pre-gather regions.

llvm/lib/CodeGen/ScheduleDAGInstrs.cpp
martien-de-jong: I guess so. I was happy that initSUnit didn't need to care. At the time I hadn't thought about adding SUnits after edge creation. If we can pre-reserve the total number of sunits with some interface I'm fine.

llvm/lib/Target/AIE/AIEBaseSubtarget.cpp
martien-de-jong: Hmm. now you have me worried about reallocation of the region array. pre-gather would imply pre-allocation, making the test 'obviously correct'

llvm/lib/Target/AIE/AIEInterBlockScheduling.h
martien-de-jong: ok. They aren't. And I know why.

llvm/lib/Target/AIE/AIEHazardRecognizer.cpp
martien-de-jong: Ok. ref order is a good reason. Maybe my main confusion is that this function should live somewhere else.

llvm/lib/Target/AIE/AIEMachineScheduler.cpp
martien-de-jong: and we don't like asserts? :-b

gbossu: I mean that I would get to the same behaviour: the error would be caught and not silently ignored.

llvm/lib/Target/AIE/AIEInterBlockScheduling.h
```diff
@@ -166,23 +166,46 @@ class Region {
   // The instruction that starts the next region, if any
   MachineInstr *ExitInstr = nullptr;
 
+  MachineBasicBlock *BB = nullptr;
+
+  /// Instructions that are already scheduled at the top, e.g. an swp epilogue.
+  /// Those should not be re-ordered by the scheduler.
+  ArrayRef<MachineBundle> TopFixedBundles;
+
+  /// Instructions that are already scheduled at the bottom, e.g. an swp
```
andcarminati, line 175: nit: a swp prologue.

llvm/lib/Target/AIE/AIEHazardRecognizer.cpp
gbossu: It should. But I blame AIE1 ;)

llvm/lib/Target/AIE/AIEInterBlockScheduling.cpp
gbossu: This is gone now :)

gbossu:
Any more comments @andcarminati @martien-de-jong ?

llvm/lib/Target/AIE/AIEBaseSubtarget.cpp
```diff
@@ -275,27 +275,14 @@ class RegionEndEdges : public ScheduleDAGMutation {
...
-          auto *Scheduler =
-              static_cast<AIEScheduleDAGMI *>(DAG)->getSchedImpl();
-          auto &InterBlock = Scheduler->getInterBlock();
-          unsigned InsertedCycles =
-              InterBlock.getBlockState(DAG->getBB()).BottomInsert.size();
-          PatchCycles =
-              PatchCycles >= InsertedCycles ? PatchCycles - InsertedCycles : 0;
-        }
-        EdgeLatency = std::max(EdgeLatency, PatchCycles);
+        const unsigned ZOLDistance = 8;
```
andcarminati, line 284: For the future: this magic 8 could be retrieved from TII as well.

llvm/lib/Target/AIE/AIEBaseSubtarget.cpp
```diff
@@ -318,6 +305,56 @@ class RegionEndEdges : public ScheduleDAGMutation {
...
+
+    // Then, create dependencies between "free" and "fixed" instructions
+    auto IsFreeSU = [Scheduler](const SUnit &SU) {
+      return Scheduler->isFreeSU(SU);
+    };
+    ArrayRef<AIE::MachineBundle> BotFixedBundles =
+        CurRegion.getBotFixedBundles();
+    for (SUnit &FreeSU : make_filter_range(DAG->SUnits, IsFreeSU)) {
+      const MachineInstr &MI = *FreeSU.getInstr();
+      MachineInstr *FixedDepMI =
```
andcarminati, line 349: super nit:

To help readability, we could extend with overloaded cast operators:

```
 struct InstrAndCycle {
   MachineInstr *MI = nullptr;
   int Cycle;
+  operator MachineInstr *() const { return MI; }
+  operator int() const { return Cycle; }
 };
```
Then we can just use:

```
MachineInstr *FixedDepMI = AIE::findEarliestRef(MI, BotFixedBundles, BotFixedBundles.size());
```

And also:

```
Earliest = findEarliestRef(MI, TopBundles, Earliest);
```

llvm/lib/Target/AIE/AIEBundle.h
```diff
@@ -103,6 +105,16 @@ template <class I> class Bundle {
       MetaInstrs.push_back(Instr);
       return;
     }
+
+    // Keep track of BUNDLE instructions. They need to be cleaned up when
+    // de-bundling before re-bundling. See applyBundles()
+    if (Instr->getOpcode() == TargetOpcode::BUNDLE) {
```
andcarminati, line 111: super nit: `Instr->isBundle()`

llvm/lib/Target/AIE/AIEMachineScheduler.cpp
```diff
@@ -1250,6 +1333,33 @@ void llvm::AIEPostRASchedStrategy::buildGraph(ScheduleDAGMI &DAG, AAResults *AA,
...
+  DEBUG_BLOCKS(dbgs() << "Adding Fixed MI: " << MI);
+  DEBUG_BLOCKS(dbgs() << "  DAG size=" << DAG->SUnits.size()
+                      << " capacity=" << DAG->SUnits.capacity() << "\n");
+  assert(!(IsTop && FirstBotFixedSU) && "Top-fixed SUnits must be added first");
+  assert(DAG->SUnits.size() < DAG->SUnits.capacity() &&
+         "SUnits need to be re-allocated.");
+  unsigned SUNum = DAG->initSUnit(MI).value();
+  SUnit &SU = DAG->SUnits[SUNum];
+
+  if (IsTop) {
```
andcarminati, line 1350: It is really smart that you are preparing the way for epilogue merging (in some small details) ;-)

andcarminati:
Hi @gbossu, I finished my review and left few suggestions. Nice piece of engineering that will make a lot more difference for the post-swp as we increase its use.

llvm/lib/Target/AIE/AIEBundle.h
gbossu: Oh I remember now, the reason I chose `getOpcode() == ...` is because `AIE::Bundle` is a template class that should work with both `MachineInstr` and `MCInstr`. The latter does not provide `::isBundle()`

llvm/lib/CodeGen/ScheduleDAGInstrs.cpp
```diff
@@ -633,8 +637,6 @@ std::optional<unsigned> ScheduleDAGInstrs::initSUnit(MachineInstr &MI) {
 
 void ScheduleDAGInstrs::initSUnits() {
   ScheduleDAG::clearDAG();
```
martien-de-jong, line 639: nit: this looks at odds with the clearDAG in `ScheduleDAGInstrs::enterRegion`

llvm/lib/Target/AIE/AIEBaseSubtarget.cpp
martien-de-jong: Perhaps move it to `getBottom()` then?

llvm/lib/Target/AIE/AIEInterBlockScheduling.cpp
```diff
@@ -970,6 +979,40 @@ bool InterBlockEdges::isPostBoundaryNode(SUnit *SU) const {
...
+
+  // Verify that all fixed instructions are at the right place in the MBB
+  assert(TopFixedBundles.empty() || Begin == BB->begin());
+  assert(TopFixedBundles.empty() ||
+         all_of(TopFixedBundles.back().Instrs, [FreeBegin](
+                                                   const MachineInstr *MI) {
+           return getBundleStart(MI->getIterator()) == std::prev(FreeBegin);
+         }));
+  assert(BotFixedBundles.empty() || End == BB->end());
+  assert(
```
martien-de-jong, line 1002: Strange, the formatting diff relative to TopFixedBundles.

llvm/lib/Target/AIE/AIEInterBlockScheduling.h
martien-de-jong: I agree with `an es-double-u-pee`

llvm/lib/CodeGen/ScheduleDAGInstrs.cpp
gbossu: I could remove that one now, I just thought it does not hurt. But I agree it does look strange

llvm/lib/Target/AIE/AIEInterBlockScheduling.h
```diff
@@ -166,23 +166,46 @@ class Region {
...
+  inline iterator_range<fixed_iterator> top_fixed_instrs() const {
+    fixed_iterator FixedEnd = std::next(BB->begin(), TopFixedBundles.size());
+    return make_range(BB->begin(), FixedEnd);
   }
-  std::vector<MachineInstr *>::const_iterator end() const {
-    return SemanticOrder.end();
+  ArrayRef<MachineBundle> getTopFixedBundles() const { return TopFixedBundles; }
+
+  /// Iterate over the instructions that are fixed at the bottom. Typically
+  /// those represent a SWP prologue.
```
martien-de-jong, line 202: nit: *an SWP* ;-)

llvm/lib/Target/AIE/AIEMachineScheduler.cpp
```diff
@@ -544,6 +544,13 @@ void AIEPostRASchedStrategy::enterMBB(MachineBasicBlock *MBB) {
 void AIEPostRASchedStrategy::commitBlockSchedule(MachineBasicBlock *BB) {
   auto &BS = InterBlock.getBlockState(BB);
 
+  // TODO: Update assert when the fixed instructions become part of the
+  // scheduling region.
+  assert(BS.getRegions().empty() ||
+         0 == BS.getTop().getTopFixedBundles().size());
```
martien-de-jong, line 613: There's no empty() for an ArrayRef?

llvm/lib/Target/AIE/AIEMachineScheduler.cpp
```diff
@@ -1146,16 +1153,14 @@ void AIEScheduleDAGMI::recordDbgInstrs(const Region &CurrentRegion) {
   // We connect any Debug machine instruction to the instruction before it.
   // if there is no instruction before it, it is recorded in FirstDbgValue;
   MachineInstr *DbgMI = nullptr;
-  for (auto MII = CurrentRegion.end(), MIE = CurrentRegion.begin(); MII != MIE;
-       --MII) {
-    MachineInstr *Prev = *std::prev(MII);
+  for (MachineInstr *MI : reverse(CurrentRegion.getFreeInstructions())) {
     if (DbgMI) {
-      DbgValues.emplace_back(DbgMI, Prev);
+      DbgValues.emplace_back(DbgMI, MI);
```
martien-de-jong, line 1233: This looks like a semantic change. Was it wrong?

llvm/lib/Target/AIE/AIEHazardRecognizer.cpp
```diff
@@ -227,8 +228,9 @@ AIEHazardRecognizer::AIEHazardRecognizer(
 
 namespace llvm {
 void applyFormatOrdering(AIE::MachineBundle &Bundle, const VLIWFormat &Format,
-                         MachineInstr *BundleRoot,
                          MachineBasicBlock::iterator InsertPoint) {
+  assert(Bundle.SlotMap.size() == Bundle.Instrs.size() &&
+         "Bundle has instructions without slot");
```
martien-de-jong, line 233: or slots without instructions....

llvm/lib/Target/AIE/AIEBaseSubtarget.cpp
```diff
@@ -318,6 +305,56 @@ class RegionEndEdges : public ScheduleDAGMutation {
...
+  void apply(ScheduleDAGInstrs *DAG) override {
+    AIEPostRASchedStrategy *Scheduler =
+        static_cast<AIEScheduleDAGMI *>(DAG)->getSchedImpl();
+    auto *TII = static_cast<const AIEBaseInstrInfo *>(DAG->TII);
+    auto *ItinData = DAG->MF.getSubtarget().getInstrItineraryData();
+    const BlockState &BS =
+        Scheduler->getInterBlock().getBlockState(DAG->getBB());
+    const Region &CurRegion = BS.getCurrentRegion();
+
+    // First, create SUnits for all "fixed" instructions
```
martien-de-jong, line 326: Perhaps mention that DistToExitSU counts cycles?

llvm/lib/Target/AIE/AIEMachineScheduler.cpp
gbossu: This is implementing your suggestion at https://github.com/Xilinx/llvm-aie/pull/236#discussion_r1846513192 :) No semantic change

llvm/lib/Target/AIE/AIEBaseSubtarget.cpp
```diff
@@ -318,6 +305,56 @@ class RegionEndEdges : public ScheduleDAGMutation {
...
+    unsigned DistToExitSU = 0;
```
martien-de-jong: I think this loop deserves more comment, e.g. what DistToExitSU means, whether we iterate over bundles or instructions, etc.

llvm/lib/Target/AIE/AIEMachineScheduler.cpp
gbossu: Hmm, there is. I guess I wanted to leave the number `0` so I can easily replace it with e.g. `BS.BottomInsert.size()` in a future commit.

llvm/lib/Target/AIE/AIEMachineScheduler.cpp
```diff
@@ -479,27 +480,78 @@ SUnit *AIEPostRASchedStrategy::pickNodeAndCycle(
...
-  if (Zone.getCurrCycle() >= RegionBottomUpCycles - 1)
+  // Top-down scheduling does not support DeltaCycles
+  if (Zone.isTop() || Zone.getCurrCycle() >= RegionBottomUpCycles - 1)
     return 0;
   return std::min({int(RegionBottomUpCycles - 1 - Zone.getCurrCycle()),
                    int(getAIEHazardRecognizer(Zone)->getMaxLookAhead()),
                    BottomUpDelta.getValue()});
 }
 
+unsigned getNumEmittedInstrs(ScheduleDAGMI *DAG, bool IsTop) {
```
martien-de-jong, line 492: I could do with some description. I also hope that std::distance is cheap here.

llvm/lib/Target/AIE/AIEMachineScheduler.cpp
```diff
@@ -479,27 +480,78 @@ SUnit *AIEPostRASchedStrategy::pickNodeAndCycle(
...
+  if (IsTop)
+    return DAG->top().isValid() ? std::distance(DAG->begin(), DAG->top()) : 0;
+  return DAG->bottom().isValid() ? std::distance(DAG->bottom(), DAG->end()) : 0;
+}
+
+SUnit *AIEPostRASchedStrategy::getNextUnscheduledFixedInstr(
+    const SchedBoundary &Zone) const {
+  if (Zone.isTop())
+    return nullptr;
+  const Region &Reg = InterBlock.getBlockState(CurMBB).getCurrentRegion();
```
martien-de-jong, line 502: Please describe what happens here. I'm wondering why we don't run across the SUnits by index. We know where they are, in what order they should be scheduled and what their scheduled cycle should be.

llvm/lib/Target/AIE/AIEMachineScheduler.cpp
```diff
@@ -479,27 +480,78 @@ SUnit *AIEPostRASchedStrategy::pickNodeAndCycle(
...
+
 bool AIEPostRASchedStrategy::isAvailableNode(SUnit &SU, SchedBoundary &Zone,
                                              bool /*VerifyReadyCycle*/) {
+  // Note we use signed integers to avoid wrap-around behavior.
+  const int MinDelta = -getMaxDeltaCycles(Zone);
+  const int ReadyCycle = std::max(Zone.getCurrCycle(), SU.BotReadyCycle);
+  const int CurrCycle = Zone.getCurrCycle();
+
+  // If the Zone has remaining fixed instructions, only one SU is available.
+  if (SUnit *FixedSU = getNextUnscheduledFixedInstr(Zone)) {
```
martien-de-jong, line 543: This could be called for multiple SU, right? That sounds pretty expensive. In particular, any instruction in the currently fixed bundle complies, not just the next one.

martien-de-jong: Also, we just check availabilty of SU. The comment "only one SU is available" is not enforced here, it just says that *this* SU is available.

martien-de-jong: How can we guarantee that isAvailableNode is, and will in future be called for the fixed instructions? I have a hunch that that hinges on the fact that we have no 'normal' dependences blocking them. But that may also mean that we will be calling getNextUnscheduledFixedInstr O(N**2) times to find the first in N candidates N times.

llvm/lib/Target/AIE/AIEMachineScheduler.cpp
```diff
@@ -1237,6 +1298,11 @@ void llvm::AIEPostRASchedStrategy::buildGraph(ScheduleDAGMI &DAG, AAResults *AA,
                                               PressureDiffs *PDiffs,
                                               LiveIntervals *LIS,
                                               bool TrackLaneMasks) {
+
+  // Let's save the DAG already instead of waiting for initialize().
+  // Some DAG mutators might require a DAG to be set.
+  this->DAG = &DAG;
```
martien-de-jong, line 1308: Yes indeed!

llvm/lib/Target/AIE/AIEBaseSubtarget.cpp
```diff
@@ -318,6 +305,56 @@ class RegionEndEdges : public ScheduleDAGMutation {
   };
 };
 
+/// This Mutator is responsible for emitting "fixed" SUnits at the top or bottom
+/// of the region. These special SUnits require a specific cycle and cannot be
+/// placed freely by the scheduler.
+///
+/// Here, these special SUnits get created from Region::top_fixed_instrs() or
+/// Region::bot_fixed_instrs(), and dependencies are created between "free" and
+/// "fixed" SUnits.
```
martien-de-jong, line 314: Perhaps spend some words on the mechanism. As I understand it, we only add edges to ExitSU with the right latency, and we make sure that  pickOnlyChoice will not return Free nodes before exhausting the Fixed ones.

llvm/lib/Target/AIE/AIEBaseSubtarget.cpp
martien-de-jong: Probably clear enough with some more narrative at the top.

llvm/lib/Target/AIE/AIEMachineScheduler.cpp
gbossu: Correct, it will be called for every pending instruction. But i don't think `getNextUnscheduledFixedInstr` is very expensive. Would you like me to cache it?

llvm/lib/Target/AIE/AIEMachineScheduler.cpp
gbossu: `/// Returns the number of emitted instructions in the Top or Bot zone`?

llvm/lib/Target/AIE/AIEMachineScheduler.cpp
```diff
@@ -476,27 +480,78 @@ SUnit *AIEPostRASchedStrategy::pickNodeAndCycle(
...
                    BottomUpDelta.getValue()});
 }
 
+unsigned getNumEmittedInstrs(ScheduleDAGMI *DAG, bool IsTop) {
+  if (IsTop)
+    return DAG->top().isValid() ? std::distance(DAG->begin(), DAG->top()) : 0;
+  return DAG->bottom().isValid() ? std::distance(DAG->bottom(), DAG->end()) : 0;
+}
+
+SUnit *AIEPostRASchedStrategy::getNextUnscheduledFixedInstr(
```
andcarminati, line 498: super nit: `getNextUnscheduledFixedSU`

llvm/lib/Target/AIE/AIEMachineScheduler.cpp
gbossu: I think that the easiest way for the fixed instructions to be interleaved is to have them scheduled as "normally" as possible. That means they get dependencies to other SUnits, and are part of the normal pending/available queues. This is appealing because "free" SUnits can be interleaved seamlessly. The only real place that I have to tweak is here: to determine what goes in the available queue.

llvm/lib/Target/AIE/AIEMachineScheduler.cpp
martien-de-jong: I'm just scared of O(N**2) if N might be ~100. That would be e.g. 16 cycles of 6 instructions. 
I'd leave it for now, but keep it in mind if we start spending serious compile time.

llvm/lib/Target/AIE/AIEMachineScheduler.cpp
martien-de-jong: No worries, I agree on that. I was just thinking that the next SU could be easily found by stepping an iterator across the fixed unit each time we a get a call back that a node is scheduled. Alternatively, that we make all the nodes whose time is due available in one go and have a very simple priority function. I don't think that single choice call adds much benefit if the pending queue is overloaded with all the fixed candidates.

martien-de-jong (dismissed):

llvm/lib/Target/AIE/AIEMachineScheduler.cpp
gbossu: I'm now chaining the fixed `SUnits` together instead of having them all connected to `ExitSU`. That should solve most of the complexity problems :)

martien-de-jong (approved):

PR #238, opened 2024-11-18
Fix G_TRUNC assert in S20Narrowing

abhinay-anubola:
The assert is causing a crash in the code. To prevent this, disabling the S20 narrowing by returning false when the corresponding case occurs.

### DIFF
--- a/llvm/lib/Target/AIE/AIECombinerHelper.cpp
+++ b/llvm/lib/Target/AIE/AIECombinerHelper.cpp
@@ -627,11 +627,10 @@ bool canNarrowUserTreeToS20(MachineRegisterInfo &MRI, InstrNode Start,
     }
     switch (Use.getOpcode()) {
     case TargetOpcode::G_TRUNC:
-      // Sanity check that we are not truncating into less than 20 bits and
-      // losing precision. If this happens this means we missed an extension
-      // from that small type back to S20 to feed into our ptr.add intrinsics.
-      assert(MRI.getType(Use.getOperand(0).getReg()).getScalarSizeInBits() >=
-             20);
+      // Check if the scalar size of the operand's type is at least 20 bits,
+      // this ensures that the G_TRUNC can be safely converted to a COPY.
+      if (MRI.getType(Use.getOperand(0).getReg()).getScalarSizeInBits() < 20)
+        return false;
       [[fallthrough]];
     case TargetOpcode::G_PTR_ADD:
     case TargetOpcode::G_STORE: // Data operand is later modified to S20 type
--- a/llvm/test/CodeGen/AIE/aie2/GlobalISel/prelegalizercombiner-s20-narrowing.mir
+++ b/llvm/test/CodeGen/AIE/aie2/GlobalISel/prelegalizercombiner-s20-narrowing.mir
@@ -781,3 +781,92 @@ body:             |
     $r0 = COPY %11
     G_BR %bb.2
 ...
+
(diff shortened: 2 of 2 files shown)


llvm/lib/Target/AIE/AIECombinerHelper.cpp
```diff
@@ -630,9 +630,8 @@ bool canNarrowUserTreeToS20(MachineRegisterInfo &MRI, InstrNode Start,
       // Sanity check that we are not truncating into less than 20 bits and
       // losing precision. If this happens this means we missed an extension
       // from that small type back to S20 to feed into our ptr.add intrinsics.
-      assert(MRI.getType(Use.getOperand(0).getReg()).getScalarSizeInBits() >=
-             20);
-      [[fallthrough]];
```
gbossu, line 635: I'd be happy if you add a .mir test that shows a G_TRUNC that is not to `s20`

llvm/lib/Target/AIE/AIECombinerHelper.cpp
```diff
@@ -630,9 +630,8 @@ bool canNarrowUserTreeToS20(MachineRegisterInfo &MRI, InstrNode Start,
       // Sanity check that we are not truncating into less than 20 bits and
       // losing precision. If this happens this means we missed an extension
       // from that small type back to S20 to feed into our ptr.add intrinsics.
```
gbossu: The comment also needs to be updated. This is not a sanity check, this is simply checking that the `G_TRUNC` can easily be converted to a `COPY`

llvm/lib/Target/AIE/AIECombinerHelper.cpp
```diff
@@ -627,12 +627,10 @@ bool canNarrowUserTreeToS20(MachineRegisterInfo &MRI, InstrNode Start,
...
-      // Sanity check that we are not truncating into less than 20 bits and
-      // losing precision. If this happens this means we missed an extension
-      // from that small type back to S20 to feed into our ptr.add intrinsics.
-      assert(MRI.getType(Use.getOperand(0).getReg()).getScalarSizeInBits() >=
-             20);
-      [[fallthrough]];
+      // Check if the scalar size of the operand's type is at least 20 bits,
+      // this ensures that the G_TRUNC can be safely converted to a COPY.
+      if (MRI.getType(Use.getOperand(0).getReg()).getScalarSizeInBits() < 20)
+        return false;
```
andcarminati, line 633: We have a warning now:
```
../llvm/lib/Target/AIE/AIECombinerHelper.cpp:634:5: warning: unannotated fall-through between switch labels [-Wimplicit-fallthrough]
    case TargetOpcode::G_PTR_ADD:
    ^
../llvm/lib/Target/AIE/AIECombinerHelper.cpp:634:5: note: insert 'LLVM_FALLTHROUGH;' to silence this warning
    case TargetOpcode::G_PTR_ADD:
    ^
    LLVM_FALLTHROUGH; 
../llvm/lib/Target/AIE/AIECombinerHelper.cpp:634:5: note: insert 'break;' to avoid fall-through
    case TargetOpcode::G_PTR_ADD:
    ^
    break; 
```

abhinay-anubola: Adding `[[fallthrough]];`

gbossu (approved):
LGTM, thanks for investigating this issue.

andcarminati (approved):

PR #239, opened 2024-11-19
[AIE2] Fix VMOV instruction itinerary

krishnamtibrewala:
Changes done in this PR were identified by @konstantinschwarz for fixing the DIFF issue in FullyConnect_aie2_bf16

ISSUE : In a specific case the compiler was assuming that VMOV has a latency of 2, i.e. it can still use the old value of $wlX when written from $qX reg, but since the DST reg was of $wlX type which uses the bypass path and overwrites the value one cycle earlier, leading to wrong value being read.

### DIFF
--- a/llvm/lib/Target/AIE/AIE2GenFixupInstrInfo.td
+++ b/llvm/lib/Target/AIE/AIE2GenFixupInstrInfo.td
@@ -762,7 +762,9 @@ let Itinerary = II_VMOV_W in {
                            ItinRegClassPair<II_VMOV_W_WMH_WML,[OperandRegClass<0, eWH>, OperandRegClass<1, eWL>]>,						   
                            ItinRegClassPair<II_VMOV_W_WML_WMH,[OperandRegClass<0, eWL>, OperandRegClass<1, eWH>]>,
                            ItinRegClassPair<II_VMOV_W_WML_Q, [OperandRegClass<0, eWL>, OperandRegClass<1, mQQm>]>,
-                           ItinRegClassPair<II_VMOV_W_WMH_Q, [OperandRegClass<0, eWH>, OperandRegClass<1, mQQm>]>] in {	
+                           ItinRegClassPair<II_VMOV_W_WMH_Q, [OperandRegClass<0, eWH>, OperandRegClass<1, mQQm>]>,
+                           ItinRegClassPair<II_VMOV_W_Q_WML, [OperandRegClass<0, mQQm>, OperandRegClass<1, eWL>]>,
+                           ItinRegClassPair<II_VMOV_W_Q_WMH, [OperandRegClass<0, mQQm>, OperandRegClass<1, eWH>]>] in {	
 	def VMOV_mv_w : AIE2_mv_w_inst_mv< (outs OP_mMvAMWQDst:$dst), (ins OP_mMvAMWQSrc:$src),
 		"vmov", "$dst, $src">;
   }
--- a/llvm/lib/Target/AIE/AIE2GenFixupInstrInfo.td
+++ b/llvm/lib/Target/AIE/AIE2GenFixupInstrInfo.td
@@ -760,7 +760,9 @@ let Itinerary = II_VMOV_W in {
                            ItinRegClassPair<II_VMOV_W_WMH_WMH,[OperandRegClass<0, eWH>, OperandRegClass<1, eWH>]>,
                            ItinRegClassPair<II_VMOV_W_WML_WML,[OperandRegClass<0, eWL>, OperandRegClass<1, eWL>]>,
                            ItinRegClassPair<II_VMOV_W_WMH_WML,[OperandRegClass<0, eWH>, OperandRegClass<1, eWL>]>,						   
-                           ItinRegClassPair<II_VMOV_W_WML_WMH,[OperandRegClass<0, eWL>, OperandRegClass<1, eWH>]>] in {	
+                           ItinRegClassPair<II_VMOV_W_WML_WMH,[OperandRegClass<0, eWL>, OperandRegClass<1, eWH>]>,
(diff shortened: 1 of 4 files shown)


llvm/lib/Target/AIE/AIE2GenFixupInstrInfo.td
```diff
@@ -760,7 +760,9 @@ let Itinerary = II_VMOV_W in {
                            ItinRegClassPair<II_VMOV_W_WMH_WMH,[OperandRegClass<0, eWH>, OperandRegClass<1, eWH>]>,
                            ItinRegClassPair<II_VMOV_W_WML_WML,[OperandRegClass<0, eWL>, OperandRegClass<1, eWL>]>,
                            ItinRegClassPair<II_VMOV_W_WMH_WML,[OperandRegClass<0, eWH>, OperandRegClass<1, eWL>]>,						   
-                           ItinRegClassPair<II_VMOV_W_WML_WMH,[OperandRegClass<0, eWL>, OperandRegClass<1, eWH>]>] in {	
+                           ItinRegClassPair<II_VMOV_W_WML_WMH,[OperandRegClass<0, eWL>, OperandRegClass<1, eWH>]>,
+                           ItinRegClassPair<II_VMOV_W_WML_Q, [OperandRegClass<0, eWL>, OperandRegClass<1, mQQm>]>,
+                           ItinRegClassPair<II_VMOV_W_WMH_Q, [OperandRegClass<0, eWH>, OperandRegClass<1, mQQm>]>] in {	
 	def VMOV_mv_w : AIE2_mv_w_inst_mv< (outs OP_mMvAMWQDst:$dst), (ins OP_mMvAMWQSrc:$src),
```
gbossu, line 768: Could you think of an easy way to check that the `II_VMOV_W` itinerary is never used by the postmisched? Doing so can result in problems like the one you describe. Maybe in `getSchedClass` we should assert if we fallback to the default itinerary because none of the operands matched?

llvm/test/CodeGen/AIE/aie2/schedule/mov_bypass.mir
```diff
@@ -121,6 +121,33 @@ body:             |
...
+    ; CHECK-LABEL: name: bypass_wl_q
+    ; CHECK: $wl1 = VMOV_mv_w killed $q0
+    ; CHECK-NEXT: $wl0 = VMOV_mv_w killed $wl1
+    ; CHECK-NEXT: NOP
+    $wl1 = VMOV_mv_w $q0
+    $wl0 = VMOV_mv_w $wl1
+...
+
+---
+name:            no_bypass_wl_q
```
andcarminati: Do you mean  `no_bypass_wh_q`?

llvm/lib/Target/AIE/AIE2GenFixupInstrInfo.td
krishnamtibrewala: is that specific for `II_VMOV_W` or for all the reg-based itinerary ? 

If its specific to II_VMOV_W itinerary I can update the tablegen flow, where in an extra field if marked with an instruction will introduce an assert. Also doing this specifically for `postmisched` will be tricky since we do not have any info in the newly autogenerated `getSchedClass` function. 

Doing it for all is easy but will require adding lot of itinerary manually for MOV instructions since we use conservative itinerary in cases when we are writing to single register like SP, LR , LE .... and so on.

gbossu: I think we should aim to have "fully-covering" reg-based itineraries actually. But yes, I don't know how much work that might be to actually make all of our itineraries complete. I think it's fine for now.

llvm/lib/Target/AIE/AIE2GenFixupInstrInfo.td
```diff
@@ -760,7 +760,9 @@ let Itinerary = II_VMOV_W in {
                            ItinRegClassPair<II_VMOV_W_WMH_WMH,[OperandRegClass<0, eWH>, OperandRegClass<1, eWH>]>,
                            ItinRegClassPair<II_VMOV_W_WML_WML,[OperandRegClass<0, eWL>, OperandRegClass<1, eWL>]>,
                            ItinRegClassPair<II_VMOV_W_WMH_WML,[OperandRegClass<0, eWH>, OperandRegClass<1, eWL>]>,						   
-                           ItinRegClassPair<II_VMOV_W_WML_WMH,[OperandRegClass<0, eWL>, OperandRegClass<1, eWH>]>] in {	
+                           ItinRegClassPair<II_VMOV_W_WML_WMH,[OperandRegClass<0, eWL>, OperandRegClass<1, eWH>]>,
+                           ItinRegClassPair<II_VMOV_W_WML_Q, [OperandRegClass<0, eWL>, OperandRegClass<1, mQQm>]>,
+                           ItinRegClassPair<II_VMOV_W_WMH_Q, [OperandRegClass<0, eWH>, OperandRegClass<1, mQQm>]>] in {
```
gbossu: What about VMOV from Q to WL? I assume the output can use the MOV bypass then. Do we need new `II_VMOV_W_Q_WML` and `II_VMOV_W_Q_WMH` itineraries?

krishnamtibrewala: As per the ISA, Move slot bypass are only used when we are reading from either a `X` or `WL` register.

gbossu: What I meant is that

Below, the bypass should be taken
```
    $wl1 = VMOV_mv_w $q0
    $q1= VMOV_mv_w $wl1
```

But not here:
```
    $hl1 = VMOV_mv_w $q0
    $q1= VMOV_mv_w $wh1
```

However, I'm not sure that will be the case due to how we define the itineraries. In any case, it is worth testing :)

gbossu: Fixed in the second commit

andcarminati (approved):
LGTM.

PR #240, opened 2024-11-20
[AIE2] Fix memory access cycle for TM instructions

katerynamuts:
Exploring the architecture design, it was found that the instructions that store to and load from the tile memory (TM) map access the TM at cycle 4 instead of cycle 5. This PR changes MemoryCycle in AIE2Schedule.td file from 5 to 4 for such instructions.

The fix changes some tests. After checking the changed tests with @martien-de-jong, it became clear that changing the memory cycle changes the first and last memory cycle for the TM instructions from 5 to 4 which are used in `AIEBaseInstrInfo::getConservativeMemoryLatency` used in `MaxLatencyFinder`. As a  result, in the tests an additional nop is added because of the worst-case analysis used for inter-block scheduling.


llvm/test/CodeGen/AIE/aie2/schedule/interblock/memdeps.mir
```diff
@@ -34,6 +34,7 @@ body:             |
   ; CHECK-NEXT:   VST_SRS_D8_S32_ag_idx_imm killed $p0, 0, killed $cm0, killed $s0, implicit-def $srsrs_of, implicit $crsat, implicit $crrnd, implicit $crsrssign
   ; CHECK-NEXT:   NOP
   ; CHECK-NEXT:   NOP
+  ; CHECK-NEXT:   NOP
```
gbossu, line 37: As discussed offline, that extra might hurt QoR. I'd be happy to measure that impact.

If the results aren't good, we could extend
```
  int getMinFirstMemoryCycle() const;
  int getMaxFirstMemoryCycle() const;
  int getMinLastMemoryCycle() const;
  int getMaxLastMemoryCycle() const;
```

into e.g.
```
  int getMinFirstMemoryCycle(unsigned AddrSpace = 0) const;
  int getMaxFirstMemoryCycle(unsigned AddrSpace = 0) const;
  int getMinLastMemoryCycle(unsigned AddrSpace = 0) const;
  int getMaxLastMemoryCycle(unsigned AddrSpace = 0) const;
```

Taking the example of AIE2:
 - `getMinFirstMemoryCycle(AIE2::AddressSpaces::none)` should return `4`
 - `getMinFirstMemoryCycle(AIE2::AddressSpaces::DM)` should return `5`
 - `getMinFirstMemoryCycle(AIE2::AddressSpaces::TM)` should return `4`
 
But then I guess we'll have the same problem because many load/stores will have the default `none` address space, even though they actually are in `DM`. That would be quite a big change if we also need to infer the `AddressSpaces::DM` address space.

krishnamtibrewala: Why do you think it will be a big change if we need to infer `none` to `DM` ?

Note : access to tile memory is always annotated at intrinsic level so we will always infer that

krishnamtibrewala:
Hi @katerynamuts did you find this info in the architecture file that you had shared with me a month back ?

katerynamuts:
> Hi @katerynamuts did you find this info in the architecture file that you had shared with me a month back ?

@krishnamtibrewala yes

llvm/test/CodeGen/AIE/aie2/schedule/interblock/memdeps.mir
gbossu: I mean it will be an invasive change, because tests that go from IR to MIR will get different MMOs if the default becomes `DM` instead of `none`

PR #241, opened 2024-11-20
[AIE2] Enable up-combining for VLD + UNPACK

andcarminati:
If we can't delay VLD, we can try to antecipate UNPACK. This approach can be extended to other selection combiners.

### DIFF
--- a/llvm/lib/Target/AIE/AIE2InstructionSelector.cpp
+++ b/llvm/lib/Target/AIE/AIE2InstructionSelector.cpp
@@ -1604,14 +1604,22 @@ bool AIE2InstructionSelector::selectG_AIE_LOAD_UNPACK(
     MachineInstr &UNPACKI, MachineRegisterInfo &MRI) {
   Register LoadResult = (std::next(UNPACKI.uses().begin()))->getReg();
   MachineInstr *LoadOp = getDefIgnoringCopiesAndBitcasts(LoadResult, MRI);
+  // Should we build the instruction at load's position?
+  bool ShouldAdvanceOp = false;
 
   assert(LoadOp && "Expected SSA.");
 
   // Do not try to combine if one of the load's defs is used by another
   // instruction between the load and the VUNPACK or if there is a store
   // between the load and the VUNPACK.
-  if (!canDelayMemOp(*LoadOp, UNPACKI, MRI))
-    return false;
+  if (!canDelayMemOp(*LoadOp, UNPACKI, MRI)) {
+    // If we cannot delay the load, we can try to advance the combined
+    // instruction to the load's position.
+    if (canAdvanceOp(*LoadOp, UNPACKI, MRI))
+      ShouldAdvanceOp = true;
+    else
+      return false;
+  }
 
   if (!canCombineUNPACKLoad(*LoadOp, UNPACKI, MRI) ||
       LoadOp->getParent() != UNPACKI.getParent() || !MRI.hasOneUse(LoadResult))
@@ -1627,6 +1635,9 @@ bool AIE2InstructionSelector::selectG_AIE_LOAD_UNPACK(
 
   Register DstReg = UNPACKI.getOperand(0).getReg();
   Register SignReg = UNPACKI.getOperand(3).getReg();
+
+  if (ShouldAdvanceOp)
+    MIB.setInstr(*LoadOp);
 
   auto NewInstr = MIB.buildInstr(LSO->ISelOpcode);
 
--- a/llvm/lib/Target/AIE/AIECombinerHelper.cpp
+++ b/llvm/lib/Target/AIE/AIECombinerHelper.cpp
@@ -63,12 +63,12 @@ MachineInstr *findPreIncMatch(MachineInstr &MemI, MachineRegisterInfo &MRI,
   return nullptr;
 }
 
-/// Checks if any operand of \a Use is defined by \a MI.
-/// This is not transitive: it will not look at how the uses of \a MI are
+/// Checks if any operand of \a MI is defined by \a Def.
+/// This is not transitive: it will not look at how the uses of \a Def are
 /// defined.
-bool isUseOf(const MachineInstr &MI, const MachineInstr &Use) {
-  for (auto &Defs : Use.defs()) {
-    for (auto &MIUse : MI.uses()) {
+bool isUseOf(const MachineInstr &MI, const MachineInstr &Def) {
+  for (auto &Defs : Def.all_defs()) {
+    for (auto &MIUse : MI.all_uses()) {
       if (MIUse.isReg() && Defs.getReg() == MIUse.getReg())
         return true;
     }
@@ -117,6 +117,30 @@ bool llvm::canDelayMemOp(MachineInstr &MemI, MachineInstr &Dest,
            !MI.isSafeToMove(nullptr, SawStore);
   };
   return none_of(InstrRange, UnsafeToMovePast);
+}
+
+/// \return true if \a Dest can be moved just after \a MemI in order to allow
+/// combining
+bool llvm::canAdvanceOp(MachineInstr &MemI, MachineInstr &Dest,
+                        const MachineRegisterInfo &MRI) {
+  assert(Dest.getOpcode() != TargetOpcode::G_INTRINSIC_W_SIDE_EFFECTS &&
+         "Cannot advance Dest MI with side effects");
+  assert(!Dest.mayLoadOrStore() && "Cannot advance load/store Dest MI");
+  if (MemI.getParent() != Dest.getParent())
+    return false;
+  auto MII = std::next(MemI.getIterator());
+  auto MIE = Dest.getIterator();
+  auto InstrRange = make_range(MII, MIE);
+  auto UnsafeToMoveBefore = [&](const MachineInstr &MI) {
+    // Conditions that indicate it is unsafe to move:
+    // 1 - G_INTRINSIC_W_SIDE_EFFECTS without explicit output, which may include
+    // writing to a control register.
+    // 2 - Crossing the definition of an input operand of Dest.
+    return ((MI.getOpcode() == TargetOpcode::G_INTRINSIC_W_SIDE_EFFECTS &&
+             MI.defs().empty()) ||
+            isUseOf(Dest, MI));
+  };
+  return none_of(InstrRange, UnsafeToMoveBefore);
 }
 
 /// Find the def instruction for \p Reg, folding away any trivial copies and
--- a/llvm/test/CodeGen/AIE/aie2/GlobalISel/inst-select-no-combine-vldb_unpack.mir
+++ b/llvm/test/CodeGen/AIE/aie2/GlobalISel/inst-select-no-combine-vldb_unpack.mir
@@ -8,66 +8,90 @@
 
 # RUN: llc -mtriple aie2 -run-pass=instruction-select %s -verify-machineinstrs -o - | FileCheck %s
 
-# This tests that we don't combine if one of the load's defs is used before the VUNPACK instruction
+
+# This tests that we don't combine if one of the load's defs (data) is used after the VUNPACK instruction (no single use)
+
 ---
-name:            VLD_UNPACK_use
+name:            VLD_UNPACK_use_after
 alignment:       16
 legalized:       true
 regBankSelected: true
 body:             |
   bb.1.entry:
     liveins: $p0
-    ; CHECK-LABEL: name: VLD_UNPACK_use
+    ; CHECK-LABEL: name: VLD_UNPACK_use_after
     ; CHECK: liveins: $p0
     ; CHECK-NEXT: {{  $}}
     ; CHECK-NEXT: [[COPY:%[0-9]+]]:ep = COPY $p0
     ; CHECK-NEXT: [[COPY1:%[0-9]+]]:er = COPY $r0
     ; CHECK-NEXT: [[COPY2:%[0-9]+]]:em = COPY [[COPY1]]
     ; CHECK-NEXT: [[VLD_pstm_pseudo:%[0-9]+]]:vec256, [[VLD_pstm_pseudo1:%[0-9]+]]:ep = VLD_pstm_pseudo [[COPY]], [[COPY2]] :: (load (<32 x s8>))
-    ; CHECK-NEXT: $m1 = COPY [[VLD_pstm_pseudo1]]
-    ; CHECK-NEXT: [[COPY3:%[0-9]+]]:ep = COPY $m1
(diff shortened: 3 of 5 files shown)


andcarminati:
QoR:

```
|--------------------------------------------------------------|------------|----------|--------------|
| Core_Compute_Cycle_Count                                     | aie-public | This PR. | Total diff   |
|--------------------------------------------------------------|------------|----------|--------------|
| Softmax_bf16_0                                               |       6350 |     6351 | SAME(+0.02%) |
|--------------------------------------------------------------|------------|----------|--------------|
| ThresholdedRelu_aie2_int8                                    |        865 |      849 | IMPR(-1.85%) |
|--------------------------------------------------------------|------------|----------|--------------|
| Erf_aie2_int8_0                                              |       2554 |     2490 | IMPR(-2.51%) |
|--------------------------------------------------------------|------------|----------|--------------|
| Erf_aie2_int8_0_ptr_interface                                |       2533 |     2469 | IMPR(-2.53%) |
|--------------------------------------------------------------|------------|----------|--------------|
| HardswishAsHardsigmoid_aie2_0                                |       1368 |     1240 | IMPR(-9.36%) |
|--------------------------------------------------------------|------------|----------|--------------|
| Hardswish_aie2_0                                             |       1368 |     1240 | IMPR(-9.36%) |
|--------------------------------------------------------------|------------|----------|--------------|
| Averege diff                                                 |            | -0.07%   | -0.07%       |
|--------------------------------------------------------------|------------|----------|--------------|
| Diff stdev                                                   |            |     0.75 |         0.75 |
|--------------------------------------------------------------|------------|----------|--------------|
| Quantile #1                                                  |            | +0.00%   | +0.00%       |
|--------------------------------------------------------------|------------|----------|--------------|
| Quantile #2                                                  |            | +0.00%   | +0.00%       |
|--------------------------------------------------------------|------------|----------|--------------|
| Quantile #3                                                  |            | +0.00%   | +0.00%       |
|--------------------------------------------------------------|------------|----------|--------------|
| Quantile #4                                                  |            | +0.00%   | +0.00%       |
|--------------------------------------------------------------|------------|----------|--------------|
| Quantile #5                                                  |            | +0.00%   | +0.00%       |
|--------------------------------------------------------------|------------|----------|--------------|
| Quantile #6                                                  |            | +0.00%   | +0.00%       |
|--------------------------------------------------------------|------------|----------|--------------|
| Quantile #7                                                  |            | +0.00%   | +0.00%       |
|--------------------------------------------------------------|------------|----------|--------------|
| Quantile #8                                                  |            | +0.00%   | +0.00%       |
|--------------------------------------------------------------|------------|----------|--------------|
| Quantile #9                                                  |            | +0.00%   | +0.00%       |
|--------------------------------------------------------------|------------|----------|--------------|

```

llvm/lib/Target/AIE/AIE2InstructionSelector.cpp
```diff
@@ -1604,14 +1604,19 @@ bool AIE2InstructionSelector::selectG_AIE_LOAD_UNPACK(
...
 
   assert(LoadOp && "Expected SSA.");
 
   // Do not try to combine if one of the load's defs is used by another
   // instruction between the load and the VUNPACK or if there is a store
   // between the load and the VUNPACK.
-  if (!canDelayMemOp(*LoadOp, UNPACKI, MRI))
-    return false;
+  if (!canDelayMemOp(*LoadOp, UNPACKI, MRI)) {
+    if (canAntecipateOp(*LoadOp, UNPACKI, MRI))
```
gbossu: Nit: ant*i*cipate
I'm wondering if we should name that e.g. `canAdvanceOp` to be more symetric?

llvm/lib/Target/AIE/AIECombinerHelper.cpp
```diff
@@ -63,11 +63,11 @@ MachineInstr *findPreIncMatch(MachineInstr &MemI, MachineRegisterInfo &MRI,
...
 }
 
-/// Checks if any operand of \a Use is defined by \a MI.
-/// This is not transitive: it will not look at how the uses of \a MI are
+/// Checks if any operand of \a MI is defined by \a Def.
+/// This is not transitive: it will not look at how the uses of \a Def are
 /// defined.
-bool isUseOf(const MachineInstr &MI, const MachineInstr &Use) {
-  for (auto &Defs : Use.defs()) {
+bool isUseOf(const MachineInstr &MI, const MachineInstr &Def) {
```
gbossu, line 69: Thanks, that change was on my todo list :D

llvm/lib/Target/AIE/AIE2InstructionSelector.cpp
andcarminati: Sure! Nice suggestion.

llvm/lib/Target/AIE/AIECombinerHelper.cpp
```diff
@@ -63,11 +63,11 @@ MachineInstr *findPreIncMatch(MachineInstr &MemI, MachineRegisterInfo &MRI,
...
+  for (auto &Defs : Def.defs()) {
```
gbossu: Does that iterate on implicit defs as well? I'm never sure

llvm/lib/Target/AIE/AIECombinerHelper.cpp
```diff
@@ -119,6 +119,21 @@ bool llvm::canDelayMemOp(MachineInstr &MemI, MachineInstr &Dest,
   return none_of(InstrRange, UnsafeToMovePast);
 }
 
+/// \return true if \a Dest can be moved just after \a MemI in order to allow
+/// combining
+bool llvm::canAntecipateOp(MachineInstr &MemI, MachineInstr &Dest,
+                           const MachineRegisterInfo &MRI) {
```
gbossu: Maybe assert that we don't try to move an instruction (`Dest`) that has side effects or loads/stores?

llvm/test/CodeGen/AIE/aie2/GlobalISel/inst-select-no-combine-vldb_unpack.mir
```diff
@@ -10,24 +10,23 @@
 
 # This tests that we don't combine if one of the load's defs is used before the VUNPACK instruction
```
gbossu: That comment is outdated. Now we do combine 💪

llvm/lib/Target/AIE/AIECombinerHelper.cpp
andcarminati: No, just explicit ones! To cover all definitions, we need `all_defs()`

llvm/test/CodeGen/AIE/aie2/GlobalISel/inst-select-no-combine-vldb_unpack.mir
andcarminati: Moved to the other file (with positive tests)

llvm/lib/Target/AIE/AIE2InstructionSelector.cpp
```diff
@@ -1628,6 +1633,9 @@ bool AIE2InstructionSelector::selectG_AIE_LOAD_UNPACK(
   Register DstReg = UNPACKI.getOperand(0).getReg();
   Register SignReg = UNPACKI.getOperand(3).getReg();
 
+  if (ShouldAntecipate)
```
F-Stuckmann: should it not be `ShouldAnticipate` ? 
Is there a more descriptive name for it? Like `CanJoinLoadAndUse` ?

andcarminati: I changed the name and also included a comment.

llvm/lib/Target/AIE/AIECombinerHelper.cpp
gbossu: I guess we should be careful about implicit defs as well. The instruction might write to a control register that is used by a later instruction. I don't think that case can happen in AIE2, but who knows.

However, I'm now thinking about a case like this:
```
%0:vec256 = VLD
$crSat = ...
%1:vec512 = VUNPACK %0, implicit $crSat
```
There we cannot advance the VUNPACK, because it would then use a different value for `$crSat`.

andcarminati: Yes, this could be a case if we extend for VUPS case...

andcarminati: The best, in this case, is to iterate over all uses and defs, like:

```
bool isUseOf(const MachineInstr &MI, const MachineInstr &Def) {
  for (auto &Defs : Def.all_defs()) {
    for (auto &MIUse : MI.all_uses()) {
      if (MIUse.isReg() && Defs.getReg() == MIUse.getReg())
        return true;
    }
  }
  return false;
}
```

gbossu: Oh alright, it's VPACK that uses `crSat`, not VUNPACK. Still, the latter could use `crUnpackSign`, so we can end up with the same problem. I think we need to use `all_uses` and `all_defs` like you proposed.

gbossu (commented):
LGTM, but before merging, we need to be a bit more careful about implicit uses/defs I think. Otherwise I think it's nice and concise code :)

llvm/lib/Target/AIE/AIECombinerHelper.cpp
```diff
@@ -119,6 +119,24 @@ bool llvm::canDelayMemOp(MachineInstr &MemI, MachineInstr &Dest,
...
+  assert(!Dest.mayLoadOrStore() && "Cannot advance load/store Dest MI");
+  if (MemI.getParent() != Dest.getParent())
+    return false;
+  auto MII = std::next(MemI.getIterator());
+  auto MIE = Dest.getIterator();
+  auto InstrRange = make_range(MII, MIE);
+  auto UnsafeToMoveBefore = [&](const MachineInstr &MI) {
+    return (isUseOf(Dest, MI));
+  };
+  return none_of(InstrRange, UnsafeToMoveBefore);
```
krishnamtibrewala, line 143: Nice way of using `none_of` 🚀

andcarminati:
Hi @gbossu, all comments were addressed. Thank you for the review ;-).

llvm/lib/Target/AIE/AIECombinerHelper.cpp
```diff
@@ -63,12 +63,12 @@ MachineInstr *findPreIncMatch(MachineInstr &MemI, MachineRegisterInfo &MRI,
...
-/// This is not transitive: it will not look at how the uses of \a MI are
+/// Checks if any operand of \a MI is defined by \a Def.
+/// This is not transitive: it will not look at how the uses of \a Def are
 /// defined.
-bool isUseOf(const MachineInstr &MI, const MachineInstr &Use) {
-  for (auto &Defs : Use.defs()) {
-    for (auto &MIUse : MI.uses()) {
+bool isUseOf(const MachineInstr &MI, const MachineInstr &Def) {
+  for (auto &Defs : Def.all_defs()) {
+    for (auto &MIUse : MI.all_uses()) {
```
gbossu, line 71: Could you add a test to check that the `VPACK`/`VUNPACK` isn't advanced when one of its implicit inputs is defined in between the load/store and the op?
E.g.
```
%0 = VLD ...
VST ... ; prevent delaying VLD
$crSign = ...
%1 = VPACK %0, implicit $crSign
```

andcarminati: I included a test for VUNPACK with a control register write in between. I am also looking explicitly to `G_INTRINSIC_W_SIDE_EFFECTS` when we try to advance an instruction.

andcarminati:
After this change:
```
  for (auto &Defs : Def.all_defs()) {
    for (auto &MIUse : MI.all_uses()) {
```

No side-effects were observed in QoR.

llvm/test/CodeGen/AIE/aie2/GlobalISel/inst-select-no-combine-vldb_unpack.mir
```diff
@@ -8,66 +8,90 @@
...
-    ; CHECK-NEXT: [[COPY2:%[0-9]+]]:er = COPY $r0
-    ; CHECK-NEXT: [[COPY3:%[0-9]+]]:em = COPY [[COPY2]]
-    ; CHECK-NEXT: [[COPY4:%[0-9]+]]:vec256 = COPY $amll0
-    ; CHECK-NEXT: [[VLD_pstm_pseudo:%[0-9]+]]:vec256, [[VLD_pstm_pseudo1:%[0-9]+]]:ep = VLD_pstm_pseudo [[COPY]], [[COPY3]] :: (load (<32 x s8>))
-    ; CHECK-NEXT: VST_dmw_sts_w_ag_idx_imm [[COPY4]], [[COPY1]], 0 :: (store (<32 x s8>))
+    ; CHECK-NEXT: [[COPY1:%[0-9]+]]:er = COPY $r0
+    ; CHECK-NEXT: [[COPY2:%[0-9]+]]:em = COPY [[COPY1]]
+    ; CHECK-NEXT: [[VLD_pstm_pseudo:%[0-9]+]]:vec256, [[VLD_pstm_pseudo1:%[0-9]+]]:ep = VLD_pstm_pseudo [[COPY]], [[COPY2]] :: (load (<32 x s8>))
+    ; CHECK-NEXT: $crunpacksign = COPY [[COPY1]]
     ; CHECK-NEXT: [[VUNPACK_S16_S8_:%[0-9]+]]:vec512 = VUNPACK_S16_S8 [[VLD_pstm_pseudo]]
```
gbossu, line 86: Super-nit: we are writing to `crSign`, but `VUNPACK_S16_S8` does not read it. If we get smarter, I guess we will be able to combine. I guess what we should really check is `VST + PACK` with a `set_crsat()` in between.

gbossu (approved):
LGTM!

llvm/test/CodeGen/AIE/aie2/GlobalISel/inst-select-no-combine-vldb_unpack.mir
andcarminati: Sure it could be improved even more. When we are combining/selecting, this csr write in-between is just an intrinsic call with a constant operand representing the register. We can parse this call against the registers that are consumed by the final instruction (we need a mapping for this). About the test, `VUNPACK_S16_S8` uses indeed the `crunpacksign`, but if we change the register, the result will be the same.

As this PR solves the problems of the tracked benchmarks, we can keep it as is and look to `set_csr()` in the future if we see this as a blocker in some benchmarks.

PR #242, opened 2024-11-20
Martien.exhaustiveswp

martien-de-jong:
Set the baseline for GEMM and Conv2D_bf16
Add a provision to try multiple strategies for a single II.
Add a critical path heuristic

For best effect on QoR, the pre-pipeliner and the prescheduler should be skipped. This is forced by 

#pragma clang loop pipeliner(disable)

in the source, or globally using --enable-pipeliner=0 --enable-misched=0

QOR:
I ran a like-for like comparison with the base aie-public version The CSV reporting failed, but the result json file shows significant overall improvements (up to >10%)  with only a few minor sub-percent regressions 

This is the tail of the diff. Trust me the regression head is tiny in comparison

image


llvm/test/CodeGen/AIE/aie2/schedule/postpipeliner/conv2d_bf16-1.mir
```diff
@@ -0,0 +1,241 @@
...
+# See https://llvm.org/LICENSE.txt for license information.
+# SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+#
+# (c) Copyright 2024 Advanced Micro Devices, Inc. or its affiliates
+
+# RUN: llc --mtriple=aie2 -O2 --start-before=postmisched %s \
+# RUN:   --debug-only=postpipeliner-summary -o - | FileCheck %s
+
+
+# derived from gemm_bf16_0
```
andcarminati: nit: conv2d_bf16_0

llvm/test/CodeGen/AIE/aie2/schedule/postpipeliner/conv2d_bf16.mir
```diff
@@ -0,0 +1,242 @@
...
+# See https://llvm.org/LICENSE.txt for license information.
+# SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+#
+# (c) Copyright 2024 Advanced Micro Devices, Inc. or its affiliates
+
+# RUN: llc --mtriple=aie2 -O2 --start-before=postmisched %s \
+# RUN:   --debug-only=postpipeliner-summary -o - | FileCheck %s
+
+
+# derived from gemm_bf16_0
```
andcarminati: nit: conv2d_bf16_0

llvm/lib/Target/AIE/AIEPostPipeliner.cpp
```diff
@@ -557,78 +564,83 @@ PostPipelinerStrategy::~PostPipelinerStrategy() {}
...
   bool better(const SUnit &A, const SUnit &B) override {
-    auto &IA = Info[A.NodeNum];
-    auto &IB = Info[B.NodeNum];
-    if (FromTop && SuccSiblingScheduled.count(A.NodeNum) >
-                       SuccSiblingScheduled.count(B.NodeNum)) {
-      return true;
-    }
-    if (FromTop) {
-      return IA.LCDLatest < IB.LCDLatest;
+    int PBias = FromTop ? 0 : 3;
```
martien-de-jong: No, not happy with this. I want to virtualize the FromTop and FromBottom differences. Also, FromBottom should be completed to mirror FromTop.

llvm/lib/Target/AIE/AIEPostPipeliner.cpp
```diff
@@ -518,7 +525,7 @@ bool PostPipeliner::scheduleFirstIteration(PostPipelinerStrategy &Strategy) {
 
     scheduleNode(SU, Actual);
     Info[N].Scheduled = true;
-    LLVM_DEBUG(dbgs() << "Scoreboard\n"; Scoreboard.dumpFull(););
+    DEBUG_FULL(dbgs() << "Scoreboard\n"; Scoreboard.dumpFull(););
```
martien-de-jong, line 471: Note: dumping the scoreboard is pretty expensive, it brings compile-time up from tens of milliseconds to seconds

llvm/lib/Target/AIE/AIEPostPipeliner.cpp
```diff
@@ -463,12 +466,16 @@ int PostPipeliner::mostUrgent(PostPipelinerStrategy &Strategy) {
...
   for (int K = 0; K < NTotalInstrs; K++) {
     auto &N = Info[K];
-    N.LastEarliestPusher = -1;
-    N.LastLatestPusher = -1;
+    if (ResetCritical) {
+      N.NumPushedEarliest = 0;
+      N.NumPushedLatest = 0;
+      N.LastEarliestPusher = -1;
+      N.LastLatestPusher = -1;
+    }
```
martien-de-jong: These parameters represents the path that pushed Earliest of the failing node beyond its modulo cycle in the previous iteration. If we don't reset it, we can use it to provide look-ahead in a rerun.

llvm/lib/Target/AIE/AIEPostPipeliner.cpp
```diff
@@ -557,78 +564,83 @@ PostPipelinerStrategy::~PostPipelinerStrategy() {}
...
-  // but chances are really slim.
-  uint64_t spice(uint64_t Value) {
-    return Value * 0x1555555555555555ull + 0x123456789abcdefull;
-  }
-  // Maximum length 63 bits LFSR
-  int randomBit() {
-    const uint64_t BitIn = (RandomState >> 62) ^ (RandomState >> 61);
-    RandomState = (RandomState << 1) | (BitIn & 1);
-    return RandomState & 1;
-  }
```
martien-de-jong: This will disappear in the final PR to be merged.

llvm/lib/Target/AIE/AIEPostPipeliner.cpp
```diff
@@ -557,78 +564,83 @@ PostPipelinerStrategy::~PostPipelinerStrategy() {}
...
+    return false;
   }
   bool fromTop() override { return FromTop; }
+
   void selected(const SUnit &N) override {
-    switch (Mode) {
-    case Modulation::Alternate:
-      FromTop = !FromTop;
-      break;
-    case Modulation::Split:
```
martien-de-jong: OneWay can be promoted to Split, by putting off the toggle indefinitely.

llvm/lib/Target/AIE/AIEPostPipeliner.cpp
```diff
@@ -641,10 +653,10 @@ class ConfigStrategy : public PostPipelinerStrategy {
...
                  int Length, enum Modulation Mode, bool FromTop, uint64_t Param)
       : PostPipelinerStrategy(DAG, Info, Length), Mode(Mode), FromTop(FromTop),
-        Param(Param), RandomState(spice(Param)) {
-    if (Mode == Modulation::Split) {
-      Count = Length * Param / 1000;
-    }
+        Param(Param) {
+    Priority[0] = (Param >> 0) & 0x3;
+    Priority[1] = (Param >> 2) & 0x3;
+    Priority[2] = (Param >> 4) & 0x3;
```
martien-de-jong: Not really elegant. Inspired by an opaque chromosome value to be used in a genetic algorithm.

llvm/lib/Target/AIE/AIEPostPipeliner.cpp
```diff
@@ -688,16 +704,25 @@ bool PostPipeliner::tryHeuristics() {
...
-    LLVM_SUMMARY(dbgs() << "--- Strategy " << S.name());
+    resetSchedule(/*ResetCritical=*/true);
+    DEBUG_SUMMARY(dbgs() << "--- Strategy " << S.name());
+    if (scheduleFirstIteration(S) && scheduleOtherIterations()) {
+      DEBUG_SUMMARY(dbgs() << " found II=" << II << "\n");
+      return true;
+    }
+
+    DEBUG_SUMMARY(dbgs() << " failed\n");
+    resetSchedule(/*ResetCritial=*/false);
```
martien-de-jong: Note: we could run multiple times without resetting the critical path to highlight more problematic paths and point out 'multi-critical nodes'. We'd need to accumulate criticality in a sane manner (and probably think about possible overflow of those numbers)

llvm/lib/Target/AIE/AIEPostPipeliner.cpp
```diff
@@ -190,29 +190,35 @@ void PostPipeliner::computeLoopCarriedParameters() {
   }
 
   // Propagate Earliest upstream, initialize Latest
+  // Unrestricted: last cycle of last stage
+  const int Latest = NCopies * II - 1;
   for (int K = 0; K < NInstr; K++) {
     const int K2 = K + NInstr;
     const int Earliest = Info[K2].Earliest - II;
```
gbossu: Checking: this could be negative

llvm/lib/Target/AIE/AIEPostPipeliner.cpp
```diff
@@ -190,29 +190,35 @@ void PostPipeliner::computeLoopCarriedParameters() {
   }
 
   // Propagate Earliest upstream, initialize Latest
+  // Unrestricted: last cycle of last stage
+  const int Latest = NCopies * II - 1;
```
gbossu: I find it confusing to initialize Latest here. Maybe at least move it in its own loop?

But I'm also wondering what this brings to have a better "upper-bound" like `NCopies * II - 1`. Why not directly initialize `Latest` to `UINT_MAX` in `NodeInfo`?

llvm/test/CodeGen/AIE/aie2/schedule/postpipeliner/conv2d_bf16-1.mir
```diff
@@ -36,67 +36,61 @@
...
-  ; CHECK-NEXT:    and r3, r3, r0; mov p4, p7; vmac.f bmh5, bmh5, x1, x5, r29
-  ; CHECK-NEXT:    add r3, r3, #34; vmac.f bml2, bml2, x3, x5, r29
-  ; CHECK-NEXT:    vmac.f bml0, bml0, x10, x5, r29
-  ; CHECK-NEXT:    vmac.f bmh1, bmh1, x8, x9, r29
-  ; CHECK-NEXT:    vmac.f bmh0, bmh0, x1, x9, r29
+  ; CHECK-NEXT:    vshuffle x3, x3, x5, r24; vmac.f bmh1, bmh1, x8, x9, r29
+  ; CHECK-NEXT:    mov r3, p0; vmac.f bmh0, bmh0, x1, x9, r29
+  ; CHECK-NEXT:    and r3, r3, r0; mov p4, p7; vmac.f bmh3, bmh3, x3, x9, r29
+  ; CHECK-NEXT:    add r3, r3, #34; vmac.f bmh2, bmh2, x10, x9, r29
+  ; CHECK-NEXT:    vmac.f bmh7, bmh7, x8, x5, r29
```
gbossu: That looks a lot more compact :)

llvm/lib/Target/AIE/AIEBaseSubtarget.cpp
```diff
@@ -661,6 +661,8 @@ AIEBaseSubtarget::getPostRAMutationsImpl(const Triple &TT) {
   std::vector<std::unique_ptr<ScheduleDAGMutation>> Mutations;
   Mutations.emplace_back(std::make_unique<LockDelays>());
   if (!TT.isAIE1()) {
+    if (EnableWAWStickyRegisters)
+      Mutations.emplace_back(std::make_unique<WAWStickyRegistersEdges>());
```
gbossu, line 708: We need to check the QoR impact of that, this touches everything scheduled by the postmissched, not just the post-pipeliner.

llvm/test/CodeGen/AIE/aie2/schedule/status_regs/srWAW.mir
```diff
@@ -1,14 +1,17 @@
+# NOTE: Assertions have been autogenerated by utils/update_mir_test_checks.py UTC_ARGS: --version 4
```
gbossu: That test was not auto-generated. Can you revert it to how it was?

llvm/lib/Target/AIE/AIEPostPipeliner.cpp
```diff
@@ -33,6 +36,58 @@ class PostPipelineDumper : public PipelineScheduleVisitor {
...
+llvm::raw_ostream &operator<<(llvm::raw_ostream &OS, const SlotCounts &Val) {
+  OS << "{ ";
+  const char *Sep = "";
+  for (int I = 0; I < Val.size(); I++) {
+    OS << Sep << Val[I];
+    Sep = ", ";
+  }
+  OS << " }";
+  return OS;
+}
```
gbossu: I think this new `SlotCounts` should be lifted to its own header. This can be useful beyond the `MachinePipeliner`

llvm/lib/Target/AIE/AIEPostPipeliner.cpp
```diff
@@ -106,32 +161,24 @@ bool PostPipeliner::canAccept(MachineBasicBlock &LoopBlock) {
...
-    while (C < NInstr && (Scoreboard[C] & Slots)) {
-      C++;
-    }
-    if (C >= NInstr) {
-      MII = NInstr;
-      break;
-    }
-    Scoreboard[C] |= Slots;
-    MII = std::max(MII, C + 1);
+    Counts += getSlotCounts(MI, TII);
```
gbossu, line 127: Nice!

llvm/lib/Target/AIE/AIEPostPipeliner.cpp
```diff
@@ -146,7 +193,20 @@ void PostPipeliner::scheduleNode(SUnit &SU, int Cycle) {
     const int NewEarliest = Cycle + Latency;
     if (NewEarliest > Info[SNum].Earliest) {
       Info[SNum].Earliest = NewEarliest;
-      LLVM_DEBUG(dbgs() << SNum << " to " << Info[SNum].Earliest << "; ");
+      LLVM_DEBUG(dbgs() << SNum << " to " << Info[SNum].Earliest << " -; ");
+    }
+  }
+  for (auto &Dep : SU.Preds) {
```
gbossu, line 155: I think I'll need a walkthrough for that commit, there seems to be quite a lot happening on top of refactoring.

llvm/test/CodeGen/AIE/aie2/schedule/postpipeliner/gemm-1.mir
```diff
@@ -0,0 +1,231 @@
...
+  ; CHECK-NEXT:    vmac.f bml5, bml5, x8, x3, r3
+  ; CHECK-NEXT:    vmac.f bml6, bml6, x0, x3, r3
+  ; CHECK-NEXT:    vmac.f bml4, bml4, x6, x3, r3
+  ; CHECK-NEXT:    nop
+  ; CHECK-NEXT:    nop
+  ; CHECK-NEXT:    nop
+  ; CHECK-NEXT:    nop
+  ; CHECK-NEXT:    nop
+  ; CHECK-NEXT:    nop
+  ; CHECK-NEXT:    nop
```
andcarminati: The original `vst.conv`'s can be also included, because they can be used a baseline for the epilogue merging.

llvm/lib/Target/AIE/AIEPostPipeliner.cpp
martien-de-jong: Agree. Working on it.

llvm/lib/Target/AIE/AIEPostPipeliner.cpp
martien-de-jong: This has been changed. We now have negative values representing the (reverse) distance to the end. It matters since we also use it to define the range where to insert.

llvm/test/CodeGen/AIE/aie2/schedule/postpipeliner/conv2d_bf16-1.mir
```diff
@@ -0,0 +1,238 @@
...
+# (c) Copyright 2024 Advanced Micro Devices, Inc. or its affiliates
+
+# RUN: llc --mtriple=aie2 -O2 --start-before=postmisched %s \
+# RUN:   --debug-only=postpipeliner-summary -o - | FileCheck %s
+
+
+# derived from conv2d_bf16_0
+
+--- |
+  define dso_local void @gemm(ptr addrspace(5) noalias nocapture writeonly %d, ptr addrspace(6) noalias nocapture readonly %s, i32 noundef %n) local_unnamed_addr #0 {
```
andcarminati: nit: this `conv2d` is still `gemm` ;-)

llvm/lib/Target/AIE/AIEPostPipeliner.cpp
martien-de-jong: Yes, I think this can be negative. In particular, [K2].Earliest only gets >0 if there's a loop carried dependence. However, this is only used to max, and is int, so it doesn't matter.

llvm/lib/Target/AIE/AIEPostPipeliner.h
```diff
@@ -147,6 +147,10 @@ class PostPipeliner {
   /// on the second iteration.
   void computeLoopCarriedParameters();
 
+  /// Try all heuristics, stop at the first that fits the II
+  /// If it returns true, a valid schedule is laid down in Info.
+  bool tryHeuristics();
```
martien-de-jong, line 192: Hmm. Wrong commit.

llvm/lib/Target/AIE/AIEPostPipeliner.cpp
```diff
@@ -188,54 +189,141 @@ int PostPipeliner::fit(MachineInstr *MI, int First, int Last, int II) {
...
-      for (auto &Dep : SU.Preds) {
-        int P = Dep.getSUnit()->NodeNum;
-        int NewLatest = Latest - Dep.getSignedLatency();
-        if (NewLatest < Info[P].Latest) {
-          Info[P].Latest = NewLatest;
-          Changed = true;
-        }
+  }
+
+  // Make Earliest of the second iteration push up Latest of the first
```
martien-de-jong, line 305: Putting this in LCDLatest in order not to pollute the 'clean' Latest

llvm/lib/Target/AIE/AIEPostPipeliner.cpp
```diff
@@ -188,54 +189,141 @@ int PostPipeliner::fit(MachineInstr *MI, int First, int Last, int II) {
...
+    auto &Me = Info[K];
+    const int Latest = Info[K].Latest;
+    for (auto &Dep : SU.Preds) {
+      if (Dep.getKind() != SDep::Data) {
+        continue;
+      }
+      int P = Dep.getSUnit()->NodeNum;
+      auto &Pred = Info[P];
+      AddOffspring(Pred, K);
+      for (auto Offs : Me.Offspring) {
```
martien-de-jong, line 249: Yes. this is at least quadratic. My only defense is that its pruning effect may be (handwaving a bit here) exponential

llvm/lib/Target/AIE/AIEPostPipeliner.cpp
```diff
@@ -168,6 +173,12 @@ void PostPipeliner::scheduleNode(SUnit &SU, int Cycle) {
   if (Next < NTotalInstrs) {
     Info[Next].Earliest = std::max(Info[Next].Earliest, Cycle + II);
   }
+#if 0
+  int Prev = SU.NodeNum - NInstr;
+  if (Prev >= 0) {
+    Info[Prev].Latest = std::max(Info[Prev].Latest, Cycle - II);
+  }
+#endif
```
martien-de-jong: Refactoring relic. to be removed / reinstated.

llvm/lib/Target/AIE/AIEPostPipeliner.cpp
```diff
@@ -423,7 +433,6 @@ void PostPipeliner::resetSchedule(bool FullReset) {
       N.LastEarliestPusher = -1;
       N.LastLatestPusher = -1;
     }
-#endif
```
martien-de-jong: Refactoring relic.  This will be applied earlier as to be invisible

llvm/lib/Target/AIE/AIEPostPipeliner.cpp
```diff
@@ -519,6 +528,148 @@ class DefaultStrategy : public PostPipelinerStrategy {
...
+      MinLength += II;
+    }
+  }
+
+  DEBUG_SUMMARY(dbgs() << "-- MinLength=" << MinLength << "\n");
+
+  static const std::tuple<int, bool, uint64_t> Configs[] = {
+      // Loosely speaking, a lower value of the first parameter targets
+      // a lower stage count, which benefits code size.
+      {0, false, ConfigStrategy::Latest},
```
martien-de-jong: Will be extended in later commits.

llvm/test/CodeGen/AIE/aie2/schedule/postpipeliner/conv2d_bf16.mir
```diff
@@ -25,74 +25,68 @@
   ; CHECK-NEXT:    nop // Delay Slot 2
   ; CHECK-NEXT:    nop // Delay Slot 1
   ; CHECK-NEXT:  // %bb.1: // %for.body.preheader
-  ; CHECK-NEXT:    vldb wh8, [p0, #32]; nopa ; nops ; nopx ; mov p4, p2; nopv
-  ; CHECK-NEXT:    vldb wl8, [p0], m4; nopx ; mov p5, p7
```
martien-de-jong: You may want to view these in a side-by-side diff. and line up the loop labels.

llvm/test/CodeGen/AIE/aie2/schedule/postpipeliner/conv2d.mir
```diff
@@ -100,10 +100,10 @@
   ; CHECK-NEXT:    vldb wh6, [p0], m6; nopa ; nops ; nopxm ; nopv
   ; CHECK-NEXT:    vldb wl8, [p0], m6; nopa ; nops ; add r0, r5, #33; vshift.align x2, x2, s1, x8, r0; nopv
   ; CHECK-NEXT:    vldb.3d wh8, [p0], d0; nopa ; nops ; nopx ; vshuffle x9, x4, x2, r2; nopv
-  ; CHECK-NEXT:    vldb wl10, [p1], #32; nopa ; nops ; nopx ; vshuffle x3, x4, x2, r3; nopv
-  ; CHECK-NEXT:    vldb wh10, [p1], #32; nopx ; vshuffle x1, x9, x0, r8; vmac cm1, cm1, x9, x10, r4
-  ; CHECK-NEXT:    vldb wl7, [p1], #32; vshuffle x5, x3, x0, r8; vmac cm3, cm3, x3, x10, r4
-  ; CHECK-NEXT:    vldb wh7, [p1], #32; mov r6, p0; vmac cm2, cm2, x1, x10, r4
```
martien-de-jong: This diff is a case of the new, simple heuristic added at the top finding the II that was found earlier by a more complex one. Different schedule, same II.

llvm/test/CodeGen/AIE/aie2/schedule/postpipeliner/conv2d_bf16-1.mir
martien-de-jong: Actually it's conv2d. It's just called gemm ;-)

llvm/test/CodeGen/AIE/aie2/schedule/postpipeliner/conv2d_bf16-1.mir
```diff
@@ -12,8 +12,8 @@
 # derived from conv2d_bf16_0
 
 --- |
-  define dso_local void @gemm(ptr addrspace(5) noalias nocapture writeonly %d, ptr addrspace(6) noalias nocapture readonly %s, i32 noundef %n) local_unnamed_addr #0 {
-  ; CHECK-LABEL: gemm:
+  define dso_local void @conv2d(ptr addrspace(5) noalias nocapture writeonly %d, ptr addrspace(6) noalias nocapture readonly %s, i32 noundef %n) local_unnamed_addr #0 {
+  ; CHECK-LABEL: conv2d:
```
gbossu, line 16: I assume you'll squash that with the previous commit?

llvm/test/CodeGen/AIE/aie2/schedule/postpipeliner/gemm-nooffset.mir
```diff
@@ -24,65 +24,96 @@
...
+  ; CHECK-NEXT:    nop
+  ; CHECK-NEXT:    nop
+  ; CHECK-NEXT:    nop
+  ; CHECK-NEXT:    nop
+  ; CHECK-NEXT:    nop
+  ; CHECK-NEXT:    nop
+  ; CHECK-NEXT:    nop
+  ; CHECK-NEXT:    nop
+  ; CHECK-NEXT:    nop
+  ; CHECK-NEXT:    nop
```
gbossu, line 116: Can you explain the nops?

martien-de-jong: Yep, as discussed just now with Andreu. This is an almost empty third stage, extended to II.

llvm/test/CodeGen/AIE/aie2/schedule/postpipeliner/conv2d_bf16-1.mir
martien-de-jong: Yes. that slipped through. I noticed another name change to be done.

llvm/lib/Target/AIE/AIESlotCounts.cpp
```diff
@@ -0,0 +1,71 @@
+//===- AIESlotCounts.cpp - SlotCount utilit  y                            ===//
```
gbossu: Nit: `utilit  y`

llvm/lib/Target/AIE/AIESlotCounts.h
```diff
@@ -0,0 +1,55 @@
...
+
+#include "MCTargetDesc/AIEFormat.h"
+
+namespace llvm {
+namespace AIE {
+
+/// Efficient representation of slot requirements
+class SlotCounts {
+  static const int MaxSlots = 16;
+  short Counts[MaxSlots];
```
gbossu: Nit: given that `MaxSlots` is only `16`, I don't think we save so much memory going to `short`. Or do you think `SlotCounts` will be used extensively and the size difference will matter?

llvm/lib/Target/AIE/AIESlotCounts.h
```diff
@@ -0,0 +1,55 @@
+//===- AIESlotCounts.h - Resource computation utility
```
gbossu: Nit: Missing the closing `===//`

llvm/lib/Target/AIE/AIESlotCounts.cpp
```diff
@@ -0,0 +1,71 @@
...
+}
+
+SlotCounts &SlotCounts::operator+=(const SlotCounts &Other) {
+  // The common part
+  for (int I = 0; I < Size && I < Other.Size; I++) {
+    Counts[I] += Other.Counts[I];
+  }
+  // Any excess from the other
+  while (Size < Other.Size) {
+    Counts[Size++] = Other.Counts[Size];
```
gbossu: Hmm, `Counts[Size++]` will have the previous value of `Size` as index, but `Other.Counts[Size]` will use the old one? Maybe move the increment in a different statement to clear our any confusion?

llvm/test/CodeGen/AIE/aie2/schedule/postpipeliner/README
```diff
@@ -0,0 +1,18 @@
+# This file is licensed under the Apache License v2.0 with LLVM Exceptions.
+# See https://llvm.org/LICENSE.txt for license information.
+# SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+#
+# (c) Copyright 2024 Advanced Micro Devices, Inc. or its affiliates
```
gbossu, line 5: Nothing to do here, I just find it funny to have a copyright header on a README :D

llvm/lib/Target/AIE/AIESlotCounts.h
martien-de-jong: I was mainly thinking about copy-by-value time, but I think I avoided excessive number of those copies algorithmically. I will go to int.

llvm/lib/Target/AIE/AIEPostPipeliner.cpp
```diff
@@ -285,24 +288,55 @@ int PostPipeliner::mostUrgent() {
...
+      N.Earliest = 0;
+      N.Latest = -1;
+    }
+    if (FullReset) {
+      N.NumPushedEarliest = 0;
+      N.NumPushedLatest = 0;
+      N.LastEarliestPusher = -1;
+      N.LastLatestPusher = -1;
+    }
+#endif
```
martien-de-jong: Refactoring glitch. To be corrected

llvm/lib/Target/AIE/AIEPostPipeliner.cpp
```diff
@@ -183,29 +183,35 @@ void PostPipeliner::computeLoopCarriedParameters() {
   }
 
   // Propagate Earliest upstream, initialize Latest
+  // Unrestricted: last cycle of last stage
+  const int Latest = NCopies * II - 1;
   for (int K = 0; K < NInstr; K++) {
     const int K2 = K + NInstr;
```
gbossu: Nit: `KNext` or `KNextIter`?

llvm/lib/Target/AIE/AIEPostPipeliner.cpp
```diff
@@ -183,29 +183,35 @@ void PostPipeliner::computeLoopCarriedParameters() {
...
     const int Earliest = Info[K2].Earliest - II;
     Info[K].Earliest = std::max(Info[K].Earliest, Earliest);
-    // Unrestricted: Beyond the last stage.
-    Info[K].Latest = NCopies * II;
+    Info[K].Latest = Latest;
+    Info[K2].Latest = Latest;
```
gbossu: How can these two nodes have the same `Latest`? Is it a "modulo cycle"?

llvm/lib/Target/AIE/AIEPostPipeliner.cpp
```diff
@@ -183,29 +183,35 @@ void PostPipeliner::computeLoopCarriedParameters() {
...
+  bool Changed = true;
+  while (Changed) {
+    Changed = false;
+    for (int K = NInstr - 1; K >= 0; K--) {
+      SUnit &SU = DAG->SUnits[K];
+      const int Latest = Info[K].Latest;
+      for (auto &Dep : SU.Preds) {
+        int P = Dep.getSUnit()->NodeNum;
+        int NewLatest = Latest - Dep.getSignedLatency();
+        if (NewLatest < Info[P].Latest) {
```
gbossu: I would be happy with a comment. I think this means that the predecessor could be scheduled too late and that would then delay `SU`, which would then not meet its `Latest`.

Also: Could we end up with a predecessor that has a lower `Latest` than `Earliest`? If not, could we add an assert?

llvm/lib/Target/AIE/AIEPostPipeliner.cpp
```diff
@@ -244,20 +255,27 @@ int PostPipeliner::mostUrgent() {
...
+    if (Info[N].Latest < Info[Ref].Latest) {
+      return true;
+    }
+
+    return false;
+  };
+
   int Best = -1;
   LLVM_DEBUG(dbgs() << "Available:");
   for (int K = FirstUnscheduled; K < NInstr; K++) {
```
gbossu: Checking: starting at `FirstUnscheduled` is an optimization, and we could also have scheduled nodes past `FirstUnscheduled`.

llvm/lib/Target/AIE/AIEPostPipeliner.cpp
martien-de-jong: Actually, I think that the K2 value was never used. Probably overcautious in initializing it. Earliest just defaults to zero, Latest is dependent on the length.
In a later commit we switch to a standard relative value of -1.

llvm/lib/Target/AIE/AIEPostPipeliner.cpp
martien-de-jong: Correct. We have an explicit check for the latter case.

llvm/lib/Target/AIE/AIEPostPipeliner.cpp
martien-de-jong: This computation of Latest mimics that of Earliest. If a predecessor pushes SU too far, the schedule doesn't fit in the given linear schedule.

Latest can be smaller then Earliest, and this would say statically that the targeted linear schedule is too short.
Note that we move to an end-of-linear-schedule-relative representation of Latest in a later commit. Latest will be strictly negative, and we round Length up to a multiple of II to make Length + Latest >= Earliest.
Also, we are going to use [Earliest, Length + Latest] to fit it in the modulo scoreboard. If that range becomes empty, we fail, moving on to the next strategy or to a bigger II.

llvm/lib/Target/AIE/AIEPostPipeliner.cpp
```diff
@@ -184,13 +186,12 @@ void PostPipeliner::computeLoopCarriedParameters() {
 
   // Propagate Earliest upstream, initialize Latest
   // Unrestricted: last cycle of last stage
-  const int Latest = NCopies * II - 1;
+  const int Latest = -1;
```
gbossu: Nit: Discussed offline: I'd be happy to have that start at `0` and grow, much like the cycles of the `Bot` zone in `MachineScheduler`

llvm/lib/Target/AIE/AIEPostPipeliner.cpp
```diff
@@ -248,34 +249,36 @@ void dumpGraph(int NInstr, const std::vector<NodeInfo> &Info,
...
-
-    return false;
+    int N = SU->NodeNum;
+    return N < NInstr && !Info[N].Scheduled;
   };
 
   int Best = -1;
   LLVM_DEBUG(dbgs() << "Available:");
-  for (int K = FirstUnscheduled; K < NInstr; K++) {
+  for (int K = FirstUnscheduled; K <= LastUnscheduled; K++) {
```
gbossu, line 404: Super-Nit: Use a `filter_range` to only iterate on the non-scheduled nodes. Then we do not have to verify `Info[K].Scheduled` inside the loop.

llvm/lib/Target/AIE/AIEPostPipeliner.cpp
```diff
@@ -285,24 +288,40 @@ int PostPipeliner::mostUrgent() {
...
 
-bool PostPipeliner::scheduleFirstIteration() {
+void PostPipeliner::resetSchedule(bool FullReset) {
+  Scoreboard.clear();
+  for (int K = 0; K < NTotalInstrs; K++) {
+    auto &N = Info[K];
+    N.Earliest = 0;
+    N.Latest = -1;
+    N.Cycle = 0;
+    N.Scheduled = false;
```
gbossu: How about `Info[K] = NodeInfo();`? Then there is a single place where we write all those default values.

llvm/lib/Target/AIE/AIEPostPipeliner.cpp
```diff
@@ -350,6 +369,16 @@ bool PostPipeliner::scheduleOtherIterations() {
   return true;
 }
 
+class DefaultStrategy : public PostPipelinerStrategy {
```
gbossu, line 516: Nice! I like that this is NFC :)

llvm/lib/Target/AIE/AIEPostPipeliner.h
```diff
@@ -51,6 +51,14 @@ class NodeInfo {
   // The latest cycle at which this can be scheduled. This is a negative value
   // relative to the length of the linear schedule.
   int Latest = -1;
+
+  // Record critical path components
+  // The Pred/Succ that pushed my Earliest/Latest
+  int LastEarliestPusher = -1;
+  int LastLatestPusher = -1;
```
gbossu: Those should be `std::optional` :)

gbossu: Also, why do we compute the "last" pusher, and not the "worst"/"most critical" pusher?

gbossu: Hmm, looking at the code, I think we do find the "worst" pushers.

llvm/lib/Target/AIE/AIEPostPipeliner.cpp
```diff
@@ -411,6 +419,8 @@ bool PostPipeliner::schedule(ScheduleDAGMI &TheDAG, int InitiationInterval) {
   NTotalInstrs = TheDAG.SUnits.size();
   assert(NTotalInstrs % NInstr == 0);
   NCopies = NTotalInstrs / NInstr;
+  LLVM_DEBUG(dbgs() << "PostPipeliner: try scheduling NCopies=" << NCopies
+                    << " x II=" << InitiationInterval << "\n");
```
gbossu: Nit: Maybe squash those logging improvements with those in the previous commit?

llvm/lib/Target/AIE/AIEPostPipeliner.cpp
martien-de-jong: I think a filter range is just a way to verify a condition in a loop. I reserve its use to write a generic iteration template, not to define a single instance.

llvm/lib/Target/AIE/AIEPostPipeliner.cpp
martien-de-jong: But I will need to refactor the static part of Info, otherwise I can't partially reset.

llvm/lib/Target/AIE/AIEPostPipeliner.cpp
martien-de-jong: nit: Since I use Earliest and Latest together to define the interval in which to schedule, the name Latest is a misfit if it runs the other way. It also requires the top-down or bottom up context to apply the proper negations. The code is fully separated from the machine scheduler and the concept fully separated from the picking order.

llvm/lib/Target/AIE/AIEPostPipeliner.cpp
```diff
@@ -188,54 +189,141 @@ int PostPipeliner::fit(MachineInstr *MI, int First, int Last, int II) {
...
-  // Note that we don't have to clean the effects of this exploration,
-  // since the real scheduling will overwrite Cycle, and the ultimate Earliest
-  // will never be less than we compute here.
-
+void PostPipeliner::computeForward() {
+  // The forward order defines a topological sort, so we can compute
+  // Earliest and Ancestors in a single forward sweep
+  // Also initialize Latest to the last cycle in the last stage
+  const int Latest = -1;
   for (int K = 0; K < NInstr; K++) {
```
gbossu, line 200: Nit: I'd find all those traversals much easier to read if we didn't have to use array indexing.
E.g. something like
```
for (NodeInfo &Node : Info) {
  ...
  for (NodeInfo &Pred : Node.preds()) { ... }
}
```

llvm/lib/Target/AIE/AIEPostPipeliner.cpp
```diff
@@ -188,54 +189,141 @@ int PostPipeliner::fit(MachineInstr *MI, int First, int Last, int II) {
...
-  // propagating its Earliest back to the first iteration
-  // Note that we don't have to clean the effects of this exploration,
-  // since the real scheduling will overwrite Cycle, and the ultimate Earliest
-  // will never be less than we compute here.
-
+void PostPipeliner::computeForward() {
+  // The forward order defines a topological sort, so we can compute
+  // Earliest and Ancestors in a single forward sweep
+  // Also initialize Latest to the last cycle in the last stage
+  const int Latest = -1;
```
gbossu: Could this be part of `NodeInfo`'s constructor instead?

llvm/lib/Target/AIE/AIEPostPipeliner.cpp
```diff
@@ -188,54 +189,141 @@ int PostPipeliner::fit(MachineInstr *MI, int First, int Last, int II) {
...
-void PostPipeliner::computeLoopCarriedParameters() {
-  // We schedule the first iteration, only using earliest. This updates
-  // earliest of the successors. Any successor in the second iteration
-  // represents a loop carried dependence, and we account for that by
-  // propagating its Earliest back to the first iteration
-  // Note that we don't have to clean the effects of this exploration,
-  // since the real scheduling will overwrite Cycle, and the ultimate Earliest
-  // will never be less than we compute here.
-
+void PostPipeliner::computeForward() {
```
gbossu, line 197: I think it would make sense to split into two functions: `computeEarliest` and `computeAncestors`, those are distinct loops anyway.

llvm/lib/Target/AIE/AIEPostPipeliner.cpp
```diff
@@ -188,54 +189,141 @@ int PostPipeliner::fit(MachineInstr *MI, int First, int Last, int II) {
...
+  bool Changed = false;
+
+  auto AddOffspring = [&Changed](NodeInfo &Info, int E) {
+    if (Info.Offspring.insert(E).second) {
+      Changed = true;
+    }
+  };
+
+  // Traversing backwards will speed convergence a bit
+  for (int K = NInstr - 1; K >= 0; K--) {
```
gbossu, line 238: Nit: Again, something like `for (NodeInfo &Node : reverse(Info))` would be clearer that having to deal with indices.

llvm/lib/Target/AIE/AIEPostPipeliner.cpp
```diff
@@ -188,54 +189,141 @@ int PostPipeliner::fit(MachineInstr *MI, int First, int Last, int II) {
...
-  // Unrestricted: last cycle of last stage
-  const int Latest = -1;
+bool PostPipeliner::computeBackward() {
+  bool Changed = false;
+
+  auto AddOffspring = [&Changed](NodeInfo &Info, int E) {
+    if (Info.Offspring.insert(E).second) {
+      Changed = true;
+    }
+  };
```
gbossu, line 235: Nice way to hide `Changed` out of the way :)

llvm/lib/Target/AIE/AIEPostPipeliner.cpp
```diff
@@ -188,54 +189,141 @@ int PostPipeliner::fit(MachineInstr *MI, int First, int Last, int II) {
...
+  computeForward();
+
+  // Backward properties like Latest and Offspring.
+  // Use a fixpoint loop, because plain reversed order may not be topological
+  // for predecessors
+  while (computeBackward()) {
+    /* EMPTY */;
+  }
+
+  // Adjust Earliest and Latest with resource requirements
```
gbossu: Nit: mention that this is to bias them towards more realistic values.

llvm/lib/Target/AIE/AIEPostPipeliner.cpp
```diff
@@ -188,54 +189,141 @@ int PostPipeliner::fit(MachineInstr *MI, int First, int Last, int II) {
...
+    SlotCounts OSlots;
+    for (int O : Me.Offspring) {
+      auto &SU = DAG->SUnits[O];
+      OSlots += getSlotCounts(*SU.getInstr(), TII);
+    }
+    LLVM_DEBUG(dbgs() << "SU" << K << " : " << Info[K].Earliest << " - "
+                      << Info[K].Latest << " " << ASlots << " " << OSlots
+                      << "\n");
+    Me.Earliest = std::max(Me.Earliest, ASlots.max());
+    Me.Latest = std::min(Me.Latest, -OSlots.max());
```
gbossu: Not subjective whatsoever: minuses are confusing me :)

llvm/lib/Target/AIE/AIEPostPipeliner.cpp
gbossu: At least all this should be in `NodeInfo::reset()`, not in `PostPipeliner`.

llvm/lib/Target/AIE/AIEPostPipeliner.cpp
```diff
@@ -188,54 +189,141 @@ int PostPipeliner::fit(MachineInstr *MI, int First, int Last, int II) {
...
   }
 
-  LLVM_DEBUG(for (int K = 0; K < NInstr; K++) {
-    dbgs() << "SU" << K << " : " << Info[K].Earliest << " - " << Info[K].Latest
-           << "\n";
-  });
+  // Save the static values for ease of reset
+  for (auto &N : Info) {
+    N.StaticEarliest = N.Earliest;
+    N.StaticLatest = N.Latest;
```
gbossu, line 328: Instead of having `Latest`, `StaticLatest` and `LCDLatests`, could we only have `Latest` and `StaticLatest`? If some strategies need LCDLatest, I guess they could just call something like `biasLatestWithLCDs()`?

llvm/lib/Target/AIE/AIEPostPipeliner.cpp
gbossu: As I said, I won't block the PR for that, I just find the symmetry appealing if we use something like `TopEarliest` that belongs in `[0,Length)`, and `BotLatest` that has the same interval. But that is purely subjective. My only valid argument is that if we go this way, we would mimic the concepts of the `MachineScheduler`, and I think it's important in a large codebase to share concepts.

llvm/lib/Target/AIE/AIEPostPipeliner.cpp
martien-de-jong: Hmm. That might work now, but that is because we don't propagate LCDLatest backwards(yet), which makes it a local, relatively cheap computation. The situation is asymmetric relative to Earliest, where we do propagate it in Earliest itself. The root cause of this asymmetry is that we use the LCDs to constrain the scheduling intervals at the 'Earliest' side. If we would also constrain it at the Latest side, we constrain it twice, which might lead to suboptimality.
Summarizing, LCDLatest is a static heuristic value guiding TopDown scheduling, and might be difficult to compute if we want to propagate backwards. Latest is absolute and modified by the actual scheduling, and StaticLatest provides an expensive static component of Latest.

martien-de-jong: Note that the cost to which 'relatively cheap'  applies here will be taken quite often, since it is computed for every comparison in the available queue.

llvm/lib/Target/AIE/AIEPostPipeliner.cpp
martien-de-jong: It is nothing compared to the confusion that missing minus signs cause me Would you like me to make SlotCounts signed and supply a min() method?

llvm/lib/Target/AIE/AIEPostPipeliner.cpp
martien-de-jong: But reverse Info runs over all Info nodes (NTotalInstr) Moreover, indices connect SUnits and Info

llvm/lib/Target/AIE/AIEPostPipeliner.cpp
martien-de-jong: Yes, now that is relative, that should work.

llvm/lib/Target/AIE/AIEPostPipeliner.h
martien-de-jong: By design, yes. Would a name change help? Suggestions will be given serious consideration.

llvm/lib/Target/AIE/AIEBaseSubtarget.cpp
martien-de-jong: I see no worrying changes in QoR.

llvm/lib/Target/AIE/AIEPostPipeliner.cpp
gbossu: I think this just means we need helpers like `iterator_range<NodeInfo> first_nodes()`, and `SUnit getSUnit(NodeInfo &Node)`. Again, this is a nit, nothing blocking :) But I think this would hide away implementation details and this means I don't have to remember what `K` and `NInstr` mean.

llvm/lib/Target/AIE/AIEPostPipeliner.h
gbossu: I guess I have a slight preference for `Worst` yes, but I don't really mind. Just having `EarliestPusher` also looks good to me. I can't find a better name at least.
Also, could this be a `NodeInfo *EarliestPusher = nullptr` instead of `std::optional<int>`?

llvm/lib/Target/AIE/AIEPostPipeliner.cpp
```diff
@@ -494,6 +499,148 @@ class DefaultStrategy : public PostPipelinerStrategy {
...
+    LCDLatest
+  };
+  static const int Shift = 3;
+  static const int Size = 1 << Shift;
+  static const int Mask = Size - 1;
+
+private:
+  const std::string Name;
+  std::set<int> SuccSiblingScheduled;
+  std::function<bool(const SUnit &A, const SUnit &B)> Discriminators[Size] = {
```
gbossu: Nit: Might be more readable to define all those functions separately, and then have e.g.
```
std::function<bool(const SUnit &A, const SUnit &B)> Discriminators[Size] = {
  &CompareNone,
  &CompareNodeNum,
  &CompareLatest,
  ...
};
```

llvm/lib/Target/AIE/AIEPostPipeliner.cpp
```diff
@@ -494,6 +499,148 @@ class DefaultStrategy : public PostPipelinerStrategy {
...
+
+  DEBUG_SUMMARY(dbgs() << "-- MinLength=" << MinLength << "\n");
+
+  static const std::tuple<int, bool, uint64_t> Configs[] = {
+      // Loosely speaking, a lower value of the first parameter targets
+      // a lower stage count, which benefits code size.
+      {0, false, ConfigStrategy::Latest},
+  };
+  int H = 0;
+  for (auto &[Extra, Rerun, Param] : Configs) {
```
gbossu: Nit: Rename `Extra` into `ExtraStages`? Or better, use a struct with named attributes instead of a tuple :)

llvm/lib/Target/AIE/AIEPostPipeliner.cpp
```diff
@@ -494,6 +499,148 @@ class DefaultStrategy : public PostPipelinerStrategy {
...
+bool PostPipeliner::tryHeuristics() {
+  // The minimum length makes sure that every node has a range in which it
+  // can be scheduled
+  int MinLength = II;
+  for (int K = 0; K < NInstr; K++) {
+    auto &Node = Info[K];
+    while (Node.Earliest > Node.Latest + MinLength) {
+      MinLength += II;
+    }
+  }
```
gbossu: I find this a bit confusing to have in `tryHeuristics`. Should that be its own function like `computeMinLength` called within `schedule()` itself?

llvm/lib/Target/AIE/AIEPostPipeliner.cpp
```diff
@@ -494,6 +499,148 @@ class DefaultStrategy : public PostPipelinerStrategy {
...
+
+  DEBUG_SUMMARY(dbgs() << "-- MinLength=" << MinLength << "\n");
+
+  static const std::tuple<int, bool, uint64_t> Configs[] = {
+      // Loosely speaking, a lower value of the first parameter targets
+      // a lower stage count, which benefits code size.
+      {0, false, ConfigStrategy::Latest},
+  };
+  int H = 0;
```
gbossu: Nit: `HeuristicIndex`?

llvm/lib/Target/AIE/AIEPostPipeliner.cpp
```diff
@@ -494,6 +499,148 @@ class DefaultStrategy : public PostPipelinerStrategy {
   }
 };
 
+class ConfigStrategy : public PostPipelinerStrategy {
+public:
+  // None serves as a terminator; any component after it will be ignored.
+  enum PriorityComponent {
```
gbossu, line 530: Nit: Use an `enum class`?

llvm/lib/Target/AIE/AIEPostPipeliner.cpp
```diff
@@ -494,6 +499,148 @@ class DefaultStrategy : public PostPipelinerStrategy {
...
+    None,
+    NodeNum,
+    Latest,
+    Critical,
+    Sibling,
+    LCDLatest
+  };
+  static const int Shift = 3;
+  static const int Size = 1 << Shift;
+  static const int Mask = Size - 1;
```
gbossu: Magic values are always hard to update. As far as I understand, these are essentially implementing a vector of small types, `int4` here. But then when building the `ConfigStrategy`, we anyway convert that "vector-ish" into a `vector<int> Priotity`. How about getting rid of that size optimization, and passing an `ArrayRef<PriorityComponent>` when constructing the `ConfigStrategy`?

llvm/lib/Target/AIE/AIEPostPipeliner.cpp
```diff
@@ -519,22 +521,27 @@ class ConfigStrategy : public PostPipelinerStrategy {
...
   const std::string Name;
   std::set<int> SuccSiblingScheduled;
+  std::set<int> PredSiblingScheduled;
   std::function<bool(const SUnit &A, const SUnit &B)> Discriminators[Size] = {
       [&](const SUnit &A, const SUnit &B) { return false; },
-      [&](const SUnit &A, const SUnit &B) { return A.NodeNum < B.NodeNum; },
+      [&](const SUnit &A, const SUnit &B) {
+        return TopDown ? A.NodeNum < B.NodeNum : A.NodeNum > B.NodeNum;
+      },
       [&](const SUnit &A, const SUnit &B) {
```
gbossu: Nit: The name of that priority function is now a bit confusing, it's only `Latest` for the top-down mode.

llvm/lib/Target/AIE/AIEPostPipeliner.cpp
```diff
@@ -546,6 +553,8 @@ class ConfigStrategy : public PostPipelinerStrategy {
   };
   std::vector<int> Priority;
 
+  bool fromTop() override { return TopDown; }
```
gbossu, line 589: Nit: `isTopDown()`?

llvm/lib/Target/AIE/AIEPostPipeliner.cpp
```diff
@@ -606,21 +627,27 @@ bool PostPipeliner::tryHeuristics() {
...
 
-  static const std::tuple<int, bool, uint64_t> Configs[] = {
+  static const std::tuple<int, bool, bool, uint64_t> Configs[] = {
       // Loosely speaking, a lower value of the first parameter targets
       // a lower stage count, which benefits code size.
-      {0, false, ConfigStrategy::NodeNum},
-      {0, false, ConfigStrategy::Latest},
-      {0, true, ConfigStrategy::Critical},
-      {0, true,
+      {0, true, false, ConfigStrategy::NodeNum},
```
gbossu: Nit: `{/*ExtraStages=*/0, /*TopDown=*/true, /*Rerun=*/false, ConfigStrategy::NodeNum}`? Or at least give a top-level comment documenting the elements of the tuple?

llvm/lib/Target/AIE/AIEBaseSubtarget.cpp
gbossu: But did you compare against plain `aie-public`? From what I understand the numbers you shared were with the pre-pipeliner disabled. If the latter is enabled, there could be QoR changing when enabling that new mutator for the postmisched as well.

llvm/lib/Target/AIE/AIEPostPipeliner.h
```diff
@@ -43,14 +44,65 @@ class NodeInfo {
...
+  ScheduleDAGInstrs &DAG;
+  std::vector<NodeInfo> &Info;
+  int LatestBias = 0;
+
+public:
+  PostPipelinerStrategy(ScheduleDAGInstrs &DAG, std::vector<NodeInfo> &Info,
+                        int LatestBias)
+      : DAG(DAG), Info(Info), LatestBias(LatestBias) {};
+  virtual ~PostPipelinerStrategy() {};
+  virtual std::string name() { return "PostPipelinerStrategy"; }
```
andcarminati, line 105: nit: maybe a `StringRef` can be used here.

llvm/lib/Target/AIE/AIESlotCounts.cpp
```diff
@@ -0,0 +1,72 @@
+//===- AIESlotCounts.cpp - SlotCount utility ------------------------------===//
```
andcarminati, line 1: nit: this looks like something for the `Utils` folder.

llvm/lib/Target/AIE/AIEPostPipeliner.cpp
gbossu: Checking again because I got confused: I think the computation is not symmetric.

As far as I understand, `Latest` belongs to `[-Length, -1]`, if our offsprings require e.g. 2 MOV slots, then (assuming a latency of 0 to those successors, and `Me` does not use the MOV slot), then the last feasible cycle is indeed `-2 = ASlots.max()`, because that is the second cycle counting from the bottom. 

For top-down, `Earliest` belongs to `[0, Length - 1]`. Using the same assumptions and 2 MOV slots booked by Ancestors, then `Me` should have an `Earlierst` of `1` which is the second cycle counting from the top. Not `ASlots.max()` which would be `2`. Just saying, but having positive "bottom-up cycles" like in the `MachineScheduler` would have effortlessly given symmetric results :trollface: 

What I would suggest is:
 - Include the slots of `Me` in `ASlots` and `OSlots`. Then we can get rid of the assumptions about latency and the slots booked by `Me`
 - Then we can compute:
   -  `EarliestCycle = FirstCycle + (MinRequiredCycles - 1) = 0 + ASlots.max() - 1`
   - `LatestCycle = LastCycle - (MinRequiredCycles - 1) = -1 - (OSlots.max() - 1) = -OSlots.max()`

Please double check me, my brain has been wired another way after looking into the MachineScheduler too much 😆

gbossu: Also, I think those computations assume no negative latencies. Otherwise, we cannot say much with SlotCounts because e.g. predecessors are not necessarily scheduled before `Me`. Maybe note that down, that we are aware of this, and still think that these metrics are good enough for the main cases?

llvm/lib/Target/AIE/AIEPostPipeliner.h
```diff
@@ -43,14 +44,65 @@ class NodeInfo {
...
+  // Choose among available alternatives
+  virtual bool better(const SUnit &A, const SUnit &B) { return false; }
+  // Tweak the effective earliest
+  virtual int earliest(const SUnit &N) { return Info[N.NodeNum].Earliest; }
+  // Select from top or from bottom.
```
gbossu: I think that comment is mis-placed

llvm/lib/Target/AIE/AIEPostPipeliner.h
```diff
@@ -43,14 +44,65 @@ class NodeInfo {
...
+public:
+  PostPipelinerStrategy(ScheduleDAGInstrs &DAG, std::vector<NodeInfo> &Info,
+                        int LatestBias)
+      : DAG(DAG), Info(Info), LatestBias(LatestBias) {};
+  virtual ~PostPipelinerStrategy() {};
+  virtual std::string name() { return "PostPipelinerStrategy"; }
+  // Choose among available alternatives
+  virtual bool better(const SUnit &A, const SUnit &B) { return false; }
+  // Tweak the effective earliest
+  virtual int earliest(const SUnit &N) { return Info[N.NodeNum].Earliest; }
```
gbossu, line 109: Nit: The docstring makes it seem like `earliest()` is tweaking the value. Instead, I believe it is return a value that has already been tweaked. Could this be `const`?

Same for `latest()` below.

llvm/lib/Target/AIE/AIEPostPipeliner.h
```diff
@@ -43,14 +44,65 @@ class NodeInfo {
...
   // This includes the lowerbound of the modulo condition, i.e.
   // Earliest(N) >= Cycle(N - NInstr) + II
   int Earliest = 0;
-  // For an LCD K1 -> K2, this holds II + Earliest(K2 - NInstr) - Latency(LCD)
-  // Instructions with lower Latest have higher priority in the
-  // top down scheduling
-  int Latest = 0;
+
+  // The latest cycle at which this can be scheduled. This is a negative value
+  // relative to the length of the linear schedule.
```
gbossu, line 55: Nit: Add a note to say that `-1` represents the last cycle of the schedule, while `-Length` represents the first one.

llvm/lib/Target/AIE/AIEPostPipeliner.h
```diff
@@ -43,14 +44,65 @@ class NodeInfo {
...
+
+  // The transitive closure of my predecessors
+  std::unordered_set<int> Ancestors;
+
+  // The transitive closure of my successors
+  std::unordered_set<int> Offspring;
+
+  /// Reset the node to the values computed statically
+  /// If FullReset is true, also reset the accumulated dynamic data.
+  void reset(bool FullReset);
```
gbossu, line 90: Super-nit: `void reset(bool IncludeDynamic);`

llvm/lib/Target/AIE/AIEPostPipeliner.cpp
```diff
@@ -274,24 +397,41 @@ int PostPipeliner::mostUrgent() {
...
-bool PostPipeliner::scheduleFirstIteration() {
+void PostPipeliner::resetSchedule(bool FullReset) {
+  Scoreboard.clear();
+  for (int K = 0; K < NTotalInstrs; K++) {
+    auto &N = Info[K];
+    N.reset(FullReset);
+    if (K < NInstr) {
+      N.Earliest = N.StaticEarliest;
+      N.Latest = N.StaticLatest;
+    }
```
gbossu, line 430: Does that imply that nodes outside of the first iteration never get an `Earliest` or `Latest`? Should we make that explicit with an `optional`?

llvm/lib/Target/AIE/AIEPostPipeliner.cpp
```diff
@@ -274,24 +397,41 @@ int PostPipeliner::mostUrgent() {
   return Best;
 }
 
-bool PostPipeliner::scheduleFirstIteration() {
+void PostPipeliner::resetSchedule(bool FullReset) {
+  Scoreboard.clear();
+  for (int K = 0; K < NTotalInstrs; K++) {
+    auto &N = Info[K];
+    N.reset(FullReset);
```
gbossu, line 426: Shouldn't a "full" reset also clear the "static" variables? Or do we always want to keep those because they never change, even when targeting a different II?

llvm/lib/Target/AIE/AIEPostPipeliner.h
```diff
@@ -43,14 +44,65 @@ class NodeInfo {
   int ModuloCycle = 0;
   // Cycle / II
   int Stage = 0;
+
+  int StaticEarliest = 0;
+  int StaticLatest = -1;
```
gbossu: Can you document those? In particular what "static" means in that context. I.e. is it invariant in the strategy and the II? Or one of them?

llvm/lib/Target/AIE/AIEPostPipeliner.cpp
```diff
@@ -306,12 +446,23 @@ bool PostPipeliner::scheduleFirstIteration() {
...
+    DEBUG_FULL(dbgs() << "Scoreboard\n"; Scoreboard.dumpFull(););
   }
-  LLVM_DEBUG(dbgs() << "==== First iteration scheduled ======\n");
+  LLVM_DEBUG(dbgs() << "==== First iteration scheduled by " << Strategy.name()
+                    << "====\n");
   return true;
 }
 
+namespace {
+void dumpEarliestChain(const std::vector<NodeInfo> &Info, int N) {
```
gbossu, line 479: Checking: I think we are safe because all those `dump` helpers seem to be in anonymous namespaces, but maybe adding a `[[nodiscard]]` would be safer?

gbossu (commented):
I think I'm done with the review, let me know if you want to discuss any of the comments :)

llvm/lib/Target/AIE/AIEPostPipeliner.cpp
martien-de-jong: They get a default 0, -1 from reset, which are sane values, not sentinels. Earliest of the second iteration is used to represent LCD latencies, but this is propagated back to the first iteration.

llvm/lib/Target/AIE/AIEPostPipeliner.h
martien-de-jong: Well, these default implementations could be const, but I don't want to block a scheme where some user state  is maintained.

llvm/lib/Target/AIE/AIESlotCounts.cpp
martien-de-jong: Except that it rests on AIE facilities.

llvm/lib/Target/AIE/AIEPostPipeliner.h
martien-de-jong: But I won't. It's for debugging purposes only and it is a nice stand-alone value.

llvm/lib/Target/AIE/AIEPostPipeliner.cpp
martien-de-jong: They would have to be methods, since they capture e.g. Info.

llvm/lib/Target/AIE/AIEPostPipeliner.cpp
martien-de-jong: Yeah, historic mishap.

llvm/lib/Target/AIE/AIEPostPipeliner.cpp
gbossu: Or just named lambdas within the constructor or wherever we fill in `Discriminators`. Anyway, it's just a nit, but I feel it will become useful to have names when we start to have dozens of compare functions.

llvm/lib/Target/AIE/AIEPostPipeliner.cpp
martien-de-jong: They change on a new II, but then the entire array is cleared and reallocated.

llvm/lib/Target/AIE/AIEPostPipeliner.cpp
martien-de-jong: As discussed, MinLength is local to the strategies. But will factor into separate method.

llvm/lib/Target/AIE/AIEPostPipeliner.cpp
```diff
@@ -174,49 +194,151 @@ int PostPipeliner::fit(MachineInstr *MI, int Earliest, int NTries, int II) {
...
+
+  // Save the static values for ease of reset
+  for (auto &N : Info) {
+    N.StaticEarliest = N.Earliest;
+    N.StaticLatest = N.Latest;
+  }
+  return true;
+}
+
+int PostPipeliner::computeMinScheduleLength() {
```
gbossu: I think that could be const?

llvm/test/CodeGen/AIE/aie2/schedule/postpipeliner/conv2d_bf16-1.mir
```diff
@@ -0,0 +1,238 @@
+# NOTE: Assertions have been autogenerated by utils/update_llc_test_checks.py UTC_ARGS: --version 4
+# This file is licensed under the Apache License v2.0 with LLVM Exceptions.
+# See https://llvm.org/LICENSE.txt for license information.
+# SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+#
+# (c) Copyright 2024 Advanced Micro Devices, Inc. or its affiliates
+
+# RUN: llc --mtriple=aie2 -O2 --start-before=postmisched %s \
+# RUN:   --debug-only=postpipeliner-summary -o - | FileCheck %s
```
andcarminati, line 9: What is the purpose of `--debug-only=postpipeliner-summary` in this test?

martien-de-jong: developer convenience. update_llc_tests shows it on stderr, and the last line reports the final II.

llvm/lib/Target/AIE/AIEPostPipeliner.cpp
```diff
@@ -503,6 +523,153 @@ class DefaultStrategy : public PostPipelinerStrategy {
...
+public:
+  enum PriorityComponent {
+    NodeNum,
+    Latest,
+    Critical,
+    Sibling,
+    LCDLatest,
+    Size
+  };
+  std::string getPriorityName(PriorityComponent Component) {
```
gbossu: Nit: This should be const, maybe even static.

llvm/lib/Target/AIE/AIEPostPipeliner.h
```diff
@@ -43,14 +45,77 @@ class NodeInfo {
...
+  // Record critical path components
+  // The Pred/Succ that pushed my Earliest/Latest
+  std::optional<int> LastEarliestPusher;
+  std::optional<int> LastLatestPusher;
+  // The number of Succs/Preds whose Earliest/Latest I have pushed.
+  int NumPushedEarliest = 0;
+  int NumPushedLatest = 0;
+
+  // Latest corrected by taking Earliest of an LCD successor into account
+  int LCDLatest = -1;
```
gbossu, line 80: Any reason not to have a symmetric `LCDEarliest` for bottom-up scheduling?

llvm/lib/Target/AIE/AIEPostPipeliner.cpp
```diff
@@ -160,8 +178,10 @@ void PostPipeliner::scheduleNode(SUnit &SU, int Cycle) {
 // Check resources. We only insert at the position modulo II. Since we insert
 // all iterations separately, the resources that wrap around accumulate in the
 // overflow area, causing conflicts when inserting future iterations
-int PostPipeliner::fit(MachineInstr *MI, int Earliest, int NTries, int II) {
-  for (int C = Earliest; C < Earliest + NTries; C++) {
+int PostPipeliner::fit(MachineInstr *MI, int First, int Last, int II) {
+  const int Step = First > Last ? -1 : 1;
+  LLVM_DEBUG(dbgs() << "   " << First << ", " << Last << ", " << Step << "\n");
+  for (int C = First; C != Last; C += Step) {
```
gbossu, line 184: Maybe have `C < Last`? Or do we have a guarantee that `Step` divides `Last-First`?

llvm/lib/Target/AIE/AIEPostPipeliner.h
martien-de-jong: No. That's an oversight.

llvm/lib/Target/AIE/AIEPostPipeliner.cpp
martien-de-jong: Eh, step is -1 or 1, so divides everything. But C < Last doesn't work if Step == -1

gbossu: Oh boy, I thought that `Step` came in as an argument, but it's just defined one line above 🙈

llvm/lib/Target/AIE/AIEPostPipeliner.h
gbossu: But I guess we would not have a strategy to make use of it anyway?

gbossu (approved):
LGTM! I think this adds a lot of flexibility to the pipeliner and I'm sure we'll do great things with it 💪

llvm/lib/Target/AIE/AIEPostPipeliner.h
martien-de-jong: Correct. Leaving it for now, but we should keep in mind that LcdLatest is asymmetric wrt TopDown now.

PR #243, opened 2024-11-21
Add support for bfloat in VInsert PreLegalizerCombiner

abhinay-anubola:
Add support for bfloat in SetExtract PreLegalizerCombiner
Add support for bfloat in VInsert PreLegalizerCombiner


llvm/lib/Target/AIE/AIE2PreLegalizerCombiner.cpp
```diff
@@ -149,6 +149,19 @@ bool AIE2PreLegalizerCombinerImpl::tryToCombineVectorShiftsByZero(
 
 bool AIE2PreLegalizerCombinerImpl::tryToCombineSetExtract(
     MachineInstr &MI) const {
+  // Compares the index registers and their associated constant values from two
+  // given machine instructions.
+  auto IsEqualIdx = [&](const MachineInstr *SetMI, const MachineInstr *ExtOp) {
```
andcarminati: What about:

```
  auto IsEqualIdx = [&](const MachineInstr *SetMI, const MachineInstr *ExtMI) {
    const Register SetMIIdxReg = SetMI->getOperand(3).getReg();
    const Register ExtMIIdxReg = ExtMI->getOperand(3).getReg();
    if (SetMIIdxReg == ExtMIIdxReg)
      return true;
    auto SetMICst = getIConstantVRegValWithLookThrough(SetMIIdxReg, MRI);
    auto ExtMICst = getIConstantVRegValWithLookThrough(ExtMIIdxReg, MRI);
    return (SetMICst && ExtMICst &&
            SetMICst->Value.getZExtValue() == ExtMICst->Value.getZExtValue());
  };
```

llvm/lib/Target/AIE/AIE2PreLegalizerCombiner.cpp
```diff
@@ -149,6 +149,18 @@ bool AIE2PreLegalizerCombinerImpl::tryToCombineVectorShiftsByZero(
...
     MachineInstr &MI) const {
+  // Compares the index registers and their associated constant values from two
+  // given machine instructions.
+  auto IsEqualIdx = [&](const MachineInstr *SetMI, const MachineInstr *ExtMI) {
+    const Register SetMIIdxReg = SetMI->getOperand(3).getReg();
+    const Register ExtMIIdxReg = ExtMI->getOperand(3).getReg();
+    if (SetMIIdxReg == ExtMIIdxReg)
+      return true;
+    auto SetMICst = getIConstantVRegValWithLookThrough(SetMIIdxReg, MRI);
+    auto ExtMICst = getIConstantVRegValWithLookThrough(ExtMIIdxReg, MRI);
```
martien-de-jong: nit: declare them explicitly, or at least as auto *?

martien-de-jong: nit: first test SetMICst before finding ExtMICst

llvm/lib/Target/AIE/AIE2PreLegalizerCombiner.cpp
```diff
@@ -162,18 +174,18 @@ bool AIE2PreLegalizerCombinerImpl::tryToCombineSetExtract(
...
-    const Register ExtOpIdxReg = ExtOp->getOperand(3).getReg();
-    auto SetOpCst = getIConstantVRegValWithLookThrough(SetOpIdxReg, MRI);
-    auto ExtOpCst = getIConstantVRegValWithLookThrough(ExtOpIdxReg, MRI);
-    if (SetOpIdxReg != ExtOpIdxReg &&
-        (!SetOpCst || !ExtOpCst ||
-         SetOpCst->Value.getZExtValue() != ExtOpCst->Value.getZExtValue()))
-      return false;
-    break;
+    if (cast<GIntrinsic>(*ExtOp).getIntrinsicID() ==
+            Intrinsic::aie2_ext_I256_I512 &&
```
martien-de-jong: nit: I think a utility like isIntrinsicID(ExtOp, Intrinsic::aie2_ext_I256_I512) would help readability

llvm/lib/Target/AIE/AIE2PreLegalizerCombiner.cpp
```diff
@@ -162,18 +174,18 @@ bool AIE2PreLegalizerCombinerImpl::tryToCombineSetExtract(
...
-    const Register SetOpIdxReg = MI.getOperand(3).getReg();
-    const Register ExtOpIdxReg = ExtOp->getOperand(3).getReg();
-    auto SetOpCst = getIConstantVRegValWithLookThrough(SetOpIdxReg, MRI);
-    auto ExtOpCst = getIConstantVRegValWithLookThrough(ExtOpIdxReg, MRI);
-    if (SetOpIdxReg != ExtOpIdxReg &&
-        (!SetOpCst || !ExtOpCst ||
-         SetOpCst->Value.getZExtValue() != ExtOpCst->Value.getZExtValue()))
-      return false;
-    break;
+    if (cast<GIntrinsic>(*ExtOp).getIntrinsicID() ==
```
martien-de-jong: CHECK are we assuming that ExtOp is a GIntrinsic here? should we check it?

llvm/test/CodeGen/AIE/aie2/GlobalISel/combine-set-extract-prelegalizer.mir
```diff
@@ -83,3 +83,63 @@ body:             |
...
+    ; CHECK-NEXT: [[INT:%[0-9]+]]:_(<8 x s32>) = G_INTRINSIC intrinsic(@llvm.aie2.ext.bf256.bf512), [[COPY]](<16 x s32>), [[C]](s32)
+    ; CHECK-NEXT: [[INT1:%[0-9]+]]:_(<16 x s32>) = G_INTRINSIC intrinsic(@llvm.aie2.set.bf512.bf256), [[INT]](<8 x s32>), [[C1]](s32)
+    ; CHECK-NEXT: PseudoRET implicit $lr, implicit [[INT1]](<16 x s32>)
+    %0:_(<16 x s32>) = COPY $x0
+    %1:_(s32) = G_CONSTANT i32 0
+    %10:_(s32) = G_CONSTANT i32 1
+    %2:_(<8 x s32>) = G_INTRINSIC intrinsic(@llvm.aie2.ext.bf256.bf512), %0(<16 x s32>), %1(s32)
+    %3:_(<16 x s32>) = G_INTRINSIC intrinsic(@llvm.aie2.set.bf512.bf256), %2(<8 x s32>), %10(s32)
+    PseudoRET implicit $lr, implicit %3
+...
```
martien-de-jong: Add a test where we don't have an extract as operand?

llvm/lib/Target/AIE/AIE2PreLegalizerCombiner.cpp
krishnamtibrewala: I guess why @martien-de-jong is suggesting to first test for SetMICst before finding ExtMICst here is `getIConstantVRegValWithLookThrough` could be an expensive call due to the while loop

Function src code : https://llvm.org/doxygen/CodeGen_2GlobalISel_2Utils_8cpp_source.html#l00336

llvm/lib/Target/AIE/AIE2PreLegalizerCombiner.cpp
abhinay-anubola: Checking it above
`if (!isa<GIntrinsic>(MI) || !isa<GIntrinsic>(*ExtOp))
    return false;`

llvm/lib/Target/AIE/AIE2PreLegalizerCombiner.cpp
abhinay-anubola: Existing `isIntrinsicID` checks whether the MachineOperand is IntrinsicID or not.
Do I need to define new function for our use case?

llvm/lib/Target/AIE/AIE2PreLegalizerCombiner.cpp
abhinay-anubola: auto has been used everytime where we are using `getIConstantVRegValWithLookThrough`

gbossu: What is the type?

gbossu (commented):
I think this looks good, but I'm wondering if getting rid of `aie2_set_bf512_bf256` in the frontend and using `aie2_set_I512_I256` wouldn't be better

llvm/lib/Target/AIE/AIE2PreLegalizerCombiner.cpp
abhinay-anubola: `std::optional<ValueAndVReg>`

gbossu: It's not too bad, but yeah I guess it's okay to keep the `auto` here

llvm/lib/Target/AIE/AIE2PreLegalizerCombiner.cpp
martien-de-jong: If I see `cast<GIntrinsic>(*ExtOp).getIntrinsicID()` I get nervous, since the cast may fail if ExtOp isn't a GIntrinsic. The utility is a specialization that allows you to match a specific intrinsic in one line that can be trusted to have properly matched a GIntrinsic. I propose to create it and promote its use as much as possible.
I'm not against overloading and for the present case it feels intuitive, but perhaps a different name like `matchIntrinsic` would be more acceptable.

clang/lib/Headers/aiev2_upd_ext.h
```diff
@@ -527,9 +527,9 @@ INTRINSIC(v32bfloat16) insert(v32bfloat16 a, int idx, v16bfloat16 b) {
 // Set 256-bit portion of 512-bit register
 INTRINSIC(v32bfloat16) set_v32bfloat16(int idx, v16bfloat16 b) {
   if (idx == 0)
-    return __builtin_aiev2_set_bf512_bf256(b, 0);
+    return __builtin_aiev2_set_I512_I256(b, 0);
```
andcarminati, line 530: This looks curious, all values different from 0 can be used as index `1`. I was wondering why not simply:
```
return __builtin_aiev2_set_I512_I256(b, idx == 0 ? 0 : 1);
```

gbossu: I think those headers were partly auto-generated, and for some intrinsics we have more than two different values. So I'd assume it's easier to write a generic generator if sticking to `if`/`else`.

llvm/lib/Target/AIE/AIECombinerHelper.cpp
```diff
@@ -145,25 +145,27 @@ bool llvm::canAdvanceOp(MachineInstr &MemI, MachineInstr &Dest,
 
 /// Find the def instruction for \p Reg, folding away any trivial copies and
 /// bitcasts. May return nullptr if \p Reg is not a generic virtual register.
+/// The \p AllowMultiUse flag permits folding even if the def instruction for \p
+/// Reg, has multiple uses.
 MachineInstr *
-llvm::getDefIgnoringCopiesAndBitcasts(Register Reg,
+llvm::getDefIgnoringCopiesAndBitcasts(Register Reg, bool AllowMultiUse,
                                       const MachineRegisterInfo &MRI) {
```
gbossu, line 230: You seem to never use `AllowMultiUse=true`. Do you have future plans outside of this PR?

gbossu: Oh okay there is one use in `matchExtractConcat`. I would be curious to see if allowing multiple uses in other combiners would help as well?

llvm/lib/Target/AIE/AIECombinerHelper.cpp
```diff
@@ -95,6 +95,10 @@ bool isNonCoalesceableUseOf(const MachineInstr &MemI,
         MRI.getVRegDef(InBetweenMI.getOperand(1).getReg());
     const MachineInstr *CopyDestMI =
         &*MRI.use_instr_nodbg_begin(InBetweenMI.getOperand(0).getReg());
+    // Look through the bitcast and update CopyDestMI to the first use.
+    if (CopyDestMI && CopyDestMI->getOpcode() == TargetOpcode::G_BITCAST)
+      CopyDestMI =
+          &*MRI.use_instr_nodbg_begin(CopyDestMI->getOperand(0).getReg());
```
gbossu: Why not use `getDefIgnoringCopiesAndBitcasts`?

llvm/lib/Target/AIE/AIE2PreLegalizerCombiner.cpp
```diff
@@ -191,6 +191,7 @@ unsigned getVInsertScalarSize(unsigned IntrinsicID) {
   case Intrinsic::aie2_vinsert8_I512:
     return 8;
   case Intrinsic::aie2_vinsert16_I512:
+  case Intrinsic::aie2_vinsert16_bf512:
```
gbossu, line 155: I think we should also get rid of that intrinsic at some point

llvm/lib/Target/AIE/AIECombinerHelper.cpp
abhinay-anubola: I think we can't use `getDefIgnoringCopiesAndBitcasts`, as we are fetching the use of CopyDestReg not def.

gbossu: You could create `getUserIgnoringCopiesAndBitcasts` :) Also, do we expect a single user?

PR #244, opened 2024-11-21
[DCE] Ensure reserved simplifiable registers are live across call boundaries.

SagarMaheshwari99:


llvm/test/CodeGen/AIE/aie2/live-reserved-regs-call.ll
```diff
@@ -0,0 +1,55 @@
...
+    %0 = tail call noundef <16 x i64> @llvm.aie2.v32acc32()
+	%1 = tail call noundef <32 x i8> @llvm.aie2.I256.v32.acc32.srs(<16 x i64> %0, i32 1, i32 0)
+	ret void
+}
+
+declare <32 x i8> @llvm.aie2.I256.v32.acc32.srs(<16 x i64>, i32, i32)
+
+declare <16 x i64> @llvm.aie2.v32acc32()
+
+declare void @llvm.aie2.set.ctrl.reg(i32, i32)
```
gbossu, line 64: I'm fine with these tests, but please also write an `.mir` "unit-test" that just runs the DCE pass. See `llvm/test/CodeGen/AIE/aie2/GlobalISel/dead-mi-elim.mir`

llvm/lib/CodeGen/DeadMachineInstructionElim.cpp
```diff
@@ -157,6 +157,15 @@ bool DeadMachineInstructionElimImpl::runImpl(MachineFunction &MF,
...
   bool AnyChanges = false;
 
+  // Cache the reserved registers that can be simplified.
+  BitVector ReservedRegs = MRI->getReservedRegs();
+  SmallVector<MCPhysReg, 16> SimplifiableReservedRegs;
+  for (MCPhysReg PhysReg : ReservedRegs.set_bits()) {
+    if (MRI->canSimplifyPhysReg(PhysReg)) {
+      SimplifiableReservedRegs.push_back(PhysReg);
+    }
+  }
```
gbossu, line 165: Nit: outline in something like `SmallVector<MCPhysReg, 16> getSimplifiableReservedRegs(const MachineRegisterInfo &MRI);`

llvm/test/CodeGen/AIE/aie2/live-reserved-regs-call.ll
```diff
@@ -0,0 +1,55 @@
+; NOTE: Assertions have been autogenerated by utils/update_llc_test_checks.py UTC_ARGS: --version 4
+; RUN: llc -mtriple=aie2 -O2 --issue-limit=1 -verify-machineinstrs %s -o - 2>&1 | FileCheck %s
+
+define void @my_test() {
```
krishnamtibrewala: super nit : better name of the function

llvm/test/CodeGen/AIE/aie2/live-reserved-regs-call.ll
```diff
@@ -0,0 +1,55 @@
+; NOTE: Assertions have been autogenerated by utils/update_llc_test_checks.py UTC_ARGS: --version 4
+; RUN: llc -mtriple=aie2 -O2 --issue-limit=1 -verify-machineinstrs %s -o - 2>&1 | FileCheck %s
+
```
krishnamtibrewala, line 9: please add a comment describing what we test here, after months just by file name it becomes tricky to remember the use/need of the test :)

llvm/test/CodeGen/AIE/aie2/live-reserved-regs-call.ll
```diff
@@ -0,0 +1,55 @@
+; NOTE: Assertions have been autogenerated by utils/update_llc_test_checks.py UTC_ARGS: --version 4
+; RUN: llc -mtriple=aie2 -O2 --issue-limit=1 -verify-machineinstrs %s -o - 2>&1 | FileCheck %s
```
andcarminati, line 8: Please, add the copyright note here.

llvm/test/CodeGen/AIE/aie2/live-reserved-regs-call.ll
SagarMaheshwari99: done.

llvm/test/CodeGen/AIE/aie2/live-reserved-regs-call.ll
SagarMaheshwari99: done.

llvm/lib/CodeGen/DeadMachineInstructionElim.cpp
SagarMaheshwari99: done.

llvm/test/CodeGen/AIE/aie2/live-reserved-regs-call.ll
SagarMaheshwari99: done.

llvm/test/CodeGen/AIE/aie2/live-reserved-regs-call.ll
SagarMaheshwari99: Added.

llvm/test/CodeGen/AIE/aie2/GlobalISel/dead-mi-elim.mir
```diff
@@ -547,3 +536,82 @@ body:             |
...
+    $crsat = COPY %0:er
+    $crupssign = COPY %1:er
+    PseudoJ_jump_imm %bb.1
+
+  bb.1:
+    liveins: $p0
+    %2:mss = COPY %0:er
+    %3:ep = COPY $p0
+    %4:acc1024 = VLDA_UPS_S32_D8_ag_idx_imm %2:mss, %3:ep, 0, implicit-def $srups_of, implicit $crsat, implicit $crupssign
+    $crupssign = MOV_scalar_imm10_pseudo 0
```
gbossu, line 616: Can you add a test with a call instruction? E.g. `PseudoJL 0`

SagarMaheshwari99: Right, added.

llvm/test/CodeGen/AIE/aie2/GlobalISel/dead-mi-elim.mir
```diff
@@ -547,3 +536,106 @@ body:             |
...
+    ; CHECK: liveins: $r1, $p0
+    ; CHECK-NEXT: {{  $}}
+    ; CHECK-NEXT: [[MOV_RLC_imm10_pseudo:%[0-9]+]]:er = MOV_RLC_imm10_pseudo 0
+    ; CHECK-NEXT: $crsat = COPY [[MOV_RLC_imm10_pseudo]]
+    ; CHECK-NEXT: $crrnd = COPY [[MOV_RLC_imm10_pseudo]]
+    ; CHECK-NEXT: PseudoJL 0
+    %0:er = MOV_RLC_imm10_pseudo 0
+    $crsat = COPY %0:er
+    $crrnd = COPY %0:er
+    PseudoJL 0
```
gbossu: That test seems to already work even without the fix?

gbossu: Maybe you need to add a regmask?

llvm/lib/CodeGen/DeadMachineInstructionElim.cpp
```diff
@@ -165,10 +180,8 @@ bool DeadMachineInstructionElimImpl::eliminateDeadMI(MachineFunction &MF) {
 
     // Reserved registers are considered always live, so consider them as
     // live-outs for MBB. Inside MBB, dead assignments can still be detected.
-    for (MCPhysReg PhysReg : MRI->getReservedRegs().set_bits()) {
-      if (MRI->canSimplifyPhysReg(PhysReg)) {
-        LivePhysRegs.addReg(PhysReg);
-      }
```
martien-de-jong, line 171: What does addLiveOuts do? I can imagine it collects liveOuts from the successor blocks, but since it may not have visited successors yet, these may not be complete. For the unvisited blocks it should add the conservative set, and I hope that includes the simpliiable ones. That would be the way to get the control registers live at function return

gbossu: This uses the liveins from the successor blocks, those are there for each MBB (the virtregrewriter inserts them when dropping the "live range" view). But those do not contain reserved registers.

llvm/test/CodeGen/AIE/aie2/GlobalISel/dead-mi-elim.mir
SagarMaheshwari99: right, added it.

llvm/lib/CodeGen/DeadMachineInstructionElim.cpp
martien-de-jong: Oh, we get them live anyway from the loop below. But if we would want to be more precise (i.e. run a fix-point loop, only inserting live registers at calls and return), this will be suboptimal.

llvm/lib/CodeGen/DeadMachineInstructionElim.cpp
```diff
@@ -187,6 +200,15 @@ bool DeadMachineInstructionElimImpl::eliminateDeadMI(MachineFunction &MF) {
       }
 
       LivePhysRegs.stepBackward(MI);
+
+      // If the instruction is a call, ensure reserved registers are made live.
+      // This also ensures that reserved registers are preserved across call
+      // boundaries.
```
martien-de-jong: It confuses me to think of this as being 'preserved across call boundaries', It's only into the call that the value is preserved, which is 'normal' liveness. Nothing is going to preserve the value in the register to survive the call. The callee is free to redefine that register, which may then have a use in the rest of the block. However, this preservation means that the return of the callee uses the callee's definitions, it is not arranged for here.

gbossu: I would just rephrase as `If the instruction is a call, conservatively assume that it reads reserved registers.`

gbossu (approved):
LGTM, thanks for the fix!

martien-de-jong (approved):

PR #246, opened 2024-12-04
Combine (shl (and x, imm1), imm2) to (shl x, imm2)

abhinay-anubola:
- Generic combiner that combines
  ```
  %t = G_AND %x, imm1
  %root = G_SHL %t, imm2
  ```
  into
  ```
  %root = G_SHL %x, imm2
  ```
  Where (~imm1 << imm2) = 0

- Updated existing tests and added new tests accordingly.


andcarminati:
It will be nice to have MIR tests applying the exact pattern in terms of G_MIR instructions.

llvm/lib/CodeGen/GlobalISel/CombinerHelper.cpp
```diff
@@ -2553,6 +2553,34 @@ void CombinerHelper::applyCombineTruncOfExt(
   MI.eraseFromParent();
 }
 
+bool CombinerHelper::matchCombineShlOfAnd(MachineInstr &MI, Register &Reg) {
+  assert(MI.getOpcode() == TargetOpcode::G_SHL && "Expected a G_SHL");
+  Register SrcReg = MI.getOperand(1).getReg();
+  Register ShftReg = MI.getOperand(2).getReg();
```
krishnamtibrewala: nit: const Register ...

llvm/lib/CodeGen/GlobalISel/CombinerHelper.cpp
```diff
@@ -2553,6 +2553,34 @@ void CombinerHelper::applyCombineTruncOfExt(
...
+
+  const auto ShftVal = getIConstantVRegSExtVal(ShftReg, MRI);
+  if (!ShftVal)
+    return false;
+  MachineInstr *SrcMI = MRI.getVRegDef(SrcReg);
```
krishnamtibrewala: This is unsafe, getVRegDef can return nullptr (https://llvm.org/doxygen/MachineRegisterInfo_8cpp_source.html#l00407)
Add a check to make sure SrcMI is not a nullptr.

llvm/lib/CodeGen/GlobalISel/CombinerHelper.cpp
```diff
@@ -2553,6 +2553,34 @@ void CombinerHelper::applyCombineTruncOfExt(
...
+  if (SrcMI->getOpcode() != TargetOpcode::G_AND)
```
krishnamtibrewala: Can you add a comment to explain why we check for not a G_AND ?

llvm/lib/CodeGen/GlobalISel/CombinerHelper.cpp
```diff
@@ -2553,6 +2553,34 @@ void CombinerHelper::applyCombineTruncOfExt(
...
+    return false;
+
+  Register AndRHS = SrcMI->getOperand(2).getReg();
```
krishnamtibrewala: nit : const Register

llvm/lib/CodeGen/GlobalISel/CombinerHelper.cpp
```diff
@@ -2553,6 +2553,48 @@ void CombinerHelper::applyCombineTruncOfExt(
...
+  const Register ShftReg = MI.getOperand(2).getReg();
+  const LLT SrcTy = MRI.getType(SrcReg);
+  const unsigned Size = SrcTy.getSizeInBits();
+
+  const auto Shift = getIConstantVRegSExtVal(ShftReg, MRI);
+  if (!Shift)
+    return false;
+
+  const unsigned ShiftVal = Shift.value();
+  if (ShiftVal >= Size / 2) {
```
andcarminati: nit: sometimes is better to invert the `if` to have an early return here (reduce indentation).

andcarminati:
Hi @abhinay-anubola, the implementation looks good! Do you have some impact measurement (QoR)?

abhinay-anubola:
There are no significant changes, all changes are around than 0.1%

llvm/lib/CodeGen/GlobalISel/CombinerHelper.cpp
```diff
@@ -2553,6 +2553,47 @@ void CombinerHelper::applyCombineTruncOfExt(
   MI.eraseFromParent();
 }
 
+bool CombinerHelper::matchCombineShlOfAnd(MachineInstr &MI, Register &Reg) {
+  // We're trying to match the following pattern:
+  //   %t = G_AND %x, (2^imm - 1)
```
martien-de-jong: nit: I think this can be generalized to include G_SEXT, G_ZEXT, etc, anything that only touches the upper wordsize-imm bits.

llvm/lib/CodeGen/GlobalISel/CombinerHelper.cpp
```diff
@@ -2553,6 +2553,47 @@ void CombinerHelper::applyCombineTruncOfExt(
...
+  const Register SrcReg = MI.getOperand(1).getReg();
+  const Register ShftReg = MI.getOperand(2).getReg();
+  const LLT SrcTy = MRI.getType(SrcReg);
+  const unsigned Size = SrcTy.getSizeInBits();
+
+  const auto Shift = getIConstantVRegSExtVal(ShftReg, MRI);
+  if (!Shift)
+    return false;
+  const unsigned ShiftVal = Shift.value();
+  if (ShiftVal < Size / 2)
```
martien-de-jong: I think this can be more general. Shift discards 'imm' upper bits of its source value, so if the AND only touches those bits, it is redundant. That means that the AND can be removed if its immediate has only one bits in the lower wordsize - imm bits. effectively, `(~andimm) << shiftimm == 0`.

martien-de-jong: And this generalizes to shiftright (both versions) with very little change.

llvm/lib/CodeGen/GlobalISel/CombinerHelper.cpp
abhinay-anubola: `zext(trunc x) ->x ` combine already exist. `sext(trunc x) -> x` combine is in this [PR](https://github.com/Xilinx/llvm-aie/pull/233)

llvm/lib/CodeGen/GlobalISel/CombinerHelper.cpp
abhinay-anubola: Modified shiftleft accordingly, but there a combine for shiftright which does similar.
`matchBitfieldExtractFromShrAnd`

llvm/lib/CodeGen/GlobalISel/CombinerHelper.cpp
```diff
@@ -2553,6 +2553,42 @@ void CombinerHelper::applyCombineTruncOfExt(
...
+  // Where (~imm1 << imm2) = 0
+  assert(MI.getOpcode() == TargetOpcode::G_SHL && "Expected a G_SHL");
+  const Register DstReg = MI.getOperand(0).getReg();
+  const Register SrcReg = MI.getOperand(1).getReg();
+  const LLT SrcTy = MRI.getType(SrcReg);
+  const unsigned Size = SrcTy.getSizeInBits();
+
+  // Try to match shl (and x, imm1), imm2
+  int64_t ShiftImm, AndImm;
+  int64_t Mask = (Size < 64) ? (1LL << Size) - 1 : -1;
```
andcarminati: nit: move this mask to just before the use.

andcarminati: I also suggest to use unsigned masks whenever is possible. I this way, can we use something like `uint64_t Mask = ~0ULL >> (64-Size);` here?

llvm/lib/CodeGen/GlobalISel/CombinerHelper.cpp
```diff
@@ -2553,6 +2553,42 @@ void CombinerHelper::applyCombineTruncOfExt(
...
+                       m_ICst(ShiftImm))))
+    return false;
+  // Check if AndImm has bits set only in positions that will be shifted out by
+  // ShiftImm. If any significant bits remain after the shift, the AND operation
+  // cannot be removed.
+  uint64_t Mask = ~0ULL >> (64 - Size);
+  if (((~AndImm << ShiftImm) & Mask) != 0)
+    return false;
+
+  return true;
```
andcarminati: nit: can we remove this `if` and simply return the negation of its condition?

andcarminati:
I would like to suggest updating the commit message to include the conditions that must be met in order to apply the combiner. It could be similar to what you mentioned in your initial comment. This will be helpful for future reference. The PR is in very good shape!

llvm/test/CodeGen/AArch64/GlobalISel/combine-shl-and.mir
```diff
@@ -0,0 +1,152 @@
...
+    ; CHECK-NEXT: $w0 = COPY [[SHL]](s32)
+    %0:_(s32) = COPY $w0
+    %1:_(s32) = G_CONSTANT i32 8
+    %2:_(s32) = G_CONSTANT i32 255
+    %3:_(s32) = G_AND %0, %2
+    %4:_(s32) = G_SHL %3, %1
+    $w0 = COPY %4(s32)
+...
+---
+# Negative test case: Here mask 255 is not of shift value i.e. 16 bits.
```
andcarminati: nit: this comment sounds a bit strange.

llvm/lib/CodeGen/GlobalISel/CombinerHelper.cpp
```diff
@@ -2553,6 +2553,39 @@ void CombinerHelper::applyCombineTruncOfExt(
...
+  assert(MI.getOpcode() == TargetOpcode::G_SHL && "Expected a G_SHL");
+  const Register DstReg = MI.getOperand(0).getReg();
+  const Register SrcReg = MI.getOperand(1).getReg();
+  const LLT SrcTy = MRI.getType(SrcReg);
+  const unsigned Size = SrcTy.getSizeInBits();
+
+  // Try to match shl (and x, imm1), imm2
+  int64_t ShiftImm, AndImm;
+  if (!mi_match(DstReg, MRI,
+                m_GShl(m_OneNonDBGUse(m_GAnd(m_Reg(Reg), m_ICst(AndImm))),
```
martien-de-jong, line 2572: nit: I don't think the mOneNonDBGUse condition is tested, but I also don't think it's necessary. We just remove a use of the AND. If it's the last use, DCE will remove it; if it's not, it will serve the other uses.

martien-de-jong (approved):
Would be nice if this could be upstreamed to get the AArch64 and AMDGPU updates in mainstream. That would require to isolate the AIE changes in a separate commit.

PR #248, opened 2024-12-11
[Driver][LLD] Turn off use of default page alignment.

SagarMaheshwari99:


clang/test/Driver/aie2/aie2-toolchain.c
```diff
@@ -46,3 +46,7 @@
 // NMAGIC: "{{[^"]*}}ld.lld{{[^"]*}}"
 // NMAGIC-SAME: "--nmagic"
 
+// RUN: %clang %s -### --target=aie2-none-unknown-elf -Wl,--no-nmagic -ccc-install-dir %S/../Inputs/basic_aie_tree/bin 2>&1 \
+// RUN:   | FileCheck -check-prefix=NONMAGIC %s
+// NONMAGIC: "{{[^"]*}}ld.lld{{[^"]*}}"
+// NONMAGIC-SAME: "--nmagic{{.*}}--no-nmagic"
```
martien-de-jong, line 53: We want to test that users can shoot themselves in the foot?

martien-de-jong (dismissed):

clang/test/Driver/aie2/aie2-toolchain.c
```diff
@@ -39,3 +39,14 @@
 // RUN: %clang %s -### --target=aie2-none-unknown-elf -ccc-install-dir  %S/../Inputs/basic_aie_tree/bin 2>&1 \
 // RUN:   | FileCheck -check-prefix=C-INCLUDES %s
 // C-INCLUDES: "-internal-externc-isystem" "{{.*}}include{{/|\\\\}}aie2-none-unknown-elf"
+
+// Turn off use of default page alignment.
```
andcarminati: nit: Verify whether the default page alignment feature is disabled.

(we are not turning it off here)

clang/test/Driver/aie2/aie2-toolchain.c
SagarMaheshwari99: yes.

clang/test/Driver/aie2/aie2-toolchain.c
SagarMaheshwari99: Actually we are. It might seem counter-intuitive at first with `no` in the prefix of `--no-nmagic`, but when we use `--nmagic`, we are saying that we are deviating from the standard practice of page aligning the sections (standard for most general purpose systems anyways).  So `no` in the flag is to prohibit that deviation and does not actually signify page alignment itself. 

Edit: Ah, never mind, I think you were talking about the comment. I will update it.

andcarminati (approved):
LGTM.

PR #251, opened 2025-01-07
Extend ZOL to recognize static bounded loops

abhinay-anubola:
Extended existing `std::optional<int64_t> getMinTripCount(const Loop *L)` to get iteration count using Scalar Evolution analysis.

### DIFF
--- a/llvm/lib/Target/AIE/AIE2TargetTransformInfo.cpp
+++ b/llvm/lib/Target/AIE/AIE2TargetTransformInfo.cpp
@@ -119,19 +119,14 @@ bool AIE2TTIImpl::isHardwareLoopProfitable(Loop *L, ScalarEvolution &SE,
     return false;
 
   if (!ForceHLGeneration) {
-    if (const MDNode *LoopID = L->getLoopID()) {
-      std::optional<int64_t> MinTripCount = getMinTripCount(LoopID);
-      if (MinTripCount) {
-        // Reject HL for this case.
-        if (*MinTripCount <= MinIterCountHLReject) {
-          return false;
-        }
-      } else {
-        // We have metadata, but not iteration information.
+    std::optional<int64_t> MinTripCount = getMinTripCount(L, &SE);
+    if (MinTripCount) {
+      // Reject HL for this case.
+      if (*MinTripCount <= MinIterCountHLReject) {
         return false;
       }
     } else {
-      // We don't have metadata.
+      // We have metadata, but not iteration information.
       return false;
     }
   }
--- a/llvm/lib/Transforms/Utils/LoopUtils.cpp
+++ b/llvm/lib/Transforms/Utils/LoopUtils.cpp
@@ -447,7 +447,15 @@ std::optional<int64_t> llvm::getMinTripCount(const MDNode *LoopID) {
   return std::nullopt;
 }
 
-std::optional<int64_t> llvm::getMinTripCount(const Loop *L) {
+std::optional<int64_t> llvm::getMinTripCount(const Loop *L,
+                                             ScalarEvolution *SE) {
+  if (SE && L->isRotatedForm()) {
+    if (SE->hasLoopInvariantBackedgeTakenCount(L)) {
+      const SCEV *BackedgeTakenCount = SE->getBackedgeTakenCount(L);
+      if (const SCEVConstant *CT = dyn_cast<SCEVConstant>(BackedgeTakenCount))
+        return CT->getValue()->getSExtValue() + 1;
+    }
+  }
   return getMinTripCount(L->getLoopID());
 }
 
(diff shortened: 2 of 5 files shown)


llvm/lib/Target/AIE/Utils/AIELoopUtils.cpp
```diff
@@ -44,6 +44,15 @@ std::optional<int64_t> getMinTripCount(const MachineBasicBlock &LoopBlock) {
   return MinTripCount;
 }
 
+std::optional<int64_t> getMinTripCount(const Loop *L, ScalarEvolution &SE) {
```
gbossu: I think we could extend the original `getMinTripCount(const Loop *L)` in the base LLVM LoopUtils with an optional `ScalarEvolution *SE = nullptr` parameter instead

llvm/lib/Target/AIE/Utils/AIELoopUtils.cpp
```diff
@@ -44,6 +44,15 @@ std::optional<int64_t> getMinTripCount(const MachineBasicBlock &LoopBlock) {
...
+  if (SE.hasLoopInvariantBackedgeTakenCount(L)) {
+    const SCEV *BackedgeTakenCount = SE.getBackedgeTakenCount(L);
+    if (const SCEVConstant *CT = dyn_cast<SCEVConstant>(BackedgeTakenCount))
+      return CT->getValue()->getSExtValue() + 1;
```
gbossu: I think the "backedge-taken count" is one less than what we call the "iteration count" only for rotated loops. We might then want to check that the loop is in rotated form. (And maybe extra things like: single latch and exit block)

llvm/lib/Target/AIE/AIE2TargetTransformInfo.cpp
```diff
@@ -119,19 +119,14 @@ bool AIE2TTIImpl::isHardwareLoopProfitable(Loop *L, ScalarEvolution &SE,
...
-        // Reject HL for this case.
-        if (*MinTripCount <= MinIterCountHLReject) {
-          return false;
-        }
-      } else {
-        // We have metadata, but not iteration information.
+    std::optional<int64_t> MinTripCount = AIELoopUtils::getMinTripCount(L, SE);
+    if (MinTripCount) {
+      // Reject HL for this case.
+      if (*MinTripCount <= MinIterCountHLReject) {
```
martien-de-jong, line 125: I think this criterium deserves reconsideration. This is meant to allow the prepipeliner to create a SWP schedule with guarded prologues, but totally blocks the postpipeliner. This will make it impossible to force the postpipeliner by a pragma.
I would suggest to only return false here if there's no pragma forcing the postpipeliner.

llvm/lib/Target/AIE/Utils/AIELoopUtils.cpp
andcarminati: Maybe it is better. I see 3 different implementations now for `getMinTripCount`. In can be a bit confusing.

llvm/lib/Target/AIE/Utils/AIELoopUtils.cpp
abhinay-anubola: I think using `L->isRotatedForm()` would suffice, as internally it is checking for single latch and exit block.

llvm/lib/Target/AIE/AIE2TargetTransformInfo.cpp
abhinay-anubola: These checks were not added by me, however if you want me to remove the `*MinTripCount <= MinIterCountHLReject` condition I can remove it.

martien-de-jong: As discussed offline. The comment was meant to raise awareness in the team, not to address it in this PR. Also, changing the default value of MinIterCountHLReject would suffice.

llvm/lib/Transforms/Utils/LoopUtils.cpp
```diff
@@ -447,7 +447,15 @@ std::optional<int64_t> llvm::getMinTripCount(const MDNode *LoopID) {
...
-std::optional<int64_t> llvm::getMinTripCount(const Loop *L) {
+std::optional<int64_t> llvm::getMinTripCount(const Loop *L,
+                                             ScalarEvolution *SE) {
+  if (SE && L->isRotatedForm()) {
+    if (SE->hasLoopInvariantBackedgeTakenCount(L)) {
+      const SCEV *BackedgeTakenCount = SE->getBackedgeTakenCount(L);
+      if (const SCEVConstant *CT = dyn_cast<SCEVConstant>(BackedgeTakenCount))
+        return CT->getValue()->getSExtValue() + 1;
+    }
+  }
```
martien-de-jong, line 458: I realize now this isn't used by the softwarepipeliners. We may want to have a follow-up to find the maximum stage count for a ZOL without a pragma but with a constant LoopStart.

martien-de-jong (approved):

PR #254, opened 2025-01-09
[AIE2P] Add initial AIE2P/Strix support

konstantinschwarz:
This PR introduces initial support for AIE2P/Strix, including frontend intrinsics, instruction selection, compiler-rt builtins, and runtimes (libc/libcxx).
Support for BFP16 data types and LLD will be added in follow-up PRs.


andcarminati:
Oh, I frequently run into those same mismatches between clan-format versions.

konstantinschwarz:
> Oh, I frequently run into those same mismatches between clan-format versions.

The `Triple.{h.cpp}` files are unfortunately not clang-formatted in upstream LLVM.
Formatting them in a downstream project leads to awful conflicts, so in the past we just accepted them and force merged changes to these files. I'd suggest we do the same here

PR #256, opened 2025-01-10
[AIE] Add missing addressing intrinsics using dims_2d_t/dims_3d_t

khallouh:

PR #263, opened 2025-01-15
Prune all Python versions from hosted tool cache prior to building

philippjh:
In #256 we were seeing the following build failure:

```
CMake Error at /usr/share/cmake-3.22/Modules/FindPackageHandleStandardArgs.cmake:230 (message):
  Could NOT find Python3 (missing: Interpreter) (Required is at least version
  "3.6")

      Reason given by package: 
          Interpreter: Cannot run the interpreter "/__t/Python/3.11.11/x64/bin/python3"
```

It turned out, that the problem is coming from Python requiring a newer GLIBC version inside the container.

```
/__t/Python/3.11.11/x64/bin/python3: /lib/x86_64-linux-gnu/libm.so.6: version `GLIBC_2.38' not found (required by /__t/Python/3.11.11/x64/lib/libpython3.11.so.1.0)
```

However, spawning the container locally and downloading the used Python 3.11.11 from GitHub's own [Python release list](https://github.com/actions/python-versions/), I wasn't able to reproduce this problem. It turned out, that the hosted tool cache, for some reasons, holds a distribution which is linked against the newer GLIBC and the `setup-python` action would not recognize that and (re-)fetch the Python package. 

This PR effectively disables the caching by pruning all previously fetched Python versions from the hosted tool cache.

### DIFF
--- a/.github/workflows/llvm-project-tests.yml
+++ b/.github/workflows/llvm-project-tests.yml
@@ -80,6 +80,10 @@ jobs:
         uses: llvm/actions/setup-windows@main
         with:
           arch: amd64
+      - name: Remove Python in hosted tool cache
+        if: runner.os == 'Linux'
+        run: |
+          rm -Rf ${RUNNER_TOOL_CACHE}/Python
       # On Windows, starting with win19/20220814.1, cmake choose the 32-bit
       # python3.10.6 libraries instead of the 64-bit libraries when building
       # lldb.  Using this setup-python action to make 3.10 the default


.github/workflows/llvm-project-tests.yml
```diff
@@ -80,6 +80,10 @@ jobs:
         uses: llvm/actions/setup-windows@main
         with:
           arch: amd64
+      - name: Remove Python in hosted tool cache
+        if: runner.os == 'Linux'
+        run: |
+          rm -Rf ${RUNNER_TOOL_CACHE}/Python
```
julio-gorge, line 86: This is quite the nuclear 💣option, have we tried passing a specific Python 3.11 version to `setup-python@v4` to force it to upgrade existing Python 3.11 installation to a certain `3.11.<patch>` version, or using `check-latest: true`?

```yaml
check-latest:
    description: "Set this option if you want the action to check for the latest available version that satisfies the version spec."
    default: false
```

```yaml
uses: actions/setup-python@v4
        with:
          python-version: ${{ inputs.python_version }}
          check-latest: true
```

philippjh: I agree. However, explicitly setting the requested Python version to `3.11.11` did not work, I was still seeing the glibc error from the distribution present in the tool cache. I'm not sure about `check-latest`, though. I reckon, that this switch is only functional, if you loosely specify the requested Python version, e.g., `3.11` instead of `3.11.11`. In this case the routine resolves to the latest release, that still satisfies the version specifier (e.g., `3.11 -> 3.11.11`). The [code](https://github.com/actions/setup-python/blob/main/src/find-python.ts#L48) of the action also supports this claim, imo.

julio-gorge (approved):
Good to get devs unblocked. Would be nice to address this another way in the future.

.github/workflows/llvm-project-tests.yml
philippjh: As the issue was popping up in another PR, I had [the opportunity to confirm](https://github.com/Xilinx/llvm-aie/actions/runs/12793980492/job/35668058825?pr=269), that both setting `check-latest` to `true` and enforcing the requested Python version to be `3.11.11` is not sufficient to resolve this issue. I advocate to proceed with the changes proposed in the PR to unblock developers.

philippjh: [Upgrading](https://github.com/Xilinx/llvm-aie/actions/runs/12805363704/job/35701615544?pr=269) `setup-python` to latest `v5` also does not fix the issue.

andcarminati (approved):
LGTM.

philippjh:
We're proceeding with #275, which contains the upstream fix for the issue this PR is solving.

PR #264, opened 2025-01-15
[AIE2P] Fix Accumulator Register Bank assignment for G_STORE

khallouh:
For `G_STORE`, we look at the bank of the register in the defining instruction but we assume that this will be at index zero, which has been the case for AIE2P's intrinsics defining accumulators but not true in general.

### DIFF
--- a/llvm/lib/Target/AIE/aie2p/AIE2PRegisterBankInfo.cpp
+++ b/llvm/lib/Target/AIE/aie2p/AIE2PRegisterBankInfo.cpp
@@ -764,13 +764,12 @@ AIE2PRegisterBankInfo::getInstrMapping(const MachineInstr &MI) const {
     Register VReg = MI.getOperand(0).getReg();
     if (!VReg)
       break;
-    MachineInstr *DefMI = MRI.getVRegDef(VReg);
     LLT Type = MRI.getType(VReg);
-    auto *RB = getRegBank(DefMI->getOperand(0).getReg(), MRI, TRI);
+    auto DefRegBank = getRegBank(VReg, MRI, TRI);
     // Add an additional typesize check to avoid matching the following case.
     // %3:_(<16 x s64>), %4:_(s32) = llvm.aie2p.scd.expand.ACC1024.incr),
     // %1(s32) G_STORE %4(s32), %2(p0) :: (store (s32))
-    if (RB == &AIE2P::AccRegBank && Type.getSizeInBits() > 256) {
+    if (DefRegBank == &AIE2P::AccRegBank && Type.getSizeInBits() > 256) {
       OpRegBankIdx[0] = getAccPartialMappingIdx(MRI.getType(VReg));
       OpRegBankIdx[1] = PMI_PTR;
       return AIEBaseRegisterBankInfo::getInstrMappingFinal(MI, Cost, OpSize,


llvm/lib/Target/AIE/aie2p/AIE2PRegisterBankInfo.cpp
```diff
@@ -587,6 +587,17 @@ bool AIE2PRegisterBankInfo::isUseAccInsn(const MachineRegisterInfo &MRI,
...
+static int getVRegIndexInInstr(const MachineInstr &MI, unsigned VReg) {
+  for (unsigned i = 0, e = MI.getNumOperands(); i != e; ++i) {
+    const MachineOperand &MO = MI.getOperand(i);
+    if (MO.isReg() && MO.getReg() == VReg) {
+      return i;
+    }
+  }
+  return -1;
+}
+
```
abnikant: This looks good, can you add a test ?

llvm/lib/Target/AIE/aie2p/AIE2PRegisterBankInfo.cpp
```diff
@@ -766,11 +777,13 @@ AIE2PRegisterBankInfo::getInstrMapping(const MachineInstr &MI) const {
       break;
     MachineInstr *DefMI = MRI.getVRegDef(VReg);
     LLT Type = MRI.getType(VReg);
-    auto *RB = getRegBank(DefMI->getOperand(0).getReg(), MRI, TRI);
+    auto DefRegBank = getRegBank(
+        DefMI->getOperand(getVRegIndexInInstr(*DefMI, VReg)).getReg(), MRI,
```
niwinanto: Is `DefMI->getOperand(getVRegIndexInInstr(*DefMI, VReg)).getReg()` same as `VReg`?

llvm/lib/Target/AIE/aie2p/AIE2PRegisterBankInfo.cpp
```diff
@@ -587,6 +587,17 @@ bool AIE2PRegisterBankInfo::isUseAccInsn(const MachineRegisterInfo &MRI,
...
 
+// Return the operand index of VReg in MI
+static int getVRegIndexInInstr(const MachineInstr &MI, unsigned VReg) {
+  for (unsigned i = 0, e = MI.getNumOperands(); i != e; ++i) {
+    const MachineOperand &MO = MI.getOperand(i);
+    if (MO.isReg() && MO.getReg() == VReg) {
+      return i;
+    }
+  }
+  return -1;
```
andcarminati: What about a `std::optional`? This -1 looks dangerous regarding alternative uses.

llvm/lib/Target/AIE/aie2p/AIE2PRegisterBankInfo.cpp
khallouh: `DefMI->getOperand(getVRegIndexInInstr(*DefMI, VReg)).getReg()` is an operand and `VReg` is a register. So they cannot be the same.

llvm/lib/Target/AIE/aie2p/AIE2PRegisterBankInfo.cpp
khallouh: I can't add a test triggering this bug because there is no intrinsic that ouputs an accumulator at index zero. but we can't assume that `VReg` is always defined at index 0 of `DefMI` as we don't know what DefMI will be. 
( See replaced line:  `auto *RB = getRegBank(DefMI->getOperand(0).getReg(), MRI, TRI)` )

llvm/lib/Target/AIE/aie2p/AIE2PRegisterBankInfo.cpp
niwinanto: No, `DefMI->getOperand(getVRegIndexInInstr(*DefMI, VReg)).getReg()` is a register. There is a tail `getReg()`

khallouh: Ah you're right, there is `.getReg()`. They should be the same.

khallouh: Updated.

andcarminati (approved):
LGTM.

llvm/lib/Target/AIE/aie2p/AIE2PRegisterBankInfo.cpp
```diff
@@ -764,13 +764,12 @@ AIE2PRegisterBankInfo::getInstrMapping(const MachineInstr &MI) const {
     Register VReg = MI.getOperand(0).getReg();
     if (!VReg)
       break;
-    MachineInstr *DefMI = MRI.getVRegDef(VReg);
     LLT Type = MRI.getType(VReg);
-    auto *RB = getRegBank(DefMI->getOperand(0).getReg(), MRI, TRI);
+    auto DefRegBank = getRegBank(VReg, MRI, TRI);
     // Add an additional typesize check to avoid matching the following case.
     // %3:_(<16 x s64>), %4:_(s32) = llvm.aie2p.scd.expand.ACC1024.incr),
     // %1(s32) G_STORE %4(s32), %2(p0) :: (store (s32))
```
niwinanto, line 771: Since you are at it , could you fix this typo.
`// G_STORE %4(s32), %2(p0) :: (store (s32))`

llvm/lib/Target/AIE/aie2p/AIE2PRegisterBankInfo.cpp
```diff
@@ -764,13 +764,12 @@ AIE2PRegisterBankInfo::getInstrMapping(const MachineInstr &MI) const {
...
-    if (RB == &AIE2P::AccRegBank && Type.getSizeInBits() > 256) {
+    if (DefRegBank == &AIE2P::AccRegBank && Type.getSizeInBits() > 256) {
```
niwinanto, line 772: Now I am bit more confused. if vreg is already accregbank, why we need the special handling. `return AIEBaseRegisterBankInfo::getInstrMapping(MI);` itself enough right?

Before this change, at least we are looking at a different register than vreg for come cases, but I dont know the usecase.

niwinanto: @khallouh You missed this comment.

khallouh: I think we were looking at VReg before the change but we were assuming it would always be at index 0. Do you have an example where they are different or any case that I missed?

niwinanto: The comment above the changed line
```
    // %3:_(<16 x s64>), %4:_(s32) = llvm.aie2p.scd.expand.ACC1024.incr),
    //G_STORE %4(s32), %2(p0) :: (store (s32))
```
Before change we check for ` %3:_(<16 x s64>)` is `DefRegBank ` and now `%4(s32)`.

Also, as I mentioned above, if  `%4(s32)` is the indented register, `AIEBaseRegisterBankInfo::getInstrMapping(MI);` might be enough.

khallouh: `%4(s32)` is not the intended register and `AIEBaseRegisterBankInfo::getInstrMapping(MI);` is already called for that case.
That comment is now outdated. Most probably that type check (`Type.getSizeInBits() > 256`) was just a hack around the above case. @abnikant can you confirm?
```
    // Add an additional typesize check to avoid matching the following case.
    // %3:_(<16 x s64>), %4:_(s32) = llvm.aie2p.scd.expand.ACC1024.incr),
    // %1(s32) G_STORE %4(s32), %2(p0) :: (store (s32))
    if (RB == &AIE2P::AccRegBank && Type.getSizeInBits() > 256) {
      OpRegBankIdx[0] = getAccPartialMappingIdx(MRI.getType(VReg));
      OpRegBankIdx[1] = PMI_PTR;
      return AIEBaseRegisterBankInfo::getInstrMappingFinal(MI, Cost, OpSize,
                                                           OpRegBankIdx);
    }
```
We don't need the type check anymore now that the bug is fixed. That example precisely shows what this fixes where if it weren't for that check (`Type.getSizeInBits() > 256`) we would be assigning the accumulator bank to a store of a scalar. Now we are looking at the correct `%4(s32)` and it won't have the accumulator bank assigned which means we will return  `AIEBaseRegisterBankInfo::getInstrMapping(MI);` exactly as we expect. 

I opened https://github.com/Xilinx/llvm-aie/pull/285 to remove the redundant type check and the outdated comment.

abnikant: yes, true.

niwinanto: @abnikant Why we dont use `isUseAccInsn` for G_STORE case?

PR #266, opened 2025-01-15
[AIE2P] Update `AIE2PGenInstrInfo.td`

mludevid:
This PR updates the instruction definition of all instructions that use the loading fifo register. I also updated the respective decoder methods, the itinerary definitions, the binary output tests and any other test using these instructions. I had to disable some run lines in the binary output tests, because the assembly parser isn't working yet. We should be able to reenable them once @andcarminati merges the new subregister definitions for all of these instructions. For now I commented the old subregister definitions out, since we cannot define the new ones until the data structure changes to subregisters get merged.
I also used the opportunity to remove some old redundant test files.


llvm/test/CodeGen/AIE/aie2p/BinaryOutput/vldb.mir
```diff
@@ -7,9 +7,10 @@
 #
 # RUN: llc %llcflags --filetype=obj -o %t
 # RUN: llvm-objdump --triple=aie2p -dr --no-print-imm-hex %t | FileCheck %s
-# RUN: llc %llcflags --filetype=asm -o %t2
-# RUN: llvm-mc -triple aie2p -filetype=obj -o %t %t2
-# RUN: llvm-objdump --triple=aie2p -dr --no-print-imm-hex %t | FileCheck %s
+# TODO: llc %llcflags --filetype=asm -o %t2
+# TODO: llvm-mc -triple aie2p -filetype=obj -o %t %t2
+# TODO: llvm-objdump --triple=aie2p -dr --no-print-imm-hex %t | FileCheck %s
+# Change TODOs to RUN once tied registers are implemented for POP and FILL instructions
```
khallouh: As discussed offline, these encoding tests don't pass because we are missing some `AsmParser` support not because of the tied register implementation from @andcarminati. So I would keep them split as before so that we still have regression tests for the instruction that are actually supported.

mludevid: Done. I split up the tests again, kept the RUN lines in the `vlda` and `vldb` test files and removed the comment.

katerynamuts (approved):
LGTM

khallouh (approved):

PR #283, opened 2025-01-17
[AIE2P] Enable legalization of 4x128-bit G_CONCAT_VECTORS

konstantinschwarz:


llvm/lib/Target/AIE/AIELegalizerHelper.cpp
```diff
@@ -1174,7 +1173,6 @@ bool AIELegalizerHelper::legalizeG_CONCAT_VECTORS(LegalizerHelper &Helper,
   const LLT DstTy = MRI.getType(DstReg);
   const LLT SrcTy = MRI.getType(SrcReg);
   assert(DstTy.isVector() && SrcTy.isVector() && "Expected vector types");
-  assert(SrcTy.getSizeInBits() >= 256 && "Input vector size does not match!");
```
khallouh, line 1177: Do we want to allow `G_CONCAT_VECTORS` of 4x64-bit? I don't think we select the resulting concats of 2x64-bit.

llvm/lib/Target/AIE/AIELegalizerHelper.cpp
```diff
@@ -1162,8 +1162,7 @@ bool AIELegalizerHelper::legalizeG_SELECT(LegalizerHelper &Helper,
 }
 
 // We legalize concat vector of 2 inputs. So, anything above we need to split
-// it. So far expect only 4 input. 1024bit vector from 4 256bit register and
-// 2048 accumulator register from 4 512bit registers.
+// it. So far expect only 4 input.
```
andcarminati, line 1165: nit: the comment seems to be contradictory. Can we make it a bit more clear?

llvm/lib/Target/AIE/AIELegalizerHelper.cpp
konstantinschwarz: You mean `G_CONCAT_VECTORS %0(<4 x s64>), %1(<4 x s64>), %2(<4 x s64>), %3(<4 x s64>)`?

khallouh: Yes, is this supported?

khallouh: Oh no, I mean since the PR name mentions 4x128 but also allows 4x64-bit by removing this assert. So is something like this supported?
`%4:_(<8 x s32>) = G_CONCAT_VECTORS %0(<2 x s32>), %1(<2 x s32>), %2(<2 x s32>), %3(<2 x s32>)`

llvm/lib/Target/AIE/AIELegalizerHelper.cpp
martien-de-jong: Looking at the implementation, it is a helper to compose a flat > 2 elements concat out of pairwise concats.
I think it could be easily extended to arbitrary powers of two and applied recursively. 
The case of two elements is the only one that is legalized without transformation and will end the recursion.

khallouh: I have this already in one of my local branches using an iterative approach. It is needed to legalize 16-byte aligned 512-bit vector loads into 4x128-bit loads using the `fewerElements` action (See WIP PR https://github.com/Xilinx/llvm-aie/pull/321). This handles an arbitrary power-of-2 number of inputs:
```
bool AIELegalizerHelper::legalizeG_CONCAT_VECTORS(LegalizerHelper &Helper,
                                                  MachineInstr &MI) const {
  MachineIRBuilder &MIRBuilder = Helper.MIRBuilder;
  MachineRegisterInfo &MRI = *MIRBuilder.getMRI();

  const Register DstReg = MI.getOperand(0).getReg();
  const Register SrcReg = MI.getOperand(1).getReg();
  const LLT SrcTy = MRI.getType(SrcReg);
  assert(DstTy.isVector() && SrcTy.isVector() && "Expected vector types");
  assert(SrcTy.getSizeInBits() >= 128 && "Input vector size does not  match!");

  const LLT DstVecEltTy = DstTy.getElementType();
  const unsigned ElTySize = DstVecEltTy.getSizeInBits();
  SmallVector<Register, 4> Regs;

  for (unsigned I = 1; I < MI.getNumOperands(); ++I) {
    Regs.push_back(MI.getOperand(I).getReg());
  }

  while (Regs.size() > 1) {
    SmallVector<Register, 4> NewRegs;
    for (unsigned I = 0; I < Regs.size(); I += 2) {
      if (I + 1 < Regs.size()) {
        const LLT SplitTy = LLT::fixed_vector(
            MRI.getType(Regs[I]).getNumElements() * 2, ElTySize);
        const Register NewReg = (Regs.size() == 2)
                                    ? DstReg
                                    : MRI.createGenericVirtualRegister(SplitTy);
        MIRBuilder.buildConcatVectors({NewReg}, {Regs[I], Regs[I + 1]});
        NewRegs.push_back(NewReg);
      } else {
        NewRegs.push_back(Regs[I]);
      }
    }
    Regs = std::move(NewRegs);
  }

  MI.eraseFromParent();
  return true;
}
```

konstantinschwarz:
Closing in favor of https://github.com/Xilinx/llvm-aie/pull/334

PR #287, opened 2025-01-20
[AIE2P] Support CopyPhysReg between L and EXP registers as well as FIFO and Vectors/Accumulators

khallouh:


niwinanto (dismissed):
LGTM!

llvm/lib/Target/AIE/aie2p/AIE2PInstrInfo.cpp
```diff
@@ -615,8 +615,8 @@ void AIE2PInstrInfo::copyPhysReg(MachineBasicBlock &MBB,
             TRI.getSubReg(DstReg, AIE2P::sub_odd_vecmaskexp_704))
         .addReg(TRI.getSubReg(SrcReg, AIE2P::sub_odd_vecmaskexp_704),
                 getKillRegState(KillSrc));
-  } else if ((AIE2P::eLdFifoRegRegClass.contains(SrcReg)) &&
-             (AIE2P::eLdFifoRegRegClass.contains(DstReg))) {
+  } else if ((AIE2P::FIFO1024RegClass.contains(SrcReg)) &&
+             (AIE2P::FIFO1024RegClass.contains(DstReg))) {
```
gbossu, line 619: THis seems unrelated, does that fix something?

khallouh: It supports sf registers as well   `def FIFO1024 : AIE2PVector1024RegisterClass<(add eLdFifoReg, sf)>;`

khallouh: Added a separate commit for this and mentioned it in the PR name

llvm/test/CodeGen/AIE/aie2p/run-physreg-copy.mir
```diff
@@ -274,3 +274,32 @@ body: |
...
+body: |
+  bb.0:
+    ; CHECK-LABEL: name: test_copy_sf_to_lf
+    ; CHECK: $lfl0 = VMOV_alu_mv_mv_x $sfl
+    ; CHECK-NEXT: $lfh0 = VMOV_alu_mv_mv_x $sfh
+  $lf0 = COPY $sf
+...
+
+---
+name:            test_copy_sf_to_vec
```
gbossu: Do we have similar tests for
 - vec to lf/sf
 - acc to lf/sf

?

khallouh: We did not have them. I added them

llvm/test/CodeGen/AIE/aie2p/run-physreg-copy.mir
```diff
@@ -274,3 +274,52 @@ body: |
...
+  $lf0 = COPY $cml0
+...
+
+---
+name:            test_copy_fifo_to_acc
+body: |
+  bb.0:
+    ; CHECK-LABEL: name: test_copy_fifo_to_acc
+    ; CHECK: $bmll0 = VMOV_alu_mv_mv_x $lfl0
+    ; CHECK-NEXT: $bmlh0 = VMOV_alu_mv_mv_x $lfh0
```
gbossu, line 323: Why do we have so many `if` cases  in the implementation if we always end up with the same opcode?

gbossu: ~Looking at instrinfo, that opcode only seems to handle X and BM registers~
Wrong: this seems fine.

khallouh: The opcode handles fifo registers as well. See below:

`def VMOV_alu_mv_mv_x : AIE2P_inst_mv_instr32 <(outs OP_mMvBMXDst:$dst), (ins OP_mMvBMXSrc:$src)`
```
def mMvBMXDst : AIE2PVector512RegisterClass<(add mBMs, mXm, mFifoHLReg)>;
def mMvBMXSrc : AIE2PVector512RegisterClass<(add mMvBMXDst)>;
```

khallouh: We have a different `if` case for `ACC1024` and `VEC1024` because the sub-register offsets are different. I don't see how we could merge those if statements.

gbossu: Clarified: There is a bug upstream that prevents us from using the same subreg indices for all of the 512-bit registers. Let's keep all the if cases with the distinct subregs so far.

gbossu (approved):

PR #288, opened 2025-01-20
 [AIE2P] Add support for a new FIFO Register Bank

khallouh:
- This introduces a new register bank for Fifo registers needed to properly/easily select the target instructions using those registers.
- A similar use-def analysis to the one already in place for the Accumulator Register Bank assignment is used to determine assignments to the Fifo Register Bank. 
- This handling is meant to be revisited in a follow-up to factor out the similar logic/code between the accumulator bank and the fifo bank assignment.


llvm/include/llvm/IR/IntrinsicsAIE2P.td
```diff
@@ -535,6 +535,12 @@ def int_aie2p_vbcst_shuffle16 : ClangBuiltin<"__builtin_aie2p_vbcst_shuffle16">,
 def int_aie2p_vbcst_shuffle32 : ClangBuiltin<"__builtin_aie2p_vbcst_shuffle32">, AIE2PVBcstShfl_I32;
 def int_aie2p_vbcst_shuffle64 : ClangBuiltin<"__builtin_aie2p_vbcst_shuffle64">, AIE2PVBcstShfl_I64;
 
+// Fifo loads
+class AIE2PFIFO_LD_FILL : Intrinsic<[llvm_ptr_ty, llvm_v32i32_ty, llvm_i32_ty],
+     [llvm_ptr_ty, llvm_v32i32_ty, llvm_i32_ty],
+     [IntrReadMem]>;
+def int_aie2p_fifo_ld_fill: AIE2PFIFO_LD_FILL;
+
```
abnikant, line 543: Do you want to keep this in a separate commit, and add a test ?

llvm/lib/Target/AIE/aie2p/AIE2PRegisterInfo.cpp
```diff
@@ -315,6 +315,16 @@ AIE2PRegisterInfo::getMinClassForRegBank(const RegisterBank &RB, LLT Ty) const {
       llvm_unreachable("Unsupported register size.");
     }
   }
+  case AIE2P::FifoRegBankID: {
+    switch (Ty.getSizeInBits()) {
+    case 512:
+      return AIE2P::mFifoHLRegRegClass;
```
abnikant: should we just define FIFO512/FIFO1024 ?

llvm/include/llvm/IR/IntrinsicsAIE2P.td
khallouh: There is a test already (See below). This is added to illustrate/test `isUsedAsFifoRegInIntrinsic` and `hasFifoInput` which are part of the register bank selection support. Or do you mean another kind of test?
```
---
name:            test_fifo_ld_fill
legalized:       true
body:             |
  bb.1.entry:
    ; GREEDY-LABEL: name: test_fifo_ld_fill
    ; GREEDY: [[INT:%[0-9]+]]:ptrregbank(p0), [[INT1:%[0-9]+]]:fiforegbank(<32 x s32>), [[INT2:%[0-9]+]]:gprregbank(s32) = G_INTRINSIC_W_SIDE_EFFECTS intrinsic(@llvm.aie2p.fifo.ld.fill), %3:ptrregbank(p0), %4:fiforegbank(<32 x s32>), %5:gprregbank(s32)
    ; GREEDY-NEXT: G_STORE [[INT1]](<32 x s32>), %6:ptrregbank(p0) :: (store (<32 x s32>))
    ; GREEDY-NEXT: G_STORE [[INT2]](s32), %7:ptrregbank(p0) :: (store (s32))
    ; GREEDY-NEXT: G_STORE [[INT]](p0), %8:ptrregbank(p0) :: (store (p0), align 4)
    ; GREEDY-NEXT: PseudoRET implicit $lr, implicit $cml0
    ;
    ; FAST-LABEL: name: test_fifo_ld_fill
    ; FAST: [[INT:%[0-9]+]]:ptrregbank(p0), [[INT1:%[0-9]+]]:fiforegbank(<32 x s32>), [[INT2:%[0-9]+]]:gprregbank(s32) = G_INTRINSIC_W_SIDE_EFFECTS intrinsic(@llvm.aie2p.fifo.ld.fill), %3:ptrregbank(p0), %4:fiforegbank(<32 x s32>), %5:gprregbank(s32)
    ; FAST-NEXT: G_STORE [[INT1]](<32 x s32>), %6:ptrregbank(p0) :: (store (<32 x s32>))
    ; FAST-NEXT: G_STORE [[INT2]](s32), %7:ptrregbank(p0) :: (store (s32))
    ; FAST-NEXT: G_STORE [[INT]](p0), %8:ptrregbank(p0) :: (store (p0), align 4)
    ; FAST-NEXT: PseudoRET implicit $lr, implicit $cml0
    %7:_(p0), %8:_(<32 x s32>), %9:_(s32) = G_INTRINSIC_W_SIDE_EFFECTS intrinsic(@llvm.aie2p.fifo.ld.fill), %4:_(p0), %5:_(<32 x s32>), %6:_(s32)
    G_STORE %8:_(<32 x s32>), %1:_(p0) :: (store (<32 x s32>))
    G_STORE %9:_(s32), %3:_(p0) :: (store (s32))
    G_STORE %7:_(p0), %0:_(p0) :: (store (p0))
    PseudoRET implicit $lr, implicit $cml0
...
```

abnikant:
PR looks good to me. See, if you can address some of the comments and also create a jira to refactor/reuse the existing logic used for accumulator banks for assigning fifo banks.

llvm/include/llvm/IR/IntrinsicsAIE2P.td
abnikant: This I saw. I mean since we are adding llvm intrinsic, adding a lowering test will make it complete or is it that you just added it here to verify the regbank assignment logic?

llvm/lib/Target/AIE/aie2p/AIE2PRegisterBankInfo.cpp
```diff
@@ -112,6 +116,9 @@ RegisterBankInfo::ValueMapping AIE2PGenRegisterBankInfo::ValMappings[]{
     {&PartMappings[PMI_VREG128 - PMI_Min], 1},
     // 29:
     {&PartMappings[PMI_ACC2048 - PMI_Min], 1},
+    // 30:
+    {&PartMappings[PMI_FIFO512 - PMI_Min], 1},
+    {&PartMappings[PMI_FIFO1024 - PMI_Min], 1},
```
niwinanto, line 121: NIt: 31

llvm/lib/Target/AIE/aie2p/AIE2PRegisterBankInfo.cpp
```diff
@@ -645,6 +731,11 @@ AIE2PRegisterBankInfo::getInstrMapping(const MachineInstr &MI) const {
         OpRegBankIdx[Idx] = getAccPartialMappingIdx(Type);
         continue;
       }
+      if (isUsedAsFifoRegInIntrinsic(MRI, MI, MI.getOperand(Idx).getReg())) {
+        LLT Type = MRI.getType(MI.getOperand(Idx).getReg());
+        OpRegBankIdx[Idx] = getFifoPartialMappingIdx(Type);
+        continue;
+      }
```
niwinanto, line 738: What about handling of `G_IMPLICIT_DEF`, `G_UNMERGE_VALUES` etc inside this function?

llvm/lib/Target/AIE/aie2p/AIE2PRegisterBankInfo.cpp
```diff
@@ -759,27 +853,40 @@ AIE2PRegisterBankInfo::getInstrMapping(const MachineInstr &MI) const {
...
+    if (isFifoPhysRegMapping) {
+      OpRegBankIdx[0] = getFifoPartialMappingIdx(Type);
+      OpRegBankIdx[1] = PMI_PTR;
+      return AIEBaseRegisterBankInfo::getInstrMappingFinal(MI, Cost, OpSize,
+                                                           OpRegBankIdx);
+    }
     return AIEBaseRegisterBankInfo::getInstrMapping(MI);
   }
   case TargetOpcode::G_STORE: {
     // Check if the store is fed by acc instructions.
```
niwinanto: Nit: Update the comment.

llvm/lib/Target/AIE/aie2p/AIE2PRegisterBankInfo.cpp
```diff
@@ -734,15 +825,18 @@ AIE2PRegisterBankInfo::getInstrMapping(const MachineInstr &MI) const {
   }
   case TargetOpcode::G_LOAD: {
     bool isAccRegMapping = false;
+    bool isFifoPhysRegMapping = false;
     // Check if that load feeds acc instructions.
```
niwinanto: Nit: comment

llvm/lib/Target/AIE/aie2p/AIE2PRegisterBankInfo.cpp
```diff
@@ -734,15 +825,18 @@ AIE2PRegisterBankInfo::getInstrMapping(const MachineInstr &MI) const {
...
     // Check if we already know the register bank.
     auto *RB = getRegBank(UseCandidate, MRI, TRI);
-    if (RB == &AIE2P::AccRegBank)
+    if (RB == &AIE2P::AccRegBank || RB == &AIE2P::FifoRegBank)
       return AIEBaseRegisterBankInfo::getInstrMapping(MI);
     if (isUseAccInsn(MRI, TRI, UseCandidate))
       isAccRegMapping = true;
+    if (isUseFifoInsn(MRI, TRI, UseCandidate))
+      isFifoPhysRegMapping = true;
     // size of accu vector on aie2p >= 512.
```
niwinanto: Nit:Comment

llvm/lib/Target/AIE/aie2p/AIE2PRegisterInfo.cpp
```diff
@@ -315,6 +315,16 @@ AIE2PRegisterInfo::getMinClassForRegBank(const RegisterBank &RB, LLT Ty) const {
       llvm_unreachable("Unsupported register size.");
     }
   }
+  case AIE2P::FifoRegBankID: {
```
niwinanto, line 318: Not sure this change is related. Seems used for instruction selection.

niwinanto (commented):
Mostly looks good, few nits.

llvm/include/llvm/IR/IntrinsicsAIE2P.td
khallouh: Yes, this is just a unit test to verify the regbank assignment logic, I don't need to add the C intrinsic/builtin and IR lowering. The fifo intrinsics will come later all together with their end-to-end support.

llvm/lib/Target/AIE/aie2p/AIE2PRegisterBankInfo.cpp
khallouh: We don't add the comment for value mappings of the same bank. See Above.

llvm/lib/Target/AIE/aie2p/AIE2PRegisterInfo.cpp
khallouh: I see that it is also used in RegisterBankSelect methods. In any case, I think it makes sense to implement this target hook as we add the register bank.

llvm/lib/Target/AIE/aie2p/AIE2PRegisterBankInfo.cpp
khallouh: I created this ticket AIECC-916 to see if this is needed along with the refactoring work if the logic is the same. Otherwise it would require its own ticket.

abnikant (approved):
LGTM!

llvm/include/llvm/IR/IntrinsicsAIE2P.td
```diff
@@ -535,6 +535,12 @@ def int_aie2p_vbcst_shuffle16 : ClangBuiltin<"__builtin_aie2p_vbcst_shuffle16">,
 def int_aie2p_vbcst_shuffle32 : ClangBuiltin<"__builtin_aie2p_vbcst_shuffle32">, AIE2PVBcstShfl_I32;
 def int_aie2p_vbcst_shuffle64 : ClangBuiltin<"__builtin_aie2p_vbcst_shuffle64">, AIE2PVBcstShfl_I64;
 
+// Fifo loads
+class AIE2PFIFO_LD_FILL : Intrinsic<[llvm_ptr_ty, llvm_v32i32_ty, llvm_i32_ty],
+     [llvm_ptr_ty, llvm_v32i32_ty, llvm_i32_ty],
+     [IntrReadMem]>;
+def int_aie2p_fifo_ld_fill: AIE2PFIFO_LD_FILL;
```
gbossu, line 542: Is this new intrinsic just for testing, and the other FIFO intrinsics will come in future PRs?

llvm/lib/Target/AIE/aie2p/AIE2PRegisterBankInfo.cpp
```diff
@@ -365,6 +387,24 @@ AIE2PRegisterBankInfo::getInstrAlternativeMappings(
   return AIEBaseRegisterBankInfo::getInstrAlternativeMappings(MI);
 }
 
+// Check if FifoRegCandidate is one of the fifo operands of the intrinsic
+static bool isUsedAsFifoRegInIntrinsic(const MachineRegisterInfo &MRI,
```
gbossu, line 391: Nit: Why not just `isUsedAsFifoReg`? I don't think it matter whether the user is an intrinsic or a plain instruction. What matters is that this should be mapped to the FIFO bank.

llvm/lib/Target/AIE/aie2p/AIE2PRegisterBankInfo.cpp
```diff
@@ -595,6 +665,22 @@ bool AIE2PRegisterBankInfo::isUseAccInsn(const MachineRegisterInfo &MRI,
                 });
 }
 
+// Check if RegOp is used as a fifo register.
+bool AIE2PRegisterBankInfo::isUseFifoInsn(const MachineRegisterInfo &MRI,
```
gbossu, line 669: The comment is clear, but the name not so much :) What would you think of `isUsedAsFifoReg()`?

llvm/lib/Target/AIE/aie2p/AIE2PRegisterBankInfo.cpp
```diff
@@ -595,6 +665,22 @@ bool AIE2PRegisterBankInfo::isUseAccInsn(const MachineRegisterInfo &MRI,
...
+                                          const TargetRegisterInfo &TRI,
+                                          const Register RegOp) const {
```
gbossu, line 671: Please just call this `Reg`, this is not an operand or opcode

llvm/lib/Target/AIE/aie2p/AIE2PRegisterBankInfo.cpp
gbossu: Hmm okay, this is called by `hasFifoInput` when the user is an intrinsic. ANd I think it does make sense to keep that in a separate function as you did, because who knows, future AIE versions might also have FIFO regs, but different intrinsics.

TLDR: No action needed

llvm/lib/Target/AIE/aie2p/AIE2PRegisterBankInfo.cpp
gbossu: Also, it's prefered not to enfore `const` on by-value parameters, see https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#con1-by-default-make-objects-immutable

llvm/lib/Target/AIE/aie2p/AIE2PRegisterBankInfo.cpp
gbossu: I think it would be nice to harmonise all those names, we have `isUseFifoInsn`, `hasFifoInput` and `hasFifoInput`. I'd suggest:
```
/// Check if \p Reg is used as a FIFO register by any instruction
bool isUsedAsFifoReg(const MachineRegisterInfo &MRI, Register Reg);

/// Check if \p Reg is used as a FIFO register by \p MI
bool isUsedAsFifoRegInInstr(const MachineInstr &MI, const MachineRegisterInfo &MRI, Register Reg);

/// Check if \p Reg is used as a FIFO register by the \p MI intrinsic
bool isUsedAsFifoRegInIntrinsic(const GIntrinsic &MI, const MachineRegisterInfo &MRI, Register Reg);
```

llvm/lib/Target/AIE/aie2p/AIE2PRegisterBankInfo.cpp
khallouh: This PR is merged now. We have the same for accumulators, there is a follow up to refactor that code and addresse any improvements. This can also be addressed there.

llvm/lib/Target/AIE/aie2p/AIE2PRegisterBankInfo.cpp
khallouh: This makes sense but this is already merged. We can address such improvements in the followup where we refactor code between accumulator and fifo handling. Most of these comments apply to that code as well.

llvm/include/llvm/IR/IntrinsicsAIE2P.td
khallouh: Right. Just for the unit test

llvm/lib/Target/AIE/aie2p/AIE2PRegisterBankInfo.cpp
khallouh: FYI @abnikant

llvm/lib/Target/AIE/aie2p/AIE2PRegisterBankInfo.cpp
khallouh: FYI @abnikant

PR #290, opened 2025-01-20
[AIE2P] Select 1024-bit and 512-bit load/store in the FIFO register bank

khallouh:


abnikant (commented):
Can you add mir tests for G_STORE fifos?

gbossu:
I would be happy if you could address the remaining comments from https://github.com/Xilinx/llvm-aie/pull/288 here. I think the code would really gain in clarity, especially with https://github.com/Xilinx/llvm-aie/pull/288#discussion_r1925359230

khallouh:
I addressed your comments from the private repo @mludevid

khallouh:
> I would be happy if you could address the remaining comments from #288 here. I think the code would really gain in clarity, especially with [#288 (comment)](https://github.com/Xilinx/llvm-aie/pull/288#discussion_r1925359230)

This PR aims to support 512-bit and 1024-bit load/store. Since the PR for the FIFO bank #288 was merged before those comments, they can be addressed in the follow-up which I mentioned (to refactor the code between accumulator and fifo handling)

khallouh:
> Can you add mir tests for G_STORE fifos?

Done

mludevid (approved):
LGTM. Thanks for addressing my comments @khallouh!

katerynamuts (approved):

PR #272, opened 2025-01-16
[AIE2P] Combine G_SHUFFLE_VECTOR into G_AIE_VSEL

katerynamuts:

PR #294, opened 2025-01-21
[AIE2P] Fix combine G_SHUFFLE_VECTOR into G_AIE_VSEL

katerynamuts:
There are some new comments in this merged PR: https://github.com/Xilinx/llvm-aie/pull/272
This PR addresses all comments from the previous PR.

### DIFF
--- a/llvm/lib/Target/AIE/AIECombinerHelper.cpp
+++ b/llvm/lib/Target/AIE/AIECombinerHelper.cpp
@@ -19,6 +19,7 @@
 #include "llvm/CodeGen/GlobalISel/Utils.h"
 #include "llvm/CodeGen/MachineBasicBlock.h"
 #include "llvm/CodeGen/MachineInstr.h"
+#include "llvm/CodeGen/MachineInstrBuilder.h"
 #include "llvm/CodeGen/MachineRegionInfo.h"
 #include "llvm/CodeGen/MachineRegisterInfo.h"
 #include "llvm/CodeGen/TargetOpcodes.h"
@@ -1266,17 +1267,6 @@ void llvm::applyPadVector(MachineInstr &MI, MachineRegisterInfo &MRI,
   MI.eraseFromParent();
 }
 
-void llvm::applyVSel(
-    MachineInstr &MI, MachineRegisterInfo &MRI, MachineIRBuilder &B,
-    std::tuple<Register, Register, Register, uint64_t> &MatchInfo) {
-  B.setInstrAndDebugLoc(MI);
-  const AIEBaseInstrInfo &AIETII = (const AIEBaseInstrInfo &)B.getTII();
-  auto [DstVecReg, Src1Reg, Src2Reg, Mask] = MatchInfo;
-  B.buildInstr(AIETII.getGenericVSelOpcode(), {DstVecReg},
-               {Src1Reg, Src2Reg, Mask});
-  MI.eraseFromParent();
-}
-
 /// Match something like this:
 ///  %68:_(s32) = G_CONSTANT i32 0
 ///  %93:_(s32) = G_CONSTANT i32 1
@@ -1785,9 +1775,9 @@ bool llvm::matchShuffleToBroadcast(MachineInstr &MI, MachineRegisterInfo &MRI,
   return true;
 }
 
-bool llvm::matchShuffleToVSel(
-    MachineInstr &MI, MachineRegisterInfo &MRI,
-    std::tuple<Register, Register, Register, uint64_t> &MatchInfo) {
+bool llvm::matchShuffleToVSel(MachineInstr &MI, MachineRegisterInfo &MRI,
+                              const AIEBaseInstrInfo &TII,
+                              BuildFnTy &MatchInfo) {
   assert(MI.getOpcode() == TargetOpcode::G_SHUFFLE_VECTOR);
   const Register DstReg = MI.getOperand(0).getReg();
   const Register Src1Reg = MI.getOperand(1).getReg();
@@ -1796,7 +1786,8 @@ bool llvm::matchShuffleToVSel(
 
   const LLT DstTy = MRI.getType(DstReg);
   const LLT Src1Ty = MRI.getType(Src1Reg);
-  if (Src1Ty.getSizeInBits() != 512)
+  if (Src1Ty.getSizeInBits() != 512 ||
+      Src1Ty.getElementType() == LLT::scalar(64))
     return false;
 
   const unsigned NumDstElems = DstTy.getNumElements();
@@ -1805,30 +1796,35 @@ bool llvm::matchShuffleToVSel(
     return false;
 
   // Check that the shuffle mask can be converted into VSel mask:
+  // The mask contains only -1
+  if (std::all_of(Mask.begin(), Mask.end(),
+                  [&](int Value) { return Value == -1; })) {
+    return false;
+  }
+
   // 1. The shuffle mask doesn't contain indices that correspond to the same
   // index in Src1 and Src2, i.e., for each i only the i-th element from Src1 or
   // the i-th element from Src2 is used.
   // 2. The mask indices modulo the number of elements are in strictly ascending
   // order.
-  int PrevIdx = Mask[0] % NumSrcElems;
-  const size_t NumElems = Mask.size();
-  for (unsigned I = 1; I < NumElems; I++) {
-    int CurrIdx = Mask[I] % NumSrcElems;
-    if (CurrIdx <= PrevIdx)
+  uint64_t DstMask = 0;
+  const size_t NumMaskElems = Mask.size();
+  for (unsigned I = 0; I < NumMaskElems; I++) {
+    int Idx = Mask[I];
+    if (Idx == -1 || Idx == (int)I)
+      continue;
+
+    if ((unsigned)Idx == I + NumSrcElems)
+      DstMask |= uint64_t(1) << I;
+    else
       return false;
   }
 
-  // Create the mask
-  unsigned long long DstMask = 0;
-  for (unsigned I = 0; I < NumElems; I++) {
-    int Idx = Mask[I];
-    if (Idx >= (int)NumSrcElems) {
-      unsigned long long ElemMask = 1 << I;
-      DstMask |= ElemMask;
-    }
-  }
-
-  MatchInfo = std::make_tuple(DstReg, Src1Reg, Src2Reg, DstMask);
+  MatchInfo = [=, &TII](MachineIRBuilder &B) {
+    MachineInstrBuilder MaskReg = B.buildConstant(LLT::scalar(32), DstMask);
+    const unsigned VSelOpc = TII.getGenericVSelOpcode();
+    B.buildInstr(VSelOpc, {DstReg}, {Src1Reg, Src2Reg, MaskReg});
+  };
   return true;
 }
 
--- a/llvm/test/CodeGen/AIE/GlobalISel/prelegalizercombiner-shuffle-vector.mir
+++ b/llvm/test/CodeGen/AIE/GlobalISel/prelegalizercombiner-shuffle-vector.mir
@@ -120,112 +120,227 @@ body:             |
     PseudoRET implicit $lr, implicit %0
 ...
 ---
+name: shuffle_vector_all_undef
+alignment:       16
+exposesReturnsTwice: false
+legalized:       false
+tracksRegLiveness: true
+body:             |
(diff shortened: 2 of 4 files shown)


llvm/lib/Target/AIE/AIECombinerHelper.cpp
```diff
@@ -1791,32 +1791,57 @@ bool llvm::matchShuffleToVSel(
 
   const LLT DstTy = MRI.getType(DstReg);
   const LLT Src1Ty = MRI.getType(Src1Reg);
-  if (Src1Ty.getSizeInBits() != 512)
+  if (Src1Ty.getSizeInBits() != 512 || Src1Ty == LLT::scalar(64))
```
niwinanto: We should be checking `Src1Ty.getElementType() == LLT::scalar(64)`, I mean we should bail out for`<8 x s64>`. Also, good idea to have a test.

llvm/lib/Target/AIE/AIECombinerHelper.cpp
```diff
@@ -1791,32 +1791,57 @@ bool llvm::matchShuffleToVSel(
...
+  // The mask contains only -1
+  if (I == NumElems)
+    return false;
+
+  int CurrIdx = Mask[I] % NumSrcElems;
+  if (CurrIdx <= PrevIdx)
+    return false;
+
+  PrevIdx = CurrIdx;
+  ++I;
```
niwinanto: I think this is not really required, for loop below should handle.

llvm/lib/Target/AIE/AIECombinerHelper.cpp
niwinanto: I can see there is a test already, not sure how it works. May be mask value is wrong.

llvm/test/CodeGen/AIE/GlobalISel/prelegalizercombiner-shuffle-vector.mir
```diff
@@ -207,23 +292,38 @@ exposesReturnsTwice: false
...
+    ; CHECK-LABEL: name: shuffle_vector_vsel_invalid_s64
+    ; CHECK: liveins: $r0, $wl4, $x2
+    ; CHECK-NEXT: {{  $}}
+    ; CHECK-NEXT: [[COPY:%[0-9]+]]:_(<8 x s64>) = COPY $x2
+    ; CHECK-NEXT: [[DEF:%[0-9]+]]:_(<8 x s64>) = G_IMPLICIT_DEF
+    ; CHECK-NEXT: [[SHUF:%[0-9]+]]:_(<8 x s64>) = G_SHUFFLE_VECTOR [[DEF]](<8 x s64>), [[COPY]], shufflemask(8, 1, 2, 3, 11, 12, 13, 14)
+    ; CHECK-NEXT: PseudoRET implicit $lr, implicit [[SHUF]](<8 x s64>)
+    %1:_(<8 x s64>) = COPY $x2
+    %8:_(<8 x s64>) = G_IMPLICIT_DEF
+    %0:_(<8 x s64>) = G_SHUFFLE_VECTOR %8(<8 x s64>), %1, shufflemask(8, 1, 2, 3, 11, 12, 13, 14)
```
niwinanto: mask value should be `shufflemask(8, 1, 2, 3, 12, 13, 14, 15)`

llvm/lib/Target/AIE/AIECombinerHelper.cpp
katerynamuts: You're right, the mask was wrong. I fixed it.

llvm/lib/Target/AIE/AIECombinerHelper.cpp
katerynamuts: Without increasing `I` we would consider `Mask[I]` used in ` int CurrIdx = Mask[++I] % NumSrcElems;` and I think it is wrong. But I made `++I` in a more elegant way: I removed `++I` and changed `int CurrIdx = Mask[I] % NumSrcElems;` to `int CurrIdx = Mask[I++] % NumSrcElems;` .

llvm/test/CodeGen/AIE/GlobalISel/prelegalizercombiner-shuffle-vector.mir
```diff
@@ -180,25 +271,19 @@ exposesReturnsTwice: false
...
-    ; CHECK-NEXT: PseudoRET implicit $lr, implicit $x0
+    ; CHECK-NEXT: [[DEF:%[0-9]+]]:_(<16 x s32>) = G_IMPLICIT_DEF
+    ; CHECK-NEXT: [[SHUF:%[0-9]+]]:_(<16 x s32>) = G_SHUFFLE_VECTOR [[DEF]](<16 x s32>), [[COPY]], shufflemask(0, 1, 2, 3, 16, 21, 22, 23, 24, 25, 26, 27, 28, 29, 30, 31)
+    ; CHECK-NEXT: PseudoRET implicit $lr, implicit [[SHUF]](<16 x s32>)
     %1:_(<16 x s32>) = COPY $x2
-    %4:_(<8 x s32>) = COPY $wl4
-    %3:_(<4 x s32>) = G_AIE_UNPAD_VECTOR %4(<8 x s32>)
-    %8:_(<16 x s32>) = G_AIE_PAD_VECTOR_UNDEF %3(<4 x s32>)
+    %8:_(<16 x s32>) = G_IMPLICIT_DEF
     %0:_(<16 x s32>) = G_SHUFFLE_VECTOR %8(<16 x s32>), %1, shufflemask(0, 1, 2, 3, 16, 21, 22, 23, 24, 25, 26, 27, 28, 29, 30, 31)
```
martien-de-jong, line 298: I find the name of this test strange. No index is repeated, rather, vsel can't reach index 16 in position 4.

llvm/test/CodeGen/AIE/GlobalISel/prelegalizercombiner-shuffle-vector.mir
```diff
@@ -207,25 +292,40 @@ exposesReturnsTwice: false
...
+    ; CHECK-LABEL: name: shuffle_vector_vsel_invalid_s64
+    ; CHECK: liveins: $r0, $wl4, $x2
+    ; CHECK-NEXT: {{  $}}
+    ; CHECK-NEXT: [[COPY:%[0-9]+]]:_(<8 x s64>) = COPY $x2
+    ; CHECK-NEXT: [[DEF:%[0-9]+]]:_(<8 x s64>) = G_IMPLICIT_DEF
+    ; CHECK-NEXT: [[SHUF:%[0-9]+]]:_(<8 x s64>) = G_SHUFFLE_VECTOR [[DEF]](<8 x s64>), [[COPY]], shufflemask(8, 1, 2, 3, 12, 13, 14, 15)
+    ; CHECK-NEXT: PseudoRET implicit $lr, implicit [[SHUF]](<8 x s64>)
+    %1:_(<8 x s64>) = COPY $x2
+    %8:_(<8 x s64>) = G_IMPLICIT_DEF
+    %0:_(<8 x s64>) = G_SHUFFLE_VECTOR %8(<8 x s64>), %1, shufflemask(8, 1, 2, 3, 12, 13, 14, 15)
```
martien-de-jong, line 342: Isn't this equivalent to {16, 17, 2, 3, 4, 5, 6, 7, 24, 25, 26, 27, 28, 29, 30, 31} on 16 x 32 ?

llvm/lib/Target/AIE/AIECombinerHelper.cpp
```diff
@@ -1810,18 +1811,37 @@ bool llvm::matchShuffleToVSel(
   // the i-th element from Src2 is used.
   // 2. The mask indices modulo the number of elements are in strictly ascending
   // order.
```
martien-de-jong, line 1809: Could this be simplified by saying that shufflemask[i] should be -1, i or i + elemcount?
I'm assuming -1 is don't care, and we choose it to be i, leading to a zero bit in the select mask.
We then just need to run one loop, returning false as soon as we find an illegal index

llvm/test/CodeGen/AIE/GlobalISel/prelegalizercombiner-shuffle-vector.mir
katerynamuts: 16 is the index 0 of the second vector, that's what I meant here :) Any suggestions for the better name? Nothing comes to my mind.

llvm/test/CodeGen/AIE/GlobalISel/prelegalizercombiner-shuffle-vector.mir
katerynamuts: yes, it is but here I wanted to test that we never substitute `G_SHUFFLE_VECTOR ` with `G_AIE_VSEL` for ``s64` because during instruction selection we map `G_AIE_VSEL` to `VSEL` where the words can be 8, 16, or 32-bit wide.

llvm/lib/Target/AIE/AIECombinerHelper.cpp
```diff
@@ -1805,30 +1795,44 @@ bool llvm::matchShuffleToVSel(
...
+
     if (Idx >= (int)NumSrcElems) {
       unsigned long long ElemMask = 1 << I;
       DstMask |= ElemMask;
     }
   }
 
-  MatchInfo = std::make_tuple(DstReg, Src1Reg, Src2Reg, DstMask);
+  MatchInfo = [=, &TII](MachineIRBuilder &B) {
+    auto Cst = B.buildConstant(LLT::scalar(32), DstMask);
```
andcarminati: nit: `Register MaskReg =`

llvm/lib/Target/AIE/AIECombinerHelper.cpp
```diff
@@ -1805,30 +1795,44 @@ bool llvm::matchShuffleToVSel(
...
+  int I = -1;
+  if (!std::all_of(Mask.begin(), Mask.end(), [&](int Value) {
+        ++I;
+        return (Value == -1 || Value == I || Value == I + (int)NumSrcElems);
+      })) {
+    return false;
   }
 
   // Create the mask
   unsigned long long DstMask = 0;
```
andcarminati: I think we should replace this (and the next one) type by the standard type `uint64_t`.

martien-de-jong: Although unsigned long long has been longer in the C standard and is guaranteed to be at least 64 bits.

llvm/lib/Target/AIE/AIECombinerHelper.cpp
```diff
@@ -1805,30 +1796,44 @@ bool llvm::matchShuffleToVSel(
...
   }
 
   // Create the mask
-  unsigned long long DstMask = 0;
-  for (unsigned I = 0; I < NumElems; I++) {
+  uint64_t DstMask = 0;
+  const size_t NumMaskElems = Mask.size();
+  for (unsigned I = 0; I < NumMaskElems; I++) {
     int Idx = Mask[I];
+    if (Idx == -1)
```
martien-de-jong: Actually I had meant to put the full check of the above `std::all_of` in here; we can safely return false if  the condition doesn't hold; `Value == I + (int)NumSrcElems` can be reused as the test to insert a one bit.
Also, the fact that we have to capture `I` by reference makes it a bit unnatural to use std::all_of.
```
if (Idx == -1 || Idx == I)
  continue;
else if (Idx == I + NumSrcElements)
 DstMask |= uint64_t(1) << I;
else
  return false;
```

llvm/test/CodeGen/AIE/GlobalISel/prelegalizercombiner-shuffle-vector.mir
martien-de-jong: I view it as element 16 not suitable for position 4. lane_mismatch? unaligned_index?

llvm/test/CodeGen/AIE/GlobalISel/prelegalizercombiner-shuffle-vector.mir
martien-de-jong: Ok. Park it for now as an improvement story?

llvm/lib/Target/AIE/AIECombinerHelper.cpp
katerynamuts: ohh, I got it. Thanks! I changed it.

martien-de-jong (approved):

PR #296, opened 2025-01-21
[AIE2P] Remove assert from matchShuffleToVSel

katerynamuts:


abnikant (approved):
Thanks. The example is going through now.

PR #292, opened 2025-01-21
[AIE2P] Implemented VST.SRS combine

katerynamuts:

PR #280, opened 2025-01-17
[AIE2P] Implemented VST.PACK combine

katerynamuts:

PR #299, opened 2025-01-22
[AIE2P] Implemented VLDB.UNPACK combine

katerynamuts:


llvm/lib/Target/AIE/AIE2InstructionSelector.cpp
```diff
@@ -978,7 +985,7 @@ bool AIE2InstructionSelector::selectG_AIE_LOAD_UNPACK(
       return false;
   }
 
-  if (!canCombineUNPACKLoad(*LoadOp, UNPACKI, MRI) ||
+  if (/*!canCombineUNPACKLoad(*LoadOp, UNPACKI, MRI) ||*/
```
andcarminati: Why is this commented?

andcarminati: We no further check if `LSO` is `nullopt`.

katerynamuts: Good question! I uncommented it and I also added this check for aie2p because it was missing there.

llvm/lib/Target/AIE/AIE2InstructionSelector.cpp
```diff
@@ -990,6 +997,9 @@ bool AIE2InstructionSelector::selectG_AIE_LOAD_UNPACK(
   std::optional<LoadStoreOpcodes> LSO = getCombinedOpcodeUNPACKLoad(
       AMI->MemI, UNPACKI, AMI->ImmediateOffset, MRI);
 
+  if (!LSO.has_value())
+    return false;
```
mludevid: We precisely call `canCombineUNPACKLoad` before, to avoid `getCombinedOpcodeUNPACKLoad` to ever fail here. I appreciate the additional check, but I would prefer this to be an assert instead.

llvm/lib/Target/AIE/aie2p/AIE2PInstructionSelector.cpp
```diff
@@ -2578,6 +2588,296 @@ bool AIE2PInstructionSelector::selectG_AIE_STORE_CONV(
...
+  Register SignReg = UNPACKI.getOperand(3).getReg();
+
+  auto SignVal = getIConstantVRegValWithLookThrough(SignReg, MRI);
+  bool ConstantSign = SignVal ? true : false;
+  std::optional<LoadStoreOpcodes> LSO = getCombinedOpcodeUNPACKLoad(
+      *LoadOp, UNPACKI, AMI->ImmediateOffset,
+      ConstantSign ? SignVal.value().Value == 0x1 : false);
+
+  if (!LSO.has_value())
+    return false;
```
mludevid: As mentioned in another comment, please use an assert instead.

llvm/lib/Target/AIE/aie2p/AIE2PInstructionSelector.cpp
```diff
@@ -2578,6 +2588,296 @@ bool AIE2PInstructionSelector::selectG_AIE_STORE_CONV(
...
+
+  assert(LoadOp && "Expected SSA.");
+
+  // Do not try to combine if one of the load's defs is used by another
+  // instruction between the load and the VUNPACK or if there is a store
+  // between the load and the VUNPACK.
+  if (!canDelayMemOp(*LoadOp, UNPACKI, MRI)) {
+    // If we cannot delay the load, we can try to advance the combined
+    // instruction to the load's position.
+    if (canAdvanceOp(*LoadOp, UNPACKI, MRI))
```
mludevid, line 3727: Nice! Is there any reason we don't do this for the other combines?

llvm/lib/Target/AIE/aie2p/AIE2PInstructionSelector.cpp
```diff
@@ -2578,6 +2588,296 @@ bool AIE2PInstructionSelector::selectG_AIE_STORE_CONV(
...
+    MIB.setInstr(UNPACKI);
+
+  // Selects the size of the UNPACK instructions
+  // 0 – Source is 4 bits
+  // 1 – Source is 8 bits
+  switch (cast<GIntrinsic>(UNPACKI).getIntrinsicID()) {
+  case Intrinsic::aie2p_unpack_I512_I8_I4:
+  case Intrinsic::aie2p_unpack_I1024_I8_I4:
+    MIB.buildInstr(AIE2P::MOV_scalar_imm11_pseudo, {AIE2P::crUnpackSize}, {})
+        .addImm(0);
```
mludevid: Two thoughts on this:
1. As I also mention in the VST.PACK PR (#280), we have mutliple places where we have to set a constant. Could we please extract that into a function?
2. Is there a reason why we are not using `getMvSclMultiSlotPseudoOpcode` here?

llvm/lib/Target/AIE/aie2p/AIE2PInstructionSelector.cpp
```diff
@@ -2578,6 +2588,296 @@ bool AIE2PInstructionSelector::selectG_AIE_STORE_CONV(
...
+
+  UNPACKI.eraseFromParent();
+
+  // Erasing the load instruction breaks later on in the selection code. That is
+  // because we keep an iterator on erased instructions. This breaks while
+  // trying to eliminate a trivially dead instruction which requires access to
+  // its memory operands which have been erased, thus leading to a seg fault. To
+  // remedy this, we keep the load to be removed by the trivial dead code
+  // elimination and we make sure to assign a new virtual register definition to
+  // its live operands to respect SSA.
```
mludevid: Nit: I think at this point, we no longer need this comment, we use `makeDeadMI` in multiple instruction selection methods.

llvm/lib/Target/AIE/aie2p/AIE2PInstructionSelector.cpp
```diff
@@ -2578,6 +2588,296 @@ bool AIE2PInstructionSelector::selectG_AIE_STORE_CONV(
...
+  const std::optional<APInt> NoImmediate = {};
+  bool IsSigned = false;
+  return getCombinedOpcodeUNPACKLoad(MemOp, CombOp, NoImmediate, IsSigned)
+      .has_value();
+}
+
+bool AIE2PInstructionSelector::selectG_AIE_LOAD_UNPACK(
+    MachineInstr &UNPACKI, MachineRegisterInfo &MRI) {
+  Register LoadResult = (std::next(UNPACKI.uses().begin()))->getReg();
+  MachineInstr *LoadOp = getDefIgnoringCopiesAndBitcasts(LoadResult, MRI);
```
mludevid, line 3715: Why are we only using `getDefIgnoringCopiesAndBitcasts` here? Could we use this in all other combiners as well?

llvm/lib/Target/AIE/aie2p/AIE2PInstructionSelector.cpp
```diff
@@ -2578,6 +2588,296 @@ bool AIE2PInstructionSelector::selectG_AIE_STORE_CONV(
...
+        return LoadStoreOpcodes{
+            AIE2P::VLDB_UNPACK_dmw_ldb_unpack_idx_imm_unpackSign1,
+            AlwaysFitsImmediateRange,
+            {}};
+      }
+    case AIE2P::G_AIE_OFFSET_LOAD:
+      switch (CombOpInstID) {
+      case Intrinsic::aie2p_unpack_I1024_I8_I4:
+      case Intrinsic::aie2p_unpack_I1024_I16_I8:
+        FitsImmediateRange = checkImmediateRange<4, 64>(Immediate);
```
mludevid, line 3556: Shouldn't it be `checkImmediateRange<4, 32>`? If so, please apply these changes to all 512 bit loads (1024 when unpacked) in this function.

llvm/test/CodeGen/AIE/aie2p/GlobalIsel/inst-select-indexed-vldb_unpack.mir
mludevid, line 1: Similarly to #292, the offsets in the tests need to be updated, because the immediate ranges have changed from AIE2 to AIE2P. Please also fix them in the post-increment test file.

mludevid (commented):
Everything else seems good. The test files need to be checked again once they have been updated.

llvm/lib/Target/AIE/aie2p/AIE2PInstructionSelector.cpp
katerynamuts: I don't think so. @andcarminati did it a couple of months ago for aie2 and suggested to do it also for aie2p, which is absolutely makes sense!

llvm/lib/Target/AIE/aie2p/AIE2PInstructionSelector.cpp
katerynamuts: I would rather keep it because if someone reads exactly this function then the comment helps to understand why we do it.

llvm/lib/Target/AIE/aie2p/AIE2PInstructionSelector.cpp
katerynamuts: No, `<4, 64>` is correct, I doublechecked.

llvm/lib/Target/AIE/AIEBaseInstructionSelector.cpp
mludevid, line 1: Please update the year in the file header.

llvm/lib/Target/AIE/AIEBaseInstructionSelector.h
mludevid, line 1: Please update the year in the file header.

llvm/lib/Target/AIE/AIE2InstructionSelector.cpp
mludevid, line 1: Could you also use `AIEBaseInstructionSelector::setCtrlRegister` in `selectSetControlRegister`?

llvm/lib/Target/AIE/aie2p/AIE2PInstructionSelector.cpp
mludevid, line 1: As in `AIE2InstructionSelector.cpp`: Could you also use `AIEBaseInstructionSelector::setCtrlRegister` in `selectSetControlRegister`?

llvm/lib/Target/AIE/aie2p/AIE2PInstructionSelector.cpp
```diff
@@ -1268,7 +1273,7 @@ bool AIE2PInstructionSelector::selectG_AIE_LOAD_UPS(MachineInstr &UPSI,
 
   // First use is the G_INTRINSIC_W_SIDE_EFFECTS ID
   Register LoadResult = (std::next(UPSI.uses().begin()))->getReg();
-  MachineInstr *LoadOp = MRI.getUniqueVRegDef(LoadResult);
+  MachineInstr *LoadOp = getDefIgnoringCopiesAndBitcasts(LoadResult, MRI);
```
mludevid, line 1309: Can we add a test for this change?

llvm/lib/Target/AIE/aie2p/AIE2PInstructionSelector.cpp
```diff
@@ -2672,7 +2677,7 @@ bool AIE2PInstructionSelector::selectG_AIE_STORE_CONV(
   // differ.
 
   Register ConvResult = (StoreI.uses().begin())->getReg();
-  MachineInstr *ConvOp = MRI.getUniqueVRegDef(ConvResult);
+  MachineInstr *ConvOp = getDefIgnoringCopiesAndBitcasts(ConvResult, MRI);
```
mludevid, line 3470: Same thing as above: Could you add a test for this change?

llvm/lib/Target/AIE/aie2p/AIE2PInstructionSelector.cpp
mludevid: Thanks for implementing this!

llvm/lib/Target/AIE/aie2p/AIE2PInstructionSelector.cpp
mludevid: You are totally right! It was the other ones that needed to be updated, as you have already done.

llvm/test/CodeGen/AIE/aie2p/GlobalIsel/inst-select-postinc-vldb_unpack.mir
```diff
@@ -0,0 +1,681 @@
...
+    ; CHECK-NEXT: $crunpacksize = MOV_scalar_imm11_pseudo 0
+    ; CHECK-NEXT: [[VLDB_UNPACK_dmw_ldb_unpack_pstm_nrm_unpackSign1_2:%[0-9]+]]:vec512, [[VLDB_UNPACK_dmw_ldb_unpack_pstm_nrm_unpackSign1_3:%[0-9]+]]:ep = VLDB_UNPACK_dmw_ldb_unpack_pstm_nrm_unpackSign1 [[VLDB_UNPACK_dmw_ldb_unpack_pstm_nrm_imm_unpackSign1_5]], [[MOV_PD_imm11_pseudo]], implicit $crunpacksize, implicit $unpacksign1 :: (load (<32 x s8>))
+    ; CHECK-NEXT: PseudoRET implicit $lr, implicit [[VLDB_UNPACK_dmw_ldb_unpack_pstm_nrm_unpackSign1_]], implicit [[VLDB_UNPACK_dmw_ldb_unpack_pstm_nrm_imm_unpackSign1_]], implicit [[VLDB_UNPACK_dmw_ldb_unpack_pstm_nrm_imm_unpackSign1_2]], implicit [[VLDB_UNPACK_dmw_ldb_unpack_pstm_nrm_imm_unpackSign1_4]], implicit [[VLDB_UNPACK_dmw_ldb_unpack_pstm_nrm_unpackSign1_2]]
+    %0:ptrregbank(p0) = COPY $p0
+    %1:gprregbank(s32) = COPY $r0
+    %7:modregbank(s20) = G_TRUNC %1
+    %8:modregbank(s20) = G_CONSTANT i20 0
+    %9:modregbank(s20) = G_CONSTANT i20 -256
+    %10:modregbank(s20) = G_CONSTANT i20 224
+    %11:modregbank(s20) = G_CONSTANT i20 256
```
mludevid: Is there a reason you skipped the offset test with half a step size?

llvm/lib/Target/AIE/aie2p/AIE2PInstructionSelector.cpp
```diff
@@ -2708,6 +2713,293 @@ bool AIE2PInstructionSelector::selectG_AIE_STORE_CONV(
...
+
+  assert(((getLoadStoreSize(MemOp) == 256 &&
+           (CombOpInstID == Intrinsic::aie2p_unpack_I512_I16_I8 ||
+            CombOpInstID == Intrinsic::aie2p_unpack_I512_I8_I4)) ||
+          (getLoadStoreSize(MemOp) == 512 &&
+           (CombOpInstID == Intrinsic::aie2p_unpack_I1024_I16_I8 ||
+            CombOpInstID == Intrinsic::aie2p_unpack_I1024_I8_I4))) &&
+         "Unexpected VLDB.UNPACK size");
+
+  unsigned ISelOpcode;
```
martien-de-jong: I would declare a local version where and when it is needed. Otherwise, please initialize it to 'the illegal opcode'

llvm/lib/Target/AIE/aie2p/AIE2PInstructionSelector.cpp
```diff
@@ -2708,6 +2713,293 @@ bool AIE2PInstructionSelector::selectG_AIE_STORE_CONV(
...
+                         : AIE2P::VLDB_UNPACK_dmx_ldb_unpack_idx_unpackSign1;
+        return LoadStoreOpcodes{ISelOpcode, FitsImmediateRange, {}};
+      case Intrinsic::aie2p_unpack_I512_I8_I4:
+      case Intrinsic::aie2p_unpack_I512_I16_I8:
+        FitsImmediateRange = checkImmediateRange<4, 32>(Immediate);
+        ISelOpcode = FitsImmediateRange
+                         ? AIE2P::VLDB_UNPACK_dmw_ldb_unpack_idx_imm_unpackSign1
+                         : AIE2P::VLDB_UNPACK_dmw_ldb_unpack_idx_unpackSign1;
+        return LoadStoreOpcodes{ISelOpcode, FitsImmediateRange, {}};
+      }
```
martien-de-jong, line 3568: I know the default case doesn't occur, but could we at least put a break between the cases of the outer switch? (if only to help the poor C++ compiler.)

llvm/lib/Target/AIE/aie2p/AIE2PInstructionSelector.cpp
martien-de-jong: I think explaining it once at the definition of makeDeadMI would suffice. It's just one vscode click away. I also think that the comment is overly woolly. 'We can not erase this since it would invalidate active iterators' says it all.

llvm/lib/Target/AIE/aie2p/AIE2PInstructionSelector.cpp
```diff
@@ -2708,6 +2713,293 @@ bool AIE2PInstructionSelector::selectG_AIE_STORE_CONV(
...
+
+bool AIE2PInstructionSelector::canCombineUNPACKLoad(MachineInstr &MemOp,
+                                                    MachineInstr &CombOp,
+                                                    MachineRegisterInfo &MRI) {
+  Register LoadResult = (std::next(CombOp.uses().begin()))->getReg();
+  if (MemOp.getParent() != CombOp.getParent() || !MRI.hasOneUse(LoadResult))
+    return false;
+
+  const std::optional<APInt> NoImmediate = {};
+  bool IsSigned = false;
```
martien-de-jong: nit: `const bool IsSigned = false;`

llvm/lib/Target/AIE/aie2p/AIE2PInstructionSelector.cpp
```diff
@@ -2708,6 +2713,293 @@ bool AIE2PInstructionSelector::selectG_AIE_STORE_CONV(
...
+      }
+    }
+  }
+  return {};
+}
+
+bool AIE2PInstructionSelector::canCombineUNPACKLoad(MachineInstr &MemOp,
+                                                    MachineInstr &CombOp,
+                                                    MachineRegisterInfo &MRI) {
+  Register LoadResult = (std::next(CombOp.uses().begin()))->getReg();
```
martien-de-jong: I don't understand this. AFAICS, MemOp is the Load. Why isn't LoadResult the register of its first def?

llvm/lib/Target/AIE/aie2p/AIE2PInstructionSelector.cpp
```diff
@@ -2708,6 +2713,293 @@ bool AIE2PInstructionSelector::selectG_AIE_STORE_CONV(
...
+          (getLoadStoreSize(MemOp) == 512 &&
+           (CombOpInstID == Intrinsic::aie2p_unpack_I1024_I16_I8 ||
+            CombOpInstID == Intrinsic::aie2p_unpack_I1024_I8_I4))) &&
+         "Unexpected VLDB.UNPACK size");
+
+  unsigned ISelOpcode;
+  const bool AlwaysFitsImmediateRange = true;
+  const bool NoImmediate = false;
+  bool FitsImmediateRange = false;
+  if (IsSigned) {
```
martien-de-jong, line 3534: AFAICS, the else branch differs only in selecting a different opcode. I would arrange for that with a small function unsigned selectSignedness(int Opcode, IsSigned)

llvm/test/CodeGen/AIE/aie2p/GlobalIsel/inst-select-postinc-vldb_unpack.mir
katerynamuts: No, I think I just missed it,  I added it to the test.

llvm/lib/Target/AIE/aie2p/AIE2PInstructionSelector.cpp
katerynamuts: I cannot create local versions in this case, I would need to give all of them different names, so I initialized it with `UINT_MAX`.

llvm/lib/Target/AIE/aie2p/AIE2PInstructionSelector.cpp
katerynamuts: it is a good idea and I think it makes sense to do it for all combines but for now I would leave it as it is and implement these functions for all combines in a follow-up PR.

llvm/lib/Target/AIE/aie2p/AIE2PInstructionSelector.cpp
```diff
@@ -3515,6 +3529,294 @@ bool AIE2PInstructionSelector::selectG_AIE_STORE_CONV(
...
+
+  addAddressingMode(NewInstr, *AMI, LSO->FitsImmediateRange, false, MRI);
+
+  NewInstr.cloneMemRefs(*LoadOp);
+
+  if (!ConstantSign)
+    setUnsetCtrlRegister(MIB, *NewInstr, MRI, AIE2P::unpackSign0, SignReg);
+
+  UNPACKI.eraseFromParent();
+  makeDeadMI(*LoadOp, MRI);
```
martien-de-jong, line 3789: CHECK: we have checked that the unpack was the only use of the load

katerynamuts: yes, it is checked in `canCombineUNPACKLoad`.

mludevid: Since we also allow copies and bitcasts between the load and the unpack, I think it's also important to note, that `getDefIgnoringCopiesAndBitcasts` also makes sure that there is only one user of the result of the `COPY` or the `BITCAST`.

llvm/lib/Target/AIE/aie2p/AIE2PInstructionSelector.cpp
martien-de-jong: I think block scope was invented for that. But I see it's not your fault...

llvm/lib/Target/AIE/aie2p/AIE2PInstructionSelector.cpp
```diff
@@ -778,9 +785,7 @@ bool AIE2PInstructionSelector::selectSetControlRegister(
       break;
     }
 
-    MachineInstrBuilder MI =
-        MIB.buildInstr(AIE2P::MOV_scalar_imm11_pseudo, {CtrlReg}, {})
-            .addImm(SrcConstVal);
+    MachineInstrBuilder MI = setCtrlRegister(MIB, CtrlReg, SrcConstVal);
```
martien-de-jong, line 788: Tip: This set of changes might have comprised a separate commit.

martien-de-jong (approved):

PR #300, opened 2025-01-22
[AIE2P] Fix EX registers' Calling Convention to follow AIE ABI

khallouh:
AIE2P's ABI requires that pair EX registers are passed before odd ones. This fixes the call lowering to follow this requirement.


niwinanto (approved):
LGTM!

PR #301, opened 2025-01-23
[AIE2P]Handle stack spill offset out of bound values

niwinanto:
This PR handles, offset values bigger than `ST_dms_sts_spill/LDA_dms_lda_spill` can encode. Idea is to move the offset and sp register to pointer register and load/store through this point register. One problem is, frame finalization and prolog/epilog happens well after register allocation and finding physical register is difficult. However, we can use register scavenger to find free register for us. This does not mean, we always find success, in that scenario, we should create an emergency spill slot that register scavenger can make use of. How ever, we decided to not create this for AIE, since it can have impact on QoR. Also, we believe, over time, we improve the code selection, stack size will reduce and we might not even face this problem.

Also, extended `expandSpillPseudo` to handle spills using register offset.

Note:
Only `LDA_dms_lda_spill`, `ST_dms_sts_spill` and other pseudos made using these supported. Other instruction can encode more offset and might not run into this issue.


llvm/lib/Target/AIE/aie2p/AIE2PRegisterInfo.cpp
```diff
@@ -141,15 +151,72 @@ bool AIE2PRegisterInfo::eliminateFrameIndex(MachineBasicBlock::iterator II,
...
+  case AIE2P::ST_DS_SPILL: {
+    if (isEncodableAsNegativeInt<9, 4>(Offset)) {
+      MI.getOperand(FIOperandNum).ChangeToImmediate(Offset);
+    } else {
+      Register SPReg =
+          MF.getRegInfo().createVirtualRegister(&AIE2P::ePRegClass);
+      BuildMI(MBB, II, DL, TII->get(TII->getMvSclOpcode()), SPReg)
+          .addReg(getStackPointerRegister());
+      TII->expandSpillPseudo(MI, TRI, /*SubRegOffsetAlign=*/Align(4), SPReg,
+                             Offset);
```
gbossu, line 197: Could we maybe always spill instructions here, regardless of the offset? I find it a bit confusing that some instructions are expanded here, and some in a different pass.

llvm/lib/Target/AIE/AIEBaseInstrInfo.cpp
```diff
@@ -420,7 +422,7 @@ void AIEBaseInstrInfo::expandSpillPseudo(MachineInstr &MI,
   const MachineOperand &RegOp = MI.getOperand(0);
   Register Reg = RegOp.getReg();
   unsigned RegFlags = computeRegStateFlags(RegOp);
-  int64_t Offset = MI.getOperand(1).getImm();
+  int64_t Offset = OffsetVal ? OffsetVal : MI.getOperand(1).getImm();
```
gbossu: I think it would be clearer to always pass a valid offset, and stop relying on `MI.getOperand(1).getImm()`

llvm/lib/Target/AIE/AIEBaseInstrInfo.h
```diff
@@ -533,15 +534,19 @@ struct AIEBaseInstrInfo : public TargetInstrInfo {
...
   /// will essentially split the register being handled into its sub-registers,
   /// until there is an actual instruction that can handle them.
   /// In case sub-registers don't have nice offset, those can be aligned using
   /// \ref SubRegOffsetAlign.
+  /// In case offset value is above the encodable limit provide with a pointer
+  /// register points to stack pointer and actual offset value.
   void expandSpillPseudo(MachineInstr &MI, const TargetRegisterInfo &TRI,
-                         Align SubRegOffsetAlign = Align(1)) const;
+                         Align SubRegOffsetAlign = Align(1),
+                         Register SPReg = {}, int64_t OffsetVal = 0) const;
```
gbossu: Nit: Maybe use `Register SPReg = MCRegister::NoRegister`. Or a `std::optional<Register>`

llvm/lib/Target/AIE/aie2p/AIE2PRegisterInfo.cpp
```diff
@@ -141,15 +151,72 @@ bool AIE2PRegisterInfo::eliminateFrameIndex(MachineBasicBlock::iterator II,
...
+        auto StoreMI = BuildMI(MBB, II, DL, TII->get(AIE2P::ST_dms_sts_idx))
+                           .addReg(MI.getOperand(0).getReg())
+                           .addReg(SPReg)
+                           .addReg(OffsetReg);
+        StoreMI.cloneMemRefs(MI);
+      }
+      II->removeFromParent();
+    }
+    return false;
+  }
```
gbossu: I think that is a sign that we should create `LDA_R_SPILL` and `ST_R_SPILL` pseudo instructions. No need for special casing then.

llvm/lib/Target/AIE/aie2p/AIE2PRegisterInfo.cpp
```diff
@@ -141,15 +151,72 @@ bool AIE2PRegisterInfo::eliminateFrameIndex(MachineBasicBlock::iterator II,
...
+                             Offset);
+    }
+    return false;
+  }
   case AIE2P::LDA_dmv_lda_q_spill:
   case AIE2P::VLDA_128_dmv_lda_w_spill:
   case AIE2P::VLDA_dmw_lda_w_spill:
   case AIE2P::VLDA_dmx_lda_bm_spill:
   case AIE2P::VLDA_dmx_lda_fifohl_spill:
   case AIE2P::VLDA_dmx_lda_x_spill:
```
gbossu: What about all these other instructions, can we always encode the immediate?

llvm/lib/Target/AIE/aie2p/AIE2PRegisterInfo.cpp
niwinanto: Basically we need 2 registers, one for stack pointer and one for holding the offset value. Every split of pseudo instructions require updating the register value. How ever, register scavenger has more rigid contains, cannot define again(like ssa). We need to create the register for every split. In short, we need to create offset register based on need(how many split).

llvm/lib/Target/AIE/aie2p/AIE2PRegisterInfo.cpp
niwinanto: Other than FIFO and PLFR, I have verified them in the past, they all have more encoding bits. FIFO and PLFR, if needed will address in a followup PR.

llvm/lib/Target/AIE/AIEBaseInstrInfo.cpp
niwinanto: I am not sure there is a way to get offset other than `MI.getOperand(1).getImm()`, from `postrapseudos`

llvm/lib/Target/AIE/AIEBaseInstrInfo.h
niwinanto: Done!

llvm/lib/Target/AIE/aie2p/AIE2PRegisterInfo.cpp
niwinanto: Hmm. yes and no. I would like to keep it this way for now.

llvm/lib/Target/AIE/aie2p/AIE2PInstrInfo.cpp
```diff
@@ -1005,6 +1005,19 @@ AIE2PInstrInfo::getSpillPseudoExpandInfo(const MachineInstr &MI) const {
...
+AIEBaseInstrInfo::AIERegOffsetSpillInstrInfo
+AIE2PInstrInfo::getRegOffsetSpillInstrInfoFromImmOffset(
+    const unsigned OpCode) const {
+  switch (OpCode) {
+  case AIE2P::ST_dms_sts_spill:
+    return {AIE2P::ST_dms_sts_idx, AIE2P::MOVXM, &AIE2P::eDJRegClass};
+  case AIE2P::LDA_dms_lda_spill:
+    return {AIE2P::LDA_dms_lda_idx, AIE2P::MOVXM, &AIE2P::eDJRegClass};
+  default:
+    llvm_unreachable("Un-implemented");
```
andcarminati: nit: a more comprehensive message.

llvm/lib/Target/AIE/aie2p/AIE2PInstrInfo.cpp
```diff
@@ -1005,6 +1005,19 @@ AIE2PInstrInfo::getSpillPseudoExpandInfo(const MachineInstr &MI) const {
   llvm_unreachable("Un-implemented");
 }
 
+AIEBaseInstrInfo::AIERegOffsetSpillInstrInfo
+AIE2PInstrInfo::getRegOffsetSpillInstrInfoFromImmOffset(
+    const unsigned OpCode) const {
```
andcarminati: nit: Opcode

llvm/lib/Target/AIE/aie2p/AIE2PRegisterInfo.cpp
```diff
@@ -23,6 +23,7 @@
 #include "llvm/CodeGen/RegisterScavenging.h"
 #include "llvm/CodeGen/TargetFrameLowering.h"
 #include "llvm/CodeGen/TargetRegisterInfo.h"
+#include "llvm/Support/Debug.h"
```
andcarminati: nit: remove this include.

llvm/lib/Target/AIE/aie2p/AIE2PRegisterInfo.cpp
```diff
@@ -141,15 +151,72 @@ bool AIE2PRegisterInfo::eliminateFrameIndex(MachineBasicBlock::iterator II,
...
+    // an emergency slot in `processFunctionBeforeFrameFinalized`
+
+    // Note that LDB path does not support SPILL instructions
+
   case AIE2P::LDA_dms_lda_spill:
+  case AIE2P::ST_dms_sts_spill: {
+    if (isEncodableAsNegativeInt<9, 4>(Offset)) {
+      MI.getOperand(FIOperandNum).ChangeToImmediate(Offset);
+    } else {
+      Register OffsetReg =
```
andcarminati: Why we have different rules here (directly expansion and pseudo expansion)? I mean, I can understand because we have real instructions and pseudos, but it could help if we can have some comments.

llvm/lib/Target/AIE/aie2p/AIE2PRegisterInfo.cpp
andcarminati: I agree with that we could have a // TODO for this change.

llvm/lib/Target/AIE/aie2p/AIE2PRegisterInfo.cpp
andcarminati: I would help if every pseudo could end with _pseudo. Nothing to do with this PR ;-)

llvm/lib/Target/AIE/aie2p/AIE2PFrameLowering.cpp
```diff
@@ -93,3 +95,19 @@ void AIE2PFrameLowering::determineCalleeSaves(MachineFunction &MF,
...
     SavedRegs.set(FPReg);
 }
+
+void AIE2PFrameLowering::processFunctionBeforeFrameFinalized(
+    MachineFunction &MF, RegScavenger *RS) const {
+  const TargetRegisterInfo *RegInfo = MF.getSubtarget().getRegisterInfo();
+  MachineFrameInfo &MFI = MF.getFrameInfo();
+  const TargetRegisterClass *RC1 = &AIE2P::ePRegClass;
+  const TargetRegisterClass *RC2 = &AIE2P::eDJRegClass;
+  if (!isInt<12>(MFI.estimateStackSize(MF))) {
```
andcarminati, line 107: nit: comment about the purpose of this code. This contradicts: 

> How ever, we decided to not create this for AIE, since it can have impact on QoR. Also, we believe, over time, we improve the code selection, stack size will reduce and we might not even face this problem.

?

andcarminati:
Nice work, some comments left, mostly to give some explanations! I expect to trigger this code only for exceptional cases ;-)

llvm/lib/Target/AIE/aie2p/AIE2PFrameLowering.cpp
niwinanto: That was our initial plan, later we ended up in failures because RS could not find the free register.

andcarminati (dismissed):
LGTM.

llvm/lib/Target/AIE/aie2p/AIE2PRegisterInfo.cpp
gbossu: I understand why we need to expand before `postrapseudos`, my point was to now completely stop expanding spill pseudos in `postrapseudos` and instead expand **all of them** (even those that have small offsets) here in `eliminateFrameIndex()`. This would make the code a lot more straightforward to understand, otherwise it will be a confusing implementation detail to remember in a month's time.

llvm/lib/Target/AIE/AIEBaseInstrInfo.cpp
gbossu: This goes with my previous comment here https://github.com/Xilinx/llvm-aie/pull/301#discussion_r1927011983 to make the expansion more consistent. If we do everything from elimintateFrameIndex, we have access to the `Offset`. if you cannot do that, I'd say to at least make `OffsetVal` an `std::optional`.

llvm/lib/Target/AIE/aie2p/AIE2PRegisterInfo.cpp
gbossu: > Hmm. yes and no

Could you give me details? Generally, I think reducing the amount of custom c++ code helps readability (and makes potential bugs easier to catch)

llvm/lib/Target/AIE/aie2p/AIE2PRegisterInfo.cpp
niwinanto: I perceived it differently. Now it is clear. I can remove spill expansion from `postrapseudos`

llvm/lib/Target/AIE/AIEBaseInstrInfo.cpp
niwinanto: Now this comment make sense. Since AIE2 uses it as well, I will go ahead with `std::optional`

llvm/lib/Target/AIE/aie2p/AIE2PRegisterInfo.cpp
niwinanto: Yes I agree, reducing the custom c++ code is good. But, I felt creating a pseudo instruction just to get rid of a small custom code bit overwhelming

llvm/lib/Target/AIE/AIEBaseInstrInfo.cpp
```diff
@@ -420,7 +420,8 @@ void AIEBaseInstrInfo::expandSpillPseudo(MachineInstr &MI,
   const MachineOperand &RegOp = MI.getOperand(0);
   Register Reg = RegOp.getReg();
   unsigned RegFlags = computeRegStateFlags(RegOp);
-  int64_t Offset = MI.getOperand(1).getImm();
+  int64_t Offset =
+      OffsetVal.has_value() ? OffsetVal.value() : MI.getOperand(1).getImm();
```
gbossu: Super-nit: `*OffsetVal` instead of `OffsetVal.value()`, otherwise there are runtime checks twice.

llvm/lib/Target/AIE/aie2p/AIE2PRegisterInfo.cpp
```diff
@@ -141,15 +150,74 @@ bool AIE2PRegisterInfo::eliminateFrameIndex(MachineBasicBlock::iterator II,
...
+    }
+    return false;
+  }
+  case AIE2P::VLDA_L_SPILL:
+  case AIE2P::VST_L_SPILL:
+  case AIE2P::LDA_D_SPILL:
+  case AIE2P::ST_D_SPILL:
+  case AIE2P::LDA_DS_SPILL:
+  case AIE2P::ST_DS_SPILL: {
+    if (isEncodableAsNegativeInt<9, 4>(Offset)) {
```
gbossu, line 188: Nit: maybe leave a comment that if `Offset` for those pseudo instructions is valid, then all of the "derived" offsets after expanding into `LDA_dms_lda_spill`/`ST_dms_sts_spill` will be valid as well.

llvm/lib/Target/AIE/aie2p/AIE2PFrameLowering.cpp
gbossu, line 107: Could you you maybe still leave a comment in the code about the `!isInt<12>(MFI.estimateStackSize(MF))`? I'm not sure I understand what it's for.

llvm/lib/Target/AIE/aie2p/AIE2PInstrInfo.cpp
```diff
@@ -1119,24 +1117,6 @@ bool AIE2PInstrInfo::expandPostRAPseudo(MachineInstr &MI) const {
     MI.eraseFromParent();
     return true;
   }
```
gbossu, line 1125: Nit leave a note to say that some pseudos like "spill/reload" pseudos are already expanded in `eliminateFrameIndex`.

llvm/lib/Target/AIE/aie2p/AIE2PRegisterInfo.cpp
```diff
@@ -202,6 +216,7 @@ bool AIE2PRegisterInfo::eliminateFrameIndex(MachineBasicBlock::iterator II,
   case AIE2P::ST_DS_SPILL: {
     if (isEncodableAsNegativeInt<9, 4>(Offset)) {
       MI.getOperand(FIOperandNum).ChangeToImmediate(Offset);
+      TII->expandSpillPseudo(MI, TRI, /*SubRegOffsetAlign=*/Align(4));
```
gbossu, line 190: Curious: Could that somehow invalidate iterators that the caller of `eliminateFrameIndex()` uses to go over the instructions?

gbossu (commented):
I think this looks good. I left a couple of comments with questions. Would be happy to get some answers there to make sure I understand the PR correctly :)

llvm/lib/Target/AIE/aie2p/AIE2PFrameLowering.cpp
niwinanto: Checking for stack size greater than `isInt<12>`, then there is a possibility of spill using instructions which cannot encode its offset value. So conservatively, allocating the emergency spill slot. Added a comment.

llvm/lib/Target/AIE/aie2p/AIE2PRegisterInfo.cpp
niwinanto: Its a valid point. We should be returning `true` value for invalid pointers. We were fortunate there was a `if (!Op.isFI()) continue;` check, only costing us couple more operand iteration of the deleted instruction. I have fixed it.

llvm/lib/Target/AIE/aie2p/AIE2PFrameLowering.cpp
gbossu: Makes sense, I'd just write that as a comment in the code :)

gbossu (approved):
LGTM

PR #274, opened 2025-01-16
[AIE2P] Legalization and instruction selection support for G_FADD/G_FSUB

katerynamuts:

PR #302, opened 2025-01-23
[AIE2P] Combine G_SHUFFLE_VECTOR to G_AIE_EXTRACT_SUBVECTOR

katerynamuts:


llvm/lib/Target/AIE/AIECombinerHelper.cpp
```diff
@@ -1761,8 +1761,13 @@ bool llvm::matchShuffleToBroadcast(MachineInstr &MI, MachineRegisterInfo &MRI,
   if (Src1Ty.getSizeInBits() != 64 && Src1Ty.getSizeInBits() != 32) {
     return false;
   }
+
   const unsigned NumDstElems = DstTy.getNumElements();
   const unsigned NumSrcElems = Src1Ty.getNumElements();
+  if (NumDstElems < NumSrcElems) {
```
andcarminati, line 1761: I guess that this could prevent a broadcast of an element of a bigger vector into a smaller one, no?

llvm/test/CodeGen/AIE/GlobalISel/prelegalizercombiner-shuffle-vector.mir
```diff
@@ -227,3 +227,129 @@ body:             |
     $x0 = COPY %0(<16 x s32>)
     PseudoRET implicit $lr, implicit $x0
 ...
+---
```
andcarminati: What about this case:

```
---
name: shuffle_vector_to_extract_subvec_4x8Dst
tracksRegLiveness: true
body:             |
  bb.1:
    liveins: $l0, $l1
    %1:_(<8 x s8>) = COPY $l0
    %2:_(<8 x s8>) = COPY $l1
    %0:_(<4 x s8>) = G_SHUFFLE_VECTOR %1(<8 x s8>), %2(<8 x s8>), shufflemask(8, 9, 10, 11)
```

Now we just discard, but we could return the first subreg of the second src register.

What do you think?

llvm/lib/Target/AIE/AIECombinerHelper.cpp
```diff
@@ -1826,3 +1831,77 @@ bool llvm::matchShuffleToVSel(
...
+/// %1:_(<16 x s16>) = COPY $wl0
+/// %2:_(s32) = G_CONSTANT i32 1
+/// %3:_(<4 x s16>) = G_AIE_EXTRACT_SUBVECTOR %1(<16 x s16>), %2(s32)
+bool llvm::matchShuffleToExtractSubvec(MachineInstr &MI,
+                                       MachineRegisterInfo &MRI,
+                                       const AIEBaseInstrInfo &TII,
+                                       BuildFnTy &MatchInfo) {
+  assert(MI.getOpcode() == TargetOpcode::G_SHUFFLE_VECTOR);
+  const Register DstReg = MI.getOperand(0).getReg();
+  const Register Src1Reg = MI.getOperand(1).getReg();
```
andcarminati, line 1929: I see a nice opportunity here by considering extracts from the second input operand.

llvm/lib/Target/AIE/AIECombinerHelper.cpp
katerynamuts: yes, but as far as I can see this combine `matchShuffleToBroadcast` doesn't cover this case.  And without this check It  fails for the following test:
`%0:_(<4 x s8>) = G_SHUFFLE_VECTOR %1(<8 x s8>), %2(<8 x s8>), shufflemask(-1, 5, -1, 3)`
because `createDuplicatePatternMask` returns a mask of size `NumDstElems / NumSrcElems=0`. 

In general, this PR is just to combine `G_SHUFFLE_VECTOR ` to `G_AIE_EXTRACT_SUBVECTOR` but there will be a follow-up PR which will extract a subvector and broadcast it and I think I can try to cover this case (`matchShuffleToBroadcast`) there.

llvm/test/CodeGen/AIE/GlobalISel/prelegalizercombiner-shuffle-vector.mir
katerynamuts: I also thought about it but @konstantinschwarz said there is a PR in upstream for canonicalization, i.e., if we have a mask with all indices from the second source vector, then the source vectors are switched and the indices are changed so that they correspond to the new first source vector (previously, source vector 2). It means that we implement all combines only for the source vector 1.

llvm/lib/Target/AIE/AIEInstrGISel.td
```diff
@@ -126,3 +126,10 @@ def G_AIE_VSHIFT_RIGHT : AIEGenericInstruction {
   let InOperandList = (ins type0:$src1, type0:$src2, type1:$shift_amt);
   let hasSideEffects = false;
 }
+
+// Extract 32-bit or 64-bit subvector.
+def G_AIE_EXTRACT_SUBVECTOR : AIEGenericInstruction {
```
niwinanto, line 131: Since we are only extracting 32/64-bit, is it really subvector extract? Can we use `G_AIE_SEXT_EXTRACT_VECTOR_ELT`?

llvm/lib/Target/AIE/AIECombinerHelper.cpp
```diff
@@ -1885,3 +1890,77 @@ bool llvm::matchShuffleToExtractBroadcast(MachineInstr &MI,
...
+
+  // Unlikely to select into a subregister copy
+  if (NumSrc1Elems % NumDstElems != 0)
+    return false;
+
+  auto CheckExtractMask = [=](unsigned Start, unsigned NumElems) -> bool {
+    auto ExtractMask = createSequentialMask(Start, NumElems, 0);
+
+    for (unsigned I = 0; I < NumDstElems; I++) {
+      if (Mask[I] == -1)
```
andcarminati: nit: comment about ignoring undefs.

llvm/lib/Target/AIE/AIECombinerHelper.cpp
```diff
@@ -1885,3 +1890,77 @@ bool llvm::matchShuffleToExtractBroadcast(MachineInstr &MI,
...
+        continue;
+
+      if (Mask[I] != ExtractMask[I])
```
andcarminati: What about a more compact implementation:

```
  auto CheckExtractMask = [&](unsigned Start, unsigned NumElems) -> bool {
    auto ExtractMask = createSequentialMask(Start, NumElems, 0);

    return std::equal(Mask.begin(), Mask.end(), ExtractMask.begin(),
                      ExtractMask.end(), [&](const int LHS, const int RHS) {
                        return (LHS == -1 || LHS == RHS);
                      });
  };
```

llvm/lib/Target/AIE/AIEInstrGISel.td
andcarminati: Maybe we could reuse and extend to include immediate usage. G_AIE_SEXT_EXTRACT_VECTOR_ELT also expands the index to a register, so as is, it is not interesting.

llvm/lib/Target/AIE/aie2p/AIE2PInstrPatterns.td
```diff
@@ -719,6 +719,28 @@ def : Pat<(int_aie2p_vinsert64_bf512 VEC512:$s1, eR29:$idx, eL:$s0),
 def : Pat<(int_aie2p_vinsert32_accfloat ACC512:$s1, eR29:$idx, eR:$s0),
            (COPY_TO_REGCLASS (VINSERT_32_mR29_insert (COPY_TO_REGCLASS ACC512:$s1, VEC512), eR29:$idx, eR:$s0), ACC512)>;
 
+// VEXTRACT
```
niwinanto, line 804: We are selecting like this `(<4 x s8>) = G_AIE_EXTRACT_SUBVECTOR [[COPY]](<8 x s8>), [[C]](s32)` in the prelegalizer. Output type differs from selection patterns below.

llvm/lib/Target/AIE/AIEInstrGISel.td
katerynamuts: The comment must be changed. It extracts 32/64 bit vectors: `4 x s8` or  `2 x s16`  for 32-bit and `8 x s8`, `4 x s16` and `2 x s32` for 64-bit. I discussed it already with @konstantinschwarz to extend G_AIE_SEXT_EXTRACT_VECTOR_ELT to cover also output vectors not only elements but we came to conclusion that it would be confusing when G_AIE_SEXT_EXTRACT_VECTOR_ELT extracts a vector and not an element, so we decided to introduce G_AIE_EXTRACT_SUBVECTOR  for extracting vectors.

katerynamuts:
The only native source vector type for `G_AIE_EXTRACT_SUBVECTOR` is 512-bit vectors. The output of `G_AIE_EXTRACT_SUBVECTOR` is always a 32- or 64-bit vector.  I updated the combiner method `matchShuffleToExtractSubvec`, so that it also supports the source vector types other than 512 bits. And I updated the instruction selection for `G_AIE_EXTRACT_SUBVECTOR`.

llvm/lib/Target/AIE/AIEInstrGISel.td
```diff
@@ -126,3 +126,10 @@ def G_AIE_VSHIFT_RIGHT : AIEGenericInstruction {
   let InOperandList = (ins type0:$src1, type0:$src2, type1:$shift_amt);
   let hasSideEffects = false;
 }
+
+// Extract 32-bit or 64-bit subvector.
```
niwinanto: Can we include 16-bit as well, which can extract <2 x 8>?

llvm/lib/Target/AIE/AIECombinerHelper.cpp
```diff
@@ -1885,3 +1890,151 @@ bool llvm::matchShuffleToExtractBroadcast(MachineInstr &MI,
...
+                          ImplicitDef, ImplicitDef, ImplicitDef, ImplicitDef});
+      } else if (Src1TySize == 64) {
+        B.buildConcatVectors(
+            {NewSrcReg}, {Src1Reg, ImplicitDef, ImplicitDef, ImplicitDef,
+                          ImplicitDef, ImplicitDef, ImplicitDef, ImplicitDef});
+      } else if (Src1TySize == 128) {
+        B.buildConcatVectors({NewSrcReg},
+                             {Src1Reg, ImplicitDef, ImplicitDef, ImplicitDef});
+      } else { // Src1TySize == 256
+        B.buildConcatVectors({NewSrcReg}, {Src1Reg, ImplicitDef});
```
niwinanto: Can we do better, to avoid code duplication. May be use for loop to generate concat vector Ops.

llvm/lib/Target/AIE/AIECombinerHelper.cpp
```diff
@@ -1885,3 +1890,151 @@ bool llvm::matchShuffleToExtractBroadcast(MachineInstr &MI,
...
+      if (SubIdx < (int)NumSubVectors / 2) {
+        B.buildUnmerge({NewSrcReg, UnusedSubReg1}, Src1Reg);
+
+      } else {
+        B.buildUnmerge({UnusedSubReg1, NewSrcReg}, Src1Reg);
+      }
+    }
+
+    if (Src1TySize == 2048) {
+      NewSubIdx = SubIdx % (NumSubVectors / 4);
```
niwinanto: Nit: May be using separate variable for `NumSubVectors / 4` increase readability.

llvm/lib/Target/AIE/AIECombinerHelper.cpp
```diff
@@ -1885,3 +1890,151 @@ bool llvm::matchShuffleToExtractBroadcast(MachineInstr &MI,
...
+
+  // 32, 64, 126, 256, 512, 1024, or 2048 bit source vectors are converted to
+  // 512 bit vectors
+  const unsigned Src1ElmtSize = Src1Ty.getElementType().getSizeInBits();
+  const unsigned Src1Vec512BitLen = 512 / Src1ElmtSize;
+  const LLT NewSrc1Ty = LLT::fixed_vector(Src1Vec512BitLen, Src1ElmtSize);
+  const Register NewSrcReg = MRI.createGenericVirtualRegister(NewSrc1Ty);
+
+  // 32, 64, 126, 256 bit source vectors
+  if (Src1TySize < 512) {
```
niwinanto: Why we need to support 32 and 64 source vector?

llvm/test/CodeGen/AIE/GlobalISel/prelegalizercombiner-shuffle-vector.mir
```diff
@@ -287,3 +287,255 @@ body:             |
...
+    ; CHECK-NEXT: {{  $}}
+    ; CHECK-NEXT: [[COPY:%[0-9]+]]:_(<4 x s8>) = COPY $r0
+    ; CHECK-NEXT: [[DEF:%[0-9]+]]:_(<4 x s8>) = G_IMPLICIT_DEF
+    ; CHECK-NEXT: [[CONCAT_VECTORS:%[0-9]+]]:_(<64 x s8>) = G_CONCAT_VECTORS [[COPY]](<4 x s8>), [[DEF]](<4 x s8>), [[DEF]](<4 x s8>), [[DEF]](<4 x s8>), [[DEF]](<4 x s8>), [[DEF]](<4 x s8>), [[DEF]](<4 x s8>), [[DEF]](<4 x s8>), [[DEF]](<4 x s8>), [[DEF]](<4 x s8>), [[DEF]](<4 x s8>), [[DEF]](<4 x s8>), [[DEF]](<4 x s8>), [[DEF]](<4 x s8>), [[DEF]](<4 x s8>), [[DEF]](<4 x s8>)
+    ; CHECK-NEXT: [[C:%[0-9]+]]:_(s32) = G_CONSTANT i32 0
+    ; CHECK-NEXT: [[AIE_EXTRACT_SUBVECTOR:%[0-9]+]]:_(<4 x s8>) = G_AIE_EXTRACT_SUBVECTOR [[CONCAT_VECTORS]](<64 x s8>), [[C]](s32)
+    ; CHECK-NEXT: PseudoRET implicit $lr, implicit [[AIE_EXTRACT_SUBVECTOR]](<4 x s8>)
+    %1:_(<4 x s8>) = COPY $r0
+    %2:_(<4 x s8>) = COPY $r1
+    %0:_(<4 x s8>) = G_SHUFFLE_VECTOR %1(<4 x s8>), %2(<4 x s8>), shufflemask(0, 1, 2, 3)
```
niwinanto: It seems `%0:_(<4 x s8>) = COPY %1` is more optimal.
Not sure this is a phase ordering problem or we are not supporting already.

llvm/lib/Target/AIE/AIEInstrGISel.td
katerynamuts: I discussed it with @konstantinschwarz and we decided to introduce  `G_AIE_EXTRACT_SUBVECTOR` to cover, e.g., the following case: `shufflemask <0,1,0,1,0,1> -> bcst (extract (src <0,1>))`.  And the input for `bcst` is 32 or 64 bit register, so  at least for now we allow `G_AIE_EXTRACT_SUBVECTOR` to extract 32 and 64 bit vectors because this combine is currently the only place where we use `G_AIE_EXTRACT_SUBVECTOR`.

llvm/lib/Target/AIE/AIECombinerHelper.cpp
katerynamuts: Why not?

llvm/test/CodeGen/AIE/GlobalISel/prelegalizercombiner-shuffle-vector.mir
katerynamuts: It is a good point but we don't support this combine (G_SHUFFLE_VECTOR->COPY) yet.

llvm/lib/Target/AIE/AIECombinerHelper.cpp
niwinanto: The minimum extract subvector size is 32, so what does extract 32 from 32-bit vector means.

katerynamuts: This covers this case: ` %0:_(<4 x s8>) = G_SHUFFLE_VECTOR %1(<4 x s8>), %2(<4 x s8>), shufflemask(0, 1, 2, 3)`

llvm/test/CodeGen/AIE/GlobalISel/prelegalizercombiner-shuffle-vector.mir
katerynamuts: But maybe I can cover this case in my function and then this will be more optimal and I won't need support for 32 bit vectors

katerynamuts: I added this case to my implementation

llvm/lib/Target/AIE/AIECombinerHelper.cpp
```diff
@@ -1885,3 +1890,144 @@ bool llvm::matchShuffleToExtractBroadcast(MachineInstr &MI,
...
+      NewSubIdx = SubIdx % NumSubVectors512Bits;
+      if (SubIdx < (int)NumSubVectors512Bits) {
+        B.buildUnmerge({NewSrcReg, UnusedSubReg1}, Src1Reg);
+      } else {
+        B.buildUnmerge({UnusedSubReg1, NewSrcReg}, Src1Reg);
+      }
+    }
+
+    if (Src1TySize == 2048) {
+      const unsigned NumSubVectors512Bits = NumSubVectors / 4;
```
andcarminati: I think this code can be simplified. Just one unmerge in the end.

llvm/lib/Target/AIE/AIEBaseInstrInfo.h
```diff
@@ -145,6 +145,10 @@ struct AIEBaseInstrInfo : public TargetInstrInfo {
   virtual unsigned getGenericVShiftOpcode() const {
     llvm_unreachable("Target didn't implement getGenericVShiftOpcode!");
   }
+  /// Return the opcode to be used for subvector extraction.
+  virtual unsigned getGenericExtractSubvectorOpcode() const {
+    llvm_unreachable("Target didn't implement getGenericVSelOpcode!");
```
martien-de-jong: nit: incorrect function name in the message

llvm/lib/Target/AIE/AIECombinerHelper.cpp
```diff
@@ -1885,3 +1890,144 @@ bool llvm::matchShuffleToExtractBroadcast(MachineInstr &MI,
   };
   return true;
 }
+
+/// Match something like this:
+///  %1:_(<16 x s16>) = COPY $wl0
+///  %2:_(<16 x s16>) = COPY $wl1
+///  %0:_(<4 x s16>) = G_SHUFFLE_VECTOR %1(<16 x s16>), %2(<16 x s16>),
+///  shufflemask(4, 5, 6, 7)
```
martien-de-jong, line 1911: Copy paste of 16 x s16 here? I expect 4 x s16

llvm/lib/Target/AIE/AIECombinerHelper.cpp
```diff
@@ -1885,3 +1890,144 @@ bool llvm::matchShuffleToExtractBroadcast(MachineInstr &MI,
...
+      SubIdx = SubVecIdx;
+      break;
+    }
+  }
+
+  // Not an extract pattern
+  if (SubIdx == -1)
+    return false;
+
+  if (Src1TySize == DstTy.getSizeInBits()) {
```
niwinanto: Even though its a simple case, I think it needs to be a separate combine with more tests. Can be even generic combine that we could upstream.

llvm/test/CodeGen/AIE/GlobalISel/prelegalizercombiner-shuffle-vector.mir
```diff
@@ -287,3 +287,265 @@ body:             |
...
+tracksRegLiveness: true
+body:             |
+  bb.1:
+    liveins: $l0, $l1
+    ; CHECK-LABEL: name: shuffle_vector_to_extract_subvec_64BitSrc_32BitDst
+    ; CHECK: liveins: $l0, $l1
+    ; CHECK-NEXT: {{  $}}
+    ; CHECK-NEXT: [[COPY:%[0-9]+]]:_(<4 x s16>) = COPY $l0
+    ; CHECK-NEXT: [[DEF:%[0-9]+]]:_(<4 x s16>) = G_IMPLICIT_DEF
+    ; CHECK-NEXT: [[CONCAT_VECTORS:%[0-9]+]]:_(<32 x s16>) = G_CONCAT_VECTORS [[COPY]](<4 x s16>), [[DEF]](<4 x s16>), [[DEF]](<4 x s16>), [[DEF]](<4 x s16>), [[DEF]](<4 x s16>), [[DEF]](<4 x s16>), [[DEF]](<4 x s16>), [[DEF]](<4 x s16>)
```
niwinanto: Not sure we handle `G_CONCAT_VECTORS` for this types.

llvm/test/CodeGen/AIE/GlobalISel/prelegalizercombiner-shuffle-vector.mir
```diff
@@ -287,3 +287,265 @@ body:             |
...
+name: shuffle_vector_to_extract_subvec_2048BitSrc
+tracksRegLiveness: true
+body:             |
+  bb.1:
+    liveins: $dm0, $dm1
+    ; CHECK-LABEL: name: shuffle_vector_to_extract_subvec_2048BitSrc
+    ; CHECK: liveins: $dm0, $dm1
+    ; CHECK-NEXT: {{  $}}
+    ; CHECK-NEXT: [[COPY:%[0-9]+]]:_(<64 x s32>) = COPY $dm0
+    ; CHECK-NEXT: [[UV:%[0-9]+]]:_(<16 x s32>), [[UV1:%[0-9]+]]:_(<16 x s32>), [[UV2:%[0-9]+]]:_(<16 x s32>), [[UV3:%[0-9]+]]:_(<16 x s32>) = G_UNMERGE_VALUES [[COPY]](<64 x s32>)
```
niwinanto, line 549: As far as I can see `G_UNMERGE_VALUES` for this not legal.

llvm/lib/Target/AIE/AIECombinerHelper.cpp
martien-de-jong: Isn't that equivalent to `%0 = COPY %1`?

llvm/lib/Target/AIE/AIECombinerHelper.cpp
```diff
@@ -1885,3 +1890,144 @@ bool llvm::matchShuffleToExtractBroadcast(MachineInstr &MI,
...
+  const Register Src1Reg = MI.getOperand(1).getReg();
+  const ArrayRef<int> Mask = MI.getOperand(3).getShuffleMask();
+
+  const LLT DstTy = MRI.getType(DstReg);
+  const LLT Src1Ty = MRI.getType(Src1Reg);
+  const unsigned Src1TySize = Src1Ty.getSizeInBits();
+
+  if (!DstTy.isVector() || !Src1Ty.isVector() || Src1TySize < 32 ||
+      (DstTy.getSizeInBits() != 32 && DstTy.getSizeInBits() != 64))
+    return false;
```
martien-de-jong, line 1949: Could we parameterize this function with all the sizes that are explicit constants here? I guess 32 is BitSize, 64 is 2 x BitSize, 512 is SomeFactor * BitSize, etc.
Since this is in a generic helper, it would make sense to cut the function into smaller, more useful methods.

llvm/test/CodeGen/AIE/GlobalISel/prelegalizercombiner-shuffle-vector.mir
andcarminati: It looks we cannot legalize. With 4 inputs we can custom legalize.

llvm/lib/Target/AIE/AIECombinerHelper.cpp
```diff
@@ -1885,3 +1890,144 @@ bool llvm::matchShuffleToExtractBroadcast(MachineInstr &MI,
...
+  // Not an extract pattern
+  if (NumSrc1Elems < NumDstElems)
+    return false;
+
+  // Unlikely to select into a subregister copy
+  if (NumSrc1Elems % NumDstElems != 0)
+    return false;
+
+  auto CheckExtractMask = [=](unsigned Start, unsigned NumElems) -> bool {
+    auto ExtractMask = createSequentialMask(Start, NumElems, 0);
```
martien-de-jong: I would just write this as a for-loop reading Mask, returning false as soon as an index doesn't match. And perhaps lift it to a top level static method checkSequentialMask(Mask, StartPos, NumElems, StartIdx).

llvm/lib/Target/AIE/AIECombinerHelper.cpp
```diff
@@ -1885,3 +1890,144 @@ bool llvm::matchShuffleToExtractBroadcast(MachineInstr &MI,
...
+  for (unsigned SubVecIdx = 0; SubVecIdx < NumSubVectors; SubVecIdx++) {
+    if (CheckExtractMask(SubVecIdx * NumDstElems, NumDstElems)) {
+      SubIdx = SubVecIdx;
+      break;
+    }
+  }
+
+  // Not an extract pattern
+  if (SubIdx == -1)
+    return false;
```
martien-de-jong, line 1977: This would read good as a lambda, returning SubVecIdx from the loop and -1 at the end.

```
const int SubIdx = getSubVectorIndx(...);
if (SubIdx < 0) {
  return false;
}
```

llvm/lib/Target/AIE/AIECombinerHelper.cpp
```diff
@@ -1885,3 +1890,144 @@ bool llvm::matchShuffleToExtractBroadcast(MachineInstr &MI,
...
+      auto Cst = B.buildConstant(LLT::scalar(32), SubIdx);
+      B.buildInstr(Opc, {DstReg}, {NewSrcReg, Cst});
+    };
+    return true;
+  }
+
+  // 1024 and 2048 bit source vectors
+  const Register UnusedSubReg1 = MRI.createGenericVirtualRegister(NewSrc1Ty);
+  const Register UnusedSubReg2 = MRI.createGenericVirtualRegister(NewSrc1Ty);
+  const Register UnusedSubReg3 = MRI.createGenericVirtualRegister(NewSrc1Ty);
```
martien-de-jong: Could we bring the Src1TypeSize test outside, have two different MatchInfo lambdas and only create the VRs that we need? Note that we may create unused VRs now.

llvm/lib/Target/AIE/AIECombinerHelper.cpp
martien-de-jong: Yes, put the UnusedSubregs in an array and build the operand vector from it.

llvm/test/CodeGen/AIE/GlobalISel/prelegalizercombiner-shuffle-vector.mir
andcarminati: But for this type, we need a different solution:

```
llvm::AIELegalizerHelper::legalizeG_CONCAT_VECTORS(llvm::LegalizerHelper &, llvm::MachineInstr &) const: Assertion `SrcTy.getSizeInBits() >= 256 && "Input vector size does not match!"' failed.
```

niwinanto: I think, subregister copy is enough for the 64 bit source. No need to build a 512-bit register and extract 32-bit element from it.

llvm/test/CodeGen/AIE/GlobalISel/prelegalizercombiner-shuffle-vector.mir
```diff
@@ -287,3 +287,265 @@ body:             |
...
+    ; CHECK-NEXT: [[C:%[0-9]+]]:_(s32) = G_CONSTANT i32 0
+    ; CHECK-NEXT: [[AIE_EXTRACT_SUBVECTOR:%[0-9]+]]:_(<2 x s32>) = G_AIE_EXTRACT_SUBVECTOR [[UV]](<16 x s32>), [[C]](s32)
```
andcarminati, line 551: We have:
```
LLVM ERROR: unable to legalize instruction: %3:_(<16 x s32>), %4:_(<16 x s32>), %5:_(<16 x s32>), %6:_(<16 x s32>) = G_UNMERGE_VALUES %0:_(<64 x s32>) (in function: shuffle_vector_to_extract_subvec_2048BitSrc)
```

For this, we need two unmerges:

1024, 1024 = G_UNMERGE_VALUES 2048
512, 512 = G_UNMERGE_VALUES 1024

And then take the correct 512 to extract.

llvm/lib/Target/AIE/AIECombinerHelper.cpp
katerynamuts: yes, I've added this already to my implementation. See this test: https://github.com/Xilinx/llvm-aie/pull/302/files#:~:text=%2D%2D%2D-,name%3A%20shuffle_vector_to_extract_subvec_32BitSrc,-tracksRegLiveness%3A%20true

llvm/lib/Target/AIE/AIECombinerHelper.cpp
katerynamuts: ok, I'll just add a check and do nothing in my combine for this case.

llvm/test/CodeGen/AIE/GlobalISel/prelegalizercombiner-shuffle-vector.mir
katerynamuts: I discussed it with @konstantinschwarz, for now we don't need support for so small source vectors, so I just put an assertion for these cases and cover only the source vectors sizes greater than or equal to 256 bits.

llvm/test/CodeGen/AIE/GlobalISel/prelegalizercombiner-shuffle-vector.mir
katerynamuts: yes, it is illegal, I added the legalization for it but the legalization is also a part of this PR https://github.com/Xilinx/llvm-aie/pull/274

llvm/test/CodeGen/AIE/GlobalISel/prelegalizercombiner-shuffle-vector.mir
katerynamuts: I added the legalization for it to this PR but the legalization is also a part of this PR https://github.com/Xilinx/llvm-aie/pull/274

katerynamuts:
I added the legalization part for `G_UNMERGE_VALUES`, For now we don't need support for the source vector sizes less than 256 , so I just put an assertion for these cases and cover only the source vectors sizes greater than or equal to 256 bits, i.e., that we don't need to extend `G_CONCAT_VECTORS`. I also ran the tests with `-start-before=aie2p-prelegalizer-combiner` without `-stop-after` and there are no errors.

llvm/test/CodeGen/AIE/GlobalISel/prelegalizercombiner-shuffle-vector.mir
```diff
@@ -373,14 +373,15 @@ body:             |
...
     ; CHECK-NEXT: {{  $}}
-    ; CHECK-NEXT: [[COPY:%[0-9]+]]:_(<4 x s64>) = COPY $wl0
-    ; CHECK-NEXT: [[C:%[0-9]+]]:_(s32) = G_CONSTANT i32 3
-    ; CHECK-NEXT: [[AIE_SEXT_EXTRACT_VECTOR_ELT:%[0-9]+]]:_(s32) = G_AIE_SEXT_EXTRACT_VECTOR_ELT [[COPY]](<4 x s64>), [[C]](s32)
-    ; CHECK-NEXT: [[AIE_BROADCAST_VECTOR:%[0-9]+]]:_(<8 x s64>) = G_AIE_BROADCAST_VECTOR [[AIE_SEXT_EXTRACT_VECTOR_ELT]](s32)
-    ; CHECK-NEXT: PseudoRET implicit $lr, implicit [[AIE_BROADCAST_VECTOR]](<8 x s64>)
-    %0:_(<4 x s64>) = COPY $wl0
-    %1:_(<4 x s64>) = COPY $wl1
-    %2:_(<8 x s64>) = G_SHUFFLE_VECTOR %0(<4 x s64>), %1(<4 x s64>), shufflemask(3, 3, -1, -1, 3, 3, 3, 3)
+    ; CHECK-NEXT: [[COPY:%[0-9]+]]:_(<8 x s32>) = COPY $wl0
```
andcarminati, line 376: This delivers a nice piece of code:

```
	nopa	;		nopb	;		nops	;		ret	lr;		nopm	;		nopv	
	mova	r0, #2                          //  Delay Slot 5
	vextract.32	 r0, x0, r0, vaddsign1  //  Delay Slot 4
	nop	                                //  Delay Slot 3
	vbcst.32	 x0, r0                         //  Delay Slot 2
```

llvm/lib/Target/AIE/AIECombinerHelper.cpp
```diff
@@ -1881,3 +1885,154 @@ bool llvm::matchShuffleToExtractBroadcast(MachineInstr &MI,
...
+
+  // Not an extract pattern
+  if (NumSrc1Elems < NumDstElems)
+    return false;
+
+  // Unlikely to select into a subregister copy
+  if (NumSrc1Elems % NumDstElems != 0)
+    return false;
+
+  auto GetSubIdx = [=]() -> std::optional<int> {
```
andcarminati: nit: you can capture by reference here to avoid a mask copy.

llvm/lib/Target/AIE/AIECombinerHelper.cpp
```diff
@@ -1881,3 +1885,154 @@ bool llvm::matchShuffleToExtractBroadcast(MachineInstr &MI,
...
+bool llvm::matchShuffleToExtractSubvec(MachineInstr &MI,
+                                       MachineRegisterInfo &MRI,
+                                       const AIEBaseInstrInfo &TII,
+                                       BuildFnTy &MatchInfo) {
+  // To parametrize this function we define some generic constants. Ideally,
+  // these constants should be defined in TII with the corresponding methods,
+  // e.g., TII->getGPRSize().
+  // The current values are valid for AIE2P.
+  const unsigned GPRSize = 32;
+  const unsigned VecRegSize = 256;
```
andcarminati, line 1926: Should those parameters be tied to TII? We already query for the opcode.

llvm/lib/Target/AIE/AIECombinerHelper.cpp
```diff
@@ -1881,3 +1885,154 @@ bool llvm::matchShuffleToExtractBroadcast(MachineInstr &MI,
...
+
+/// To convert to:
+/// %1:_(<16 x s16>) = COPY $wl0
+/// %2:_(s32) = G_CONSTANT i32 1
+/// %3:_(<4 x s16>) = G_AIE_EXTRACT_SUBVECTOR %1(<16 x s16>), %2(s32)
+bool llvm::matchShuffleToExtractSubvec(MachineInstr &MI,
+                                       MachineRegisterInfo &MRI,
+                                       const AIEBaseInstrInfo &TII,
+                                       BuildFnTy &MatchInfo) {
+  // To parametrize this function we define some generic constants. Ideally,
```
andcarminati: I think we should go to this generic approach. This is the only non-generic part of the code.

llvm/lib/Target/AIE/AIECombinerHelper.cpp
andcarminati: super-nit: As we use optional, we can use unsigned here.

llvm/lib/Target/AIE/AIECombinerHelper.cpp
```diff
@@ -1881,3 +1885,154 @@ bool llvm::matchShuffleToExtractBroadcast(MachineInstr &MI,
...
+      B.buildConcatVectors({NewSrcReg}, ConcatOps);
+      auto Cst = B.buildConstant(LLT::scalar(GPRSize), SubIdx.value());
+      B.buildInstr(Opc, {DstReg}, {NewSrcReg, Cst});
+    };
+    return true;
+  }
+
+  // Source vectors with the size greater than the native source vector size
+  MatchInfo = [=, &MRI](MachineIRBuilder &B) {
+    int NewSubIdx = -1;
```
andcarminati: nit: declare and initialize in the same line below, perhaps using `unsigned`.

llvm/lib/Target/AIE/AIECombinerHelper.cpp
```diff
@@ -1881,3 +1885,154 @@ bool llvm::matchShuffleToExtractBroadcast(MachineInstr &MI,
...
+    const unsigned SizeCoefficient = Src1TySize / ExtractSubvecNativeSrcSize;
+    const unsigned NumSubVectorsNativeSize = NumSubVectors / SizeCoefficient;
+    NewSubIdx = SubIdx.value() % NumSubVectorsNativeSize;
+
+    SmallVector<Register, 4> SubRegs;
+    for (unsigned I = 0; I < SizeCoefficient - 1; ++I) {
```
andcarminati: nit: Instead of shifting the vector to insert, we can insert in place:
```
    unsigned NewSrcRegPosition =  SubIdx.value() / NumSubVectorsNativeSize;
    for (unsigned I = 0; I < SizeCoefficient; ++I) {
      if (I == NewSrcRegPosition)
        SubRegs.push_back(NewSrcReg);
      else 
        SubRegs.push_back(MRI.createGenericVirtualRegister(NewSrc1Ty));
    }
```

llvm/lib/Target/AIE/aie2p/AIE2PLegalizerInfo.cpp
```diff
@@ -524,22 +524,31 @@ AIE2PLegalizerInfo::AIE2PLegalizerInfo(const AIE2PSubtarget &ST)
...
             // Unmerge of sources <= 32-bit are legalized into bit arithmetic
             // below
             if (SrcTy.getSizeInBits() <= 32)
               return false;
 
-            // Unmerge results that are not vector register sized are legalized
-            // to EXTRACT_VECTOR_ELT
-            if (!isVectorRegisterSized(DstTy))
-              return false;
+            if (SrcTy.isVector() && DstTy.isScalar()) {
```
andcarminati, line 532: Check: this is part of other PR, but it is here now for completeness, right?

katerynamuts: The part inside this `if` was already there, I just add this `if` to extend `fewerElementsIf` for this case: `SrcTy.isVector() && DstTy.isVector()`.

andcarminati: In case, it would be nice to have a test for this change.

llvm/lib/Target/AIE/aie2p/AIE2PLegalizerInfo.cpp
```diff
@@ -524,22 +524,31 @@ AIE2PLegalizerInfo::AIE2PLegalizerInfo(const AIE2PSubtarget &ST)
...
+              // legalized to EXTRACT_VECTOR_ELT
+              if (!isVectorRegisterSized(DstTy))
+                return false;
+
+              // Split unmerges with more than two scalar results into multiple
+              // unmerges with two results
+              return SrcTy.getNumElements() > 2;
+            }
+
+            if (SrcTy.isVector() && DstTy.isVector()) {
```
andcarminati, line 543: nit:
```
            if (SrcTy.isVector() && DstTy.isVector()) {
              // Unmerges into 2 subvectors are legal
              return !(2 * DstTy.getSizeInBits() == SrcTy.getSizeInBits());
            }
```

llvm/lib/Target/AIE/AIE2InstrInfo.cpp
```diff
@@ -1620,3 +1620,9 @@ Register AIE2InstrInfo::getMSStatusReg() const { return AIE2::srMS0; }
 Register AIE2InstrInfo::getPackSignCReg() const { return AIE2::crPackSign; }
 
 Register AIE2InstrInfo::getUnpackSignCReg() const { return AIE2::crUnpackSign; }
+
+unsigned AIE2InstrInfo::getGPRSize() const { return 32; }
```
andcarminati: nit: maybe `getScalarRegSize`.

llvm/lib/Target/AIE/AIEBaseInstrInfo.h
```diff
@@ -145,6 +145,11 @@ struct AIEBaseInstrInfo : public TargetInstrInfo {
   virtual unsigned getGenericVShiftOpcode() const {
     llvm_unreachable("Target didn't implement getGenericVShiftOpcode!");
   }
+  /// Return the opcode to be used for subvector extraction.
+  virtual unsigned getGenericExtractSubvectorOpcode() const {
```
F-Stuckmann, line 149: I know this is probably the wrong place and time to note this, but it would make sense to have shared opcodes across all AIE targets and only specify the opcodes that differ. Therefore we would have all generic optimization by default enabled in every architecture we have, instead of manually adding support for each architecture by providing target specific hooks.

katerynamuts: absolutely agree, but it should be done in a separate PR. :)

llvm/lib/Target/AIE/AIECombinerHelper.cpp
```diff
@@ -1881,3 +1885,148 @@ bool llvm::matchShuffleToExtractBroadcast(MachineInstr &MI,
...
+  const unsigned Src1TySize = Src1Ty.getSizeInBits();
+
+  if (!DstTy.isVector() || !Src1Ty.isVector() || Src1TySize < GPRSize ||
+      (DstTy.getSizeInBits() != GPRSize &&
+       DstTy.getSizeInBits() != 2 * GPRSize))
+    return false;
+
+  // Currently, we cannot concat vectors of the size less than vector register
+  // size.
+  assert(Src1TySize >= VecRegSize && "Unsupported source register size");
```
katerynamuts: @andcarminati what do you think about this ? is it better to have an assert here or just return so that we don't crash if we have this case?

llvm/lib/Target/AIE/AIEBaseInstrInfo.h
andcarminati: I think that the name can be the same, but internally they can have different numbers. So we need `AIE2::` and `AIE2p::`.

llvm/lib/Target/AIE/AIECombinerHelper.cpp
andcarminati: As this is a match function, it is better to not crash! You are right.

llvm/lib/Target/AIE/AIEBaseInstrInfo.h
F-Stuckmann: yeah, but there should be a `AIEBASE::` opcode collection. 

Then all combiners etc. work on the `AIEBASE::` as  much as possible and only target specific Opcodes get own hooks (like fifo load/stores)

That's why this is not the point in time to do such a large overhaul, but it may help us in the future

andcarminati (approved):
LGTM. Nice work!

PR #305, opened 2025-01-23
Improvements in resource conflict detection related to slots

martien-de-jong:
This started out by the realization that XM is a separate slot from X and M, and that the only mechanism that prevents them from occupying the same bundle is the format table. 

ResourceCycle will make a quick check whether there's overlap in the slot bits, but will miss an X XM overlap.

The idea is that we can go through the complete format list and find the implications from XM to X and M (All formats containing XM don't allow X or M)

First, we have simplified the generated tables by removing the target-specific SlotInfo and SlotKind classes. That removes the need for MCFormats.h to include all generated format tables. This aligns with the tablegen idiom that we have universal data type whose data is supplied as an opaque bunch of untyped data.

Trying to use the denser 'ConflictBits' in the check pointed out a problem, which is the nop slot as used by AIE2 to disambiguate  the two stand-alone formats for XM, on in 48 bits and one in 64 bits. The latter one is only used for bundle padding. That nop slot only occurs in the 16 nop format and the 64 bit XM format, and hence is implied by A, B, S, X and M slots. 
We added the notion of an artificial slot to be able to exclude it from the conflict bits.

**Status**

- Clean-up of slot and format interfaces
- Conflict bits generated and used.
- NFC for  existing unit tests
- Use a O(1) order lookup for format availability
- Strip some unused data from the field description table in AIEMCFormats.h


llvm/utils/TableGen/CodeGenFormat.cpp
```diff
@@ -974,8 +974,6 @@ void TGFieldLayout::resolveFieldsGlobalOffsets() {
   GlobalOffsets.RightOffset = GlobalOffsets.LeftOffset + this->getSize() - 1;
 }
 
-unsigned TGTargetSlots::SlotID = 0;
```
martien-de-jong, line 978: nit copyright change here.

llvm/lib/Target/AIE/MCTargetDesc/AIEMCFormats.h
```diff
@@ -26,6 +26,8 @@
 
 namespace llvm {
 
+static const int SLOT_UNKNOWN = -1;
+
```
martien-de-jong, line 30: nit cpyrgt change here

llvm/utils/TableGen/CodeGenFormat.h
```diff
@@ -404,9 +404,8 @@ class TGTargetSlots {
         [&Label](const RecordSlot &RS) { return RS.second.SlotName == Label; });
   }
 
-  const_iterator cbegin() const { return Slots.begin(); }
-
-  const_iterator cend() const { return Slots.end(); }
+  const_iterator begin() const { return Slots.begin(); }
+  const_iterator end() const { return Slots.end(); }
```
martien-de-jong, line 437: There's probably a reason, e.g. being able to use range-based syntax.

llvm/lib/Target/AIE/MCTargetDesc/AIEMCFormats.h
```diff
@@ -75,21 +75,31 @@ class MCSlotInfo {
...
+      : SlotName(SlotName), Size(Size), SlotOccupancy(Bits), ConflictBits(Bits),
+        NopOpc(NopOpc) {}
 
   constexpr MCSlotInfo(int Kind, const char *SlotName, unsigned Size,
                        SlotBits Bits, unsigned NopOpc)
-      : SlotName(SlotName), Size(Size), SlotOccupancy(Bits), NopOpc(NopOpc) {}
+      : SlotName(SlotName), Size(Size), SlotOccupancy(Bits), ConflictBits(Bits),
+        NopOpc(NopOpc) {}
+  constexpr MCSlotInfo(int Kind, const char *SlotName, unsigned Size,
+                       SlotBits Bits, SlotBits Conflicts, unsigned NopOpc)
```
martien-de-jong, line 97: nit: it looks as we could lose Kind plus related constructors.

llvm/utils/TableGen/CodeGenFormat.cpp
```diff
@@ -220,7 +216,6 @@ void CodeGenFormat::run(raw_ostream &o) {
 
     for (TGInstrLayout &Inst : InstFormats)
       if (Inst.isPacketFormat()) {
-        Inst.computeSlotSet();
```
martien-de-jong, line 223: We've moved this out to a second pass, since we first should process all formats before we can compute the conflict bits.

llvm/utils/TableGen/CodeGenFormat.cpp
```diff
@@ -1171,16 +1171,15 @@ void TGTargetSlots::emitTargetSlotClass(raw_ostream &o) const {
 
   o << "class " << TargetClassName << " : public MC" << GenSlotInfoName << "\n"
     << "{\n"
-    << "  const " << TargetEnumName << " Kind;\n"
```
mludevid, line 1176: Please also remove the `const std::string TargetEnumName = ...`.

llvm/utils/TableGen/CodeGenFormat.cpp
```diff
@@ -631,15 +630,14 @@ void TGInstrLayout::emitFlatTree(ConstTable &FieldsHierarchy,
...
 
-    FieldsHierarchy << TargetKindName << '(' << TargetKindName << "::";
     if (SlotInfo)
-      FieldsHierarchy << SlotInfo->getEnumerationString();
+      FieldsHierarchy << "MCSlotKind(" << TargetKindName
+                      << "::" << SlotInfo->getEnumerationString() << ")";
     else
-      FieldsHierarchy << DefaultSlotInfo->second.getEnumerationString();
-    FieldsHierarchy << ") }";
+      FieldsHierarchy << "SLOT_UNKNOWN";
```
mludevid, line 694: Can we create a global `static const` for `"SLOT_UNKONWN"` as well? It is used in a few places...

llvm/utils/TableGen/CodeGenFormat.cpp
```diff
@@ -251,52 +249,91 @@ void CodeGenFormat::run(raw_ostream &o) {
...
+      if (Format.isPacketFormat()) {
+        uint64_t Slots = Format.getSlotSet();
+        if (Slots & ThisSlot) {
+          Excluded &= ~Slots & AllSlots;
+        }
+      }
+    }
+    Slot.setConflictBits(ThisSlot | Excluded);
+  }
+}
```
martien-de-jong, line 330: Ehh, this got lost in rebase. This is NOT a cosmetic detail.

llvm/utils/TableGen/CodeGenFormat.cpp
martien-de-jong: I think we have one in AIEMCFormats.h. This is its use.

llvm/lib/Target/AIE/AIEBundle.h
```diff
@@ -91,9 +91,15 @@ template <class I> class Bundle {
     // Verify there is a format that can accommodate the new slots
     MCSlotKind Slot = FormatInterface->getSlotKind(InstOpCode);
     assert(Slot != MCSlotKind());
-    SlotBits NewSlots = FormatInterface->getSlotInfo(Slot)->getSlotSet();
-    return (OccupiedSlots & NewSlots) == 0 &&
-           FormatInterface->getPacketFormats().getFormat(OccupiedSlots |
+
+    auto *SlotInfo = FormatInterface->getSlotInfo(Slot);
+    // ConflictBits is a fast predictor of missing formats
+    SlotBits ConflictBits = SlotInfo->getConflictSet();
```
martien-de-jong, line 97: Using ConflictBits here is easy. It's a pity that this is not the biggest hotspot. The more frequently used conflict interfaces in AIEHazardRecognizer are more of a problem.

llvm/utils/TableGen/CodeGenFormat.cpp
martien-de-jong: Actually, the entire function disappears later on.

llvm/lib/Target/AIE/AIEHazardRecognizer.cpp
```diff
@@ -125,6 +125,7 @@ FuncUnitWrapper &FuncUnitWrapper::operator|=(const FuncUnitWrapper &Other) {
 bool FuncUnitWrapper::conflict(const FuncUnitWrapper &Other) const {
   if ((Slots & Other.Slots) != 0 || (MemoryBanks & Other.MemoryBanks) != 0 ||
       (MemObjectsBits & Other.MemObjectsBits) != 0 ||
+      (Conflicts & Other.Slots) != 0 || (Slots & Other.Conflicts) != 0 ||
```
martien-de-jong, line 131: This was the easiest way. We misuse FuncUnitWrapper to hold the auxiliary data in Conflicts, which is a straight copy of the SlotInfo field. Conflicts isn't part of the scoreboard state, just of the incoming new FuncUnitWrapper.
-- well, it will get merged into the scoreboard after all, it keeps the merge operation associative.

llvm/lib/Target/AIE/AIEHazardRecognizer.cpp
```diff
@@ -482,6 +483,15 @@ static SlotBits getSlotSet(const MCInstrDesc &Desc,
   return IgnoreUnkownSlotSets ? 0 : ~0;
 }
 
+static SlotBits getConflictSet(const MCInstrDesc &Desc,
+                               const AIEBaseMCFormats &Formats) {
+  MCSlotKind SlotKind = Formats.getSlotKind(Desc.getOpcode());
+  if (SlotKind != MCSlotKind())
+    return Formats.getSlotInfo(SlotKind)->getConflictSet();
+
+  return 0;
```
martien-de-jong, line 495: If we can't retrieve conflict bits, we fall back to the existing slot test.

llvm/lib/Target/AIE/AIEHazardRecognizer.cpp
```diff
@@ -540,25 +551,26 @@ bool AIEHazardRecognizer::checkConflict(
...
   // Verify format hazards
-  FuncUnitWrapper EmissionCycle(SlotSet);
+  FuncUnitWrapper EmissionCycle(SlotSet, ConflictSet);
   if (EmissionCycle.conflict(Scoreboard[DeltaCycles]))
     return true;
 
   // Verify memory bank and shared object hazards
   if (!MemoryAccessCycles.empty()) {
-    FuncUnitWrapper MemoryAccessCycle(/*SlotSet=*/0, MemoryBanks,
+    FuncUnitWrapper MemoryAccessCycle(/*SlotSet=*/0, 0, MemoryBanks,
```
martien-de-jong, line 576: The danger of weakly typed default parameters. Without this change, it compiled, but all memory bank tests broke.

llvm/lib/Target/AIE/AIEHazardRecognizer.h
```diff
@@ -72,6 +72,9 @@ class FuncUnitWrapper {
 
   /// The occupied slots. This is currently redundant with Bundle
   SlotBits Slots = 0;
+  /// Conflicts are just for speeding up conflict detection. They may be present
+  /// in cycles to be merged, but they will not be merged into the scoreboard.
+  SlotBits Conflicts = 0;
```
martien-de-jong, line 77: TODO: I need to clear these bits. Moreover, we will merge the bits, since we can't know whether we are incrementally building a cycle to add to the scoreboard, or merging into the scoreboard -- both are done using `operator |`.

martien-de-jong:
Some of the cleanup commits have been put into a separate PR. The rest will resume in a new PR.

PR #306, opened 2025-01-23
[AIEX] AIEClusterBaseAddress: re-order G_PTR_ADD to maximise address chaining

gbossu:
Tiny improvement that re-orders G_PTR_ADD so that those with constant offsets are considered first. This enables more chaining. We used to break the chain at the first non-const G_PTR_ADD .

### DIFF
--- a/llvm/lib/Target/AIE/AIEClusterBaseAddress.cpp
+++ b/llvm/lib/Target/AIE/AIEClusterBaseAddress.cpp
@@ -47,7 +47,6 @@
 #include "AIE.h"
 #include "llvm/CodeGen/GlobalISel/CSEInfo.h"
 #include "llvm/CodeGen/GlobalISel/CSEMIRBuilder.h"
-#include "llvm/CodeGen/GlobalISel/GenericMachineInstrs.h"
 #include "llvm/CodeGen/GlobalISel/MachineIRBuilder.h"
 #include "llvm/CodeGen/MachineFunction.h"
 #include "llvm/CodeGen/MachineFunctionPass.h"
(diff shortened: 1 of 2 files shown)


llvm/lib/Target/AIE/AIEClusterBaseAddress.cpp
```diff
@@ -74,6 +73,61 @@ static cl::opt<bool> EnableChainsForVectorLdSt(
...
+/// PTR_ADD chains.
+bool optimisePostIncrements(Register BasePtr, ArrayRef<MachineInstr *> PtrAdds,
+                            const MachineRegisterInfo &MRI,
+                            MachineIRBuilder &MIB,
+                            GISelObserverWrapper &Observer) {
+  bool Changed = false;
+
+  // Look for the following sequence:
+  // %0 = G_PTR_ADD %100, %101
+  // %1 = G_PTR_ADD %100, 32
```
F-Stuckmann, line 87: CHECK: For the following the pass does not move the constant to the top, correct? Does this pattern appear in the kernels?

```
%0 = G_PTR_ADD %100, %101
%1 = G_PTR_ADD %100, %102
%2 = G_PTR_ADD %100, 32
```

F-Stuckmann: nevermind, I did get to the while loop

llvm/lib/Target/AIE/AIEClusterBaseAddress.cpp
```diff
@@ -74,6 +73,61 @@ static cl::opt<bool> EnableChainsForVectorLdSt(
     cl::desc("Enable ptradd chaining for vector loads and stores."));
 
 namespace {
+
+/// Try and re-order PTR_ADD instructions to maximise the size of constant
+/// PTR_ADD chains.
+bool optimisePostIncrements(Register BasePtr, ArrayRef<MachineInstr *> PtrAdds,
```
andcarminati: nit: this `BasePtr` seems to be unused.

llvm/lib/Target/AIE/AIEClusterBaseAddress.cpp
```diff
@@ -74,6 +73,61 @@ static cl::opt<bool> EnableChainsForVectorLdSt(
...
+    Observer.changingInstr(*PtrAdd);
+    MIB.setInstr(*PtrAdd);
+    Register NewOffsetReg =
+        MIB.buildConstant(LLT::scalar(20), CstOffset->Value.getSExtValue())
+            .getReg(0);
+    PtrAdd->getOperand(2).setReg(NewOffsetReg);
+    Observer.changedInstr(*PtrAdd);
+    Observer.changingInstr(OutputUser);
+    OutputUser.getOperand(2).setReg(OffsetReg);
+    OutputUser.clearFlag(MachineInstr::NoUWrap);
```
andcarminati, line 123: If we have such flag, should we also add it to the `PtrAdd`? Maybe a test could be also interesting.

gbossu: You can see it in the test that the `nuw` flag disappears. I could probably "swap" the flags as you suggest, but I just preferred to be conservative. I don't think the flags matter much in MIR.

andcarminati: Indeed, fair enough!

andcarminati:
This PR looks good. I left just one small nit.

llvm/lib/Target/AIE/AIEClusterBaseAddress.cpp
gbossu: Fixed

andcarminati (approved):
LGTM.

PR #308, opened 2025-01-23
[AIE2P] End-to-End support of fifo store intrinsics

khallouh:


llvm/include/llvm/IR/IntrinsicsAIE2P.td
```diff
@@ -591,4 +591,43 @@ def int_aie2p_sqrtf : ClangBuiltin<"__builtin_aie2p_sqrtf">, AIE2PNLF;
...
+     [llvm_ptr_ty, llvm_v64i8_ty, llvm_v8i8_ty, llvm_v32i32_ty, llvm_i32_ty],
+     [IntrWriteMem, IntrArgMemOnly]>;
+def int_aie2p_fifo_st_push_576_bfp16: AIE2PFIFO_ST_PUSH_576_BFP16;
+def int_aie2p_fifo_st_push_544_bfp16: AIE2PFIFO_ST_PUSH_576_BFP16;
+class AIE2PFIFO_ST_PUSH_512_BFP16 : Intrinsic<[llvm_ptr_ty, llvm_v32i32_ty, llvm_i32_ty],
+     [llvm_ptr_ty, llvm_v16i32_ty, llvm_v32i32_ty, llvm_i32_ty],
+     [IntrWriteMem, IntrArgMemOnly]>;
+def int_aie2p_fifo_st_push_512_bfp16: AIE2PFIFO_ST_PUSH_512_BFP16;
+
+// FLUSH normal
```
gbossu, line 645: Nit: I think this is referred to as "sequential" in our ISA.

llvm/lib/Target/AIE/aie2p/AIE2PInstructionSelector.cpp
```diff
@@ -348,6 +349,18 @@ bool AIE2PInstructionSelector::select(MachineInstr &I) {
     case Intrinsic::aie2p_v64accfloat_to_v64bfp16ebs16:
     case Intrinsic::aie2p_v64bfp16ebs8_to_v64bfp16ebs16:
       return selectVCONVbfp16(I, MRI);
+    case Intrinsic::aie2p_fifo_st_push_576_bfp16:
+    case Intrinsic::aie2p_fifo_st_push_544_bfp16:
+    case Intrinsic::aie2p_fifo_st_push_512_bfp16:
+    case Intrinsic::aie2p_fifo_st_flush:
+    case Intrinsic::aie2p_fifo_st_flush_conv:
+    case Intrinsic::aie2p_fifo_st_flush_1d:
```
gbossu, line 372: Would that be equivalent to what you call a normal `aie2p_fifo_st_flush` followed by a ptr increment? Same for 2d/3d?

llvm/test/CodeGen/AIE/aie2p/GlobalIsel/inst-select-fifo-stores.mir
```diff
@@ -0,0 +1,391 @@
...
+# This file is licensed under the Apache License v2.0 with LLVM Exceptions.
+# See https://llvm.org/LICENSE.txt for license information.
+# SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+#
+# (c) Copyright 2025 Advanced Micro Devices, Inc. or its affiliates
+# RUN: llc -mtriple aie2p -run-pass=instruction-select %s -verify-machineinstrs -o - | FileCheck %s
+
+
+---
+name:            _Z18test_fifo_st_resetRPDv64_DB8_S0_R12fifo_state_t
```
gbossu: please use human readable names :)

llvm/include/llvm/IR/IntrinsicsAIE2P.td
gbossu: Although that name is confusing, because apparently the pointer is not modified...

llvm/test/CodeGen/AIE/aie2p/GlobalIsel/regbankselect-fifo-insn.mir
```diff
@@ -4,7 +4,7 @@
 # See https://llvm.org/LICENSE.txt for license information.
 # SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
 #
-# (c) Copyright 2024 Advanced Micro Devices, Inc. or its affiliates
+# (c) Copyright 2025 Advanced Micro Devices, Inc. or its affiliates
```
gbossu, line 7: Nit: 2024-2025

llvm/include/llvm/IR/IntrinsicsAIE2P.td
khallouh: The instruction does update the pointer and the instruction is called "normal" in the architecture description file. See below.
`def VST_FLUSH_512_CONV_normal_flush : AIE2P_inst_st_instr32 <(outs mStFifo:$fifo_reg_out, mPfs:$ptr_out, mR26_fifo_st:$avail_out), (ins mStFifo:$fifo_reg, mPfs:$ptr, mR26_fifo_st:$avail)`

gbossu: Then I trust you :)

llvm/test/CodeGen/AIE/aie2p/GlobalIsel/inst-select-fifo-stores.mir
khallouh: Could this mean I am superhuman? haha. These tests were generated, hence the mangled names. We have quite a few like this already and I agree it is not ideal.

llvm/test/CodeGen/AIE/aie2p/GlobalIsel/regbankselect-fifo-insn.mir
khallouh: This file was actually added in 2025 :)

llvm/lib/Target/AIE/aie2p/AIE2PInstructionSelector.cpp
khallouh: I would hope so. And the architecture file calls it normal ;-)

llvm/test/CodeGen/AIE/aie2p/GlobalIsel/inst-select-fifo-stores.mir
gbossu: Maybe just pipe the file through `c++filt`?

llvm/lib/Target/AIE/aie2p/AIE2PInstructionSelector.cpp
gbossu: I was just wondering because we would then only need the `aie2p_fifo_st_flush` instrinsic/builtin. And we could get the 1d/2d/3d versions using instruction selection combiners like we do for "normal" load/stores. But then we would probably need new `G_AIE_FIFO_xxxx` opcodes, so it's probably not worth to do this now.

clang/lib/CodeGen/CGBuiltin.cpp
```diff
@@ -22305,6 +22328,111 @@ Value *CodeGenFunction::EmitAIEBuiltinExpr(unsigned BuiltinID,
 
     return Builder.CreateDefaultAlignedStore(Exp, ExpAddr);
   }
+  case AIE::BI__builtin_aie2p_fifo_st_push_576_bfp16:
+  case AIE::BI__builtin_aie2p_fifo_st_push_512_bfp16:
+  case AIE::BI__builtin_aie2p_fifo_st_push_544_bfp16:
+  case AIE::BI__builtin_aie2p_fifo_st_flush_conv:
+  case AIE::BI__builtin_aie2p_fifo_st_flush: {
```
gbossu, line 22362: Could you remind me why we need the custom builtin->intrinsic lowering? Is that because of our `i20` types?

clang/lib/CodeGen/CGBuiltin.cpp
```diff
@@ -22305,6 +22328,111 @@ Value *CodeGenFunction::EmitAIEBuiltinExpr(unsigned BuiltinID,
...
+    Value *FifoAddr;
+    Value *AvailAddr;
+    if (BuiltinID == AIE::BI__builtin_aie2p_fifo_st_flush ||
+        BuiltinID == AIE::BI__builtin_aie2p_fifo_st_flush_conv) {
+      FifoAddr = EmitLValue(E->getArg(1)).getPointer(*this);
+      AvailAddr = EmitLValue(E->getArg(2)).getPointer(*this);
+    } else if (BuiltinID == AIE::BI__builtin_aie2p_fifo_st_push_512_bfp16) {
+      FifoAddr = EmitLValue(E->getArg(2)).getPointer(*this);
+      AvailAddr = EmitLValue(E->getArg(3)).getPointer(*this);
+    } else {
```
gbossu, line 22386: Nit: I guess this `else` case is handling `BI__builtin_aie2p_fifo_st_push_576_bfp16` or `BI__builtin_aie2p_fifo_st_push_544_bfp16`. Maybe make that an assert?

clang/lib/CodeGen/CGBuiltin.cpp
```diff
@@ -22305,6 +22328,111 @@ Value *CodeGenFunction::EmitAIEBuiltinExpr(unsigned BuiltinID,
...
+    Function *F = CGM.getIntrinsic(IntrinsicID);
+    Value *Val = Builder.CreateCall(F, Ops);
+
+    Value *Ptr = Builder.CreateExtractValue(Val, 0);
+    Value *Fifo = Builder.CreateExtractValue(Val, 1);
+    Value *Avail = Builder.CreateExtractValue(Val, 2);
+    Value *PtrAddr = EmitLValue(E->getArg(0)).getPointer(*this);
+
+    Value *FifoAddr;
+    Value *AvailAddr;
```
gbossu: Please initialize those to `nullptr`

clang/lib/CodeGen/CGBuiltin.cpp
```diff
@@ -22305,6 +22328,111 @@ Value *CodeGenFunction::EmitAIEBuiltinExpr(unsigned BuiltinID,
...
+        BuiltinID == AIE::BI__builtin_aie2p_fifo_st_flush_conv_1d_byte) {
+      NumAddrIncs = 1;
+    } else if (BuiltinID == AIE::BI__builtin_aie2p_fifo_st_flush_2d_byte ||
+               BuiltinID == AIE::BI__builtin_aie2p_fifo_st_flush_conv_2d_byte) {
+      NumAddrIncs = 4;
+    } else if ((BuiltinID == AIE::BI__builtin_aie2p_fifo_st_flush_3d_byte ||
+                BuiltinID ==
+                    AIE::BI__builtin_aie2p_fifo_st_flush_conv_3d_byte)) {
+      NumAddrIncs = 7;
+    }
```
gbossu, line 22421: Nit: Use a `switch`?

clang/lib/CodeGen/CGBuiltin.cpp
khallouh: Yes, namely. The other reason is the indirect return of the fifo state.

clang/test/CodeGen/aie/aie2p/aie2p-ldst.cpp
```diff
@@ -82,3 +83,4086 @@ v128uint4 test_pack_v128i4_v128i8(v128uint8 v, int sign) {
...
+// CHECK-NEXT:    [[TMP11:%.*]] = trunc i32 [[INC2:%.*]] to i20
+// CHECK-NEXT:    [[TMP12:%.*]] = tail call { ptr, <32 x i32>, i32, i20, i20 } @llvm.aie2p.fifo.st.flush.3d(ptr [[TMP0]], <32 x i32> [[TMP1]], i32 [[TMP2]], i20 [[TMP3]], i20 [[TMP4]], i20 [[TMP6]], i20 [[TMP7]], i20 [[TMP8]], i20 [[TMP10]], i20 [[TMP11]])
+// CHECK-NEXT:    [[TMP13:%.*]] = extractvalue { ptr, <32 x i32>, i32, i20, i20 } [[TMP12]], 0
+// CHECK-NEXT:    [[TMP14:%.*]] = extractvalue { ptr, <32 x i32>, i32, i20, i20 } [[TMP12]], 1
+// CHECK-NEXT:    [[TMP15:%.*]] = extractvalue { ptr, <32 x i32>, i32, i20, i20 } [[TMP12]], 2
+// CHECK-NEXT:    [[TMP16:%.*]] = extractvalue { ptr, <32 x i32>, i32, i20, i20 } [[TMP12]], 3
+// CHECK-NEXT:    [[TMP17:%.*]] = zext i20 [[TMP16]] to i32
+// CHECK-NEXT:    [[TMP18:%.*]] = extractvalue { ptr, <32 x i32>, i32, i20, i20 } [[TMP12]], 4
+// CHECK-NEXT:    [[TMP19:%.*]] = zext i20 [[TMP18]] to i32
+// CHECK-NEXT:    store i32 [[TMP17]], ptr [[COUNT1]], align 4
```
gbossu, line 217: Do we have similar code for AIE2, or do we directly store an `i20`?

gbossu: Hmm I checked for `__builtin_aiev2_add_2d`, and we also have a zext. And we have no test in AIE2 for that builtin btw 🙈

clang/test/CodeGen/aie/aie2p/aie2p-ldst.cpp
```diff
@@ -82,3 +83,4086 @@ v128uint4 test_pack_v128i4_v128i8(v128uint8 v, int sign) {
...
+// CHECK-NEXT:    store i32 [[TMP13]], ptr [[COUNT1]], align 4
+// CHECK-NEXT:    store <32 x i32> [[TMP10]], ptr [[S]], align 128
+// CHECK-NEXT:    store i32 [[TMP11]], ptr [[POS1_I]], align 64
+// CHECK-NEXT:    store ptr [[TMP9]], ptr [[P]], align 4
+// CHECK-NEXT:    ret void
+//
+void test_fifo_st_flush_conv_2d_byte(v128int4 *&p, fifo_state_t &s, int off,
+                                     int size1, addr_t &count1, int inc1) {
+  return fifo_st_flush_conv_2d_byte(p, s, off, size1, count1, inc1);
+}
```
gbossu, line 400: Nit: would be nice to skip lines, otherwise it's hard to see which check lines correspond to which function.

clang/test/CodeGen/aie/aie2p/aie2p-ldst.cpp
```diff
@@ -82,3 +83,4086 @@ v128uint4 test_pack_v128i4_v128i8(v128uint8 v, int sign) {
...
+// CHECK-NEXT:    [[TMP5:%.*]] = extractvalue { ptr, <32 x i32>, i32 } [[TMP4]], 0
+// CHECK-NEXT:    [[TMP6:%.*]] = extractvalue { ptr, <32 x i32>, i32 } [[TMP4]], 1
+// CHECK-NEXT:    [[TMP7:%.*]] = extractvalue { ptr, <32 x i32>, i32 } [[TMP4]], 2
+// CHECK-NEXT:    store <32 x i32> [[TMP6]], ptr [[S]], align 128
+// CHECK-NEXT:    store i32 [[TMP7]], ptr [[POS1_I]], align 64
+// CHECK-NEXT:    store ptr [[TMP5]], ptr [[P]], align 4
+// CHECK-NEXT:    ret void
+//
+void test_fifo_st_reset(v128uint4 *&p, v128uint4 v, fifo_state_t &s) {
+  return fifo_st_push(p, v, s);
```
gbossu: The test says `reset`, but we call `push`

clang/test/CodeGen/aie/aie2p/aie2p-ldst.cpp
```diff
@@ -82,3 +83,4086 @@ v128uint4 test_pack_v128i4_v128i8(v128uint8 v, int sign) {
...
+// CHECK-NEXT:    [[TMP4:%.*]] = extractvalue { ptr, <32 x i32>, i32 } [[TMP3]], 0
+// CHECK-NEXT:    [[TMP5:%.*]] = extractvalue { ptr, <32 x i32>, i32 } [[TMP3]], 1
+// CHECK-NEXT:    [[TMP6:%.*]] = extractvalue { ptr, <32 x i32>, i32 } [[TMP3]], 2
+// CHECK-NEXT:    store <32 x i32> [[TMP5]], ptr [[S]], align 128
+// CHECK-NEXT:    store i32 [[TMP6]], ptr [[POS1_I]], align 64
+// CHECK-NEXT:    store ptr [[TMP4]], ptr [[P]], align 4
+// CHECK-NEXT:    ret void
+//
+void test_fifo_st_reset(v16uint32 *&p, v16uint32 v, fifo_state_t &s) {
+  return fifo_st_push(p, v, s);
```
gbossu: Same here, we do not call reset

clang/test/CodeGen/aie/aie2p/aie2p-ldst.cpp
```diff
@@ -82,3 +83,4086 @@ v128uint4 test_pack_v128i4_v128i8(v128uint8 v, int sign) {
...
+// CHECK-NEXT:    store i32 [[TMP13]], ptr [[COUNT1]], align 4
+// CHECK-NEXT:    store <32 x i32> [[TMP10]], ptr [[S]], align 128
+// CHECK-NEXT:    store i32 [[TMP11]], ptr [[POS1_I]], align 64
+// CHECK-NEXT:    store ptr [[TMP9]], ptr [[P]], align 4
+// CHECK-NEXT:    ret void
+//
+void test_fifo_st_flush_2d_byte(v16uint32 *&p, fifo_state_t &s, int off,
+                                int size1, addr_t &count1, int inc1) {
+  return fifo_st_flush_2d_byte(p, s, off, size1, count1, inc1);
+}
```
gbossu, line 2652: Nit: This really is a huge file and it is pretty much impossible to review. It seems that in many cases, only the vector type changes. Do you think it would be possible to templatise all those `test_xxxx` functions and run the same file multiple times with e.g. `-DTYPE=v16uint32`, `-DTYPE=v16int32`, etc.

At least I think that would work for the "flush" tests, because the vector type does not seem to appear in the check lines. For "push", we might have so write a test for each type, but there aren't that many push variants, so I think that's fine.

llvm/test/CodeGen/AIE/aie2p/ldst-fifo-stores.ll
```diff
@@ -0,0 +1,7292 @@
+; NOTE: Assertions have been autogenerated by utils/update_llc_test_checks.py
+;
+; This file is licensed under the Apache License v2.0 with LLVM Exceptions.
+; See https://llvm.org/LICENSE.txt for license information.
+; SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+;
+; (c) Copyright 2025 Advanced Micro Devices, Inc. or its affiliates
+; RUN: llc < %s -verify-machineinstrs -mtriple=aie2p | FileCheck %s
```
gbossu, line 8: I think many of the tests here are identical and only differ in the function name. Could you trim that file?

llvm/test/CodeGen/AIE/aie2p/ldst-fifo-stores.ll
```diff
@@ -0,0 +1,7292 @@
...
+
+%struct.v64bfp16ebs8 = type <{ <64 x i8>, <8 x i8> }>
+%struct.v64bfp16ebs16 = type <{ <64 x i8>, <8 x i8> }>
+
+; Function Attrs: mustprogress nounwind memory(write, argmem: readwrite, inaccessiblemem: none)
+define dso_local void @_Z18test_fifo_st_resetRPDv64_DB8_S0_R12fifo_state_t(ptr nocapture nonnull align 4 dereferenceable(4) %p, <64 x i8> noundef %v, ptr nocapture nonnull align 64 dereferenceable(256) %s) local_unnamed_addr #2 {
```
gbossu, line 14: Note: this is also calling push here, not reset.

clang/test/CodeGen/aie/aie2p/aie2p-ldst.cpp
khallouh: I tried it but there is still the function label line which depends on the type.

llvm/test/CodeGen/AIE/aie2p/GlobalIsel/inst-select-fifo-stores.mir
khallouh: `c++filt` could not demangle all of them for some reason. I've done it manually :)

clang/lib/CodeGen/CGBuiltin.cpp
khallouh: Done

clang/lib/CodeGen/CGBuiltin.cpp
khallouh: Done

clang/lib/CodeGen/CGBuiltin.cpp
khallouh: Done

clang/test/CodeGen/aie/aie2p/aie2p-ldst.cpp
khallouh: Yes

clang/test/CodeGen/aie/aie2p/aie2p-ldst.cpp
khallouh: Done

clang/test/CodeGen/aie/aie2p/aie2p-ldst.cpp
khallouh: Fixed

llvm/test/CodeGen/AIE/aie2p/ldst-fifo-stores.ll
khallouh: I trimmed the file.

llvm/lib/Target/AIE/aie2p/AIE2PInstructionSelector.cpp
```diff
@@ -3166,6 +3179,156 @@ bool AIE2PInstructionSelector ::selectVSHUFFLE_BFP(MachineInstr &I,
   return constrainSelectedInstRegOperands(*MI, TII, TRI, RBI);
 }
 
+unsigned int getStoreFifoOpcode(MachineInstr &I) {
```
niwinanto, line 4010: I would suggest move this to `AIE2PInstructionInfo` where we have similar APIs, or even extend the `getOpCode`

clang/test/CodeGen/aie/aie2p/aie2p-ldst.cpp
```diff
@@ -82,3 +84,4206 @@ v128uint4 test_pack_v128i4_v128i8(v128uint8 v, int sign) {
...
+// CHECK-LABEL: @_Z18test_fifo_st_resetRPDv64_DB8_S0_R12fifo_state_t(
+// CHECK-NEXT:  entry:
+// CHECK-NEXT:    [[TMP0:%.*]] = load ptr, ptr [[P:%.*]], align 4, !tbaa [[TBAA2:![0-9]+]]
+// CHECK-NEXT:    [[TMP1:%.*]] = bitcast <64 x i8> [[V:%.*]] to <16 x i32>
+// CHECK-NEXT:    [[TMP2:%.*]] = load <32 x i32>, ptr [[S:%.*]], align 64, !tbaa [[TBAA6:![0-9]+]]
+// CHECK-NEXT:    [[TMP3:%.*]] = tail call { ptr, <32 x i32>, i32 } @llvm.aie2p.fifo.st.push.512.bfp16(ptr [[TMP0]], <16 x i32> [[TMP1]], <32 x i32> [[TMP2]], i32 0)
+// CHECK-NEXT:    [[TMP4:%.*]] = extractvalue { ptr, <32 x i32>, i32 } [[TMP3]], 0
+// CHECK-NEXT:    [[TMP5:%.*]] = extractvalue { ptr, <32 x i32>, i32 } [[TMP3]], 1
+// CHECK-NEXT:    store <32 x i32> [[TMP5]], ptr [[S]], align 128
+// CHECK-NEXT:    store ptr [[TMP4]], ptr [[P]], align 4
```
gbossu, line 100: Shouldn't the new availability also be stored back?

clang/test/CodeGen/aie/aie2p/aie2p-ldst.cpp
```diff
@@ -82,3 +84,4206 @@ v128uint4 test_pack_v128i4_v128i8(v128uint8 v, int sign) {
...
+// CHECK-NEXT:    [[TMP4:%.*]] = extractvalue { ptr, <32 x i32>, i32 } [[TMP3]], 0
+// CHECK-NEXT:    [[TMP5:%.*]] = extractvalue { ptr, <32 x i32>, i32 } [[TMP3]], 1
+// CHECK-NEXT:    [[TMP6:%.*]] = extractvalue { ptr, <32 x i32>, i32 } [[TMP3]], 2
+// CHECK-NEXT:    store <32 x i32> [[TMP5]], ptr [[S]], align 128
+// CHECK-NEXT:    store i32 [[TMP6]], ptr [[POS1_I]], align 64
+// CHECK-NEXT:    store ptr [[TMP4]], ptr [[P]], align 4
+// CHECK-NEXT:    ret void
+//
+void test_fifo_st_flush_bare(v128int4 *&p, fifo_state_t &s) {
+  return fifo_st_flush_bare(p, s);
```
gbossu, line 247: What is the difference with a standard `fifo_st_flush`?

khallouh: As far as I can understand from the API, there is no difference.

clang/test/CodeGen/aie/aie2p/aie2p-ldst.cpp
khallouh: Yes, that was a bug. It should be fixed now!

llvm/lib/Target/AIE/aie2p/AIE2PInstructionSelector.cpp
khallouh: You mean to move it to `AIE2PInstrInfo`? Because it is already in `AIE2PInstructionInfo`

clang/lib/CodeGen/CGBuiltin.cpp
gbossu: Super-nit: Also above for defining `FifoAddr` and `AvailAddr`.

clang/lib/Headers/aie2p_ldst.h
```diff
@@ -81,4 +81,118 @@ INTRINSIC(v64uint16) unpack(v64uint8 v) { return unpack(v, __SIGN_UNSIGNED); }
 INTRINSIC(v128int8) unpack(v128int4 v) { return unpack(v, __SIGN_SIGNED); }
 INTRINSIC(v128uint8) unpack(v128uint4 v) { return unpack(v, __SIGN_UNSIGNED); }
 
+#define FIFO_ST_PUSH_NORMAL(T)                                                 \
+  INTRINSIC(void) fifo_st_reset(T *&p, T v, fifo_state_t &s) {                 \
+    int pos = 0;                                                               \
+    sparse_fifo_t &fifo = s.fifo;                                              \
+    __builtin_aie2p_fifo_st_push_512_bfp16((int *&)p, v, fifo, pos);           \
+  }                                                                            \
```
gbossu, line 90: I'm not sure I understand what `fifo_st_reset` is supposed to be doing. I'm assuming it's the very first store of a sequence, meaning the availability must be reset to 0.  If that's the case, shouldn't we have this instead?
```
int &pos = s.pos;
pos = 0;
...
```
Or just:
```
  INTRINSIC(void) fifo_st_reset(T *&p, T v, fifo_state_t &s) {                 \
    s.pos = 0;                                                                        \
    __builtin_aie2p_fifo_st_push_512_bfp16((int *&)p, v, s.fifo, s.pos);     \
  }
```

khallouh: I fixed this already. Did you look at the last fixup commit?

khallouh: This diff is from the first commit before the fix.

gbossu: (Same for the other `fifo_st_reset`)

clang/test/CodeGen/aie/aie2p/aie2p-ldst.cpp
gbossu: I think it's still not stored back. See https://github.com/Xilinx/llvm-aie/pull/308/commits/e30d6efc40f2368a9d4becdb4c55d13f57fb2768#r1930335808

khallouh: But I see it is stored here in the fixup commit [be6981a0f00aa860](https://github.com/Xilinx/llvm-aie/pull/308/commits/be6981a0f00aa860350434d867947755b4298a11)

clang/lib/Headers/aie2p_ldst.h
```diff
@@ -81,4 +81,118 @@ INTRINSIC(v64uint16) unpack(v64uint8 v) { return unpack(v, __SIGN_UNSIGNED); }
...
+FIFO_ST_NORMAL(v32bfloat16)
+FIFO_ST_NORMAL(v16float)
+FIFO_ST_NORMAL(v128int4)
+FIFO_ST_NORMAL(v128uint4)
+FIFO_ST_NORMAL(v64int8)
+FIFO_ST_NORMAL(v64uint8)
+FIFO_ST_NORMAL(v32int16)
+FIFO_ST_NORMAL(v32uint16)
+FIFO_ST_NORMAL(v16int32)
+FIFO_ST_NORMAL(v16uint32)
```
konstantinschwarz: Should `#undef` these macros after their usage

clang/lib/Headers/aie2p_ldst.h
```diff
@@ -81,4 +81,118 @@ INTRINSIC(v64uint16) unpack(v64uint8 v) { return unpack(v, __SIGN_UNSIGNED); }
 INTRINSIC(v128int8) unpack(v128int4 v) { return unpack(v, __SIGN_SIGNED); }
 INTRINSIC(v128uint8) unpack(v128uint4 v) { return unpack(v, __SIGN_UNSIGNED); }
 
+#define FIFO_ST_PUSH_NORMAL(T)                                                 \
+  INTRINSIC(void) fifo_st_reset(T *&p, T v, fifo_state_t &s) {                 \
```
konstantinschwarz: The API also supports `volatile` qualified pointers. Could you add those variants as well?

abhinay-anubola:
It would be good if you could squash the fixup commit with the original commit. This will help keep the commit history clean.

khallouh:
> It would be good if you could squash the fixup commit with the original commit. This will help keep the commit history clean.

I will do that later. Just wanted to show what the fixup commit fixed and when.

clang/lib/Headers/aie2p_ldst.h
konstantinschwarz: Correction: I meant `restrict`, not `volatile`

khallouh: I have added the versions with memory bank annotations and the `restrict` qualifier. See the last fixup commit.

clang/lib/Headers/aie2p_ldst.h
khallouh: Done

clang/test/CodeGen/aie/aie2p/aie2p-ldst.cpp
```diff
@@ -5,7 +5,7 @@
 // See https://llvm.org/LICENSE.txt for license information.
 // SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
 //
-// (c) Copyright 2024 Advanced Micro Devices, Inc. or its affiliates
+// (c) Copyright 2024-2025 Advanced Micro Devices, Inc. or its affiliates
```
gbossu, line 8: Could you add some tests to see how FIFO operations to a given address bank get lowered into IR? Maybe just bank A, no need to do them all.

clang/test/CodeGen/aie/aie-bank-annotation.cpp
```diff
@@ -5,11 +5,12 @@
...
 // SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
 //
-// (c) Copyright 2024 Advanced Micro Devices, Inc. or its affiliates
+// (c) Copyright 2025 Advanced Micro Devices, Inc. or its affiliates
 //
 //===----------------------------------------------------------------------===//
-// Test to check the bank annotation for AIE2, specifically the translation of __aie_dm_resource_* to the correct address space.
+// Test to check the bank annotation for AIE, specifically the translation of __aie_dm_resource_* to the correct address space.
 // RUN: %clang -O1 --target=aie2 -nostdlibinc -S -emit-llvm %s -o - | FileCheck %s
+// RUN: %clang -O1 --target=aie2p -nostdlibinc -S -emit-llvm %s -o - | FileCheck %s
```
gbossu, line 13: Noice

gbossu (dismissed):
LGTM. I cannot verify every line, but what I've seen is correct and our regression suite seems to confirm that. Nice work!

gbossu (approved):

PR #314, opened 2025-01-28
[AIE2P] Optimize G_PHI regbankselect for fifo/acc

andcarminati:
Observation: this PR should include more tests after merging more fifo intrinsics.


llvm/test/CodeGen/AIE/aie2p/GlobalIsel/regbankselect-phi.mir
```diff
@@ -0,0 +1,206 @@
...
+    G_BR %bb.2
+
+  bb.2:
+    PseudoRET implicit $lr
+
+...
+
+# Test 3:  The instruction  %19:_(p0) = G_PHI %0(p0), %bb.0, %27(p0), %bb.1
+# should not be mapped to accregbank. The result of the PHI node is used by an instruction
+# that is considered an `accululator-use instruction` but the considered result is just a scalar
```
martien-de-jong: nit accumulator

andcarminati: Oh, did I really write this? Good catch!

llvm/test/CodeGen/AIE/aie2p/GlobalIsel/regbankselect-phi.mir
```diff
@@ -0,0 +1,206 @@
...
+    %27:_(p0) = G_PTR_ADD %19, %5
+    %64:_(s32) = G_ADD %18, %63
+    %69:_(s32) = G_ICMP intpred(eq), %64(s32), %4
+    G_BRCOND %69(s32), %bb.1
+    G_BR %bb.2
+
+  bb.2:
+    PseudoRET implicit $lr
+
+...
```
gbossu, line 270: Nit: Maybe add a test where a PHI is used both as a PHI and an ACC. I think we will favour the FIFO mapping at this point.

llvm/lib/Target/AIE/aie2p/AIE2PRegisterBankInfo.cpp
```diff
@@ -966,6 +976,36 @@ AIE2PRegisterBankInfo::getInstrMapping(const MachineInstr &MI) const {
     }
     return AIEBaseRegisterBankInfo::getInstrMapping(MI);
   }
+  case TargetOpcode::G_PHI: {
```
andcarminati: This is a really special-case to handle. If we map to vector reg bank, we may block optimal SWP scheduling's after.  This is also a special case because one single PHI node may contain inputs mapped to fifo, accumulator and vector bank at the same time. The way it is mapped is also different, we handle only the output and inputs are related the inputs.

llvm/test/CodeGen/AIE/aie2p/GlobalIsel/regbankselect-phi.mir
andcarminati: Do you mean by both FIFO and ACC? Yes, we should favor the FIFO because we check it first. I hope we will not face this situation, according to the benchmarks I saw, what do you think?

khallouh: Why should we favor the FIFO bank?

llvm/test/CodeGen/AIE/aie2p/GlobalIsel/regbankselect-phi.mir
```diff
@@ -0,0 +1,282 @@
...
+
+# Test 3:  The instruction  %19:_(p0) = G_PHI %0(p0), %bb.0, %27(p0), %bb.1
+# should not be mapped to accregbank. The result of the PHI node is used by an instruction
```
khallouh, line 202: Maybe add a similar test for a fifo incompatible type?

llvm/test/CodeGen/AIE/aie2p/GlobalIsel/regbankselect-phi.mir
```diff
@@ -0,0 +1,282 @@
...
+    G_BRCOND %69(s32), %bb.1
+    G_BR %bb.2
+
+  bb.2:
+    PseudoRET implicit $lr
+
+...
+
+# Test 4:  The instruction  %19:_(p0) = G_PHI %0(p0), %bb.0, %27(p0), %bb.1
+# is also used as PHI input.
```
khallouh, line 273: Nit: do you mean? `%23:_(<32 x s32>) = G_PHI %66(<32 x s32>), %bb.0, %28(<32 x s32>), %bb.1`

andcarminati: Oh, true.

llvm/lib/Target/AIE/aie2p/AIE2PRegisterBankInfo.cpp
andcarminati: Handled in a different way now.

llvm/lib/Target/AIE/aie2p/AIE2PRegisterBankInfo.cpp
```diff
@@ -729,56 +733,62 @@ bool AIE2PRegisterBankInfo::hasFifoInput(const MachineInstr &MI,
...
+    if (Depth > MaxDepthBankSearch)
+      return false;
+    // skip copies
+    if (UseOpcode == TargetOpcode::G_BITCAST || IsCopyToVReg(UseMI) ||
+        UseMI.isPHI()) {
+      Register DefReg = UseMI.getOperand(0).getReg();
+      if (DefReg.isPhysical())
+        continue;
+      if (isUseFifoInsn(MRI, TRI, DefReg, Depth + 1))
+        return true;
```
khallouh, line 801: This is great that we now look into the real use. Can you add a test for this with bitcasts and copies? I don't think we have that for the fifo bank.

llvm/lib/Target/AIE/aie2p/AIE2PRegisterBankInfo.cpp
```diff
@@ -729,56 +733,62 @@ bool AIE2PRegisterBankInfo::hasFifoInput(const MachineInstr &MI,
...
+      return false;
+    // skip copies
+    if (UseOpcode == TargetOpcode::G_BITCAST || IsCopyToVReg(UseMI) ||
+        UseMI.isPHI()) {
+      Register DefReg = UseMI.getOperand(0).getReg();
+      if (DefReg.isPhysical())
+        continue;
+      if (isUseAccInsn(MRI, TRI, DefReg, Depth + 1))
+        return true;
+    } else if (MRI.getType(RegOp).getSizeInBits() == 2048 ||
```
khallouh: This makes sense for 2048-bit vectors. Does it address a specific case that we did not handle or is this just to return earlier?

khallouh:
I left a couple of comments, where a couple of more tests would be nice. Otherwise, this looks good to me.

llvm/lib/Target/AIE/aie2p/AIE2PRegisterBankInfo.cpp
andcarminati: In this case, I just kept the same check that we had before. I can investigate if we can remove.

khallouh: After taking a look myself, it seems this just serves as an early return. Functionally, I don't think it's needed. We can keep it but in that case it can be moved up out of the loop since the size of `RegOp` doesn't depend on its uses (UseMI).

andcarminati: I see the same, I will drop this check, in this way we can improve for a further refactor (integration with fifo handling code).

andcarminati:
> I left a couple of comments, where a couple of more tests would be nice. Otherwise, this looks good to me.

Hi @khallouh , thank you for the review. I hope I addressed all suggestions.

llvm/lib/Target/AIE/aie2p/AIE2PRegisterBankInfo.cpp
```diff
@@ -745,56 +749,61 @@ bool AIE2PRegisterBankInfo::hasFifoInput(const MachineInstr &MI,
...
 
-  // Now check for accumulator-usage using the updated TracedRegOp
-  return any_of(MRI.use_nodbg_instructions(TracedRegOp),
-                [&](const MachineInstr &UseMI) {
-                  // Check if this instruction uses the accumulator register
-                  return (MRI.getType(TracedRegOp).getSizeInBits() == 2048) ||
-                         usesAccReg(UseMI, MRI, TRI, TracedRegOp);
-                });
+    // skip copies
+    if (UseOpcode == TargetOpcode::G_BITCAST || IsCopyToVReg(UseMI) ||
```
khallouh, line 765: Nit: ...copies, bitcasts and phis

llvm/lib/Target/AIE/aie2p/AIE2PRegisterBankInfo.cpp
andcarminati: Done.

llvm/test/CodeGen/AIE/aie2p/GlobalIsel/regbankselect-phi.mir
andcarminati: Done! We have a implicit pointer definition in `test_fifo.ld.fill.implicit.ptr`.

llvm/test/CodeGen/AIE/aie2p/GlobalIsel/regbankselect-phi.mir
andcarminati: I feel that this is a corner case situation. If we see this case in the future we can handle properly.

khallouh (dismissed):
LGTM!

llvm/lib/Target/AIE/aie2p/AIE2PRegisterBankInfo.cpp
```diff
@@ -745,56 +749,61 @@ bool AIE2PRegisterBankInfo::hasFifoInput(const MachineInstr &MI,
...
+  for (auto &UseMI : MRI.use_nodbg_instructions(RegOp)) {
+    const unsigned UseOpcode = UseMI.getOpcode();
+
+    // skip copies, bitcasts and phis
+    if (UseOpcode == TargetOpcode::G_BITCAST || IsCopyToVReg(UseMI) ||
+        UseMI.isPHI()) {
+      Register DefReg = UseMI.getOperand(0).getReg();
+      if (DefReg.isPhysical())
+        continue;
+      if (isUseFifoInsn(MRI, TRI, DefReg, Depth + 1))
```
F-Stuckmann, line 800: `isUseAccInsn` and `isUseFifoInsn` are very similar.
Could you pass the different functions as a function pointers? Then you don't duplicate code.

```

[isUseFifoInsn(const MachineRegisterInfo &MRI,
                                          const TargetRegisterInfo &TRI,
                                          const Register RegOp,
                                          unsigned Depth) const {
    return RecursiveOperandSearch(MRI, TRI, RegOp, depth, isUseFifoInsn, hasFifoInput) 
}

bool AIE2PRegisterBankInfo::isUseAccInsn(const MachineRegisterInfo &MRI,
                                         const TargetRegisterInfo &TRI,
                                         const Register &RegOp,
                                         unsigned Depth) const {
    return RecursiveOperandSearch(MRI, TRI, RegOp, depth, isUseAccInsn, usesAccReg) 
}
```

andcarminati: Indeed they are asking for it. However we need a bit more discussions to harmonize with https://github.com/Xilinx/llvm-aie/pull/325.

andcarminati: If possible, I would like to shift this refactor to the aforementioned PR.

llvm/lib/Target/AIE/aie2p/AIE2PRegisterBankInfo.cpp
```diff
@@ -745,56 +749,61 @@ bool AIE2PRegisterBankInfo::hasFifoInput(const MachineInstr &MI,
...
 
-  // Create a non-const copy of RegOp for tracing
-  Register TracedRegOp = RegOp;
+  auto IsCopyToVReg = [](const MachineInstr &MI) {
+    return (MI.isCopy() && MI.getOperand(0).getReg().isVirtual());
+  };
 
-  // Trace RegOp to its actual register, updating it if necessary
-  traceToActualReg(TracedRegOp);
+  for (auto &UseMI : MRI.use_nodbg_instructions(RegOp)) {
```
niwinanto, line 761: Do we really need to check for all the uses?

andcarminati: Discussed offline. Now this cases is covered with the test case `test_fifo.ld.fill.store`.

llvm/lib/Target/AIE/aie2p/AIE2PRegisterBankInfo.cpp
F-Stuckmann: that does sound like the correct place to do this

khallouh (approved):

PR #316, opened 2025-01-29
[AIE2P] Handle large stack offset for PLFR

niwinanto:
This change handles large stack offset for PLFR, which split the pseudo instruction again to pseudo which needs special handling.

### DIFF
--- a/llvm/lib/Target/AIE/AIEBaseInstrInfo.cpp
+++ b/llvm/lib/Target/AIE/AIEBaseInstrInfo.cpp
@@ -457,7 +457,10 @@ void AIEBaseInstrInfo::expandSpillPseudo(
         EI.SubRegIndex ? Register(TRI.getSubReg(Reg, EI.SubRegIndex)) : Reg;
     if (MI.hasOrderedMemoryRef() || SEH.isLiveReg(SubReg)) {
       MachineInstrBuilder MIB;
-      if (SPReg) {
+      // If the SPReg is set, we need to spill using register offset. However,
+      // for pseudo instructions, we skip this step since we will come back to
+      // it once we expand the pseudo instruction into multiple instructions.
+      if (SPReg && !get(EI.ExpandedOpCode).isPseudo()) {
         auto OffsetRegInfo =
             getRegOffsetSpillInstrInfoFromImmOffset(EI.ExpandedOpCode);
         MachineFunction &MF = *MI.getMF();
@@ -502,7 +505,7 @@ void AIEBaseInstrInfo::expandSpillPseudo(
   // E.g. In AIE2 eDS spills are expanded into 2 eD spills, which are themselves
   // expanded into 4 other spills.
   for (MachineInstr *ExpandedMI : ExpandedInsts)
-    expandSpillPseudo(*ExpandedMI, TRI, SubRegOffsetAlign);
+    expandSpillPseudo(*ExpandedMI, TRI, SubRegOffsetAlign, SPReg);
 }
 
 /// Collect all the "atomic" sub-registers that make up \ref Reg.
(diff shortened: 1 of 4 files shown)


llvm/lib/Target/AIE/AIEBaseInstrInfo.cpp
```diff
@@ -457,7 +457,7 @@ void AIEBaseInstrInfo::expandSpillPseudo(
         EI.SubRegIndex ? Register(TRI.getSubReg(Reg, EI.SubRegIndex)) : Reg;
     if (MI.hasOrderedMemoryRef() || SEH.isLiveReg(SubReg)) {
       MachineInstrBuilder MIB;
-      if (SPReg) {
+      if (SPReg && !isSpillPseudoOpcode(EI.ExpandedOpCode)) {
```
gbossu: Would that be equivalent to `SPReg && !TII.get(EI.ExpandedOpCode).isPseudo()`? I'd like to avoid defining a new hook that is very similar to `getSpillPseudoExpandInfo`.

niwinanto: Nice, I didnt know we could get pseudo information from opcode. Will remove the hook.

niwinanto: Done.

llvm/lib/Target/AIE/AIEBaseInstrInfo.cpp
```diff
@@ -457,7 +457,7 @@ void AIEBaseInstrInfo::expandSpillPseudo(
         EI.SubRegIndex ? Register(TRI.getSubReg(Reg, EI.SubRegIndex)) : Reg;
     if (MI.hasOrderedMemoryRef() || SEH.isLiveReg(SubReg)) {
       MachineInstrBuilder MIB;
-      if (SPReg) {
+      if (SPReg && !get(EI.ExpandedOpCode).isPseudo()) {
```
andcarminati, line 463: Please, add a comment here to state the goal of this change.

llvm/test/CodeGen/AIE/aie2p/eliminate-frame-index.mir
```diff
@@ -413,3 +413,42 @@ body:             |
...
+    ; CHECK-LABEL: name: test_fifo_plfr_bigstack
+    ; CHECK: frame-setup PADDXM_pstm_sp_imm 2240, implicit-def $sp, implicit $sp
+    ; CHECK-NEXT: $p0 = MOV_alu_mv_mv_mv_scl $sp
+    ; CHECK-NEXT: $m0 = MOVXM -2240
+    ; CHECK-NEXT: $p0 = PADD_mod_pseudo $p0, killed $m0
+    ; CHECK-NEXT: $p0 = MOV_alu_mv_mv_mv_scl $sp
+    ; CHECK-NEXT: $dj0 = MOVXM -2240
+    ; CHECK-NEXT: $lfl0 = VLDA_dmx_lda_fifohl_idx $p0, killed $dj0
+    ; CHECK-NEXT: $dj0 = MOVXM -2176
+    ; CHECK-NEXT: $lfh0 = VLDA_dmx_lda_fifohl_idx $p0, killed $dj0
```
gbossu, line 435: I thought you said in a previous PR that only scalar ld/st can have an out-of-range immediate. Maybe I got confused?

niwinanto: You are right, other ld/st has more encoding bits, so spilling through register offset is not really needed. But here, `VST_PLFR_SPILL` composed of `LDA_dms_lda_idx` and `VST_dmx_sts_fifohl_idx`. To make the implementation simple, I just spill both using register offset, even though `VST_dmx_sts_fifohl_idx` can encode this offset.

gbossu: Hmm yes, unless we expand the pseudos before frame elimination, that is difficult/tedious to generate optimal code. Hopefully we won't spill with large offsets very often.

gbossu (approved):

llvm/test/CodeGen/AIE/aie2p/eliminate-frame-index.mir
niwinanto: Yes, hope code selection will be better over the time and we might not hit this often.

PR #317, opened 2025-01-29
[AIE2P] RegBankSelect G_UNMERGE_VALUES for ACC1024

andcarminati:
This overcome some inefficiency found in several benchmarks.

### DIFF
--- a/llvm/lib/Target/AIE/aie2p/AIE2PRegisterBankInfo.cpp
+++ b/llvm/lib/Target/AIE/aie2p/AIE2PRegisterBankInfo.cpp
@@ -819,16 +819,28 @@ AIE2PRegisterBankInfo::getInstrMapping(const MachineInstr &MI) const {
                                                          OpRegBankIdx);
   }
   case TargetOpcode::G_UNMERGE_VALUES: {
+    assert(MI.getNumOperands() == 3 &&
+           "Unsupported number of operands for G_UNMERGE_VALUES\n");
     const Register SrcReg = MI.getOperand(2).getReg();
     const LLT SrcTy = MRI.getType(SrcReg);
-    if (SrcTy.getSizeInBits() == 2048) {
+
+    auto *RB = getRegBank(SrcReg, MRI, TRI);
+    if ((RB == &AIE2P::AccRegBank) &&
+        (SrcTy.getSizeInBits() == 1024 || SrcTy.getSizeInBits() == 2048)) {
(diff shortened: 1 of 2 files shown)


abnikant (dismissed):
LGTM!

llvm/lib/Target/AIE/aie2p/AIE2PRegisterBankInfo.cpp
```diff
@@ -829,6 +830,17 @@ AIE2PRegisterBankInfo::getInstrMapping(const MachineInstr &MI) const {
                               getValueMapping(PMI_ACC2048, 2048)}),
           /*NumOperands*/ 3);
     }
+
+    auto *RB = getRegBank(SrcReg, MRI, TRI);
```
niwinanto, line 827: Could you combine this with `if` block just above.  You can use `getPartialMappingIdx`

andcarminati: Sure ;-) done!

llvm/lib/Target/AIE/aie2p/AIE2PRegisterBankInfo.cpp
```diff
@@ -821,14 +821,24 @@ AIE2PRegisterBankInfo::getInstrMapping(const MachineInstr &MI) const {
   case TargetOpcode::G_UNMERGE_VALUES: {
     const Register SrcReg = MI.getOperand(2).getReg();
     const LLT SrcTy = MRI.getType(SrcReg);
-    if (SrcTy.getSizeInBits() == 2048) {
+
+    auto *RB = getRegBank(SrcReg, MRI, TRI);
+    if ((RB == &AIE2P::AccRegBank) &&
+        (SrcTy.getSizeInBits() == 1024 || SrcTy.getSizeInBits() == 2048)) {
```
niwinanto, line 829: Do you have a wrong parenthesis? Is it `(RB == &AIE2P::AccRegBank && SrcTy.getSizeInBits() == 1024) || SrcTy.getSizeInBits() == 2048)`

andcarminati: Clarified offline. If we are assigning acc bank is because the input is already acc reg bank. So the comparison makes the `if` a bit more symmetric.

niwinanto (approved):
LGTM!

PR #318, opened 2025-01-29
[AIE2P][simplify CR] added simplification of control registers

F-Stuckmann:


llvm/lib/Target/AIE/AIEBaseRegisterInfo.cpp
```diff
@@ -0,0 +1,22 @@
+//===-- AIEBaseRegisterInfo.cpp - Common AIE Register Info -*- C++ -*------===//
```
andcarminati: Check: it looks like a class implementation, but just a command line option for now.

gbossu: Maybe `llvm/lib/Target/AIE/AIEBaseTargetMachine.cpp` would be a better location. What do you think?

llvm/test/CodeGen/AIE/aie2p/GlobalIsel/post-select-optimize-erase-physregs.mir
```diff
@@ -0,0 +1,896 @@
...
+#
+# (c) Copyright 2025 Advanced Micro Devices, Inc. or its affiliates
+
+# RUN: llc -mtriple aie2p \
+# RUN: -run-pass=dead-mi-elimination,aie-post-select-optimize,dead-mi-elimination \
+# RUN: --aie-simplify-crsr-edges=true -verify-machineinstrs %s -o - | FileCheck %s
+
+# Test 1 - Remove identity copies.
+# Transform -
+# %0 = COPY $crsat               |   |    |   %0 = COPY $crsat
```
andcarminati: I think it is `crfpmask` instead on `crsat` in this first test.

F-Stuckmann: you have eagle eyes, yeah I missed that

F-Stuckmann:
just a rebase to get the unit tests to work again

llvm/lib/Target/AIE/aie2p/AIE2PRegisterInfo.cpp
```diff
@@ -32,6 +32,8 @@ using namespace llvm;
 
 #define DEBUG_TYPE "aie2p-reg-info"
 
+extern cl::opt<bool> SimplifyCRSRRegs;
```
martien-de-jong, line 35: Not for now, but Yes, we should have a central header file with these declarations somewhere in AIEUtils. That would also be the perfect place to elaborate on the effects and the reasoning behind the defaults.

F-Stuckmann: I 100% agree

martien-de-jong (approved):

PR #321, opened 2025-01-30
[AIE2P] Fix align requirement for wide vector load/store and split 32 aligned load/store instead of scalarizing them 

khallouh:


andcarminati:
Nice, I think we need the same for AIE2.

konstantinschwarz:
> Nice, I think we need the same for AIE2.

AIE2 doesn't have 512-bit load/store instructions, so those will be broken down into 256-bit accesses which only require 32-byte alignment. I think AIE2 is fine

andcarminati:
Nice, I think we need the same for AIE2.

> > Nice, I think we need the same for AIE2.
> 
> AIE2 doesn't have 512-bit load/store instructions, so those will be broken down into 256-bit accesses which only require 32-byte alignment. I think AIE2 is fine

The problem there is a bit different. For AIE2 we are considering alignments of 32 bits (probably a misunderstanding of how alignment is expressed during legalization.).

khallouh:
This now splits the large loads/stores when the alignment is 256 bits.

llvm/test/CodeGen/AIE/GlobalISel/legalize-vector-load.mir
```diff
@@ -116,12 +116,23 @@ body:             |
...
+    ; AIE2-NEXT: {{  $}}
+    ; AIE2-NEXT: [[COPY:%[0-9]+]]:_(p0) = COPY $p0
+    ; AIE2-NEXT: [[LOAD:%[0-9]+]]:_(<64 x s8>) = G_LOAD [[COPY]](p0) :: (load (<64 x s8>), align 32)
+    ; AIE2-NEXT: PseudoRET implicit $lr, implicit [[LOAD]](<64 x s8>)
+    ;
+    ; AIE2P-LABEL: name: test_v64int8
+    ; AIE2P: liveins: $p0
+    ; AIE2P-NEXT: {{  $}}
+    ; AIE2P-NEXT: [[COPY:%[0-9]+]]:_(p0) = COPY $p0
+    ; AIE2P-NEXT: [[LOAD:%[0-9]+]]:_(<32 x s8>) = G_LOAD [[COPY]](p0) :: (load (<32 x s8>))
```
martien-de-jong, line 130: I assume we don't need the alignment anymore, but it seems wrong that it disappears. It's still 32 byte aligned, isn't it?

khallouh: yeah, it always disappears post legalizer, even if we just use `.legal()`, not sure why

llvm/lib/Target/AIE/aie2p/AIE2PLegalizerInfo.cpp
```diff
@@ -400,19 +400,44 @@ AIE2PLegalizerInfo::AIE2PLegalizerInfo(const AIE2PSubtarget &ST)
...
+            // == 128)
+            //   return true;
+            return false;
+          },
+          [=](const LegalityQuery &Query) {
+            const LLT &MemoryType = Query.MMODescrs[0].MemoryTy;
+            const uint64_t Alignment = Query.MMODescrs[0].AlignInBits;
+            const unsigned SplitFactor = MemoryType.getSizeInBits() / Alignment;
+            return std::make_pair(
+                0, LLT::fixed_vector(MemoryType.getNumElements() / SplitFactor,
```
andcarminati, line 437: nit: `return std::make_pair(0, MemoryType.divide(SplitFactor));`

llvm/lib/Target/AIE/aie2p/AIE2PLegalizerInfo.cpp
```diff
@@ -400,19 +400,44 @@ AIE2PLegalizerInfo::AIE2PLegalizerInfo(const AIE2PSubtarget &ST)
...
+          {AccV32S64, P0, AccV32S64, 512},
           {S128, P0, S128, 16},
       })
+      // Split 256-bit aligned load/store of 512-bit or larger vectors into
+      // legal 256-bit loads with 32-byte alignment.
+      .fewerElementsIf(
+          [=](const LegalityQuery &Query) {
+            const LLT &MemoryType = Query.MMODescrs[0].MemoryTy;
+            const uint64_t Alignment = Query.MMODescrs[0].AlignInBits;
+            if (MemoryType.getSizeInBits() >= 512 && Alignment == 256)
```
andcarminati: nit: `return (MemoryType.getSizeInBits() >= 512 && Alignment == 256);`

andcarminati: Should we also consider `MemoryType.getSizeInBits() % 256 == 0`, or at least assert?

llvm/lib/Target/AIE/aie2p/AIE2PLegalizerInfo.cpp
andcarminati: Note that, if some unaligned scalar load appears (let's say s512), divide will give us a valid scalar type, not replacing the type by a vector.

khallouh: Good point but I don't think `fewerElements` works with scalars. I can assert that we have a vector here. Otherwise, we will simply scalarize as we already do and currently, it will fail during the process because we can't legalize the resulting merge and unmerge opcodes.

llvm/lib/Target/AIE/aie2p/AIE2PLegalizerInfo.cpp
khallouh: We can't assert because we have other types above and it will break for them but we can return false instead

andcarminati (approved):
LGTM.

PR #322, opened 2025-01-30
[AIE2][AIE2P] Make i64 a legal register type

konstantinschwarz:
This is needed to be able to write TableGen selection patterns using this type.
We still want to pass 64-bit values as 2 x 32-bit GPRs in CallLowering.
This should be NFC, apart from making i64 available in TableGen patterns.

### DIFF
--- a/clang/test/CodeGen/aie/aie2p/aie2p-upd-ext-intrinsic.cpp
+++ b/clang/test/CodeGen/aie/aie2p/aie2p-upd-ext-intrinsic.cpp
@@ -5122,10 +5122,15 @@ v64accfloat test_set_v64accfloat (int idx, v16accfloat b)
 // CHECK-LABEL: define dso_local inreg noundef <64 x float> @_Z11test_concatDv16_u10__accfloatS_S_S_(
 // CHECK-SAME: <16 x float> inreg noundef [[A0:%.*]], <16 x float> inreg noundef [[A1:%.*]], <16 x float> inreg noundef [[A2:%.*]], <16 x float> inreg noundef [[A3:%.*]]) local_unnamed_addr #[[ATTR0]] {
 // CHECK-NEXT:  entry:
-// CHECK-NEXT:    [[TMP0:%.*]] = shufflevector <16 x float> [[A0]], <16 x float> [[A1]], <32 x i32> <i32 0, i32 1, i32 2, i32 3, i32 4, i32 5, i32 6, i32 7, i32 8, i32 9, i32 10, i32 11, i32 12, i32 13, i32 14, i32 15, i32 16, i32 17, i32 18, i32 19, i32 20, i32 21, i32 22, i32 23, i32 24, i32 25, i32 26, i32 27, i32 28, i32 29, i32 30, i32 31>
-// CHECK-NEXT:    [[TMP1:%.*]] = shufflevector <16 x float> [[A2]], <16 x float> [[A3]], <32 x i32> <i32 0, i32 1, i32 2, i32 3, i32 4, i32 5, i32 6, i32 7, i32 8, i32 9, i32 10, i32 11, i32 12, i32 13, i32 14, i32 15, i32 16, i32 17, i32 18, i32 19, i32 20, i32 21, i32 22, i32 23, i32 24, i32 25, i32 26, i32 27, i32 28, i32 29, i32 30, i32 31>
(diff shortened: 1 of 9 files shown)


clang/test/CodeGen/aie/aie2p/aie2p-upd-ext-intrinsic.cpp
```diff
@@ -5122,10 +5122,15 @@ v64accfloat test_set_v64accfloat (int idx, v16accfloat b)
...
-// CHECK-NEXT:    [[TMP1:%.*]] = shufflevector <16 x float> [[A2]], <16 x float> [[A3]], <32 x i32> <i32 0, i32 1, i32 2, i32 3, i32 4, i32 5, i32 6, i32 7, i32 8, i32 9, i32 10, i32 11, i32 12, i32 13, i32 14, i32 15, i32 16, i32 17, i32 18, i32 19, i32 20, i32 21, i32 22, i32 23, i32 24, i32 25, i32 26, i32 27, i32 28, i32 29, i32 30, i32 31>
-// CHECK-NEXT:    [[TMP2:%.*]] = shufflevector <32 x float> [[TMP0]], <32 x float> [[TMP1]], <64 x i32> <i32 0, i32 1, i32 2, i32 3, i32 4, i32 5, i32 6, i32 7, i32 8, i32 9, i32 10, i32 11, i32 12, i32 13, i32 14, i32 15, i32 16, i32 17, i32 18, i32 19, i32 20, i32 21, i32 22, i32 23, i32 24, i32 25, i32 26, i32 27, i32 28, i32 29, i32 30, i32 31, i32 32, i32 33, i32 34, i32 35, i32 36, i32 37, i32 38, i32 39, i32 40, i32 41, i32 42, i32 43, i32 44, i32 45, i32 46, i32 47, i32 48, i32 49, i32 50, i32 51, i32 52, i32 53, i32 54, i32 55, i32 56, i32 57, i32 58, i32 59, i32 60, i32 61, i32 62, i32 63>
-// CHECK-NEXT:    ret <64 x float> [[TMP2]]
+// CHECK-NEXT:    [[TMP0:%.*]] = bitcast <16 x float> [[A0]] to <8 x i64>
```
konstantinschwarz, line 5125: Adding `i64` as a legal register type has the side-effect of making it a legal integer type.
This in turn triggers (or rather prevents) a InstCombine optimization.
Independent of this change I started to rewrite our upd/ext intrinsics a while ago, to limit the amount of bitcasts we see everywhere.
I think this change just makes that work even more important. FYI @katerynamuts, this is what we discussed recently.

martien-de-jong (approved):

llvm/lib/Target/AIE/AIEBaseISelLowering.cpp
```diff
@@ -534,3 +534,28 @@ void AIEBaseTargetLowering::alignFirstVASlot(CCState &CCInfo) {
 MVT AIEBaseTargetLowering::getVectorIdxTy(const DataLayout &DL) const {
   return MVT::i32;
 }
+
+unsigned AIEBaseTargetLowering::getNumRegistersForCallingConv(
+    LLVMContext &Context, CallingConv::ID CC, EVT VT) const {
+  if (VT == MVT::i64 || VT == MVT::f64) {
```
khallouh, line 540: Why is the check for `f64` needed since you didn't make it legal?

clang/test/CodeGen/aie/aie2p/aie2p-upd-ext-intrinsic.cpp
khallouh: Funnily enough, These bitcasts result in better code in the end due to the better register bank selection. We still rely on the default `getPartialMappingIdx` which will map the resulting `G_CONCAT_VECTORS` to the accumulator bank saving us 4 cross-bank copies in the end.

khallouh:
```
// %bb.0:                               // %entry
	nopa	;		nopb	;		ret	lr;		nopm	;		nops	
	vmov	bmll0, bmll1                    //  Delay Slot 5
	vmov	bmlh0, bmll2                    //  Delay Slot 4
	vmov	bmhl0, bmll3                    //  Delay Slot 3
	vmov	bmhh0, bmll4                    //  Delay Slot 2
	nop	    
``` 
vs
```
// %bb.0:                               // %entry
	nopa	;		nopb	;		nops	;		nopx	;		vmov	x0, bmll1;		nopv	
	vmov	x1, bmll2
	vmov	x2, bmll3
	ret	lr;		vmov	x3, bmll4
	vmov	bmll0, x0                       //  Delay Slot 5
	vmov	bmlh0, x1                       //  Delay Slot 4
	vmov	bmhl0, x2                       //  Delay Slot 3
	vmov	bmhh0, x3                       //  Delay Slot 2
```

PR #326, opened 2025-01-31
[AIE2P] Combine G_SHUFFLE_VECTOR into ExtractSubvector+Broadcast

katerynamuts:


llvm/lib/Target/AIE/AIECombinerHelper.cpp
```diff
@@ -1853,19 +1765,18 @@ static std::optional<int> getUniqueIndex(ArrayRef<int> Mask) {
...
 ///         %2:_(<8 x s64>) = G_SHUFFLE_VECTOR %X(<4 x s64>), %1(<4 x s64>),
 ///         shufflemask(3, 3, 3, 3, 3, 3, 3, 3)
 /// To :    %2:_(<8 x s64>) = G_AIE_BROADCAST_VECTOR %X(<4 x s64>)
-bool llvm::matchShuffleToExtractBroadcast(MachineInstr &MI,
+static bool matchShuffleToVecEltBroadcast(MachineInstr &MI,
                                           MachineRegisterInfo &MRI,
                                           const AIEBaseInstrInfo &TII,
                                           BuildFnTy &MatchInfo) {
-  assert(MI.getOpcode() == TargetOpcode::G_SHUFFLE_VECTOR);
   ArrayRef<int> Mask = MI.getOperand(3).getShuffleMask();
```
F-Stuckmann, line 1827: why did you remove the assert?

llvm/lib/Target/AIE/AIECombinerHelper.cpp
```diff
@@ -1976,58 +1956,210 @@ bool llvm::matchShuffleToExtractSubvec(MachineInstr &MI,
...
+                                       MachineRegisterInfo &MRI,
+                                       const AIEBaseInstrInfo &TII,
+                                       BuildFnTy &MatchInfo) {
+  const Register DstReg = MI.getOperand(0).getReg();
+  const Register Src1Reg = MI.getOperand(1).getReg();
+  ArrayRef<int> Mask = MI.getOperand(3).getShuffleMask();
+
+  const LLT DstTy = MRI.getType(DstReg);
+  const LLT Src1Ty = MRI.getType(Src1Reg);
+  if (Src1Ty.getSizeInBits() != 64 && Src1Ty.getSizeInBits() != 32) {
```
F-Stuckmann, line 2118: shouldn't the there also be a `Src1Ty == DstTy` check?

llvm/lib/Target/AIE/AIECombinerHelper.cpp
katerynamuts: because this assert is already in `matchShuffleToBroadcast` which calls this method `matchShuffleToVecEltBroadcast`, so this assert would be just duplicated.

llvm/lib/Target/AIE/AIECombinerHelper.cpp
katerynamuts: I don't know, I didn't actually touch this part it was  before, the method was only called `matchShuffleToBroadcast`. Sometimes git diff is not helpful. :)

F-Stuckmann: but we also have it in its' sister function `matchShuffleToVSel`. It may make sense to also add it here

F-Stuckmann: How would extension work (zero extend or sign extend) ? I think the check here would help to not have this confusion

katerynamuts: It does not matter whether we choose the sign or zero-extending instruction, because there is nothing to extend for 32- and 64-bit, right?

F-Stuckmann: I think this function  here could take care of it `checkExtractSubvectorPrerequisites`.

It's sister functions like `matchShuffleToSubvecBroadcast` also have it

llvm/lib/Target/AIE/AIECombinerHelper.cpp
F-Stuckmann: but then it should also be removed in for example  `matchShuffleToVSel`.
I like to know, that there is an assert that will trip, if somebody calls the function wrong.

llvm/lib/Target/AIE/AIECombinerHelper.cpp
F-Stuckmann: Hardware wise going from 32 bit to 64 bit, you do have a choice if you zero extend (unsigned int) or sign extend (signed int). However, I do not know how AIE handles this

My thinking was: we could the following and the pattern matching would have to decide how to fill the MSBs.

`%0:_(<16 x s64>) = G_SHUFFLE_VECTOR %1(<2 x s32>), %2, shufflemask(0, 1, 0, 1, 0, 1, 0, 1, 0, 1, 0, 1, 0, 1, 0, 1)`

llvm/test/CodeGen/AIE/GlobalISel/prelegalizercombiner-shuffle-vector.mir
```diff
@@ -593,3 +593,84 @@ body:             |
...
+    ; CHECK-NEXT: [[AIE_UNPAD_VECTOR:%[0-9]+]]:_(<16 x s16>) = G_AIE_UNPAD_VECTOR [[AIE_BROADCAST_VECTOR]](<32 x s16>)
+    ; CHECK-NEXT: PseudoRET implicit $lr, implicit [[AIE_UNPAD_VECTOR]](<16 x s16>)
+    %1:_(<16 x s16>) = COPY $wl0
+    %2:_(<16 x s16>) = COPY $wl1
+    %0:_(<16 x s16>) = G_SHUFFLE_VECTOR %1(<16 x s16>), %2(<16 x s16>), shufflemask(4, 5, 6, 7, 4, 5, 6, 7, 4, 5, 6, 7, 4, 5, 6, 7)
+    PseudoRET implicit $lr, implicit %0
+...
+# The test is not covered by ExtractSubvecBroadcast because we can extract subvectors only of 32- or 64-bit sizes.
+---
+name: shuffle_vector_to_extract_broadcast_subvec_invalid_exctract_subvec_srcsize
```
andcarminati, line 770: I guess that this will be the next combiner. This pattern can be spotted in several benchmarks.

llvm/lib/Target/AIE/AIECombinerHelper.cpp
```diff
@@ -1904,6 +1815,94 @@ static bool checkSequentialMask(const ArrayRef<int> Mask, unsigned StartValue,
...
+  const unsigned VecRegSize = TII.getBasicVecRegSize();
+
+  const unsigned SrcTySize = SrcTy.getSizeInBits();
+
+  if (!DstTy.isVector() || !SrcTy.isVector() || SrcTySize < GPRSize ||
+      (DstTy.getSizeInBits() != GPRSize &&
+       DstTy.getSizeInBits() != 2 * GPRSize))
+    return false;
+
+  // Currently, we cannot concat vectors of the size less than vector register
```
andcarminati: concat or extract?

llvm/lib/Target/AIE/AIECombinerHelper.cpp
```diff
@@ -1904,6 +1815,94 @@ static bool checkSequentialMask(const ArrayRef<int> Mask, unsigned StartValue,
...
+  // size.
+  if (SrcTySize < VecRegSize)
+    return false;
+
+  //  This should be handled by a separate combine that copies SrcReg to
+  //  DstReg.
+  if (SrcTySize == DstTy.getSizeInBits())
```
andcarminati: nit:

```
return !(SrcTySize == DstTy.getSizeInBits());
```

llvm/lib/Target/AIE/AIECombinerHelper.cpp
katerynamuts: The check in `matchShuffleToVSel` is needed because it is not called from  `matchShuffleToBroadcast`.

llvm/lib/Target/AIE/AIECombinerHelper.cpp
katerynamuts: > I think this function here could take care of it `checkExtractSubvectorPrerequisites`.
> 
> It's sister functions like `matchShuffleToSubvecBroadcast` also have it

I can add this check also here

llvm/lib/Target/AIE/AIECombinerHelper.cpp
F-Stuckmann: got it, I missed that they were just local helper functions and not callable from outside.

llvm/test/CodeGen/AIE/GlobalISel/prelegalizercombiner-shuffle-vector.mir
katerynamuts: yes, I just added this test to show that it is still not covered by the current combines.

llvm/lib/Target/AIE/AIECombinerHelper.cpp
katerynamuts: extract :)

llvm/lib/Target/AIE/AIECombinerHelper.cpp
katerynamuts: I've tried to use also here `checkExtractSubvectorPrerequisites` but the check conditions in `checkExtractSubvectorPrerequisites` are much stronger than required in this method `matchShuffleToVecBroadcast` and then some tests that are expected to be covered by this method are just result in `G_SHUFFLE_VECTOR` without any combine. So I would leave the current `matchShuffleToVecBroadcast` method as it is .

llvm/lib/Target/AIE/AIECombinerHelper.cpp
```diff
@@ -1904,6 +1814,91 @@ static bool checkSequentialMask(const ArrayRef<int> Mask, unsigned StartValue,
   return true;
 }
 
+/// Check prerequisites to extract a subvector
+static bool checkExtractSubvectorPrerequisites(const AIEBaseInstrInfo &TII,
+                                               const LLT DstTy,
+                                               const LLT SrcTy) {
+  const unsigned ScalarRegSize = TII.getScalarRegSize();
+  const unsigned VecRegSize = TII.getBasicVecRegSize();
+
```
martien-de-jong: nit: drop newline?

llvm/lib/Target/AIE/AIECombinerHelper.cpp
```diff
@@ -1904,6 +1814,91 @@ static bool checkSequentialMask(const ArrayRef<int> Mask, unsigned StartValue,
...
+    return false;
+
+  // Currently, we cannot extract vectors of the size less than vector register
+  // size.
+  if (SrcTySize < VecRegSize)
+    return false;
+
+  //  This should be handled by a separate combine that copies SrcReg to
+  //  DstReg.
+  return !(SrcTySize == DstTy.getSizeInBits());
```
martien-de-jong: nit: use `!=` ?

llvm/lib/Target/AIE/AIECombinerHelper.cpp
```diff
@@ -1976,58 +1952,210 @@ bool llvm::matchShuffleToExtractSubvec(MachineInstr &MI,
...
+
+      // Create the sequential mask that covers NumSplatMaskLen
+      llvm::SmallVector<int, 16> SubvecMask =
+          llvm::createSequentialMask(Mask[0], NumSplatMaskLen, 0);
+      // Replicate the splat mask to cover the entire dst register
+      llvm::SmallVector<int, 16> SplatMask =
+          createDuplicatePatternMask(SubvecMask, NumDstElems / NumSplatMaskLen);
+
+      if (SplatMask.size() != Mask.size())
+        return std::nullopt;
```
martien-de-jong: Could we predict those sizes and return before creating them? Also note that this check is already part of isEquivalentMasks

llvm/lib/Target/AIE/AIECombinerHelper.cpp
```diff
@@ -1976,58 +1952,210 @@ bool llvm::matchShuffleToExtractSubvec(MachineInstr &MI,
...
+  const unsigned NumSrcElems = Src1Ty.getNumElements();
+  const unsigned NumElems = Mask.size();
+
+  auto GetSubvecMask = [=]() -> std::optional<llvm::SmallVector<int, 16>> {
+    // Find the splat mask pattern, start with length 2 and then power of 2.
+    for (unsigned NumSplatMaskLen = 2;
+         NumSplatMaskLen <= NumSrcElems && NumSplatMaskLen <= NumElems;
+         NumSplatMaskLen *= 2) {
+      if (Mask[0] % NumSplatMaskLen != 0)
+        return std::nullopt;
```
martien-de-jong, line 2042: nit: this will preclude a -1 in the first position

llvm/lib/Target/AIE/AIECombinerHelper.cpp
```diff
@@ -1976,58 +1952,210 @@ bool llvm::matchShuffleToExtractSubvec(MachineInstr &MI,
...
+
+      // Create the sequential mask that covers NumSplatMaskLen
+      llvm::SmallVector<int, 16> SubvecMask =
+          llvm::createSequentialMask(Mask[0], NumSplatMaskLen, 0);
```
martien-de-jong: nit: we miss the opportunity to use -1 if it is the same in all chunks.

llvm/lib/Target/AIE/AIECombinerHelper.cpp
```diff
@@ -1976,58 +1952,210 @@ bool llvm::matchShuffleToExtractSubvec(MachineInstr &MI,
...
+    buildBroadcastVector(B, MRI, Extract.getReg(0), DstReg);
+  };
+  return true;
+}
+
+/// Match something like this:
+///  %1:_(<2 x s32>) = COPY $l0
+///  %2:_(<2 x s32>) = G_IMPLICIT_DEF
+///  %0:_(<16 x s32>) = G_SHUFFLE_VECTOR %1(<2 x s32>), %2, shufflemask(0, 1, 0,
+///  1, 0, 1, 0, 1, 0, 1, 0, 1, 0, 1, 0, 1)
```
martien-de-jong: nit: put the mask on one line?

llvm/lib/Target/AIE/AIECombinerHelper.cpp
```diff
@@ -1976,58 +1952,210 @@ bool llvm::matchShuffleToExtractSubvec(MachineInstr &MI,
...
+  if (NumDstElems < NumSrcElems) {
+    return false;
+  }
+
+  // Create the mask that covers all elements of Src1
+  auto Src1Mask = llvm::createSequentialMask(0, NumSrcElems, 0);
+
+  // Replicate the Src1Mask to cover the entire dst register
+  auto SplatMask =
+      createDuplicatePatternMask(Src1Mask, NumDstElems / NumSrcElems);
```
martien-de-jong: I would suggest to have a mask matching function that calls a std::function bool check(I, Mask[I]); on each element of Mask, unless it is don't care. The createDuplicate() translates to a straightforward modulo on the index. Then we avoid creating all those explicit masks, and will fail really fast on masks that don't match for low indices.

llvm/lib/Target/AIE/AIECombinerHelper.cpp
```diff
@@ -1976,58 +1952,210 @@ bool llvm::matchShuffleToExtractSubvec(MachineInstr &MI,
...
+  for (unsigned I = 0; I < NumDstElems - 1; I++) {
+    if (Mask[I] == -1)
+      continue;
+
+    if (Mask[I] != SplatMask[I])
+      return false;
   }
```
martien-de-jong, line 2130: I'm also wondering why we wouldn't use isEquivalentMask() here.

llvm/lib/Target/AIE/AIECombinerHelper.cpp
```diff
@@ -1976,58 +1952,210 @@ bool llvm::matchShuffleToExtractSubvec(MachineInstr &MI,
...
+  const unsigned NumSrcElems = Src1Ty.getNumElements();
+  if (NumDstElems != NumSrcElems)
+    return false;
+
+  // Check that the shuffle mask can be converted into VSel mask:
+  // The mask contains only -1
+  if (std::all_of(Mask.begin(), Mask.end(),
+                  [&](int Value) { return Value == -1; })) {
+    return false;
+  }
```
martien-de-jong: This will also be caught by the loop below, resulting in a 0 VSEL mask. As such, the comment above is strange.

llvm/lib/Target/AIE/AIECombinerHelper.cpp
```diff
@@ -1904,6 +1814,91 @@ static bool checkSequentialMask(const ArrayRef<int> Mask, unsigned StartValue,
...
+                                               const LLT DstTy,
+                                               const LLT SrcTy) {
+  const unsigned ScalarRegSize = TII.getScalarRegSize();
+  const unsigned VecRegSize = TII.getBasicVecRegSize();
+
+  const unsigned SrcTySize = SrcTy.getSizeInBits();
+
+  if (!DstTy.isVector() || !SrcTy.isVector() || SrcTySize < ScalarRegSize ||
+      (DstTy.getSizeInBits() != ScalarRegSize &&
+       DstTy.getSizeInBits() != 2 * ScalarRegSize))
```
martien-de-jong: I guess delcaring DstTySize can be justified as well

llvm/lib/Target/AIE/AIECombinerHelper.cpp
martien-de-jong: Current use could just be covered by a generating function, int maskVal(int I).

llvm/lib/Target/AIE/AIECombinerHelper.cpp
katerynamuts: fixed

llvm/lib/Target/AIE/AIECombinerHelper.cpp
katerynamuts: Great idea! When I simplified it, then it became just `NumDstElems != MaskSize` and I could move it outside `GetSubvecMask`. I would rather leave the check in `isEquivalentMasks` for the case if anyone else calls `isEquivalentMasks` without checking the size of the masks.

llvm/lib/Target/AIE/AIECombinerHelper.cpp
katerynamuts: done

katerynamuts: I also simplified it in a similar way in `matchShuffleToSubvecBroadcast`

llvm/lib/Target/AIE/AIECombinerHelper.cpp
katerynamuts: Removed

llvm/lib/Target/AIE/AIECombinerHelper.cpp
katerynamuts: I've also added some tests for these cases

llvm/lib/Target/AIE/AIECombinerHelper.cpp
```diff
@@ -1976,58 +1951,204 @@ bool llvm::matchShuffleToExtractSubvec(MachineInstr &MI,
...
+  if (matchShuffleToVecBroadcast(MI, MRI, TII, MatchInfo))
+    return true;
+  if (matchShuffleToVecEltBroadcast(MI, MRI, TII, MatchInfo))
+    return true;
+  if (matchShuffleToSubvecBroadcast(MI, MRI, TII, MatchInfo))
+    return true;
+  return false;
+}
+
+bool llvm::matchShuffleToVSel(MachineInstr &MI, MachineRegisterInfo &MRI,
```
andcarminati: Is there a way to minimize the diffs? Some functions are appearing as new but they were just moved and updated with few lines. Maybe if we keep the original order it can facilitate the review.

katerynamuts: I was really trying to do this but some functions are local, so they have to be defined before they are used and I had to move them up.

llvm/lib/Target/AIE/AIECombinerHelper.cpp
```diff
@@ -1976,58 +1951,204 @@ bool llvm::matchShuffleToExtractSubvec(MachineInstr &MI,
...
+         SplatMaskLen <= NumSrcElems && SplatMaskLen <= MaskSize;
+         SplatMaskLen *= 2) {
+      if (Mask[0] != -1 && Mask[0] % SplatMaskLen != 0)
+        return std::nullopt;
+
+      // Find the start value of the splat mask
+      int SplatMaskStart = -1;
+      for (int MaskElem : Mask) {
+        if (MaskElem != -1) {
+          SplatMaskStart = MaskElem - SplatMaskStart - 1;
```
martien-de-jong: I can't get my head around this. I think you're trying to find the '4' mask value from the example in the header even if it is -1. I *think* that value could derive from any non(-1) value in the validity check below, provided that we derive the validity from the SplatMaskLen, not from SplatMaskStart.
What makes this loop here difficult to read is that there's no clear invariant related to SplatMaskStart.

martien-de-jong: Ultimately, we are looking for a nice sawtooth function the period and amplitude of which are defined by SplatMaskLen and the height of which should be an an integer multiple of SplatMaskLen. Each index defines a phase of Idx % SplatMaskLen, and hence the first non-(-1) value fully defines height by dividing by SplatMaskLen.

llvm/lib/Target/AIE/AIECombinerHelper.cpp
```diff
@@ -1976,58 +1951,204 @@ bool llvm::matchShuffleToExtractSubvec(MachineInstr &MI,
...
+    buildBroadcastVector(B, MRI, Extract.getReg(0), DstReg);
+  };
+  return true;
+}
+
+/// Match something like this:
+///  %1:_(<2 x s32>) = COPY $l0
+///  %2:_(<2 x s32>) = G_IMPLICIT_DEF
+///  %0:_(<16 x s32>) = G_SHUFFLE_VECTOR %1(<2 x s32>), %2,
+///  shufflemask(0, 1, 0, 1, 0, 1, 0, 1, 0, 1, 0, 1, 0, 1, 0, 1)
```
martien-de-jong, line 2057: I think this also fits the pattern from `matchShuffleToSubvecBroadcast`.
How do we distinguish?

llvm/lib/Target/AIE/AIECombinerHelper.cpp
martien-de-jong: The generating function would be` F(Idx) = Height + Idx % SplatMaskLen`, or `Height = F(Idx) - Idx % SplatMaskLen`;
At the first non (-1) value we compute Height, checking that `Height % SplatMaskLen == 0` All further values should be -1 or match the now complete function definition.

llvm/lib/Target/AIE/AIECombinerHelper.cpp
```diff
@@ -1976,58 +1951,204 @@ bool llvm::matchShuffleToExtractSubvec(MachineInstr &MI,
...
+      DstMask |= uint64_t(1) << I;
+    else
+      return false;
+  }
+
+  if (DstMask == 0)
+    return false;
+
+  MatchInfo = [=, &TII](MachineIRBuilder &B) {
+    MachineInstrBuilder MaskReg = B.buildConstant(LLT::scalar(64), DstMask);
```
konstantinschwarz: This is changed to `LLT::scalar(64)` for which we have no instruction selection patterns.
Why do you have to change this to 64-bit constants?
How do we handle `<64 x s8>` selects? Do we have prelegalizercombiner tests for this type?

katerynamuts: oh you're right, I changed it back! But then I think we should change the definition of `DstMask` from `uint64_t DstMask = 0;` to `uint32_t DstMask = 0;`

llvm/lib/Target/AIE/AIECombinerHelper.cpp
katerynamuts: I changed it according to your suggestion. There is a separate commit which to be squashed.

martien-de-jong (dismissed):

llvm/lib/Target/AIE/AIECombinerHelper.cpp
katerynamuts: Fixed in PR https://github.com/Xilinx/llvm-aie/pull/338

martien-de-jong (approved):

PR #327, opened 2025-01-31
[AIE2P] Alias Analysis

F-Stuckmann:
Added Alias Analysis for AIE2p.


llvm/test/CodeGen/AIE/alias-analysis-fifo-vst-vld-scheduling.mir
```diff
@@ -0,0 +1,214 @@
+# NOTE: Assertions have been autogenerated by utils/update_llc_test_checks.py
+#
+# This file is licensed under the Apache License v2.0 with LLVM Exceptions.
+# See https://llvm.org/LICENSE.txt for license information.
+# SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+#
+# (c) Copyright 2025 Advanced Micro Devices, Inc. or its affiliates
+# RUN: llc -mtriple=aie2p --start-after=machine-scheduler -o -  %s | FileCheck %s
+
+# no alias conflicts between load and stores
```
F-Stuckmann: Note: For the unit test to properly show the parallel scheduling of the load and stores, MMOs are needed.

llvm/lib/Target/AIE/AIEBaseAliasAnalysis.cpp
```diff
@@ -75,6 +76,32 @@ static bool isAIEPtrAddIntrinsic(Intrinsic::ID ID, unsigned &InPtrIdx) {
   switch (ID) {
   case Intrinsic::aie2_add_2d:
   case Intrinsic::aie2_add_3d:
+  case Intrinsic::aie2p_add_2d:
```
andcarminati, line 79: I guess we need just `isAIEPtrAddIntrinsic`.

andcarminati: Plus: aie2p_add_3d and aie2p_add_2d for the other cases.

The next two functions are intended to cover loops with load and stores on the same base pointer. It is not clear that this will be the case.

F-Stuckmann: but then we would also have to change the return type of `getAddIntrinsicOps` and `getAddIntrinsicCounterOps`.
Additionally it would complicate the the handling of the functions.

```
for (unsigned Op : getAddIntrinsicCounterOps(ID)) {
    Args.CounterArgs.push_back(IntrinsicCall->getArgOperand(Op));
  }
```
It now just gets jumped over since we return an empty Vector. I like this elegant solution

andcarminati: Oh, now I got your idea. Nice, we can go with it.

llvm/test/CodeGen/AIE/alias-analysis-fifo-vst-vld-scheduling.mir
martien-de-jong: Apart from that, I think it would make sense to test a fifo load against a normal store and a fifo store against a normal load. You can use a long load latency to give the scheduler a good reason to move it up past a fifo store. They don't need to run in parallel, they only need to be independent.

llvm/test/CodeGen/AIE/alias-analysis-pointer-disambiguation-aie2p.ll
```diff
@@ -0,0 +1,669 @@
+;
+; This file is licensed under the Apache License v2.0 with LLVM Exceptions.
+; See https://llvm.org/LICENSE.txt for license information.
+; SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+;
+; (c) Copyright 2025 Advanced Micro Devices, Inc. or its affiliates
+; RUN: opt -mtriple=aie2p -passes=aa-eval -print-all-alias-modref-info -disable-output < %s 2>&1 | FileCheck %s --check-prefix=ENABLED-AIE-AA-DIS
+; RUN: opt -mtriple=aie2p -passes=aa-eval -print-all-alias-modref-info -disable-output \
+; RUN:    --aie-alias-analysis-disambiguation=false < %s 2>&1 | FileCheck %s --check-prefix=DISABLED-AIE-AA-DES
+
```
andcarminati: Is is possible to include a test that shows that we are conservative for fifo operations using the same pointer?

llvm/test/CodeGen/AIE/alias-analysis-fifo-vst-vld-scheduling.mir
```diff
@@ -0,0 +1,214 @@
...
+
+--- |
+  ; ModuleID = 'llvm/test/CodeGen/AIE/alias-analysis-fifo-vst-vld-scheduling.ll'
```
andcarminati: nit: remove this MuduleID.

llvm/test/CodeGen/AIE/alias-analysis-fifo-vst-vld-scheduling.mir
andcarminati: I think we need to check with opt and `-print-all-alias-modref-info`. Scheduling will be a very indirect way to test.

llvm/lib/Target/AIE/AIEBaseAliasAnalysis.cpp
```diff
@@ -93,6 +120,36 @@ static SmallVector<unsigned, 5> getAddIntrinsicOps(Intrinsic::ID ID) {
   case Intrinsic::aie2_add_3d:
     return {1, 2, 3, 4, /*,count1*/ 6,
             /*,count2*/};
+  case Intrinsic::aie2p_add_2d:
+    return {1, 2, 3 /*,count*/};
+  case Intrinsic::aie2p_add_3d:
+    return {1, 2, 3, 4, /*,count1*/ 6,
+            /*,count2*/};
```
khallouh, line 107: Why are count1 and count2 mentioned in the comment if this function is supposed to exclude counters?

llvm/lib/Target/AIE/AIEBaseAliasAnalysis.cpp
```diff
@@ -93,6 +120,36 @@ static SmallVector<unsigned, 5> getAddIntrinsicOps(Intrinsic::ID ID) {
...
+  case Intrinsic::aie2p_fifo_st_flush_1d_conv:
+  case Intrinsic::aie2p_fifo_st_flush_2d:
+  case Intrinsic::aie2p_fifo_st_flush_2d_conv:
+  case Intrinsic::aie2p_fifo_st_flush_3d:
+  case Intrinsic::aie2p_fifo_st_flush_3d_conv:
+  case Intrinsic::aie2p_fifo_st_flush_conv:
+  case Intrinsic::aie2p_fifo_st_push_512_bfp16:
+  case Intrinsic::aie2p_fifo_st_push_544_bfp16:
+  case Intrinsic::aie2p_fifo_st_push_576_bfp16:
+    return {};
```
khallouh: Why is return empty for all fifo intrinsics? Aren't there also operands that change the output pointer here as well?

llvm/lib/Target/AIE/AIEBaseAliasAnalysis.cpp
```diff
@@ -107,6 +164,35 @@ static SmallVector<unsigned, 5> getAddIntrinsicCounterOps(Intrinsic::ID ID) {
...
```
khallouh: Same as for `getAddIntrinsicOps` Aren't there counter operands in those intrinsics?

llvm/test/CodeGen/AIE/alias-analysis-pointer-disambiguation-aie2p.ll
```diff
@@ -0,0 +1,669 @@
+;
+; This file is licensed under the Apache License v2.0 with LLVM Exceptions.
+; See https://llvm.org/LICENSE.txt for license information.
+; SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+;
+; (c) Copyright 2025 Advanced Micro Devices, Inc. or its affiliates
+; RUN: opt -mtriple=aie2p -passes=aa-eval -print-all-alias-modref-info -disable-output < %s 2>&1 | FileCheck %s --check-prefix=ENABLED-AIE-AA-DIS
+; RUN: opt -mtriple=aie2p -passes=aa-eval -print-all-alias-modref-info -disable-output \
+; RUN:    --aie-alias-analysis-disambiguation=false < %s 2>&1 | FileCheck %s --check-prefix=DISABLED-AIE-AA-DES
```
khallouh: Could you move this test under the aie2p folder? and the original test for aie2 under aie2p folder as well?

khallouh:
How about extending `getUnderlyingObjectAIEIntrinsic` in `ValueTracking.cpp`? Is it not needed with our pass anymore?

llvm/lib/Target/AIE/AIEBaseAliasAnalysis.cpp
F-Stuckmann: `count1` in `aie2p_add_3d` would be at index 5, which is excluded.

It is just a nice way to visualize if we miss an index.

llvm/test/CodeGen/AIE/alias-analysis-pointer-disambiguation-aie2p.ll
F-Stuckmann: yes, I'll add it together with the fifo load/stores and non-fifo load store test

llvm/test/CodeGen/AIE/alias-analysis-fifo-vst-vld-scheduling.mir
F-Stuckmann: I added fifo and non-fifo load stores to the unit tests.
Additionally I will change unit-test to a opt unittest, that directly checks the alias info.

llvm/test/CodeGen/AIE/alias-analysis-pointer-disambiguation-aie2p.ll
```diff
@@ -0,0 +1,669 @@
...
+; ENABLED-AIE-AA-DIS:  NoAlias:      i8* %10, i8* %3
+; ENABLED-AIE-AA-DIS:  NoAlias:      i8* %10, i8* %7
+; ENABLED-AIE-AA-DIS:  NoAlias:      i8* %14, i8* %ptr_load
+; ENABLED-AIE-AA-DIS:  NoAlias:      i8* %14, i8* %ptr_store
+; ENABLED-AIE-AA-DIS:  NoAlias:      i8* %14, i8* %3
+; ENABLED-AIE-AA-DIS:  NoAlias:      i8* %14, i8* %7
+; ENABLED-AIE-AA-DIS:  MayAlias:     i8* %10, i8* %14
+
+; DISABLED-AIE-AA-DES-LABEL: Function: test
+; DISABLED-AIE-AA-DES:  MayAlias:      i8* %ptr_load, i8* %ptr_store
```
khallouh: Curious, was this meant to be called `DISABLED-AIE-AA-DIS` like `ENABLED-AIE-AA-DIS` or does `DES` mean something else other than "disambiguate"?

llvm/lib/Target/AIE/AIEBaseAliasAnalysis.cpp
F-Stuckmann: No we aren't using the counter operands, because the intrinsics don't have any. We would have to do our own Address-tracking logic of how much the pointer was actually changed, which is tricky and not the scope of this PR.

The empty return values (in `getAddIntrinsicOps` and `getAddIntrinsicCounterOps`)  are just a nicer way opting-out of the counter following mechanism, see

https://github.com/Xilinx/llvm-aie/pull/327#discussion_r1937372405

llvm/lib/Target/AIE/AIEBaseAliasAnalysis.cpp
F-Stuckmann: see https://github.com/Xilinx/llvm-aie/pull/327#discussion_r1939334888

llvm/test/CodeGen/AIE/alias-analysis-pointer-disambiguation-aie2p.ll
F-Stuckmann: yeah

llvm/test/CodeGen/AIE/alias-analysis-pointer-disambiguation-aie2p.ll
F-Stuckmann: I find it easier to visually differentiate the two cases, if the difference is closer to where i actually look (may alias/ptr etc)

F-Stuckmann:
> How about extending `getUnderlyingObjectAIEIntrinsic` in `ValueTracking.cpp`? Is it not needed with our pass anymore?

you are right, we may default to performing Alias Analysis with Valuetracking in some cases. I will add it

llvm/lib/Target/AIE/AIEBaseAliasAnalysis.cpp
```diff
@@ -75,6 +76,32 @@ static bool isAIEPtrAddIntrinsic(Intrinsic::ID ID, unsigned &InPtrIdx) {
...
+  case Intrinsic::aie2p_fifo_st_flush_2d:
+  case Intrinsic::aie2p_fifo_st_flush_2d_conv:
+  case Intrinsic::aie2p_fifo_st_flush_3d:
+  case Intrinsic::aie2p_fifo_st_flush_3d_conv:
+  case Intrinsic::aie2p_fifo_st_flush_conv:
+  case Intrinsic::aie2p_fifo_st_push_512_bfp16:
+  case Intrinsic::aie2p_fifo_st_push_544_bfp16:
+  case Intrinsic::aie2p_fifo_st_push_576_bfp16:
     InPtrIdx = 0;
     return true;
```
martien-de-jong, line 86: Would there be a way to share this list / operand number with ValueTracking?

F-Stuckmann: I don't think we should add `AIE Backend` dependencies to the `LLVMAnalysis`. That will probably break a lot of unit tests. The other backends also just go through the intrinsics header files

llvm/test/CodeGen/AIE/aie2p/alias-analysis-fifo-vst-vld-scheduling.ll
```diff
@@ -0,0 +1,248 @@
+;
```
khallouh: Why is the test called scheduling when it runs alias analysis?

llvm/test/CodeGen/AIE/aie2p/alias-analysis-fifo-vst-vld-scheduling.ll
```diff
@@ -0,0 +1,248 @@
...
+declare { ptr, <32 x i32>, i32 } @llvm.aie2p.fifo.ld.fill(ptr, <32 x i32>, i32) #1
+
+; Function Attrs: nounwind memory(none)
+declare <32 x float> @llvm.aie2p.v32bf16.to.v32accfloat(<32 x bfloat>) #2
+
+; Function Attrs: nounwind memory(argmem: read)
+declare { <64 x i8>, ptr, <32 x i32>, i32 } @llvm.aie2p.fifo.ld.pop.unaligned(ptr, <32 x i32>, i32) #1
+
+; Function Attrs: nounwind memory(argmem: read)
+declare { <64 x i8>, ptr, <32 x i32>, i32, i20 } @llvm.aie2p.fifo.ld.pop.2d.unaligned(ptr, <32 x i32>, i32, i20, i20, i20, i20) #1
```
khallouh: Could you test the other intrinsics? namely pop.1d, pop.3d, flush.3d, etc

llvm/test/CodeGen/AIE/aie2p/alias-analysis-fifo-vst-vld-scheduling.ll
F-Stuckmann: yes the name is out-of-date. I changed it

llvm/test/CodeGen/AIE/aie2p/alias-analysis-fifo-vst-vld-scheduling.ll
```diff
@@ -0,0 +1,248 @@
...
+  %2 = extractvalue { ptr, <32 x i32>, i32 } %0, 1
+  %3 = extractvalue { ptr, <32 x i32>, i32 } %0, 2
+  %4 = call { ptr, <32 x i32>, i32 } @llvm.aie2p.fifo.st.flush(ptr %1, <32 x i32> %2, i32 %3)
+  %5 = extractvalue { ptr, <32 x i32>, i32 } %4, 0
+  %6 = extractvalue { ptr, <32 x i32>, i32 } %4, 1
+  %7 = extractvalue { ptr, <32 x i32>, i32 } %4, 2
+  ret void
+}
+
+; verify that regular load/store do not alias with fifo load/store
```
andcarminati: Under the assumption that we have noalias.

llvm/test/CodeGen/AIE/aie2p/alias-analysis-fifo-vst-vld-scheduling.ll
```diff
@@ -0,0 +1,248 @@
+;
+; This file is licensed under the Apache License v2.0 with LLVM Exceptions.
+; See https://llvm.org/LICENSE.txt for license information.
+; SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+;
+; (c) Copyright 2025 Advanced Micro Devices, Inc. or its affiliates
+; RUN: opt -mtriple=aie2p -print-all-alias-modref-info -disable-output -passes=aa-eval < %s 2>&1  | FileCheck %s
+
+; no alias conflicts between fifo load and fifo stores
```
andcarminati: Maybe write a bit more about the constraints that we have here. Like which operations use which pointer and its attributes.

llvm/test/CodeGen/AIE/aie2p/alias-analysis-fifo-vst-vld-scheduling.ll
F-Stuckmann: I added more comments to describe what the test is supposed to do

llvm/test/CodeGen/AIE/aie2p/alias-analysis-fifo-vst-vld-scheduling.ll
F-Stuckmann: I added more comments

llvm/test/CodeGen/AIE/aie2p/alias-analysis-fifo-vst-vld-scheduling.ll
F-Stuckmann: I updated coverage unittests for the intrinsics

llvm/lib/Analysis/ValueTracking.cpp
```diff
@@ -6303,6 +6303,29 @@ static const Value *getUnderlyingObjectAIEIntrinsic(const Value *V) {
...
+      case Intrinsic::aie2p_fifo_ld_pop_576_bfp16:
+      case Intrinsic::aie2p_fifo_ld_pop_1d_unaligned:
+      case Intrinsic::aie2p_fifo_ld_pop_544_1d_bfp16:
+      case Intrinsic::aie2p_fifo_ld_pop_576_1d_bfp16:
+      case Intrinsic::aie2p_fifo_ld_pop_2d_unaligned:
+      case Intrinsic::aie2p_fifo_ld_pop_3d_unaligned:
+      case Intrinsic::aie2p_fifo_ld_pop_544_2d_bfp16:
+      case Intrinsic::aie2p_fifo_ld_pop_576_2d_bfp16:
+      case Intrinsic::aie2p_fifo_ld_pop_544_3d_bfp16:
+      case Intrinsic::aie2p_fifo_ld_pop_576_3d_bfp16:
```
gbossu, line 6321: Super-Nit: The ordering is quite confusing and it's difficult to check if all intrinsics are present.

llvm/lib/Analysis/ValueTracking.cpp
```diff
@@ -6303,6 +6303,29 @@ static const Value *getUnderlyingObjectAIEIntrinsic(const Value *V) {
...
+      case Intrinsic::aie2p_fifo_st_flush_1d:
+      case Intrinsic::aie2p_fifo_st_flush_1d_conv:
+      case Intrinsic::aie2p_fifo_st_flush_2d:
+      case Intrinsic::aie2p_fifo_st_flush_2d_conv:
+      case Intrinsic::aie2p_fifo_st_flush_3d:
+      case Intrinsic::aie2p_fifo_st_flush_3d_conv:
+      case Intrinsic::aie2p_fifo_st_flush_conv:
+      case Intrinsic::aie2p_fifo_st_push_512_bfp16:
+      case Intrinsic::aie2p_fifo_st_push_544_bfp16:
+      case Intrinsic::aie2p_fifo_st_push_576_bfp16:
```
gbossu, line 6335: If think this is missing ld_fill

llvm/lib/Target/AIE/AIEBaseAliasAnalysis.cpp
```diff
@@ -75,6 +76,32 @@ static bool isAIEPtrAddIntrinsic(Intrinsic::ID ID, unsigned &InPtrIdx) {
...
+  case Intrinsic::aie2p_fifo_st_flush_1d:
+  case Intrinsic::aie2p_fifo_st_flush_1d_conv:
+  case Intrinsic::aie2p_fifo_st_flush_2d:
+  case Intrinsic::aie2p_fifo_st_flush_2d_conv:
+  case Intrinsic::aie2p_fifo_st_flush_3d:
+  case Intrinsic::aie2p_fifo_st_flush_3d_conv:
+  case Intrinsic::aie2p_fifo_st_flush_conv:
+  case Intrinsic::aie2p_fifo_st_push_512_bfp16:
+  case Intrinsic::aie2p_fifo_st_push_544_bfp16:
+  case Intrinsic::aie2p_fifo_st_push_576_bfp16:
```
gbossu: Did you check whether that is useful? I believe this is old code that is now superseded by the one in `ValueTracking.cpp`

llvm/lib/Target/AIE/AIEBaseAliasAnalysis.cpp
```diff
@@ -93,6 +120,37 @@ static SmallVector<unsigned, 5> getAddIntrinsicOps(Intrinsic::ID ID) {
...
+    return {};
```
gbossu: Discussed offline: We should only care about 2d/3d addressing here.

llvm/test/CodeGen/AIE/aie2/alias-analysis-pointer-disambiguation.ll
```diff
@@ -0,0 +1,669 @@
+;
+; This file is licensed under the Apache License v2.0 with LLVM Exceptions.
+; See https://llvm.org/LICENSE.txt for license information.
+; SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+;
+; (c) Copyright 2024 Advanced Micro Devices, Inc. or its affiliates
+; RUN: opt -mtriple=aie2 -passes=aa-eval -print-all-alias-modref-info -disable-output < %s 2>&1 | FileCheck %s --check-prefix=AIE-AA-DIS-ENABLE
+; RUN: opt -mtriple=aie2 -passes=aa-eval -print-all-alias-modref-info -disable-output \
+; RUN:    --aie-alias-analysis-disambiguation=false < %s 2>&1 | FileCheck %s --check-prefix=AIE-AA-DIS-DISABLE
```
gbossu, line 9: Nit: I would only run with `--aie-alias-analysis-disambiguation=true`, that file is big enough already :)

llvm/test/CodeGen/AIE/aie2p/alias-analysis-fifo-vst-vld.ll
```diff
@@ -0,0 +1,1541 @@
+;
+; This file is licensed under the Apache License v2.0 with LLVM Exceptions.
+; See https://llvm.org/LICENSE.txt for license information.
+; SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+;
+; (c) Copyright 2025 Advanced Micro Devices, Inc. or its affiliates
+; RUN: opt -mtriple=aie2p -print-all-alias-modref-info -disable-output -passes=aa-eval < %s 2>&1  | FileCheck %s
+
+; Verify no alias conflicts between fifo load and fifo stores, when load and 
+; store use different (noalias) pointers
```
gbossu: Could you think of a way to trim that file? There are really lots of CHECK lines and functions to read through.

I also do not see the important CHECK lines with `NoAlias` in some functions. I think those are the only ones we really care about. `NoModRef`/`Just Mod` is not really what we want to test here.

F-Stuckmann: When we have function calls (i.e. Intrinsics), we don't get the `NoAlias/MayAlias` output. Instead we get `NoModRef/Just Mod/Just Ref/Both ModRef`

llvm/lib/Analysis/ValueTracking.cpp
F-Stuckmann: I was going for a 1D, 2D, 3D ordering, but I suppose I could also order them According to bitwidth/functionality

llvm/lib/Analysis/ValueTracking.cpp
F-Stuckmann: yes, I missed that one

llvm/lib/Target/AIE/AIEBaseAliasAnalysis.cpp
F-Stuckmann: Funnily enough, either location works, but it's cleaner to only have the non-2d/3d operations in `ValueTracking,cpp`

llvm/test/CodeGen/AIE/aie2/alias-analysis-pointer-disambiguation.ll
F-Stuckmann: I just moved this file from `llvm/test/CodeGen/AIE/` into the `aie2`, so I'll leave it here. 
I'll use the `--aie-alias-analysis-disambiguation=true` on the newly generated file in `aie2p` and the unit test for the fifo load/stores

llvm/test/CodeGen/AIE/aie2p/alias-analysis-fifo-vst-vld.ll
F-Stuckmann: I am using `--aie-alias-analysis-disambiguation=true` now to summarize the unit test results

llvm/lib/Target/AIE/AIEBaseAliasAnalysis.cpp
gbossu: Discussed offline: `ValueTracking.cpp` is the preferred place for tracking a pointer through intrinsics.

AIEBaseAliasAnalysis.cpp also does some tracking, but that should be deprecated and removed. It should only be used to disambiguate addresses based on the same pointer when 2d/3d increments are involved. For FIFO ld/st, we cannot even do that, because the internal address used by the HW for loading/storing cannot be easily known.

llvm/test/CodeGen/AIE/aie2p/alias-analysis-fifo-vst-vld.ll
```diff
@@ -0,0 +1,174 @@
...
+; (c) Copyright 2025 Advanced Micro Devices, Inc. or its affiliates
+; RUN: opt -mtriple=aie2p -print-all-alias-modref-info -disable-output -passes=aa-eval < %s 2>&1  | FileCheck %s
+
+
+
+; Coverage: verify AA for every single fifo load/store intrinsic, under 
+; the assumption that load and store operate on different (noalias) pointers
+
+; CHECK-LABEL: coverageTestConcurrentLoad
+; CHECK: NoAlias:      i32* %L.0, i32* %S.0
```
gbossu, line 15: Nice!

gbossu (approved):
LGTM

PR #331, opened 2025-02-04
MachineMemOperands and bank annotations for FIFO ld/st

gbossu:
Better review commit by commit. Many details were discussed offline, but there are some details in the commits :)

The interesting test updates for addrspace related changes are in `clang/test/CodeGen/aie/aie2p/aie2p-bank-annotation.cpp`

> Without these changes, we lose the bank annotations pretty much immediately because our intrinsics only support the default addrspace, and there seems to be super conservative type checks on clang builtins that prevent us to pass references to pointers in other address spaces. The final IR is also pretty bad and SROA will not manage to eliminate alloca instructions for bank-annotated pointers that are used by fifo intrinsics.
>
> Here's my solution:
> - Change all intrinsics signatures so that they take and return a llvm_anyptr_ty instead of llvm_ptr_ty. This means we will now need to specify the right types when creating the intrinsics. We would now have e.g. `call { ptr addrspace(5), <32 x i32>, i32 } @llvm.aie2p.fifo.st.push.576.bfp16.p5.p5(ptr addrspace(5) %0, <64 x i8> %v, <8 x i8> %e, <32 x i32> %fifo, i32 %avail)` instead of `call { ptr, <32 x i32>, i32 } @llvm.aie2p.fifo.st.push.576.bfp16(ptr %0, <64 x i8> %v, <8 x i8> %e, <32 x i32> %fifo, i32 %avail)` . There are slight changes in the clang->IR layer to do to get there, but that's it, ISel remains identical.
> - Change builtins to accept bank-qualified pointer refences. I tried many things, but I could essentially only get it to work when disabling type-checking for these builtins. This is also what AArch64 and other targets do for builtins that are meant to work for multiple address spaces. This means we can now pass a `v64bfp16ebs8_unaligned __aie_dm_resource_a *` pointer to e.g. `fifo_st_push`, and the addrspace will remain. There will be no addrspacecast .


clang/lib/CodeGen/CGBuiltin.cpp
```diff
@@ -22153,6 +22153,16 @@ Value *CodeGenFunction::EmitAIE1BuiltinExpr(unsigned BuiltinID,
...
 
+void insertImplicitCasts(MutableArrayRef<llvm::Value *> Args,
+                         llvm::FunctionType &FTy, CGBuilderTy &Builder) {
+  for (unsigned ArgNum = 0; ArgNum < Args.size(); ++ArgNum) {
+    llvm::Value *Arg = Args[ArgNum];
+    llvm::Type *RequiredType = FTy.getParamType(ArgNum);
+    if (Arg->getType() != RequiredType)
+      Args[ArgNum] = Builder.CreateBitCast(Arg, RequiredType);
+  }
+}
```
gbossu, line 22164: Note: We now need to manually insert bitcasts because type checks are disabled.

clang/include/clang/Basic/BuiltinsAIE2P.def
```diff
@@ -323,29 +323,31 @@ BUILTIN(__builtin_aie2p_divstep, "vUi&Ui&Ui", "nc")
...
-BUILTIN(__builtin_aie2p_fifo_ld_pop_576_bfp16, "vi*&V32i&i&V64c&V8c&", "nc")
-BUILTIN(__builtin_aie2p_fifo_ld_pop_1d_512_unaligned, "V64ci*&V32i&i&i", "nc")
-BUILTIN(__builtin_aie2p_fifo_ld_pop_1d_544_bfp16, "vi*&V32i&i&iV64c&V8c&", "nc")
-BUILTIN(__builtin_aie2p_fifo_ld_pop_1d_576_bfp16, "vi*&V32i&i&iV64c&V8c&", "nc")
-BUILTIN(__builtin_aie2p_fifo_ld_pop_2d_512_unaligned, "V64ci*&V32i&i&iii&i", "nc")
-BUILTIN(__builtin_aie2p_fifo_ld_pop_2d_544_bfp16, "vi*&V32i&i&iii&iV64c&V8c&", "nc")
-BUILTIN(__builtin_aie2p_fifo_ld_pop_2d_576_bfp16, "vi*&V32i&i&iii&iV64c&V8c&", "nc")
-BUILTIN(__builtin_aie2p_fifo_ld_pop_3d_512_unaligned, "V64ci*&V32i&i&iii&iii&i", "nc")
-BUILTIN(__builtin_aie2p_fifo_ld_pop_3d_544_bfp16, "vi*&V32i&i&iii&iii&iV64c&V8c&", "nc")
-BUILTIN(__builtin_aie2p_fifo_ld_pop_3d_576_bfp16, "vi*&V32i&i&iii&iii&iV64c&V8c&", "nc")
```
gbossu, line 351: Note: two changes here:
 - Change the `int *` into a `void *` which I believe is less confusing
 - use `"t"` option to disable type checks

llvm/lib/CodeGen/GlobalISel/IRTranslator.cpp
```diff
@@ -2761,9 +2761,12 @@ bool IRTranslator::translateCall(const User &U, MachineIRBuilder &MIRBuilder) {
   TargetLowering::IntrinsicInfo Info;
   // TODO: Add a GlobalISel version of getTgtMemIntrinsic.
   if (TLI->getTgtMemIntrinsic(Info, CI, *MF, ID)) {
-    Align Alignment = Info.align.value_or(
-        DL->getABITypeAlign(Info.memVT.getTypeForEVT(F->getContext())));
-    LLT MemTy = Info.memVT.isSimple()
+    Align Alignment =
+        Info.align.has_value()
+            ? *Info.align
+            : DL->getABITypeAlign(Info.memVT.getTypeForEVT(F->getContext()));
```
martien-de-jong, line 2767: This is equivalent to what it was, no?

llvm/lib/Target/AIE/aie2p/AIE2PISelLowering.cpp
```diff
@@ -73,3 +74,55 @@ bool AIE2PTargetLowering::functionArgumentNeedsConsecutiveRegisters(
...
+  case Intrinsic::aie2p_fifo_ld_pop_576_2d_bfp16:
+  case Intrinsic::aie2p_fifo_ld_pop_576_3d_bfp16:
+  case Intrinsic::aie2p_fifo_ld_pop_576_bfp16:
+    // The HW does a 512-bit load from somewhere between addr-63 and addr+128
+    // depending on the FIFO availability and the input alignment.
+    // A conservative access range would be [addr-64, addr+192)
+    // To be safe, we prefer "unknown-size".
+    Info.memVT = MVT::Other;
+    Info.ptrVal = I.getArgOperand(0);
+    Info.align = Align(1); // Can we somehow recover the original alignment?
```
martien-de-jong, line 102: I don't know whether it makes a difference, but I suppose the underlying load is vector-aligned?

llvm/lib/Target/AIE/aie2p/AIE2PISelLowering.h
```diff
@@ -31,6 +31,10 @@ class AIE2PTargetLowering : public AIEBaseTargetLowering {
   // that properly implements CTLZ.
   // I wonder why this isn't the default.
```
martien-de-jong, line 32: Why do I always read ctrl-Z here?

llvm/test/CodeGen/AIE/aie2p/GlobalIsel/irtranslator-fifo-ldst.ll
```diff
@@ -7,7 +7,10 @@
 ; (c) Copyright 2025 Advanced Micro Devices, Inc. or its affiliates
 
 ; RUN: llc -mtriple=aie2p -stop-after=irtranslator -global-isel -verify-machineinstrs %s -o - 2>&1 | FileCheck %s --check-prefix=GMIR
+; RUN: llc -mtriple=aie2p -stop-after=instruction-select -global-isel -verify-machineinstrs %s -o - 2>&1 | FileCheck %s --check-prefix=ISEL
 
+; This test verifies that MMos are created by the IRTranslator for FIFO ld/st
```
martien-de-jong: nit MMOs

clang/lib/CodeGen/CGBuiltin.cpp
```diff
@@ -22153,6 +22153,16 @@ Value *CodeGenFunction::EmitAIE1BuiltinExpr(unsigned BuiltinID,
   return nullptr;
 }
 
+void insertImplicitCasts(MutableArrayRef<llvm::Value *> Args,
+                         llvm::FunctionType &FTy, CGBuilderTy &Builder) {
+  for (unsigned ArgNum = 0; ArgNum < Args.size(); ++ArgNum) {
+    llvm::Value *Arg = Args[ArgNum];
+    llvm::Type *RequiredType = FTy.getParamType(ArgNum);
+    if (Arg->getType() != RequiredType)
```
martien-de-jong, line 22161: check: pointer equality is equivalent to type equality

clang/lib/CodeGen/CGBuiltin.cpp
```diff
@@ -22392,7 +22402,10 @@ Value *CodeGenFunction::EmitAIEBuiltinExpr(unsigned BuiltinID,
 
     llvm::Intrinsic::ID IntrinsicID = getAIEIntrinsicFunction(BuiltinID, Arch);
     assert(IntrinsicID != Intrinsic::not_intrinsic);
-    Function *F = CGM.getIntrinsic(IntrinsicID);
+    Function *F =
+        CGM.getIntrinsic(IntrinsicID, {Ops[0]->getType(), Ops[0]->getType()});
```
martien-de-jong: Add a comment that this resolves overloading?

llvm/lib/CodeGen/GlobalISel/IRTranslator.cpp
gbossu: Not entirely: This allows lazy evaluation, which is important because `Info.memVT.getTypeForEVT()` will fail for `MVT::Other`

llvm/lib/Target/AIE/aie2p/AIE2PISelLowering.cpp
gbossu: Correct, but the internal result will be shifted to act as a byte-aligned load (of less than 512-bit then)

clang/lib/CodeGen/CGBuiltin.cpp
gbossu: For every occurrence? I'll check if I can find a sensible place.

martien-de-jong (commented):
LGTM

clang/lib/CodeGen/CGBuiltin.cpp
gbossu: I now have:
```
    llvm::Type *OverloadedTy = Ops[0]->getType();
    llvm::Intrinsic::ID IntrinsicID = getAIEIntrinsicFunction(BuiltinID, Arch);
    assert(IntrinsicID != Intrinsic::not_intrinsic);
    Function *F = CGM.getIntrinsic(IntrinsicID, {OverloadedTy, OverloadedTy});
```
Let me know if you find it clearer.

clang/test/CodeGen/aie/aie2p/aie2p-bank-annotation.cpp
```diff
@@ -26,12 +26,10 @@
 // CHECK-NEXT:    [[TMP0:%.*]] = load <32 x i32>, ptr [[S]], align 64, !tbaa [[TBAA6:![0-9]+]]
 // CHECK-NEXT:    [[TMP1:%.*]] = load i32, ptr [[POS1_I]], align 64, !tbaa [[TBAA7:![0-9]+]]
 // CHECK-NEXT:    [[TMP2:%.*]] = tail call { ptr, <32 x i32>, i32 } @llvm.aie2p.fifo.st.push.576.bfp16(ptr [[P_ADDR_SROA_0_0_P_ADDR_SROA_0_0_P_ADDR_SROA_0_0_P_ADDR_SROA_0_0_]], <64 x i8> [[V_COERCE_FCA_0_EXTRACT_I]], <8 x i8> [[V_COERCE_FCA_1_EXTRACT_I]], <32 x i32> [[TMP0]], i32 [[TMP1]])
-// CHECK-NEXT:    [[TMP3:%.*]] = extractvalue { ptr, <32 x i32>, i32 } [[TMP2]], 0
-// CHECK-NEXT:    [[TMP4:%.*]] = extractvalue { ptr, <32 x i32>, i32 } [[TMP2]], 1
-// CHECK-NEXT:    [[TMP5:%.*]] = extractvalue { ptr, <32 x i32>, i32 } [[TMP2]], 2
-// CHECK-NEXT:    store <32 x i32> [[TMP4]], ptr [[S]], align 128
-// CHECK-NEXT:    store i32 [[TMP5]], ptr [[POS1_I]], align 64
-// CHECK-NEXT:    store ptr [[TMP3]], ptr [[P_ADDR_SROA_0]], align 4
+// CHECK-NEXT:    [[TMP3:%.*]] = extractvalue { ptr, <32 x i32>, i32 } [[TMP2]], 1
```
martien-de-jong: CHECK, lost one to DCE as part of -O2

martien-de-jong (approved):

PR #332, opened 2025-02-04
[AIE2P] Support fifo loads with extra storage

khallouh:


llvm/test/CodeGen/AIE/aie2p/fifo-loads.ll
```diff
@@ -624,6 +624,98 @@ entry:
...
+define dso_local noundef <64 x i8> @_Z17test_fifo_ld_popxRPDv64_hR12fifo_state_t(ptr nocapture nonnull align 4 dereferenceable(4) %p, ptr nocapture nonnull align 64 dereferenceable(256) %s) local_unnamed_addr #0 {
+; CHECK-LABEL: _Z17test_fifo_ld_popxRPDv64_hR12fifo_state_t:
+; CHECK:         .p2align 4
+; CHECK-NEXT:  // %bb.0: // %entry
+; CHECK-NEXT:    lda p0, [p0, #0]; mov dj0, #128
+; CHECK-NEXT:    lda r24, [p1, dj0]
+; CHECK-NEXT:    nop
+; CHECK-NEXT:    nop
+; CHECK-NEXT:    vlda lfl0, [p1, #0]
+; CHECK-NEXT:    vlda x0, [p1, #192]; mov p2, p0
```
khallouh, line 636: We select the vector load because of the mapping to the vector bank instead of the fifo bank. This is because we introduce a copy during InstructionSelect to map the `extra fifo` input and output operands of the llvm intrinsic to the implicit use and def of `lfe` register in the instruction. The reason being we don't currently look into the uses beyond copies, bitcasts and phi nodes when determining the register bank assignments.

gbossu:
I'd be happy if you could look at the changes I've made in https://github.com/Xilinx/llvm-aie/pull/331/files and already apply them here for the new frontend intrinsics, builtins, and LLVM intrinsics. LMK if you need more details :)

khallouh:
> I'd be happy if you could look at the changes I've made in https://github.com/Xilinx/llvm-aie/pull/331/files and already apply them here for the new frontend intrinsics, builtins, and LLVM intrinsics. LMK if you need more details :)

Updated

clang/test/CodeGen/aie/aie2p/aie2p-ldst.cpp
```diff
@@ -6404,3 +6405,613 @@ v32int16 test_fifo_ld_pop_3d_byte(v32int16 __aie_dm_resource_b *p, fifo_state_t
...
+// CHECK-NEXT:    store <16 x i32> [[TMP9]], ptr [[EXTRA3_I_I]], align 64
+// CHECK-NEXT:    store <32 x i32> [[TMP7]], ptr [[S]], align 128
+// CHECK-NEXT:    store i32 [[TMP8]], ptr [[POS1_I_I]], align 64
+// CHECK-NEXT:    store ptr [[TMP6]], ptr [[P]], align 4
+// CHECK-NEXT:    [[TMP10:%.*]] = bitcast <64 x i8> [[TMP5]] to <16 x float>
+// CHECK-NEXT:    ret <16 x float> [[TMP10]]
+//
+v16float test_fifo_ld_popx(v16float *&p, fifo_state_t &s) {
+  return fifo_ld_popx(p, s);
+}
```
gbossu, line 7017: Could you add some tests with memory bank annotations in `clang/test/CodeGen/aie/aie2p/aie2p-bank-annotation.cpp`?

~And extend `llvm/test/CodeGen/AIE/aie2p/GlobalIsel/irtranslator-fifo-ldst.ll` accordingly~

khallouh: I think I extended `irtranslator-fifo-ldst.ll` already.

clang/test/CodeGen/aie/aie2p/aie2p-bank-annotation.cpp
```diff
@@ -14,51 +14,131 @@
...
+// CHECK-NEXT:    ret <64 x i8> [[TMP23]]
 //
-void test_intrinsic_annotated_pointer(v64bfp16ebs8_unaligned __aie_dm_resource_a * p, v64bfp16ebs8 v,
+v128uint4 test_intrinsic_annotated_pointer(v64bfp16ebs8_unaligned __aie_dm_resource_a * p_unaligned, v128int4 __aie_dm_resource_b *p, v64bfp16ebs8 v,
                        fifo_state_t &s) {
-  return fifo_st_push(p, v, s);
+  fifo_ld_fill(p_unaligned, s);
+  fifo_ld_fillx(p_unaligned, s);
+  fifo_st_push(p_unaligned, fifo_ld_pop(p_unaligned, s), s);
+  return  fifo_ld_popx(p, s);
```
gbossu, line 152: Why not test the `x` variants in a separate function? I think this would make the scope of this commit clearer.

llvm/test/CodeGen/AIE/aie2p/fifo-loads.ll
```diff
@@ -624,6 +624,98 @@ entry:
...
+  %5 = extractvalue { ptr, <32 x i32>, i32, <16 x i32> } %4, 0
+  %6 = extractvalue { ptr, <32 x i32>, i32, <16 x i32> } %4, 1
+  %7 = extractvalue { ptr, <32 x i32>, i32, <16 x i32> } %4, 2
+  %8 = extractvalue { ptr, <32 x i32>, i32, <16 x i32> } %4, 3
+  store <16 x i32> %8, ptr %extra3.i, align 64
+  store <32 x i32> %6, ptr %s, align 128
+  store i32 %7, ptr %pos1.i, align 64
+  store ptr %5, ptr %p, align 4
+  ret void
+}
```
gbossu, line 716: Nit: These IR functions could be made a lot more concise and better focus on the fillx/popx intrinsics.

clang/lib/Headers/aie2p_aie_api_compat.h
```diff
@@ -659,34 +659,6 @@ inline __attribute__((always_inline)) v64int8 pack_conf(v64int16 , crsat_t );
...
-inline __attribute__((always_inline)) v8cint32 fifo_ld_popx(v8cint32 *& , fifo_state_t & );
-inline __attribute__((always_inline)) v8cint32 fifo_ld_popx(v8cint32 *& , fifo_state_t & , int , int );
-inline __attribute__((always_inline)) v32bfloat16 fifo_ld_popx(v32bfloat16 *& , fifo_state_t & );
-inline __attribute__((always_inline)) v32bfloat16 fifo_ld_popx(v32bfloat16 *& , fifo_state_t & , int , int );
-inline __attribute__((always_inline)) v16float fifo_ld_popx(v16float *& , fifo_state_t & );
-inline __attribute__((always_inline)) v16float fifo_ld_popx(v16float *& , fifo_state_t & , int , int );
-inline __attribute__((always_inline)) void fifo_ld_fillx(v64bfp16ebs8_unaligned *& , fifo_state_t & );
-inline __attribute__((always_inline)) void fifo_ld_fillx(v64bfp16ebs8_unaligned *& , fifo_state_t & , int , int );
-inline __attribute__((always_inline)) void fifo_ld_fillx(v64bfp16ebs16_unaligned *& , fifo_state_t & );
-inline __attribute__((always_inline)) void fifo_ld_fillx(v64bfp16ebs16_unaligned *& , fifo_state_t & , int , int );
```
gbossu, line 689: Could that be squashed with the commit where you introduce the real definitions?

clang/test/CodeGen/aie/aie2p/aie2p-bank-annotation.cpp
khallouh: To keep a smaller test file and I noticed we didn't have a test for fill and pop so I thought I could test them all together. I can split them to 3 different tests if you prefer.

llvm/test/CodeGen/AIE/aie2p/fifo-loads.ll
khallouh: They also work as an end to end test since they come directly from the `aie2p-ldst.cpp` test file and show what kind of code we generate from there (See vector load above instead of fifo load)

clang/test/CodeGen/aie/aie2p/aie2p-bank-annotation.cpp
khallouh: I split them

clang/lib/Headers/aie2p_aie_api_compat.h
khallouh: Done

llvm/test/CodeGen/AIE/aie2p/fifo-loads.ll
gbossu: Sure, but I think there's a lot of "un-interesting" instructions in the test that blur what we really want to test end to end: the fifo intrinsics. Nothing to change :)

llvm/test/CodeGen/AIE/aie2p/GlobalIsel/inst-select-fifo-loads.mir
```diff
@@ -29,6 +29,62 @@ body:             |
...
+    liveins:
+    ; CHECK-LABEL: name: ld_popx
+    ; CHECK: [[DEF:%[0-9]+]]:eps = IMPLICIT_DEF
+    ; CHECK-NEXT: [[DEF1:%[0-9]+]]:eldfiforeg = IMPLICIT_DEF
+    ; CHECK-NEXT: [[DEF2:%[0-9]+]]:erf2 = IMPLICIT_DEF
+    ; CHECK-NEXT: [[DEF3:%[0-9]+]]:fifo512 = IMPLICIT_DEF
+    ; CHECK-NEXT: [[DEF4:%[0-9]+]]:mr30_fifo_step_e1 = IMPLICIT_DEF
+    ; CHECK-NEXT: $lfe = COPY [[DEF3]]
+    ; CHECK-NEXT: [[VLDB_POPX_512_:%[0-9]+]]:vec512, [[VLDB_POPX_512_1:%[0-9]+]]:eps, [[VLDB_POPX_512_2:%[0-9]+]]:eldfiforeg, [[VLDB_POPX_512_3:%[0-9]+]]:erf2 = VLDB_POPX_512 [[DEF4]], [[DEF4]], [[DEF]], [[DEF1]], [[DEF2]], implicit-def $lfe, implicit-def $srfifo_uf, implicit $lfe
+    ; CHECK-NEXT: [[COPY:%[0-9]+]]:mextra = COPY $lfe
```
gbossu, line 76: This `mextra` reg class is a singleton class. We might run into register allocation problems when multiple POPX instructions are used with different "extra" parameters.

in a follow-up, we might want to extend `AIE2PRegisterInfo::getLargestLegalSuperClass` for all FIFO registers. That will give greedy the chance to inflate the RC to a bigger one and not run out of registers.

clang/test/CodeGen/aie/aie2p/aie2p-bank-annotation.cpp
```diff
@@ -129,6 +152,37 @@ v128uint4 test_intrinsic_annotated_pointer_reference_fillx_popx(v64bfp16ebs8_una
...
+// CHECK-NEXT:    [[TMP14:%.*]] = extractvalue { <64 x i8>, ptr addrspace(6), <32 x i32>, i32 } [[TMP10]], 3
+// CHECK-NEXT:    store <32 x i32> [[TMP13]], ptr [[S]], align 128
+// CHECK-NEXT:    store i32 [[TMP14]], ptr [[POS1_I]], align 64
+// CHECK-NEXT:    store ptr addrspace(6) [[TMP12]], ptr [[P]], align 4
+// CHECK-NEXT:    ret <64 x i8> [[TMP11]]
+//
+v128uint4 test_intrinsic_annotated_pointer_reference_fill_pop(v64bfp16ebs8_unaligned __aie_dm_resource_a * &p_unaligned, v128int4 __aie_dm_resource_b *&p, fifo_state_t &s) {
+  fifo_ld_fill(p_unaligned, s);
+  return  fifo_ld_pop(p, s);
+}
```
gbossu, line 185: Can you squash this with the first commit?

gbossu, line 185: Can you squash this with the first commit?

gbossu: Oh that's for the standard pop/fill, not the "x" variants. Then it's fine.

llvm/test/CodeGen/AIE/aie2p/GlobalIsel/inst-select-fifo-loads.mir
gbossu: Note: For AIE2; we created the `ercr` class for rather similar reasons. It contains GPRs and all CR registers (which are not allocatable). Here I think we could create e.g. `xlfe` class which contains `X` regs and `lfe`. We could select the vreg copies to and from `lfe` to use that class instead of a singleton.

clang/test/CodeGen/aie/aie2p/aie2p-bank-annotation.cpp
```diff
@@ -34,8 +34,39 @@ void test_intrinsic_annotated_pointer(v64bfp16ebs8_unaligned __aie_dm_resource_a
...
+// CHECK-NEXT:    [[TMP8:%.*]] = extractvalue { <64 x i8>, ptr addrspace(6), <32 x i32>, i32, <16 x i32> } [[TMP7]], 0
+// CHECK-NEXT:    [[TMP9:%.*]] = extractvalue { <64 x i8>, ptr addrspace(6), <32 x i32>, i32, <16 x i32> } [[TMP7]], 2
+// CHECK-NEXT:    [[TMP10:%.*]] = extractvalue { <64 x i8>, ptr addrspace(6), <32 x i32>, i32, <16 x i32> } [[TMP7]], 3
+// CHECK-NEXT:    [[TMP11:%.*]] = extractvalue { <64 x i8>, ptr addrspace(6), <32 x i32>, i32, <16 x i32> } [[TMP7]], 4
+// CHECK-NEXT:    store <16 x i32> [[TMP11]], ptr [[EXTRA3_I_I]], align 64
+// CHECK-NEXT:    store <32 x i32> [[TMP9]], ptr [[S]], align 128
+// CHECK-NEXT:    store i32 [[TMP10]], ptr [[POS1_I_I]], align 64
+// CHECK-NEXT:    ret <64 x i8> [[TMP8]]
+//
+v128uint4 test_intrinsic_annotated_pointer_fillx_popx(v64bfp16ebs8_unaligned __aie_dm_resource_a * p_unaligned, v128int4 __aie_dm_resource_b *p, fifo_state_t &s) {
```
gbossu, line 62: Curious: if we declared `fifo_state_t &s` with a `__restrict` attribute, would the intermediate stores go away?

clang/test/CodeGen/aie/aie2p/aie2p-ldst.cpp
```diff
@@ -6404,3 +6405,613 @@ v32int16 test_fifo_ld_pop_3d_byte(v32int16 __aie_dm_resource_b *p, fifo_state_t
...
+v128int4 test_fifo_ld_popx(v128int4 *&p, fifo_state_t &s) {
+  return fifo_ld_popx(p, s);
+}
+
+// CHECK-LABEL: @_Z17test_fifo_ld_popxRPDv64_DU8_R12fifo_state_tii(
+// CHECK-NEXT:  entry:
+// CHECK-NEXT:    [[POS1_I:%.*]] = getelementptr inbounds i8, ptr [[S:%.*]], i20 128
+// CHECK-NEXT:    [[EXTRA3_I:%.*]] = getelementptr inbounds i8, ptr [[S]], i20 192
+// CHECK-NEXT:    [[SHL_I:%.*]] = shl i32 [[STEP:%.*]], 6
+// CHECK-NEXT:    [[OR_I:%.*]] = or i32 [[SHL_I]], [[MASK:%.*]]
```
gbossu, line 6562: Checking: I guess we are then expecting the mask to use a max of 6 bits?

gbossu (approved):
LGTM! I think that there are still too many "fully-fledged" tests for them to be useful. As I previously suggested, it would probably help to instead have a couple of concise tests that one can really look at to verify the generated IR from C++. And then a "completeness" test where all types and combinations are verified, but without `CHECK` lines. One could just look at the exit status of clang. But so be it :)

clang/test/CodeGen/aie/aie2p/aie2p-bank-annotation.cpp
khallouh: As I remember, yes.

clang/test/CodeGen/aie/aie2p/aie2p-ldst.cpp
khallouh: Yes, the mask is 6 bits and the step is 3 bits but they are both passed in a GPR (r30)

PR #333, opened 2025-02-04
[AIE2P] Enable S20 Narrowing for FIFO Loads and Stores

abhinay-anubola:
Added FIFO load and store intrinsics as S20 consumers in S20Narrowing
Added and updated the tests for the same

### DIFF
--- a/llvm/lib/Target/AIE/AIECombine.td
+++ b/llvm/lib/Target/AIE/AIECombine.td
@@ -21,7 +21,7 @@ def combine_globalval_offset : GICombineRule<
 def combine_S20NarrowingOpt_instruction_queue : GIDefMatchData<"std::set<InstrNode>">;
 def combine_S20NarrowingOpt : GICombineRule<
   (defs root:$root, combine_S20NarrowingOpt_instruction_queue:$matchinfo),
-  (match (wip_match_opcode G_PTR_ADD, G_INTRINSIC):$root,
+  (match (wip_match_opcode G_PTR_ADD, G_INTRINSIC, G_INTRINSIC_W_SIDE_EFFECTS):$root,
           [{ return matchS20NarrowingOpt(*${root}, MRI, ${matchinfo}); }]),
   (apply [{  applyS20NarrowingOpt(*${root}, MRI, B, Observer, Helper, ${matchinfo});}])
 >;
--- a/llvm/lib/Target/AIE/AIECombinerHelper.cpp
+++ b/llvm/lib/Target/AIE/AIECombinerHelper.cpp
@@ -631,6 +631,60 @@ static bool canProduceS20(const MachineRegisterInfo &MRI,
   }
 }
 
+/// Checks if the intrinsic natively consumes S20 for scalar inputs.
+static bool isNativeS20ConsumerIntrinsic(const unsigned IntrinsicID,
+                                         std::optional<unsigned> OperandIdx) {
+  static const std::map<unsigned, const std::set<unsigned>> S20OpIndices = {
+      {Intrinsic::aie2_add_2d, {4, 5, 6, 7}},
+      {Intrinsic::aie2_add_3d, {5, 6, 7, 8, 9, 10, 11}},
+      {Intrinsic::aie2p_add_2d, {4, 5, 6, 7}},
+      {Intrinsic::aie2p_add_3d, {5, 6, 7, 8, 9, 10, 11}},
+      {Intrinsic::aie2p_fifo_st_flush_1d, {7}},
+      {Intrinsic::aie2p_fifo_st_flush_1d_conv, {7}},
+      {Intrinsic::aie2p_fifo_ld_pop_1d_unaligned, {8}},
+      {Intrinsic::aie2p_fifo_st_flush_2d, {8, 9, 10, 11}},
+      {Intrinsic::aie2p_fifo_st_flush_2d_conv, {8, 9, 10, 11}},
+      {Intrinsic::aie2p_fifo_ld_pop_544_1d_bfp16, {9}},
+      {Intrinsic::aie2p_fifo_ld_pop_576_1d_bfp16, {9}},
+      {Intrinsic::aie2p_fifo_ld_pop_2d_unaligned, {9, 10, 11, 12}},
+      {Intrinsic::aie2p_fifo_st_flush_3d, {9, 10, 11, 12, 13, 14, 15}},
+      {Intrinsic::aie2p_fifo_st_flush_3d_conv, {9, 10, 11, 12, 13, 14, 15}},
+      {Intrinsic::aie2p_fifo_ld_pop_544_2d_bfp16, {10, 11, 12, 13}},
+      {Intrinsic::aie2p_fifo_ld_pop_576_2d_bfp16, {10, 11, 12, 13}},
+      {Intrinsic::aie2p_fifo_ld_pop_3d_unaligned, {10, 11, 12, 13, 14, 15, 16}},
+      {Intrinsic::aie2p_fifo_ld_pop_544_3d_bfp16, {11, 12, 13, 14, 15, 16, 17}},
+      {Intrinsic::aie2p_fifo_ld_pop_576_3d_bfp16,
+       {11, 12, 13, 14, 15, 16, 17}}};
+
+  auto It = S20OpIndices.find(IntrinsicID);
+  if (It == S20OpIndices.end())
+    return false;
+
+  if (!OperandIdx) {
+    return true;
+  }
+  const std::set<unsigned> &Indices = It->second;
+  return Indices.find(*OperandIdx) != Indices.end();
+}
+
+/// Checks if the instruction natively consumes S20 for scalar inputs.
+static bool
+isNativeS20Consumer(const MachineInstr &MI,
+                    std::optional<unsigned> OperandIdx = std::nullopt) {
+  switch (MI.getOpcode()) {
+  case TargetOpcode::G_PTR_ADD:
+    return true;
+  case TargetOpcode::G_INTRINSIC:
+  case TargetOpcode::G_INTRINSIC_W_SIDE_EFFECTS: {
+    const unsigned IntrinsicID = cast<GIntrinsic>(MI).getIntrinsicID();
+    return isNativeS20ConsumerIntrinsic(IntrinsicID, OperandIdx);
+  }
+
+  default:
+    return false;
+  }
+}
+
 /// The function checks if the node can be adapted to produce an S20 value, and
 /// recursively looks forward to see if all users can be changed to consume S20
 /// inputs.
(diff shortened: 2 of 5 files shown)


llvm/lib/Target/AIE/AIECombine.td
```diff
@@ -9,7 +9,7 @@
 //===----------------------------------------------------------------------===//
 include "llvm/Target/GlobalISel/Combine.td"
 
-// AIE-specific offset folding for G_GLOBAL_VALUE.
+// AIE-specific offset folding for G_GLOBALllvm/lib/Target/AIE/AIECombine.td_VALUE.
```
martien-de-jong: editor glitch?

llvm/lib/Target/AIE/AIECombinerHelper.cpp
```diff
@@ -621,6 +621,29 @@ static bool canProduceS20(const MachineRegisterInfo &MRI,
...
+  case Intrinsic::aie2p_fifo_ld_pop_1d_unaligned:
+  case Intrinsic::aie2p_fifo_ld_pop_2d_unaligned:
+  case Intrinsic::aie2p_fifo_ld_pop_3d_unaligned:
+  case Intrinsic::aie2p_fifo_ld_pop_544_1d_bfp16:
+  case Intrinsic::aie2p_fifo_ld_pop_544_2d_bfp16:
+  case Intrinsic::aie2p_fifo_ld_pop_544_3d_bfp16:
+  case Intrinsic::aie2p_fifo_ld_pop_576_1d_bfp16:
+  case Intrinsic::aie2p_fifo_ld_pop_576_2d_bfp16:
+  case Intrinsic::aie2p_fifo_ld_pop_576_3d_bfp16:
+  case Intrinsic::aie2p_fifo_st_flush_2d:
```
gbossu: We now have more intrinsics available, maybe double check they are all present

llvm/lib/Target/AIE/AIECombinerHelper.cpp
```diff
@@ -671,16 +694,14 @@ bool canNarrowUserTreeToS20(MachineRegisterInfo &MRI, InstrNode Start,
...
-      switch (cast<GIntrinsic>(Use).getIntrinsicID()) {
-      case Intrinsic::aie2_add_2d:
-      case Intrinsic::aie2_add_3d:
-      case Intrinsic::aie2p_add_2d:
-      case Intrinsic::aie2p_add_3d:
-        continue;
-      default:
+    case TargetOpcode::G_INTRINSIC_W_SIDE_EFFECTS: {
+      const unsigned IntrinsicID = cast<GIntrinsic>(Use).getIntrinsicID();
+      if (!isNativeS20ConsumerIntrinsic(IntrinsicID)) {
```
gbossu: Nice you factored that out!

llvm/test/CodeGen/AIE/aie2p/GlobalIsel/prelegalizercombiner-s20-narrowing.mir
```diff
@@ -612,3 +612,326 @@ body:             |
...
+    PseudoRET implicit $lr, implicit %5(<64 x s8>)
+...
+
+---
+name:            FIFO_LOAD_2D_unaligned
+legalized:       false
+tracksRegLiveness: true
+body:             |
+  bb.1:
+    liveins: $p0, $p1, $p2
```
andcarminati: nit: `$p2` is not used.

llvm/lib/Target/AIE/AIECombinerHelper.cpp
martien-de-jong: I'm wondering whether all scalar operands of these instructions are accurately qualified as 'S20'. It at least seems likely that they don't consume more than 20 bits.

andcarminati: See:
```
%4 = tail call { ptr, <32 x i32>, i32, <64 x i8>, <8 x i8> } @llvm.aie2p.fifo.ld.pop.576.1d.bfp16(ptr %0, <32 x i32> %1, i32 %2, i20 %3)
```

gbossu: > I'm wondering whether all scalar operands of these instructions are accurately qualified as 'S20

The availability register is indeed an `s32`, and we should leave it as is.

llvm/test/CodeGen/AIE/aie2p/GlobalIsel/prelegalizercombiner-s20-narrowing.mir
```diff
@@ -612,3 +612,326 @@ body:             |
...
+    G_STORE %2(s32), %0(p0) :: (store (s32))
+
+  bb.1:
+    successors: %bb.1(0x80000000); %bb.1(100.00%)
+
+    %6:_(s32) = G_PHI %2(s32), %bb.0, %7(s32), %bb.1
+    %8:_(s32) = G_PHI %2(s32), %bb.0, %9(s32), %bb.1
+    %10:_(s20) = G_TRUNC %6(s32)
+    %11:_(s20) = G_TRUNC %8(s32)
+    %12:_(p0), %13:_(<32 x s32>), %14:_(s32), %15:_(s20), %16:_(s20), %17:_(<64 x s8>), %18:_(<8 x s8>) = G_INTRINSIC_W_SIDE_EFFECTS intrinsic(@llvm.aie2p.fifo.ld.pop.544.3d.bfp16), %0(p0), %5(<32 x s32>), %2(s32), %1(s20), %1(s20), %11(s20), %1(s20), %1(s20), %10(s20), %1(s20), %3(<64 x s8>), %4(<8 x s8>)
```
andcarminati, line 800: Can we have a test where `%0(p0), %5(<32 x s32>), %2(s32),` are also taken from phi nodes?

llvm/lib/Target/AIE/AIECombinerHelper.cpp
```diff
@@ -671,16 +694,14 @@ bool canNarrowUserTreeToS20(MachineRegisterInfo &MRI, InstrNode Start,
     case TargetOpcode::G_STORE: // Data operand is later modified to S20 type
```
martien-de-jong, line 734: This switch seems to be an extension to `isNativeS20Consumer`.  I guess the difference is given by 'can be adapted to'.  Could we delegate the trivial cases to that predicate function?

llvm/lib/Target/AIE/AIECombinerHelper.cpp
gbossu: Re-opening: I think @martien-de-jong 's comment needs to be addressed.

This function should only return `true` for actual s20 operands. FIFO ld/st have an s32 availability operand, `isNativeS20ConsumerIntrinsic` should return `false` for this one. Maybe change the prototype to something like `isNativeS20ConsumerIntrinsic(unsigned IntrinsicID, unsigned OperandIdx)`?

F-Stuckmann: I believe the actual Operand extraction happens in `getOperandsToNarrow`, where `@llvm.aie2p.fifo.ld.pop.576.1d.bfp16` would not provide the `i32` operand as a `i20` operand to handle. Here we just check if the MI has Operands that are i20 and could profit from the combine.

llvm/test/CodeGen/AIE/aie2p/GlobalIsel/prelegalizercombiner-s20-narrowing.mir
```diff
@@ -612,3 +612,488 @@ body:             |
...
+    %10:_(<32 x s32>) = G_PHI %3(<32 x s32>), %bb.0, %11(<32 x s32>), %bb.1
+    %12:_(s20) = G_TRUNC %4(s32)
+    %13:_(s20) = G_TRUNC %6(s32)
+    %14:_(p0), %11:_(<32 x s32>), %9:_(s32), %15:_(s20), %16:_(s20) = G_INTRINSIC_W_SIDE_EFFECTS intrinsic(@llvm.aie2p.fifo.st.flush.3d), %0(p0), %10(<32 x s32>), %8(s32), %1(s20), %1(s20), %13(s20), %1(s20), %12(s20), %1(s20), %12(s20)
+    %5:_(s32) = G_ZEXT %15(s20)
+    %7:_(s32) = G_ZEXT %16(s20)
+    G_BR %bb.1
+...
+
+---
```
gbossu, line 1014: Nit: Maybe leave a comment explaining for what we are testing in that particular case.

llvm/test/CodeGen/AIE/aie2p/ldst-fifo-stores.ll
```diff
@@ -333,16 +333,16 @@ define dso_local void @_Z31test_fifo_st_flush_conv_2d_byteRPDv64_DB8_R12fifo_sta
...
+; CHECK-NEXT:    lda p4, [p0, #0]
+; CHECK-NEXT:    lda dc0, [p2, #0]; mov dj1, #128
 ; CHECK-NEXT:    lda r26, [p1, dj1]
 ; CHECK-NEXT:    vlda sfh, [p1, #64]
-; CHECK-NEXT:    nop
 ; CHECK-NEXT:    mov m0, r0
 ; CHECK-NEXT:    mov p3, p2
 ; CHECK-NEXT:    mov dn0, r1
 ; CHECK-NEXT:    mov dj0, r2
+; CHECK-NEXT:    mov p2, p4
```
gbossu, line 356: We seem to have more pointer movs now. Do you understand why?

abhinay-anubola: Yes, the extra pointer moves are there because after my change live range of p2 got extended. This means p2 is still in use when we need to store a new value in it, so the compiler cannot overwrite it directly. Instead, it assigns the new value to a temporary register (p4) and then moves it into p2 once it's safe to do so.

gbossu: Hmm, right, it would have been better for the scheduler to keep the ordering of the scalar loads as
```
lda dc0, [p2, #0]
lda p2, [p0, #0];
```

Now that they are reversed, we do need a new pointer, and then a copy because `vst.flush.512.2d` needs to consume  `$p2`.

Nothing you can easily change here I think.

llvm/lib/Target/AIE/AIECombinerHelper.cpp
```diff
@@ -621,6 +621,60 @@ static bool canProduceS20(const MachineRegisterInfo &MRI,
   }
 }
 
+/// Checks if the intrinsic natively consumes S20 for scalar inputs.
+static bool isNativeS20ConsumerIntrinsic(const unsigned IntrinsicID,
+                                         std::optional<unsigned> OperandIdx) {
+  static const std::map<unsigned, const std::set<unsigned>> S20OpIndices = {
+      {Intrinsic::aie2_add_2d, {4, 5, 6, 7}},
```
andcarminati, line 638: Can we keep all aie2-specific intrinsics first?

gbossu (dismissed):
LGTM!

gbossu (approved):

PR #323, opened 2025-01-31
[AIE2P] Support wide vector postinc 2D/3D, and offset load/store

abnikant:
1. Enable combine_load_store_increment, combine_offset_load_store_ptradd and combine_offset_load_store_share_ptradd
2. Support wide vector POSTINC, POSTINC_2D, POSTINC_3D load and store.
3. Fixup , offset opcode.
4. Enable fifo combined load/store incr and offset load/store for 512-bits.
5. Add tests for combining and load/store.

PR #337, opened 2025-02-06
[AIE2P] Improve `RegbankSelect` handling for load/store offset and post-increment addressing modes

abnikant:
Improved register bank assignments (for accumulator and FIFO register banks) to support load/store post-increment, post-increment 2D, post-increment 3D, and offset addressing modes.


llvm/lib/Target/AIE/aie2p/AIE2PRegisterBankInfo.cpp
```diff
@@ -251,32 +251,124 @@ AIE2PRegisterBankInfo::getInstrAlternativeMappings(
...
+    // Base alternative mapping for sizes up to 256.
+    // For 2048-bit sizes, mapping is restricted to the accumulator bank.
+    // Also, if the instruction has any implicit defs/uses, leave it alone.
+    if (Size <= 256 || Size == 2048 ||
+        MI.getNumOperands() != MI.getNumExplicitOperands())
       break;
+
+    // Prepare the three mapping indices.
+    std::vector<AIEBaseRegisterBankInfo::PartialMappingIdx> MappingIndices = {
+        getPartialMappingIdx(Ty), getAccPartialMappingIdx(Ty),
```
khallouh: Can we just use `getVecPartialMappingIdx` here?

llvm/lib/Target/AIE/aie2p/AIE2PRegisterBankInfo.cpp
```diff
@@ -251,32 +251,124 @@ AIE2PRegisterBankInfo::getInstrAlternativeMappings(
...
+          static_cast<AIEBaseRegisterBankInfo::PartialMappingIdx>(VecIdx);
+      for (unsigned Idx = 0; Idx < NumOperands; ++Idx) {
+        if (MI.getOperand(Idx).isReg() && MI.getOperand(Idx).getReg()) {
+          LLT Ty = MRI.getType(MI.getOperand(Idx).getReg());
+          if (!Ty.isValid())
+            continue;
+          OpdsMapping[Idx] = getValueMapping(OpRegBankIdx[Idx], OpSize[Idx]);
+        }
+      }
+      const InstructionMapping &VRegMapping = getInstructionMapping(
```
khallouh: Nit: I find the name `VRegMapping` to be a bit confusing here since this is supposed to be an alternative mapping

llvm/lib/Target/AIE/aie2p/AIE2PRegisterBankInfo.cpp
```diff
@@ -251,32 +251,124 @@ AIE2PRegisterBankInfo::getInstrAlternativeMappings(
...
   switch (MI.getOpcode()) {
   case TargetOpcode::G_LOAD:
-  case TargetOpcode::G_STORE: {
+  case AIE2P::G_AIE_OFFSET_LOAD:
+  case AIE2P::G_AIE_POSTINC_LOAD:
+  case AIE2P::G_AIE_POSTINC_2D_LOAD:
+  case AIE2P::G_AIE_POSTINC_3D_LOAD: {
+    const unsigned NumOperands = MI.getNumOperands();
+    const unsigned FirstSrcIdx = MI.getNumExplicitDefs();
+    unsigned MappingID = DefaultMappingID;
```
khallouh: Any reason you start at `DefaultMappingID` instead of 1 for your alternative mapping IDs? DefaultMappingID is a UINT_MAX and you are incrementing it. It also mentions:
```
Identifier used when the related instruction mapping instance is generated by target independent code.
Make sure not to use that identifier to avoid possible collision.
```

llvm/lib/Target/AIE/aie2p/AIE2PRegisterBankInfo.cpp
```diff
@@ -251,32 +251,124 @@ AIE2PRegisterBankInfo::getInstrAlternativeMappings(
...
+    // Base alternative mapping for sizes up to 256.
+    // For 2048-bit sizes, mapping is restricted to the accumulator bank.
+    // Also, if the instruction has any implicit defs/uses, leave it alone.
+    if (Size <= 256 || Size == 2048 ||
+        MI.getNumOperands() != MI.getNumExplicitOperands())
+      break;
+
+    // Prepare the three mapping indices.
+    std::vector<AIEBaseRegisterBankInfo::PartialMappingIdx> MappingIndices = {
+        getPartialMappingIdx(Ty), getAccPartialMappingIdx(Ty),
```
khallouh: Same as for the loads, Can't we use `getVecPartialMappingIdx` instead of `getPartialMappingIdx` from the base which still relies on the element size?

llvm/lib/Target/AIE/aie2p/AIE2PRegisterBankInfo.cpp
```diff
@@ -251,32 +251,124 @@ AIE2PRegisterBankInfo::getInstrAlternativeMappings(
...
+    SmallVector<const ValueMapping *, 8> OpdsMapping(NumOperands);
+    SmallVector<unsigned, 4> OpSize(NumOperands, 20);
+
+    // Initialize first few indices based on VRegOpIdx
+    for (unsigned I = 0; I < VecRegOpIdx; ++I) {
+      OpRegBankIdx[I] = (I == 0) ? PMI_PTR : PMI_MOD;
+    }
+
+    if ((VecRegOpIdx + 1) < NumOperands)
+      OpRegBankIdx[VecRegOpIdx + 1] = PMI_PTR;
```
khallouh: Can we simply assert this `(VecRegOpIdx + 1) < NumOperands` instead and directly assign `PMI_PTR`? I think all of the store opcodes handled here should verify this condition.

llvm/lib/Target/AIE/aie2p/AIE2PRegisterBankInfo.cpp
```diff
@@ -251,32 +251,124 @@ AIE2PRegisterBankInfo::getInstrAlternativeMappings(
...
+          static_cast<AIEBaseRegisterBankInfo::PartialMappingIdx>(VecIdx);
+      for (unsigned Idx = 0; Idx < NumOperands; ++Idx) {
+        if (MI.getOperand(Idx).isReg() && MI.getOperand(Idx).getReg()) {
+          LLT Ty = MRI.getType(MI.getOperand(Idx).getReg());
+          if (!Ty.isValid())
+            continue;
+          OpdsMapping[Idx] = getValueMapping(OpRegBankIdx[Idx], OpSize[Idx]);
+        }
+      }
+      const InstructionMapping &VRegMapping = getInstructionMapping(
```
khallouh: Same as for the loads. I would use a different name since this is not always a vector bank mapping

llvm/lib/Target/AIE/aie2p/AIE2PRegisterBankInfo.cpp
```diff
@@ -705,7 +800,8 @@ bool AIE2PRegisterBankInfo::hasFifoInput(const MachineInstr &MI,
                                          const Register RegOp) const {
   auto *RI = static_cast<const AIEBaseRegisterInfo *>(
       MI.getParent()->getParent()->getSubtarget().getRegisterInfo());
-  switch (MI.getOpcode()) {
+  const unsigned Op = MI.getOpcode();
+  switch (Op) {
```
khallouh: I had addressed a comment on my PR to directly use `switch (MI.getOpcode())` since `Op` for 'opcode' can be confusing as we also have a `RegOp` argument for 'operand' this time. Could you keep it and do the same for `usesAccReg` or at least change the name to avoid confusion with RegOp?

llvm/lib/Target/AIE/aie2p/AIE2PRegisterBankInfo.cpp
```diff
@@ -717,8 +813,11 @@ bool AIE2PRegisterBankInfo::hasFifoInput(const MachineInstr &MI,
       return true;
     break;
   }
-  case TargetOpcode::G_STORE: {
-    auto *RB = getRegBank(MI.getOperand(0).getReg(), MRI, TRI);
+  case TargetOpcode::G_STORE:
+  case AIE2P::G_AIE_OFFSET_STORE:
+  case AIE2P::G_AIE_POSTINC_STORE: {
+    const unsigned OpIdx = MI.getNumExplicitDefs();
```
khallouh: How about 2D and 3D?

llvm/lib/Target/AIE/aie2p/AIE2PRegisterBankInfo.cpp
```diff
@@ -687,8 +779,11 @@ bool AIE2PRegisterBankInfo::usesAccReg(const MachineInstr &MI,
...
```
khallouh: How about 2D and 3D?

llvm/lib/Target/AIE/aie2p/AIE2PRegisterBankInfo.cpp
```diff
@@ -935,71 +1050,117 @@ AIE2PRegisterBankInfo::getInstrMapping(const MachineInstr &MI) const {
...
+    const unsigned FirstSrcIdx = MI.getNumExplicitDefs();
+
     MachineMemOperand *MMO = *MI.memoperands_begin();
     const unsigned MemSize = 8 * MMO->getSize().getValue();
     // Handle the example case as below
     /* %2:_(p0) = G_FRAME_INDEX %stack.0
        G_STORE %1(<32 x s32>), %2(p0)
        %4:_(<32 x s32>) = G_LOAD %2(p0) */
+    // Size of accumulator and fifo vectors on aie2p >= 512.
     if (MemSize >= 512) {
```
khallouh: Could you remind me why the case above was handled according to the `MemSize` and why it should be greater than 512? 
Edit: clarified

llvm/lib/Target/AIE/aie2p/AIE2PRegisterBankInfo.cpp
```diff
@@ -935,71 +1050,117 @@ AIE2PRegisterBankInfo::getInstrMapping(const MachineInstr &MI) const {
...
-      // e.g. UseCandidate:_(p0) = G_FRAME_INDEX %stack.0
-      UseCandidate = DefMI->getOperand(0).getReg();
-      Type = MRI.getType(MI.getOperand(0).getReg());
+      if (DefMI) {
+        UseCandidate = getUseCandidateReg(*DefMI);
+        Type = MRI.getType(MI.getOperand(0).getReg());
+      }
     }
+
+    // Update mapping flags based on the refined candidate.
```
khallouh: I think if this just to address the case above, we can move it up inside the corresponding if case, or keep it here and remove it from above the if case. We don't have to ~~update~~ to do it twice.

llvm/lib/Target/AIE/aie2p/AIE2PRegisterBankInfo.cpp
```diff
@@ -781,6 +880,22 @@ bool AIE2PRegisterBankInfo::isUseFifoInsn(const MachineRegisterInfo &MRI,
                 });
 }
 
+// Given an instruction that produces a pointer used in a load operation,
+// return the candidate pointer register that should be traced
+// for determining the register bank mapping.
+const Register getUseCandidateReg(const MachineInstr &DefMI) {
```
khallouh: Nit: `getUseCandidatePointer`

llvm/lib/Target/AIE/aie2p/AIE2PRegisterBankInfo.cpp
khallouh: Ok, I just saw the comment, this based on the size of the acc/fifo on the target. In that case, can we just return earlier for anything smaller than 512?

llvm/lib/Target/AIE/aie2p/AIE2PRegisterBankInfo.cpp
```diff
@@ -935,71 +1050,117 @@ AIE2PRegisterBankInfo::getInstrMapping(const MachineInstr &MI) const {
...
+    for (unsigned I = 0; I < VecRegOpIdx; ++I) {
+      OpRegBankIdx[I] = (I == 0) ? PMI_PTR : PMI_MOD;
     }
-    return AIEBaseRegisterBankInfo::getInstrMapping(MI);
+    OpRegBankIdx[VecRegOpIdx] =
+        static_cast<AIEBaseRegisterBankInfo::PartialMappingIdx>(MappingIdx);
+
+    // Assign PMI_PTR to the next operand if within bounds
+    if ((VecRegOpIdx + 1) < NumOperands)
+      OpRegBankIdx[VecRegOpIdx + 1] = PMI_PTR;
```
khallouh: Same as above, can we simply assert here instead of the `if`?

llvm/lib/Target/AIE/aie2p/AIE2PRegisterBankInfo.cpp
abnikant: done

llvm/lib/Target/AIE/aie2p/AIE2PRegisterBankInfo.cpp
abnikant: oh! missed updating the switch case, added now.

llvm/lib/Target/AIE/aie2p/AIE2PRegisterBankInfo.cpp
abnikant: done.

llvm/lib/Target/AIE/aie2p/AIE2PRegisterBankInfo.cpp
abnikant: yes, makes sense.

llvm/lib/Target/AIE/aie2p/AIE2PRegisterBankInfo.cpp
abnikant: done

llvm/test/CodeGen/AIE/aie2p/GlobalIsel/regbankselect-fifo-insn.mir
```diff
@@ -976,6 +976,249 @@ body:             |
...
+    ; FAST-LABEL: name: offset_load_fifo512
+    ; FAST: liveins: $p0
+    ; FAST-NEXT: {{  $}}
+    ; FAST-NEXT: [[COPY:%[0-9]+]]:ptrregbank(p0) = COPY $p0
+    ; FAST-NEXT: [[C:%[0-9]+]]:modregbank(s20) = G_CONSTANT i20 448
+    ; FAST-NEXT: [[AIE_OFFSET_LOAD:%[0-9]+]]:fiforegbank(<8 x s64>) = G_AIE_OFFSET_LOAD [[COPY]](p0), [[C]](s20) :: (load (<8 x s64>))
+    ; FAST-NEXT: $lfl0 = COPY [[AIE_OFFSET_LOAD]](<8 x s64>)
+    %1:_(p0) = COPY $p0
+    %2:_(s20) = G_CONSTANT i20 448
+    %0:_(<8 x s64>) = G_AIE_OFFSET_LOAD %1:_(p0), %2:_(s20) :: (load (<8 x s64>))
```
khallouh, line 975: Could you add a test for G_AIE_OFFSET_STORE?

llvm/lib/Target/AIE/aie2p/AIE2PRegisterBankInfo.cpp
```diff
@@ -951,71 +1069,118 @@ AIE2PRegisterBankInfo::getInstrMapping(const MachineInstr &MI) const {
...
        G_STORE %1(<32 x s32>), %2(p0)
        %4:_(<32 x s32>) = G_LOAD %2(p0) */
-    if (MemSize >= 512) {
-      Register PtrReg = MI.getOperand(1).getReg();
-      MachineInstr *DefMI = MRI.getVRegDef(PtrReg);
-      // e.g. UseCandidate:_(p0) = G_FRAME_INDEX %stack.0
-      UseCandidate = DefMI->getOperand(0).getReg();
+    Register PtrReg = MI.getOperand(FirstSrcIdx).getReg();
+    MachineInstr *DefMI = MRI.getVRegDef(PtrReg);
+    if (DefMI) {
```
khallouh: Isn't that guaranteed by SSA?

khallouh: ~~Which seems to make the following block above kind of dead code~~ as we would update the use candidate every time
```
    if (isUseAccInsn(MRI, TRI, UseCandidate))
      isAccRegMapping = true;
    if (isUseFifoInsn(MRI, TRI, UseCandidate))
      isFifoPhysRegMapping = true;
```

khallouh: Ah ok so we would update the use candidate every time but we still need the first block for when the the load feeds acc/fifo instructions directly and not through the pointer.

llvm/lib/Target/AIE/aie2p/AIE2PRegisterBankInfo.cpp
```diff
@@ -951,71 +1069,118 @@ AIE2PRegisterBankInfo::getInstrMapping(const MachineInstr &MI) const {
...
     const unsigned MemSize = 8 * MMO->getSize().getValue();
+    // Size of accumulator and fifo vectors on aie2p >= 512.
+    // for MemSize < 512, fallback to base instruction mapping.
+    if (MemSize < 512)
+      return AIEBaseRegisterBankInfo::getInstrMapping(MI);
+
     // Handle the example case as below
     /* %2:_(p0) = G_FRAME_INDEX %stack.0
        G_STORE %1(<32 x s32>), %2(p0)
        %4:_(<32 x s32>) = G_LOAD %2(p0) */
```
khallouh: I'm not sure if we need to handle this case and update the use candidate. Here is the code after instruction-select for `test_v16x32_spill_reload` without this change:
```
  bb.0.entry:
    %0:fifo512 = COPY $lfl1
    VST_dmx_sts_fifohl_spill %0, %stack.0, implicit $sp :: (store (<16 x s32>))
    %3:vec512 = VLDA_dmx_lda_x_spill %stack.0, implicit $sp :: (dereferenceable load (<16 x s32>))
    VST_dmx_sts_x_spill %3, %stack.1, implicit $sp :: (volatile store (<16 x s32>))
    %4:fifo512 = VLDA_dmx_lda_fifohl_spill %stack.1, implicit $sp :: (volatile dereferenceable load (<16 x s32>))
    $lfl0 = COPY %4
    PseudoRET implicit $lr, implicit $lfl0
```
We simply use a vector load and store instead of fifo load and store, no cross bank copies are introduced.

llvm/lib/Target/AIE/aie2p/AIE2PRegisterBankInfo.cpp
```diff
@@ -251,32 +251,125 @@ AIE2PRegisterBankInfo::getInstrAlternativeMappings(
   const MachineRegisterInfo &MRI = MF.getRegInfo();
   switch (MI.getOpcode()) {
   case TargetOpcode::G_LOAD:
-  case TargetOpcode::G_STORE: {
+  case AIE2P::G_AIE_OFFSET_LOAD:
+  case AIE2P::G_AIE_POSTINC_LOAD:
```
khallouh, line 259: How about `G_AIE_OFFSET_ZEXTLOAD` and `G_AIE_OFFSET_SEXTLOAD`?

llvm/lib/Target/AIE/aie2p/AIE2PRegisterBankInfo.cpp
```diff
@@ -951,71 +1069,118 @@ AIE2PRegisterBankInfo::getInstrMapping(const MachineInstr &MI) const {
     }
     return AIEBaseRegisterBankInfo::getInstrMapping(MI);
   }
-  case TargetOpcode::G_LOAD: {
+  case TargetOpcode::G_LOAD:
+  case AIE2P::G_AIE_OFFSET_LOAD:
+  case AIE2P::G_AIE_POSTINC_LOAD:
```
khallouh, line 1050: How about `G_AIE_OFFSET_ZEXTLOAD` and `G_AIE_OFFSET_SEXTLOAD`?

llvm/lib/Target/AIE/AIE2InstrInfo.cpp
```diff
@@ -229,6 +229,13 @@ unsigned AIE2InstrInfo::getOffsetMemOpcode(unsigned BaseMemOpcode) const {
   llvm_unreachable("not a generic load/store");
 }
 
+bool AIE2InstrInfo::isGenericOffsetMemOpcode(unsigned Opcode) const {
```
andcarminati: I would say that all `G_AIE*` are target specific. I feel that `Generic` (besides of `G`) can be a bit confusing

khallouh (commented):
Since you add the regbankselect support for AIE's load and store opcodes in general. It would be nice to test the other assignments not just fifo512 and acc512. 
For scalars offset and postinc, there is these 2 tests for AIE2 which you can extend for AIE2P
`regbankselect-offset-postinc-store.mir` and `regbankselect-offset-postinc-load.mir`. I couldn't find any for vectors or 2D/3D

khallouh:
Note: all my review comments are concerning the 2nd commit regarding RegBankSelect. For the 1st commit (enabling combines), there is another PR https://github.com/Xilinx/llvm-aie/pull/323

llvm/lib/Target/AIE/aie2p/AIE2PRegisterBankInfo.cpp
```diff
@@ -951,71 +1069,118 @@ AIE2PRegisterBankInfo::getInstrMapping(const MachineInstr &MI) const {
...
+  case TargetOpcode::G_STORE:
+  case AIE2P::G_AIE_OFFSET_STORE:
+  case AIE2P::G_AIE_POSTINC_STORE:
+  case AIE2P::G_AIE_POSTINC_2D_STORE:
+  case AIE2P::G_AIE_POSTINC_3D_STORE: {
+    // Select the operand index for VecReg.
+    const unsigned VecRegOpIdx = MI.getNumExplicitDefs();
+    Register VecReg = MI.getOperand(VecRegOpIdx).getReg();
+    if (!VecReg)
       break;
```
khallouh: Why not return the base implementation?

khallouh: Also, can we return earlier for any type smaller than 512 like we do for the loads? since we are trying to assign either the accumulator or fifo bank.

llvm/test/CodeGen/AIE/aie2p/GlobalIsel/regbankselect-fifo-insn.mir
abnikant: added tests.

llvm/lib/Target/AIE/aie2p/AIE2PRegisterBankInfo.cpp
abnikant: Indeed. I have removed this special case. I don't find a case where it will help in better register bank assignment. We can always add if there is a need.

llvm/lib/Target/AIE/aie2p/AIE2PRegisterBankInfo.cpp
abnikant: We are falling back to Base Implementation for Size < 512.

llvm/lib/Target/AIE/AIE2InstrInfo.cpp
abnikant: As mentioned in other PR, we need to update other hooks as well to remove 'Generic'. Possibly we can do it in other/separate  PR?

llvm/lib/Target/AIE/aie2p/AIE2PRegisterBankInfo.cpp
abnikant: yes, now we return early for Size < 512.

abnikant:
> Since you add the regbankselect support for AIE's load and store opcodes in general. It would be nice to test the other assignments not just fifo512 and acc512. For scalars offset and postinc, there is these 2 tests for AIE2 which you can extend for AIE2P `regbankselect-offset-postinc-store.mir` and `regbankselect-offset-postinc-load.mir`. I couldn't find any for vectors or 2D/3D

enabled these for aie2p and also added tests for vector. 2D/3D tests were already added for accumulator and fifo registers.

llvm/test/CodeGen/AIE/aie2p/GlobalIsel/regbankselect-accumulator-insn.mir
```diff
@@ -205,7 +205,7 @@ body:             |
     ; FAST-NEXT: [[FRAME_INDEX:%[0-9]+]]:ptrregbank(p0) = G_FRAME_INDEX %stack.0
     ; FAST-NEXT: [[FRAME_INDEX1:%[0-9]+]]:ptrregbank(p0) = G_FRAME_INDEX %stack.1
     ; FAST-NEXT: G_STORE [[COPY]](<16 x s32>), [[FRAME_INDEX]](p0) :: (store (<16 x s32>))
-    ; FAST-NEXT: [[LOAD:%[0-9]+]]:accregbank(<16 x s32>) = G_LOAD [[FRAME_INDEX]](p0) :: (dereferenceable load (<16 x s32>))
+    ; FAST-NEXT: [[LOAD:%[0-9]+]]:vregbank(<16 x s32>) = G_LOAD [[FRAME_INDEX]](p0) :: (dereferenceable load (<16 x s32>))
```
niwinanto, line 208: Are we regressing here?

khallouh: See my comment https://github.com/Xilinx/llvm-aie/pull/337#discussion_r1948766897

llvm/lib/Target/AIE/aie2p/AIE2PRegisterBankInfo.cpp
```diff
@@ -255,32 +255,125 @@ AIE2PRegisterBankInfo::getInstrAlternativeMappings(
...
+    unsigned MappingID = 1;
+    unsigned const Cost = 1;
+
     Register DstReg = MI.getOperand(0).getReg();
     const LLT Ty = MRI.getType(DstReg);
-    unsigned Size = getSizeInBits(DstReg, MRI, TRI);
-    // Base Alternative Mapping for size <= 256.
-    if (Size <= 256)
-      break;
+    const unsigned Size = getSizeInBits(DstReg, MRI, TRI);
```
khallouh: Nit: DstSize

llvm/lib/Target/AIE/aie2p/AIE2PRegisterBankInfo.cpp
```diff
@@ -255,32 +255,125 @@ AIE2PRegisterBankInfo::getInstrAlternativeMappings(
...
+    // Select the operand index for VecReg.
+    const unsigned VecRegOpIdx = MI.getNumExplicitDefs();
+    unsigned MappingID = 1;
+    unsigned const Cost = 1;
 
-    AltMappings.push_back(&VRegMapping);
-    AltMappings.push_back(&AccRegMapping);
+    Register VecReg = MI.getOperand(VecRegOpIdx).getReg();
+    const LLT Ty = MRI.getType(VecReg);
+    const unsigned Size = getSizeInBits(VecReg, MRI, TRI);
```
khallouh: Nit: VecRegTy and VecRegSize

llvm/lib/Target/AIE/aie2p/AIE2PRegisterBankInfo.cpp
khallouh: nit: DstRegTy

llvm/lib/Target/AIE/aie2p/AIE2PRegisterBankInfo.cpp
```diff
@@ -964,71 +1066,105 @@ AIE2PRegisterBankInfo::getInstrMapping(const MachineInstr &MI) const {
...
-    }
+    // Size of accumulator and fifo vectors on aie2p >= 512.
+    // for MemSize < 512, fallback to base instruction mapping.
+    if (MemSize < 512)
+      return AIEBaseRegisterBankInfo::getInstrMapping(MI);
+
+    // If the register already belongs to a known bank, just return.
+    if (getRegBank(UseCandidate, MRI, TRI) == &AIE2P::AccRegBank ||
+        getRegBank(UseCandidate, MRI, TRI) == &AIE2P::FifoRegBank)
+      return AIEBaseRegisterBankInfo::getInstrMapping(MI);
```
khallouh, line 1071: Should we also return if we know it is in the vector bank?

llvm/lib/Target/AIE/aie2p/AIE2PRegisterBankInfo.cpp
```diff
@@ -964,71 +1066,105 @@ AIE2PRegisterBankInfo::getInstrMapping(const MachineInstr &MI) const {
...
+  case AIE2P::G_AIE_OFFSET_STORE:
+  case AIE2P::G_AIE_POSTINC_STORE:
+  case AIE2P::G_AIE_POSTINC_2D_STORE:
+  case AIE2P::G_AIE_POSTINC_3D_STORE: {
+    // Select the operand index for VecReg.
+    const unsigned VecRegOpIdx = MI.getNumExplicitDefs();
+    Register VecReg = MI.getOperand(VecRegOpIdx).getReg();
+    // Size of accumulator and fifo vectors on aie2p >= 512.
+    // for VecSize < 512, fallback to base instruction mapping.
+    if (!VecReg || MRI.getType(VecReg).getSizeInBits() < 512)
```
khallouh: Why do you need to check if(!VecReg)?

llvm/lib/Target/AIE/aie2p/AIE2PRegisterBankInfo.cpp
```diff
@@ -964,71 +1066,105 @@ AIE2PRegisterBankInfo::getInstrMapping(const MachineInstr &MI) const {
...
+      return AIEBaseRegisterBankInfo::getInstrMapping(MI);
+
+    auto *DefRegBank = getRegBank(VecReg, MRI, TRI);
+    // Only process if VecReg is defined in Accumulator or Fifo register banks.
+    if (DefRegBank != &AIE2P::AccRegBank && DefRegBank != &AIE2P::FifoRegBank)
+      return AIEBaseRegisterBankInfo::getInstrMapping(MI);
```
khallouh, line 1124: Why not process the vector bank ones as well?

llvm/lib/Target/AIE/aie2p/AIE2PRegisterBankInfo.cpp
niwinanto: As I understands we are handling this
```
    %1:_(<16 x s32>) = COPY $lfl1
    %2:_(p0) = G_FRAME_INDEX %stack.0
    G_STORE %1(<16 x s32>), %2(p0) :: (store (<16 x s32>))
    %4:_(<16 x s32>) = G_LOAD %2(p0) :: (dereferenceable load (<16 x s32>))
..
..
uses %4
```
Only issue I can see is, if we want to select `uses %4` instruction based on `%4` we might hit vector variant of instruction always and need cross bank copy always if there are other variants we can select.

Also, if you would like to remove this special case you can remove `AIE2PRegisterBankInfo::hasFifoInput` and corresponding accumulator function. It seems dead code.
```
  case TargetOpcode::G_STORE:
  case AIE2P::G_AIE_OFFSET_STORE:
  case AIE2P::G_AIE_POSTINC_STORE:
  case AIE2P::G_AIE_POSTINC_2D_STORE:
  case AIE2P::G_AIE_POSTINC_3D_STORE: {
    const unsigned OpIdx = MI.getNumExplicitDefs();
    auto *RB = getRegBank(MI.getOperand(OpIdx).getReg(), MRI, TRI);
    if (RB == &AIE2P::FifoRegBank)
      return true;
    break;
  }
```

niwinanto: Here is an example usecase
```
    ; FAST-LABEL: name: test_v16acc32_spill_reload
    ; FAST: [[COPY:%[0-9]+]]:accregbank(<16 x s32>) = COPY $bmll1
    ; FAST-NEXT: [[COPY1:%[0-9]+]]:gprregbank(s32) = COPY $r0
    ; FAST-NEXT: [[FRAME_INDEX:%[0-9]+]]:ptrregbank(p0) = G_FRAME_INDEX %stack.0
    ; FAST-NEXT: G_STORE [[COPY]](<16 x s32>), [[FRAME_INDEX]](p0) :: (store (<16 x s32>))
    ; FAST-NEXT: [[LOAD:%[0-9]+]]:vregbank(<16 x s32>) = G_LOAD [[FRAME_INDEX]](p0) :: (dereferenceable load (<16 x s32>))
    ; FAST-NEXT: [[ADD:%[0-9]+]]:vregbank(<16 x s32>) = G_ADD [[LOAD]], [[LOAD]]
    ; FAST-NEXT: [[COPY2:%[0-9]+]]:accregbank(<16 x s32>) = COPY [[ADD]](<16 x s32>)
    ; FAST-NEXT: G_INTRINSIC_W_SIDE_EFFECTS intrinsic(@llvm.aie2p.mcd.write.acc32), [[COPY2]](<16 x s32>), [[COPY1]](s32)
    ; FAST-NEXT: PseudoRET implicit $lr
    %1:_(<16 x s32>) = COPY $bmll1
    %3:gprregbank(s32) = COPY $r0
    %2:_(p0) = G_FRAME_INDEX %stack.0
    G_STORE %1(<16 x s32>), %2(p0) :: (store (<16 x s32>))
    %4:_(<16 x s32>) = G_LOAD %2(p0) :: (dereferenceable load (<16 x s32>))
    %5:_(<16 x s32>) = G_ADD %4, %4
    G_INTRINSIC_W_SIDE_EFFECTS intrinsic(@llvm.aie2p.mcd.write.acc32), %5(<16 x s32>), %3(s32)
    PseudoRET implicit $lr
```

abnikant: > Here is an example usecase
> 
> ```
>     ; FAST-LABEL: name: test_v16acc32_spill_reload
>     ; FAST: [[COPY:%[0-9]+]]:accregbank(<16 x s32>) = COPY $bmll1
>     ; FAST-NEXT: [[COPY1:%[0-9]+]]:gprregbank(s32) = COPY $r0
>     ; FAST-NEXT: [[FRAME_INDEX:%[0-9]+]]:ptrregbank(p0) = G_FRAME_INDEX %stack.0
>     ; FAST-NEXT: G_STORE [[COPY]](<16 x s32>), [[FRAME_INDEX]](p0) :: (store (<16 x s32>))
>     ; FAST-NEXT: [[LOAD:%[0-9]+]]:vregbank(<16 x s32>) = G_LOAD [[FRAME_INDEX]](p0) :: (dereferenceable load (<16 x s32>))
>     ; FAST-NEXT: [[ADD:%[0-9]+]]:vregbank(<16 x s32>) = G_ADD [[LOAD]], [[LOAD]]
>     ; FAST-NEXT: [[COPY2:%[0-9]+]]:accregbank(<16 x s32>) = COPY [[ADD]](<16 x s32>)
>     ; FAST-NEXT: G_INTRINSIC_W_SIDE_EFFECTS intrinsic(@llvm.aie2p.mcd.write.acc32), [[COPY2]](<16 x s32>), [[COPY1]](s32)
>     ; FAST-NEXT: PseudoRET implicit $lr
>     %1:_(<16 x s32>) = COPY $bmll1
>     %3:gprregbank(s32) = COPY $r0
>     %2:_(p0) = G_FRAME_INDEX %stack.0
>     G_STORE %1(<16 x s32>), %2(p0) :: (store (<16 x s32>))
>     %4:_(<16 x s32>) = G_LOAD %2(p0) :: (dereferenceable load (<16 x s32>))
>     %5:_(<16 x s32>) = G_ADD %4, %4
>     G_INTRINSIC_W_SIDE_EFFECTS intrinsic(@llvm.aie2p.mcd.write.acc32), %5(<16 x s32>), %3(s32)
>     PseudoRET implicit $lr
> ```

yes, we have some cases in mllib too, I noticed that while running mllib. I will bring back this code. I am already working on to refine this.

khallouh: It would be great if you could also include a reduced test case from mllib @abnikant

llvm/lib/Target/AIE/aie2p/AIE2PRegisterBankInfo.cpp
```diff
@@ -255,32 +255,125 @@ AIE2PRegisterBankInfo::getInstrAlternativeMappings(
   const MachineRegisterInfo &MRI = MF.getRegInfo();
   switch (MI.getOpcode()) {
   case TargetOpcode::G_LOAD:
-  case TargetOpcode::G_STORE: {
+  case AIE2P::G_AIE_OFFSET_LOAD:
```
niwinanto, line 258: Do we really care about the alternative mapping? Since the cost is same, I think we always stick to the instruction mapping. May be you can check with `--check-prefix=COMMON,GREEDY` in the test. I believe both fast and greedy produces the same result.

llvm/lib/Target/AIE/aie2p/AIE2PRegisterBankInfo.cpp
```diff
@@ -737,8 +834,13 @@ bool AIE2PRegisterBankInfo::hasFifoInput(const MachineInstr &MI,
...
   }
-  case TargetOpcode::G_STORE: {
-    auto *RB = getRegBank(MI.getOperand(0).getReg(), MRI, TRI);
+  case TargetOpcode::G_STORE:
+  case AIE2P::G_AIE_OFFSET_STORE:
+  case AIE2P::G_AIE_POSTINC_STORE:
+  case AIE2P::G_AIE_POSTINC_2D_STORE:
+  case AIE2P::G_AIE_POSTINC_3D_STORE: {
+    const unsigned OpIdx = MI.getNumExplicitDefs();
+    auto *RB = getRegBank(MI.getOperand(OpIdx).getReg(), MRI, TRI);
```
niwinanto: I am not sure `OpIdx` is the right value. As I understand, we would like to check whether the vector to be store is a `FifoReg`. Now we look for the pointer and it will always return false.

abnikant: why it is a pointer operand ?

abnikant: This Idx should point to the first source operand.

niwinanto: Yes, you are right. I got confused between LOAD.

llvm/lib/Target/AIE/aie2p/AIE2PRegisterBankInfo.cpp
```diff
@@ -255,32 +255,125 @@ AIE2PRegisterBankInfo::getInstrAlternativeMappings(
...
+    const unsigned Size = getSizeInBits(DstReg, MRI, TRI);
 
-    // If the instruction has any implicit-defs or uses,
-    // do not mess with it.
-    if (MI.getNumOperands() != 2)
+    // Base alternative mapping for sizes up to 256.
+    // For 2048-bit sizes, mapping is restricted to the accumulator bank.
+    // Also, if the instruction has any implicit defs/uses, leave it alone.
+    if (Size <= 256 || Size == 2048 ||
+        MI.getNumOperands() != MI.getNumExplicitOperands())
```
andcarminati: Is there any valid case of `MI.getNumOperands() != MI.getNumExplicitOperands()` considering loads?

llvm/lib/Target/AIE/aie2p/AIE2PRegisterBankInfo.cpp
```diff
@@ -255,32 +255,125 @@ AIE2PRegisterBankInfo::getInstrAlternativeMappings(
...
+    SmallVector<const ValueMapping *, 8> OpdsMapping(NumOperands);
+    SmallVector<unsigned, 4> OpSize(NumOperands, 20);
+
+    // Initialize first few indices based on VRegOpIdx
+    for (unsigned I = 0; I < VecRegOpIdx; ++I) {
+      OpRegBankIdx[I] = (I == 0) ? PMI_PTR : PMI_MOD;
+    }
+
+    // Assign PMI_PTR to the next operand if within bounds
+    assert(((VecRegOpIdx + 1) < NumOperands) && "RegOpIdx + 1 out of bound");
```
andcarminati, line 356: nit: `VecRegOpIdx + 1 out of bounds`

llvm/lib/Target/AIE/aie2p/AIE2PRegisterBankInfo.cpp
abnikant: I couldn't reproduce with the mllib tests I tried. I think we can drop it now and take it later if there is a need?

llvm/lib/Target/AIE/aie2p/AIE2PRegisterBankInfo.cpp
abnikant: done

llvm/lib/Target/AIE/aie2p/AIE2PRegisterBankInfo.cpp
abnikant: yes.

llvm/lib/Target/AIE/aie2p/AIE2PRegisterBankInfo.cpp
abnikant: not needed. dropped.

llvm/lib/Target/AIE/aie2p/AIE2PRegisterBankInfo.cpp
abnikant: we do the analysis to assign acc and fifo regbanks.

llvm/lib/Target/AIE/aie2p/AIE2PRegisterBankInfo.cpp
abnikant: yes, we do need in many cases, and also cost might change later.

llvm/lib/Target/AIE/aie2p/AIE2PRegisterBankInfo.cpp
abnikant: not, needed. I dropped this check.

llvm/lib/Target/AIE/aie2p/AIE2PRegisterBankInfo.cpp
abnikant: Fixed.

abnikant:
Also, I ran mllib 3 days back, there was no extra failure. I have fired the run again, will post the result once done.

llvm/lib/Target/AIE/aie2p/AIE2PRegisterBankInfo.cpp
niwinanto: Could you update the tests removing `--check-prefix=GREEDY` and  `--check-prefix=GREEDY` , I do not see any tests using alternative mappings.

abnikant: is your comment only pointing to load/store variants tests or other opcode as well?

llvm/test/CodeGen/AIE/aie2p/GlobalIsel/regbankselect-vector.mir
```diff
@@ -0,0 +1,548 @@
...
+    ; GREEDY-NEXT: $x0 = COPY [[COPY]](<16 x s32>)
+    ; GREEDY-NEXT: PseudoRET implicit $lr, implicit $x0
+    ;
+    ; FAST-LABEL: name: test_v16int32
+    ; FAST: [[COPY:%[0-9]+]]:vregbank(<16 x s32>) = COPY $x1
+    ; FAST-NEXT: $x0 = COPY [[COPY]](<16 x s32>)
+    ; FAST-NEXT: PseudoRET implicit $lr, implicit $x0
+    %0:_(<16 x s32>) = COPY $x1
+    $x0 = COPY %0(<16 x s32>)
+    PseudoRET implicit $lr, implicit $x0
```
khallouh: This test seems unrelated to load and store

llvm/test/CodeGen/AIE/aie2p/GlobalIsel/regbankselect-vector.mir
```diff
@@ -0,0 +1,548 @@
...
+    ; GREEDY-NEXT: $y0 = COPY [[COPY]](<32 x s32>)
+    ; GREEDY-NEXT: PseudoRET implicit $lr, implicit $y0
+    ;
+    ; FAST-LABEL: name: test_v32int32
+    ; FAST: [[COPY:%[0-9]+]]:vregbank(<32 x s32>) = COPY $y1
+    ; FAST-NEXT: $y0 = COPY [[COPY]](<32 x s32>)
+    ; FAST-NEXT: PseudoRET implicit $lr, implicit $y0
+    %0:_(<32 x s32>) = COPY $y1
+    $y0 = COPY %0(<32 x s32>)
+    PseudoRET implicit $lr, implicit $y0
```
khallouh: This test seems unrelated. Don't we have this already?

abnikant:
[result_with_regbank.txt](https://github.com/user-attachments/files/18825186/result_with_regbank.txt)
[result_without_regbank.txt](https://github.com/user-attachments/files/18825188/result_without_regbank.txt)
There are 3 extra passes with the register bank assignments : 
Transpose_aie2_bf16_120_pad  [without this change it timed out]
Transpose_aie2_bf16_201 [without this change it was an error in aiesimulator]
CastI32Bf16_AIE2P_bfloat16 [it was a diff at two data points - last bit differed by 1, now it a pass] - Edit - investigated this - generated code in with and without are same, only difference is in input data, so it seems depending on input data/ref the last bit differs at two data points

All of these needs an analysis why these are passing now, but these could be taken as a separate analysis/work. 

Edit -2 : both Transpose_aie2_bf16_120_pad and Transpose_aie2_bf16_201  pass when I re-run

llvm/lib/Target/AIE/aie2p/AIE2PRegisterBankInfo.cpp
```diff
@@ -810,6 +911,22 @@ AIE2PRegisterBankInfo::getPreferredRegBankForVectorTy(
   return std::nullopt;
 }
 
+// Given an instruction that produces a pointer used in a load operation,
+// return the candidate pointer register that should be traced
+// for determining the register bank mapping.
+const Register getUseCandidatePointer(const MachineInstr &DefMI) {
```
khallouh: This is not used anymore

llvm/test/CodeGen/AIE/aie2p/GlobalIsel/regbankselect-vector.mir
abnikant: this test is unrelated, removed.

llvm/lib/Target/AIE/aie2p/AIE2PRegisterBankInfo.cpp
abnikant: > As I understands we are handling this
> 
> ```
>     %1:_(<16 x s32>) = COPY $lfl1
>     %2:_(p0) = G_FRAME_INDEX %stack.0
>     G_STORE %1(<16 x s32>), %2(p0) :: (store (<16 x s32>))
>     %4:_(<16 x s32>) = G_LOAD %2(p0) :: (dereferenceable load (<16 x s32>))
> ..
> ..
> uses %4
> ```
> 
> Only issue I can see is, if we want to select `uses %4` instruction based on `%4` we might hit vector variant of instruction always and need cross bank copy always if there are other variants we can select.
> 
> Also, if you would like to remove this special case you can remove `AIE2PRegisterBankInfo::hasFifoInput` and corresponding accumulator function. It seems dead code.
> 
> ```
>   case TargetOpcode::G_STORE:
>   case AIE2P::G_AIE_OFFSET_STORE:
>   case AIE2P::G_AIE_POSTINC_STORE:
>   case AIE2P::G_AIE_POSTINC_2D_STORE:
>   case AIE2P::G_AIE_POSTINC_3D_STORE: {
>     const unsigned OpIdx = MI.getNumExplicitDefs();
>     auto *RB = getRegBank(MI.getOperand(OpIdx).getReg(), MRI, TRI);
>     if (RB == &AIE2P::FifoRegBank)
>       return true;
>     break;
>   }
> ```

cleaned up this code.

khallouh (approved):
LGTM

llvm/lib/Target/AIE/aie2p/AIE2PRegisterBankInfo.cpp
niwinanto: Here is the tracking ticket: https://jira.xilinx.com/browse/AIECC-937

PR #339, opened 2025-02-06
[AIEX] Combine G_SHUFFLE_VECTOR to UNMERGE

katerynamuts:
If we cannot combine `G_SHUFFLE_VECTOR `to `G_AIE_EXTRACT_SUBVECTOR`, then we try to combine it into `UNMERGE`

### DIFF
--- a/llvm/lib/Target/AIE/AIECombinerHelper.cpp
+++ b/llvm/lib/Target/AIE/AIECombinerHelper.cpp
@@ -1874,8 +1874,8 @@ static bool checkExtractSubvectorPrerequisites(const AIEBaseInstrInfo &TII,
       (DstTySize != ScalarRegSize && DstTySize != 2 * ScalarRegSize))
     return false;
 
-  // Currently, we cannot extract vectors of the size less than vector register
-  // size.
+  // Currently, we cannot extract vectors for the case when the size of the
+  // source vector is less than the basic vector register size (of the target).
   if (SrcTySize < VecRegSize)
     return false;
 
@@ -1945,6 +1945,38 @@ buildExtractSubvector(MachineIRBuilder &B, MachineRegisterInfo &MRI,
 }
 
 /// Match something like this:
+///  %1:_(<16 x s32>) = COPY $x0
+///  %2:_(<16 x s32>) = COPY $x1
+///  %0:_(<8 x s32>) = G_SHUFFLE_VECTOR %1(<16 x s32>), %2(<16 x s32>),
+///  shufflemask(8, 9, 10, 11, 12, 13, 14, 15)
+///  PseudoRET implicit $lr, implicit %0
+
+/// To convert to:
+/// %1:_(<16 x s32>) = COPY $x0
+/// %2:_(<8 x s32>), %3:_(<8 x s32>) = G_UNMERGE_VALUES %1(<16 x s32>)
+/// PseudoRET implicit $lr, implicit %3(<8 x s32>)
+static bool matchShuffleToUnmerge(MachineInstr &MI, MachineRegisterInfo &MRI,
+                                  BuildFnTy &MatchInfo, unsigned SubIdx,
+                                  unsigned NumSubVectors) {
+  const Register DstReg = MI.getOperand(0).getReg();
+  const Register Src1Reg = MI.getOperand(1).getReg();
+  const LLT DstTy = MRI.getType(DstReg);
+
+  // TODO: Select into G_EXTRACT_SUBVECTOR once it is more widely supported
+  MatchInfo = [=, &MRI](MachineIRBuilder &B) {
+    SmallVector<Register, 4> SubVecs;
+    for (unsigned I = 0; I < NumSubVectors; I++) {
+      if (I == (unsigned)SubIdx)
+        SubVecs.push_back(DstReg);
+      else
+        SubVecs.push_back(MRI.createGenericVirtualRegister(DstTy));
+    }
+    B.buildUnmerge(SubVecs, Src1Reg);
+  };
+  return true;
+}
+
+/// Match something like this:
 ///  %1:_(<16 x s16>) = COPY $wl0
 ///  %2:_(<16 x s16>) = COPY $wl1
 ///  %0:_(<4 x s16>) = G_SHUFFLE_VECTOR %1(<16 x s16>), %2(<16 x s16>),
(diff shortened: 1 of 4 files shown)


llvm/lib/Target/AIE/AIECombinerHelper.cpp
```diff
@@ -2033,3 +2028,68 @@ bool llvm::matchShuffleToExtractSubvec(MachineInstr &MI,
...
+    return false;
+
+  //  This should be handled by a separate combine that copies Src1Reg to
+  //  DstReg.
+  if (Src1TySize == DstTy.getSizeInBits())
+    return false;
+
+  const unsigned NumDstElems = DstTy.getNumElements();
+  const unsigned NumSrc1Elems = Src1Ty.getNumElements();
+  const unsigned NumSubVectors = NumSrc1Elems / NumDstElems;
```
khallouh: Nit: You could move this assignment after the 2 following early returns

llvm/lib/Target/AIE/AIECombinerHelper.cpp
```diff
@@ -2033,3 +2028,68 @@ bool llvm::matchShuffleToExtractSubvec(MachineInstr &MI,
...
+  ArrayRef<int> Mask = MI.getOperand(3).getShuffleMask();
+
+  const LLT DstTy = MRI.getType(DstReg);
+  const LLT Src1Ty = MRI.getType(Src1Reg);
+  const unsigned Src1TySize = Src1Ty.getSizeInBits();
+
+  if (!DstTy.isVector() || !Src1Ty.isVector())
+    return false;
+
+  // Boolean vectors are unlikely to select into subregister copy
```
khallouh: Why is that?

llvm/lib/Target/AIE/AIECombinerHelper.cpp
```diff
@@ -2033,3 +2028,68 @@ bool llvm::matchShuffleToExtractSubvec(MachineInstr &MI,
...
+    for (unsigned SubVecIdx = 0; SubVecIdx < NumSubVectors; ++SubVecIdx) {
+      if (checkSequentialMask(Mask, SubVecIdx * NumDstElems, NumDstElems)) {
+        return SubVecIdx;
+      }
+    }
+
+    return std::nullopt;
+  };
+
+  std::optional<unsigned> SubIdx = GetSubIdx();
```
khallouh: Nit: SubVecExtractIdx and GetSubVecExtractIdx

llvm/lib/Target/AIE/AIECombinerHelper.cpp
```diff
@@ -2033,3 +2028,68 @@ bool llvm::matchShuffleToExtractSubvec(MachineInstr &MI,
...
+
+  // Not an extract pattern
+  if (!SubIdx)
```
khallouh: Not a sub vector extract pattern, I think it could be an extract pattern but one that can't be selected to a sub reg copy.

khallouh: Ok, I see now that this doesn't only match to sub-register copies but also sub vector elements through vextract instructions, but still it could be a different extract pattern that doesn't map to either sub-reg copy or element extract.

llvm/lib/Target/AIE/AIECombinerHelper.cpp
```diff
@@ -1917,73 +1949,36 @@ static bool checkSequentialMask(const ArrayRef<int> Mask, unsigned StartValue,
...
-  if (!DstTy.isVector() || !Src1Ty.isVector() || Src1TySize < GPRSize ||
-      (DstTy.getSizeInBits() != GPRSize &&
-       DstTy.getSizeInBits() != 2 * GPRSize))
+  if (Src1TySize < GPRSize || (DstTy.getSizeInBits() != GPRSize &&
+                               DstTy.getSizeInBits() != 2 * GPRSize))
     return false;
 
-  // Currently, we cannot concat vectors of the size less than vector register
-  // size.
+  // Currently, we cannot concat vectors for the case when the size of the
```
khallouh: Nit: we cannot extract vectors...

khallouh: and "..less than the basic vector register (of the target)"

llvm/test/CodeGen/AIE/GlobalISel/prelegalizercombiner-shuffle-vector.mir
```diff
@@ -615,3 +615,57 @@ body:             |
...
+name: shuffle_vector_unmerge_lo
+tracksRegLiveness: true
+body:             |
+  bb.1:
+    liveins: $x0, $x1
+    ; CHECK-LABEL: name: shuffle_vector_unmerge_lo
+    ; CHECK: liveins: $x0, $x1
+    ; CHECK-NEXT: {{  $}}
+    ; CHECK-NEXT: [[COPY:%[0-9]+]]:_(<16 x s32>) = COPY $x0
+    ; CHECK-NEXT: [[AIE_UNPAD_VECTOR:%[0-9]+]]:_(<8 x s32>) = G_AIE_UNPAD_VECTOR [[COPY]](<16 x s32>)
```
khallouh, line 800: Nit: Maybe add a comment saying that this is a special case where g_unmerge is itself combined to g_aie_unpad_vector

khallouh: Also maybe add a test with an unmerge of 4 outputs and a bigger size so that you don't run into the `G_AIE_UNPAD_VECTOR ` case

llvm/test/CodeGen/AIE/aie2p/vmac.ll
```diff
@@ -259,116 +169,26 @@ define dso_local inreg noundef <64 x i32> @_Z27test_addmsc_4x16_16x16_confDv64_h
...
+; CHECK-NEXT:    vunpack y1, x4, unpacksign0; or r0, r1, r0
+; CHECK-NEXT:    or r1, r3, r2; mov crunpacksize, #0
+; CHECK-NEXT:    or r1, r1, r4; mov unpacksign0, r16
+; CHECK-NEXT:    mova r4, #13; or r1, r1, r5; mov r2, #50
+; CHECK-NEXT:    lshl r2, r6, r4; vshuffle x0, x0, x0, r2
+; CHECK-NEXT:    mova r3, #200; or r1, r1, r0; vmov wl4, wh0
+; CHECK-NEXT:    or r1, r1, r3
+; CHECK-NEXT:    or r0, r0, r2; vmsc dm0, dm1, x4, y3,r1
+; CHECK-NEXT:    or r0, r0, r3
+; CHECK-NEXT:    ret lr; vaddmsc dm0, dm0, dm2, x0, y1,r0
```
khallouh, line 187: Nice code size reduction!

llvm/test/CodeGen/AIE/GlobalISel/prelegalizercombiner-shuffle-vector.mir
```diff
@@ -615,3 +615,57 @@ body:             |
...
+    ; CHECK-NEXT: PseudoRET implicit $lr, implicit [[AIE_UNPAD_VECTOR]](<8 x s32>)
+    %1:_(<16 x s32>) = COPY $x0
+    %2:_(<16 x s32>) = COPY $x1
+    %0:_(<8 x s32>) = G_SHUFFLE_VECTOR %1(<16 x s32>), %2(<16 x s32>), shufflemask(0, 1, 2, 3, 4, 5, 6, 7)
```
khallouh, line 804: Here we test an output of 256 bits. More generally, it would be nice to also test one of 128, 512 and 1024 bits.

llvm/lib/Target/AIE/AIECombinerHelper.cpp
```diff
@@ -2033,3 +2028,68 @@ bool llvm::matchShuffleToExtractSubvec(MachineInstr &MI,
...
+  if (!DstTy.isVector() || !Src1Ty.isVector())
+    return false;
+
+  // Boolean vectors are unlikely to select into subregister copy
+  if (DstTy.getElementType() == LLT::scalar(1))
+    return false;
+
+  //  This should be handled by a separate combine that copies Src1Reg to
+  //  DstReg.
+  if (Src1TySize == DstTy.getSizeInBits())
```
khallouh, line 2083: Nit: It would be nice to have a test for this case even if we don't combine.

khallouh (commented):
Looks good to me! Only a few small suggestions and more tests would be nice.

llvm/lib/Target/AIE/AIECombinerHelper.cpp
```diff
@@ -1907,6 +1907,38 @@ static bool checkSequentialMask(const ArrayRef<int> Mask, unsigned StartValue,
   return true;
 }
 
+/// Match something like this:
+///  %1:_(<16 x s32>) = COPY $x0
+///  %2:_(<16 x s32>) = COPY $x1
+///  %0:_(<8 x s32>) = G_SHUFFLE_VECTOR %1(<16 x s32>), %2(<16 x s32>),
+///  shufflemask(8, 9, 10, 11, 12, 13, 14, 15)
+///  PseudoRET implicit $lr, implicit %0
```
martien-de-jong, line 1952: If we want to operate on the result, we probably want to keep it in a 512 bit register, and then this can be implemented with a shift.

llvm/lib/Target/AIE/AIECombinerHelper.cpp
katerynamuts: It simply means that it is not an extract pattern for this specific match function

llvm/lib/Target/AIE/AIECombinerHelper.cpp
katerynamuts: We don't need this. Removed.

llvm/lib/Target/AIE/AIECombinerHelper.cpp
katerynamuts: but I think it should be done not in the combine, the combine should result in the same type as the original `G_SHUFFLE_VECTOR`

khallouh (approved):
LGTM!

PR #340, opened 2025-02-06
Register re-allocation tweaks

gbossu:
This is a follow-up of the work from @F-Stuckmann and @martien-de-jong. It uses the least-recently-used approach for re-allocating registers. This limits WAW dependencies and gives more freedom to the post-pipeliner.

On top of this, there are two extra commits to allow:
 - Re-allocation of GPRs
 - Aggressive re-allocation: i.e. by de-allocation live-through registers.

QoR:
 - AIE2: 0.3% cycle count improvements on aie-public. There are a couple of regressions, but we get them back if forcing the post-pipeliner. This PR is also necessary to get the optimal schedule for GEMM_bf16 and Conv2D_int8
 - AIE2p: 4.2% cycle count improvement, because the register renamer was completely disabled.


llvm/lib/Target/AIE/AIEWawRegRewriter.cpp
```diff
@@ -96,14 +96,15 @@ class AIEWawRegRewriter : public MachineFunctionPass {
   BitVector getDefinedPhysRegs(const MachineBasicBlock *MBB) const;
 
   /// Returns true if the physical register \p Reg was replaced
-  bool replaceReg(const Register Reg, RoundRobin &Registers);
+  bool replaceReg(const Register Reg, RoundRobin &Registers,
+                  BitVector &UsedUnits);
```
gbossu, line 107: Note: I will split this commit, it does a bit too much, and the introduction of `BitVector &UsedUnits` is only needed for the next commit.

llvm/lib/Target/AIE/AIEWawRegRewriter.cpp
```diff
@@ -249,103 +242,141 @@ bool AIEWawRegRewriter::renameMBBPhysRegs(const MachineBasicBlock *MBB) {
...
+      }
+    }
+    for (auto &[MO, Org] : Candidates) {
+      auto VReg = MO->getReg();
+      assignReg(VReg, Org);
+    }
+  };
+
+  // Least-Recently-Used list. Append new uses at the end, search from the start
+  std::list<MCPhysReg> LRURegisters;
```
martien-de-jong, line 327: RoundRobin?

martien-de-jong: I guess the declaration can be closer to its first use.

llvm/lib/Target/AIE/AIEWawRegRewriter.cpp
```diff
@@ -380,8 +396,7 @@ AIEWawRegRewriter::getReplacementPhysReg(const Register VReg,
     if (IK == LiveRegMatrix::IK_Free) {
       // Move it to the end of the list. We return, so don't have to
       // care about invalidation
-      LRURegisters.erase(It);
-      LRURegisters.emplace_back(PhysReg);
+      moveRegAndAliasesBack(It, LRURegisters, TRI);
```
martien-de-jong: The aliases are probably in another regclass. I'm not sure anymore why we filtered on regclass, but if we want to move all registers in the alias set of PhysReg back, it would be cheaper to do an 'isInAliasSet(PhysReg)' than to find all aliases separately.

gbossu: `LRURegisters` can contain registers of multiple register classes, so we need to filter out the phys regs that are not compatible.

I'm not sure I understand the comment about `isInAliasSet(PhysReg)`. Here, if we choose `$cm0`, I want to move `bmh0` and `bml0` to the back of the list as well to make sure they are not picked when we need to re-allocate a `BM` register.

llvm/lib/Target/AIE/AIEWawRegRewriter.cpp
```diff
@@ -394,9 +423,32 @@ AIEWawRegRewriter::getReplacementPhysReg(const Register VReg,
     }
     LiveRegMatrix::InterferenceKind IK = LRM->checkInterference(LI, PhysReg);
     if (IK == LiveRegMatrix::IK_Free) {
+      // If the chosen physical register has already been used and the vreg to
+      // allocate is defined at a point where another vreg gets killed, prefer
+      // reusing the assignment of the killed reg.
```
martien-de-jong, line 472: I have a feeling that this would be counterproductive in our GPR Conv2D case, which is a chain of three single-def-single use chains.

llvm/lib/Target/AIE/AIEWawRegRewriter.cpp
```diff
@@ -248,7 +253,12 @@ bool AIEWawRegRewriter::renameMBBPhysRegs(const MachineBasicBlock *MBB) {
         continue;
 
       if (isWorthRenaming(Reg, VRegWithCopies)) {
-        Candidates.emplace_back(&MO, VRM->getPhys(Reg));
+        MCRegister AssignedPhysReg = VRM->getPhys(Reg);
```
martien-de-jong, line 259: const?

llvm/lib/Target/AIE/AIEWawRegRewriter.cpp
```diff
@@ -257,15 +267,27 @@ bool AIEWawRegRewriter::renameMBBPhysRegs(const MachineBasicBlock *MBB) {
...
+      unassignReg(VReg);
     auto *RC = MRI->getRegClass(VReg);
     RegClasses.insert(RC);
-    LLVM_DEBUG(dbgs() << "Candidate " << printReg(VReg, TRI, 0, MRI) << " ("
-                      << TRI->getName(Org) << ")\n");
   }
   LLVM_DEBUG(dbgs() << "Renaming " << Candidates.size() << " candidates in "
                     << RegClasses.size() << " classes\n");
 
+  // If requested, unassign MBB's liveins as well to get even more freedom
```
martien-de-jong, line 280: nit: this is not the same as live-through as described in the commit comment. However, these will definitely interfere with all loop-carried values, so are worth to be given lower priority.

llvm/lib/Target/AIE/AIEWawRegRewriter.cpp
gbossu: Agreed, that's why it's only done once we start to re-use registers. Otherwise they are still distributed "round robin"

llvm/lib/Target/AIE/AIEWawRegRewriter.cpp
gbossu: The MBB's liveins will include live-through, as well as LCDs, you are correct. I could make a distinction and only de-allocate purely live-through registers.

llvm/lib/Target/AIE/AIEWawRegRewriter.cpp
martien-de-jong: It's just that we do a find over the full list for each alias. I'm assuming that sweeping it once to move all aliases back would be cheaper. No big deal.

llvm/lib/Target/AIE/AIEWawRegRewriter.cpp
```diff
@@ -394,9 +423,32 @@ AIEWawRegRewriter::getReplacementPhysReg(const Register VReg,
...
     LiveRegMatrix::InterferenceKind IK = LRM->checkInterference(LI, PhysReg);
     if (IK == LiveRegMatrix::IK_Free) {
+      // If the chosen physical register has already been used and the vreg to
+      // allocate is defined at a point where another vreg gets killed, prefer
+      // reusing the assignment of the killed reg.
+      if (std::optional<Register> KilledReg =
+              getKilledRegAtSingledDefPoint(VReg, *MRI);
+          KilledReg &&
+          any_of(TRI->regunits(PhysReg),
+                 [&UsedUnits](MCRegUnit RU) { return UsedUnits.test(RU); })) {
```
martien-de-jong: nit: this is far beyond my limits to reasonably nest statements in an if condition.

llvm/test/CodeGen/AIE/aie2/ra/waw_reg_renaming_loop.mir
```diff
@@ -509,6 +509,110 @@ body:             |
...
+  ; CHECK-NEXT:   renamable $wl7, renamable $p0 = VLD_pstm_pseudo killed renamable $p0, renamable $m3
+  ; CHECK-NEXT:   renamable $wh7, renamable $p0 = VLD_pstm_pseudo killed renamable $p0, renamable $m4
+  ; CHECK-NEXT:   renamable $wl9, renamable $p0 = VLD_pstm_pseudo killed renamable $p0, renamable $m3
+  ; CHECK-NEXT:   renamable $wh9, renamable $p0 = VLD_pstm_pseudo killed renamable $p0, renamable $m4
+  ; CHECK-NEXT:   renamable $wl11, renamable $p0 = VLD_pstm_pseudo killed renamable $p0, renamable $m3
+  ; CHECK-NEXT:   $wh11, $p0, $dc1, $dc5 = VLD_3D_pseudo killed $p0, $d1_3d
+  ; CHECK-NEXT:   renamable $x5 = VSHUFFLE killed renamable $x5, renamable $x5, renamable $r2
+  ; CHECK-NEXT:   renamable $x7 = VSHUFFLE killed renamable $x7, renamable $x7, renamable $r2
+  ; CHECK-NEXT:   renamable $x9 = VSHUFFLE killed renamable $x9, renamable $x9, renamable $r2
+  ; CHECK-NEXT:   renamable $x11 = VSHUFFLE killed renamable $x11, renamable $x11, renamable $r2
```
martien-de-jong, line 563: Nice, especially that this still is seen as a single use

llvm/lib/Target/AIE/AIEWawRegRewriter.cpp
martien-de-jong: I'm fine with it. If you really want, you could adjust the comment.

llvm/lib/Target/AIE/AIE2TargetMachine.cpp
```diff
@@ -161,8 +161,10 @@ bool AIE2PassConfig::addRegAssignAndRewriteOptimized() {
     addPass(createAIESuperRegRewriter());
   }
   addPass(createGreedyRegisterAllocator());
-  if (EnableWAWRegRewrite)
+  if (EnableWAWRegRewrite) {
     addPass(createAIEWawRegRewriter());
+    addPass(createGreedyRegisterAllocator());
```
martien-de-jong, line 163: Curious, is this expensive, or is it proportional to the number of registers it needs to allocate?

llvm/test/CodeGen/AIE/aie2/ra/waw_reg_renaming_aggressive.mir
```diff
@@ -0,0 +1,132 @@
+# NOTE: Assertions have been autogenerated by utils/update_mir_test_checks.py UTC_ARGS: --version 4
+# NOTE: Example file for Write After Write Register Renaming in Loop test
+#
+# This file is licensed under the Apache License v2.0 with LLVM Exceptions.
+# See https://llvm.org/LICENSE.txt for license information.
+# SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+#
+# (c) Copyright 2025 Advanced Micro Devices, Inc. or its affiliates
+# unit test for the WAW register renaming pass and check edge cases
+# RUN: llc -mtriple=aie2 -verify-machineinstrs --start-before=greedy --stop-after=virtregrewriter \
```
martien-de-jong, line 10: nit: Do we have a motivating example for aie2p?

llvm/lib/Target/AIE/AIEWawRegRewriter.cpp
```diff
@@ -249,103 +242,141 @@ bool AIEWawRegRewriter::renameMBBPhysRegs(const MachineBasicBlock *MBB) {
...
+      if (!replaceReg(VReg, Registers)) {
+        LLVM_DEBUG(dbgs() << "Renaming VR" << Register::virtReg2Index(VReg)
+                          << " failed\n");
+        return false;
+      }
+    }
+    return true;
+  };
+
+  // Reapply the original allocation to the intersection of RC and Candidates
```
andcarminati: I am a bit confused on what we call intersection here.

llvm/lib/Target/AIE/AIEWawRegRewriter.cpp
```diff
@@ -249,103 +250,247 @@ bool AIEWawRegRewriter::renameMBBPhysRegs(const MachineBasicBlock *MBB) {
...
-        // BlockedPhysRegs, so that an additional replacement will not cause a
-        // WAW, which this pass is trying to remove.
-        UsedPhysRegs[VRM->getPhys(Reg)] = true;
+      if (isWorthRenaming(Reg, VRegWithCopies)) {
+        MCRegister AssignedPhysReg = VRM->getPhys(Reg);
+        Candidates.emplace_back(&MO, AssignedPhysReg);
+        LLVM_DEBUG(dbgs() << "Candidate " << printReg(Reg, TRI, 0, MRI) << ":"
+                          << TRI->getRegClassName(MRI->getRegClass(Reg)) << " ("
+                          << TRI->getName(AssignedPhysReg) << ")\n");
+        assert(AssignedPhysReg != VirtRegMap::NO_PHYS_REG);
```
F-Stuckmann: why not check AssignedPhysReg.isVirtual()?

Do we want to rename StackSlots?

llvm/lib/Target/AIE/AIE2TargetMachine.cpp
F-Stuckmann, line 163: why are we rerunning greedy?

gbossu: This will typically only run for registers that have been de-allocated (the live-throughs when "aggressive re-allocation" is enabled). So pretty cheap, or free if "aggressive re-allocation" is disabled.

llvm/test/CodeGen/AIE/aie2/ra/waw_reg_renaming_aggressive.mir
gbossu: Not that I know of (so far)

llvm/lib/Target/AIE/AIEWawRegRewriter.cpp
gbossu: Yes I think this could just be `// Reapply the original allocation to all Candidates`

llvm/lib/Target/AIE/AIEWawRegRewriter.cpp
```diff
@@ -249,103 +250,247 @@ bool AIEWawRegRewriter::renameMBBPhysRegs(const MachineBasicBlock *MBB) {
...
+                                                   BitVector &UsedUnits) const {
+  assert(VReg.isVirtual() && "Reg has to be a virtual register");
+
+  LLVM_DEBUG(dbgs() << "     Try to re-assign" << printReg(VReg, TRI) << "\n");
+  const TargetRegisterClass *RC = MRI->getRegClass(VReg);
+  LiveInterval &LI = LIS->getInterval(VReg);
+  // The order in Registers represent the LeastRecentlyUsed, which will
+  // distribute register use evenly
+  for (auto It = LRURegisters.begin(); It != LRURegisters.end(); ++It) {
+    MCPhysReg PhysReg = *It;
```
F-Stuckmann, line 463: Instead of using Physical Register as a whole for the round robin scheme, we could use RegUnits. 
this way, we would differentiate between different registers, that partially populate the same RegUnits, i.e. wl/h, X and y registers.

Unsure if this is necessary at this point.

gbossu: I think you are suggesting to use a list of RegUnits, instead of a list of MCRegister? I have to say I do prefer the latter, because it makes it really straightforward to retrieve the least-recently used register.

llvm/lib/Target/AIE/AIEWawRegRewriter.cpp
gbossu: I'm not sure I get this, `AssignedPhysReg` is a physreg, not a virtual one. Here I just want to make sure that every candidate for renaming does have phys reg assigned.

llvm/lib/Target/AIE/AIE2TargetMachine.cpp
F-Stuckmann, line 163: should greedy also run after `createAIEWawRegRewriter `in `AIE2p`?

llvm/lib/Target/AIE/AIEWawRegRewriter.cpp
```diff
@@ -249,103 +250,247 @@ bool AIEWawRegRewriter::renameMBBPhysRegs(const MachineBasicBlock *MBB) {
...
     LiveRegMatrix::InterferenceKind IK = LRM->checkInterference(LI, PhysReg);
-    if (IK == LiveRegMatrix::IK_Free)
+    if (IK == LiveRegMatrix::IK_Free) {
+      // If the chosen physical register has already been used and the vreg to
+      // allocate is defined at a point where another vreg gets killed, prefer
+      // reusing the assignment of the killed reg.
+      if (std::optional<Register> KilledReg =
+              getKilledRegAtSingledDefPoint(VReg, *MRI);
+          KilledReg &&
+          any_of(TRI->regunits(PhysReg),
```
F-Stuckmann: What happens if we have multiple RegUnits per PhysReg? Why don't all have to be used already?

llvm/lib/Target/AIE/AIEWawRegRewriter.cpp
andcarminati, line 327: I think this is closer to Least-Recently-Defined. LRU. Could we have any additional gain by implementing a full LRU approach? I mean, updating the list considering the uses as well?

llvm/lib/Target/AIE/AIE2TargetMachine.cpp
gbossu: Definitely, in general, the `AIEWawRegRewriter` pass does not even run for AIE2p. I assumed that the pass pipeline is shared, but that part is not. Good catch!

llvm/lib/Target/AIE/AIEWawRegRewriter.cpp
gbossu: I think this is meant as "least recently used for re-allocation". I could rename that to "least recently allocated" maybe?

llvm/test/CodeGen/AIE/aie2/ra/waw_reg_renaming_loop.mir
```diff
@@ -509,6 +509,110 @@ body:             |
     PseudoRET implicit $lr
 ...
 
+# This is a simplified example taken from the GEMM_bf16 kernel.
+# When we get to the last 4 VSHUFFLE, all 12 X registers have already been
+# distributed. Those VSHUFFLE represent a kill point for other X registers.
+# We should then prefer "re-using" the killed registers instead of us starting
+# anew from x0, x2, x4 and x6.
```
andcarminati, line 516: nit: a new

llvm/lib/Target/AIE/AIEWawRegRewriter.cpp
andcarminati: I will be happy if we can extract at least one lambda/predicate like `WasUsedBefore(PhysReg)` ;-)

llvm/lib/Target/AIE/AIEWawRegRewriter.cpp
```diff
@@ -360,6 +365,30 @@ bool AIEWawRegRewriter::replaceReg(const Register VReg,
...
+/// at the point \p VReg gets defined.
+std::optional<Register>
+getKilledRegAtSingledDefPoint(Register VReg, const MachineRegisterInfo &MRI) {
+  MachineOperand *MO = MRI.getOneDef(VReg);
+  if (!MO)
+    return std::nullopt;
+
+  MachineInstr &DefMI = *MO->getParent();
+  auto OnlyUsedByInstr = [&MRI](Register Reg, const MachineInstr &MI) {
+    return all_of(MRI.use_instructions(Reg),
```
andcarminati, line 418: Check: we consider that the register is killed because we have a single user: DefMI.

How about multiple uses, provide that all instructions that also use the register dom DefMI?

llvm/lib/Target/AIE/AIEWawRegRewriter.cpp
gbossu: This checks if any of the reg units have been used for renaming. Essentially I consider `d0_3d` used if any of its regunits has been used, be it `d0_3d` itself, or `d0`, or `m0`, or `dn4` etc.

llvm/lib/Target/AIE/AIEWawRegRewriter.cpp
gbossu: I guess that it could be extended at some point, but having multiple users will typically make the live range bigger. So far I want to favour small live ranges if they are to share a physical register. If it isn't sufficient for future benchmarks, we will adjust :)

llvm/test/CodeGen/AIE/aie2/ra/waw_reg_renaming_loop.mir
gbossu: https://dictionary.cambridge.org/us/dictionary/english/anew 🤓

andcarminati: Wow, now I will also use it ;-)

llvm/lib/Target/AIE/AIEWawRegRewriter.cpp
andcarminati: Maybe for a future reference.

llvm/lib/Target/AIE/AIEWawRegRewriter.cpp
martien-de-jong: Basically codifying that any register overlap will lead to a dependence

llvm/lib/Target/AIE/AIEWawRegRewriter.cpp
martien-de-jong: It still looks funny that this value is defined in the VirtRegMap scope. How does it know it conflicts with the MCRegister encoding? Would `VRM->hasPhys(Reg) `be more suitable?

llvm/test/CodeGen/AIE/aie2/ra/waw_reg_renaming_aggressive.mir
martien-de-jong: I would be very happy if it could improve SigmoidTemplated_bf16_0

llvm/lib/Target/AIE/AIEWawRegRewriter.cpp
```diff
@@ -326,6 +328,11 @@ bool AIEWawRegRewriter::renameMBBPhysRegs(const MachineBasicBlock *MBB) {
   // For each reg class, allocate the candidates in round-robin fashion.
   // If we fail, we fall back to the original allocation
   BitVector RegSet{TRI->getNumRegs()};
```
martien-de-jong: nit: more descriptive name?

llvm/lib/Target/AIE/AIEWawRegRewriter.cpp
```diff
@@ -355,7 +362,10 @@ bool AIEWawRegRewriter::isWorthRenaming(const Register &Reg,
   if (!VRM->hasPhys(Reg))
     return false;
 
-  if (!TRI->isVecOrAccRegClass(*(MRI->getRegClass(Reg))))
+  // Only consider vec/acc registers as candidates, and optionally GPRs.
+  if (!TRI->isVecOrAccRegClass(*(MRI->getRegClass(Reg))) &&
+      (!GPRRealloc ||
+       !TRI->getGPRRegClass(*MF)->hasSubClassEq(MRI->getRegClass(Reg))))
```
martien-de-jong: nit: factor this in a `isCandidateRegClass()` lambda? I would at least rework it to have a single negation at top level

llvm/lib/Target/AIE/AIEWawRegRewriter.cpp
gbossu: `NO_PHYS_REG` is the sentinel value used by VRM to say "nothing is assigned", so that scope makes sense to me. I'll use `VRM->hasPhys(Reg)` nonetheless :)

llvm/lib/Target/AIE/AIE2TargetMachine.cpp
gbossu: Added a new commit for AIE2p

gbossu:
I think I addressed all the comments. :) I also gave some QoR numbers in the PR description.

llvm/lib/Target/AIE/AIEWawRegRewriter.cpp
martien-de-jong: It's using the common acronym LRU from caching strategies.  Indeed, it is referring to the allocation in a forward traversal, not as defining a register.

llvm/test/CodeGen/AIE/aie2p/schedule/postpipeliner/end-to-end.ll
```diff
@@ -20,27 +20,29 @@ define <32 x i16> @zol(i32 %n, ptr %p) {
...
-; CHECK-NEXT:    vadd.16 x0, x2, x0
+; CHECK-NEXT:    vadd.16 x2, x0, x2
+; CHECK-NEXT:    vadd.16 x2, x0, x2
+; CHECK-NEXT:    vadd.16 x2, x0, x2
+; CHECK-NEXT:    vadd.16 x2, x0, x2
+; CHECK-NEXT:    vadd.16 x2, x0, x2
+; CHECK-NEXT:    vadd.16 x2, x0, x2
+; CHECK-NEXT:    vadd.16 x2, x0, x2
+; CHECK-NEXT:    nop
+; CHECK-NEXT:    vmov x0, x2
```
martien-de-jong, line 45: nit: no net effect in the loop, but ABI constraint ignored.

martien-de-jong (dismissed):

martien-de-jong (approved):

PR #341, opened 2025-02-06
[AIE2P] Instruction select G_TRUNC for vector operands

niwinanto:
We can instruction select `G_TRUNC` in to `VSHUFFLE` instruction. `VSHUFFLE` instruction supports de-interleaved modes which can be translated to G_TRUNC operation.

For an example, `Dst = VSHUFFLE Src0, Src1, Mode0` is equalant to a `shuffle mask of <0, 2, 4, ...>` on concatenated output of `Src0` and `Src1`.


llvm/lib/Target/AIE/aie2p/AIE2PInstructionSelector.cpp
```diff
@@ -4784,7 +4787,46 @@ bool AIE2PInstructionSelector::selectVST_FIFO(MachineInstr &I,
...
+        MIB.buildInstr(AIE2P::MOV_RLC_imm11_pseudo, {&AIE2P::eRRegClass}, {})
+            .addImm(4)
+            .getReg(0);
+    auto MI = MIB.buildInstr(AIE2P::VSHUFFLE_vec_shuffle_x, {DstReg},
+                             {SrcReg512Lo, SrcReg512Hi, ModeReg});
+    MI->dump();
+    I.eraseFromParent();
+    return constrainSelectedInstRegOperands(*MI, TII, TRI, RBI);
+  }
+  default:
```
martien-de-jong: We are called for Size >= 512, we only handle 1024?

niwinanto: Yes, this was work in progress. Now ready for review.

llvm/lib/Target/AIE/aie2p/AIE2PInstrPatterns.td
```diff
@@ -1145,3 +1145,56 @@ defm : Extract_512<i32, v64i8, (i32 c6u:$idx), VEXTRACT_8_vec_extract_imm_vaddSi
...
+
+// G_TRUNC
+
+// The G_TRUNC operation can be efficiently implemented using a VSHUFFLE
+// instruction. Modes 0, 2, and 4 correspond to the deinterleaved operation of
+// 1, 2, and 4 bytes respectively on the concatenated src0 and src1 input
+// vectors of the VSHUFFLE instruction.
+
+// src accumulator register bank
+def : Pat<(v16i32 (trunc (v16i64 ACC1024:$s1))),
```
konstantinschwarz: Looks like these patterns could be refactored into a common class?

llvm/test/CodeGen/AIE/extractelement.ll
```diff
@@ -72,10 +72,10 @@ define signext i8 @extract_v16i8_signext(<16 x i8> %v) nounwind {
 ; AIE2P-LABEL: extract_v16i8_signext:
 ; AIE2P:         .p2align 4
 ; AIE2P-NEXT:  // %bb.0:
-; AIE2P-NEXT:    ret lr
+; AIE2P-NEXT:    nopa ; nopb ; nops ; ret lr; nopm ; nopv
```
konstantinschwarz, line 75: We don't handle 128-bit vectors properly in the calling convention.
For AIE2, we have custom code to handle passing 128-bit vectors in 256-bit registers: https://github.com/Xilinx/llvm-aie/blob/aie-public/llvm/lib/Target/AIE/AIE2ISelLowering.cpp#L58

@SagarMaheshwari99 is working on a fix

llvm/lib/Target/AIE/AIELegalizerHelper.cpp
```diff
@@ -1529,4 +1529,36 @@ bool AIELegalizerHelper::legalizeG_AIE_EXTRACT_VECTOR_ELT(
   return true;
 }
 
+bool AIELegalizerHelper::legalizeG_TRUNC(LegalizerHelper &Helper,
+                                         MachineInstr &MI) const {
+  MachineIRBuilder &MIRBuilder = Helper.MIRBuilder;
+  MachineRegisterInfo &MRI = *MIRBuilder.getMRI();
+
+  const Register DstReg = MI.getOperand(0).getReg();
```
konstantinschwarz: Nit: can be written as:
const auto [DstReg, DstVecTy, SrcReg, SrcVecTy] = MI.getFirst2RegLLTs();

llvm/lib/Target/AIE/AIELegalizerHelper.cpp
```diff
@@ -1529,4 +1529,36 @@ bool AIELegalizerHelper::legalizeG_AIE_EXTRACT_VECTOR_ELT(
...
+  const LLT DstVecTy = MRI.getType(DstReg);
+  const unsigned SrcVecSize = SrcVecTy.getSizeInBits();
+
+  assert(SrcVecSize == 256 && "Expected G_TRUNC input vector size is 256!");
+
+  const AIEBaseInstrInfo *II = ST.getInstrInfo();
+  const unsigned PadOpc = II->getGenericPadVectorOpcode();
+  const unsigned UnpadOpc = II->getGenericUnpadVectorOpcode();
+
+  const LLT NewPadRegTy = LLT::fixed_vector(SrcVecTy.getNumElements() * 2,
```
konstantinschwarz: Should use
```
    const unsigned BasicVecSize = II()->getBasicVectorBitSize();
```
to get the size to pad

llvm/lib/Target/AIE/aie2p/AIE2PInstrPatterns.td
```diff
@@ -1145,3 +1145,56 @@ defm : Extract_512<i32, v64i8, (i32 c6u:$idx), VEXTRACT_8_vec_extract_imm_vaddSi
...
+  (EXTRACT_SUBREG
+    (VSHUFFLE_vec_shuffle_x ACC512:$s1, ACC512:$s1, (MOV_RLC_imm11_pseudo (i32 4))), 
+    sub_256_lo)>;
+
+def : Pat<(v16i16 (trunc (v16i32 ACC512:$s1))),
+  (EXTRACT_SUBREG
+    (VSHUFFLE_vec_shuffle_x ACC512:$s1, ACC512:$s1, (MOV_RLC_imm11_pseudo (i32 2))), 
+    sub_256_lo)>;
+
+// src vector register bank
```
konstantinschwarz: Why is there no pattern for
`v16i32 (trunc (v16i64 ...))`
on the vector register bank?

llvm/lib/Target/AIE/AIELegalizerHelper.cpp
```diff
@@ -1529,4 +1529,36 @@ bool AIELegalizerHelper::legalizeG_AIE_EXTRACT_VECTOR_ELT(
...
+  const Register SrcReg = MI.getOperand(1).getReg();
+  const LLT SrcVecTy = MRI.getType(SrcReg);
+  const LLT DstVecTy = MRI.getType(DstReg);
+  const unsigned SrcVecSize = SrcVecTy.getSizeInBits();
+
+  assert(SrcVecSize == 256 && "Expected G_TRUNC input vector size is 256!");
+
+  const AIEBaseInstrInfo *II = ST.getInstrInfo();
+  const unsigned PadOpc = II->getGenericPadVectorOpcode();
+  const unsigned UnpadOpc = II->getGenericUnpadVectorOpcode();
```
martien-de-jong: Conventionally called TII, TargetInstructionInfo

llvm/lib/Target/AIE/aie2p/AIE2PLegalizerInfo.cpp
```diff
@@ -256,7 +256,35 @@ AIE2PLegalizerInfo::AIE2PLegalizerInfo(const AIE2PSubtarget &ST)
       .clampScalar(0, S32, S32)
       .clampScalar(1, S32, S32);
 
-  getActionDefinitionsBuilder(G_TRUNC).alwaysLegal();
+  getActionDefinitionsBuilder(G_TRUNC)
+      .legalIf([=](const LegalityQuery &Query) {
+        const LLT &SrcTy = Query.Types[1];
+        const LLT &DstTy = Query.Types[0];
+        return SrcTy.isVector() && DstTy.isVector() &&
+               SrcTy.getSizeInBits() > 256 && SrcTy.getSizeInBits() < 2048 &&
```
martien-de-jong: nit: I think we can just as easily pin this down to 512 and 1024.

llvm/lib/Target/AIE/aie2p/AIE2PInstrPatterns.td
```diff
@@ -1145,3 +1145,56 @@ defm : Extract_512<i32, v64i8, (i32 c6u:$idx), VEXTRACT_8_vec_extract_imm_vaddSi
...
+// The G_TRUNC operation can be efficiently implemented using a VSHUFFLE
+// instruction. Modes 0, 2, and 4 correspond to the deinterleaved operation of
+// 1, 2, and 4 bytes respectively on the concatenated src0 and src1 input
+// vectors of the VSHUFFLE instruction.
+
+// src accumulator register bank
+def : Pat<(v16i32 (trunc (v16i64 ACC1024:$s1))),
+  (VSHUFFLE_vec_shuffle_x
+   (EXTRACT_SUBREG ACC1024:$s1, sub_512_acc_lo ),
+   (EXTRACT_SUBREG ACC1024:$s1, sub_512_acc_lo ),
```
andcarminati: What happens to the hi subreg?

llvm/lib/Target/AIE/aie2p/AIE2PInstructionSelector.cpp
```diff
@@ -4784,7 +4772,26 @@ bool AIE2PInstructionSelector::selectVST_FIFO(MachineInstr &I,
...
   return false;
 }
-
+bool AIE2PInstructionSelector::selectG_TRUNC(MachineInstr &I,
+                                             MachineRegisterInfo &MRI) {
+  Register SrcReg = I.getOperand(1).getReg();
+  LLT SrcTy = MRI.getType(SrcReg);
+  unsigned Size = SrcTy.getSizeInBits();
+  // G_TRUNC S32 <- S64
+  if (Size == 64) {
```
khallouh: Nit: SrcSize

llvm/test/CodeGen/AIE/aie2p/GlobalIsel/inst-select-trunc.mir
```diff
@@ -0,0 +1,151 @@
...
+#
+# This file is licensed under the Apache License v2.0 with LLVM Exceptions.
+# See https://llvm.org/LICENSE.txt for license information.
+# SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+#
+# (c) Copyright 2025 Advanced Micro Devices, Inc. or its affiliates
+# RUN: llc -mtriple aie2p -run-pass=instruction-select %s -verify-machineinstrs -o - | FileCheck %s
+
+---
+name:            v16s32_trunc_v16s64_acc1024
```
khallouh, line 11: Could you add instruction select tests for scalar source? I noticed we don't have them for AIE2P

llvm/lib/Target/AIE/aie2p/AIE2PInstructionSelector.cpp
```diff
@@ -4784,7 +4772,26 @@ bool AIE2PInstructionSelector::selectVST_FIFO(MachineInstr &I,
...
+    MachineInstrBuilder MI = MIB.buildInstr(TargetOpcode::COPY, {DstReg}, {})
+                                 .addReg(SrcReg, 0, AIE2P::sub_l_even);
+    I.eraseFromParent();
+    return selectCopy(*MI.getInstr(), MRI);
+  } else if (SrcTy.isVector()) {
+    assert(Size >= 512 && "Invalid vector size for G_TRUNC source vector!");
+    return selectImpl(I, *CoverageInfo);
+  } else {
+    I.setDesc(TII.get(TargetOpcode::COPY));
+    return selectCopy(I, MRI);
```
khallouh, line 5045: Curious, which legal case can we select to a simple copy?

llvm/test/CodeGen/AIE/aie2p/GlobalIsel/inst-select-trunc.mir
```diff
@@ -0,0 +1,151 @@
...
+    ; CHECK: [[DEF:%[0-9]+]]:acc1024 = IMPLICIT_DEF
+    ; CHECK-NEXT: [[MOV_RLC_imm11_pseudo:%[0-9]+]]:er = MOV_RLC_imm11_pseudo 4
+    ; CHECK-NEXT: [[COPY:%[0-9]+]]:mwbfsrs = COPY [[DEF]].sub_512_acc_lo
+    ; CHECK-NEXT: [[COPY1:%[0-9]+]]:mwbfsrs = COPY [[DEF]].sub_512_acc_lo
+    ; CHECK-NEXT: [[COPY2:%[0-9]+]]:mxm = COPY [[COPY1]]
+    ; CHECK-NEXT: [[COPY3:%[0-9]+]]:mxn = COPY [[COPY]]
+    ; CHECK-NEXT: [[VSHUFFLE_vec_shuffle_x:%[0-9]+]]:vec512 = VSHUFFLE_vec_shuffle_x [[COPY2]], [[COPY3]], [[MOV_RLC_imm11_pseudo]]
+    ; CHECK-NEXT: PseudoRET implicit $lr, implicit [[VSHUFFLE_vec_shuffle_x]]
+    %1:accregbank(<16 x s64>) = G_IMPLICIT_DEF
+    %0:vregbank(<16 x s32>) = G_TRUNC %1(<16 x s64>)
```
khallouh, line 26: What if it gets mapped to the accumulator bank? Could we select `VSHUFFLE_vec_shuffle_bm` in that case?

khallouh (commented):
Do we need custom handling of `G_TRUNC` in `RegBankSelect`? Not sure when we will map which bank (we don't have `regbankselect` tests). 
If `G_TRUNC` happens to be mapped to the accumulator bank, we won't be able to select with the current patterns.

llvm/lib/Target/AIE/aie2p/AIE2PInstrPatterns.td
niwinanto: You are absolutely right.

llvm/lib/Target/AIE/aie2p/AIE2PInstrPatterns.td
niwinanto: Its a good idea. Done.

llvm/lib/Target/AIE/AIELegalizerHelper.cpp
niwinanto: This is really nice. Done.

llvm/lib/Target/AIE/aie2p/AIE2PInstrPatterns.td
niwinanto: You are right. I missed it.

llvm/test/CodeGen/AIE/aie2p/GlobalIsel/inst-select-trunc.mir
niwinanto: Probably I would like to keep it for a followup PR, I want to focus only with vector operands with this PR.

llvm/lib/Target/AIE/aie2p/AIE2PInstructionSelector.cpp
niwinanto: Not from my head, need to investigate.

llvm/test/CodeGen/AIE/aie2p/GlobalIsel/inst-select-trunc.mir
niwinanto: You are right, accumulator bank is possible for this type and the pattern was missing. Included the pattern.

niwinanto:
> Do we need custom handling of `G_TRUNC` in `RegBankSelect`? Not sure when we will map which bank (we don't have `regbankselect` tests). If `G_TRUNC` happens to be mapped to the accumulator bank, we won't be able to select with the current patterns.

Irrespective of the source register bank, we should be able to instruction select the correct instruction. If accumulator register bank, cross bank copies will be inserted to use vector register bank. And for output type accumulator register bank, one pattern was missing and I did include that as well. Please see the tests.

llvm/test/CodeGen/AIE/aie2p/GlobalIsel/inst-select-trunc.mir
khallouh: We already have the test in aie2 `aie2/GlobalISel/inst-select-gtrunc.mir`, you could just enable it for AIE2P but up to you.

llvm/lib/Target/AIE/aie2p/AIE2PInstructionSelector.cpp
khallouh: I checked with Konstantin, this was meant for the trunc from 32 to 20 bits. Maybe we could (here or in a follow-up for scalars) check explicitly for that case and return false otherwise, so that we don't only fail later in CopyPhysReg for other unsupported cases.

niwinanto: So, lets do it it a followup PR along with scalar tests. AIECC-925

llvm/lib/Target/AIE/aie2p/AIE2PLegalizerInfo.cpp
```diff
@@ -256,7 +256,36 @@ AIE2PLegalizerInfo::AIE2PLegalizerInfo(const AIE2PSubtarget &ST)
...
+                   SrcTy.getElementType().getSizeInBits();
+      })
+      .legalIf([=](const LegalityQuery &Query) {
+        const LLT &SrcTy = Query.Types[1];
+        const LLT &DstTy = Query.Types[0];
+        return SrcTy.isScalar() && DstTy.isScalar();
+      })
+      .customIf([=](const LegalityQuery &Query) {
+        const LLT &SrcTy = Query.Types[1];
+        return SrcTy.isVector() && SrcTy.getSizeInBits() == 256;
```
khallouh: Do we want to check if the the Src size is double the Dst size like we do for the legal case above?

niwinanto:
> > Do we need custom handling of `G_TRUNC` in `RegBankSelect`? Not sure when we will map which bank (we don't have `regbankselect` tests). If `G_TRUNC` happens to be mapped to the accumulator bank, we won't be able to select with the current patterns.
> 
> Irrespective of the source register bank, we should be able to instruction select the correct instruction. If accumulator register bank, cross bank copies will be inserted to use vector register bank. And for output type accumulator register bank, one pattern was missing and I did include that as well. Please see the tests.

I did further investigate the register bank selection for G_TRUNC and found we always assign vector register bank to destination operand. So if we want to assign accumulator bank, we might need to use the use-def approach like we do for other special cases. However, we cannot select it using tablegen to *_bm variant of vshuffle, because `AIE2PAcc512RegisterClass` has only `[v16i32, v8i64, v16f32]` types. We might need to add `v32i16, v64i8` to the list. @konstantinschwarz @khallouh Do you think we should go for it? or ok with the current PR, that is one additional cross bank copy incase of accumulator destination operand. Please have a look at the new commit which has the registerbank tests.
Also, greedy register bank selection was failing because we had alternative mapping to gpr bank irrespective of vector/scalar.

khallouh:
> > > Do we need custom handling of `G_TRUNC` in `RegBankSelect`? Not sure when we will map which bank (we don't have `regbankselect` tests). If `G_TRUNC` happens to be mapped to the accumulator bank, we won't be able to select with the current patterns.
> > 
> > 
> > Irrespective of the source register bank, we should be able to instruction select the correct instruction. If accumulator register bank, cross bank copies will be inserted to use vector register bank. And for output type accumulator register bank, one pattern was missing and I did include that as well. Please see the tests.
> 
> I did further investigate the register bank selection for G_TRUNC and found we always assign vector register bank to destination operand. So if we want to assign accumulator bank, we might need to use the use-def approach like we do for other special cases. However, we cannot select it using tablegen to *_bm variant of vshuffle, because `AIE2PAcc512RegisterClass` has only `[v16i32, v8i64, v16f32]` types. We might need to add `v32i16, v64i8` to the list. @konstantinschwarz @khallouh Do you think we should go for it? or ok with the current PR, that is one additional cross bank copy incase of accumulator destination operand. Please have a look at the new commit which has the registerbank tests. Also, greedy register bank selection was failing because we had alternative mapping to gpr bank irrespective of vector/scalar.

For me its fine as is for now. We can keep the pattern and survive with the cross bank copies. We can optimize this in a follow-up. Maybe you could add a TODO and create a ticket to keep track of this.

llvm/lib/Target/AIE/aie2p/AIE2PRegisterBankInfo.cpp
```diff
@@ -381,6 +381,13 @@ AIE2PRegisterBankInfo::getInstrAlternativeMappings(
     AltMappings.push_back(&AccRegMapping);
     return AltMappings;
   }
+  case TargetOpcode::G_TRUNC: {
+    Register SrcReg = MI.getOperand(1).getReg();
+    LLT SrcTy = MRI.getType(SrcReg);
+    if (SrcTy.isVector())
+      return InstructionMappings();
```
khallouh: Nit: `return RegisterBankInfo::getInstrAlternativeMappings(MI); `

niwinanto:
> > > > Do we need custom handling of `G_TRUNC` in `RegBankSelect`? Not sure when we will map which bank (we don't have `regbankselect` tests). If `G_TRUNC` happens to be mapped to the accumulator bank, we won't be able to select with the current patterns.
> > > 
> > > 
> > > Irrespective of the source register bank, we should be able to instruction select the correct instruction. If accumulator register bank, cross bank copies will be inserted to use vector register bank. And for output type accumulator register bank, one pattern was missing and I did include that as well. Please see the tests.
> > 
> > 
> > I did further investigate the register bank selection for G_TRUNC and found we always assign vector register bank to destination operand. So if we want to assign accumulator bank, we might need to use the use-def approach like we do for other special cases. However, we cannot select it using tablegen to *_bm variant of vshuffle, because `AIE2PAcc512RegisterClass` has only `[v16i32, v8i64, v16f32]` types. We might need to add `v32i16, v64i8` to the list. @konstantinschwarz @khallouh Do you think we should go for it? or ok with the current PR, that is one additional cross bank copy incase of accumulator destination operand. Please have a look at the new commit which has the registerbank tests. Also, greedy register bank selection was failing because we had alternative mapping to gpr bank irrespective of vector/scalar.
> 
> For me its fine as is for now. We can keep the pattern and survive with the cross bank copies. We can optimize this in a follow-up. Maybe you could add a TODO and create a ticket to keep track of this.

Here is the ticket AIECC-927

khallouh (approved):
LGTM

PR #345, opened 2025-02-07
[AIE2/2P] Fix DONE latency at region end.

SagarMaheshwari99:


llvm/lib/Target/AIE/AIEBaseInstrInfo.h
```diff
@@ -176,6 +176,13 @@ struct AIEBaseInstrInfo : public TargetInstrInfo {
   }
   /// Check whether Opc represents a lock instruction
   virtual bool isLock(unsigned Opc) const { return false; }
+
+  /// Check whether Opc represents a DONE instruction
+  virtual bool isDone(unsigned Opc) const { return false; }
```
martien-de-jong: nit: I hope no one needs to call this from outside. This one could go and instances could be private or inlined.

martien-de-jong (dismissed):

llvm/lib/Target/AIE/AIEBaseInstrInfo.h
```diff
@@ -176,6 +176,13 @@ struct AIEBaseInstrInfo : public TargetInstrInfo {
...
+
+  /// Get "implicit" latency for special instructions.
+  virtual unsigned getImplicitLatency(const MachineInstr &) const { return 0; }
```
gbossu: Can you document what that means?

llvm/lib/Target/AIE/AIEBaseSubtarget.cpp
```diff
@@ -280,6 +280,10 @@ class RegionEndEdges : public ScheduleDAGMutation {
         EdgeLatency = DelaySlots + 1;
       }
 
+      // "Implicit" latency for special instructions.
+      const unsigned ImplicitLatency = TII->getImplicitLatency(MI);
+      EdgeLatency = std::max(EdgeLatency, ImplicitLatency);
```
gbossu, line 290: FYI @mludevid you might have done something similar when working on semaphores.

llvm/lib/Target/AIE/AIEBaseInstrInfo.h
andcarminati: I think we can have a design where we need just one method to be reimplemented. Something like:
```
virtual std::optional<unsigned> getDoneLatency(...)
```
In this case, we can have a base implementation for `getImplicitLatency` that can be reimplemented by the target only if needed.

martien-de-jong: @andcarminati, The idea was to have a more generic name excluding 'Done'. It's basically an extra latency that can be given to the exit edge. But yes, we would only have one generic method that fills it in.

andcarminati: Hi, my idea is to avoid encode the latency knowledge separately. If we have a new done, we have two places to change.

SagarMaheshwari99: The aim was to encode all "special latencies" into one "getImplicitLatency", so that we don't touch AIEBaseSubTarget frequently. Is there another way other than having a separate function for Opcode to achieve all in one getImplicitLatency?

SagarMaheshwari99: made private.

llvm/lib/Target/AIE/AIEBaseInstrInfo.h
SagarMaheshwari99: done.

llvm/lib/Target/AIE/AIEBaseInstrInfo.h
andcarminati: Just as a suggestion, I was thinking in something like:

```
// Return an optional latency if Opc is a done instruction.
std::optional<unsigned>AIE2InstrInfo::getDoneLatency(unsigned Opc) const {
 return (Opc == AIE2::DONE) ? 6 : std::nullopt
}

// This can be in the base class.
unsigned AIE2InstrInfo::getImplicitLatency(const MachineInstr &MI) const {
  if (auto OptLatency = getDoneLatency(MI.getOpcode()))
    return *OptLatency;

  return 0;
}
```
New targets can just implement `getDoneLatency` which will encode opcode + latency in the same place.

As I said, just a suggestion to have a single source of information here.

SagarMaheshwari99: Right I think you mean `if (auto OptLatency = getDoneLatency(MI.getOpcode()))`

SagarMaheshwari99: I have made the changes.

SagarMaheshwari99: I have removed this function.

llvm/lib/Target/AIE/aie2p/AIE2PInstrInfo.cpp
```diff
@@ -1219,6 +1219,12 @@ bool AIE2PInstrInfo::isLock(unsigned Opc) const {
   return false;
 }
 
+// Return an optional latency if Opc is DONE.
+std::optional<unsigned>
+AIE2PInstrInfo::getDoneLatency(const unsigned Opc) const {
+  return (Opc == AIE2P::DONE) ? std::optional<unsigned>(6) : std::nullopt;
```
F-Stuckmann, line 1227: nit: add a comment where the magic number comes from

SagarMaheshwari99: Done.

F-Stuckmann (approved):

PR #346, opened 2025-02-11
[AIEX] Combine G_SHUFFLE_VECTOR to COPY

katerynamuts:


llvm/lib/Target/AIE/AIECombinerHelper.cpp
```diff
@@ -2266,3 +2266,43 @@ bool llvm::matchShuffleToBroadcast(MachineInstr &MI, MachineRegisterInfo &MRI,
...
+  const LLT Src1Ty = MRI.getType(Src1Reg);
+  if (DstTy != Src1Ty)
+    return false;
+
+  const unsigned NumSrcElems = Src1Ty.isVector() ? Src1Ty.getNumElements() : 1;
+  if (Mask.size() != NumSrcElems)
+    return false;
+
+  // If the mask has only -1 (undef), do nothing
+  auto allUndefs = [&NumSrcElems](const ArrayRef<int> &Mask) -> bool {
```
martien-de-jong: capture by value?

llvm/test/CodeGen/AIE/GlobalISel/prelegalizercombiner-shuffle-vector.mir
```diff
@@ -907,24 +905,88 @@ body:             |
...
-# which is not implemented yet.
+
+# Combine G_SHUFFLE_VECTOR into COPY
+---
+name: shuffle_vector_to_copy_scalar
+tracksRegLiveness: true
+body:             |
+  bb.1:
+    ; CHECK-LABEL: name: shuffle_vector_to_copy_scalar
+    ; CHECK: [[C:%[0-9]+]]:_(s32) = G_CONSTANT i32 0
```
martien-de-jong, line 916: nit: not really a copy

llvm/test/CodeGen/AIE/GlobalISel/prelegalizercombiner-shuffle-vector.mir
```diff
@@ -907,24 +905,88 @@ body:             |
...
+    ; CHECK-NEXT: PseudoRET implicit $lr, implicit [[C]](s32)
+    %1:_(s32) = G_CONSTANT i32 0
+    %2:_(s32) = G_CONSTANT i32 1
+    %0:_(s32) = G_SHUFFLE_VECTOR %1(s32), %2(s32), shufflemask(0)
```
martien-de-jong, line 920: We don't use <1 x s32> ?

llvm/test/CodeGen/AIE/GlobalISel/prelegalizercombiner-shuffle-vector.mir
```diff
@@ -907,24 +905,88 @@ body:             |
...
-    ; CHECK-NEXT: [[C:%[0-9]+]]:_(s32) = G_CONSTANT i32 0
-    ; CHECK-NEXT: [[AIE_VSEL:%[0-9]+]]:_(<16 x s32>) = G_AIE_VSEL [[COPY]], [[COPY1]], [[C]](s32)
-    ; CHECK-NEXT: PseudoRET implicit $lr, implicit [[AIE_VSEL]](<16 x s32>)
+    ; CHECK-NEXT: PseudoRET implicit $lr, implicit [[COPY]](<16 x s32>)
     %1:_(<16 x s32>) = COPY $x0
     %2:_(<16 x s32>) = COPY $x1
     %0:_(<16 x s32>) = G_SHUFFLE_VECTOR %1(<16 x s32>), %2(<16 x s32>), shufflemask(0, 1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11, 12, 13, 14, 15)
     PseudoRET implicit $lr, implicit %0
 ...
+# Note: currently not commutable
```
martien-de-jong, line 939: I think this is exactly the same as matching a sawtooth with period, amplitude and height parameters that we saw recently.  May I suggest to factor that code out, and build the commutable version by checking the resulting height and select the appropriate operand?

martien-de-jong: Tip: I would create a base class that generalizes a mask match. The instances will fix free parameters, like the height in this example, while matching. A base class for a sawtooth has period, amplitude and height as free parameters. Specializations require for instance Period == amplitude. Different constructors can pre-fix specific free parameters.

llvm/test/CodeGen/AIE/GlobalISel/prelegalizercombiner-shuffle-vector.mir
```diff
@@ -907,24 +905,88 @@ body:             |
...
+    ; CHECK-NEXT: {{  $}}
+    ; CHECK-NEXT: [[COPY:%[0-9]+]]:_(<2 x s32>) = COPY $l0
+    ; CHECK-NEXT: [[COPY1:%[0-9]+]]:_(<2 x s32>) = COPY $l1
+    ; CHECK-NEXT: [[SHUF:%[0-9]+]]:_(<2 x s32>) = G_SHUFFLE_VECTOR [[COPY]](<2 x s32>), [[COPY1]], shufflemask(1, 2)
+    ; CHECK-NEXT: PseudoRET implicit $lr, implicit [[SHUF]](<2 x s32>)
+    %1:_(<2 x s32>) = COPY $l0
+    %2:_(<2 x s32>) = COPY $l1
+    %0:_(<2 x s32>) = G_SHUFFLE_VECTOR %1(<2 x s32>), %2(<2 x s32>), shufflemask(1, 2)
+    PseudoRET implicit $lr, implicit %0
+...
```
martien-de-jong, line 992: Perhaps add a boundary case test, e.g. a mask with eight elements, which only match in the first seven positions

llvm/lib/Target/AIE/AIECombinerHelper.cpp
andcarminati: `AllUndefs`.

llvm/test/CodeGen/AIE/GlobalISel/prelegalizercombiner-shuffle-vector.mir
martien-de-jong: A negative amplitude would indicate possibilities for shuffle modes that reverse.

llvm/lib/Target/AIE/AIECombinerHelper.cpp
```diff
@@ -2266,3 +2266,43 @@ bool llvm::matchShuffleToBroadcast(MachineInstr &MI, MachineRegisterInfo &MRI,
     return true;
   return false;
 }
+
+bool llvm::matchShuffleToCopy(MachineInstr &MI, MachineRegisterInfo &MRI,
+                              BuildFnTy &MatchInfo) {
+  assert(MI.getOpcode() == TargetOpcode::G_SHUFFLE_VECTOR);
+
+  const Register DstReg = MI.getOperand(0).getReg();
```
andcarminati, line 2339: Can we use something like this:

```
const auto [DstReg, DstTy, Src1Reg, Src1Ty] = MI.getFirst2RegLLTs();
```
?

andcarminati: But I guess that this could trigger some clang frontend bug....

llvm/lib/Target/AIE/AIECombinerHelper.cpp
```diff
@@ -2266,3 +2266,43 @@ bool llvm::matchShuffleToBroadcast(MachineInstr &MI, MachineRegisterInfo &MRI,
...
+        return false;
+    }
+    return true;
+  };
+
+  if (allUndefs(Mask))
+    return false;
+
+  // Check that the mask is sequential
+  for (unsigned I = 0; I < NumSrcElems; ++I) {
```
andcarminati: We also check sequential masks in `matchShuffleToVecEltBroadcast`. I feel that we can refactor some code here (good for new future combiners).

llvm/test/CodeGen/AIE/GlobalISel/prelegalizercombiner-shuffle-vector.mir
katerynamuts: For `<1 x s32>`, I'm getting the error: "invalid number of vector elements"

llvm/lib/Target/AIE/AIECombinerHelper.cpp
katerynamuts: With this I am getting the following error: " error: reference to local binding 'DstReg' declared in enclosing function 'llvm::matchShuffleToCopy'
  MatchInfo = [=](MachineIRBuilder &B) { B.buildCopy(DstReg, Src1Reg); };"

and the same error for `Src1Reg`.

andcarminati: I think that this is a clang bug.

martien-de-jong: That means that you can explain every line from the C++ standard related to structural bindings? Respect.

llvm/test/CodeGen/AIE/GlobalISel/prelegalizercombiner-shuffle-vector.mir
katerynamuts: Done

llvm/lib/Target/AIE/AIECombinerHelper.cpp
katerynamuts: Done

llvm/lib/Target/AIE/AIECombinerHelper.h
```diff
@@ -34,6 +34,44 @@ struct AIELoadStoreCombineMatchData {
...
+  unsigned Height = 0;
+  /// Negative amplitude can be used for reverse mask patterns.
+  int Amplitude = 1;
+};
+
+class SequentialPeriodicMaskMatch : public BaseMaskMatch {
+public:
+  SequentialPeriodicMaskMatch(unsigned Height, unsigned Period)
+      : BaseMaskMatch{Height, Period} {}
+};
```
martien-de-jong: These classes don't add much in this way. We could also have created BaseMaskMatch with the constructor used here.

llvm/lib/Target/AIE/AIECombinerHelper.h
```diff
@@ -34,6 +34,44 @@ struct AIELoadStoreCombineMatchData {
...
+  static std::optional<int> getUniqueIndex(ArrayRef<int> Mask);
+
+protected:
+  unsigned getMaskValue(unsigned Idx) const {
+    unsigned BaseIdx = Period == 0 ? Idx : Idx % Period;
+    return Height + BaseIdx * Amplitude;
+  }
+
+  unsigned Period = 0;
+  unsigned Height = 0;
```
martien-de-jong, line 60: I had imagined these as std::optional<unsigned>, indicating that they were free parameters. Then, the match function, isValidMask, would use this freedom to make it match the given mask. Matching on a mask without dont-care values, it would fix Height on the first element, the Amplitude on the second, and period when the sawtooth ended. It would return false the first time it couldn't adjust a free parameter. If you would know the period that you were looking for, you would initialize it with concrete values using a constructor, or a setter function.
I guess that for our purposes, this generalized sawtooth is a reasonable base, although we could make the base class even more generic. For instance, we could match the different transpose modes for VSHUFFLE in this form as well.

martien-de-jong: But don't pursue this idea at this point. It is already good to have it factored this far.

llvm/lib/Target/AIE/AIECombinerHelper.cpp
```diff
@@ -52,14 +52,75 @@ cl::opt<bool> CombineVecShiftByZero(
...
+bool BaseMaskMatch::isValidMask(ArrayRef<int> Mask) const {
+  bool FirstNotUndef = true;
+  for (unsigned Idx = 0; Idx < Mask.size(); ++Idx) {
+    if (Mask[Idx] == -1)
+      continue;
+
+    // Find the start value of the mask
+    if (FirstNotUndef) {
+      // Get the start value
+      const unsigned MaskStart = Mask[Idx] - Idx % Period;
```
niwinanto: I could reproduce the test fails with gcc build and the issue is here. `Period` can be zero and it results in divide by zero.

katerynamuts: Oh right, thanks, @niwinanto! After cleaning the ccache, I still couldn't reproduce it.

llvm/lib/Target/AIE/AIECombinerHelper.h
katerynamuts: Right, I removed the derived classes for now, if we need them, we can add them later.

llvm/lib/Target/AIE/AIECombinerHelper.cpp
```diff
@@ -52,14 +52,75 @@ cl::opt<bool> CombineVecShiftByZero(
...
+    if (Mask[Idx] == -1)
+      continue;
+
+    // Find the start value of the mask
+    if (FirstNotUndef) {
+      // Get the start value
+      const unsigned MaskStart = Mask[Idx] - Idx % (Period == 0 ? 1 : Period);
+
+      if (MaskStart != Height)
+        return false;
```
niwinanto: Can this be 
```
      if (Mask[Idx] != Idx % (Period == 0 ? 1 : Period))
        return false;
```

niwinanto: Discussed offline, it is clear to me now.

martien-de-jong (approved):

martien-de-jong:
I think I'll generalize this a bit more after it has been merged.

PR #349, opened 2025-02-12
[AIEX] Fix ExpandPostRAPseudos and PostSelectOptimize passes to satisfy MachineVerifier

khallouh:


llvm/lib/Target/AIE/AIEPostSelectOptimize.cpp
```diff
@@ -377,6 +377,23 @@ static bool tryToDuplicateLoadUse(
     MachineInstr *DefReg = MRI.getUniqueVRegDef(Reg);
 
     MachineBasicBlock::iterator InsertPoint = DefReg->getIterator();
+
+    // AIE's PseudoMove instruction takes compound register classes which
+    // contains registers of different sizes. We need to use the right classes
+    // to avoid the MachineVerifier complaining about mismatching sizes.
+    auto ConstrainAddrRegClass = [&](Register Reg) {
```
andcarminati: Maybe just:
```
return (MRI.constrainRegClass(Reg, TRI->getAddrAs32BitDCRegClass()) ||
          MRI.constrainRegClass(Reg, TRI->getAddrAs32BitDNRegClass()) ||
          MRI.constrainRegClass(Reg, TRI->getAddrAs32BitDJRegClass()) ||
          MRI.constrainRegClass(Reg, TRI->getAddrAs32BitMRegClass()) ||
          MRI.constrainRegClass(Reg, TRI->getAddrAs32BitPRegClass()));
```

llvm/lib/Target/AIE/aie2p/AIE2PInstrInfo.cpp
```diff
@@ -1162,8 +1162,7 @@ bool AIE2PInstrInfo::expandPostRAPseudo(MachineInstr &MI) const {
   case AIE2P::PseudoMove: {
     Register Dst = MI.getOperand(0).getReg();
     Register Src = MI.getOperand(1).getReg();
-    BuildMI(MBB, MI, DL, get(AIE2P::MOV_alu_mv_mv_mv_scl), Dst)
-        .addReg(Src, getKillRegState(true));
+    BuildMI(MBB, MI, DL, get(AIE2P::MOV_alu_mv_mv_mv_scl), Dst).addReg(Src);
```
andcarminati: What about keeping the existing state, like `getKillRegState(MI.getOperand(1).isKill())` ?

llvm/lib/Target/AIE/AIEPostSelectOptimize.cpp
```diff
@@ -377,6 +377,23 @@ static bool tryToDuplicateLoadUse(
...
+      if (MRI.constrainRegClass(Reg, TRI->getAddrAs32BitDCRegClass()) ||
+          MRI.constrainRegClass(Reg, TRI->getAddrAs32BitDNRegClass()) ||
+          MRI.constrainRegClass(Reg, TRI->getAddrAs32BitDJRegClass()) ||
+          MRI.constrainRegClass(Reg, TRI->getAddrAs32BitMRegClass()) ||
+          MRI.constrainRegClass(Reg, TRI->getAddrAs32BitPRegClass()))
```
andcarminati: I think M, N and J are enough here.

andcarminati:
Nice investigation! I would like recommend two things:

1. Update the dates in the test headers.
2. Enable machine verifier on the tests.

llvm/test/CodeGen/AIE/aie2p/GlobalIsel/inst-select-vconv-bfp16.mir
```diff
@@ -28,7 +28,7 @@ body:             |
     %1:accregbank(<64 x s32>) = COPY $dm0
     %2:vregbank(<64 x s8>) = COPY $x1
     %3:gprregbank(<8 x s8>) = COPY $e1
-    %0:vregbank(<64 x s8>), %4:gprregbank(<8 x s8>) = G_INTRINSIC_W_SIDE_EFFECTS intrinsic(@llvm.aie2p.v64accfloat.to.v64bfp16ebs16), %2(<64 x s8>), %3(<8 x s8>), %1(<64 x s32>)
```
martien-de-jong, line 31: Random copy paste junk? Do we still need %3 and %1?

llvm/lib/Target/AIE/aie2p/AIE2PInstrInfo.cpp
martien-de-jong: I guess we can't steal/copy the entire Src operand, even if we throw it away in the next line?

llvm/lib/Target/AIE/AIEBaseRegisterInfo.h
```diff
@@ -99,6 +99,22 @@ struct AIEBaseRegisterInfo : public TargetRegisterInfo {
...
+  }
+  virtual const TargetRegisterClass *getAddrAs32BitPRegClass() const {
+    llvm_unreachable("Target didn't implement getAddrAs32BitPRegClass!");
+  }
+  virtual const TargetRegisterClass *getAddrAs32BitMRegClass() const {
+    llvm_unreachable("Target didn't implement getAddrAs32BitMRegClass!");
+  }
+  virtual const TargetRegisterClass *getAddrAs32BitDCRegClass() const {
+    llvm_unreachable("Target didn't implement getAddrAs32BitDCRegClass!");
+  }
```
martien-de-jong: nit: It looks as if these may be AIE2XX specific. Perhaps put the entire constrain sequence in a virtual method, like constrainSubwordRegister?

andcarminati: I like this idea. +1.

llvm/test/CodeGen/AIE/aie2p/GlobalIsel/inst-select-vconv-bfp16.mir
khallouh: Yes, this test was wrong.  I fixed it. The input should have been a copy of $dm0 not $x1. the verify doesn't detect such problems :/

llvm/lib/Target/AIE/AIEBaseRegisterInfo.h
khallouh: Done

llvm/lib/Target/AIE/AIEPostSelectOptimize.cpp
khallouh: Removed P but kept C

llvm/lib/Target/AIE/AIEPostSelectOptimize.cpp
```diff
@@ -377,6 +377,13 @@ static bool tryToDuplicateLoadUse(
     MachineInstr *DefReg = MRI.getUniqueVRegDef(Reg);
 
     MachineBasicBlock::iterator InsertPoint = DefReg->getIterator();
+
+    // AIE's PseudoMove instruction takes compound register classes which
+    // contains registers of different sizes. We need to use the right classes
+    // to avoid the MachineVerifier complaining about mismatching sizes.
+    assert(TRI->constrainAddrRegClass(MRI, Reg) &&
+           "Expected an addressing register class");
```
gbossu: I'm not sure I understand the need for a new hook. Why can't we do the same as in ISel and constrain the operands to the largest superclass of the existing class, and the one in the instruction descriptor? We would then get a 32-bit class, because `MvScl` class is a compound 32-bit class.

khallouh: You are right. I updated to use `constrainSelectedInstRegOperands`

andcarminati (approved):
LGTM.

PR #360, opened 2025-02-17
[AIE2P] Legalize and select VMUL.f from G_FMUL

khallouh:


llvm/lib/Target/AIE/AIELegalizerHelper.cpp
```diff
@@ -1157,6 +1157,71 @@ bool AIELegalizerHelper::legalizeG_FABS(LegalizerHelper &Helper,
...
+  MachineRegisterInfo &MRI = *MIRBuilder.getMRI();
+
+  const Register DstReg = MI.getOperand(0).getReg();
+  Register SrcLHS = MI.getOperand(1).getReg();
+  Register SrcRHS = MI.getOperand(2).getReg();
+
+  assert(MRI.getType(DstReg) == LLT::scalar(16) &&
+         "Expected bfloat16 type in custom legalization.");
+
+  const LLT InsertVecLLT = ST.isAIE2P() ? V64BF16 : V64BF16;
```
andcarminati: Same type?

llvm/lib/Target/AIE/AIELegalizerHelper.cpp
```diff
@@ -1157,6 +1157,71 @@ bool AIELegalizerHelper::legalizeG_FABS(LegalizerHelper &Helper,
   return true;
 }
 
+bool AIELegalizerHelper::legalizeG_FMUL(LegalizerHelper &Helper,
+                                        MachineInstr &MI) const {
+  MachineIRBuilder &MIRBuilder = Helper.MIRBuilder;
+  MachineRegisterInfo &MRI = *MIRBuilder.getMRI();
+
+  const Register DstReg = MI.getOperand(0).getReg();
+  Register SrcLHS = MI.getOperand(1).getReg();
```
andcarminati: nit: const for both.

llvm/lib/Target/AIE/AIELegalizerHelper.cpp
khallouh: Yeah it's still a WIP. I didn't care about AIE2. I will remove. I still need to make some patterns work then I will update and make it ready for review.

llvm/test/CodeGen/AIE/aie2p/GlobalIsel/legalize-fmul.mir
```diff
@@ -0,0 +1,108 @@
...
+    ; CHECK-NEXT: [[C:%[0-9]+]]:_(s32) = G_CONSTANT i32 0
+    ; CHECK-NEXT: [[DEF:%[0-9]+]]:_(<64 x s16>) = G_IMPLICIT_DEF
+    ; CHECK-NEXT: [[AIE_INSERT_VECTOR_ELT:%[0-9]+]]:_(<64 x s16>) = G_AIE_INSERT_VECTOR_ELT [[DEF]], [[TRUNC]](s16), [[C]](s32)
+    ; CHECK-NEXT: [[AIE_INSERT_VECTOR_ELT1:%[0-9]+]]:_(<64 x s16>) = G_AIE_INSERT_VECTOR_ELT [[DEF]], [[TRUNC1]](s16), [[C]](s32)
+    ; CHECK-NEXT: [[DEF1:%[0-9]+]]:_(<32 x s32>) = G_IMPLICIT_DEF
+    ; CHECK-NEXT: [[BITCAST:%[0-9]+]]:_(<32 x s32>) = G_BITCAST [[AIE_INSERT_VECTOR_ELT]](<64 x s16>)
+    ; CHECK-NEXT: [[CONCAT_VECTORS:%[0-9]+]]:_(<64 x s32>) = G_CONCAT_VECTORS [[BITCAST]](<32 x s32>), [[DEF1]](<32 x s32>)
+    ; CHECK-NEXT: [[BITCAST1:%[0-9]+]]:_(<32 x s32>) = G_BITCAST [[AIE_INSERT_VECTOR_ELT1]](<64 x s16>)
+    ; CHECK-NEXT: [[CONCAT_VECTORS1:%[0-9]+]]:_(<64 x s32>) = G_CONCAT_VECTORS [[BITCAST1]](<32 x s32>), [[DEF1]](<32 x s32>)
+    ; CHECK-NEXT: [[FMUL:%[0-9]+]]:_(<64 x s32>) = G_FMUL [[CONCAT_VECTORS]], [[CONCAT_VECTORS1]]
```
khallouh: We also have vmul.f that operates on X registers, I will change to legalize and select that one instead for less register pressure.

llvm/lib/Target/AIE/AIELegalizerHelper.cpp
khallouh: I modify them

llvm/test/CodeGen/AIE/aie2p/GlobalIsel/legalize-fmul.mir
khallouh: Updated

llvm/lib/Target/AIE/aie2p/AIE2PInstrPatterns.td
```diff
@@ -40,6 +40,8 @@ class VecConf {
   int BMODE_16x16_b = 1;
   int BMODE_32x16   = 0;
```
martien-de-jong, line 41: Funny to have aliases here.

llvm/lib/Target/AIE/aie2p/AIE2PInstrPatterns.td
```diff
@@ -40,6 +40,8 @@ class VecConf {
...
 
+  int VARIANT_BF16xBF16_1_elem_1 = 1;
```
martien-de-jong: Sounds as if there are more variants. List them all in one go?

llvm/lib/Target/AIE/aie2p/AIE2PInstrPatterns.td
```diff
@@ -59,6 +61,7 @@ class VecConf {
 }
 
 def accfp32_vecconf : VecConf { let amode = AMODE_FP32; let bmode = BMODE_16x16; }
+def mulbf16_vecconf : VecConf { let amode = AMODE_FP32; let bmode = BMODE_16x16; let cmode = VARIANT_BF16xBF16_1_elem_1; }
```
martien-de-jong: Since this is a local definition, I wouldn't mind using CMODE as prefix.

llvm/lib/Target/AIE/aie2p/AIE2PInstrPatterns.td
```diff
@@ -222,6 +225,26 @@ def : Pat<(fadd ACC2048:$acc1, ACC2048:$acc2),
...
+                (VMUL_f_vmul_bf_vmul_bf_core_Y_Y VEC1024:$vec1, VEC1024:$vec2, (i32 mulbf16_vecconf.ConfBits)),
+                sub_1024_acc_lo)), 
+            sub_512_lo,
+            (VCONV_bf16_fp32_mv_x_srs_bf 
+              (EXTRACT_SUBREG
+                (VMUL_f_vmul_bf_vmul_bf_core_Y_Y VEC1024:$vec1, VEC1024:$vec2, (i32 mulbf16_vecconf.ConfBits)),
+                sub_1024_acc_hi)), 
+            sub_512_hi))>;
+
+def : Pat<(v32bf16 (fmul v32bf16:$vec1, v32bf16:$vec2)),
```
martien-de-jong, line 245: This isn't a standard legalization?

llvm/lib/Target/AIE/aie2p/AIE2PLegalizerInfo.cpp
```diff
@@ -225,12 +225,17 @@ AIE2PLegalizerInfo::AIE2PLegalizerInfo(const AIE2PSubtarget &ST)
 
   getActionDefinitionsBuilder(G_FABS).customFor({S16, S32, S64}).scalarize(0);
 
+  getActionDefinitionsBuilder(G_FMUL)
+      .legalFor({V64S16, V32S16})
+      .customFor({S16})
```
niwinanto, line 235: Do we need to retain `.clampScalar`?

llvm/test/CodeGen/AIE/GlobalISel/legalize-fmul.mir
```diff
@@ -5,7 +5,6 @@
 # (c) Copyright 2024 Advanced Micro Devices, Inc. or its affiliates
 
 # RUN: llc -mtriple aie2 -run-pass=legalizer %s -verify-machineinstrs -o - | FileCheck -DVER=2 --check-prefix=COMMON --check-prefix=AIE2 %s
-# RUN: llc -mtriple aie2p -run-pass=legalizer %s -verify-machineinstrs -o - | FileCheck -DVER=2p --check-prefix=COMMON --check-prefix=AIE2P %s
```
niwinanto, line 8: We still have AIE2P checkline in the test. You could also remove `-DVER=2 --check-prefix=COMMON`

llvm/test/CodeGen/AIE/aie2p/GlobalIsel/legalize-fmul.mir
```diff
@@ -0,0 +1,80 @@
+# NOTE: Assertions have been autogenerated by utils/update_mir_test_checks.py UTC_ARGS: --version 4
+# This file is licensed under the Apache License v2.0 with LLVM Exceptions.
+# See https://llvm.org/LICENSE.txt for license information.
+# SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+#
+# (c) Copyright 2024 Advanced Micro Devices, Inc. or its affiliates
+
+# RUN: llc -mtriple aie2p -run-pass=legalizer %s -verify-machineinstrs -o - | FileCheck %s
```
niwinanto, line 8: Would be nice to include the libcall tests as well.

llvm/lib/Target/AIE/aie2p/AIE2PLegalizerInfo.cpp
niwinanto, line 235: It would be nice to have a comment to explain why we would customize this for `s16`. I dont really get the context here.

khallouh: We have custom legalization for S16 now, no need to clamp it to S32/S64. Any other scalar should be illegal

khallouh: We don't have an instruction to multiply bf16 scalars, so instead of using an inefficient and potentially unsafe libcall (e.g. in the case of hardware loops) we need custom legalization by inserting the bf16 scalar into a vector, perform the element wise multiplication with `VMUL.f` and extract the bf16 scalar again. I can add this explanation as a comment.

martien-de-jong: It's the same as for FADD / FSUB. We implement a scalar multiplication by a full element by element vector mul.

llvm/lib/Target/AIE/aie2p/AIE2PInstrPatterns.td
khallouh: For this case, I don't know any but for the wider `v64bf16` case above we could possibly use `.fewerElements` to keep only one pattern. I will try it.

llvm/lib/Target/AIE/aie2p/AIE2PLegalizerInfo.cpp
niwinanto: I mean `.clampScalar(0, S16, S64)`

llvm/lib/Target/AIE/aie2p/AIE2PInstrPatterns.td
khallouh: I could but I'm not sure if we will ever be able to use all of them them in any patterns.

llvm/lib/Target/AIE/AIELegalizerHelper.cpp
```diff
@@ -1157,6 +1158,51 @@ bool AIELegalizerHelper::legalizeG_FABS(LegalizerHelper &Helper,
...
+         "Expected bfloat16 type in custom legalization.");
+
+  Register SrcLHS = MI.getOperand(1).getReg();
+  Register SrcRHS = MI.getOperand(2).getReg();
+
+  const LLT InsertVecLLT = V32BF16;
+  const unsigned InsertEltOpc =
+      ST.getInstrInfo()->getGenericInsertVectorEltOpcode();
+
+  const Register IdxReg = MIRBuilder.buildConstant(S32, 0).getReg(0);
```
martien-de-jong: Wouldn't it be cheaper to broadcast? Or is this picked up by a push.lo?

llvm/test/CodeGen/AIE/aie2p/GlobalIsel/legalize-fmul.mir
khallouh: We already didn't have them but I will add them while at it.

llvm/lib/Target/AIE/aie2p/AIE2PLegalizerInfo.cpp
khallouh: but why? the only float type under 16 bits we have is bfloat (aka S16)

llvm/lib/Target/AIE/AIELegalizerHelper.cpp
```diff
@@ -1157,6 +1158,51 @@ bool AIELegalizerHelper::legalizeG_FABS(LegalizerHelper &Helper,
...
+
+  MachineIRBuilder &MIRBuilder = Helper.MIRBuilder;
+  MachineRegisterInfo &MRI = *MIRBuilder.getMRI();
+
+  const Register DstReg = MI.getOperand(0).getReg();
+  assert(MRI.getType(DstReg) == LLT::scalar(16) &&
+         "Expected bfloat16 type in custom legalization.");
+
+  Register SrcLHS = MI.getOperand(1).getReg();
+  Register SrcRHS = MI.getOperand(2).getReg();
```
martien-de-jong, line 1173: I think the bulk of this code could be shared with other scalar binary legalizers.

llvm/lib/Target/AIE/aie2p/AIE2PLegalizerInfo.cpp
niwinanto: True. Just pointing that we deviate from old behavior, s128 to s64 or s8 to s32. But you are right, it does not make sense for these types.

llvm/lib/Target/AIE/aie2p/AIE2PInstrPatterns.td
```diff
@@ -222,6 +225,26 @@ def : Pat<(fadd ACC2048:$acc1, ACC2048:$acc2),
 def : Pat<(fsub ACC2048:$acc1, ACC2048:$acc2),
           (VSUB_f_vmac_cm2_add_reg ACC2048:$acc1, ACC2048:$acc2, (i32 accfp32_vecconf.ConfBits))>;
 
+// MUL
+def : Pat<(v64bf16 (fmul v64bf16:$vec1, v64bf16:$vec2)),
```
andcarminati, line 232: Check: We are performing the same multiplication twice: one for extract lo and other to extract hi. I guess we cannot express an optimized reuse of the same VMUL here, right?

llvm/lib/Target/AIE/aie2p/AIE2PInstrPatterns.td
martien-de-jong: It looks like the translation of a hardware enumeration into tablegen speak. I'm hoping that one day we'll have a single point of definition for these, and the full list would make them more recognisable.

llvm/lib/Target/AIE/aie2p/AIE2PInstrPatterns.td
khallouh: I don't know if we can do this in the pattern but CSE takes care of it

llvm/lib/Target/AIE/aie2p/AIE2PInstrPatterns.td
khallouh: Discussed offline: If we split in the legalizer we will need 2 multiplications (X_X variant) one for each part whereas if we do it here in ISel we can use just one multiplication using the Y_Y variant

llvm/lib/Target/AIE/AIELegalizerHelper.cpp
khallouh: a vbcst would be cheaper than vinsert because it doesn't require a gpr for the index (since we currently don't select the vinsert with immediate zero instruction) plus it would require only 2 vector registers instead of 3. I will change to use vbcst

llvm/test/CodeGen/AIE/GlobalISel/legalize-fmul.mir
khallouh: Done

llvm/test/CodeGen/AIE/aie2p/GlobalIsel/legalize-fmul.mir
khallouh: Added those

llvm/lib/Target/AIE/AIELegalizerHelper.cpp
khallouh: Implemented using VBCST

llvm/lib/Target/AIE/aie2p/AIE2PInstrPatterns.td
khallouh: ran CSE in the selection test

martien-de-jong (approved):

PR #361, opened 2025-02-17
[AIEX] Bail out from waw rewriter for spilled register

niwinanto:
There is an upstream bug which stop us from  unassigning the spilled register after greedy. So we bail out for such cases from aie-waw-rewriter pass.
https://github.com/llvm/llvm-project/issues/48255


llvm/lib/Target/AIE/AIEWawRegRewriter.cpp
```diff
@@ -33,6 +33,7 @@
 #include "llvm/CodeGen/TargetSubtargetInfo.h"
 #include "llvm/MC/MCRegister.h"
 #include "llvm/Support/Debug.h"
+#include "llvm/Support/ErrorHandling.h"
```
andcarminati: Is this include necessary?

llvm/test/CodeGen/AIE/aie2p/ra/waw_reg_renaming_block_spill.mir
```diff
@@ -0,0 +1,194 @@
+# NOTE: Assertions have been autogenerated by utils/update_mir_test_checks.py UTC_ARGS: --version 4
+# NOTE: Example file for Write After Write Register Renaming in Loop test
+#
+# This file is licensed under the Apache License v2.0 with LLVM Exceptions.
+# See https://llvm.org/LICENSE.txt for license information.
+# SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+#
+# (c) Copyright 2025 Advanced Micro Devices, Inc. or its affiliates
+
+# Workaround WAW register renaming pass for llvm bug #48911. We must not crash here.
```
F-Stuckmann: nit: Add a comment which register is now skipped due to being a spill register

F-Stuckmann:
LGTM

llvm/lib/Target/AIE/AIEWawRegRewriter.cpp
```diff
@@ -253,7 +253,12 @@ bool AIEWawRegRewriter::renameMBBPhysRegs(const MachineBasicBlock *MBB) {
       // WAW dependency resolution
       if (LastVRegDef[Reg] != &MI)
         continue;
-
+      // See llvm bug #48911.
+      // Skip reassign if a register has originated from such split.
```
martien-de-jong: I would appreciate a brief inline explanation of 'such split' if possible/feasible.

martien-de-jong (approved):

llvm/lib/Target/AIE/AIEWawRegRewriter.cpp
```diff
@@ -253,7 +253,12 @@ bool AIEWawRegRewriter::renameMBBPhysRegs(const MachineBasicBlock *MBB) {
       // WAW dependency resolution
       if (LastVRegDef[Reg] != &MI)
         continue;
-
+      // See llvm bug #48911.
+      // Skip reassign if a register has originated from InlineSpiller.
+      // FIXME: Remove the workaround when bug #48911 is fixed.
+      if (VRM->getPreSplitReg(Reg)) {
+        continue;
+      }
```
gbossu, line 261: I'm wondering if we shouldn't abort altogether? It seems dangerous to do further manipulations to the LiveRegMatrix if isn't up to date. Or is this really safe to just skip this register, but continue with the other ones?

PR #362, opened 2025-02-18
[AIE2p] Register re-allocation for BFP16 regs

gbossu:
Enable register re-allocation for EX/EY registers. Needed for Conv2D_bfp16

### DIFF
--- a/llvm/lib/Target/AIE/aie2p/AIE2PRegisterInfo.cpp
+++ b/llvm/lib/Target/AIE/aie2p/AIE2PRegisterInfo.cpp
@@ -591,6 +591,11 @@ bool AIE2PRegisterInfo::isVecOrAccRegClass(
   if (AIE2P::ACC2048RegClass.hasSubClassEq(&RC))
     return true;
 
+  // BFP16 registers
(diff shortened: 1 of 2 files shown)


llvm/lib/Target/AIE/aie2p/AIE2PRegisterInfo.cpp
```diff
@@ -591,6 +591,11 @@ bool AIE2PRegisterInfo::isVecOrAccRegClass(
   if (AIE2P::ACC2048RegClass.hasSubClassEq(&RC))
     return true;
 
+  // BFP16 registers
+  if (AIE2P::VEC576RegClass.hasSubClassEq(&RC) ||
+      AIE2P::eEYRegClass.hasSubClassEq(&RC))
```
F-Stuckmann, line 596: nice, we're also checking the "double registers"

F-Stuckmann (approved):

llvm/lib/Target/AIE/aie2p/AIE2PRegisterInfo.cpp
khallouh: Nit: define a new register class `VEC1152` with `eEY` and use it here?

PR #367, opened 2025-02-24
[AIEX] Prevent libcall in ZOL.

martien-de-jong:
This is not in full detail, but I guess it will cover most cases for now. It can be easily adjusted and extended.

It would still be nice if we could lower away ZOL between instruction selection and register allocation, which should be feasible.


llvm/lib/Target/AIE/AIE2TargetTransformInfo.cpp
```diff
@@ -190,3 +44,22 @@ AIE2TTIImpl::instCombineIntrinsic(InstCombiner &IC, IntrinsicInst &II) const {
   }
   return std::nullopt;
 }
+
```
martien-de-jong, line 47: These are just small wrappers. We can differentiate by using a derived class for Common, adding virtuals if required.

llvm/lib/Target/AIE/AIEBaseTargetTransformInfo.cpp
```diff
@@ -0,0 +1,197 @@
...
+  case Instruction::FAdd:
+  case Instruction::FSub:
+  case Instruction::FMul:
+  case Instruction::FNeg:
+    return Ty->isBFloatTy();
+  case Instruction::FCmp:
+    return I.getOperand(0)->getType()->isBFloatTy();
+  case Instruction::FDiv:
+  case Instruction::FPExt:
+  case Instruction::FPTrunc:
```
martien-de-jong: Future: bfloat16 <-> float32 could probably be handled by vector insert/extract.

llvm/lib/Target/AIE/AIEBaseTargetTransformInfo.h
```diff
@@ -55,6 +72,12 @@ template <typename T> class AIEBaseTTIImpl : public BasicTTIImplBase<T> {
     // cost?
     return TTI::TCC_Basic;
   }
+  void adjustUnrollingPreferences(Loop *L, ScalarEvolution &SE,
```
andcarminati, line 75: This is nice! This refactor is more than welcome!

llvm/lib/Target/AIE/AIEBaseTargetTransformInfo.cpp
```diff
@@ -0,0 +1,197 @@
...
+  Type *Ty = I.getType();
+  if (Ty->isDoubleTy()) {
+    return false;
+  }
+  switch (I.getOpcode()) {
+  case Instruction::FAdd:
+  case Instruction::FSub:
+  case Instruction::FMul:
+  case Instruction::FNeg:
+    return Ty->isBFloatTy();
```
niwinanto: Do you think we should allow for vector types?

llvm/test/CodeGen/AIE/aie2p/zol/no-rt-calls.ll
```diff
@@ -0,0 +1,37 @@
...
+; RUN: llc -O2 --mtriple=aie2p %s -o - --stop-before=irtranslator | \
+; RUN:   FileCheck %s
+
+; We have float operations in the loop body, and we check whether ZOL loops
+; were created by looking for the hallmark intrinsics
+
+; CHECK-NOT: llvm.set.loop.iterations
+; CHECK-NOT: llvm.loop.decrement
+
+define dso_local float @factorial(i32 noundef %n) local_unnamed_addr {
```
andcarminati, line 17: I think we can also have a test for the positive case, when we generate a HL for bfloat mul, form example. Not exactly in this file ;-)

llvm/lib/Target/AIE/aie2p/AIE2PTargetTransformInfo.cpp
```diff
@@ -9,124 +9,26 @@
...
-    LLVM_DEBUG(dbgs() << "AIE Loops: Loop is JNZD candidate\n");
-    HWLoopInfo.CounterInReg = true;
-  }
-  return true;
+  return Common.isHardwareLoopProfitable(L, SE, AC, LibInfo, HWLoopInfo);
 }
+
+bool AIE2PTTIImpl::isProfitableOuterLSR(const Loop &L) const {
+  return Common.isProfitableOuterLSR(L);
+}
```
andcarminati: nit: new line.

llvm/lib/Target/AIE/AIEBaseTargetTransformInfo.cpp
martien-de-jong: Oh, you're right!

llvm/lib/Target/AIE/AIEBaseTargetTransformInfo.cpp
```diff
@@ -0,0 +1,205 @@
...
+  Type *Ty = I.getType();
+  if (Ty->getScalarType()->isDoubleTy()) {
+    return false;
+  }
+  switch (I.getOpcode()) {
+  case Instruction::FAdd:
+  case Instruction::FSub:
+  case Instruction::FMul:
+  case Instruction::FNeg:
+    if (!Ty->getScalarType()->isBFloatTy()) {
```
andcarminati, line 73: Wouldn't it be better to always allow for ZOL for vector types? As standard libcalls are always for scalar types. In this case we can remove this getScalarType.

llvm/lib/Target/AIE/AIEBaseTargetTransformInfo.cpp
martien-de-jong: Oh, it already is. Will allow it here.

llvm/lib/Target/AIE/AIEBaseTargetTransformInfo.cpp
martien-de-jong: Ehm, perhaps we're not on the same page. getScalarType() is a no-op for scalartypes, and looks though vector types to get to the element type. 
Vectors with unsupported element types will still be lowered to libcalls. This phrase seems to be the idiom to indicate a particular scalartype, be it stand-alone of in a vector.

llvm/lib/Target/AIE/AIE2TargetTransformInfo.cpp
```diff
@@ -190,3 +44,22 @@ AIE2TTIImpl::instCombineIntrinsic(InstCombiner &IC, IntrinsicInst &II) const {
...
+bool AIE2TTIImpl::isHardwareLoopProfitable(Loop *L, ScalarEvolution &SE,
+                                           AssumptionCache &AC,
+                                           TargetLibraryInfo *LibInfo,
+                                           HardwareLoopInfo &HWLoopInfo) {
+  return Common.isHardwareLoopProfitable(L, SE, AC, LibInfo, HWLoopInfo);
+}
+
+bool AIE2TTIImpl::isProfitableOuterLSR(const Loop &L) const {
+  return Common.isProfitableOuterLSR(L);
+}
```
andcarminati: super last nit: new line.

andcarminati (approved):
LGTM.

PR #368, opened 2025-02-24
[AIE2P] Implement build subvector with vsel

martien-de-jong:


llvm/lib/Target/AIE/AIECombinerHelper.cpp
```diff
@@ -1828,6 +1828,20 @@ bool llvm::matchBroadcastElement(MachineInstr &MI, MachineRegisterInfo &MRI,
   return true;
 }
 
+static void buildUnmergeVector(MachineIRBuilder &B, MachineRegisterInfo &MRI,
+                               Register DstReg, Register SrcReg,
+                               unsigned NumSubVectors, unsigned SubIdx) {
+  const LLT DstTy = MRI.getType(DstReg);
```
andcarminati, line 1822: This is present in the other PR as well. I think we could assert that `NumSubVectors > 1`, otherwise  `buildUnmerge` can crash in a strange way.

andcarminati:
It would be nice to have some QoR numbers for this PR.

llvm/lib/Target/AIE/AIECombinerHelper.cpp
martien-de-jong: Agree, although It looks as if buildUnmerge has the same assert. That doesn't look like 'in a strange way' to me.

llvm/lib/Target/AIE/AIECombinerHelper.cpp
```diff
@@ -1841,25 +1855,36 @@ bool llvm::matchShuffleToVSel(MachineInstr &MI, MachineRegisterInfo &MRI,
...
   const Register Src1Reg = MI.getOperand(1).getReg();
   const Register Src2Reg = MI.getOperand(2).getReg();
   ArrayRef<int> Mask = MI.getOperand(3).getShuffleMask();
 
   const LLT DstTy = MRI.getType(DstReg);
   const LLT Src1Ty = MRI.getType(Src1Reg);
-  if (Src1Ty.getSizeInBits() != 512 ||
-      Src1Ty.getElementType() == LLT::scalar(64))
+  if (Src1Ty.getSizeInBits() != BasicVectorBitSize ||
+      Src1Ty.getElementType() == LLT::scalar(DoubleScalarRegSize))
```
khallouh: I suppose this is because we can't directly select VSEL.64 which the architecture lacks. Could we then return false for anything bigger than 32? (e.g. 128 size is not supported either)

llvm/lib/Target/AIE/AIECombinerHelper.cpp
```diff
@@ -1841,25 +1855,36 @@ bool llvm::matchShuffleToVSel(MachineInstr &MI, MachineRegisterInfo &MRI,
...
-  if (NumDstElems != NumSrcElems)
+  if (NumDstElems > NumSrcElems)
     return false;
 
   if (MaskMatch::isMaskWithAllUndefs(Mask))
     return false;
 
+  std::optional<unsigned> MaskHeight = MaskMatch::getHeight(Mask, /*Period*/ 0);
+  if (!MaskHeight)
+    return false;
```
khallouh, line 1875: Can we now get rid of the `isMaskWithAllUndefs` check? because getHeight also checks if we have a mask with all Undefs and it will return nullptr in that case.

llvm/lib/Target/AIE/AIECombinerHelper.cpp
```diff
@@ -1868,8 +1893,8 @@ bool llvm::matchShuffleToVSel(MachineInstr &MI, MachineRegisterInfo &MRI,
   // order.
   uint64_t DstMask = 0;
   const size_t NumMaskElems = Mask.size();
-  for (unsigned I = 0; I < NumMaskElems; I++) {
-    int Idx = Mask[I];
+  for (unsigned I = *MaskHeight; I < *MaskHeight + NumMaskElems; I++) {
+    const int Idx = Mask[I];
```
khallouh: Discussed offline, starting at MaskHeight would go out of the mask range. This is not needed to handle this vsel pattern anyway (when number of src elements is bigger than dst elements), so any generalization (potentially using VSHIFT) can be added later if needed.

llvm/lib/Target/AIE/AIECombinerHelper.cpp
martien-de-jong: Right. And actually this is not defined by DoubleScalarRegSize, which is only logically restricting the size of the vsel condition vector.

llvm/lib/Target/AIE/AIECombinerHelper.cpp
martien-de-jong: Agree. VSEL can only use I or I+NumSrcElems. I will simplify this code.

llvm/lib/Target/AIE/AIECombinerHelper.cpp
martien-de-jong: We don't use getHeight anymore. I leave the check in, because I'm unsure about combine priorities. AS indicated in the comment, the entire instruction should be eliminated.

khallouh (approved):
LGTM

PR #369, opened 2025-02-24
[AIEX] Enhance combiner to support post-inc-load requiring moving multiple instructions 

krishnamtibrewala:
Enhance the combiner to identify the potential to move multiple dependent instructions below G_PTR_ADD to enable post-inc combining.

Say between a dependent LOAD & PTR_ADD you have multiple instructions that use the value loaded by the LOAD instruction.

First we try to identify the dependent instructions chain, which uses the definition of LOAD, and see if we can move the dependency chain below PTR_ADD such that we open up the potential to combine LOAD & PTR_ADD to POST_INC_LOAD.


llvm/lib/Target/AIE/AIECombinerHelper.cpp
```diff
@@ -196,6 +202,73 @@ bool llvm::canDelayMemOp(MachineInstr &MemI, MachineInstr &Dest,
...
+/// instructions.
+void DependentInstrInRange(MachineInstr &SrcMI, MachineInstr &DestMI,
+                           MachineRegisterInfo &MRI,
+                           std::vector<MachineInstr *> &DependentInstrs) {
+  auto SrcMIIter = SrcMI.getIterator();
+  auto DestMIIter = DestMI.getIterator();
+  for (auto &Defs : SrcMI.all_defs()) {
+    const Register Reg = Defs.getReg();
+    for (auto &Use : MRI.use_nodbg_instructions(Reg)) {
+      if (Use.getParent() != DestMI.getParent())
```
martien-de-jong, line 310: It would be nice to document a precondition that Src and Dest should be in the same block and assert it at entry. Perhaps also that Src dominates Dest

llvm/lib/Target/AIE/AIECombinerHelper.cpp
```diff
@@ -196,6 +202,73 @@ bool llvm::canDelayMemOp(MachineInstr &MemI, MachineInstr &Dest,
...
+/// be moved just before \a Dest in order to allow post-increment combining
+bool llvm::canDelayMemOpWithExtraInstr(
+    MachineInstr &MemI, MachineInstr &Dest, MachineRegisterInfo &MRI,
+    std::vector<MachineInstr *> &DependentInstrs) {
+  if (MemI.getParent() != Dest.getParent())
+    return false;
+  DependentInstrInRange(MemI, Dest, MRI, DependentInstrs);
+
+  if (DependentInstrs.empty())
+    return false;
```
martien-de-jong: Isn't 'no dependent instructions' the ideal case? Why return false?

krishnamtibrewala: Basically if its empty then yes there are no dependent case and should be ideal but that case is covered by the previous "else if" case.

https://github.com/Xilinx/llvm-aie/pull/369/files#diff-3b392dc8a295886ef3ff3adbae88fa9158036aef8fd17ef8c571b370a0d3f21c:~:text=(canDelayMemOp(MemI%2C%20PtrAddInsertLoc%2C%20MRI))

llvm/lib/Target/AIE/AIECombinerHelper.cpp
```diff
@@ -196,6 +202,73 @@ bool llvm::canDelayMemOp(MachineInstr &MemI, MachineInstr &Dest,
...
+
+  // Sort DependentInstrs in descending order based on distance from MemI, this
+  // is required since func DependentInstrInRange(...) collects the dependent
+  // instructions in order based on defs and use of those defs.
+  auto MemIIter = MemI.getIterator();
+  std::sort(DependentInstrs.begin(), DependentInstrs.end(),
+            [MemIIter](MachineInstr *A, MachineInstr *B) {
+              return std::distance(MemIIter, A->getIterator()) >
+                     std::distance(MemIIter, B->getIterator());
```
martien-de-jong: It sounds expensive  (> O(N^2)) to call this from a sort routine. I guess we could precompute a distance map. In fact, DependentInstrInRange has the distance to Dest available, so it could fill a map rather than a vector.

llvm/lib/Target/AIE/AIECombinerHelper.cpp
```diff
@@ -196,6 +202,73 @@ bool llvm::canDelayMemOp(MachineInstr &MemI, MachineInstr &Dest,
...
+  if (!canDelayMemOp(*DependentInstrs.front(), Dest, MRI))
+    return false;
+  // Check if we can move the rest of the dependent instructions after the
+  // previous one, this inherently checks if we can move the dependent
+  // instruction below the Dest instruction
+  for (size_t i = 0; i < DependentInstrs.size() - 1; ++i)
+    if (!canDelayMemOp(*DependentInstrs[i + 1], *DependentInstrs[i], MRI))
+      return false;
+
+  return true;
```
martien-de-jong, line 346: What I don't like here is that we have a very layered view of dependences. I think the basic condition is: the destination should not be reachable from the source in the dependence graph over any of DependentInstrs. `canDelayMemOp()` is one way of coding a dependence edge in that graph, and the use def chains are another one.
Also, the formulation is not the most direct.  I think you actually want to know `canAdvanceDest(MI, DestMI)` for all MI in DependentInstrs

martien-de-jong: Src -> I1 -> I2 -> I3 
Src -> Dst

krishnamtibrewala: What I have tried to check here using the existing functions is if I can pack these dependence graphs together.

By together here, I mean next to each other.

Using the same  Src -> I1 -> I2 -> I3  andSrc -> Dstexample but with others Ix in between Src/Des/I1-3, I check if it's safe to pack them as Src , I1, I2, I3 , Dst in mist, of Ix if so, I move belowI1, I2, I3athe Dst and Src just above Dst

The reason for taking this approach was that I1 -> I2 -> I3 could only be moved as a group, and checking that in a single shot without actually moving them in the "match" phase of the combiner would have been tricky.

llvm/lib/Target/AIE/AIECombinerHelper.cpp
```diff
@@ -196,6 +202,88 @@ bool llvm::canDelayMemOp(MachineInstr &MemI, MachineInstr &Dest,
...
+      if (std::distance(SrcMIIter, UseIter) >
+          std::distance(SrcMIIter, DestMIIter))
+        break;
+      DependentInstrs.push_back(&Use);
+      DependentInstrInRange(Use, DestMI, MRI, DependentInstrs);
+    }
+  }
+}
+
+/// \return true and \a DependentInstrs vecto to move below Dest, if \a MemI can
```
andcarminati: nit: vecto

llvm/lib/Target/AIE/AIECombinerHelper.cpp
```diff
@@ -196,6 +202,88 @@ bool llvm::canDelayMemOp(MachineInstr &MemI, MachineInstr &Dest,
...
+  for (auto *MI : DependentInstrs) {
+    int Distance = std::distance(MemIIter, MI->getIterator());
+    assert(Distance >= 0 && "Distance should be positive");
+    InstrWithDistances.emplace_back(MI, Distance);
+  }
+
+  // Sort DependentInstrs in descending order based on distance from MemI, this
+  // is required since func DependentInstrInRange(...) collects the dependent
+  // instructions in order based on defs and use of those defs.
+  std::sort(InstrWithDistances.begin(), InstrWithDistances.end(),
```
andcarminati: Could we use dominance relation to in-place sort instructions directly on DependentInstrs? We could eliminate two fors here.

llvm/lib/Target/AIE/AIECombinerHelper.cpp
```diff
@@ -196,6 +202,88 @@ bool llvm::canDelayMemOp(MachineInstr &MemI, MachineInstr &Dest,
...
+    DependentInstrs.push_back(MI);
+
+  // Check if we can move the dependent instructions after Dest in order to
+  // enable post-increment combining.
+  if (!canDelayMemOp(*DependentInstrs.front(), Dest, MRI))
+    return false;
+  // Check if we can move the rest of the dependent instructions after the
+  // previous one, this inherently checks if we can move the dependent
+  // instruction below the Dest instruction
+  for (size_t i = 0; i < DependentInstrs.size() - 1; ++i)
```
andcarminati: nit: I = 0 ....

llvm/lib/Target/AIE/AIECombinerHelper.cpp
```diff
@@ -196,6 +202,88 @@ bool llvm::canDelayMemOp(MachineInstr &MemI, MachineInstr &Dest,
...
+            });
+
+  // Update DependentInstrs with sorted instructions
+  DependentInstrs.clear();
+  for (const auto &[MI, Distance] : InstrWithDistances)
+    DependentInstrs.push_back(MI);
+
+  // Check if we can move the dependent instructions after Dest in order to
+  // enable post-increment combining.
+  if (!canDelayMemOp(*DependentInstrs.front(), Dest, MRI))
```
andcarminati: I think this check is the only that matters here. Maybe you can add a test that fails on the next for.

krishnamtibrewala: Only this wont be sufficient (kingly note the front() will give the last instruction in the dependency chain)
Why we need the next check is explained here https://github.com/Xilinx/llvm-aie/pull/369#discussion_r1968031152

llvm/lib/Target/AIE/AIECombinerHelper.cpp
krishnamtibrewala: Thanks for the suggestion @andcarminati  have created a `set` that uses dominance relation as comparison to in-place sort instructions directly on DependentInstrs.

krishnamtibrewala:
Hi @andcarminati can you re-review this again. Thank you.

llvm/lib/Target/AIE/AIECombinerHelper.cpp
```diff
@@ -277,6 +285,67 @@ bool llvm::canDelayMemOp(MachineInstr &MemI, MachineInstr &Dest,
...
+  auto SrcMIIter = SrcMI.getIterator();
+  auto DestMIIter = DestMI.getIterator();
+  for (auto &Defs : SrcMI.all_defs()) {
+    const Register Reg = Defs.getReg();
+    for (auto &Use : MRI.use_nodbg_instructions(Reg)) {
+      if (Use.getParent() != DestMI.getParent())
+        continue;
+      auto UseIter = Use.getIterator();
+      if (std::distance(SrcMIIter, UseIter) >
+          std::distance(SrcMIIter, DestMIIter))
```
andcarminati: Nice! But I see `std::distance(SrcMIIter, DestMIIter)` as invariant, we could pre-compute.

llvm/test/CodeGen/AIE/aie2p/GlobalIsel/combine-postinc-load.mir
```diff
@@ -0,0 +1,128 @@
+# NOTE: Assertions have been autogenerated by utils/update_mir_test_checks.py UTC_ARGS: --version 4
+#
+# This file is licensed under the Apache License v2.0 with LLVM Exceptions.
+# See https://llvm.org/LICENSE.txt for license information.
+# SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+#
+# (c) Copyright 2023-2024 Advanced Micro Devices, Inc. or its affiliates
```
andcarminati: nit 2025

llvm/test/CodeGen/AIE/aie2p/GlobalIsel/combine-postinc-load.mir
```diff
@@ -0,0 +1,128 @@
...
+        %7:_(<32 x s32>) = G_INTRINSIC intrinsic(@llvm.aie2p.v32bf16.to.v32accfloat), %6:_(<32 x s16>)
+        %8:_(p0) = G_PTR_ADD %5:_, %2:_(s20)
+        %9:_(<32 x s16>) = G_LOAD %8:_(p0) :: (load (<32 x s16>), align 32, addrspace 6)
+        %10:_(<32 x s32>) = G_INTRINSIC intrinsic(@llvm.aie2p.v32bf16.to.v32accfloat), %9:_(<32 x s16>)
+        %11:_(<64 x s32>) = G_CONCAT_VECTORS %7:_(<32 x s32>), %10:_(<32 x s32>)
+        %12:_(p0), %13:_(s20), %14:_(s20) = G_INTRINSIC intrinsic(@llvm.aie2p.add.3d), %5:_(p0), %1:_(s20), %1:_(s20), %1:_(s20), %1:_(s20), %1:_(s20), %1:_(s20), %1:_(s20)
+        %15:_(<64 x s32>) = G_INTRINSIC_W_SIDE_EFFECTS intrinsic(@llvm.aie2p.ACC2048.accfloat.add.conf), %4:_(<64 x s32>), %11:_(<64 x s32>), %3:_(s32)
+        PseudoRET implicit $lr, implicit %12, implicit %11
+...
+
```
andcarminati, line 75: Maybe a short description will be nice.

llvm/lib/Target/AIE/AIECombinerHelper.cpp
```diff
@@ -277,6 +285,67 @@ bool llvm::canDelayMemOp(MachineInstr &MemI, MachineInstr &Dest,
...
+        break;
+      DependentInstrs.insert(&Use);
+      DependentInstrInRange(Use, DestMI, MRI, DependentInstrs);
```
andcarminati, line 316: Is there a chance for this to enter in loop? What happens if `Use` is a Phi node? We could have a test, something like:
```
a = Phi (c) ...
b = load p0
c = add a, b
p1 = padd p0
```
We could have a test with such a scenario.

llvm/lib/Target/AIE/AIECombinerHelper.cpp
```diff
@@ -552,9 +639,12 @@ void llvm::applyLdStInc(MachineInstr &MI, MachineRegisterInfo &MRI,
   // Debug Loc: Debug Loc of LOAD STORE: MI
   B.setDebugLoc(MI.getDebugLoc());
   auto NewInstr = B.buildInstr(MatchData.CombinedInstrOpcode);
-  for (auto *Instr : MatchData.ExtraInstrsToMove) {
+  for (auto *Instr : MatchData.ExtraInstrsToMoveBefore) {
     Instr->moveBefore(NewInstr);
   }
+  for (auto *Instr : MatchData.ExtraInstrsToMoveAfter) {
+    Instr->moveBefore(&*std::next(NewInstr.getInstr()->getIterator()));
```
andcarminati: Can we wrap this in a lambda? Just to be more clear to the eyes that move after is the same as move before the next.

llvm/lib/Target/AIE/AIECombinerHelper.cpp
krishnamtibrewala: Have added  a test and explain why we will not get into a loop.

PR #370, opened 2025-02-25
[AIE2P][AIE2] Combine single diff lane G_BUILD_VECTOR into G_AIE_BROADCAST and G_INSERT_VECTOR_ELT

abhinay-anubola:


llvm/lib/Target/AIE/AIECombinerHelper.cpp
```diff
@@ -1252,6 +1252,88 @@ bool llvm::applySplatVector(MachineInstr &MI, MachineRegisterInfo &MRI,
...
+  case 512:
+  case 1024:
+  case 2048:
+    break;
+  default:
+    // unimplemented
+    return false;
+  }
+  // Map to hold unique registers and their operand positions
+  DenseMap<Register, SmallSet<unsigned, 4>> UniqueRegs;
```
martien-de-jong: nit: 4 will be frequently too small. Actually you only need to maintain {register, count, (last) index}. As soon as you have two counts > 1 you can return false.

llvm/test/CodeGen/AIE/GlobalISel/prelegalizercombiner-splat-vector.mir
```diff
@@ -68,19 +68,19 @@ body:             |
...
     ; CHECK-LABEL: name: test_build_vector_256_32bit_scl_invalid
     ; CHECK: [[COPY:%[0-9]+]]:_(s32) = COPY $r0
     ; CHECK-NEXT: [[COPY1:%[0-9]+]]:_(s32) = COPY $r1
-    ; CHECK-NEXT: [[BUILD_VECTOR:%[0-9]+]]:_(<8 x s32>) = G_BUILD_VECTOR [[COPY]](s32), [[COPY]](s32), [[COPY]](s32), [[COPY1]](s32), [[COPY]](s32), [[COPY]](s32), [[COPY]](s32), [[COPY]](s32)
+    ; CHECK-NEXT: [[BUILD_VECTOR:%[0-9]+]]:_(<8 x s32>) = G_BUILD_VECTOR [[COPY1]](s32), [[COPY]](s32), [[COPY]](s32), [[COPY1]](s32), [[COPY]](s32), [[COPY]](s32), [[COPY]](s32), [[COPY]](s32)
     ; CHECK-NEXT: $wl0 = COPY [[BUILD_VECTOR]](<8 x s32>)
     %1:_(s32) = COPY $r0
     %2:_(s32) = COPY $r1
-    %3:_(<8 x s32>) = G_BUILD_VECTOR %1:_(s32), %1:_(s32), %1:_(s32), %2:_(s32), %1:_(s32), %1:_(s32), %1:_(s32), %1:_(s32)
+    %3:_(<8 x s32>) = G_BUILD_VECTOR %2:_(s32), %1:_(s32), %1:_(s32), %2:_(s32), %1:_(s32), %1:_(s32), %1:_(s32), %1:_(s32)
```
martien-de-jong, line 84: FYI, This can be done with two broadcasts and a VSEL. Kateryna's PR would pick up a G_SHUFFLE_VECTOR on two vector inputs. Don't know whether that could also take scalars as a vector of 1 element.

llvm/lib/Target/AIE/AIECombinerHelper.cpp
```diff
@@ -1252,6 +1252,88 @@ bool llvm::applySplatVector(MachineInstr &MI, MachineRegisterInfo &MRI,
   return true;
 }
 
+// Match something like:
+// %0:_(<32 x s16>) = G_BUILD_VECTOR %1:(s16), %2:(s16), %2:(s16) ... x32
```
martien-de-jong: I would make the example more generic, and put %1 in another position.

llvm/lib/Target/AIE/AIECombinerHelper.cpp
```diff
@@ -1252,6 +1252,88 @@ bool llvm::applySplatVector(MachineInstr &MI, MachineRegisterInfo &MRI,
...
+  Register SplatReg, DifferingReg;
+  unsigned DifferingIndex;
+
+  // Identify splat (multiple uses) and differing (single use) registers
+  for (const auto &[Reg, Indices] : UniqueRegs) {
+    if (Indices.size() == 1) {
+      DifferingReg = Reg;
+      DifferingIndex = *Indices.begin();
+    } else {
+      SplatReg = Reg;
```
martien-de-jong, line 1296: perhaps assert that the size is elementcount - 1?

llvm/lib/Target/AIE/AIECombinerHelper.cpp
```diff
@@ -1252,6 +1252,88 @@ bool llvm::applySplatVector(MachineInstr &MI, MachineRegisterInfo &MRI,
...
+  // vector pattern. More than 2 registers won't match; 1 unique register would
+  // be a splat vector combine
+  if (UniqueRegs.size() != 2)
+    return false;
+
+  Register SplatReg, DifferingReg;
+  unsigned DifferingIndex;
+
+  // Identify splat (multiple uses) and differing (single use) registers
+  for (const auto &[Reg, Indices] : UniqueRegs) {
```
andcarminati: Nice! I think we can generalize this in the future if we observe similar correlated patterns, like 2 or 3 singleton elements.

llvm/test/CodeGen/AIE/GlobalISel/prelegalizercombiner-splat-vector.mir
```diff
@@ -140,19 +140,19 @@ body:             |
 ...
 
 # Invalid Vector for broadcast.
-# As one of the element in G_BUILD_VECTOR is different from others, ideally this should not converted into G_AIE_BROADCAST_VECTOR
+# As two of the elements in G_BUILD_VECTOR is different from others, ideally this should not converted into G_AIE_BROADCAST_VECTOR
```
andcarminati: It looks like one extra instruction and we can solve this as well ;-). Let's keep on the radar.

andcarminati:
Nice work @abhinay-anubola! If you could provide some QoR numbers it could be nice!

llvm/lib/Target/AIE/AIECombinerHelper.cpp
```diff
@@ -1252,6 +1252,88 @@ bool llvm::applySplatVector(MachineInstr &MI, MachineRegisterInfo &MRI,
...
+    AIESingleDiffLaneBuildVectorMatchData &MatchInfo) {
+  B.setInstrAndDebugLoc(MI);
+  const Register DstVecReg = MatchInfo.DstVecReg;
+  const LLT DstVecRegTy = MRI.getType(DstVecReg);
+  const Register BcstDstReg = MRI.createGenericVirtualRegister(DstVecRegTy);
+  const LLT S32 = LLT::scalar(32);
+
+  buildBroadcastVector(B, MRI, MatchInfo.SplatReg, BcstDstReg);
+  const Register IdxReg =
+      B.buildConstant(S32, MatchInfo.DifferingIndex).getReg(0);
```
martien-de-jong, line 1323: I can imagine that we make this pattern also cover the regular broadcast, by making the insert conditional on the number of different elements, 1 or 2, that we have seen.

llvm/test/CodeGen/AIE/GlobalISel/prelegalizercombiner-splat-vector.mir
martien-de-jong: Yes, and a consecutive initial range could be done with a number of push.lo / push.hi

llvm/lib/Target/AIE/AIECombinerHelper.cpp
abhinay-anubola: We can’t add an assert here since we haven’t fully validated the register counts beforehand. Maybe we can return false by checking the above condition.

llvm/test/CodeGen/AIE/aie2p/GlobalIsel/prelegalizercombiner-vector-broadcast.mir
```diff
@@ -50,13 +50,13 @@ body:             |
...
     ; CHECK-NEXT: [[SHUF:%[0-9]+]]:_(<16 x s32>) = G_SHUFFLE_VECTOR [[BUILD_VECTOR]](<16 x s32>), [[BUILD_VECTOR1]], shufflemask(1, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0)
     ; CHECK-NEXT: $x0 = COPY [[SHUF]](<16 x s32>)
     %1:_(s32) = COPY $r0
     %2:_(s32) = COPY $r1
     %3:_(<16 x s32>) = G_BUILD_VECTOR %2:_(s32), %2:_(s32), %2:_(s32), %1:_(s32), %1:_(s32), %1:_(s32), %1:_(s32), %1:_(s32), %1:_(s32), %1:_(s32), %1:_(s32), %1:_(s32), %1:_(s32), %1:_(s32), %1:_(s32), %1:_(s32)
-    %4:_(<16 x s32>) = G_BUILD_VECTOR %1:_(s32), %2:_(s32), %1:_(s32), %1:_(s32), %1:_(s32), %1:_(s32), %1:_(s32), %1:_(s32), %1:_(s32), %1:_(s32), %1:_(s32), %1:_(s32), %1:_(s32), %1:_(s32), %1:_(s32), %1:_(s32)
+    %4:_(<16 x s32>) = G_BUILD_VECTOR %1:_(s32), %2:_(s32), %1:_(s32), %1:_(s32), %1:_(s32), %2:_(s32), %1:_(s32), %1:_(s32), %1:_(s32), %1:_(s32), %1:_(s32), %1:_(s32), %1:_(s32), %1:_(s32), %1:_(s32), %1:_(s32)
```
khallouh, line 59: You need to update the comments above these tests. It is not only invalid because it is not a splat vector anymore.

llvm/test/CodeGen/AIE/GlobalISel/prelegalizercombiner-splat-vector.mir
khallouh: G_BUILD_VECTOR is turned into a sequence of G_AIE_ADD_VECTOR_ELT_HI (VPUSH) by the legalizer so I don't think it would be picked up by the PR you're referring to

llvm/test/CodeGen/AIE/aie2p/GlobalIsel/prelegalizercombiner-vector-broadcast.mir
khallouh: But I'm not an expert in the nomenclature, maybe splat doesn't only refer to broadcast-only patterns but also broadcast+insert ones

llvm/lib/Target/AIE/AIECombinerHelper.cpp
```diff
@@ -1252,6 +1252,93 @@ bool llvm::applySplatVector(MachineInstr &MI, MachineRegisterInfo &MRI,
...
+    AIESingleDiffLaneBuildVectorMatchData &MatchInfo) {
+  assert(MI.getOpcode() == TargetOpcode::G_BUILD_VECTOR &&
+         "Expected a G_BUILD_VECTOR");
+
+  const Register DstVecReg = MI.getOperand(0).getReg();
+  const LLT DstVecTy = MRI.getType(DstVecReg);
+  const unsigned DstVecSize = DstVecTy.getSizeInBits();
+
+  switch (DstVecSize) {
+  case 256:
```
khallouh, line 1260: Could we extend to support 128-bit size as well? We keep accumulating gaps in our support for 128-bit types.

llvm/lib/Target/AIE/AIECombinerHelper.cpp
```diff
@@ -1252,6 +1252,93 @@ bool llvm::applySplatVector(MachineInstr &MI, MachineRegisterInfo &MRI,
...
+    if (RegInfo.first == 1) {
+      DifferingReg = Reg;
+      DifferingIndex = RegInfo.second;
+    } else {
+      SplatReg = Reg;
+    }
+  }
+  // Validate that one register was used exactly once
+  if (!DifferingReg.isValid() || !SplatReg.isValid())
+    return false;
```
khallouh, line 1301: I'm not sure I understand the reason for this check. Can we have a test for this case?

llvm/lib/Target/AIE/AIECombinerHelper.cpp
```diff
@@ -1252,6 +1252,93 @@ bool llvm::applySplatVector(MachineInstr &MI, MachineRegisterInfo &MRI,
...
+
+  // Check that the splat register is properly defined
+  const MachineInstr *SplatRegDef = getDefIgnoringCopies(SplatReg, MRI);
+  if (!SplatRegDef || SplatRegDef->getOpcode() == TargetOpcode::G_IMPLICIT_DEF)
```
khallouh, line 1305: Should this check be done in a combiner? Isn't that the job of the verifier to check that registers are defined before they are used? In this case an assert would be enough...

khallouh: What's the purpose of excluding the `G_IMPLICIT_DEF` case?

khallouh:
Maybe you can add the following as an end-2-end test:
```
define void @single_diff_lane_buildvector() {
entry:
  store <8 x i32> <i32 -2139062272, i32 -2139062144, i32 -2139062144, i32 -2139062144, i32 -2139062144, i32 -2139062144, i32 -2139062144, i32 -2139062144>, ptr addrspace(6) null, align 32
  ret void
}
```

llvm/lib/Target/AIE/AIECombinerHelper.cpp
abhinay-anubola: This check confirms that the for-loop above found exactly one register (“differing”) with a single usage and one register (“splat”) with multiple usages. That pattern ensures we have precisely one differing element in the vector, rather than a full splat or multiple unique lanes, and lets us apply the single diff lane combine. If we don’t see that pattern, we return false.

llvm/lib/Target/AIE/AIECombinerHelper.cpp
abhinay-anubola: We skip G_IMPLICIT_DEF to avoid overlapping with matchBroadcastElement, which specifically looks for a G_SHUFFLE_VECTOR(G_BUILD_VECTOR (VAL, UNDEF, ...), mask<0,0,...>) pattern to combine into a G_AIE_BROADCAST.

llvm/lib/Target/AIE/AIECombinerHelper.cpp
abhinay-anubola: Iam working on unified combiner for G_BUILD_VECTOR (splat, one diff splat and symmetric build vec). I would include this 128-bit support that PR.

llvm/lib/Target/AIE/AIECombinerHelper.cpp
martien-de-jong: It's a good reason, but it requires a comment stating it here. If we don't do this, we will have a monolithic unmaintainable instruction selector like SelectionDag in no time at all.

llvm/test/CodeGen/AIE/GlobalISel/prelegalizercombiner-splat-vector.mir
```diff
@@ -68,19 +68,19 @@ body:             |
...
```
martien-de-jong: nits:
- `should not be converted`
- `two of the elements ... are`
- wrap line to 80 columns

martien-de-jong: There are more copies of this comment in other files.

llvm/test/CodeGen/AIE/aie2p/buildvector.ll
```diff
@@ -0,0 +1,66 @@
...
+; CHECK-NEXT:    movxm r1, #-2139062144
+; CHECK-NEXT:    mova r29, #0
+; CHECK-NEXT:    ret lr
+; CHECK-NEXT:    vbcst.32 x0, r1 // Delay Slot 5
+; CHECK-NEXT:    vinsert.32 x0, x0, r29, r0 // Delay Slot 4
+; CHECK-NEXT:    mova p0, #0 // Delay Slot 3
+; CHECK-NEXT:    vst wl0, [p0, #0] // Delay Slot 2
+; CHECK-NEXT:    nop // Delay Slot 1
+entry:
+  store <8 x i32> <i32 -2139062272, i32 -2139062144, i32 -2139062144, i32 -2139062144, i32 -2139062144, i32 -2139062144, i32 -2139062144, i32 -2139062144>, ptr addrspace(6) null, align 32
```
martien-de-jong: nit: Use  something recognizable as 111, 777 and 888

llvm/lib/Target/AIE/AIECombinerHelper.cpp
```diff
@@ -1240,6 +1240,93 @@ bool llvm::applySplatVector(MachineInstr &MI, MachineRegisterInfo &MRI,
...
+  default:
+    // unimplemented
+    return false;
+  }
+  // DenseMap to hold unique registers and their (count, last index)
+  DenseMap<Register, std::pair<unsigned, unsigned>> UniqueRegs;
+  const unsigned NumOps = MI.getNumOperands();
+  for (unsigned i = 1; i < NumOps; i++) {
+    const Register OpReg = MI.getOperand(i).getReg();
+    auto &RegInfo = UniqueRegs[OpReg];
```
martien-de-jong, line 1274: I'm not sure whether the default constructor initializes to zero here, which is a problem for the count. Perhaps we can use a struct with default initialized members Count and LastIndex? That also reads better in the update below.

llvm/lib/Target/AIE/AIECombinerHelper.cpp
martien-de-jong: I don't know. We have looked at all operands and checked that there are exactly two registers used. I guess that their counts should add up to the number of elements, and since one has a count of one, the other should account for the rest. Even if we haven't checked further conditions, it checks the logic of the code above.

llvm/lib/Target/AIE/AIECombinerHelper.cpp
martien-de-jong: Ok, it seems to be defined by the standard. Feel free to take up the readability argument though.

martien-de-jong (approved):
Just a few nits, take them or leave them.

llvm/lib/Target/AIE/AIECombinerHelper.cpp
abhinay-anubola: I will fix it in upcoming NFC PR

PR #371, opened 2025-02-25
[AIEX] Multi Slot tweaks for QoR Improvements

F-Stuckmann:
with this PR there are no regressions in QoR left that are attributable to multi-slot pseudo instruction materialization, therefore it is enabled by default. 

Additional changes:
- don't attempt to assign slots, if they are all mapped to a single slot (default assignment is a better choice in this case)
- empirical tweaking of default reassign slot selection: if slots are exhausted (we already assigned memory banks to each slot but we encounter another memory bank), we now start selection from `slot a` instead of `slot b`.


llvm/lib/Target/AIE/AIEMultiSlotInstrMaterializer.cpp
```diff
@@ -160,6 +166,7 @@ SlotMapping getAssignedSlots(const MachineBasicBlock &MBB,
 /// in \p SlotToBanks are used and updated.
 bool assignSlots(SlotMapping &SlotToBanks, const MachineBasicBlock &MBB,
                  const AIEBaseInstrInfo *TII, const AIEHazardRecognizer &HR) {
```
martien-de-jong, line 165: nit: this is a nice side-effect-free predicate, but the name suggests it does something. a name starting with is, has, or should would be more appropriate.

llvm/lib/Target/AIE/AIEMultiSlotInstrMaterializer.cpp
```diff
@@ -100,7 +100,7 @@ class SlotMapping {
   /// If Slots have to be reassigned (because every Slot has already been
   /// assigned to a Memory Bank), use an Index to cycle through already
   /// used Slots.
-  unsigned ReassignIndex = 0;
+  unsigned ReassignIndex = 1;
```
martien-de-jong, line 104: This now neither is the first, nor may it be the last. I think this requires a rationale in the comment. I'm afraid that's going to be something like "this is a heuristic tuned for a particular benchmark set for a particular architecture". 
Perhaps you get around that by changing the slot order  from TII? Or is that already implementing another heuristic?

F-Stuckmann: We are specifically deviating from the slot order from TII. 
I'll add a comment

llvm/test/CodeGen/AIE/aie2p/schedule/postpipeliner/multiSlotAssignment/reassign_slots.mir
```diff
@@ -0,0 +1,98 @@
...
+# (c) Copyright 2025 Advanced Micro Devices, Inc. or its affiliates
+
+
+# RUN: llc  --mtriple=aie2p -O2 --start-before=postmisched \
+# RUN:   -aie-preassign-multi-slot-instr=true -verify-machineinstrs \
+# RUN:   --aie-loop-min-tripcount=10 %s -o - | FileCheck %s
+
+
+# Note that the LLVM IR doesn't match the actual MIR code. It is just a standard
+# loop providing some pointers into different spaces to dereference.
```
martien-de-jong: As discussed, the first sentence is correct the second one isn't.

martien-de-jong (dismissed):

martien-de-jong (approved):

PR #372, opened 2025-02-26
[AIE2P] Support more instructions for 2D/3D operand splitting

niwinanto:
This PR add support for more instructions for 2d/3d splitting.
Extend the AIESplitInstructionRewriter for supporting multiple tied register sets.

### DIFF
--- a/llvm/lib/Target/AIE/AIESplitInstructionRewriter.cpp
+++ b/llvm/lib/Target/AIE/AIESplitInstructionRewriter.cpp
@@ -68,8 +68,8 @@ bool AIESplitInstrBuilder::runOnMachineFunction(MachineFunction &MF) {
     for (MachineInstr &MI : make_early_inc_range(MBB)) {
       if (std::optional<unsigned> SplitOpcode =
               TII.getOpcodeWithAtomicOperands(MI.getOpcode())) {
-        assert(TII.getTiedRegInfo(MI).size() == 1);
-        rewriteInstruction(TII.getTiedRegInfo(MI).front(), MI, *SplitOpcode);
+        rewriteInstruction(TII.getTiedRegInfoForSplitting(MI.getOpcode()), MI,
+                           *SplitOpcode);
       }
     }
   }
@@ -178,8 +178,7 @@ bool AIESplitInstrReplacer::runOnMachineFunction(MachineFunction &MF) {
     for (MachineInstr &MI : make_early_inc_range(MBB)) {
       if (std::optional<unsigned> OpcodeWithTuple =
               TII.getOpcodeWithTupleOperands(MI.getOpcode())) {
-        assert(TII.getTiedRegInfo(*OpcodeWithTuple).size() == 1);
-        rewriteInstruction(TII.getTiedRegInfo(*OpcodeWithTuple).front(), MI,
+        rewriteInstruction(TII.getTiedRegInfoForSplitting(*OpcodeWithTuple), MI,
                            *OpcodeWithTuple);
       }
     }
--- a/llvm/lib/Target/AIE/aie2p/AIE2PInstrInfo.cpp
+++ b/llvm/lib/Target/AIE/aie2p/AIE2PInstrInfo.cpp
@@ -1635,6 +1635,21 @@ AIE2PInstrInfo::getTiedRegInfo(unsigned Opcode) const {
   }
 }
 
+TiedRegOperands
+AIE2PInstrInfo::getTiedRegInfoForSplitting(unsigned Opcode) const {
+  const auto &TiedRegInfoVector = getTiedRegInfo(Opcode);
+  unsigned Size = TiedRegInfoVector.size();
+  assert(Size >= 1 && "Expected to have tied register info");
+
+  for (auto &TiedRegInfo : TiedRegInfoVector) {
+    if (TiedRegInfo.SrcOps.size() > 1)
+      continue;
+    if (TiedRegInfo.canSplitSrcOps())
+      return TiedRegInfo;
+  }
+  return {};
+}
+
 unsigned AIE2PInstrInfo::getNumBypassedCycles(
     const InstrItineraryData *ItinData, const MachineInstr &DefMI,
     unsigned DefIdx, const MachineInstr &UseMI, unsigned UseIdx) const {
--- a/llvm/lib/Target/AIE/aie2p/AIE2PInstrInfo.td
+++ b/llvm/lib/Target/AIE/aie2p/AIE2PInstrInfo.td
@@ -133,7 +133,7 @@ let hasDelaySlot = true, isBranch = true, isTerminator = true,
 include "aie2p/AIE2PMultiSlotPseudoInstrInfo.td"
 // Define _split variants for instructions using 2D registers
 class Split2DInstr<Instruction RealInst, int opidx> : SplitPseudo<RealInst,
-    opidx, (ins eM:$mod, eDN:$dim_size, eDJ:$dim_stride, eDC:$dim_count)> {}
+    opidx, (ins eM:$mod_2d, eDN:$dim_size, eDJ:$dim_stride, eDC:$dim_count)> {}
 
 foreach instr = [PADDA_2D, PADDB_2D, PADDS_2D, PADD_2D_pseudo] in
     def instr # _split : Split2DInstr<instr, /*opidx=*/3>;
@@ -192,6 +192,18 @@ foreach instr = [
   VST_2D_SRS_4x_dmx_sts_srs_dm_srsSign0,
   VST_2D_SRS_4x_dmx_sts_srs_dm_srsSign1] in
     def instr # _split : Split2DInstr<instr, /*opidx=*/5>;
+
+foreach instr = [
+     VST_FLUSH_512_2D, VST_FLUSH_512_CONV_2D] in
+    def instr # _split : Split2DInstr<instr, /*opidx=*/7>;
+
+foreach instr = [
+    VLDA_POP_512_2D, VLDA_POP_544_2D, VLDA_POP_576_2D, VLDA_POP_640_2D,
+    VLDA_POP_704_2D, VLDB_POP_512_2D, VLDB_POP_544_2D, VLDB_POP_576_2D,
+    VLDB_POP_640_2D, VLDB_POP_704_2D,
+    VLD_POP_512_2D_pseudo, VLD_POP_544_2D_pseudo, VLD_POP_576_2D_pseudo, 
+    VLD_POP_640_2D_pseudo, VLD_POP_704_2D_pseudo] in
+    def instr # _split : Split2DInstr<instr, /*opidx=*/8>;
 
 // Define _split variants for instructions using 3D registers
 class Split3DInstr<Instruction RealInst, int opidx> : SplitPseudo<RealInst,
@@ -262,6 +274,17 @@ foreach instr = [ VLDA_3D_UPS_2x_dmw_lda_ups_w2b_upsSign0,
   VST_3D_SRS_4x_dmx_sts_srs_dm_srsSign1] in   
     def instr # _split : Split3DInstr<instr, /*opidx=*/6>;
 
+foreach instr = [
+    VST_FLUSH_512_3D, VST_FLUSH_512_CONV_3D] in
+    def instr # _split : Split3DInstr<instr, /*opidx=*/8>;
+
+foreach instr = [
+    VLDA_POP_512_3D, VLDA_POP_544_3D, VLDA_POP_576_3D, VLDA_POP_640_3D,
+    VLDA_POP_704_3D, VLDB_POP_512_3D, VLDB_POP_544_3D, VLDB_POP_576_3D,
+    VLDB_POP_640_3D, VLDB_POP_704_3D,
+    VLD_POP_512_3D_pseudo, VLD_POP_544_3D_pseudo, VLD_POP_576_3D_pseudo,
+    VLD_POP_640_3D_pseudo, VLD_POP_704_3D_pseudo] in
+    def instr # _split : Split3DInstr<instr, /*opidx=*/9>;
 
 let hasSideEffects = 0, Uses = [sp] in {
 let mayLoad = false, mayStore = true in {
--- a/llvm/lib/Target/AIE/aie2p/AIE2PInstrInfo.td
+++ b/llvm/lib/Target/AIE/aie2p/AIE2PInstrInfo.td
@@ -130,18 +130,138 @@ let hasDelaySlot = true, isBranch = true, isTerminator = true,
   }
 }
 
+include "aie2p/AIE2PMultiSlotPseudoInstrInfo.td"
 // Define _split variants for instructions using 2D registers
 class Split2DInstr<Instruction RealInst, int opidx> : SplitPseudo<RealInst,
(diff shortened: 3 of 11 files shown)


llvm/test/CodeGen/AIE/aie2p/ra/split-instrs-create.mir
```diff
@@ -11,7 +11,7 @@
 # that expose sub-registers.
 
 ---
-name:            test_split_2D
```
martien-de-jong, line 14: nit copyright 2025

llvm/test/CodeGen/AIE/aie2p/ra/split-instrs-create.mir
```diff
@@ -81,6 +81,25 @@ body:             |
...
+---
+name:            test_split_2D_opIdx6
+alignment:       16
+legalized:       true
+regBankSelected: true
+selected:        true
+tracksRegLiveness: true
+body:             |
+  bb.1.entry:
+    liveins: $p0, $lf0, $r24, $d1, $plfr0
```
martien-de-jong: No CHECK lines?

llvm/lib/Target/AIE/aie2p/AIE2PInstrInfo.td
```diff
@@ -133,7 +133,7 @@ let hasDelaySlot = true, isBranch = true, isTerminator = true,
 include "aie2p/AIE2PMultiSlotPseudoInstrInfo.td"
 // Define _split variants for instructions using 2D registers
 class Split2DInstr<Instruction RealInst, int opidx> : SplitPseudo<RealInst,
-    opidx, (ins eM:$mod, eDN:$dim_size, eDJ:$dim_stride, eDC:$dim_count)> {}
+    opidx, (ins eM:$mod_2d, eDN:$dim_size, eDJ:$dim_stride, eDC:$dim_count)> {}
```
martien-de-jong, line 136: This looks a bit out of place here. Perhaps split it in a separate commit?

llvm/test/CodeGen/AIE/aie2p/ra/split-instrs-replace.mir
```diff
@@ -167,6 +167,37 @@ body:             |
...
+    ; CHECK-LABEL: name: test_split_2D_opIdx5
+    ; CHECK: liveins: $d1, $p0, $s0
+    ; CHECK-NEXT: {{  $}}
+    ; CHECK-NEXT: $bmhh0, dead $p0, dead $dc1 = VLDA_2D_UPS_2x_dmw_lda_ups_w2b_upsSign0 $s0, killed $p0, $d1, implicit-def $srups_of, implicit $crsat, implicit $crupsmode, implicit $upssign0
+    $bmhh0, dead $p0, dead $dc1 = VLDA_2D_UPS_2x_dmw_lda_ups_w2b_upsSign0_split $s0, killed $p0, $m1, $dn1, $dj1, $dc1, implicit-def $srups_of, implicit $crsat, implicit $crupsmode, implicit $upssign0
+
+...
+
 ---
 name:            PADD_3D
```
khallouh: Nit: Rename this test with the opIdx like the others to show that this is a representative of its category?

llvm/test/CodeGen/AIE/aie2p/ra/split-instrs-create.mir
```diff
@@ -5,7 +5,7 @@
 # SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
 #
 # (c) Copyright 2024 Advanced Micro Devices, Inc. or its affiliates
```
khallouh: Nit: 2024-2025

llvm/lib/Target/AIE/aie2p/AIE2PInstrInfo.td
khallouh: Why is this needed?

niwinanto: Because the name has conflict with the operand name of instruction.

llvm/lib/Target/AIE/AIESplitInstructionRewriter.cpp
```diff
@@ -68,8 +68,11 @@ bool AIESplitInstrBuilder::runOnMachineFunction(MachineFunction &MF) {
     for (MachineInstr &MI : make_early_inc_range(MBB)) {
       if (std::optional<unsigned> SplitOpcode =
               TII.getOpcodeWithAtomicOperands(MI.getOpcode())) {
-        assert(TII.getTiedRegInfo(MI).size() == 1);
-        rewriteInstruction(TII.getTiedRegInfo(MI).front(), MI, *SplitOpcode);
+        auto TiedRegInfo = TII.getTiedRegInfo(MI);
+        assert(TiedRegInfo.size() >= 1);
```
martien-de-jong: nit: this assert was checking sanity of using front(), which is basically index 0. It would translate to checking whether the retrieved 2D/3DSplitIndex is in range of [0, size). I don't know what type TiedRegInfo has, but I can also imagine that you can leave that range check to using TiedRegInfo.at()

llvm/lib/Target/AIE/aie2p/AIE2PInstrInfo.cpp
```diff
@@ -1635,6 +1638,48 @@ AIE2PInstrInfo::getTiedRegInfo(unsigned Opcode) const {
   }
 }
 
+int AIE2PInstrInfo::getTiedRegInfo2D3DSplitIndex(unsigned Opcode) const {
+
+  auto TiedRegInfoVector = getTiedRegInfo(Opcode);
+  unsigned Size = TiedRegInfoVector.size();
+  assert(TiedRegInfoVector.size() >= 1);
```
martien-de-jong: use Size?

llvm/lib/Target/AIE/aie2p/AIE2PInstrInfo.cpp
```diff
@@ -1635,6 +1638,48 @@ AIE2PInstrInfo::getTiedRegInfo(unsigned Opcode) const {
...
+
+  // For only one set of tied registers, they must be 2D/3D split???
+  if (TiedRegInfoVector.size() == 1)
+    return 0;
+
+  for (unsigned Index = 0; Index < Size; Index++) {
+    auto TiedRegInfo = TiedRegInfoVector[Index];
```
martien-de-jong: I think you could use `for (auto &TiedRegInfo : TiedRegInfoVector)`

llvm/lib/Target/AIE/aie2p/AIE2PInstrInfo.cpp
```diff
@@ -1635,6 +1638,48 @@ AIE2PInstrInfo::getTiedRegInfo(unsigned Opcode) const {
...
+    // far.
+    if (SrcOpsOperandSubRegMappingVector.size() != 1)
+      continue;
+
+    auto SubRegSplit = SrcOpsOperandSubRegMappingVector.front().SubRegsSplit;
+    unsigned SubRegSplitSize = SubRegSplit.size();
+    // check if SubRegSplit == Split2DReg or Split3DReg
+    if (SubRegSplitSize == 4 || SubRegSplitSize == 8) {
+      bool IsEqual = true;
+      auto SplitReg2D3DRef = SubRegSplitSize == 4 ? Split2DReg : Split3DReg;
```
martien-de-jong: Could you change this to a const reference, and perhaps expand the auto? You're copying by value here.

llvm/lib/Target/AIE/AIESplitInstructionRewriter.cpp
```diff
@@ -68,8 +68,11 @@ bool AIESplitInstrBuilder::runOnMachineFunction(MachineFunction &MF) {
...
+        rewriteInstruction(
+            TiedRegInfo[TII.getTiedRegInfo2D3DSplitIndex(MI.getOpcode())], MI,
+            *SplitOpcode);
```
martien-de-jong: I have a concern here that this becomes very specific for 2D/3D handling. I would be happier if TiedRegInfo would say exactly what we needed to do. 
Also note that the function itself duplicates the bulk of the TiedRegInfo record in order to check that it has the right operand. I think that's rather back to front.

llvm/lib/Target/AIE/aie2p/AIE2PInstrInfo.cpp
```diff
@@ -1635,6 +1638,48 @@ AIE2PInstrInfo::getTiedRegInfo(unsigned Opcode) const {
...
+          IsEqual = false;
+          break;
+        }
+      }
+
+      if (IsEqual)
+        return Index;
+    }
+  }
+  return -1;
```
martien-de-jong: I think this should not occur, since the caller will crash on an out-of-bounds access.

llvm/lib/Target/AIE/aie2p/AIE2PInstrInfo.cpp
```diff
@@ -1264,20 +1264,23 @@ AIE2PInstrInfo::getNumReservedDelaySlots(const MachineInstr &MI) const {
   return 0;
 }
 
+namespace {
+const SmallVector<SubRegSplit, 8> Split2DReg = {
```
martien-de-jong: I think these could just be arrays.

llvm/lib/Target/AIE/AIESplitInstructionRewriter.cpp
niwinanto: Added one more assert which make sure index is in valid range.

llvm/lib/Target/AIE/aie2p/AIE2PInstrInfo.cpp
niwinanto: True, but if it is a `SmallVector` it helps reuse with `getTiedRegInfo` function.

llvm/lib/Target/AIE/aie2p/AIE2PInstrInfo.cpp
niwinanto: Ohh, yes. Done.

llvm/lib/Target/AIE/aie2p/AIE2PInstrInfo.cpp
niwinanto: Well, then it force me to keep a record of Index value separate. Especially, when I use `continue` I may to explicitly increment its value. But its a good idea to define `TiedRegInfo` as const reference, updated.

llvm/lib/Target/AIE/aie2p/AIE2PInstrInfo.cpp
niwinanto: Completely agree. Updated.

llvm/lib/Target/AIE/aie2p/AIE2PInstrInfo.cpp
niwinanto: Well, I dont get this. At least, for an example, `VLDA_FILL_512` do not have 2D or 3D splits and we bailout from the for loop early and return `-1` if we ever call this function.

llvm/lib/Target/AIE/aie2p/AIE2PInstrInfo.cpp
martien-de-jong: But you don't use Index otherwise, or did I miss a use?

llvm/lib/Target/AIE/aie2p/AIE2PInstrInfo.cpp
martien-de-jong: But we use the return value unchecked as index into an array. It will crash?

llvm/lib/Target/AIE/aie2p/AIE2PInstrInfo.cpp
niwinanto: I am returning the Index

llvm/lib/Target/AIE/aie2p/AIE2PInstrInfo.cpp
martien-de-jong: I don't see how?

llvm/lib/Target/AIE/aie2p/AIE2PInstrInfo.cpp
niwinanto: Yes, you are right. I have an assert where we use this now.

llvm/lib/Target/AIE/aie2p/AIE2PInstrInfo.cpp
krishnamtibrewala: It might be a good idea to move this into the AIE2PInstrInfo class instead of keeping them as global.

krishnamtibrewala:
```
struct OperandSubRegMapping {
  /// Index of the operand in a MachineInstr
  unsigned OpIdx;
  /// The sub-register of a tied register that this operand uses
  unsigned SubRegIdx;

  /// Information on how to split the operand into multiple operands.
  /// For super-regs, this is the list of sub-regs to split the operand into
  /// when rewriting the instruction into SplitOpcode.
  /// For for AIE2's 2D registers, this would be:
  ///   {MOD, DIM_SIZE, DIM_STRIDE, DIM_COUNT}
  SmallVector<SubRegSplit, 8> SubRegsSplit = {};
};
```

it will be helpful to directly use the` Split2DReg / Split3DReg` as part of `OperandSubRegMapping.SubRegsSplit`

llvm/lib/Target/AIE/aie2p/AIE2PInstrInfo.cpp
```diff
@@ -1635,6 +1638,50 @@ AIE2PInstrInfo::getTiedRegInfo(unsigned Opcode) const {
...
+    // For a 2D/3D split, TiedRegInfo has only one src register information so
+    // far.
+    if (SrcOpsOperandSubRegMappingVector.size() != 1)
+      continue;
+
+    auto SrcSubRegVector =
+        SrcOpsOperandSubRegMappingVector.front().SubRegsSplit;
+    unsigned SrcSubRegVectorSize = SrcSubRegVector.size();
+    // check if SubRegSplit == Split2DReg or Split3DReg
+    if (SrcSubRegVectorSize == 4 || SrcSubRegVectorSize == 8) {
```
krishnamtibrewala: if we move `Split2DReg` & `Split3DReg` in `AIE2PInstrInfo`
Enhance `OperandSubRegMapping` to take the `Split*DReg` as reference then we can significantly simplify `getTiedRegInfo2D3DSplitIndex` and reduce the hard coded values like 4/8.

martien-de-jong (approved):
As discussed. It is fine for now, but I would have liked it to have no assumptions on number of split operands, but just blindly apply the recipe that was specified in the TiedRegInfo. Effectiverly
```
for (auto &TRI  : TiedRegInfo) {
   apply(TRI, MI);
}
```
The TiedRegInfo would be checked to be sane by the tablegen backend, and the only reason why this pass would ever change is when we add inherently different ways of splitting, e.g. two outputs to two inputs and three outputs.

llvm/lib/Target/AIE/aie2p/AIE2PInstrInfo.cpp
niwinanto: Updated the PR and now I dont explicitly check for 2D/3D register.

PR #374, opened 2025-02-27
[AIE] Add llvm-as to wheel

newling:
I'd like to create both .s and .o files, but don't want to call llc twice (on a representative workload it takes 4 seconds, opt only takes 1 seconds, so invoking llc twice will ~double compile time). So I'd like to use llc to create .s, and then llvm-as to create .o from .s.

### DIFF
--- a/clang/cmake/caches/Peano-AIE.cmake
+++ b/clang/cmake/caches/Peano-AIE.cmake
@@ -35,6 +35,7 @@ set(LLVM_INSTALL_TOOLCHAIN_ONLY ON CACHE BOOL "")
 set(LLVM_TOOLCHAIN_TOOLS
   llc
   llvm-ar
+  llvm-as
   llvm-cxxfilt
   llvm-dwarfdump
   llvm-nm


konstantinschwarz:
Hi @newling,

llvm-as doesn't create object files, its meant to create LLVM bitcode from human readable LLVM IR: https://llvm.org/docs/CommandGuide/llvm-as.html

Do you need the `.s` file for debugging purposes?
4s for running llc sounds definitely too much. Are you using a debug build?

newling:
Hi @konstantinschwarz 


> Do you need the `.s` file for debugging purposes? 4s for running llc sounds definitely too much. Are you using a debug build? llvm-as doesn't create object files, its meant to create LLVM bitcode from human readable LLVM IR: https://llvm.org/docs/CommandGuide/llvm-as.html

Ok. I guess what I want is [llvm-mc](https://llvm.org/docs/CommandGuide/llvm-mc.html)  - would this PR make sense if I replace llvm-as with llvm-mc? 

> Do you need the .s file for debugging purposes?

Yes, I find it easier to read the assembly than what is generated by `llvm-objump -d foo.o`. And so it would be a nice artifact for us to have generated in our 'end-to-end' CI. 

> 4s for running llc sounds definitely too much. Are you using a debug build?

I don't think so, it's a wheel I downloaded: https://github.com/Xilinx/llvm-aie/releases/download/nightly/llvm_aie-19.0.0.2025013001+87f2b036-py3-none-manylinux_2_27_x86_64.manylinux_2_28_x86_64.whl

The wheel's llc is run with ` -O2 --march=aie2 --function-sections  --stack-size-section` and the input `.opt.ll is about 3'000 lines.

konstantinschwarz:
> Ok. I guess what I want is [llvm-mc](https://llvm.org/docs/CommandGuide/llvm-mc.html) - would this PR make sense if I replace llvm-as with llvm-mc?

Yes that would make sense.
Note however that's a not very well tested path - we usually directly emit object files.
Of course we should be able to assemble the code that we produce, but I can imagine there are some unhandled cases.
Bug reports (or patches) welcome! :) 

> I don't think so, it's a wheel I downloaded: https://github.com/Xilinx/llvm-aie/releases/download/nightly/llvm_aie-19.0.0.2025013001+87f2b036-py3-none-manylinux_2_27_x86_64.manylinux_2_28_x86_64.whl
> 
> The wheel's llc is run with ` -O2 --march=aie2 --function-sections --stack-size-section` and the input `.opt.ll is about 3'000 lines.

Would be interesting to run llc with `--time-trace` to see where the time is spent

martien-de-jong:
I suspect a complexity problem in code selection when basic blocks are large. Large basic blocks are usually an indication of overunrolling or scalarization of vectors. It would be interesting to see the ouput of llc --stop-after=instruction-select. 
(post)-pipelining loops with large basic blocks would be another sourfce of complexity.

newling:
Hi @konstantinschwarz, @martien-de-jong 

Update: I erroneously said that the opt.ll file was 3'000 lines, it was 25'000 lines. I ran with --time-trace, and see most of the time is in "PostRA Machine Instruction Scheduler" (let me know if you'd like the trace/opt.ll file). This large .ll file was generated running `opt` at O3, where we see a lot of function inlining. With O2, the opt.ll file _is_ around 3'000 lines, and llc takes < 0.2 seconds. FWIW the file size goes up 20x (140K -> 3M) and time goes up 25x (0.14 s -> 3.8 s), so my guess is that time is spent in a algorithm that is linear in block size. 

We have been using opt with -O3 because it was producing significantly faster code. But I figured out how to get O2 to match O3's performance yesterday ( see https://github.com/nod-ai/iree-amd-aie/pull/1149 ) and so plan to stop using O3 in iree-amd-aie. Unless you see any reason we should run opt with O3 instead of O2? And so I'll probably abandon this PR because it's simpler for us to call llc twice, once for .o and once for .s, and 0.1 seconds is negligible for us. 

One question I've had for a while is: does peano own/modify the passes that are run in its `opt` at O2 and O3? If so, would it makes sense to incorporate the small change in https://github.com/nod-ai/iree-amd-aie/pull/1149 into peano directly? i.e. rather than us running the O2 pipeline, followed by CSE, maybe peano should update its O2 pipeline to run CSE at the end?

konstantinschwarz:
Ok 25'000 lines in 4s sounds less concerning, and seeing a large chunk of that time being spent in the post-ra scheduler can be expected.

>One question I've had for a while is: does peano own/modify the passes that are run in its opt at O2 and O3? If so, would it makes sense to incorporate the small change in https://github.com/nod-ai/iree-amd-aie/pull/1149 into peano directly? i.e. rather than us running the O2 pipeline, followed by CSE, maybe peano should update its O2 pipeline to run CSE at the end?

So far we haven't made any changes to the default optimization pipelines. However, a while ago I've found opt is not running CSE after the final LICM pass, leading to annoying artifacts in the code generator for some of the kernels we are looking at. I added a CSE run in my local configuration, I think its time to make that the default for AIE's optimization pipeline :)

newling:
> So far we haven't made any changes to the default optimization pipelines. However, a while ago I've found opt is not running CSE after the final LICM pass, leading to annoying artifacts in the code generator for some of the kernels we are looking at. I added a CSE run in my local configuration, I think its time to make that the default for AIE's optimization pipeline :)

Very interesting! Please let us know when it's part of the default optimization pipeline, and I'll remove it on our end.

newling:
(Closing, as we've decided the wheel contains all the tools we need at the moment).

PR #375, opened 2025-02-27
[AIE] Constrain get.ctrl.reg destination to GPR regclass

konstantinschwarz:
Previously we allowed COPYing to the larger eRCR regclass. However this class cannot be spilled easily, because control registers cannot be stored to memory directly, but need to bounce through a GPR. There is very little benefit in constraining the class to eRCR, namely allowing direct COPY from one control register to another. That case can be handled as a peephole optimization of needed.

### DIFF
--- a/llvm/lib/Target/AIE/AIE2InstructionSelector.cpp
+++ b/llvm/lib/Target/AIE/AIE2InstructionSelector.cpp
@@ -2306,7 +2306,7 @@ bool AIE2InstructionSelector::selectGetControlRegister(
   else
     llvm_unreachable("Expected const value for control register map index.");
 
-  if (!RBI.constrainGenericRegister(DstReg, AIE2::eRCRRegClass, MRI))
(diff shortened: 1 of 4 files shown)


niwinanto (approved):
LGTM!

martien-de-jong (approved):
Looks sane to me, but would you have a motivating (regalloc) unittest?

konstantinschwarz:
> Looks sane to me, but would you have a motivating (regalloc) unittest?

I think a unit test for this problem isn't possible. We currently do not handle spilling the eRCR, leading to a crash.
However instead of handling that register class, I would like to prevent the situation in the first place.
InstructionSelection is the only place where we introduce that regclass, so constraining to a different one fixes the problem.
A test that showcases the problem would hence need to start before instruction selection and stop after register allocation.
Moreover it needs to have all GPRs live to not allow spilling to GPRs.

PR #376, opened 2025-02-28
[AIE2P] Regbankselect Fixup! 

abnikant:
1) Fix the compilation error when using -regbankselect-greedy, also fixup the opsize for vetor in store and G_CONCAT_VECTORS.
2) Support missing G_UNMERGE_VALUES in AlternativeMappings.


llvm/lib/Target/AIE/aie2p/AIE2PRegisterBankInfo.cpp
```diff
@@ -346,6 +346,8 @@ AIE2PRegisterBankInfo::getInstrAlternativeMappings(
     SmallVector<PartialMappingIdx, 4> OpRegBankIdx(NumOperands, PMI_MOD);
     SmallVector<const ValueMapping *, 8> OpdsMapping(NumOperands);
     SmallVector<unsigned, 4> OpSize(NumOperands, 20);
+    // Operand[VecRegOpIdx]- vector operand.
+    OpSize[VecRegOpIdx] = VecRegSize;
```
niwinanto: Are we setting the vector size for `LOAD*` variants? DO you know why the unit tests where not complaining about this?

niwinanto: May be you could  `VecRegOpIdx ==idx ? VecRegSize: 20` at the uses, no need to create a vector.

llvm/test/CodeGen/AIE/aie2p/GlobalIsel/regbankselect-phi.mir
```diff
@@ -373,12 +373,12 @@ body:             |
   ; CHECK-NEXT:   [[PHI1:%[0-9]+]]:ptrregbank(p0) = G_PHI [[COPY]](p0), %bb.0, %8(p0), %bb.1
   ; CHECK-NEXT:   [[PHI2:%[0-9]+]]:gprregbank(s32) = G_PHI [[C]](s32), %bb.0, %10(s32), %bb.1
   ; CHECK-NEXT:   [[PHI3:%[0-9]+]]:fiforegbank(<16 x s64>) = G_PHI [[DEF]](<16 x s64>), %bb.0, %12(<16 x s64>), %bb.1
-  ; CHECK-NEXT:   [[COPY1:%[0-9]+]]:accregbank(<16 x s64>) = COPY [[PHI3]](<16 x s64>)
+  ; CHECK-NEXT:   [[COPY1:%[0-9]+]]:vregbank(<16 x s64>) = COPY [[PHI3]](<16 x s64>)
```
niwinanto, line 376: Nice! we were selecting this wrong.

llvm/lib/Target/AIE/aie2p/AIE2PRegisterBankInfo.cpp
```diff
@@ -374,6 +374,51 @@ AIE2PRegisterBankInfo::getInstrAlternativeMappings(
...
+    const LLT HalfSrcTy = SrcTy.divide(2);
+    const unsigned HalfSrcTySize = HalfSrcTy.getSizeInBits();
+    // Only the accumulator is of 2048-bit size, with the possible mapping
+    // restricted to accumulator register banks. For a 512-bit vector,
+    // the only possible mapping is to a vector register bank.
+    if (SrcSize <= 512 || SrcSize == 2048)
+      break;
+    // If the instruction has any implicit-defs or uses,
+    // do not mess with it.
+    if (MI.getNumOperands() != 3)
```
niwinanto: Probably you can assert here.

llvm/lib/Target/AIE/aie2p/AIE2PRegisterBankInfo.cpp
```diff
@@ -374,6 +374,51 @@ AIE2PRegisterBankInfo::getInstrAlternativeMappings(
...
+  case TargetOpcode::G_UNMERGE_VALUES: {
+    const Register SrcReg = MI.getOperand(2).getReg();
+    const LLT SrcTy = MRI.getType(SrcReg);
+    unsigned SrcSize = getSizeInBits(SrcReg, MRI, TRI);
+    const LLT HalfSrcTy = SrcTy.divide(2);
+    const unsigned HalfSrcTySize = HalfSrcTy.getSizeInBits();
+    // Only the accumulator is of 2048-bit size, with the possible mapping
+    // restricted to accumulator register banks. For a 512-bit vector,
+    // the only possible mapping is to a vector register bank.
+    if (SrcSize <= 512 || SrcSize == 2048)
```
niwinanto, line 385: Move these lines down. If the size check fail, we can break without executing them.
```
    const LLT HalfSrcTy = SrcTy.divide(2);
    const unsigned HalfSrcTySize = HalfSrcTy.getSizeInBits();
```

llvm/lib/Target/AIE/aie2p/AIE2PRegisterBankInfo.cpp
```diff
@@ -374,6 +374,51 @@ AIE2PRegisterBankInfo::getInstrAlternativeMappings(
     }
     return AltMappings;
   }
+  case TargetOpcode::G_UNMERGE_VALUES: {
+    const Register SrcReg = MI.getOperand(2).getReg();
```
niwinanto: Super nit: You could use `MI.getNumOperands() - 1` instead hardcoding `3`, `G_UNMERGE_VALUES` can have more than 3 operands and 3 may not be a source operand.

khallouh:
Last I checked we had only very minor improvements in a couple of QoR tests. Did you check how it looks like now? 
Also, there was a significant degradation in `Cumsum_AIE2_bfloat16` with greedy (128%), did you have a chance to look at it?

llvm/lib/Target/AIE/aie2p/AIE2PRegisterBankInfo.cpp
abnikant: yes, since we allow BaseIdxOffset as valid for size <= [RBIdx size]

llvm/lib/Target/AIE/aie2p/AIE2PRegisterBankInfo.cpp
abnikant: done

llvm/lib/Target/AIE/aie2p/AIE2PRegisterBankInfo.cpp
abnikant: yes but not in RegisterBankSelect. We expect only 3 operands at this point. However I made this change, but still assert on 3 operands.

abnikant:
> Last I checked we had only very minor improvements in a couple of QoR tests. Did you check how it looks like now? Also, there was a significant degradation in `Cumsum_AIE2_bfloat16` with greedy (128%), did you have a chance to look at it?

This PR addresses compilation issues, but I will check the QoR.

niwinanto (approved):
LGTM!

llvm/test/CodeGen/AIE/aie2p/GlobalIsel/regbankselect-unmerge-values.mir
```diff
@@ -145,3 +145,17 @@ body:             |
...
 ...
+
+---
+name:            test_unmerge_v8s32
+legalized: true
+body:             |
+  bb.1.entry:
+    ; CHECK-LABEL: name: test_unmerge_v8s32
+    ; CHECK: [[COPY:%[0-9]+]]:vregbank(<8 x s32>) = COPY $wl0
+    ; CHECK-NEXT: [[UV:%[0-9]+]]:vregbank(<4 x s32>), [[UV1:%[0-9]+]]:vregbank(<4 x s32>) = G_UNMERGE_VALUES [[COPY]](<8 x s32>)
```
khallouh, line 156: This is not using alternative mapping for because of the source size equal to 512. Could you add a test where it is used for other sizes where we can observe the difference between greedy and fast?

khallouh: Maybe there is no case where Greedy would be better because the Copies have the same cost. If so, you can ignore the above comment.

khallouh (approved):

PR #377, opened 2025-03-03
[AIE2P] Fix VADD.f and VSUB.f configuration to silence the simulator

khallouh:
The multiplication mode (cmode) should be irrelevant for an addition/substraction but it is part of the amode, bmode, cmode combination to perform an element wise multiplication according to ISA, we set it to the value of that combination (1) to silence our tools (e.g. simulator complaining about an `Invalid multiplication mode`)

### DIFF
--- a/llvm/lib/Target/AIE/aie2p/AIE2PInstrPatterns.td
+++ b/llvm/lib/Target/AIE/aie2p/AIE2PInstrPatterns.td
@@ -63,7 +63,10 @@ class VecConf {
   dag ConfBits = (MOV_RLC_imm11_pseudo (i32 all));
 }
 
-def accfp32_vecconf : VecConf { let amode = AMODE_FP32; let bmode = BMODE_16x16; }
+// The multiplication mode (cmode) should be irrelevant for an addition/substraction but it is part of 
+// the amode, bmode, cmode combination to perform an element wise multiplication according to ISA, we set it 
+// to the value of that combination (1) to silence our tools (e.g. simulator)
+def accfp32_vecconf : VecConf { let amode = AMODE_FP32; let bmode = BMODE_16x16; let cmode = 1;}
 def mulbf16_vecconf : VecConf { let amode = AMODE_FP32; let bmode = BMODE_16x16; let cmode = CMODE_BF16xBF16_1_elem_1; }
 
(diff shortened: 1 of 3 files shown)


llvm/lib/Target/AIE/aie2p/AIE2PInstrPatterns.td
```diff
@@ -63,7 +63,10 @@ class VecConf {
   dag ConfBits = (MOV_RLC_imm11_pseudo (i32 all));
 }
 
-def accfp32_vecconf : VecConf { let amode = AMODE_FP32; let bmode = BMODE_16x16; }
+// The multiplication mode (cmode) should be irrelevant for an addition/substraction but it is part of 
+// the amode, bmode, cmode combination to perform an element wise multiplication (see ISA doc), we set it
```
abnikant: I think ISA doc is not public, I guess just writing according to ISA should be sufficient ?

khallouh: Changed accordingly

abnikant (approved):
LGTM!

llvm/lib/Target/AIE/aie2p/AIE2PInstrPatterns.td
```diff
@@ -63,7 +63,10 @@ class VecConf {
   dag ConfBits = (MOV_RLC_imm11_pseudo (i32 all));
 }
 
-def accfp32_vecconf : VecConf { let amode = AMODE_FP32; let bmode = BMODE_16x16; }
+// The multiplication mode (cmode) should be irrelevant for an addition/substraction but it is part of 
+// the amode, bmode, cmode combination to perform an element wise multiplication according to ISA, we set it 
+// to the value of that combination (1) to silence our tools (e.g. simulator)
+def accfp32_vecconf : VecConf { let amode = AMODE_FP32; let bmode = BMODE_16x16; let cmode = 1;}
```
niwinanto, line 69: Would you like to fix this for AIE2 as well?

khallouh: I checked and 0 is a valid cmode there

PR #378, opened 2025-03-04
[GlobalISel][AIE2P] Fix 3 different bugs in Prelegalizer generic combiner

khallouh:
This PR addresses all 55 tests where we observed the following compile failure using `-O2 -fno-unroll-loops `
`../llvm/include/llvm/CodeGenTypes/LowLevelType.h:185: llvm::ElementCount llvm::LLT::getElementCount() const: Assertion IsVector && "cannot get number of elements on scalar/aggregate"'` `

Some context: LLVM's middle-end vector combiner creates vectors from scalars which results in a unaligned load followed by a shuffle (splat pattern: **shuffle --> extract + broadcast**). 
This is detected and undone by the generic pre-legalizer combiner `matchExtractedVectorLoad` which tries to combine the vector load and the extract into a single scalar load. Before creating the load the combiner checks for legality but the query is constructed using an initializer list `{VecEltTy, PtrTy}` inside the `LegalityQuery` constructor which leads to weird behavior where the `VecEltTy` ends up pointing to a garbage LLT value when it is eventually used by the legality checks. 

- The 1s commit showcases the bug and the 2nd commit fixes it by explicitly creating `OpTys` and then passing it to the `LegalityQuery` constructor, this ensures that the types are correctly set up and the query is properly formed.
- Now since we are running the combiner pre-legalizer, we shouldn't get to the legality check in the first place. We do because our prelegalizer combiner pass is initialized using the default value of the `isPrelegalize` attribute which is false. The 3rd commit fixes this by setting `isPrelegalize` to true in the prelegalizer pass effectively enabling this combine and potentially others.
- The 4th commit solves another bug in `getVectorElementPointer` which assumes multiple of 8 pointer sizes. This is used by the combiner to compute the address of the scalar load, resulting in wrong code. The fix consists in replacing `DL.getIndexSize(AS) * 8` with `DL.getIndexSizeInBits(AS)`.

Better to review commit by commit.


llvm/test/CodeGen/AIE/aie2p/GlobalIsel/prelegalizer-combiner-extracted-vector-load.mir
```diff
@@ -0,0 +1,104 @@
...
+    %3:_(<8 x s16>) = G_IMPLICIT_DEF
+    %5:_(p0) = COPY $p0
+    %0:_(<8 x s16>) = G_LOAD %5:_(p0) :: (load (<8 x s16>) from `ptr null`, align 4)
+    %2:_(<32 x s16>) = G_SHUFFLE_VECTOR %0:_(<8 x s16>), %3:_, shufflemask(3, 3, 3, 3, 3, 3, 3, 3, 3, 3, 3, 3, 3, 3, 3, 3, 3, 3, 3, 3, 3, 3, 3, 3, 3, 3, 3, 3, 3, 3, 3, 3)
+    G_STORE %2:_(<32 x s16>), %5:_(p0) :: (store (<32 x s16>) into `ptr addrspace(6) null`, addrspace 6)
+    PseudoRET implicit $lr
+
+...
+
+
```
niwinanto: Super nit: remove the newlines.

llvm/lib/Target/AIE/aie2p/AIE2PPreLegalizerCombiner.cpp
```diff
@@ -85,7 +85,7 @@ AIE2PPreLegalizerCombinerImpl::AIE2PPreLegalizerCombinerImpl(
     const AIE2PSubtarget &STI, MachineDominatorTree *MDT,
     const LegalizerInfo *LI)
     : Combiner(MF, CInfo, TPC, &KB, CSEInfo),
-      Helper(Observer, B, /*IsPreLegalize*/ false, &KB, MDT, LI),
+      Helper(Observer, B, /*IsPreLegalize*/ true, &KB, MDT, LI),
```
niwinanto, line 88: You should fix this for aie2 as well.

niwinanto (commented):
The change looks good to me.

llvm/lib/CodeGen/GlobalISel/CombinerHelper.cpp
```diff
@@ -1326,15 +1326,10 @@ bool CombinerHelper::matchCombineExtractedVectorLoad(MachineInstr &MI,
 
   LegalityQuery::MemDesc MMDesc(*NewMMO);
 
-  // llvm::dbgs() << "Using VecEltTy : " << VecEltTy
-  //              << " in this debug print fixes the bug below\n";
-  LegalityQuery Q = {TargetOpcode::G_LOAD, {VecEltTy, PtrTy}, {MMDesc}};
```
abnikant, line 1329: lifetime issues with temporary objects -  {VecEltTy, PtrTy}, and {MMDesc}.  use-after-free due to LegalityQuery holding dangling pointer. This fix looks good to me but there are other places in CombinerHelper with the same issue, e.g. at line 690 :if (LI->getAction({CandidateLoadOpc, {UseTy, SrcTy}, {MMDesc}})
                .Action != LegalizeActions::Legal)
do you want to fix this ? Or can you try alternative fix, fixing the constructor to capture initializer_list arguments e.g. constexpr LegalityQuery(unsigned Opcode, std::initializer_list<LLT> Types,
                        std::initializer_list<MemDesc> MMODescrs = {})
    : LegalityQuery(Opcode, SmallVector<LLT>(Types.begin(), Types.end()), 
                    SmallVector<MemDesc>(MMODescrs.begin(), MMODescrs.end())) {} ? I think your solution is good and avoid changing the constructor, but check for (and fix) more places for similar issues in CombinerHelper.cpp .

llvm/lib/CodeGen/GlobalISel/LegalizerHelper.cpp
```diff
@@ -4090,7 +4090,7 @@ Register LegalizerHelper::getVectorElementPointer(Register VecPtr, LLT VecTy,
   // Convert index to the correct size for the address space.
   const DataLayout &DL = MIRBuilder.getDataLayout();
   unsigned AS = MRI.getType(VecPtr).getAddressSpace();
-  unsigned IndexSizeInBits = DL.getIndexSize(AS) * 8;
+  unsigned IndexSizeInBits = DL.getIndexSizeInBits(AS);
```
abnikant, line 4093: looks good.

llvm/lib/CodeGen/GlobalISel/CombinerHelper.cpp
khallouh: We have the same issue with the other constructor. I fixed it in the other combine where we also pass the initializer list.

abnikant (dismissed):
LGTM!

abnikant (approved):

PR #379, opened 2025-03-05
Fix Mangling for __bf16.

SagarMaheshwari99:


clang/lib/Basic/Targets/AIE.h
```diff
@@ -121,9 +121,11 @@ class LLVM_LIBRARY_VISIBILITY AIETargetInfo : public TargetInfo {
       return false;
     return true;
   }
-  /// Return the mangled code of bfloat. this is needed, else it asserts
+  /// Return the mangled code of bfloat. This is needed, else it asserts
```
konstantinschwarz: Nit: make this comment more useful?
e.g. along the lines of:
```
// We treat the builtin __bf16 type as a vendor extension type.
// Mangle the type name using its user visible name, "bfloat16".
```

konstantinschwarz (dismissed):
I'm biased, but the change looks good to me.
Just a nit regarding the code comment

konstantinschwarz (approved):

PR #380, opened 2025-03-10
[AIEX] Handle lock delays during SWP prologue merge with pre-header 

niwinanto:
The merging of the postpipeliner prologue with the preheader block's instruction doesn't respect lock delays. This allows lock instructions to be moved across loads.

### DIFF
--- a/llvm/lib/Target/AIE/AIEBaseInstrInfo.h
+++ b/llvm/lib/Target/AIE/AIEBaseInstrInfo.h
@@ -491,6 +491,10 @@ struct AIEBaseInstrInfo : public TargetInstrInfo {
   virtual int getMaxLastMemoryCycle() const;
   /// Return cycles for memory operations of an instruction.
   virtual SmallVector<int, 2> getMemoryCycles(unsigned SchedClass) const;
+  /// Return cycles for core to stall after lock instruction.
+  virtual int getCoreStallCycleAfterLock() const;
+  /// Return cycles for core to resume after lock instruction.
+  virtual int getCoreResumeCycleAfterLock() const;
 
   /// Return the schedclass for the given instruction descriptor based on
   /// operand regclass.
--- a/llvm/lib/Target/AIE/AIEBaseSubtarget.cpp
+++ b/llvm/lib/Target/AIE/AIEBaseSubtarget.cpp
@@ -194,11 +194,9 @@ bool updateSuccLatency(SDep &SuccEdge, SUnit &PredSU, int Latency) {
 // locks/DONE.
 class LockDelays : public ScheduleDAGMutation {
   void apply(ScheduleDAGInstrs *DAG) override {
-    // FIXME: Delays for locks to reach the core aren't completely described in
-    // the ISA. The numbers are therefore conservative.
-    const int CoreStallCycle = 2;
-    const int CoreResumeCycle = 8;
     const auto *TII = static_cast<const AIEBaseInstrInfo *>(DAG->TII);
+    const int CoreStallCycle = TII->getCoreStallCycleAfterLock();
+    const int CoreResumeCycle = TII->getCoreResumeCycleAfterLock();
 
     // Iterate over all the predecessors and successors of Lock instructions
     // to increase the edge latency.
--- a/llvm/lib/Target/AIE/AIEBaseSubtarget.cpp
+++ b/llvm/lib/Target/AIE/AIEBaseSubtarget.cpp
@@ -396,8 +396,21 @@ public:
           DAG->getSUnit(&*getBundleStart(FixedDepMI->getIterator()));
       assert(FixedDepSU && "Fixed Bundle has no corresponding SU.");
       SDep Dep(&FreeSU, SDep::Artificial);
-      Dep.setLatency(
-          AIE::maxLatency(&MI, *TII, *ItinData, /*IncludeStages=*/true));
+      auto Latency =
+          AIE::maxLatency(&MI, *TII, *ItinData, /*IncludeStages=*/true);
+      if (TII->isLock(MI.getOpcode())) {
+        Dep.setLatency(std::max(
+            TII->getCoreResumeCycleAfterLock() -
+                *TII->getFirstMemoryCycle(FixedDepMI->getDesc().SchedClass) + 1,
+            Latency));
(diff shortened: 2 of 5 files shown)


llvm/lib/Target/AIE/AIEBaseInstrInfo.h
```diff
@@ -491,6 +491,12 @@ struct AIEBaseInstrInfo : public TargetInstrInfo {
   virtual int getMaxLastMemoryCycle() const;
   /// Return cycles for memory operations of an instruction.
   virtual SmallVector<int, 2> getMemoryCycles(unsigned SchedClass) const;
+  /// Return cycles for core to stall after lock instruction.
+  /// FIXME: Delays for locks to reach the core aren't completely described in
+  /// the ISA. The numbers are therefore conservative.
```
SagarMaheshwari99: We don't need the FIXME here since there are no numbers.

llvm/test/CodeGen/AIE/aie2/schedule/postpipeliner/interleave-prologue.mir
```diff
@@ -437,3 +437,88 @@ body:             |
...
+  ; CHECK-NEXT:   $lc = ADD_NC $r0, -7
+  ; CHECK-NEXT:   $ls = MOVXM_lng_cg %bb.2
+  ; CHECK-NEXT:   $le = MOVXM_lng_cg <mcsymbol .L_LEnd0>
+  ; CHECK-NEXT:   $x6, $p0 = VLDB_UNPACK_S16_S8_ag_pstm_nrm killed $p0, $m0
+  ; CHECK-NEXT:   $x6, $p0 = VLDB_UNPACK_S16_S8_ag_pstm_nrm killed $p0, $m0
+  ; CHECK-NEXT:   $x6, $p0 = VLDB_UNPACK_S16_S8_ag_pstm_nrm killed $p0, $m0
+  ; CHECK-NEXT:   BUNDLE implicit-def $x6, implicit-def $wl6, implicit-def $wh6, implicit-def $p0, implicit killed $p0, implicit $m0, implicit killed $r18 {
+  ; CHECK-NEXT:     $x6, $p0 = VLDB_UNPACK_S16_S8_ag_pstm_nrm killed $p0, $m0
+  ; CHECK-NEXT:     ACQ_mLockId_imm 49, killed renamable $r18
+  ; CHECK-NEXT:   }
```
martien-de-jong: CHECK: This shows the problem since it has overtaken the the VLDB_UNPACKs, which are part of the prologue of the pipelined loop.

llvm/lib/Target/AIE/AIEMaxLatencyFinder.cpp
```diff
@@ -126,19 +126,23 @@ InstrAndCycle findEarliestRef(const MachineInstr &SrcMI,
...
       return {/*MI=*/nullptr, Cycle};
     }
     for (MachineInstr *DstMI : Bundle.getInstrs()) {
-      LLVM_DEBUG(dbgs() << "  " << *DstMI);
+      LLVM_DEBUG(dbgs() << " " << *DstMI);
       if (depends(SrcMI, *DstMI, TRI)) {
         LLVM_DEBUG(dbgs() << "    depends in cycle=" << Cycle << "\n");
         return {DstMI, Cycle};
       }
+      if (TII->isLock(SrcMI.getOpcode()) && (DstMI->mayLoadOrStore())) {
```
martien-de-jong: I wouldn't mind putting this condition in depends(), especially since the action is the same.

martien-de-jong: We may also need LdSt -> Lock dependences. AFAIK, this is generically used in interblock scheduling, and both orders may occur.

llvm/lib/Target/AIE/AIEBaseSubtarget.cpp
```diff
@@ -194,11 +194,9 @@ bool updateSuccLatency(SDep &SuccEdge, SUnit &PredSU, int Latency) {
 // locks/DONE.
 class LockDelays : public ScheduleDAGMutation {
   void apply(ScheduleDAGInstrs *DAG) override {
-    // FIXME: Delays for locks to reach the core aren't completely described in
-    // the ISA. The numbers are therefore conservative.
-    const int CoreStallCycle = 2;
-    const int CoreResumeCycle = 8;
     const auto *TII = static_cast<const AIEBaseInstrInfo *>(DAG->TII);
+    const int CoreStallCycle = TII->getCoreStallCycleAfterLock();
+    const int CoreResumeCycle = TII->getCoreResumeCycleAfterLock();
```
martien-de-jong, line 199: Could you factor this into an NFC commit?

llvm/lib/Target/AIE/AIEBaseSubtarget.cpp
```diff
@@ -398,8 +396,14 @@ class EmitFixedSUnits : public ScheduleDAGMutation {
           DAG->getSUnit(&*getBundleStart(FixedDepMI->getIterator()));
       assert(FixedDepSU && "Fixed Bundle has no corresponding SU.");
       SDep Dep(&FreeSU, SDep::Artificial);
-      Dep.setLatency(
-          AIE::maxLatency(&MI, *TII, *ItinData, /*IncludeStages=*/true));
+      if (TII->isLock(MI.getOpcode())) {
+        Dep.setLatency(
+            TII->getCoreResumeCycleAfterLock() -
+            *TII->getFirstMemoryCycle(FixedDepMI->getDesc().SchedClass) + 1);
```
martien-de-jong: I would feel safer if we take the max of this and the 'baseline' maxLatency.

llvm/test/CodeGen/AIE/aie2/schedule/postpipeliner/interleave-prologue.mir
niwinanto: Yes, you are right. This shows the problem.

llvm/lib/Target/AIE/AIEMaxLatencyFinder.cpp
niwinanto: Yes, it is a possible order, updated.

martien-de-jong (approved):

PR #385, opened 2025-03-12
[WIP] Unified G_BUILD_VECTOR combine

abhinay-anubola:


llvm/lib/Target/AIE/AIECombinerHelper.cpp
```diff
@@ -1230,101 +1192,121 @@ static void buildBroadcastVector(MachineIRBuilder &B, MachineRegisterInfo &MRI,
...
+    MatchInfo = [=, &MRI](MachineIRBuilder &B) {
+      buildBroadcastVector(B, MRI, SplatReg, DstVecReg);
+    };
+    return true;
   }
-  // Validate that one register was used exactly once
-  if (!DifferingReg.isValid() || !SplatReg.isValid())
+  case 2:
+    // Check for single differing lane splat, as there are 2 unique registers.
+    return matchSingleDiffSplatVector(UniqueRegs, DstVecReg, MRI, MatchInfo);
```
martien-de-jong, line 1280: I'd like to take this further. We still have separate apply lambdas. Instead, we could match a splat pattern, and pack enough information in the apply lambda so that it can build the broadcast of the splat, and then, if we have different elements, perform one (or more) additional inserts. I'd make the apply lambda accept any number of additional asserts, and let the match function decide how many to allow.
The next step is then to recognize that the generic build vector inserts all elements on top of an undefined, so that we can also reuse that apply lambda to build a generic vector. Note that a broadcast is no more expensive than an insert into an undef.. The inserts could also be done by push lo or push hi.

martien-de-jong: There's more fun to be had, by imagining the different registers to be inserted in vectors, and translating the BUILD initializer to a shufflemask. By virtualizing the shuffle match and apply methods, you could then build on all of the shuffle tricks, e.g. to create two (or more) broadcast vectors for two different registers, and to pick from the two using a a (or more) VSELs. (Not now, just to illustrate reuse possibilities)

khallouh:
Could you handle the case when some are undefs? See the following pattern from Reduce where %55 is undef so we should be able to broadcast %17 and insert %52 as a diff element:
` %53:_(<64 x s8>) = G_BUILD_VECTOR %52(s8), %17(s8), %17(s8), %17(s8), %17(s8), %17(s8), %17(s8), %17(s8), %17(s8), %17(s8), %17(s8), %17(s8), %17(s8), %17(s8), %17(s8), %17(s8), %17(s8), %17(s8), %17(s8), %17(s8), %17(s8), %17(s8), %17(s8), %17(s8), %17(s8), %17(s8), %17(s8), %17(s8), %17(s8), %17(s8), %17(s8), %17(s8), %55(s8), %55(s8), %55(s8), %55(s8), %55(s8), %55(s8), %55(s8), %55(s8), %55(s8), %55(s8), %55(s8), %55(s8), %55(s8), %55(s8), %55(s8), %55(s8), %55(s8), %55(s8), %55(s8), %55(s8), %55(s8), %55(s8), %55(s8), %55(s8), %55(s8), %55(s8), %55(s8), %55(s8), %55(s8)`

PR #386, opened 2025-03-12
[AIE2P] Support Vector Insert Element of 64 bits (Legalizer, RegBankSelect, ISel) 

khallouh:


llvm/lib/Target/AIE/aie2p/AIE2PRegisterBankInfo.cpp
```diff
@@ -1270,14 +1270,26 @@ void AIE2PRegisterBankInfo::setAIEGenericInstrMapping(
...
     LLT DstType = MRI.getType(DstReg);
     unsigned Size = DstType.getSizeInBits();
+    // 512-bit size is supported using VBCST and 2048-bit using VCLR
     assert((Size == 512 || Size == 2048) &&
            "Unsupported vector size for G_AIE_BROADCAST_VECTOR!");
-    if (Size == 2048)
+    if (Size == 2048) {
       OpRegBankIdx[0] = getAccPartialMappingIdx(DstType);
-    else
+    } else
```
martien-de-jong: I'm not too happy with bracketing only one branch of an if-else. (And I guess  the brackets were inserted when another statement was temporarily added)

llvm/lib/Target/AIE/aie2p/AIE2PRegisterBankInfo.cpp
```diff
@@ -1270,14 +1270,26 @@ void AIE2PRegisterBankInfo::setAIEGenericInstrMapping(
...
+
+  case AIE2P::G_AIE_INSERT_VECTOR_ELT: {
+    Register DstReg = MI.getOperand(0).getReg();
+    LLT DstType = MRI.getType(DstReg);
+    LLT SrcType = MRI.getType(MI.getOperand(1).getReg());
+    assert(DstType == SrcType);
+    unsigned Size = DstType.getSizeInBits();
+    OpRegBankIdx[0] = getVecPartialMappingIdx(DstType);
+    OpRegBankIdx[1] = getVecPartialMappingIdx(SrcType);
+    break;
```
martien-de-jong, line 1289: Both DstReg and Size go out of scope. I guess we can initialize DstType and SrcType in an analogous way and remove Size?

llvm/test/CodeGen/AIE/aie2p/GlobalIsel/inst-select-insert-vec-elt.mir
```diff
@@ -0,0 +1,59 @@
+# NOTE: Assertions have been autogenerated by utils/update_mir_test_checks.py
+#
+# This file is licensed under the Apache License v2.0 with LLVM Exceptions.
+# See https://llvm.org/LICENSE.txt for license information.
+# SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+#
+# (c) Copyright 2024 Advanced Micro Devices, Inc. or its affiliates
```
martien-de-jong: 2025?

llvm/test/CodeGen/AIE/aie2p/GlobalIsel/inst-select-insert-vec-elt.mir
```diff
@@ -0,0 +1,59 @@
...
+    %1:gprregbank(s64) = COPY $l0
+    %2:gprregbank(s32) = G_CONSTANT i32 8
+    %3:vregbank(<8 x s64>) = G_AIE_INSERT_VECTOR_ELT %0:vregbank, %1:gprregbank(s64), %2:gprregbank(s32)
+    $x3 = COPY %3:vregbank(<8 x s64>)
+    PseudoRET implicit $lr, implicit $x3
+...
+
+
+---
+name:            512bit_syn_idx_s64_elt
```
martien-de-jong: why 'syn'?

martien-de-jong (commented):
LGTM. Just a few textual things.

llvm/test/CodeGen/AIE/aie2p/GlobalIsel/inst-select-insert-vec-elt.mir
khallouh: dyn :)

martien-de-jong (dismissed):

llvm/lib/Target/AIE/aie2p/AIE2PRegisterBankInfo.cpp
```diff
@@ -1270,6 +1270,7 @@ void AIE2PRegisterBankInfo::setAIEGenericInstrMapping(
     Register DstReg = MI.getOperand(0).getReg();
     LLT DstType = MRI.getType(DstReg);
     unsigned Size = DstType.getSizeInBits();
+    // 512-bit size is supported using VBCST and 2048-bit using VCLR
```
abnikant, line 1273: 2048-bit only in case the operands are constant zero, it will select vclr else it will result in 4 - 512 VBCST. Probably update the comment to be clear.

llvm/lib/Target/AIE/aie2p/AIE2PRegisterBankInfo.cpp
```diff
@@ -1278,6 +1279,15 @@ void AIE2PRegisterBankInfo::setAIEGenericInstrMapping(
       OpRegBankIdx[0] = getVecPartialMappingIdx(DstType);
     break;
   }
+
+  case AIE2P::G_AIE_INSERT_VECTOR_ELT: {
+    LLT DstType = MRI.getType(MI.getOperand(0).getReg());
+    LLT SrcType = MRI.getType(MI.getOperand(1).getReg());
+    assert(DstType == SrcType);
```
abnikant, line 1286: Nit: do you want to put some message in assert ?

llvm/test/CodeGen/AIE/aie2p/GlobalIsel/regbankselect-aie-insert-vector-elt.mir
```diff
@@ -28,6 +28,28 @@ body:             |
...
+body:             |
+  bb.0.entry:
+    liveins: $l0, $r0
+    ; CHECK-LABEL: name: test_8x64
+    ; CHECK: liveins: $l0, $r0
+    ; CHECK-NEXT: {{  $}}
+    ; CHECK-NEXT: [[COPY:%[0-9]+]]:vregbank(<8 x s64>) = COPY $x0
+    ; CHECK-NEXT: [[COPY1:%[0-9]+]]:gprregbank(s64) = COPY $l0
+    ; CHECK-NEXT: [[C:%[0-9]+]]:gprregbank(s32) = G_CONSTANT i32 0
+    ; CHECK-NEXT: [[AIE_INSERT_VECTOR_ELT:%[0-9]+]]:vregbank(<8 x s64>) = G_AIE_INSERT_VECTOR_ELT [[COPY]], [[COPY1]](s64), [[C]](s32)
```
abnikant, line 43: nice! no more accumulator vector for 64-bit element vector. Can you also add an end-to-end test?

llvm/lib/Target/AIE/aie2p/AIE2PRegisterBankInfo.cpp
khallouh: I don't think that's true, we don't do any splitting in Instruction selection for broadcast. I just added this comment because it wasn't clear at first, but it's unrelated here.

llvm/test/CodeGen/AIE/aie2p/GlobalIsel/regbankselect-aie-insert-vector-elt.mir
khallouh: I'm not supporting any shuffle mask here. What are you expecting in an end-2-end test? I hit this missing support while experimenting with shuffle combines, a PR is coming to make use of it.

llvm/lib/Target/AIE/aie2p/AIE2PRegisterBankInfo.cpp
abnikant: yes, we don't do it in instruction selection but in through legalizer only 2048-bit vector = G_AIE_BRODCAST_VECTOR <constant 0> went through, for all the cases we split to make it 512-bits in legalizer combiner. Not sure what the comment in regbankselect means , just wanted it to be more clear.

llvm/test/CodeGen/AIE/aie2p/GlobalIsel/regbankselect-aie-insert-vector-elt.mir
abnikant: end-to-end test is not there when we have 64-bit element insert ?

llvm/lib/Target/AIE/aie2p/AIE2PRegisterBankInfo.cpp
khallouh: G_AIE_BROADCAST_VECTOR is always legal by default. I couldn't find anywhere where we do the splitting. I can specify that VCLR is only valid with 0 but I thought that was clear.

abnikant: okay, I meant we have matchSplatVector/buildBroadcastVector, where for 2048-bit we emit G_AIE_BROADCAST_VECTOR when the operand is constant else we build using 4 512-bits vbcst. Anyway you can update or leave this comment, not very important.

llvm/test/CodeGen/AIE/aie2p/GlobalIsel/regbankselect-aie-insert-vector-elt.mir
khallouh: Added the end 2 end test for insertelement

llvm/lib/Target/AIE/aie2p/AIE2PRegisterBankInfo.cpp
khallouh: Right, coming from G_BUILD_VECTOR or G_SHUFFLE_VECTORS. What I want to say by the comment is that once we get to regbankselect (past legalizer) only 512 and 2048 are supported because they correspond to instructions VBCST and VCLR respectively. Here is something I tried which looks strange, we select VCLR even if it's not a zero constant. I guess we don't expect this to happen and if it does it will hit this assert in RegBankSelect but still.
```
---
name:            test
legalized:       true
regBankSelected: true
selected:        false
body:             |
  bb.1.entry:
    liveins: $r0
    ; CHECK-LABEL: name: 512bit_splat_vzector_32bit_elt
    ; CHECK: liveins: $r0
    ; CHECK-NEXT: {{  $}}
    ; CHECK-NEXT: [[VCLR:%[0-9]+]]:acc2048 = VCLR
    ; CHECK-NEXT: $dm0 = COPY [[VCLR]]
    ; CHECK-NEXT: PseudoRET implicit $lr, implicit $dm0
    %0:gprregbank(s32) = COPY $r0
    %1:accregbank(<32 x s64>) = G_AIE_BROADCAST_VECTOR %0:gprregbank(s32)
    $dm0 = COPY %1:accregbank(<32 x s64>)
    PseudoRET implicit $lr, implicit $dm0
...
```

martien-de-jong (dismissed):

F-Stuckmann (approved):

abnikant (approved):

PR #387, opened 2025-03-13
[AIE2P][AIE2] Parse command line globalIsel combiner rules

niwinanto:
Parse globalIsel combiner rules, so that we can enable/disable then independent.


khallouh (commented):
Nice! Can we have a test where this is used?

khallouh (approved):
LGTM

PR #389, opened 2025-03-13
[AIE2P] Track exceptions to shuffle mask validity and add extract insert element combine

khallouh:

PR #393, opened 2025-03-18
[AIE2P] Combine shuffle to concat of a number of extractable sequences

khallouh:


llvm/lib/Target/AIE/AIECombinerHelper.cpp
```diff
@@ -1834,6 +1834,22 @@ void llvm::applyOffsetLoadStoreSharePtrAdd(MachineInstr &MI,
   MI.getOperand(1).setReg(PtrAddReg);
   MI.getOperand(2).setReg(NewOffsetReg);
 }
+
+static bool isPowerOfTwoOrZero(unsigned Height) {
+  return Height == 0 || (Height > 1 && !(Height & (Height - 1)));
+}
+
+static unsigned getNumMaskUndefs(const ArrayRef<int> &Mask,
+                                 unsigned startIndex) {
```
andcarminati: nit: `StartIndex`.

llvm/lib/Target/AIE/AIECombinerHelper.cpp
```diff
@@ -2612,5 +2628,194 @@ bool llvm::matchShuffleToExtractInsertElt(MachineInstr &MI,
...
+
+// Since we can only unmerge and concat vectors of the same size, equalize the
+// extract indices, by inserting additional indices if needed.
+static SmallVector<unsigned>
+EqualizeExtractIndices(const SmallVector<unsigned> &SubvecExtractIndices,
+                       unsigned MaskSize) {
+  if (SubvecExtractIndices.size() < 2)
+    return SubvecExtractIndices;
+  // Calculate the smallest difference between consecutive indices
+  unsigned minDiff = MaskSize - SubvecExtractIndices.back();
```
andcarminati: nit: `MinDiff`.

llvm/lib/Target/AIE/AIECombinerHelper.cpp
```diff
@@ -2612,5 +2628,194 @@ bool llvm::matchShuffleToExtractInsertElt(MachineInstr &MI,
...
+  for (unsigned I = 1; I < SubvecExtractIndices.size(); ++I) {
+    unsigned diff = SubvecExtractIndices[I] - SubvecExtractIndices[I - 1];
```
andcarminati: nit: `Diff`.

llvm/lib/Target/AIE/AIECombinerHelper.h
```diff
@@ -34,6 +34,13 @@ struct AIELoadStoreCombineMatchData {
   bool RemoveInstr;
 };
 
+struct ShuffleMaskValidity {
+  bool IsValid;
+  // Holds mask indices that don't satisfy the mask constraints
+  // Only relevant if mask is valid. Reallty??.
+  SmallVector<unsigned, 4> MaskExceptions;
+};
+
```
martien-de-jong: Looks like std::optional<SmallVector<>>

llvm/lib/Target/AIE/AIECombinerHelper.cpp
```diff
@@ -1834,6 +1834,22 @@ void llvm::applyOffsetLoadStoreSharePtrAdd(MachineInstr &MI,
   MI.getOperand(1).setReg(PtrAddReg);
   MI.getOperand(2).setReg(NewOffsetReg);
 }
+
+static bool isPowerOfTwoOrZero(unsigned Height) {
+  return Height == 0 || (Height > 1 && !(Height & (Height - 1)));
```
andcarminati: nit: 

```
 static bool isPowerOfTwoOrZero(unsigned Height) {
-  return Height == 0 || (Height > 1 && !(Height & (Height - 1)));
+  return Height == 0 || (Height > 1 && has_single_bit(Height));
 }
```

llvm/lib/Target/AIE/AIECombinerHelper.cpp
```diff
@@ -2507,10 +2512,105 @@ bool llvm::matchShuffleToCopy(MachineInstr &MI, MachineRegisterInfo &MRI,
...
+    return false;
+
+  SmallVector<unsigned, 4> Exceptions = IsSequential.MaskExceptions;
+  assert(!Exceptions.empty());
+
+  unsigned MaxNumInsertions;
+  if (MI.getMF()->getTarget().getTargetTriple().isAIE2P()) {
+    // The scalarization of G_SHUFFLE_VECTOR in the legalizer is more beneficial
+    // if there are more exceptions than NumSrcElems / 2 as AIE2P's VINSERT
+    // instrutions require a move to a register used for the index unlike VPUSH.
```
martien-de-jong: nit: instructions

llvm/lib/Target/AIE/AIECombinerHelper.cpp
```diff
@@ -1834,6 +1834,22 @@ void llvm::applyOffsetLoadStoreSharePtrAdd(MachineInstr &MI,
...
   MI.getOperand(2).setReg(NewOffsetReg);
 }
+
+static bool isPowerOfTwoOrZero(unsigned Height) {
+  return Height == 0 || (Height > 1 && !(Height & (Height - 1)));
+}
+
+static unsigned getNumMaskUndefs(const ArrayRef<int> &Mask,
+                                 unsigned startIndex) {
+  unsigned count = 0;
```
andcarminati: Count

llvm/lib/Target/AIE/AIECombinerHelper.cpp
```diff
@@ -2507,10 +2512,105 @@ bool llvm::matchShuffleToCopy(MachineInstr &MI, MachineRegisterInfo &MRI,
...
+  unsigned MaxNumInsertions;
+  if (MI.getMF()->getTarget().getTargetTriple().isAIE2P()) {
+    // The scalarization of G_SHUFFLE_VECTOR in the legalizer is more beneficial
+    // if there are more exceptions than NumSrcElems / 2 as AIE2P's VINSERT
+    // instrutions require a move to a register used for the index unlike VPUSH.
+    MaxNumInsertions = NumSrcElems / 2;
+  } else
+    llvm_unreachable(
+        "MaxNumInsertions Undefined for target. Does the target's Insert "
+        "instruction take immediate indices or only registers?");
```
martien-de-jong: Can we just take a default rather than fail?

llvm/lib/Target/AIE/AIECombinerHelper.cpp
```diff
@@ -2507,10 +2512,105 @@ bool llvm::matchShuffleToCopy(MachineInstr &MI, MachineRegisterInfo &MRI,
...
+
+  if (Exceptions.size() >= MaxNumInsertions)
```
martien-de-jong, line 2623: nit: `> MaxNumInsertions`, so that the maximum is allowed?

llvm/lib/Target/AIE/AIECombinerHelper.cpp
```diff
@@ -2612,5 +2628,194 @@ bool llvm::matchShuffleToExtractInsertElt(MachineInstr &MI,
...
+getSubMaskValuesAtIndices(const llvm::SmallVector<unsigned> &Indices,
+                          const ArrayRef<int> &Mask,
+                          const MaskMatch &MaskPattern) {
+  unsigned NumSubVectors = Indices.size();
+  SmallVector<unsigned, 4> SubMask;
+  SubMask.clear();
+  SubMask.reserve(Indices.size());
+  for (unsigned Idx : Indices) {
+    if (Mask[Idx] == -1)
+      SubMask.push_back(MaskPattern.getMaskValue(Idx % NumSubVectors));
```
andcarminati: Maybe assign with a ternary operator instead of if-else.

llvm/lib/Target/AIE/AIECombinerHelper.cpp
```diff
@@ -2507,10 +2512,105 @@ bool llvm::matchShuffleToCopy(MachineInstr &MI, MachineRegisterInfo &MRI,
...
+      Register VecToExtract =
+          Mask[ExceptionIdx] < (int)NumSrcElems ? Src1Reg : Src2Reg;
+
+      int ExtractIdx = Mask[ExceptionIdx] % NumSrcElems;
+      auto ExtrElt =
+          B.buildExtractVectorElementConstant(ElemTy, VecToExtract, ExtractIdx);
+
+      auto ExceptionIdxReg = B.buildConstant(LLT::scalar(32), ExceptionIdx);
+
+      if (Exceptions.size() == 1) {
```
martien-de-jong: I think that this can be coded simpler. front can be set up in the src register to use, which rotates to the dst reg of the previous iteration. This can be setup outside of the loop. the destination register is either the original dst for back() or a new register for others. This new register becomes the new src reg to be used.

llvm/lib/Target/AIE/AIECombinerHelper.cpp
```diff
@@ -2612,5 +2628,194 @@ bool llvm::matchShuffleToExtractInsertElt(MachineInstr &MI,
...
+// Return the values in the mask for the given indices. For undef, return the
+// expected value of the provided mask pattern.
+static SmallVector<unsigned, 4>
+getSubMaskValuesAtIndices(const llvm::SmallVector<unsigned> &Indices,
+                          const ArrayRef<int> &Mask,
+                          const MaskMatch &MaskPattern) {
+  unsigned NumSubVectors = Indices.size();
+  SmallVector<unsigned, 4> SubMask;
+  SubMask.clear();
+  SubMask.reserve(Indices.size());
```
martien-de-jong: I think the clear() is superfluous. I also think the reserve is strange, given that we may be pushing less than that size.

llvm/lib/Target/AIE/AIECombinerHelper.cpp
```diff
@@ -2612,5 +2628,194 @@ bool llvm::matchShuffleToExtractInsertElt(MachineInstr &MI,
...
+
+  SmallVector<unsigned, 4> Exceptions = IsSequential.MaskExceptions;
+
+  auto GetSubvecExtractIndices =
+      [=, &Mask]() -> std::optional<SmallVector<unsigned>> {
+    SmallVector<unsigned> SubvecExtractIndices;
+    SubvecExtractIndices.push_back(0);
+    SmallVector<int, 4> CurrentMask(Mask.begin(), Mask.end());
+    unsigned CurrentHeight = *Height;
+    SmallVector<unsigned, 4> CurrentExceptions = Exceptions;
```
martien-de-jong: This is a copy of Exceptions, which is a capture-copy of Exceptions, which is a copy of IsSequential.MaskExceptions.
The latter is also capture-copied separately  to copy it back at the end of the loop.

khallouh: I can capture IsSequential.MaskExceptions by reference but I need at least one copy even if I use IsSequential.MaskExceptions directly. As for the last "copy" at the end of the loop that's not the capture-copy of Exceptions but the updated exceptions which I recursively compute inside the loop. See below
```
      ShuffleMaskValidity IsSequential =
          SequentialMask.isValidMask(ArrayRef<int>(CurrentMask));
      if (IsSequential.IsValid)
        return SubvecExtractIndices;
      CurrentExceptions = IsSequential.MaskExceptions;
```

llvm/lib/Target/AIE/AIECombinerHelper.cpp
```diff
@@ -2612,5 +2628,194 @@ bool llvm::matchShuffleToExtractInsertElt(MachineInstr &MI,
...
+EqualizeExtractIndices(const SmallVector<unsigned> &SubvecExtractIndices,
+                       unsigned MaskSize) {
+  if (SubvecExtractIndices.size() < 2)
+    return SubvecExtractIndices;
+  // Calculate the smallest difference between consecutive indices
+  unsigned minDiff = MaskSize - SubvecExtractIndices.back();
+  for (unsigned I = 1; I < SubvecExtractIndices.size(); ++I) {
+    unsigned diff = SubvecExtractIndices[I] - SubvecExtractIndices[I - 1];
+    if (diff < minDiff) {
+      minDiff = diff;
```
martien-de-jong: use std::min?

llvm/lib/Target/AIE/AIECombinerHelper.cpp
```diff
@@ -2612,5 +2628,194 @@ bool llvm::matchShuffleToExtractInsertElt(MachineInstr &MI,
...
+}
+
+// Since we can only unmerge and concat vectors of the same size, equalize the
+// extract indices, by inserting additional indices if needed.
+static SmallVector<unsigned>
+EqualizeExtractIndices(const SmallVector<unsigned> &SubvecExtractIndices,
+                       unsigned MaskSize) {
+  if (SubvecExtractIndices.size() < 2)
+    return SubvecExtractIndices;
+  // Calculate the smallest difference between consecutive indices
```
martien-de-jong: I can see what you compute, just not why. Perhaps some explanation / example?

llvm/test/CodeGen/AIE/GlobalISel/prelegalizercombiner-shuffle-vector.mir
```diff
@@ -1107,9 +1107,10 @@ body:             |
     ; CHECK: liveins: $x0, $x1
     ; CHECK-NEXT: {{  $}}
     ; CHECK-NEXT: [[COPY:%[0-9]+]]:_(<16 x s32>) = COPY $x0
-    ; CHECK-NEXT: [[COPY1:%[0-9]+]]:_(<16 x s32>) = COPY $x1
-    ; CHECK-NEXT: [[SHUF:%[0-9]+]]:_(<16 x s32>) = G_SHUFFLE_VECTOR [[COPY]](<16 x s32>), [[COPY1]], shufflemask(8, 9, 10, 11, 12, 13, 14, 15, 0, 1, 2, 3, 4, 5, 6, 7)
-    ; CHECK-NEXT: PseudoRET implicit $lr, implicit [[SHUF]](<16 x s32>)
+    ; CHECK-NEXT: [[UV:%[0-9]+]]:_(<8 x s32>), [[UV1:%[0-9]+]]:_(<8 x s32>) = G_UNMERGE_VALUES [[COPY]](<16 x s32>)
+    ; CHECK-NEXT: [[AIE_UNPAD_VECTOR:%[0-9]+]]:_(<8 x s32>) = G_AIE_UNPAD_VECTOR [[COPY]](<16 x s32>)
```
martien-de-jong, line 1111: Can't we use UV instead?

llvm/lib/Target/AIE/AIECombinerHelper.cpp
```diff
@@ -2507,10 +2528,294 @@ bool llvm::matchShuffleToCopy(MachineInstr &MI, MachineRegisterInfo &MRI,
...
+  unsigned minDiff = MaskSize - SubvecExtractIndices.back();
+  for (unsigned I = 1; I < SubvecExtractIndices.size(); ++I) {
+    unsigned diff = SubvecExtractIndices[I] - SubvecExtractIndices[I - 1];
+    if (diff < minDiff) {
+      minDiff = diff;
+    }
+  }
+  // Generate the new indices
+  SmallVector<unsigned, 4> UsableSubvecExtractIndices;
```
abnikant: I get the build error with SmallVector<unsigned, 4> , AIECombinerHelper.cpp:2675:10: error: could not convert ‘UsableSubvecExtractIndices’ from ‘SmallVector<[...],4>’ to ‘SmallVector<[...],12>’
 2675 |   return UsableSubvecExtractIndices;
      |          ^~~~~~~~~~~~~~~~~~~~~~~~~~
      |          |
      |          SmallVector<[...],4>
 with SmallVector<unsigned, 12>, build passes.

khallouh: Interesting that it tries to convert to SmallVector<unsigned, 12> for you. This is just declaring a variable.

llvm/lib/Target/AIE/AIECombinerHelper.cpp
khallouh: See comment above 

> // Since we can only unmerge and concat vectors of the same size, equalize the
> // extract indices, by inserting additional indices if needed.

My logic to extract the indices only cares about sequences (until the first exception) Here is one example mask from the test `shuffle_concat_extracted_subvectors_vecToExtract_DiffSize1`  :  16, 17, 18, 19, 20, 21, 22, 23, 0, 1, 2, 3, undef, undef, undef, undef` I get extract indices at 0 for [16,23], at 8 for (0,3), and 12 for the rest, but I want to extract subvectors of the size so I have a helper that will equalize those indices so I convert this [0, 8, 12] to [0, 4, 8, 12]

llvm/lib/Target/AIE/AIECombinerHelper.cpp
khallouh: The idea is to be reminded by the failure to check the future target features and adjust or keep that number based on that, otherwise we might forget about it before it becomes critical for QoR.

llvm/lib/Target/AIE/AIECombinerHelper.cpp
khallouh: This was changed in #389 (see below). You mean just move the loop invariant `Exceptions.size() == 1` outside? 
```
InsertSrc = (Exceptions.size() == 1 || ExceptionIdx == Exceptions.front()) ? Src1Reg : InsertDst;
InsertDst = (Exceptions.size() == 1) || ExceptionIdx == Exceptions.back()) ? DstReg : MRI.createGenericVirtualRegister(Src1Ty);
```

llvm/lib/Target/AIE/AIECombinerHelper.cpp
khallouh: I think we will be pushing exactly the size of Indices. What makes you think we may be pushing less? It is about the size of Indices not Mask: ) the latter is just used to get the values.

khallouh: So the reserve makes sense, the clear however is indeed superfluous

llvm/lib/Target/AIE/AIECombinerHelper.cpp
abnikant: error is not in declaring, but when it returns UsableSubvecExtractIndices.

llvm/test/CodeGen/AIE/GlobalISel/prelegalizercombiner-shuffle-vector.mir
khallouh: We create one unmerge at a time in an extract-like fashion so `AIE_UNPAD_VECTOR` comes from the other unmerge which was combined to a `G_AIE_UNPAD_VECTOR `. I think the implementation is more readable this way and it allows us to reuse our buildUnmerge helper. What seems to be the same computation twice here is well taken care of by CSE so in the end we get:
```
    ; CHECK-NEXT: [[COPY:%[0-9]+]]:vec512 = COPY $x0
    ; CHECK-NEXT: [[COPY1:%[0-9]+]]:ewl = COPY [[COPY]].sub_256_hi
    ; CHECK-NEXT: [[COPY2:%[0-9]+]]:ewh = COPY [[COPY]].sub_256_lo
    ; CHECK-NEXT: [[REG_SEQUENCE:%[0-9]+]]:vec512 = REG_SEQUENCE [[COPY1]], %subreg.sub_256_lo, [[COPY2]], %subreg.sub_256_hi  
```

llvm/lib/Target/AIE/AIECombinerHelper.cpp
khallouh: but the return type is `SmallVector<unsigned>` not` SmallVector<unsigned, 12>`. Curious, what are you using to build llvm?

llvm/lib/Target/AIE/AIECombinerHelper.cpp
```diff
@@ -52,17 +52,22 @@ cl::opt<bool> CombineVecShiftByZero(
     "aie-combine-vec-shift-by-zero", cl::init(true), cl::Hidden,
     cl::desc("Combine vectors shift by zero into copies."));
 
-bool MaskMatch::isValidMask(ArrayRef<int> Mask) const {
+ShuffleMaskValidity MaskMatch::isValidMask(ArrayRef<int> Mask) const {
```
niwinanto: I would recommend creating a new method, may be `getInvalidMaskExceptions` than modifying `isValidMask`

llvm/lib/Target/AIE/AIECombinerHelper.cpp
martien-de-jong: Well, I think we have done a lot of work before finding out that we're not aie2p. We only fail if all the earlier conditions pass. I guess you could notice earlier that you're not AIE2P and should not even try this.

llvm/lib/Target/AIE/AIECombinerHelper.cpp
martien-de-jong: No, I mean that the invariant is that you're adding an operation to a source register. That src register starts at the true source register and gets updated to the new register. The only decision that we need to make in the loop is not to create a new intermediate register, but use the final Dst register. That test knows it is the last one, so doesn't need to update the src register for the next iteration. Moreover, the size() == 1 will be treated like any other case.

llvm/lib/Target/AIE/AIECombinerHelper.cpp
khallouh: That's from the other PR. It was approved so I merged it as. I will bring this as a pre-commit here

llvm/lib/Target/AIE/AIECombinerHelper.cpp
khallouh: I agree! The other PR was merged though. I will add a commit here to move it up and bail earlier.

llvm/lib/Target/AIE/AIECombinerHelper.cpp
khallouh: Right, I will make the change here.

llvm/lib/Target/AIE/AIECombinerHelper.cpp
khallouh: Moved this as lambda

llvm/lib/Target/AIE/AIECombinerHelper.cpp
khallouh: Moved this as lambda as well and included the explanation

martien-de-jong (approved):
Still heavy reading, but looks good. Thanks for the comprehensive set of tests

PR #395, opened 2025-03-19
Introduce `-mno-vitis-headers` to accelerate compiler checks for RT libs

philippjh:
Configuring the CMake project for RT libs takes more time than needed as `compiler-rt` standard compile checks are conducted with the inclusion of all Vitis headers. 

This PR disables the inclusion of those headers through the `CMAKE_REQUIRED_FLAGS` cache entry inspected by the [CheckCXXSourceCompiles](https://cmake.org/cmake/help/latest/module/CheckCXXSourceCompiles.html) module that is internally used to conduct these checks. 

From my observations, this change drops the time to conduct a debug build by 1-2 hours, depending on the exact CMake configuration.


martien-de-jong (approved):
Wonderful!

PR #396, opened 2025-03-20
Decouple register classes in rewriting

martien-de-jong:
Register reallocation was split into non-overlapping register classes. That means that if one class fails, the others can still reap benefit.
The loop-aware option is an attempt at a 'circular LRU', but there were no immediate benefits/

image

### DIFF
--- a/llvm/lib/Target/AIE/AIEWawRegRewriter.cpp
+++ b/llvm/lib/Target/AIE/AIEWawRegRewriter.cpp
@@ -271,13 +271,17 @@ bool AIEWawRegRewriter::renameMBBPhysRegs(const MachineBasicBlock *MBB) {
   }
 
   // Free physregs of all candidates and register their regclasses
-  std::set<const TargetRegisterClass *> RegClasses;
+  // RegClasses is a map that tracks successful reallocation for all
+  // registers in that class. If there's a failure in a register class, all
+  // allocations for register classes that overlap with the failed ones are
+  // reverted
+  std::map<const TargetRegisterClass *, bool> RegClasses;
   for (auto &[MO, Org] : Candidates) {
     auto VReg = MO->getReg();
     if (VRM->hasPhys(VReg))
       unassignReg(VReg);
     auto *RC = MRI->getRegClass(VReg);
-    RegClasses.insert(RC);
+    RegClasses.emplace(RC, true);
   }
   LLVM_DEBUG(dbgs() << "Renaming " << Candidates.size() << " candidates in "
                     << RegClasses.size() << " classes\n");
@@ -298,6 +302,7 @@ bool AIEWawRegRewriter::renameMBBPhysRegs(const MachineBasicBlock *MBB) {
   // Reallocate all virtual registers in Candidates.
   // Return true if successful.
   auto ReAllocate = [&](OriginalAllocation &Candidates, RoundRobin &Registers) {
+    bool AnyFails = false;
     BitVector UsedUnits;
     UsedUnits.resize(TRI->getNumRegUnits());
     for (auto &[MO, Org] : Candidates) {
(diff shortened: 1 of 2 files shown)


llvm/lib/Target/AIE/AIEWawRegRewriter.cpp
```diff
@@ -295,35 +303,80 @@ bool AIEWawRegRewriter::renameMBBPhysRegs(const MachineBasicBlock *MBB) {
...
+        RegClasses[RC] = false;
       }
     }
-    return true;
+    return !AnyFails;
   };
 
-  // Reapply the original allocation to all Candidates
+  auto Overlaps = [this](MCRegister Reg, const TargetRegisterClass *RC) {
+    for (auto RCReg : RC->getRegisters()) {
```
andcarminati: Here, we could use a more compact construct instead of a for, something like:
```
return llvm::any_of(RC->getRegisters(), [&](unsigned RCReg) {
            return TRI->regsOverlap(Reg, RCReg);});
```
Or do you think it is a lambda abuse? 😅

llvm/lib/Target/AIE/AIEWawRegRewriter.cpp
```diff
@@ -295,35 +303,80 @@ bool AIEWawRegRewriter::renameMBBPhysRegs(const MachineBasicBlock *MBB) {
...
-      auto VReg = MO->getReg();
-      if (VRM->hasPhys(VReg)) {
-        unassignReg(VReg);
+    for (auto [RC, Success] : RegClasses) {
+      LLVM_DEBUG(dbgs() << "RC=" << TRI->getRegClassName(RC)
+                        << (Success ? " Succeeded\n" : " Failed\n"));
+      if (Success) {
+        continue;
+      }
+      for (auto &[MO, Org] : Candidates) {
```
andcarminati: Check: now we revert just the failed cases, not everything.

llvm/lib/Target/AIE/AIEWawRegRewriter.cpp
```diff
@@ -295,35 +303,80 @@ bool AIEWawRegRewriter::renameMBBPhysRegs(const MachineBasicBlock *MBB) {
...
+      }
+    }
+    return false;
+  };
+
+  // Reapply the original allocation to all Candidates involved in failed
+  // register classes.
   auto RevertAllocation = [&](OriginalAllocation &Candidates) {
-    // The partial allocation may conflict with the original one in ugly ways.
+    // We revert all allocations whose original have an overlap
```
andcarminati: This whole function is getting bigger and bigger with a lambda-oriented paradigm. I think we should apply some refactor in the future!

llvm/lib/Target/AIE/AIEWawRegRewriter.cpp
```diff
@@ -352,6 +405,11 @@ bool AIEWawRegRewriter::renameMBBPhysRegs(const MachineBasicBlock *MBB) {
     }
     LLVM_DEBUG(dbgs() << "\n");
   }
+
+  // Prime the LRURegisters, so that the allocation is loop-aware.
+  if (PreAlloc) {
```
andcarminati, line 426: Nice, we start with a LRURegisters "calibration".

andcarminati:
Nice improvement @martien-de-jong. Do you have some QoR numbers to see the whole behavior?

llvm/lib/Target/AIE/AIEWawRegRewriter.cpp
martien-de-jong: TBH, I find it just more compact, not more readable. I might use it if clang-format would allow me to put that lambda on three separate lines.

llvm/lib/Target/AIE/AIEWawRegRewriter.cpp
martien-de-jong: Correct. Success is reset if a register class has failed.

llvm/lib/Target/AIE/AIEWawRegRewriter.cpp
martien-de-jong: But inlining that any_of into the caller allows me to remove the lambda, which might be worth it.

llvm/lib/Target/AIE/AIEWawRegRewriter.cpp
martien-de-jong: The idea is nice. It didn't bring the desired improvement though. I think the entire LRU idea becomes less effective  at higher stage counts; more and more live ranges start to interfere even if they are not close.

llvm/lib/Target/AIE/AIEWawRegRewriter.cpp
martien-de-jong: I think I agree.  They are capturing so much that it's probably better to make them into methods.

llvm/lib/Target/AIE/AIEWawRegRewriter.cpp
```diff
@@ -218,6 +246,75 @@ AIEWawRegRewriter::getVRegWithCopies(const MachineBasicBlock &MBB) const {
...
+  auto Overlaps = [this](MCRegister Reg, const TargetRegisterClass *RC) {
+    return any_of(RC->getRegisters(), [this, Reg](MCRegister RCReg) {
+      return TRI->regsOverlap(Reg, RCReg);
+    });
+  };
+
+  // We revert all allocations whose original have an overlap
+  // with a failed register class.
+  // The partial allocation may conflict with the original one in ugly ways,
+  // which means we may not be able to reassign in arbitrary order
```
andcarminati: nit: . in the end.

llvm/lib/Target/AIE/AIEWawRegRewriter.cpp
```diff
@@ -298,32 +302,61 @@ bool AIEWawRegRewriter::renameMBBPhysRegs(const MachineBasicBlock *MBB) {
   // Reallocate all virtual registers in Candidates.
   // Return true if successful.
   auto ReAllocate = [&](OriginalAllocation &Candidates, RoundRobin &Registers) {
+    bool AnyFails = false;
```
niwinanto: Can we make it true here and just `return AnyFails;`?

llvm/lib/Target/AIE/AIEWawRegRewriter.cpp
```diff
@@ -299,12 +303,28 @@ bool AIEWawRegRewriter::renameMBBPhysRegs(const MachineBasicBlock *MBB) {
...
 
+  // Pre-allocate all virtual registers in Candidates. The sole purpose of
+  // this is to prime the LRURegisters, so that the end of the loop is
+  // considered to be near to the start. No actual allocations are made.
+  // The ultimate allocation is expected to be different from this initial
+  // run -- that's the whole purpose.
+  auto PreAllocate = [&](OriginalAllocation &Candidates,
+                         RoundRobin &LRURegisters) {
+    // This is tracking any used register unit across the entire loop.
+    BitVector UsedUnits(TRI->getNumRegUnits());
```
niwinanto: `UsedUnits` is local to the lamda, does the comment above valid?

llvm/lib/Target/AIE/AIEWawRegRewriter.cpp
```diff
@@ -100,6 +109,25 @@ class AIEWawRegRewriter : public MachineFunctionPass {
 
   bool renameMBBPhysRegs(const MachineBasicBlock *MBB);
 
+  /// Pre-allocate all virtual registers in Candidates. The sole purpose of
```
niwinanto, line 112: Really nice refactoring. File is much more easy to read now.

llvm/lib/Target/AIE/AIEWawRegRewriter.cpp
martien-de-jong: Yes, it is shared across all candidates as handled in the loop below, which constitute 'the entire loop'. Note that the second run in reAllocate needs a fresh copy, in line with preAllocate only priming LRURegisters and nothing more.

llvm/lib/Target/AIE/AIEWawRegRewriter.cpp
martien-de-jong: Not without renaming it to AllSuccess. I can live with Success initialized to true falling to false.

andcarminati (approved):
LGTM.

PR #397, opened 2025-03-20
[AIE2P] add pseudo for scalar move (in M and S slot)

martien-de-jong:
The store unit can move address related registers (pointers and modifier regs), which enables two scalar moves in one cycle. 
This PR introduces a MultiSlotPseudo to represent that alternative, and it is used to implement scalar moves that result from materializing COPYs in regalloc.
I expect this to lead to higher code density in setup of multi-dim iterators and to unload the MOV slot for the occasional pointer copy that we see in kernel loops.
In particular, gemm has two pointer copies in the inner loop. 
As a follow up, we want to pre-materialize to MOVS in loops that underutilize the S slot.

### DIFF
--- a/llvm/lib/Target/AIE/aie2p/AIE2PMultiSlotPseudoInstrInfo.td
+++ b/llvm/lib/Target/AIE/aie2p/AIE2PMultiSlotPseudoInstrInfo.td
@@ -56,6 +56,12 @@ let isMoveImm = 1, isReMaterializable = 1, isAsCheapAsAMove = 1, Itinerary = II_
       "mov_scalar_imm11_pseudo ", "$dst, $i", [MOV_alu_mv_mv_mv_cg, MOVXM]>;
 }
 
+// Pseudo Scalar MOV. The MOVS register set is a strict subset
+let Itinerary = II_MOVS, hasSideEffects = false, mayLoad = false, mayStore = false in {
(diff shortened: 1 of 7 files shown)


khallouh (dismissed):
LGTM

llvm/test/CodeGen/AIE/aie2p/run-physreg-copy.mir
```diff
@@ -14,10 +14,10 @@ name:     test_copy_d
 body: |
```
khallouh, line 14: Nit: 2024-2025

llvm/lib/Target/AIE/aie2p/AIE2PMultiSlotPseudoInstrInfo.td
```diff
@@ -56,6 +56,12 @@ let isMoveImm = 1, isReMaterializable = 1, isAsCheapAsAMove = 1, Itinerary = II_
       "mov_scalar_imm11_pseudo ", "$dst, $i", [MOV_alu_mv_mv_mv_cg, MOVXM]>;
 }
 
+// Pseudo Scalar MOV. The MOVS register set is a strict subset
+let Itinerary = II_MOVS, hasSideEffects = false, mayLoad = false, mayStore = false in {
+  def MOV_Scalar_pseudo : MultiSlot_Pseudo< (outs OP_mAguDst:$dst), (ins OP_mAguSrc:$src),
```
andcarminati: nit: should we use: `def MOV_scalar_pseudo` ?

andcarminati (approved):
LGTM.

PR #400, opened 2025-03-24
[AIE2P][AIE2] Split symmetric G_BUILD_VECTOR into two G_BUILD_VECTORs

abhinay-anubola:
Splits the symmetric G_BUILD_VECTOR into two half-sized G_BUILD_VECTOR operations and then emits a G_CONCAT_VECTORS to combine them back into the final vector.


llvm/lib/Target/AIE/AIECombine.td
```diff
@@ -9,6 +9,29 @@
 //===----------------------------------------------------------------------===//
 include "llvm/Target/GlobalISel/Combine.td"
 
+def aie_all_combines : GICombineGroup<[trivial_combines, vector_ops_combines,
```
andcarminati, line 14: Maybe we should state in a comment why we are doing this, like which combine we are excluding and why.

khallouh: Could you also post some QoR numbers and if the excluded combines affect any kernels?

llvm/lib/Target/AIE/AIECombinerHelper.cpp
```diff
@@ -1369,6 +1370,70 @@ bool llvm::applySingleDiffLaneBuildVector(
...
+                                     MachineRegisterInfo &MRI) {
+
+  assert(MI.getOpcode() == TargetOpcode::G_BUILD_VECTOR &&
+         "Expected a G_BUILD_VECTOR");
+  const Register DstVecReg = MI.getOperand(0).getReg();
+  const LLT DstVecTy = MRI.getType(DstVecReg);
+  const unsigned DstVecSize = DstVecTy.getSizeInBits();
+
+  switch (DstVecSize) {
+  case 256:
```
khallouh, line 1493: How about 128?

llvm/test/CodeGen/AIE/aie2p/buildvector.ll
```diff
@@ -40,6 +40,25 @@ entry:
...
+; CHECK-NEXT:    mova p0, #0
+; CHECK-NEXT:    vbcst.32 x0, r0
+; CHECK-NEXT:    ret lr
+; CHECK-NEXT:    vbcst.32 x2, r1 // Delay Slot 5
+; CHECK-NEXT:    vst wl0, [p0, #0] // Delay Slot 4
+; CHECK-NEXT:    mova p0, #32 // Delay Slot 3
+; CHECK-NEXT:    vst wl2, [p0, #0] // Delay Slot 2
+; CHECK-NEXT:    nop // Delay Slot 1
+entry:
+  store <16 x i32> <i32 111, i32 111, i32 111, i32 111, i32 111, i32 111, i32 111, i32 111, i32 222, i32 222, i32 222, i32 222, i32 222, i32 222, i32 222, i32 222>, ptr addrspace(6) null, align 32
```
khallouh, line 58: Could we have a similar test for each size?

llvm/lib/Target/AIE/AIECombinerHelper.cpp
abhinay-anubola: There is not much use of that here, as it splits into two 64-bit build vectors.(no splat exist for 64-bit)

llvm/lib/Target/AIE/AIECombinerHelper.cpp
```diff
@@ -1369,6 +1370,70 @@ bool llvm::applySingleDiffLaneBuildVector(
...
+  }
+
+  const unsigned NumElts = MI.getNumOperands() - 1;
+  const unsigned HalfNumElts = NumElts / 2;
+  const MachineOperand FirstOp = MI.getOperand(1);
+  const MachineOperand SecondOp = MI.getOperand(HalfNumElts + 1);
+
+  // If both halves are the same register, it's effectively a splat, and the
+  // splat vector combine already handles that case.
+  if (FirstOp.isIdenticalTo(SecondOp))
```
khallouh: We don't know if it's a splat, but yes it's not symmetric and we can return early.

llvm/lib/Target/AIE/AIECombinerHelper.cpp
```diff
@@ -1369,6 +1370,70 @@ bool llvm::applySingleDiffLaneBuildVector(
...
+    return false;
+  // Ensures that each operand in the first half matches FirstOp, and each
+  // operand in the second half matches SecondOp.
+  for (unsigned i = 2; i <= HalfNumElts; i++) {
+    if (!MI.getOperand(i).isIdenticalTo(FirstOp)) {
+      return false;
```
khallouh: Could we match cases when it is not identical to FirstOp but is undef?

khallouh: Or if FirstOp itself is undef but all others in each half are identical...

llvm/lib/Target/AIE/AIECombinerHelper.cpp
```diff
@@ -1369,6 +1370,70 @@ bool llvm::applySingleDiffLaneBuildVector(
...
+                                     MachineIRBuilder &B,
+                                     GISelChangeObserver &Observer) {
+  Register DstReg = MI.getOperand(0).getReg();
+  LLT DstTy = MRI.getType(DstReg);
+
+  B.setInstrAndDebugLoc(MI);
+  LegalizerHelper Helper(B.getMF(), Observer, B);
+  // Splits the G_BUILD_VECTOR into two half-sized G_BUILD_VECTOR operations and
+  // then emits a G_CONCAT_VECTORS to combine them back into the final vector.
+  Helper.fewerElementsVector(
```
khallouh: Nice. Checking: can we do this in a matchInfo lambda without defining this new apply function?

llvm/lib/Target/AIE/AIECombine.td
niwinanto: To be frank, I dont get why we need this change. Also, can we move this to a separate commit?

llvm/lib/Target/AIE/AIECombinerHelper.cpp
khallouh: I think we do have a use case. Splat for 64-bit (e.g. v232) is selected into `REG_SEQUENCE`. 
So `2 x REG_SEQUENCE + CONCAT` is definitely better than `4 x VPUSH + VSHIFT` ;)

llvm/lib/Target/AIE/AIECombinerHelper.cpp
```diff
@@ -1369,6 +1370,70 @@ bool llvm::applySingleDiffLaneBuildVector(
   return true;
 }
 
+// Match something like:
+// %0:_(<32 x s16>) = G_BUILD_VECTOR %1:_(s16), ... x16, %2:_(s16), ... x16
+//
+// To turn it into
+// %3:_(<32 x s16>) = G_BUILD_VECTOR %1:_(s16), ... x16
+// %4:_(<32 x s16>) = G_BUILD_VECTOR %2:_(s16), ... x16
+// %0:_(<32 x s16>) = G_CONCAT_VECTORS %3:_(<32 x s16>), %4:_(<32 x s16>)
```
martien-de-jong: It's not very clear what we're matching and why. I'm guessing that we have only two elements and we want to broadcast the halves? Are all vector lengths as intended in this example? (I would expect 16 x s16 for the sub-BUILD_VECTORs)

llvm/lib/Target/AIE/AIECombinerHelper.cpp
```diff
@@ -1369,6 +1370,70 @@ bool llvm::applySingleDiffLaneBuildVector(
...
+    break;
+  default:
+    // unimplemented
+    return false;
+  }
+
+  const unsigned NumElts = MI.getNumOperands() - 1;
+  const unsigned HalfNumElts = NumElts / 2;
+  const MachineOperand FirstOp = MI.getOperand(1);
+  const MachineOperand SecondOp = MI.getOperand(HalfNumElts + 1);
```
martien-de-jong: I think it would be nice to have a lambda getSrcOperand(unsigned I) that captures MI, gets rid of the +1 everywhere and renormalizes the for loop below to the standard `(unsigned I = 0; I < Num; I++)` form.

llvm/lib/Target/AIE/AIECombinerHelper.cpp
martien-de-jong: Yep that would call for finding the 'representative' FirstOp or SecondOp on the fly. Since we will do that twice, it's probably a good idea to factor that into a function that checks each of the subranges.

llvm/test/CodeGen/AIE/GlobalISel/prelegalizercombiner-symmetric-build-vector.mir
```diff
@@ -0,0 +1,293 @@
...
+name:            test_build_vector_512_8bit_scl
+body:             |
+  bb.1.entry:
+    ; CHECK-LABEL: name: test_build_vector_512_8bit_scl
+    ; CHECK: [[COPY:%[0-9]+]]:_(s32) = COPY $r0
+    ; CHECK-NEXT: [[COPY1:%[0-9]+]]:_(s32) = COPY $r1
+    ; CHECK-NEXT: [[AIE_BROADCAST_VECTOR:%[0-9]+]]:_(<64 x s8>) = G_AIE_BROADCAST_VECTOR [[COPY]](s32)
+    ; CHECK-NEXT: [[AIE_UNPAD_VECTOR:%[0-9]+]]:_(<32 x s8>) = G_AIE_UNPAD_VECTOR [[AIE_BROADCAST_VECTOR]](<64 x s8>)
+    ; CHECK-NEXT: [[AIE_BROADCAST_VECTOR1:%[0-9]+]]:_(<64 x s8>) = G_AIE_BROADCAST_VECTOR [[COPY1]](s32)
+    ; CHECK-NEXT: [[AIE_UNPAD_VECTOR1:%[0-9]+]]:_(<32 x s8>) = G_AIE_UNPAD_VECTOR [[AIE_BROADCAST_VECTOR1]](<64 x s8>)
```
martien-de-jong, line 121: CHECK: There are patterns that absorb the UNPAD_VECTORs (Ultimately we will have 32 x s8 subregs being merged to 64 x s8?)

llvm/lib/Target/AIE/AIECombinerHelper.cpp
khallouh: Nevermind, we will have the same result as we would need to support a concat of 64-bit in a cheaper way for this to make sense.

llvm/lib/Target/AIE/AIECombine.td
martien-de-jong: We explicit list the combines that we use, rather than getting an opaque package deal from upstream llvm.  That's useful in itself. Any annotation as to why we need a particular combine would add to that value. I hope we can intersperse comment within this tablegen list?
I agree that a separate commit would be good; perhaps even two, one copying the standard list, then removing the ones we don't want. The commit comment can explain why we don't want them.

llvm/lib/Target/AIE/AIECombinerHelper.cpp
```diff
@@ -1369,6 +1370,70 @@ bool llvm::applySingleDiffLaneBuildVector(
...
+bool llvm::matchSymmetricBuildVector(MachineInstr &MI,
+                                     MachineRegisterInfo &MRI) {
+
+  assert(MI.getOpcode() == TargetOpcode::G_BUILD_VECTOR &&
+         "Expected a G_BUILD_VECTOR");
+  const Register DstVecReg = MI.getOperand(0).getReg();
+  const LLT DstVecTy = MRI.getType(DstVecReg);
+  const unsigned DstVecSize = DstVecTy.getSizeInBits();
+
+  switch (DstVecSize) {
```
niwinanto, line 1492:
```suggestion
  if (DstVecSize < 256) 
     return false;
```

llvm/lib/Target/AIE/AIECombinerHelper.cpp
```diff
@@ -1369,6 +1370,70 @@ bool llvm::applySingleDiffLaneBuildVector(
...
+  // splat vector combine already handles that case.
+  if (FirstOp.isIdenticalTo(SecondOp))
+    return false;
+  // Ensures that each operand in the first half matches FirstOp, and each
+  // operand in the second half matches SecondOp.
+  for (unsigned i = 2; i <= HalfNumElts; i++) {
+    if (!MI.getOperand(i).isIdenticalTo(FirstOp)) {
+      return false;
+    }
+    if (!MI.getOperand(HalfNumElts + i).isIdenticalTo(SecondOp)) {
```
niwinanto: May be, no need to satisfy both half needs to be splat vectors to apply this combine, just one is enough. It helps to optimize half. Also, I can see more opportunity with some patterns, for example 3/4th with one element and 1/4th other, literately split into 2 and then into 4 giving us good optimization. What do you think?

llvm/lib/Target/AIE/AIECombine.td
abnikant: I think this solution is not scalable/maintainable, what if there are new generic combiner added? Is there no other way to filter the combiner we don't need? @konstantinschwarz ?

llvm/lib/Target/AIE/AIECombinerHelper.cpp
abhinay-anubola: These G_BUILD_VECTORs will turn into BCST once we have [support for 128-bit splat](https://github.com/Xilinx/llvm-aie/pull/401)

abhinay-anubola: Oh there is typo here, type should be <16 x s16> and I will add line about broadcast too.

llvm/lib/Target/AIE/AIECombine.td
abhinay-anubola: This feature is mentioned on [llvm.org](https://lists.llvm.org/pipermail/llvm-dev/2018-November/127642.html), but it is not yet available in LLVM.
```
def CommonPreLegalizerCombiner: GICombiner<
      "CommonGenPreLegalizerCombiner", [all_combines]> {
      let DisableRuleOption = "common-prelegalizercombiner-disable-rule";
      let Modifiers = [(disable_rule copy_prop)]
    }

Comment (andcarminati) [reply llvm/lib/Target/AIE/AIECombine.td:14]: I think that listing the combiners can help to keep the generated code stable. The goal here is to exclude one combiner that undo what we are doing in this new combiner. Also, this can help to avoid surprises in case of a llvm bump.

Comment (abnikant) [reply llvm/lib/Target/AIE/AIECombine.td:14]: okay, I know the goal, just wanted to check if we have a better way to do this (like what is mentioned above, but not available yet). I think we should write comment about the purpose of doing this and also a TODO- to verify the QoR after accommodating new combiner (when we bump to next version).

File: llvm/lib/Target/AIE/AIECombinerHelper.cpp (hunk 11 above)
Comment (andcarminati) [reply llvm/lib/Target/AIE/AIECombinerHelper.cpp]: True! if we have at least one condition, we already gain in one side.

File: llvm/lib/Target/AIE/AIECombinerHelper.cpp [hunk 12]
```diff
@@ -1385,6 +1386,72 @@ bool llvm::applySingleDiffLaneBuildVector(
...
+      return false;
+    }
+  }
+
+  // If both halves are the same register, it's effectively a splat, and the
+  // splat vector combine already handles that case.
+  if (FirstOp.isIdenticalTo(SecondOp))
+    return false;
+
+  MatchInfo = [&MI, &Observer, &MRI, DstVecTy](MachineIRBuilder &B) {
```
khallouh: This ` &MRI` capture is not used. you will probably get a warning

abhinay-anubola:
Working on more generic way to adding more patterns like quarters and single diff splat on parts.

llvm/lib/Target/AIE/AIECombine.td
martien-de-jong: I don't think that `<everything> - <what we explicitly don't want>` is any better than `<everything>` in the first place. <what we don't want> is usually the result of painstakingly finding out what is bothering us, searching through an unknown bunch of combiners we have just assumed would be doing the right thing. Maintaining the list by trying newly published ones first and  checking effects on QoR doesn't scare me as non-scalable. 
At the point where we rebase on a new llvm, the less uncontrolled change we have, the better. Once it is up and running it is natural to play with the new toys, updating the combine list based on the release notes, or differences in the generic list.

llvm/lib/Target/AIE/AIECombinerHelper.cpp
martien-de-jong: And pray no one gives us 384 bit vectors?

llvm/lib/Target/AIE/AIECombinerHelper.cpp
```diff
@@ -1385,6 +1386,72 @@ bool llvm::applySingleDiffLaneBuildVector(
...
+    if (!getSrcOperand(i).isIdenticalTo(FirstOp)) {
+      return false;
+    }
+    if (!getSrcOperand(HalfNumElts + i).isIdenticalTo(SecondOp)) {
+      return false;
+    }
+  }
+
+  // If both halves are the same register, it's effectively a splat, and the
+  // splat vector combine already handles that case.
```
martien-de-jong: If we have determined, at cost, that this is a splat vector, I would say there's no need to match this build vector again. Why don't we remove the splat combine, returning the splat transform lambda here?
In general, building vector patterns from subvector patterns is a great way of software reuse.

```
splat1 = matchSplat(0, halfsize);
splat2 = matchSplat(halfsize, size);
if (splat1 && splat2) {
  if (*splat1 == *splat2) {
    MatchInfo = Splat(*splat1);
    return true;
  } 
  MatchInfo = ConcatSplat(*splat1, *splat2);
  return true;
}
return false;  
```

We now have all the splat logic in one place; we might spot that one case is missing, which is the opportunity to implement half of this buildvector as a splat and the rest with the fall back case.

llvm/lib/Target/AIE/AIECombinerHelper.cpp
```diff
@@ -1449,7 +1445,8 @@ bool llvm::matchSymmetricBuildVector(MachineInstr &MI, MachineRegisterInfo &MRI,
         DstVecTy.changeElementCount(
             DstVecTy.getElementCount().divideCoefficientBy(2)));
   };
-  return true;
+
+  return (FirstHalfSrcReg.has_value() || SecondHalfSrcReg.has_value());
```
martien-de-jong, line 1523: Better, but we can do more. We have now reused the matching code in the full splat vector match, but we still match twice: once to find two sub-splats and build two smaller BUILD_VECTORS and then once more to properly implement the splat vectors. 
Instead, we can *remove* the original splat vector match and implement all splats here, because we have already done the full match. So if both subvectors are splats and identical, try to generate a full-width broadcast, otherwise if either of them are splats generate a concat with appropriate splat  and buildvector.
Please note that creating a concat will hide the original full splat pattern, and we may be lucky that it is tried first, or we are painstakingly recognising the full width splat from the concat pattern.

Please leave this as a follow-up story with a TODO comment in the code.

llvm/lib/Target/AIE/AIECombinerHelper.cpp
```diff
@@ -1157,6 +1158,40 @@ void llvm::applyExtractVecEltAndExt(
...
+getSplatVectorSrcReg(const MachineInstr &MI, const MachineRegisterInfo &MRI,
+                     std::pair<unsigned, unsigned> Range) {
+  auto IsUndef = [&](const MachineOperand &Op) {
+    const MachineInstr *Undef = MRI.getVRegDef(Op.getReg());
+    return Undef && Undef->getOpcode() == TargetOpcode::G_IMPLICIT_DEF;
+  };
+  const unsigned Start = Range.first;
+  const unsigned End = Range.second;
+  // First non-undef operand.
+  unsigned SrcReg = 0;
```
andcarminati: Even if the underlying data is unsigned, I would prefer to see `Register` here. Same for the return type.

andcarminati: Ok,  it was refactored, but if you don't mind to change it would be nice.

andcarminati:
Hi @abhinay-anubola, do you have some QoR numbers to share?

andcarminati (approved):
LGTM.

martien-de-jong (approved):
There are some rough edges, but let's merge this one, which covers most of the occurrences.

PR #401, opened 2025-03-24
[AIE2][AIE2P] Support 128-bit G_BUILD_VECTOR combines

abhinay-anubola:


llvm/lib/Target/AIE/AIECombinerHelper.cpp
```diff
@@ -1257,10 +1258,10 @@ static void buildBroadcastVector(MachineIRBuilder &B, MachineRegisterInfo &MRI,
     // Build the G_AIE_BROADCAST_VECTOR instruction for the 512-bit vector.
     B.buildInstr(AIETII.getGenericBroadcastVectorOpcode(), {DstVec512BitReg},
                  {SrcReg});
-    if (DstVecSize == 256) {
-      const Register UnusedSubReg = MRI.createGenericVirtualRegister(DstVecTy);
-      // Unmerge the 512-bit vector into the 256-bit destination vector.
-      B.buildUnmerge({DstVecReg, UnusedSubReg}, DstVec512BitReg);
+    if (DstVecSize == 128 || DstVecSize == 256) {
+      // Unpad the 512-bit vector to form a 128/256-bit destination vector.
+      B.buildInstr(AIETII.getGenericUnpadVectorOpcode(), {DstVecReg},
```
khallouh: Could we keep using Unmerge here? It would be more generic and we already have a combine that will turn it into unpad if the target allows it

khallouh (commented):
Looks good. Would be nice to have some end-2-end tests

khallouh (approved):
LGTM

PR #403, opened 2025-03-25
[AIE2P] Fix shuffle_concat_extracted_subvectors combine  

khallouh:
This fixes a bug where we try to extract at an index that is not a multiple of the number of detected subvector elements. See test file.
Note: The bug was found accidentally and didn't affect any of the kernels.


llvm/lib/Target/AIE/AIECombinerHelper.cpp
```diff
@@ -2745,6 +2745,9 @@ bool llvm::matchShuffleToConcatExtractedSubvectors(MachineInstr &MI,
       CurrentMask = SubMask;
       CurrentHeight = *MaskMatch::getHeight(CurrentMask, /*Period*/ 0);
 
+      if (!isPowerOfTwoOrZero((CurrentHeight) % NumSrc1Elems))
```
martien-de-jong, line 2748: nit: The precedence of tokenization is infinite. We can lose a pair of parentheses.

martien-de-jong (approved):

PR #388, opened 2025-03-13
[AIE2P] Allow G_IMPLICIT_DEF as valid operand in broadcast

abnikant:
1. ) G_BUILD_VECTOR(undef, def) -> AIE_BROADCAST_VECTOR def
2. ) G_SHUFFLE_VECTOR (AIE_BROADCAST_VECTOR, src2, shufflemask(0)) -> AIE_BROADCAST_VECTOR

PR #404, opened 2025-03-25
[AIE2P] fixup! add check for implicit_def operand in matchShuffleBcstToCopy

abnikant:
This is the fixup commit of the PR - https://github.com/Xilinx/llvm-aie/pull/388 , add the checks for G_IMPLICIT_DEF operand.

### DIFF
--- a/llvm/lib/Target/AIE/AIECombinerHelper.cpp
+++ b/llvm/lib/Target/AIE/AIECombinerHelper.cpp
@@ -2866,17 +2866,21 @@ bool llvm::matchShuffleBcstToCopy(MachineInstr &MI, MachineRegisterInfo &MRI,
 
   const Register DstReg = MI.getOperand(0).getReg();
   const Register Src1Reg = MI.getOperand(1).getReg();
+  const Register Src2Reg = MI.getOperand(2).getReg();
   const LLT DstTy = MRI.getType(DstReg);
   const LLT Src1Ty = MRI.getType(Src1Reg);
-
   if (DstTy != Src1Ty)
     return false;
 
   ArrayRef<int> Mask = MI.getOperand(3).getShuffleMask();
   const AIEBaseInstrInfo &AIETII = (const AIEBaseInstrInfo &)TII;
-  MachineInstr *SrcVec = MRI.getVRegDef(Src1Reg);
+  MachineInstr *SrcVec1 = MRI.getVRegDef(Src1Reg);
+  MachineInstr *SrcVec2 = MRI.getVRegDef(Src2Reg);
   // Check if the first source is defined by a Broadcast.
-  if (SrcVec->getOpcode() != AIETII.getGenericBroadcastVectorOpcode())
+  if (SrcVec1->getOpcode() != AIETII.getGenericBroadcastVectorOpcode())
+    return false;
+  // Check if the second source is an undef.
+  if (SrcVec2->getOpcode() != TargetOpcode::G_IMPLICIT_DEF)
     return false;
 
   const unsigned NumSrcElems = Src1Ty.isVector() ? Src1Ty.getNumElements() : 1;
--- a/llvm/test/CodeGen/AIE/aie2p/GlobalIsel/prelegalizercombiner-vector-broadcast.mir
+++ b/llvm/test/CodeGen/AIE/aie2p/GlobalIsel/prelegalizercombiner-vector-broadcast.mir
@@ -454,6 +454,24 @@ body:             |
 
 ...
 ---
+name:       test_shuffle_broadcast_combine_v64s32_invalid
+alignment:       16
+body:             |
+  bb.1.entry:
+    ; CHECK-LABEL: name: test_shuffle_broadcast_combine_v64s32_invalid
+    ; CHECK: [[C:%[0-9]+]]:_(s32) = G_CONSTANT i32 0
+    ; CHECK-NEXT: [[COPY:%[0-9]+]]:_(<64 x s32>) = COPY $dm0
+    ; CHECK-NEXT: [[AIE_BROADCAST_VECTOR:%[0-9]+]]:_(<64 x s32>) = G_AIE_BROADCAST_VECTOR [[C]](s32)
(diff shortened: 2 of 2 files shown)


llvm/lib/Target/AIE/AIECombinerHelper.cpp
```diff
@@ -2866,17 +2866,21 @@ bool llvm::matchShuffleBcstToCopy(MachineInstr &MI, MachineRegisterInfo &MRI,
...
-  MachineInstr *SrcVec = MRI.getVRegDef(Src1Reg);
+  MachineInstr *SrcVec1 = MRI.getVRegDef(Src1Reg);
+  MachineInstr *SrcVec2 = MRI.getVRegDef(Src2Reg);
   // Check if the first source is defined by a Broadcast.
-  if (SrcVec->getOpcode() != AIETII.getGenericBroadcastVectorOpcode())
+  if (SrcVec1->getOpcode() != AIETII.getGenericBroadcastVectorOpcode())
+    return false;
+  // Check if the second source is defined by an undef.
+  if (SrcVec2->getOpcode() != TargetOpcode::G_IMPLICIT_DEF)
     return false;
```
khallouh, line 2884: Could you add a test case showcasing the bug?

llvm/lib/Target/AIE/AIECombinerHelper.cpp
```diff
@@ -2866,17 +2866,21 @@ bool llvm::matchShuffleBcstToCopy(MachineInstr &MI, MachineRegisterInfo &MRI,
...
 
   const unsigned NumSrcElems = Src1Ty.isVector() ? Src1Ty.getNumElements() : 1;
```
khallouh, line 2886: I see below we return false if the mask is all undefs. Why is that if we don't care about the mask otherwise?

llvm/lib/Target/AIE/AIECombinerHelper.cpp
abnikant: done

llvm/lib/Target/AIE/AIECombinerHelper.cpp
abnikant: since all the lanes have same value in broadcast vector.

abnikant: and say if the mask is >=16 then it will choose from undef, in that case too it is okay to create a vector with a value in the source operand of the broadcast.

khallouh (approved):
LGTM

llvm/lib/Target/AIE/AIECombinerHelper.cpp
```diff
@@ -2866,17 +2866,21 @@ bool llvm::matchShuffleBcstToCopy(MachineInstr &MI, MachineRegisterInfo &MRI,
 
   const Register DstReg = MI.getOperand(0).getReg();
```
niwinanto, line 2867: I recommend checking the mask element belongs to one of the source input, than fixing it simply checking the undef behavior of other input. Or may be both, we could cover more cases.

abnikant: yes, we can generalize this for other case e.g. I have posted on the channel,  %56:_(<16 x s32>) = G_INTRINSIC intrinsic(@llvm.aie2p.vshift.I512.I512), %47(<16 x s32>), %50(<16 x s32>), %12(s32), %54(s32), debug-location !437
    %57:_(<16 x s32>) = G_AIE_BROADCAST_VECTOR %58(s32)
    %60:_(<16 x s32>) = G_SHUFFLE_VECTOR %57(<16 x s32>), %56, shufflemask(16, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0), which we can do using bcst + insert . @khallouh if you are not working on this , then I can take this and work on generalization and open another PR.

khallouh: I'm not sure if the vshift and broadcast are part of the pattern. I have added a combine for that shuffle mask to convert into a broadcast and an insert. I don't care about the shuffle use operands.

khallouh: I don't see any undefs here. How do you plan to generalize this to encompass the above pattern?

khallouh: I think it's a different pattern than the case which Niwin is referring to

abnikant: @khallouh is this pattern handled (don't see any vpsuh) by you already or in progress ? Generalization I am talking about is one lank mask is different than zero.  I understood what @niwinanto mentioned, what I am saying is we can handle Niwin's suggestion on generalizing (any source can be a broadcast) and also handling the pattern I pasted here.

niwinanto: Well, what I meant is something like 
```
///  %1:_(s32) = COPY $r0
///  %2:_(<16 x s32>) = COPY $x0
///  %3:_(<16 x s32>) = G_AIE_BROADCAST_VECTOR %1:_(s32)
///  %0:_(<16 x s32>) = G_SHUFFLE_VECTOR %3(<16 x s32>), %2(<16 x s32>),
///  shufflemask(15, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0)
/// To convert to:
///  %0:_(<16 x s32>) = G_AIE_BROADCAST_VECTOR %1(s32)
```
by checking the mask value below 16.

abnikant: yes, I thought about it but this is a fixup PR where I was expecting second source to be implicit_def. I will open a PR with the following change- 1) your example 2) any source operand is broadcast (so mask is between 16 to 31 if second source is broadcast) 3) The pattern I mentioned above.

khallouh: I don't have a PR yet. As long as you don't look for a frequent index in the shuffle mask which you broadcast and insert the rest then we have no overlap. Regarding the above pattern I'm handing  this `%60:_(<16 x s32>) = G_SHUFFLE_VECTOR %57(<16 x s32>), %56, shufflemask(16, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0),` As I said I don't care about the inputs. Moreover, we have 16 and 0 so 2 different inputs... Maybe wait for my PR and see if it still makes sense to improve my combine result.

PR #407, opened 2025-03-26
[AIE2P] Combine shuffle to broadcast + insertions

khallouh:


llvm/lib/Target/AIE/AIECombinerHelper.cpp
```diff
@@ -125,6 +125,48 @@ std::optional<int> MaskMatch::getUniqueIndex(ArrayRef<int> Mask) {
...
+  std::optional<int> FrequentValue;
+  unsigned MaxFrequency = 0;
+
+  for (const auto &Entry : FrequencyMap) {
+    unsigned AdjustedFrequency = Entry.second + DontCareCount;
+    if (AdjustedFrequency >= MinFrequency && Entry.second > MaxFrequency) {
+      FrequentValue = Entry.first;
+      MaxFrequency = Entry.second;
+    }
+  }
```
martien-de-jong, line 169: Nit: This would be clearer if we would just compute the bare maximum frequency, and then filter it using
DontCareCount and MinFrequency. My gut feeling is that computing the frequency map in a separate method would be reusable, for instance to compute the number of  values to extract / insert that could be used as a separate filter/cost parameter.

llvm/lib/Target/AIE/AIECombinerHelper.cpp
```diff
@@ -125,6 +125,48 @@ std::optional<int> MaskMatch::getUniqueIndex(ArrayRef<int> Mask) {
...
+    for (unsigned I = 0; I < Mask.size(); I++) {
+      int Idx = Mask[I];
+      if (Idx != -1 && Idx != FrequentValue) {
+        NonMatchingCount++;
+      }
+      if (Idx == FrequentValue) {
+        FrequentIdx = I;
+      }
+    }
+    return FrequentIndexResult{FrequentIdx, NonMatchingCount};
```
martien-de-jong: Here too, we could supply more raw data in a separate method rather than reducing it to the one most frequent. For instance, we may find a clever scheme to deal with the most two frequent values

llvm/lib/Target/AIE/AIECombinerHelper.cpp
```diff
@@ -2581,6 +2623,113 @@ bool llvm::matchShuffleToCopy(MachineInstr &MI, MachineRegisterInfo &MRI,
...
+  if (DstTy != Src1Ty)
+    return false;
+
+  if (!DstTy.isVector() || !Src1Ty.isVector())
+    return false;
+
+  if (DstTy.getSizeInBits() < 128)
+    return false;
+
+  const unsigned NumSrcElems = Src1Ty.isVector() ? Src1Ty.getNumElements() : 1;
```
martien-de-jong: We won't be here if Src1Ty is not a vector.

llvm/lib/Target/AIE/AIECombinerHelper.cpp
```diff
@@ -2581,6 +2623,113 @@ bool llvm::matchShuffleToCopy(MachineInstr &MI, MachineRegisterInfo &MRI,
...
+    return false;
+
+  if (MaskMatch::isMaskWithAllUndefs(Mask))
+    return false;
+
+  unsigned MinFrequency;
+  if (MI.getMF()->getTarget().getTargetTriple().isAIE2P())
+    // The scalarization of G_SHUFFLE_VECTOR in the legalizer is more beneficial
+    // if there are more exceptions than NumSrcElems / 2 as AIE2P's VINSERT
+    // instrutions require a move to a register used for the index unlike VPUSH.
```
martien-de-jong, line 2689: nit: instructions

llvm/lib/Target/AIE/AIECombinerHelper.cpp
```diff
@@ -2581,6 +2623,113 @@ bool llvm::matchShuffleToCopy(MachineInstr &MI, MachineRegisterInfo &MRI,
...
+        DstElemTy, VecToExtract, Mask[FrequentIdx] % NumSrcElems);
+    buildBroadcastVector(B, MRI, Extr.getReg(0), BroadcastVecReg);
+
+    Register InsertSrc = BroadcastVecReg;
+    Register InsertDst;
+
+    unsigned InsertionCount = 0;
+    for (unsigned Idx = 0; Idx < Mask.size(); ++Idx) {
+
+      if (Mask[Idx] == Mask[FrequentIdx])
```
martien-de-jong: I would capture the  broadcasted mask value in stead of FrequentIdx

llvm/test/CodeGen/AIE/GlobalISel/prelegalizercombiner-shuffle-vector.mir
```diff
@@ -313,10 +313,12 @@ body:             |
...
     ; CHECK-NEXT: {{  $}}
-    ; CHECK-NEXT: [[COPY:%[0-9]+]]:_(<16 x s32>) = COPY $x2
-    ; CHECK-NEXT: [[DEF:%[0-9]+]]:_(<16 x s32>) = G_IMPLICIT_DEF
-    ; CHECK-NEXT: [[SHUF:%[0-9]+]]:_(<16 x s32>) = G_SHUFFLE_VECTOR [[DEF]](<16 x s32>), [[COPY]], shufflemask(0, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1)
-    ; CHECK-NEXT: PseudoRET implicit $lr, implicit [[SHUF]](<16 x s32>)
+    ; CHECK-NEXT: [[DEF:%[0-9]+]]:_(s32) = G_IMPLICIT_DEF
+    ; CHECK-NEXT: [[AIE_BROADCAST_VECTOR:%[0-9]+]]:_(<16 x s32>) = G_AIE_BROADCAST_VECTOR [[DEF]](s32)
+    ; CHECK-NEXT: [[C:%[0-9]+]]:_(s32) = G_CONSTANT i32 0
+    ; CHECK-NEXT: [[DEF1:%[0-9]+]]:_(s32) = G_IMPLICIT_DEF
+    ; CHECK-NEXT: [[IVEC:%[0-9]+]]:_(<16 x s32>) = G_INSERT_VECTOR_ELT [[AIE_BROADCAST_VECTOR]], [[DEF1]](s32), [[C]](s32)
```
martien-de-jong, line 320: Will this be covered by VPUSH.lo?

llvm/test/CodeGen/AIE/aie2p/shufflevec.ll
```diff
@@ -221,25 +221,12 @@ define <16 x i32> @test_shuffle_vector_vsel_invalid_mask(<16 x i32> noundef %a,
...
-; CHECK-NEXT:    vpush.hi.32 x0, x0, r1
-; CHECK-NEXT:    vpush.hi.32 x0, x0, r1
+; CHECK-NEXT:    mova r29, #0; nopxm
 ; CHECK-NEXT:    ret lr
-; CHECK-NEXT:    vpush.hi.32 x0, x0, r1 // Delay Slot 5
-; CHECK-NEXT:    vpush.hi.32 x0, x0, r1 // Delay Slot 4
-; CHECK-NEXT:    vpush.hi.32 x0, x0, r1 // Delay Slot 3
-; CHECK-NEXT:    vpush.hi.32 x0, x0, r1 // Delay Slot 2
+; CHECK-NEXT:    vextract.32 r0, x2, #1, vaddsign1 // Delay Slot 5
+; CHECK-NEXT:    vextract.32 r0, x2, #0, vaddsign1 // Delay Slot 4
```
martien-de-jong: This looks strange. r0 is overwritten? or does the RAW latency equal two?

llvm/lib/Target/AIE/AIECombinerHelper.cpp
```diff
@@ -125,6 +125,48 @@ std::optional<int> MaskMatch::getUniqueIndex(ArrayRef<int> Mask) {
   return UniqOpIdx;
 }
 
+std::optional<FrequentIndexResult>
+MaskMatch::getFrequentIndexResult(ArrayRef<int> Mask, unsigned MinFrequency) {
+  std::unordered_map<int, unsigned> FrequencyMap;
```
andcarminati, line 153: This could be a vector with the same size as Mask.

llvm/test/CodeGen/AIE/aie2p/shufflevec.ll
khallouh: Yes, vextract writes in E2

llvm/test/CodeGen/AIE/GlobalISel/prelegalizercombiner-shuffle-vector.mir
khallouh: This G_INSERT_VECTOR_ELT selects to vinsert. VPUSH.lo would result from shuffle legalization if we had not combined here.

llvm/lib/Target/AIE/AIECombinerHelper.cpp
khallouh: I am just comparing mask indices and everything is captured here in the lambda. What _broadcasted mask value_ you're referring to?

martien-de-jong: [ BcstValue = Mask[FrequentIdx] ] in the capture list, or compute BCstValue outside of the lambda and use it inside it

llvm/lib/Target/AIE/AIECombinerHelper.cpp
martien-de-jong: Twice the size of Mask. But you might have many zero entries.

llvm/lib/Target/AIE/AIECombinerHelper.cpp
khallouh: Done

llvm/lib/Target/AIE/AIECombinerHelper.cpp
khallouh: Yes, I even tried handling any number of frequent values by splitting the shuffle using the `fewerElements` action if we find a frequent values in its halves but I had to include a bunch of corner cases given the legalizer/isel support we have for broadcast/insert and some generic combines kicked in as well. We will also need an appropriate cost model. I think this will be easier as we cleanup our legalization strategy for insert as we have done for extract. But since this is not critical and is part of a broader combine, it could introduce instabilities. I created this followup ticket for later AIECC-959.

llvm/lib/Target/AIE/AIECombinerHelper.cpp
```diff
@@ -136,6 +148,57 @@ std::optional<int> MaskMatch::getUniqueIndex(ArrayRef<int> Mask) {
...
+      FrequencyMap.begin(), FrequencyMap.end(),
+      [](const std::pair<int, unsigned> p1, const std::pair<int, unsigned> p2) {
+        return p1.second < p2.second;
+      });
+
+  if (MostFrequentPair->second + DontCareCount < MinFrequency) {
+    return std::nullopt;
+  }
+
+  int FrequentValue = FrequentValue = MostFrequentPair->first;
```
martien-de-jong: It may work. It looks strange though. While at it, could you use structural binding to two variables for the max_element call?

llvm/lib/Target/AIE/AIECombinerHelper.cpp
```diff
@@ -136,6 +148,57 @@ std::optional<int> MaskMatch::getUniqueIndex(ArrayRef<int> Mask) {
...
+
+  unsigned NonMatchingCount =
+      Mask.size() - MostFrequentPair->second - DontCareCount;
```
martien-de-jong: nit: I think MostFrequentPair->second + DontCareCount could be reused from a variable.

llvm/lib/Target/AIE/AIECombinerHelper.cpp
```diff
@@ -136,6 +148,57 @@ std::optional<int> MaskMatch::getUniqueIndex(ArrayRef<int> Mask) {
...
+
+  unsigned FrequentIdx = 0;
+  for (unsigned I = 0; I < Mask.size(); I++) {
+    int Idx = Mask[I];
```
martien-de-jong: nit: rename Idx to MaskValue ? Especially since we are searching for FrequentIdx.

martien-de-jong (dismissed):
Just nits. Ping we if you want to fix it and need reapproval.

llvm/lib/Target/AIE/AIECombinerHelper.cpp
khallouh: It was strange indeed. Done.

martien-de-jong (approved):

PR #408, opened 2025-03-26
Fix alias analysis bug for virtual unrollong in postpipeliner

martien-de-jong:
This is a small fix with important consequences.


andcarminati (approved):
LGTM.

llvm/test/CodeGen/AIE/aie2p/AA-unroll-iterations.mir
```diff
@@ -0,0 +1,186 @@
+# NOTE: Assertions have been autogenerated by utils/update_llc_test_checks.py UTC_ARGS: --version 4
```
niwinanto, line 1: Why we are not using `utils/update_mir_test_checks.py` here?

andcarminati: Because it is a IR test. Maybe we could rename to .ll

martien-de-jong: As explained offline. The postpipeliner is one of the last big modifiers, and I find the assembly much more readable than the MIR. (MIR bundles are split across lines; the number of CHECK lines showing the loop body doesn't fit in one screen.)

PR #409, opened 2025-03-27
Extend virtual unroll AA to cover GEPs

andcarminati:


llvm/test/CodeGen/AIE/aie2p/end-to-end/softmax-aa.ll
```diff
@@ -0,0 +1,468 @@
+; NOTE: Assertions have been autogenerated by utils/update_llc_test_checks.py
+;
+; This file is licensed under the Apache License v2.0 with LLVM Exceptions.
+; See https://llvm.org/LICENSE.txt for license information.
+; SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+;
+; (c) Copyright 2025 Advanced Micro Devices, Inc. or its affiliates
+; RUN: llc -mtriple=aie2p --aie-force-postpipeliner \
+; RUN:   -aie-preassign-multi-slot-instr=true %s -o - | FileCheck %s
```
martien-de-jong: nit: for test stability, could we make it less end-to-end? i.e. create input to postscheduler in .mir?

llvm/test/CodeGen/AIE/aie2p/end-to-end/softmax-aa.ll
```diff
@@ -0,0 +1,468 @@
...
+  %2 = shufflevector <32 x float> %1, <32 x float> poison, <16 x i32> <i32 0, i32 1, i32 2, i32 3, i32 4, i32 5, i32 6, i32 7, i32 8, i32 9, i32 10, i32 11, i32 12, i32 13, i32 14, i32 15>
+  %3 = tail call noundef <16 x bfloat> @llvm.aie2p.v16accfloat.to.v16bf16(<16 x float> %2)
+  %add.ptr.i.i.i.i28 = getelementptr inbounds i8, ptr %it_soft_out.sroa.0.131, i20 32
+  store <16 x bfloat> %3, ptr %it_soft_out.sroa.0.131, align 64
+  %4 = call i1 @llvm.loop.decrement.i32(i32 1)
+  br i1 %4, label %for.body5, label %for.cond.cleanup4, !llvm.loop !5
+}
+
+; The main goal of this test is to pipeline the code in presence of lockstep pointer uptates in the outermost loops.
```
martien-de-jong: Perhaps mention that the essence is that the pointers are only equal after traversing more than the loop's phis.

llvm/lib/Target/AIE/AIEBaseAliasAnalysis.cpp
```diff
@@ -183,32 +184,39 @@ struct IntrinsicArgsInfo {
...
+  }
 
   // Same number of call until the point of interest.
-  bool hasSameNumberOfCalls(const IntrinsicChainInfo &Other) const {
-    return IntrinsicCalls == Other.IntrinsicCalls;
+  bool hasSameNumberOfUpdates(const PtrUpdateChainInfo &Other) const {
+    return PointerUpdates == Other.PointerUpdates;
   }
 
   // Same number of calls accross the loop.
```
martien-de-jong: nit: across

llvm/test/CodeGen/AIE/aie2p/end-to-end/softmax-aa.ll
andcarminati: The problem is that we need IR anyway (because we look to it in the AA).

llvm/lib/Target/AIE/AIEBaseAliasAnalysis.cpp
```diff
@@ -332,21 +349,41 @@ trackIntrinsicChain(const Value *LastUpdate, const Value *TargetPointer) {
...
     if (!LastUpdate->getType()->isPointerTy())
       return std::nullopt; // not a pointer.
 
     if (auto *GEP = dyn_cast<GEPOperator>(LastUpdate)) {
-      if (!GEP->hasAllZeroIndices())
-        return std::nullopt; // No pointer arithmetic is allowed.
+      // We handle GEP with a single index bacause of the PostSWP.
+      if (!GEP->hasAllZeroIndices() && GEP->getNumOperands() == 2) {
+
+        if (FoundIntrinsicChain) {
```
martien-de-jong, line 361: Why can't we allow mixed chains? What's special about a single Index? To me GEP(ptr, I1, I2) would be fully analogous to add.2d(ptr, stride, count).
Apart from that, the condition seems strange, since we no longer return nullopt if we have more than two indices

llvm/test/CodeGen/AIE/aie2p/end-to-end/softmax-aa.ll
martien-de-jong: Sure, but the IR doesn't change, and will be the same if we dump the mir just before postmisched. But we will prevent updates from changed combines, matches, register allocation, and default settings for all options of these previous passes.

llvm/lib/Target/AIE/AIEBaseAliasAnalysis.cpp
andcarminati: By mixed I mean something like:

```
p = phi (p2...)
p1 = gep (p, ..)
store (p1, ...)
p2 = add2d(p1, ...)
store (p2, ...)
```

The algorithm relies on the homogeneous update property alongside the chain and between memory operations. Only GEP with zero index is allowed between intrinsic calls.

martien-de-jong:
I was expecting that we would have a set of legal operations in a pointer chain which should be the same in both chains, then a PHI in the loop block, then any legal operation or phi until we find a common pointer base, until we find an illegal operation, or until there's a difference between the chains. The first PHI node needs to be traced back to the original value in order to incorporate the iteration index reliably. I would do this in a loop that steps through both chains in parallel and performs the checks and comparisons as we traverse both of them, to the next node in the chain.

llvm/test/CodeGen/AIE/aie2p/end-to-end/softmax-aa.ll
andcarminati: In this case, the test was replaced by a .mir one that starts before postmisched.

andcarminati:
> I was expecting that we would have a set of legal operations in a pointer chain which should be the same in both chains, then a PHI in the loop block, then any legal operation or phi until we find a common pointer base, until we find an illegal operation, or until there's a difference between the chains. The first PHI node needs to be traced back to the original value in order to incorporate the iteration index reliably. I would do this in a loop that steps through both chains in parallel and performs the checks and comparisons as we traverse both of them, to the next node in the chain.

Yes, that is the purpose `isLockStepGEPChain` - except that it recursively walks upwards. We traverse GEPs, PHI and AS casts nodes only.

llvm/lib/Target/AIE/AIEBaseAliasAnalysis.cpp
martien-de-jong: As discussed offline. We could generalize the code to make it simpler and better testable, but only when time allows.

llvm/lib/Target/AIE/AIEBaseAliasAnalysis.cpp
```diff
@@ -359,36 +404,88 @@ trackIntrinsicChain(const Value *LastUpdate, const Value *TargetPointer) {
...
       LastUpdate = AIEObject;
-      ChainInfo.TotalIntrinsicCalls++;
+      ChainInfo.TotalPointerUpdates++;
+      FoundIntrinsicChain = true;
     }
   }
   return std::nullopt;
 }
 
+static bool isLockStepGEPChain(std::set<const Value *> &Visited,
```
martien-de-jong: nit: `unordered_set`?

llvm/lib/Target/AIE/AIEBaseAliasAnalysis.cpp
```diff
@@ -359,36 +404,88 @@ trackIntrinsicChain(const Value *LastUpdate, const Value *TargetPointer) {
...
+                              PhiB->getIncomingValue(I)))
+        return false;
+    }
+    return true;
+  }
+
+  const GEPOperator *GEPA = dyn_cast<GEPOperator>(ValueA);
+  const GEPOperator *GEPB = dyn_cast<GEPOperator>(ValueB);
+
+  if (GEPA && GEPB && GEPA->getNumOperands() == GEPB->getNumOperands()) {
```
martien-de-jong: early return on the negated condition?

llvm/lib/Target/AIE/AIEBaseAliasAnalysis.cpp
```diff
@@ -359,36 +404,88 @@ trackIntrinsicChain(const Value *LastUpdate, const Value *TargetPointer) {
...
+  if (Visited.count(ValueA) && Visited.count(ValueB))
+    return true;
+
+  Visited.insert(ValueA);
+  Visited.insert(ValueB);
+
+  const PHINode *PhiA = dyn_cast<PHINode>(ValueA);
+  const PHINode *PhiB = dyn_cast<PHINode>(ValueB);
+
+  if (PhiA && PhiB) {
```
martien-de-jong, line 462: Shouldn't we check they are in the same block?

martien-de-jong (approved):

andcarminati:
No QoR regressions.

PR #410, opened 2025-03-27
[AIE2P] Add shuffle mask range checks in matchShuffleBcstToCopy combiner

abnikant:
If either vector source of `G_SHUFFLE_VECTOR` is defined by `G_AIE_BROADCAST_VECTOR`, combine it into a `COPY` operation if the mask falls within a valid range.

### DIFF
--- a/llvm/lib/Target/AIE/AIECombinerHelper.cpp
+++ b/llvm/lib/Target/AIE/AIECombinerHelper.cpp
@@ -96,6 +96,16 @@ bool MaskMatch::isMaskWithAllUndefs(ArrayRef<int> Mask) {
   return true;
 }
 
+// Checks if the mask is in range, allowing some values to be
+// undef (-1), but not all.
+bool MaskMatch::isMaskWithinRangeOrUndef(ArrayRef<int> Mask, int MinValue,
+                                         int MaxValue) {
+  return llvm::any_of(Mask, [](int v) { return v != -1; }) &&
+         llvm::all_of(Mask, [=](int v) {
+           return v == -1 || (v >= MinValue && v <= MaxValue);
+         });
+}
+
 std::optional<unsigned> MaskMatch::getHeight(ArrayRef<int> Mask,
                                              unsigned Period) {
   for (unsigned I = 0; I < Mask.size(); ++I) {
@@ -2869,8 +2879,9 @@ bool llvm::matchShuffleToConcatExtractedSubvectors(MachineInstr &MI,
   return true;
 }
 
+/// Match something like this:
 ///  %1:_(s32) = COPY $r0
-///  %2:_(<16 x s32>) = G_IMPLICIT_DEF
+///  %2:_(<16 x s32>) = COPY $x0
 ///  %3:_(<16 x s32>) = G_AIE_BROADCAST_VECTOR %1:_(s32)
 ///  %0:_(<16 x s32>) = G_SHUFFLE_VECTOR %3(<16 x s32>), %2(<16 x s32>),
 ///  shufflemask(0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0)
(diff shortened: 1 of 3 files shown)


llvm/lib/Target/AIE/AIECombinerHelper.cpp
```diff
@@ -2876,23 +2884,32 @@ bool llvm::matchShuffleBcstToCopy(MachineInstr &MI, MachineRegisterInfo &MRI,
...
     return false;
 
   if (MaskMatch::isMaskWithAllUndefs(Mask))
     return false;
 
-  MatchInfo = [=](MachineIRBuilder &B) { B.buildCopy(DstReg, Src1Reg); };
+  // Determine the valid mask range
+  const int MinValue = IsSrc1Bcst ? 0 : NumSrcElems;
+  const int MaxValue = IsSrc2Bcst ? (2 * NumSrcElems - 1) : (NumSrcElems - 1);
+  bool IsValid = MaskMatch(0).isMaskWithinRange(Mask, MinValue, MaxValue);
```
andcarminati: nit: `const bool`

llvm/lib/Target/AIE/AIECombinerHelper.cpp
```diff
@@ -2876,23 +2884,32 @@ bool llvm::matchShuffleBcstToCopy(MachineInstr &MI, MachineRegisterInfo &MRI,
...
-  if (SrcVec2->getOpcode() != TargetOpcode::G_IMPLICIT_DEF)
-    return false;
+  MachineInstr *Src1Vec = MRI.getVRegDef(Src1Reg);
+  MachineInstr *Src2Vec = MRI.getVRegDef(Src2Reg);
+  unsigned BcstOpcode = AIETII.getGenericBroadcastVectorOpcode();
+  bool IsSrc1Bcst = (Src1Vec->getOpcode() == BcstOpcode);
+  bool IsSrc2Bcst = (Src2Vec->getOpcode() == BcstOpcode);
 
+  // Check if the first or second source is defined by a Broadcast.
+  if (!IsSrc1Bcst && !IsSrc2Bcst)
```
andcarminati, line 2912: I think this is not correct. The code was written with just one broadcast in mind. If we have two, it will get `MinValue = 0` and `MaxValue = (2 * NumSrcElems - 1)` (full vector), and then it will select a copy from the first. 

Please, test this case:
```
---
name:       test
alignment:       16
body:             |
  bb.1.entry:
  %1:_(s32) = COPY $r0
  %6:_(s32) = COPY $r1
  %2:_(<16 x s32>) = G_AIE_BROADCAST_VECTOR %1:_(s32)
  %3:_(<16 x s32>) = G_AIE_BROADCAST_VECTOR %6:_(s32)
  %0:_(<16 x s32>) = G_SHUFFLE_VECTOR %3:_(<16 x s32>), %2:_, shufflemask(1, 18, 18, 18, 18, 18, 18, 18, 18, 2, 18, 18, 18, 18, 18, 18)
```
The easiest way to fix is:
```
if (IsSrc1Bcst == IsSrc2Bcst)
   return false;
```

llvm/lib/Target/AIE/AIECombinerHelper.cpp
abnikant: fixed.

llvm/lib/Target/AIE/AIECombinerHelper.cpp
martien-de-jong: or in one go` if (IsSrc1Bst + IsSrc2Bcst != 1) return false;`

abnikant: yes, you are right. I was just thinking about one broadcast source. I have updated the code to support any or both sources as broadcast. Thanks for catching this.

martien-de-jong: But we could also check both ranges separately.

llvm/lib/Target/AIE/AIECombinerHelper.cpp
```diff
@@ -2876,23 +2884,50 @@ bool llvm::matchShuffleBcstToCopy(MachineInstr &MI, MachineRegisterInfo &MRI,
...
+    if (IsValidSrc1Mask)
+      SrcReg = Src1Reg;
+    else if (IsValidSrc2Mask)
+      SrcReg = Src2Reg;
+  } else {
+    if (IsSrc1Bcst && IsValidSrc1Mask)
+      SrcReg = Src1Reg;
+    else if (IsSrc2Bcst && IsValidSrc2Mask)
+      SrcReg = Src2Reg;
+  }
```
martien-de-jong: This can be further streamlined. 
1) We only have to check whether it is a valid src mask if the corresponding operand is a broadcast.
2) We only have to check the second operand if the first operand fails.
```
if (broadcast1 && inrange(mask, 0, NumSrcElems))
  SrcReg = Src1Reg;
else if (broadcas2 && inrange(mask, NumSrcElems. NumSrcElems)
  SrcReg = Src2Reg;
else 
  return false; 
```
   (using {start, size} convention for ranges here)

abnikant: refactored.

llvm/lib/Target/AIE/AIECombinerHelper.cpp
```diff
@@ -96,6 +96,13 @@ bool MaskMatch::isMaskWithAllUndefs(ArrayRef<int> Mask) {
   return true;
 }
 
+bool MaskMatch::isMaskWithinRange(ArrayRef<int> Mask, int MinValue,
+                                  int MaxValue) {
+  return llvm::all_of(Mask, [MinValue, MaxValue](int Val) {
```
niwinanto: Do you think, we should take care of undef(-1) as well?

llvm/lib/Target/AIE/AIECombinerHelper.cpp
```diff
@@ -2891,23 +2899,42 @@ bool llvm::matchShuffleBcstToCopy(MachineInstr &MI, MachineRegisterInfo &MRI,
...
     return false;
-  // Check if the second source is an undef.
-  if (SrcVec2->getOpcode() != TargetOpcode::G_IMPLICIT_DEF)
+
+  if (!DstTy.isVector() || !Src1Ty.isVector())
     return false;
 
-  const unsigned NumSrcElems = Src1Ty.isVector() ? Src1Ty.getNumElements() : 1;
+  const unsigned NumSrcElems = Src1Ty.getNumElements();
   if (Mask.size() != NumSrcElems)
```
niwinanto, line 2916: May be `Mask.size() > NumSrcElems`, we can cover more cases.

llvm/lib/Target/AIE/AIECombinerHelper.cpp
```diff
@@ -2891,23 +2899,42 @@ bool llvm::matchShuffleBcstToCopy(MachineInstr &MI, MachineRegisterInfo &MRI,
...
     return false;
 
   if (MaskMatch::isMaskWithAllUndefs(Mask))
     return false;
 
-  MatchInfo = [=](MachineIRBuilder &B) { B.buildCopy(DstReg, Src1Reg); };
+  // Determine the valid mask range
+  const int MinSrc1Value = 0, MaxSrc1Value = NumSrcElems - 1;
+  const int MinSrc2Value = NumSrcElems, MaxSrc2Value = 2 * NumSrcElems - 1;
+  const bool IsValidSrc1Mask =
```
niwinanto, line 2925: nit: `const bool IsValidSrc1Mask = IsSrc1Bcst ? MaskMatch(0).isMaskWithinRange(Mask, MinSrc1Value, MaxSrc1Value) : false;`

llvm/lib/Target/AIE/AIECombinerHelper.cpp
abnikant: do we have any practical example generating the mix of undef and def masks?

niwinanto: I dont know we have this pattern in mllib or not. But there were many patterns with mask containing def and undef values. Same example can be like this
```
/// Match something like this:
///  %1:_(s32) = COPY $r0
///  %2:_(<16 x s32>) = COPY $x0
///  %3:_(<16 x s32>) = G_AIE_BROADCAST_VECTOR %1:_(s32)
///  %0:_(<16 x s32>) = G_SHUFFLE_VECTOR %3(<16 x s32>), %2(<16 x s32>),
///  shufflemask(0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, -1, -1, -1)
/// To convert to:
///  %0:_(<16 x s32>) = G_AIE_BROADCAST_VECTOR %1(s32)
```

abnikant: okay, I have updated the function to handle undef and added some tests.

llvm/lib/Target/AIE/AIECombinerHelper.cpp
```diff
@@ -96,6 +96,16 @@ bool MaskMatch::isMaskWithAllUndefs(ArrayRef<int> Mask) {
   return true;
 }
 
+// Checks if the mask is in range, allowing some values to be
+// undef (-1), but not all.
+bool MaskMatch::isMaskWithinRangeOrUndef(ArrayRef<int> Mask, int MinValue,
+                                         int MaxValue) {
+  return llvm::any_of(Mask, [](int v) { return v != -1; }) &&
```
niwinanto, line 103: Why we need `llvm::any_of(Mask, [](int v) { return v != -1; })`?

abnikant: Since I wanted it to return false when all of the masks are defined as -1

niwinanto: Then you could remove this check
```
  if (MaskMatch::isMaskWithAllUndefs(Mask))
    return false;
```

abnikant: done

niwinanto (approved):
LGTM!

PR #411, opened 2025-03-27
[AIE2P] Move AIE2P slot decoders to a namespace to avoid link-time conflicts with AIE2 ones.

SagarMaheshwari99:
Apparently MSVC linker thought it was okay to remove one of the conflicting decoder definitions at link time. (On the other hand, it is dubious as to how this didn't show up with other compilers).


martien-de-jong (approved):
Interesting.

PR #413, opened 2025-03-28
[AIEX] Fix shared library build

konstantinschwarz:
Commit 8978b1c started using getLoopID from the Analysis library. We now need to add this library to the dependencies of the AIEUtils library.


andcarminati (approved):
LGTM.

PR #415, opened 2025-03-31
[AIE2P] Add a selection combiner for BCAST -> SHIFT

andcarminati:
Under certain conditions, we can have a VPUSH_HI.


llvm/lib/Target/AIE/aie2p/AIE2PInstrPatterns.td
```diff
@@ -862,6 +862,10 @@ def : Pat<(int_aie2p_vshift_I512_I512 VEC512:$s1, VEC512:$s2, mSs:$pre, eR:$shif
 def : Pat<(int_aie2p_vshift_bf512_bf512 VEC512:$s1, VEC512:$s2, mSs:$pre, eR:$shift),
            (VSHIFT_ALIGN VEC512:$s1, mSs:$pre, VEC512:$s2, eR:$shift)>;
 
+// Combine broadcast + shift into VPUSH_hi_64
+def : Pat<(int_aie2p_vshift_I512_I512 (v16i32 VEC512:$s0), (v16i32 (bcst_vector_node(v2i32 eL : $s1))), 0x0, 0x8),
```
F-Stuckmann, line 866: can we generalize the rule to support all the vpush variants, i.e. add 8, 16, 32?

andcarminati: Good question @F-Stuckmann. We can for sure, I will investigate if this is important for the benchmarks.

andcarminati: As those patterns will not match now, I would prefer to keep this simple. We also have different shift intrinsics that can be addressed in a follow up work. (in case we see a demand.)

niwinanto (dismissed):
LGTM! Would be nice if you can create a ticket to track other variants to support.

andcarminati:
> LGTM! Would be nice if you can create a ticket to track other variants to support.

Created: AIECC-958

niwinanto (approved):

PR #418, opened 2025-04-01
[AIE2P] Fold adjacent 32b G_EXTRACT_VECTOR_ELT

andcarminati:
Depends on https://github.com/Xilinx/llvm-aie/pull/420 to avoid spills.


llvm/lib/Target/AIE/AIECombinerHelper.cpp
```diff
@@ -2896,3 +2897,93 @@ bool llvm::matchShuffleBcstToCopy(MachineInstr &MI, MachineRegisterInfo &MRI,
...
+  const LLT DstTy = MRI.getType(DstReg);
+  const LLT Src1Ty = MRI.getType(Src1Reg);
+  const LLT S64 = LLT::scalar(64);
+  const LLT S32 = LLT::scalar(32);
+
+  if (Src1Ty.getSizeInBits() < AIETII.getBasicVectorBitSize())
+    return false;
+
+  if (DstTy != S32)
+    return false;
```
martien-de-jong, line 2942: I think it would also work for 8,8 -> 16 and 16,16 -> 32. Then we only have to find an adjacent partner of the same type.

martien-de-jong: Eh, oh, not true. We need to define two separate registers that form a register pair.

llvm/lib/Target/AIE/AIECombinerHelper.cpp
```diff
@@ -2896,3 +2897,93 @@ bool llvm::matchShuffleBcstToCopy(MachineInstr &MI, MachineRegisterInfo &MRI,
...
+  auto Index = getIConstantVRegValWithLookThrough(Src2Reg, MRI);
+
+  if (!Index)
+    return false;
+
+  unsigned IndexCst = Index->Value.getZExtValue();
+
+  // Only even odd pairs.
+  if (IndexCst % 2)
+    return false;
```
martien-de-jong, line 2955: If they are in the wrong order, we will find the pair when combining the other, but perhaps we can also allow any lane here and find the one with the lowest bit flipped. The unmerge builder can use the index to decide the right order.

llvm/lib/Target/AIE/AIECombinerHelper.cpp
```diff
@@ -2896,3 +2897,93 @@ bool llvm::matchShuffleBcstToCopy(MachineInstr &MI, MachineRegisterInfo &MRI,
...
+
+  auto FindPairedExtract = [&](unsigned PairedIndex) -> MachineInstr * {
+    for (auto &Use : MRI.use_nodbg_instructions(Src1Reg)) {
+      if (Use.getOpcode() != TargetOpcode::G_EXTRACT_VECTOR_ELT)
+        continue;
+      if (Use.getParent() != MI.getParent())
+        continue;
+      if (!Helper.dominates(MI, Use))
+        continue;
```
martien-de-jong, line 2969: I don't think one use needs to dominate the other use. As it stands, this will prevent finding out-of-order pairs

llvm/lib/Target/AIE/AIECombinerHelper.cpp
```diff
@@ -2896,3 +2897,93 @@ bool llvm::matchShuffleBcstToCopy(MachineInstr &MI, MachineRegisterInfo &MRI,
...
+      const Register UseDstReg = Use.getOperand(0).getReg();
+      const Register UseSrc2Reg = Use.getOperand(2).getReg();
+      const LLT UseDstTy = MRI.getType(UseDstReg);
+      if (UseDstTy != S32)
+        continue;
+      auto UseIndex = getIConstantVRegValWithLookThrough(UseSrc2Reg, MRI);
```
martien-de-jong: I think we could share a lot of this in a common lambda recognizing an extract of the right type and a constant lane index, shared between the two extracts.

llvm/test/CodeGen/AIE/aie2p/GlobalIsel/prelegalizercombiner-paired-extracts.mir
```diff
@@ -0,0 +1,96 @@
+# NOTE: Assertions have been autogenerated by utils/update_mir_test_checks.py
+#
+# This file is licensed under the Apache License v2.0 with LLVM Exceptions.
+# See https://llvm.org/LICENSE.txt for license information.
+# SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+#
+# (c) Copyright 2025 Advanced Micro Devices, Inc. or its affiliates
+# RUN: llc -mtriple aie2p -run-pass=aie2p-prelegalizer-combiner --aie-s20-narrowing=true %s -verify-machineinstrs -o - | FileCheck %s
```
andcarminati: remove `--aie-s20-narrowing=true`

llvm/lib/Target/AIE/AIECombinerHelper.cpp
andcarminati: As discussed offline, this helps to simplify the code in relation to the insertion point. If we reach a situation that MI don't dominates UseMI, it means that UseMI was created by other combiner in the same run. If we catch this case now, we should either:

- Move all uses of UseMI down to after MI, or
- Change the builder insertion point to UseMI.

Dominance here will avoid such complication. If we have a case where MI !dom UseMI, in the next run we will catch as UseMI dom MI, so no opportunities will be left.

llvm/lib/Target/AIE/AIECombinerHelper.cpp
```diff
@@ -2896,3 +2897,101 @@ bool llvm::matchShuffleBcstToCopy(MachineInstr &MI, MachineRegisterInfo &MRI,
...
+
+  if (Src1Ty.getSizeInBits() < AIETII.getBasicVectorBitSize())
+    return false;
+
+  if (DstTy != S32)
+    return false;
+
+  auto GetConstantIndex =
+      [&](const MachineInstr &ExtractMI) -> std::optional<unsigned> {
+    const Register Src2Reg = ExtractMI.getOperand(2).getReg();
```
martien-de-jong: nit: I wouldn't mind IndexReg

llvm/lib/Target/AIE/AIECombinerHelper.cpp
```diff
@@ -2896,3 +2897,101 @@ bool llvm::matchShuffleBcstToCopy(MachineInstr &MI, MachineRegisterInfo &MRI,
...
+  MachineInstr *NextExtractMI = FindPairedExtract(NextIndex);
+  if (!NextExtractMI)
+    return false;
+
+  MatchInfo = [=, &MRI](MachineIRBuilder &B) {
+    const Register NewExtendedDstReg = MRI.createGenericVirtualRegister(S64);
+    const Register NewIdxReg = B.buildConstant(S32, *IndexCst / 2).getReg(0);
+
+    // Bitcast from <2n x s32> to <n x s64>
+    const LLT NewType = LLT::fixed_vector(Src1Ty.getNumElements() / 2, S64);
```
martien-de-jong, line 2993: Isn't there some nice LLT method, similar to .divides(2) ? Shall we add it? (.reshapeDivideElts(2))

martien-de-jong: CHECK: do we have type constraints that say that 32 bit extract can only be done on an n x s32 vector type? Otherwise this may change the vector size.

llvm/lib/Target/AIE/AIECombinerHelper.cpp
```diff
@@ -2896,3 +2897,101 @@ bool llvm::matchShuffleBcstToCopy(MachineInstr &MI, MachineRegisterInfo &MRI,
...
+    const Register CastedVecReg = B.buildBitcast(NewType, Src1Reg).getReg(0);
+
+    const Register NextDstReg = NextExtractMI->getOperand(0).getReg();
+    // As we need to replace 2 instructions, make the second one dead,
+    // so DCE will remove it. If we don't do this, we have a SSA violation.
+    const Register NewDeadReg = MRI.cloneVirtualRegister(NextDstReg);
+    NextExtractMI->getOperand(0).setReg(NewDeadReg);
```
martien-de-jong, line 3001: nit: the comment confused me. DCE has nothing to do with it, we just do this to prevent two definitions of the same SSA value.

llvm/lib/Target/AIE/AIECombinerHelper.cpp
```diff
@@ -2896,3 +2897,101 @@ bool llvm::matchShuffleBcstToCopy(MachineInstr &MI, MachineRegisterInfo &MRI,
...
+
+    B.buildExtractVectorElement(NewExtendedDstReg, CastedVecReg, NewIdxReg);
+    auto Split = B.buildUnmerge(S32, NewExtendedDstReg);
+    const unsigned FirstOpIndex = isEven ? 0 : 1;
+    B.buildCopy(DstReg, Split->getOperand(FirstOpIndex).getReg());
+    MRI.replaceRegWith(NextDstReg,
+                       Split->getOperand(FirstOpIndex ^ 1).getReg());
```
martien-de-jong: CHECK: the asymmetry here is only because we replace the first, and have changed the destination of the other.

llvm/lib/Target/AIE/AIECombine.td
```diff
@@ -145,7 +151,8 @@ def AIE2PPreLegalizerCombiner
                                                                  combine_vector_shuffle_broadcast,
                                                                  combine_single_diff_build_vector,
                                                                  combine_vector_shuffle_to_extract_insert_elt,
-                                                                 combine_vector_shuffle_concat_extracted_subvectors]> {
+                                                                 combine_vector_shuffle_concat_extracted_subvectors,
+                                                                 combine_paired_extracts]> {
```
niwinanto, line 155: Can we include this in post legalizer combiner too, looks like legalizer might introduce some opportunities too.

niwinanto: Ohh, no. We are leagalizing `G_EXTRACT_VECTOR_ELT` to `G_AIE_SEXT_EXTRACT_VECTOR_ELT` or other instructions.

llvm/lib/Target/AIE/AIECombinerHelper.cpp
andcarminati: I simplified the code with two copies.

llvm/lib/Target/AIE/AIECombinerHelper.cpp
andcarminati: The effect is twofold: prevent SSA breakage and make the instruction dead for removal.

llvm/lib/Target/AIE/AIECombine.td
andcarminati: Discussed offline.

martien-de-jong (approved):

PR #423, opened 2025-04-03
Report core postpipeliner results as optimization remarks.

martien-de-jong:
This enables a yaml file with key statistics like stage count and initiation interval
Additionally the stage count vs minitercount checks have been moved earlier


andcarminati (approved):
LGTM.

PR #425, opened 2025-04-07
[AIEX][AIE2P] Add target support for II pragma

andcarminati:
Now we can use #pragma clang loop pipeline_initiation_interval(number) to catch more cases in the pre-SWP. The target will honor the pragma as much as possible in any rejection decision.


llvm/lib/Target/AIE/AIEBasePipelinerLoopInfo.cpp
```diff
@@ -758,7 +762,7 @@ bool ZeroOverheadLoop::preferPostPipeliner(SMSchedule &SMS) {
 
   unsigned NS = SMS.getMaxStageCount() + 1;
   int II = SMS.getInitiationInterval();
-  if (NS > LoopMaxStageCount && II < PostPipelinerCutoff) {
+  if (!HasIIPragma && NS > LoopMaxStageCount && II < PostPipelinerCutoff) {
```
martien-de-jong, line 765: I would suggest to actually use the specified II, and only accept the prepipeliner schedule if it is <= that II.

andcarminati: Discussed offline: this check is done by the MachinePipeliner.

martien-de-jong (approved):
Perhaps we need to find a nice place to document it, also that we don't fail if we don't find it (as it is now)

PR #429, opened 2025-04-08
[DRAFT][AIEX] Improve pre-sched decisions for Post-SWP candidates

andcarminati:
This is a trial to get a more consistent behavior for the PreRA Scheduler. Currently, for example. Now, we will look only to Post-SWP potential candidates to skip the pre scheduler. The affected test also shows that we could prevent pre-scheduling of non-ZOL loops with the current approach.

### DIFF
--- a/llvm/lib/Target/AIE/Utils/AIELoopUtils.cpp
+++ b/llvm/lib/Target/AIE/Utils/AIELoopUtils.cpp
@@ -190,4 +190,51 @@ bool hasUnrollPragma(const Loop *L) {
   return false;
 }
 
+/// Check if this block is a post-SWP candidate.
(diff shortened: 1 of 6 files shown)


llvm/lib/Target/AIE/Utils/AIELoopUtils.cpp
```diff
@@ -190,4 +190,51 @@ bool hasUnrollPragma(const Loop *L) {
   return false;
 }
 
+/// Check if this block is a post-SWP candidate.
+bool isPostSWPCandidate(const AIEBaseInstrInfo &TII,
```
andcarminati, line 194: Some refactoring opportunities.

F-Stuckmann:
I like where this PR is heading. We just have to check out QoR numbers

PR #432, opened 2025-04-09
[AIEX] Disable memset/cpy/move loop idioms recognition

andcarminati:
Our backend can handle those loops in a better way.


llvm/lib/Analysis/TargetLibraryInfo.cpp
```diff
@@ -879,6 +882,13 @@ static void initialize(TargetLibraryInfoImpl &TLI, const Triple &T,
     TLI.setUnavailable(LibFunc_vec_free);
   }
 
+  // Memset/memcpy/memmove to vectors are forwarded to bytewise libcalls.
+  if (T.isAIE2P()) {
```
niwinanto: Disabling here is going to affect bunch of passes, not just loop idioms recognition. Is that acceptable?

llvm/lib/Analysis/TargetLibraryInfo.cpp
```diff
@@ -879,6 +882,13 @@ static void initialize(TargetLibraryInfoImpl &TLI, const Triple &T,
...
+    TLI.setUnavailable(LibFunc_memset);
```
niwinanto: Also, I am wondering, there is no test covering this case. Can we have a test? It can be a pre-commit, will be easy to see the change.

llvm/lib/Analysis/TargetLibraryInfo.cpp
andcarminati: I only found another one: MemCpyOptimizer. Maybe I am wrong, can you point to the other ones?

niwinanto: I am not sure though, `MemCpyOptimizer` is one of them. Also, `LibFunc_memset` has occurrence in SimplifyLibCalls, DeadStoreElimination, MemoryOpRemark etc.

andcarminati: Sure, but we need to look to who queries TLI, like `TLI->has(LibFunc_memset)`.

llvm/test/CodeGen/AIE/opt/memset_idiom.ll
```diff
@@ -0,0 +1,49 @@
...
+; CHECK-LABEL: define void @test_memset(
+; CHECK-SAME: ptr nocapture writeonly [[OFM:%.*]], i32 noundef [[N:%.*]]) {
+; CHECK-NEXT:  entry:
+; CHECK-NEXT:    [[CMP3:%.*]] = icmp sgt i32 [[N]], 0
+; CHECK-NEXT:    br i1 [[CMP3]], label [[FOR_BODY_PREHEADER:%.*]], label [[FOR_COND_CLEANUP:%.*]]
+; CHECK:       for.body.preheader:
+; CHECK-NEXT:    [[TMP0:%.*]] = trunc i32 [[N]] to i20
+; CHECK-NEXT:    [[TMP1:%.*]] = shl nuw i20 [[TMP0]], 6
+; CHECK-NEXT:    call void @llvm.memset.p0.i20(ptr align 64 [[OFM]], i8 0, i20 [[TMP1]], i1 false)
+; CHECK-NEXT:    br label [[FOR_BODY:%.*]]
```
martien-de-jong: I don't get this. We do the memset and jump to the original loop, without an adjustment of the count.

martien-de-jong: Oh, sorry, the loop is left empty.

llvm/lib/Analysis/TargetLibraryInfo.cpp
martien-de-jong: Add a TODO, that we need to address the cases where memset handling is worse than the original.

llvm/test/CodeGen/AIE/opt/memset_idiom.ll
andcarminati: Yeah, we are just running LoopIdiom pass, nothing eles will change.

llvm/lib/Analysis/TargetLibraryInfo.cpp
andcarminati: Done!

andcarminati:
Results:
Core_Compute_Insn_Count
(Note that for core compute, we are now accounting for the copies internally, and not as part of initialization - so it is not a regression.)
Core_Insn_Count
Core_PMSize
Core_StackSize

clang/lib/Driver/ToolChains/AIE.cpp
```diff
@@ -187,6 +187,16 @@ void AIEToolChain::addClangTargetOptions(
 
   // Enable Loop Iteration Count Assumptions
   CC1Args.append({"-mllvm", "-enable-loop-iter-count-assumptions=true"});
+
+  bool UseBuiltins = DriverArgs.hasFlag(options::OPT_fbuiltin,
```
andcarminati, line 191: Just in case:
```
  /// hasFlag - Given an option \p Pos and its negative form \p Neg, return
  /// true if the option is present, false if the negation is present, and
  /// \p Default if neither option is given. If both the option and its
  /// negation are present, the last one wins.
```

clang/lib/Driver/ToolChains/AIE.cpp
```diff
@@ -187,6 +187,16 @@ void AIEToolChain::addClangTargetOptions(
...
+                                        options::OPT_fno_builtin, false);
+
+  // Only disable memory builtins if -fbuiltin was not used.
```
martien-de-jong: nit : or -fno-builtin

clang/lib/Driver/ToolChains/AIE.cpp
```diff
@@ -187,6 +187,16 @@ void AIEToolChain::addClangTargetOptions(
...
+  if (!UseBuiltins) {
+    CC1Args.push_back("-fno-builtin-memset");
```
martien-de-jong, line 199: Add a TODO describing the bad case, so that we can revisit later. Perhaps an end-to-end test?

andcarminati: Added an IR test. I think end-to-end here can be a bit sensible here due to possible backend changes.

martien-de-jong (approved):

PR #435, opened 2025-04-14
[AIEX] Make the decision to accept supplied II stronger

andcarminati:


llvm/lib/Target/AIE/AIEBasePipelinerLoopInfo.cpp
```diff
@@ -756,13 +756,13 @@ bool ZeroOverheadLoop::preferPostPipeliner(SMSchedule &SMS) {
   // job on multi-stage live-ranges without spilling or moving.
 
   // PostPipeliner can do nothing without tripcount > 1
-  if (MinTripCount <= 1) {
+  if (MinTripCount <= 1 || HasIIPragma) {
```
martien-de-jong, line 759: so specifying an II now is equivalent to disable prepipeliner?

andcarminati: Is the other way around: if we specify an II we don't prefer the postpipeliner. I changed because we were still preferring the postpipeliner for a user specified II when `SDep.getSignedLatency() >= II` (see last condition).

martien-de-jong: Sorry, I thought this was the canAccept() method.

martien-de-jong (approved):

PR #436, opened 2025-04-14
[AIE] Enable floatdidf builtin used by G_SITOFP

jseaton:
Hi, for `G_SITOFP` instructions you seem to now emit calls to `__floatdidf` for `int64_t` -> `double`, however this builtin is disabled.
This causes a linker error when compiling some code that contains an `int64_t` -> `double` cast. I stumbled across this while trying to reuse Arm's `cosf` implementation.
Checked vs `floatdidf_test.c` values on AIE2 hardware.

### DIFF
--- a/compiler-rt/lib/builtins/CMakeLists.txt
+++ b/compiler-rt/lib/builtins/CMakeLists.txt
@@ -423,7 +423,6 @@ list(REMOVE_ITEM aie2_GENERIC_SOURCES
   fixdfdi.c     # unsupported instruction : G_FCMP
   fixunsdfdi.c  # unsupported instruction : G_FPTOUI
   fixunssfdi.c  # unsupported instruction : G_FPTOUI
-  floatdidf.c   # unsupported instruction : G_SITOFP
   muldc3.c      # unsupported instruction : G_FCOPYSIGN
   mulsc3.c      # unsupported instruction : G_FCOPYSIGN
   udivmoddi4.c  # unsupported instruction : G_CTTZ_ZERO_UNDEF


compiler-rt/lib/builtins/CMakeLists.txt
```diff
@@ -423,7 +423,6 @@ list(REMOVE_ITEM aie2_GENERIC_SOURCES
   fixdfdi.c     # unsupported instruction : G_FCMP
   fixunsdfdi.c  # unsupported instruction : G_FPTOUI
   fixunssfdi.c  # unsupported instruction : G_FPTOUI
-  floatdidf.c   # unsupported instruction : G_SITOFP
```
martien-de-jong, line 426: Sounds like a free lunch to me. Thanks for checking this out.

martien-de-jong: If you performed the test manually, is there a possibility to add it to CI?

martien-de-jong (approved):

andcarminati:
I think we could also reabase the commit on top of an aie-public-derived branch (without the second merge commit). It will help to merge this change.

compiler-rt/lib/builtins/CMakeLists.txt
jseaton: This was on actual hardware - I'm happy to add one if you have on-device tests somewhere!

martien-de-jong: Ah, I see what you mean. Forget I asked.

PR #438, opened 2025-04-15
[AIE2P] remove redundant ptr_add chains

F-Stuckmann:
Address Chaining may insert incremental G_PTR_ADD, to reduce the immediate range in the G_PTR_ADD.
Since Post Increments replace the G_PTR_ADD, they delete their source G_PTR_ADD. 
However, Offset Load/Stores cannot delete their source G_PTR_ADD, since they do not update the Pointer. So as a result a now-unnecessary G_PTR_ADD chain may remain, see below.
`%1 = G_PTR_ADD %base, G_CONSTANT imm1`
`%2 = G_PTR_ADD %1, G_CONSTANT imm2`
` -->`
` %2 = G_PTR_ADD %base, G_CONSTANT (imm1 + imm2)`

Solution, add existing LLVM combiner **ptr_add_immed_chain** .

**Instr Count Improvement**: -0.04%
- with a single regression in non-optimized GEMM, which is caused by a suboptimal register allocation
- The optimized GEMM variants all improve performance by -0.18% up to -0.82%

**Programm Memory** : -0.25% 
- no regressions

**Stack Size**: -0.02% 
- no regressions


andcarminati:
This makes a lot of sense @F-Stuckmann, great!

niwinanto:
Could you please include a test as well to showcase the change?

llvm/lib/Target/AIE/AIECombine.td
```diff
@@ -249,6 +249,7 @@ def AIE2PostLegalizerCustomCombiner
 
 def AIE2PPostLegalizerCustomCombiner
     : GICombiner<"AIE2PPostLegalizerCustomCombinerImpl", [ combine_load_store_increment,
+                                                           ptr_add_immed_chain,
```
martien-de-jong, line 252: I think it would make sense also to add this to AIE2?

F-Stuckmann: AIE2 already added the combiner. It was only missing in AIE2p

martien-de-jong (approved):

PR #440, opened 2025-04-18
[AIE] Improve Zero-overhead loop 112-bytes padding

abnikant:
Avoid inserting extra nop bundles if there are enough bundles between ZOLSetup instructions and LoopEnd to create 112-bytes padding. It takes into account the bundles in the loop body. 
1) Added heuristic to check for available size between ZOLSetup last bundle and (LoopEnd - 1) - for the cases where number of loop bundles are >= LoopSetupDistance. 
In plan: More heuristics to be added 1) Avoid elongating if there are 112-bytes available between writing of (ls, le, and lc) and (loop end - 1), for the case where number of loop bundles < LoopSetupDistance. The EgdeLatency needs to be updated. 2) Elongate the bundles after deducting the available size.

QoR report : 
Core_PMSize
Core_Insn_Count


llvm/lib/Target/AIE/AIEBaseInstrInfo.h
```diff
@@ -435,6 +439,10 @@ struct AIEBaseInstrInfo : public TargetInstrInfo {
   // setting LE Reg
   bool isZOLSetupBundle(MachineBasicBlock::iterator MII) const;
   bool isLastZOLSetupBundleInMBB(MachineBasicBlock::iterator MII) const;
+  virtual const AIE::Bundle<MachineInstr>
```
martien-de-jong, line 448: Check: we return by value, so I guess we have the implementation of AIE::Bundle available here. Can we drop the forward declaration?

llvm/lib/Target/AIE/AIEMachineAlignment.cpp
```diff
@@ -226,7 +195,9 @@ bool AIEMachineAlignment::runOnMachineFunction(MachineFunction &MF) {
   for (auto &MBB : MF) {
     std::vector<llvm::iterator_range<MachineBasicBlock::iterator>> Regions =
         findRegions(MBB);
-    applyBundlesAlignment(Regions);
+    auto *TII = static_cast<const AIEBaseInstrInfo *>(
+        MBB.getParent()->getSubtarget().getInstrInfo());
```
martien-de-jong: This can be hoisted from the loop. MBB.getParent() is referring to MF. I wouldn't mind keeping it in a data member of  AIEMachineAlignment (provided it is reset to nullptr afterwards) so that it doesn't need to be passed.

llvm/lib/Target/AIE/AIEBaseInstrInfo.cpp
```diff
@@ -1079,20 +1079,57 @@ const PacketFormats &AIEBaseInstrInfo::getPacketFormats() const {
...
+AIEBaseInstrInfo::getZOLBundlesCount(const MachineBasicBlock &MBB) const {
+  auto First = MBB.getFirstNonDebugInstr();
+  auto Last = MBB.getLastNonDebugInstr();
+
+  // If MBB is empty or has no non-debug instructions, return 0.
+  if (First == MBB.end() || Last == MBB.end())
+    return 0;
+
+  if (isHardwareLoopEnd(Last->getOpcode()))
+    return std::distance(First, Last);
```
martien-de-jong: Doesn't this count debug instructions between First and Last as bundles?

llvm/lib/Target/AIE/AIEBaseInstrInfo.cpp
```diff
@@ -397,6 +397,39 @@ bool AIEBaseInstrInfo::isLastZOLSetupBundleInMBB(
   return true;
 }
 
+unsigned AIEBaseInstrInfo::getRegionSize(
```
andcarminati: nit: maybe `getRegionSizeInBytes`.

llvm/lib/Target/AIE/AIEBaseInstrInfo.cpp
```diff
@@ -397,6 +397,39 @@ bool AIEBaseInstrInfo::isLastZOLSetupBundleInMBB(
...
+    llvm::iterator_range<MachineBasicBlock::iterator> Region) const {
+  unsigned Size = 0;
+  LLVM_DEBUG(dbgs() << "---Region Begin---\n");
+  for (auto it = Region.begin(), end = Region.end(); it != end; ++it) {
```
andcarminati: nit: `It` and `End`

llvm/lib/Target/AIE/AIEBaseInstrInfo.cpp
```diff
@@ -1079,20 +1079,57 @@ const PacketFormats &AIEBaseInstrInfo::getPacketFormats() const {
...
   unsigned LoopSetupDistance = 0;
+  unsigned ZOLBundlesCount = 0;
+  unsigned ZOLBodyRegionsCount = 0;
   bool IsCall = false;
   auto ZOLSupport = getZOLSupport();
+  bool IsZOLBody = isZOLBody(MBB);
+  if (IsZOLBody) {
+    // Exclude the LoopEnd bundle since it must reside in its own
+    // standalone region to ensure it points to a 128-bit aligned instruction.
+    ZOLBundlesCount = getZOLBundlesCount(MBB) - 1;
```
martien-de-jong, line 1177: I think we can't guarantee that all bundles in the loop are 16 bytes. Hence, we need `ZOL->LoopSetupDistance - BodySize / 16` padding bundles.

abnikant: @martien-de-jong can you please review the updated commits? Algorithm does not expect the loop bundle size to be 16 bytes. It is counting  the number of bundles in the loop and subtract 1. If this count is < 7, then the DAG mutator create an EdgeLatency of 8 - count, and it elongate each of these bundles to 16-bytes. For the case where number the count (number of loop bundles -1) is >= 7 and then 7 of these loop bundles will be elongated provided the number of bytes between LastZOLSetup instruction and LoopEnd -1 is < 112 bytes (for more than 112-bytes, we are done)

llvm/lib/Target/AIE/AIEBaseInstrInfo.cpp
andcarminati: I would prefer to not trust `distance` here.

llvm/lib/Target/AIE/AIEBaseInstrInfo.cpp
```diff
@@ -1079,20 +1079,57 @@ const PacketFormats &AIEBaseInstrInfo::getPacketFormats() const {
   return FormatInterface->getPacketFormats();
 }
 
+unsigned
+AIEBaseInstrInfo::getZOLBundlesCount(const MachineBasicBlock &MBB) const {
+  auto First = MBB.getFirstNonDebugInstr();
+  auto Last = MBB.getLastNonDebugInstr();
+
+  // If MBB is empty or has no non-debug instructions, return 0.
```
andcarminati: can we reuse `isZOLBody` here in some way?

abnikant:
[Core_Insn_Count.csv](https://github.com/user-attachments/files/19846331/Core_Insn_Count.csv)
[Core_PMSize.csv](https://github.com/user-attachments/files/19846332/Core_PMSize.csv)
[Core_StackSize.csv](https://github.com/user-attachments/files/19846333/Core_StackSize.csv)
 QoR data w.r.t. to original run with Peano.

llvm/lib/Target/AIE/AIEBaseInstrInfo.cpp
abnikant: yes, updating this code.

llvm/lib/Target/AIE/AIEBaseInstrInfo.cpp
```diff
@@ -1112,14 +1149,26 @@ AIEBaseInstrInfo::getAlignmentBoundaries(MachineBasicBlock &MBB) const {
         LoopSetupDistance--;
       }
 
+      if (IsZOLBody && ZOLBodyRegionsCount > 0) {
+        AlgnCandidates.emplace_back(MI);
+        ZOLBodyRegionsCount--;
+      }
```
martien-de-jong: Oh, you're actually padding the bundles in the loop. I had thought you would just subtract the size of the body in units of full bundles from LoopSetupDistance, putting all padding in the preheader. That would make the algorithm logic much simpler.  Am I missing something?

abnikant: yes, I am padding the loop bundles this way we can avoid inserting extra nop bundles in most of the cases. Consider the case where preheader size after writing to the ZOL setup instruction is 0, we will end of inserting some nop bundles if the loop size is not sufficient.

llvm/test/CodeGen/AIE/aie2/elongate/zol_112bytes_elongate2.mir
```diff
@@ -36,66 +37,45 @@ body:             |
...
+  ; CHECK-NEXT:     NOP
   ; CHECK-NEXT:   }
   ; CHECK-NEXT:   BUNDLE {
-  ; CHECK-NEXT:     NOPB
-  ; CHECK-NEXT:     NOPA
-  ; CHECK-NEXT:     NOPS
-  ; CHECK-NEXT:     NOPXM
-  ; CHECK-NEXT:     NOPV
+  ; CHECK-NEXT:     NOP
   ; CHECK-NEXT:   }
```
martien-de-jong, line 59: We still have 7 bundles (cycles) here. I would have expected less. 
Never mind -- they are in the input.

llvm/test/CodeGen/AIE/aie2/schedule/postpipeliner/gemm-1.mir
```diff
@@ -35,28 +35,28 @@
...
-  ; CHECK-NEXT:    vldb wh7, [p1], m6; vmac.f bmh6, bmh6, x0, x5, r3
-  ; CHECK-NEXT:    vldb wl8, [p0, #0]; vmac.f bmh7, bmh7, x6, x5, r3
-  ; CHECK-NEXT:    vldb wh8, [p0, #32]; vmac.f bmh8, bmh8, x1, x7, r3
+  ; CHECK-NEXT:    vldb wh9, [p1], m6; nopa ; nops ; nopxm ; vmac.f bmh2, bmh2, x0, x11, r3
+  ; CHECK-NEXT:    vldb wl5, [p1], m5; nopa ; nops ; nopx ; vshuffle x3, x3, x3, r2; vmac.f bmh3, bmh3, x6, x11, r3
+  ; CHECK-NEXT:    vldb wh5, [p1], m6; nopa ; nops ; nopxm ; vmac.f bmh4, bmh4, x1, x5, r3
+  ; CHECK-NEXT:    vldb wl7, [p1], m5; nopa ; nops ; nopxm ; vmac.f bmh5, bmh5, x8, x5, r3
+  ; CHECK-NEXT:    vldb wh7, [p1], m6; nopa ; nops ; nopxm ; vmac.f bmh6, bmh6, x0, x5, r3
+  ; CHECK-NEXT:    vldb wl8, [p0, #0]; nopa ; nops ; nopxm ; vmac.f bmh7, bmh7, x6, x5, r3
+  ; CHECK-NEXT:    nopa ; vldb wh8, [p0, #32]; nopx ; vmac.f bmh8, bmh8, x1, x7, r3
```
martien-de-jong: I don't think we need these nops. The body already had more than 112 bytes.

llvm/test/CodeGen/AIE/aie2/schedule/postpipeliner/add-store.mir
```diff
@@ -12,50 +12,6 @@
 
 --- |
   define dso_local void @addStore(ptr addrspace(5) noalias nocapture writeonly %d, i32 noundef %n) local_unnamed_addr #0 {
-  ; CHECK-LABEL: addStore:
```
andcarminati: What happened here?

andcarminati: nit `update_llc_test..`

llvm/lib/Target/AIE/AIEBaseInstrInfo.cpp
martien-de-jong: Fair point.  But I'm still worried about the nops we insert in the gemm example, where the body was large enough.

llvm/test/CodeGen/AIE/aie2/schedule/postpipeliner/gemm-1.mir
abnikant: @martien-de-jong , looks like you are looking into outdated commit? I don't see any nops inserted in gemm-1.mir

llvm/lib/Target/AIE/AIEBaseInstrInfo.cpp
```diff
@@ -1046,20 +1079,90 @@ const PacketFormats &AIEBaseInstrInfo::getPacketFormats() const {
...
+    ZOLBundlesCount = getZOLBundlesCount(MBB) - 1;
+    if (ZOLBundlesCount < ZOLSupport->LoopSetupDistance)
+      ZOLBodyRegionsCount = ZOLBundlesCount;
+    else {
+      // Elongate the ZOL loop body only if the distance from the end of the
+      // ZOL setup instruction to the last bundle in the loop (excluding the
+      // final bundle) is less than 112 bytes.
+      const unsigned LoopSize = LoopSizeExcludingLastBundle(MBB);
+      const bool DistanceConstraintMet =
+          (LoopSize + PostZOLRegionSize) >= 16 * ZOLSupport->LoopSetupDistance;
```
martien-de-jong: I would like to compute PostZOLRegionSize here by traversing the preheader, possibly in a separate function/method. That would avoid the state variable living in the alignment pass, which looks slightly out of place there.

abnikant: okay, I wanted to avoid traversing the MBB again, to save some runtime.

llvm/lib/Target/AIE/AIEBaseInstrInfo.cpp
```diff
@@ -1079,14 +1182,32 @@ AIEBaseInstrInfo::getAlignmentBoundaries(MachineBasicBlock &MBB) const {
         LoopSetupDistance--;
       }
 
+      if (IsZOLBody && ZOLBodyRegionsCount > 0) {
+        AlgnCandidates.emplace_back(MI);
+        ZOLBodyRegionsCount--;
+      }
+
```
martien-de-jong: Still, I think LoopSetupDistance and ZOLBodyRegionsCount are the same thing. We either have a preheader or a body of a ZOL, not both,  and we have a number of bundles that should be expanded to 16 bytes. The body one computes that amount of bundles before hand, the preheader starts it at the last loop setup instruction.

llvm/lib/Target/AIE/AIEMachineAlignment.cpp
```diff
@@ -198,14 +167,14 @@ void AIEMachineAlignment::applyBundlesAlignment(
...
-findRegions(MachineBasicBlock &MBB) {
+std::vector<llvm::iterator_range<MachineBasicBlock::iterator>>
+AIEMachineAlignment::findRegions(MachineBasicBlock &MBB) {
   auto *TII = static_cast<const AIEBaseInstrInfo *>(
       MBB.getParent()->getSubtarget().getInstrInfo());
   MachineBasicBlock::iterator RegionBegin = MBB.begin();
   std::vector<llvm::iterator_range<MachineBasicBlock::iterator>> Regions;
   std::vector<MachineBasicBlock::iterator> AlgnCandidate =
-      TII->getAlignmentBoundaries(MBB);
+      TII->getAlignmentBoundaries(MBB, PostZOLRegionSize);
```
niwinanto: What is the use of  `PostZOLRegionSize` passing from here?

llvm/lib/Target/AIE/AIEBaseSubtarget.cpp
```diff
@@ -296,9 +298,18 @@ class RegionEndEdges : public ScheduleDAGMutation {
...
-        EdgeLatency = std::max(EdgeLatency, ZOLSupport->LoopSetupDistance + 1);
+        if (PrologueMBB && PrologueMBB->succ_size() == 1) {
+          // if we have only one MBB, it must be the loop.
+          MachineBasicBlock *LoopSucc = *PrologueMBB->successors().begin();
+          // Exclude the LoopEnd bundle since it must reside in its own
+          // standalone region to ensure it points to a 128-bit aligned
+          // instruction.
+          ZOLBundlesCount = TII->getZOLBundlesCount(*LoopSucc) - 1;
+        }
+        if (ZOLBundlesCount < ZOLSupport->LoopSetupDistance)
```
niwinanto, line 309: May be a dump question, for `ZOLBundlesCount >= ZOLSupport->LoopSetupDistance` can we use 0?

llvm/lib/Target/AIE/AIEBaseInstrInfo.cpp
martien-de-jong: We can probably regain some time by providing a pruning limit to some of these searches. For instance, if we have established that the loopsize is at least 112 bytes, we don't need to know the true size.

llvm/lib/Target/AIE/AIEBaseInstrInfo.cpp
abnikant: updated

llvm/lib/Target/AIE/AIEBaseInstrInfo.cpp
abnikant: done

llvm/lib/Target/AIE/AIEMachineAlignment.cpp
abnikant: done

llvm/lib/Target/AIE/AIEMachineAlignment.cpp
abnikant: It was used to store Region size in bytes after last ZOL setup bundle. Now it is removed and it is computed when it is needed.

llvm/lib/Target/AIE/AIEBaseInstrInfo.cpp
```diff
@@ -397,13 +398,73 @@ bool AIEBaseInstrInfo::isLastZOLSetupBundleInMBB(
   return true;
 }
 
-unsigned AIEBaseInstrInfo::getRegionSize(
+unsigned AIEBaseInstrInfo::getPostZOLRegionSize(MachineBasicBlock &MBB) const {
+  auto Last = MBB.getLastNonDebugInstr();
+  if (Last == MBB.end())
+    return 0;
+
+  for (auto MI = MBB.begin(), End = MBB.end(); MI != End; ++MI) {
```
niwinanto: Nit: `for (const auto &MI: MBB)`

llvm/lib/Target/AIE/AIEBaseInstrInfo.cpp
```diff
@@ -397,13 +398,73 @@ bool AIEBaseInstrInfo::isLastZOLSetupBundleInMBB(
...
+    if (MI->isBundle() && isZOLSetupBundle(MI) &&
```
niwinanto: What is the difference between `isZOLSetupBundle` and `isLastZOLSetupBundleInMBB`?

llvm/lib/Target/AIE/AIEBaseInstrInfo.cpp
```diff
@@ -1082,17 +1143,62 @@ const PacketFormats &AIEBaseInstrInfo::getPacketFormats() const {
...
+  unsigned DelaySlot = 0;
   // LoopSetupDistance will be set to number of instructions (7). In
   // PostRAScheduler, this is enforced by setting the exit latency in the
-  // schduler dag mutator
+  // schduler dag mutator.
   unsigned LoopSetupDistance = 0;
+  unsigned ZOLBundlesCount = 0;
   bool IsCall = false;
   auto ZOLSupport = getZOLSupport();
+  assert(ZOLSupport);
```
niwinanto: Is `getAlignmentBoundaries` used just with ZOL?

llvm/lib/Target/AIE/AIEBaseInstrInfo.cpp
```diff
@@ -1082,17 +1143,62 @@ const PacketFormats &AIEBaseInstrInfo::getPacketFormats() const {
...
+      }
+      return 0;
+    };
+    const unsigned LoopSetupDist = ZOLSupport->LoopSetupDistance;
+    // Exclude the LoopEnd bundle as it must be placed in its own standalone
+    // region to guarantee 128-bit instruction alignment. Additionally, there
+    // must be a 112-byte gap (in PM address space) between writing to the ls,
+    // le, and lc registers and the LoopEnd instruction.
+    ZOLBundlesCount = getZOLBundlesCount(MBB) - 1;
+    if (ZOLBundlesCount < LoopSetupDist)
```
niwinanto: `LoopSetupDist` and `LoopSetupDistance` are confusing. Could you please rename one?

llvm/lib/Target/AIE/AIEBaseInstrInfo.cpp
martien-de-jong: Good catch! Perhaps move the assert with the `if (IsZOLBody)` below.

abnikant: It is true that this is not used only in case of ZOL, but why this should assert even if it has no ZOL? it is a query and just returns pre-populated struct object value.

llvm/lib/Target/AIE/AIEBaseInstrInfo.cpp
martien-de-jong: There can be more ZOLSetUpBundles in a block, but only one is the last.

martien-de-jong: I guess we can streamline the code by running the loop in reverse and accumulating the region size for each bundle that we see. Then we can return the region size as soon as we hit a ZOLSetUpBundle. llvm::reverse(MBB) gives you the reverse range.

llvm/lib/Target/AIE/AIEBaseInstrInfo.cpp
```diff
@@ -397,6 +398,99 @@ bool AIEBaseInstrInfo::isLastZOLSetupBundleInMBB(
...
+      }
+      break;
+    }
+  }
+  return 0;
+}
+
+unsigned AIEBaseInstrInfo::getPostZOLRegionSizeFromZOLBody(
+    MachineBasicBlock &LoopMBB) const {
+  for (auto *Pred : LoopMBB.predecessors()) {
```
andcarminati: Should we use `AIELoopUtils::getLoopPredecessor` here?

llvm/lib/Target/AIE/AIEBaseInstrInfo.cpp
```diff
@@ -397,6 +398,99 @@ bool AIEBaseInstrInfo::isLastZOLSetupBundleInMBB(
...
+        auto Start = std::next(MI);
+        return getRegionSizeInBytes(llvm::make_range(Start, MBB.end()));
+      }
+      break;
+    }
+  }
+  return 0;
+}
+
+unsigned AIEBaseInstrInfo::getPostZOLRegionSizeFromZOLBody(
```
andcarminati: I feel that this `Post` is `Pre`, or even `PreHeader`.

llvm/lib/Target/AIE/AIEBaseInstrInfo.cpp
abnikant: updated the name.

llvm/lib/Target/AIE/AIEBaseInstrInfo.cpp
abnikant: yes, updated the code to use loop in reverse, please check.

llvm/lib/Target/AIE/AIEBaseInstrInfo.cpp
abnikant: This is computing the number of bytes after the last ZOL setup instruction in the preheader, I guess name captures that, what is the name you suggest ?

andcarminati: maybe `getPostZOLSetupRegionSize` + this comment.

llvm/lib/Target/AIE/AIEBaseInstrInfo.cpp
```diff
@@ -397,6 +398,99 @@ bool AIEBaseInstrInfo::isLastZOLSetupBundleInMBB(
...
+
+  return isHardwareLoopEnd(Last->getOpcode());
+}
+
+unsigned
+AIEBaseInstrInfo::getZOLBundlesCount(const MachineBasicBlock &MBB) const {
+  if (!isZOLBody(MBB))
+    return 0;
+
+  auto First = MBB.getFirstNonDebugInstr();
```
andcarminati, line 438: Is this check already done in `isZOLBody`? Can we simplify?

llvm/lib/Target/AIE/AIEBaseInstrInfo.cpp
abnikant: I thought this is used only when the predecessor is a loop or isSingleMBBLoop. This function is called from ZOL Body MBB and we want to get the pre-header MBB.

llvm/lib/Target/AIE/AIEBaseInstrInfo.cpp
andcarminati: Something like:
```
unsigned
AIEBaseInstrInfo::getZOLBundlesCount(const MachineBasicBlock &MBB) const {
  if (!isZOLBody(MBB))
    return 0;

  return count_if(MBB, [](const MachineInstr &MI) {
    return !MI.isDebugInstr();
  });
}
```

abnikant: updated.

llvm/lib/Target/AIE/AIEBaseInstrInfo.cpp
abnikant: done, also moved into a lamda function.

martien-de-jong (approved):

andcarminati (approved):
LGTM. Great work @abnikant !

PR #442, opened 2025-04-22
[WIP] Enhance 3D register allocation strategy 

krishnamtibrewala:
I have been investigating Conv2D_bfp16_0 and why it performs so poorly. The issue is that we spill 3D addressing registers frequently.

Upon further investigation of how we have implemented its staged register allocation and how the greedy register allocator works with registers (including sub-registers), I believe the problem lies in how LLVM treats and assumes the impact of modifications in a sub-register on the super-register. This assumption does not hold for AIE 2D/3D registers.
When we split a 2D/3D register, say "Z" (which was used by PADD_3D), we create two new virtual registers, "X" and "Y," which are treated as the same register class as "Z." Instead of "Z," "Y" is used with PADD_3D. The key point is that "X," even though it is considered a "3D" register (i.e., $d1_3d), does not need to follow the constraint of assigning sub-registers to (m1 dn1 dj1 dc1 dn5 dj5 dc5). But we end up forcing this constrain on reg-alloc.

We started by splitting the live ranges of one 3D register "Z" and ended with two 3D registers in the same "3D" register class. This has a snowball effect, as we continually split to achieve the "3D" register constraints for every new virtual register greedy-alloc ends up creating. In Conv2D_bfp16_0, we start with five 3D virtual registers and end up creating 32. Furthermore, due to the handling mechanism of sub-registers, we cannot use getLargestLegalSuperClass(...) to avoid spilling and use other scalar registers. (Note: If spilling has already occurred—common in our staged allocation when allocating 3D/2D registers—the aie-superreg-rewrite pass cannot undo its effects. It only helps in the next pass of the greedy register allocator.)

Ideally, what we could do is:
1. If we split "Z" into "X" and "Y" virtual registers, focus only on allocating "Y" (used by PADD_3D after splitting) during the first pass of staged register allocation where we only allocate 3D registers.
2. Retype the unassigned virtual register "X," which is a "3D" register only used in COPY, into the individually legal largest superclass.
3. In the final pass of our staged register allocation, we will automatically have the flexibility to move these registers to other scalar registers. If necessary, they can be further split into smaller live ranges and moved to other 20-bit registers.

The potential benefits of this approach include reduced spilling and increasing the greedy allocator's flexibility to manage 2D/3D sub-registers.

To implement this with minimal modifications to the target-independent code:

1. To focus on only the virtual registers used by 2D/3D instructions, we could extend RegAllocBase::enqueue(const LiveInterval *LI). After splitting and creating new virtual registers, RegAllocBase::enqueue(...) is called on them. Currently, we use ShouldAllocateClass(*TRI, RC), which allows us to focus only on specific reg-class like "onlyAllocate3DRegisters." We could establish a similar mechanism to check if the live range "LI" is used by a 2D/3D instruction and focus only on them.
2. To retype unassigned 3D registers that were not directly used by a 2D/3D instruction, we could extend our aie-superreg-rewrite. This pass currently works only on virtual registers with an assigned physical register. We would need to extend this to include virtual registers that are 2D/3D, not assigned, and not used by any 2D/3D instructions (this should be fairly easy).
3. There is no need for special handling to allocate unassigned sub-reg of 2D/3D registers (now as full registers) since the final pass of the GreedyRegisterAllocator will handle this.


krishnamtibrewala:
TODO : Verify if there is any negative impact of following change

Implement TRI.getSubClassWithSubReg to avoid generation of 
```
15704B	  undef %1266.sub_lo_dim:eds = COPY %1267.sub_lo_dim:eds {
	    internal %1266.sub_hi_dim_then_sub_dim_count:eds = COPY %1267.sub_hi_dim_then_sub_dim_count:eds
	    internal %1266.sub_hi_dim_then_sub_dim_size:eds = COPY %1267.sub_hi_dim_then_sub_dim_size:eds
	    internal %1266.sub_hi_dim_then_sub_dim_stride:eds = COPY %1267.sub_hi_dim_then_sub_dim_stride:eds
```

Rather force it to generate 7 copies. Basically for all the individual sub-regs.
` undef %1266.sub_lo_dim:eds = COPY %1267.sub_lo_dim:eds` Should get split into 4 individual reg copy

This will help super-reg-rewrite to split them into new virtual reg with there own live-ranges

krishnamtibrewala:
TODO :
1. Extend the work to 2D reg
2. Understand the scope of extending this 1024 bit FIFO register. CHECK : if its even possible.

llvm/lib/CodeGen/RegAllocGreedy.h
```diff
@@ -284,7 +284,8 @@ class LLVM_LIBRARY_VISIBILITY RAGreedy : public MachineFunctionPass,
   bool ReverseLocalAssignment = false;
 
 public:
-  RAGreedy(const RegClassFilterFunc F = allocateAllRegClasses);
+  RAGreedy(const RegClassFilterFunc F = allocateAllRegClasses,
+           const LiveIntervalFilterFunc LIF = allocateAllLiveIntervals);
```
andcarminati, line 288: Nice that this is o NFC if we don't provide a function!

llvm/lib/Target/AIE/aie2p/AIE2PTargetMachine.cpp
```diff
@@ -66,6 +67,173 @@ void AIE2PPassConfig::addPreRegBankSelect() {
   }
 }
 
+static bool onlyAllocateLIwith3DInstruction(MachineRegisterInfo &MRI,
```
andcarminati, line 70: nit: `onlyAllocateLIWith3DInstruction`

andcarminati: We have access to TII, so we could move this switch to that class in a AIE-specific hook. We could also use a nice name, something like `isHighPriorityLIUseInstruction` (not sure about this name).

llvm/lib/Target/AIE/aie2p/AIE2PTargetMachine.cpp
```diff
@@ -66,6 +67,173 @@ void AIE2PPassConfig::addPreRegBankSelect() {
...
+                                            const TargetInstrInfo &TII,
+                                            const LiveInterval *LI) {
+  const Register Reg = LI->reg();
+  return std::any_of(
+      MRI.use_nodbg_instructions(Reg).begin(),
+      MRI.use_nodbg_instructions(Reg).end(), [&](const MachineInstr &MI) {
+        switch (MI.getOpcode()) {
```
andcarminati, line 77: For the future: we could also encode this using TableGen.

llvm/test/CodeGen/AIE/staged-ra-rewrite.mir
```diff
@@ -71,13 +71,13 @@ body:             |
   ; AIE2P-VREGS-NEXT:   [[LDA_dms_lda_idx_imm1:%[0-9]+]]:er = LDA_dms_lda_idx_imm [[COPY1]], 4
   ; AIE2P-VREGS-NEXT:   [[LDA_dms_lda_idx_imm2:%[0-9]+]]:er = LDA_dms_lda_idx_imm [[COPY1]], 8
   ; AIE2P-VREGS-NEXT:   [[LDA_dms_lda_idx_imm3:%[0-9]+]]:er = LDA_dms_lda_idx_imm [[COPY1]], 12
-  ; AIE2P-VREGS-NEXT:   [[COPY2:%[0-9]+]]:edn = COPY [[LDA_dms_lda_idx_imm1]]
-  ; AIE2P-VREGS-NEXT:   [[COPY3:%[0-9]+]]:em = COPY [[LDA_dms_lda_idx_imm]]
-  ; AIE2P-VREGS-NEXT:   [[COPY4:%[0-9]+]]:edj = COPY [[LDA_dms_lda_idx_imm2]]
-  ; AIE2P-VREGS-NEXT:   [[COPY5:%[0-9]+]]:edc = COPY [[LDA_dms_lda_idx_imm3]]
+  ; AIE2P-VREGS-NEXT:   [[COPY2:%[0-9]+]]:edn_as_32bit = COPY [[LDA_dms_lda_idx_imm1]]
```
andcarminati, line 74: Are we mixing AIE2 and AIE2P here? I mean, there is nothing related to your implementation, but I see no `edn_as_32bit` for AIE2P. I see `spill_eDN_to_eR` instead.

andcarminati: Oh, maybe this test is failing.....

llvm/lib/Target/AIE/AIESuperRegRewriter.cpp
```diff
@@ -172,17 +180,32 @@ bool AIESuperRegRewriter::runOnMachineFunction(MachineFunction &MF) {
...
 
-  // Re-write all the collected VRegs
+  // Re-write all the collected assigned VRegs
   for (auto &[VReg, PhysReg] : AssignedPhysRegs) {
-    rewriteSuperReg(VReg, PhysReg, MRI, TRI, VRM, LRM, LIS, Indexes, DebugVars);
+    rewriteSuperReg(VReg, PhysReg, MF, MRI, TRI, VRM, LRM, LIS, Indexes,
+                    DebugVars);
+  }
+
+  // Re-write all the collected unassigned VRegs
```
andcarminati, line 205: Maybe include a description in the commit message!

llvm/lib/Target/AIE/AIESuperRegRewriter.cpp
```diff
@@ -172,17 +180,32 @@ bool AIESuperRegRewriter::runOnMachineFunction(MachineFunction &MF) {
...
+  for (auto &VReg : UnAssignedPhysRegs) {
+    MCRegister DummyPhysReg;
```
andcarminati, line 207: A brief explanation of this part will be nice!

andcarminati:
Hi @krishnamtibrewala, I see a good potential in your approach! Did you manage to get some QoR numbers?

krishnamtibrewala:
> Hi @krishnamtibrewala, I see a good potential in your approach! Did you manage to get some QoR numbers?

Thank you @andcarminati for the feedback, I will incorporate all of them. 
Currently focusing on fixing a small issue(an annoying one) that get exposed due to this strategy.  

Following are the result for Conv2D_bfp16* 
image

llvm/lib/Target/AIE/aie2p/AIE2PTargetMachine.cpp
krishnamtibrewala: Yes, the idea of exposing TII was to move switch under a AIE-specific hook.
Chose to delay that until I see some result.

llvm/test/CodeGen/AIE/staged-ra-rewrite.mir
```diff
@@ -423,17 +423,17 @@ body:             |
...
   ; AIE2P-VREGS-NEXT: bb.1:
-  ; AIE2P-VREGS-NEXT:   [[COPY4:%[0-9]+]]:edc = COPY [[MOV_PD_imm11_pseudo]]
-  ; AIE2P-VREGS-NEXT:   [[COPY5:%[0-9]+]]:edn = COPY [[COPY2]]
-  ; AIE2P-VREGS-NEXT:   [[COPY6:%[0-9]+]]:edj = COPY [[COPY3]]
-  ; AIE2P-VREGS-NEXT:   [[COPY7:%[0-9]+]]:em = COPY [[COPY1]]
-  ; AIE2P-VREGS-NEXT:   [[COPY:%[0-9]+]]:ep, [[COPY4:%[0-9]+]]:edc = PADDA_2D_split [[COPY]], [[COPY7]], [[COPY5]], [[COPY6]], [[COPY4]]
+  ; AIE2P-VREGS-NEXT:   [[COPY4:%[0-9]+]]:edc_as_32bit = COPY [[MOV_PD_imm11_pseudo]]
+  ; AIE2P-VREGS-NEXT:   [[COPY5:%[0-9]+]]:edn_as_32bit = COPY [[COPY2]]
+  ; AIE2P-VREGS-NEXT:   [[COPY6:%[0-9]+]]:edj_as_32bit = COPY [[COPY3]]
+  ; AIE2P-VREGS-NEXT:   [[COPY7:%[0-9]+]]:em_as_32bit = COPY [[COPY1]]
```
krishnamtibrewala, line 435: TODO :  Try to understand why COPY from physical reg leads to spill_* and COPY from some [[COPY*]] lead to *_32bit RC

llvm/lib/Target/AIE/aie2p/AIE2PTargetMachine.cpp
andcarminati: I guess it could work with just target instructions, because we already have the 3D filter:
```
static bool onlyAllocateTargetInstructions(MachineRegisterInfo &MRI,
                                            const TargetInstrInfo &TII,
                                            const LiveInterval *LI) {
  const Register Reg = LI->reg();
  return std::any_of(
      MRI.use_nodbg_instructions(Reg).begin(),
      MRI.use_nodbg_instructions(Reg).end(), [&](const MachineInstr &MI) {
        return MI.getOpcode() > TargetOpcode::GENERIC_OP_END;
    });
}
```

krishnamtibrewala: Not exactly sure of your approach

I wanted that after the COPY are generated by the live rage splitting logic, which will introduce more 3D regs, instructions that require the construct of 3D reg ( by construct I mean to create a 3d reg the sub reg used matters "m1 dn1 dj1 dc1 dn5 dj5 dc5") function `onlyAllocateLIwith3DInstruction(...)` focuses on such instructions only.

the function you have suggested would do the same as `createGreedyRegisterAllocator(onlyAllocate3DRegisters)`

Ahh I think by 3D filter you mean "3D" reg filter, and what I wanted was combination of 3D reg filter and 3D use instructions.

andcarminati: We already filtered by register class, so, if we consider only `MI.getOpcode() > TargetOpcode::GENERIC_OP_END`  We will exclude all copy and other generic instructions. 

BTW, with the recent bump, they changed the filter-by-class by filter-by-reg, we there is no need to change core regalloc anymore, that is a nice change ;-)

krishnamtibrewala: Ahh I see what you mean, that will be nice.

PR #443, opened 2025-04-23
[NFC] Some improvements in pipeliner compile time

martien-de-jong:
Don't iterate a strategy if nothing changed.
use RecMII to detect failure early, plus some spin-off in the graphviz dump of the DAG.
Detect empty intervals early. Especially important for alternating scheduling modes

### DIFF
--- a/llvm/lib/Target/AIE/AIEPostPipeliner.cpp
+++ b/llvm/lib/Target/AIE/AIEPostPipeliner.cpp
@@ -141,7 +141,8 @@ int PostPipeliner::getResMII(MachineBasicBlock &LoopBlock) {
 
 // This assigns Cycle of SU, Earliest of its successors and Latest of its
 // predecessors
-void PostPipeliner::scheduleNode(SUnit &SU, int Cycle) {
+void PostPipeliner::scheduleNode(SUnit &SU, int Cycle,
+                                 PostPipelinerStrategy &Strategy) {
   LLVM_DEBUG(dbgs() << "PostPipelined SU" << SU.NodeNum << " in cycle " << Cycle
                     << ": " << *SU.getInstr());
   Info[SU.NodeNum].Cycle = Cycle;
@@ -161,6 +162,7 @@ void PostPipeliner::scheduleNode(SUnit &SU, int Cycle) {
       Info[SNum].LastEarliestPusher = SU.NodeNum;
       Info[SNum].Earliest = NewEarliest;
       Info[SU.NodeNum].NumPushedEarliest++;
+      Strategy.setChanged();
     }
   }
   LLVM_DEBUG(dbgs() << "\n  Pushed preds Latest: ");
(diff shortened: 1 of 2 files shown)


llvm/lib/Target/AIE/AIEPostPipeliner.cpp
```diff
@@ -662,7 +662,12 @@ bool PostPipeliner::scheduleFirstIteration(PostPipelinerStrategy &Strategy) {
     MachineInstr *MI = SU.getInstr();
     const int Earliest = Strategy.earliest(SU);
     const int Latest = Strategy.latest(SU);
-    // Find the first cycle that fits. We try every position modulo II
+    if (Earliest > Latest) {
```
martien-de-jong, line 667: This was insidious. We would actually revert the insertion direction, since fit() computes that from the range. This would allow a latency violation, only detected when checking the other iterations. So the error was masked by an extra check. We short-circuit all of that by detecting an empty interval early.

llvm/lib/Target/AIE/AIEPostPipeliner.cpp
```diff
@@ -321,6 +321,91 @@ bool PostPipeliner::computeBackward() {
...
+    // We can't conclude this is the longest path from start to end,
+    // but we will visit all paths.
+    return 0;
+  }
+  if (Start->NodeNum > End->NodeNum) {
+    // We are topologically ordered, so any forward edge to End will start
+    // from an earlier one. We can prune, knowing we will not find a circuit.
+    return {};
+  }
+  // We memoize the result, which prevents searching path suffixes
```
andcarminati, line 343: nit: `memoize`

llvm/lib/Target/AIE/AIEPostPipeliner.h
```diff
@@ -179,6 +179,9 @@ class PostPipeliner {
   ScheduleInfo Info;
   int MinLength;
 
+  /// The length of the longest circuit in the graph
```
andcarminati: nit: `graph.`

llvm/lib/Target/AIE/AIEPostPipeliner.cpp
```diff
@@ -857,15 +979,14 @@ bool PostPipeliner::tryHeuristics() {
       DEBUG_SUMMARY(dbgs() << "--- Strategy " << S.name() << " run=" << Run
                            << "\n");
       if (scheduleFirstIteration(S) && scheduleOtherIterations(S)) {
-        DEBUG_SUMMARY(dbgs() << "    Strategy " << S.name() << " run=" << Run
-                             << " found schedule:\n");
-        const bool Success = checkStages();
```
andcarminati, line 862: redundant check?

andcarminati (dismissed):
LGTM.

llvm/lib/Target/AIE/AIEPostPipeliner.cpp
martien-de-jong: Yes, this has been made part of scheduleFirstIteration in an earlier version.

andcarminati (approved):

PR #446, opened 2025-04-23
Recognize constant trip count

serhkurt:
[AIECC-947] ticket issue is resolved.


llvm/lib/Target/AIE/AIEPostPipeliner.cpp
```diff
@@ -94,27 +96,61 @@ bool PostPipeliner::isPostPipelineCandidate(MachineBasicBlock &LoopBlock) {
...
-    if (TII->isZOLTripCountDef(MI, Pristine)) {
-      TripCountDef = &MI;
+  std::optional<int> MyTripCount; 
+  for (auto RI = Preheader->rbegin(); RI != Preheader->rend();++RI) {
+    MachineInstr &MI = *RI;
+    if (TII->isZOLTripCountDef(MI, Pristine)) {  
+      TripCountDef = &MI;   
+      Register Vreg = MI.getOperand(1).getReg();
+      for(auto RI2 = RI;RI2 !=Preheader->rend();++RI2) {
+        if (RI2->modifiesRegister(Vreg) && RI2->getNumOperands()>1) {
```
andcarminati: If an instruction `modifiesRegister(Vreg)`, but `!isMoveImmediate()` we should stop the search, because it can be anything, like a add or shift that writes to that register. In this case we cannot assume a constant value anymore.

andcarminati: I think we can also skip `RI2->getNumOperands()>1`.

serhkurt: Yesterday I noticed it as well, and force pushed that part.

llvm/lib/Target/AIE/AIEPostPipeliner.cpp
```diff
@@ -94,27 +94,55 @@ bool PostPipeliner::isPostPipelineCandidate(MachineBasicBlock &LoopBlock) {
...
-  for (auto &MI : reverse(*Preheader)) {
-    if (TII->isZOLTripCountDef(MI, Pristine)) {
-      TripCountDef = &MI;
+  std::optional<int> MyTripCount; 
+  for (auto RI = Preheader->rbegin(); RI != Preheader->rend();++RI) {
+    MachineInstr &MI = *RI;
+    if (TII->isZOLTripCountDef(MI, Pristine)) {  
+      TripCountDef = &MI;   
+      Register Vreg = MI.getOperand(1).getReg();
+      for(auto RI2 = RI;RI2 !=Preheader->rend();++RI2) {
```
martien-de-jong: Your spacing is weird. Did you install the commit hooks?

llvm/lib/Target/AIE/AIEPostPipeliner.cpp
```diff
@@ -94,27 +94,55 @@ bool PostPipeliner::isPostPipelineCandidate(MachineBasicBlock &LoopBlock) {
...
+        if (RI2->modifiesRegister(Vreg) && RI2->getNumOperands()>1) {
+          MachineOperand &Src = RI2->getOperand(1);
```
martien-de-jong: nit: we can move this into the if. Rationale: 
a- Make it clear to the reader that we don't use this value anywhere else.
b- Avoid unnecessary actions.
c- Since you have checked that it is an immediate move, you can get its immediate operand without checking the number of operands like Andreu remarked. The latter is implied.

martien-de-jong (commented):
I think you forgot to add the test. you need an explicit git add `<new test file>`

llvm/lib/Target/AIE/AIEPostPipeliner.cpp
martien-de-jong: Coding style tip: You are starting a new iteration within an existing one, at the same time ending the outer one. To do that you had to change the nice range-based for statement to an explicit iterator one to get to the 'current iterator.
Instead, you could keep the original iteration, just changing state when you find the tripcount definition. The body would look like:
```
if (!TripcountDef) {
  if (isZOLTripCountDef()) {
    TripCountDef = &MI;
    VReg = ...
  }
} else {
  if (defines(VReg)) {
    ...
    break;
  }
}
```

llvm/test/CodeGen/AIE/aie2p/schedule/postpipeliner/TestMinTripCount.ll
```diff
@@ -0,0 +1,100 @@
+; NOTE: Assertions have been autogenerated by utils/update_llc_test_checks.py UTC_ARGS: --version 4
+;
+; This file is licensed under the Apache License v2.0 with LLVM Exceptions.
+; See https://llvm.org/LICENSE.txt for license information.
+; SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+;
+; (c) Copyright 2024 Advanced Micro Devices, Inc. or its affiliates
```
martien-de-jong: nit: 2025

llvm/test/CodeGen/AIE/aie2p/schedule/postpipeliner/TestMinTripCount.ll
```diff
@@ -0,0 +1,100 @@
...
+  target datalayout = "e-m:e-p:20:32-i1:8:32-i8:8:32-i16:16:32-i32:32:32-f32:32:32-i64:32-f64:32-a:0:32-n32"
+  target triple = "aie2p-none-unknown-elf"
+
+  define dso_local noundef i32 @_Z9test_loopPi(ptr nocapture readonly %arr) local_unnamed_addr #0 {
+; CHECK-LABEL: _Z9test_loopPi:
+; CHECK:         .p2align 4
+; CHECK-NEXT:  // %bb.0: // %entry
+; CHECK-NEXT:    nopa ; nopb ; nops ; movxm ls, #.LBB0_1; nopv
+; CHECK-NEXT:    mova r0, #7; nopb ; movxm le, #.L_LEnd0
+; CHECK-NEXT:    add.nc lc, r0, #0
```
martien-de-jong: I see no evidence that the postpipeliner accepted the loop. You should change the example so that the pipeliner kicks in, which is indicated by a non-zero third operand of this instruction.
Moreover, you run the whole codegenerator, which makes the test sensitive to changes in any of the preceeding passes. Instead, you can do all the preprocessing and save the result in an mir file by running
`llc -O2 --stop-before=postmisched -o file.mir`

This can then be given as input to` llc -O2 --start-before=postmisched file.mir`

andcarminati:
I think this cpp code can give a nice test to explore:

```
int sum(int n, int *vector) {
    int sum = 0;
    for (int i = 0; i < n; i++) {
        sum += vector[i];
    }
    return sum;
}

int a[] = {1, 2, 3, 4, 5, 6};
int b[] = {6, 7, 8, 9, 10, 11};

int main() {

    return sum(6, a) + sum(6, b);
 }
```

You can use to get the assembly:
```
clang -target aie2p -O2 example-pipeline.cpp -c -o test.s -S -nostdinc++ -nostdlib++ -fno-unroll-loops
```

In this example, I guess the implementation will handle the first loop but not the second.

llvm/lib/Target/AIE/AIEPostPipeliner.cpp
```diff
@@ -94,13 +100,31 @@ bool PostPipeliner::isPostPipelineCandidate(MachineBasicBlock &LoopBlock) {
...
+
   const bool Pristine = true;
+  std::optional<int> MyTripCount;
+  TripCountDef = nullptr;
+  Register Vreg;
   for (auto &MI : reverse(*Preheader)) {
-    if (TII->isZOLTripCountDef(MI, Pristine)) {
-      TripCountDef = &MI;
-      break;
+    if (!TripCountDef) {
```
andcarminati: We need to look to all blocks from which you can reach the loop to trace de constant definition. We should also do this after we reach the conclusion that there is no `ParsedMinTripCount` (compile time performance.)

llvm/lib/Target/AIE/AIEPostPipeliner.cpp
```diff
@@ -123,6 +146,56 @@ bool PostPipeliner::isPostPipelineCandidate(MachineBasicBlock &LoopBlock) {
...
+          if (MI.isMoveImmediate()) {
+            MyTripCount = MI.getOperand(1).getImm();
+          }
+          IsNotMoveModifies = true;
+          break;
+        }
+      }
+    }
+
+    if (MyTripCount.has_value() || IsNotMoveModifies) {
```
andcarminati: Oh, if we already have `MyTripCount`, but found another change to the register, we will break and return what we already have: one block assigns a constant and other assigns something else and we take the constant.

llvm/lib/Target/AIE/AIEPostPipeliner.cpp
```diff
@@ -123,6 +146,56 @@ bool PostPipeliner::isPostPipelineCandidate(MachineBasicBlock &LoopBlock) {
...
+        if (TII->isZOLTripCountDef(MI, Pristine)) {
+          TripCountState = true;
+          Vreg = MI.getOperand(1).getReg();
+        }
+
+      } else {
+
+        if (MI.modifiesRegister(Vreg)) {
+          if (MI.isMoveImmediate()) {
+            MyTripCount = MI.getOperand(1).getImm();
```
andcarminati: If we found a `MyTripCount` in the past, and found another one now, the values should match! We can have a path with LC=4 and another with LC=1, for example. If we pipeline considering 4 (let's say we construct a 2-stage pipeline) but execute considering 1 we will have a wrong code.

llvm/lib/Target/AIE/AIEPostPipeliner.cpp
```diff
@@ -123,6 +146,56 @@ bool PostPipeliner::isPostPipelineCandidate(MachineBasicBlock &LoopBlock) {
   return true;
 }
 
+std::optional<int>
+PostPipeliner ::extractMinTripCount(MachineBasicBlock *LoopPreheader) {
+  const bool Pristine = true;
+  bool IsNotMoveModifies = false;
+  std::optional<int> MyTripCount;
+  Register Vreg;
```
andcarminati: Vreg can be understood as `as` virtual register. We can use a more descriptive name here.

llvm/lib/Target/AIE/AIEPostPipeliner.cpp
```diff
@@ -96,52 +102,54 @@ bool PostPipeliner::isPostPipelineCandidate(MachineBasicBlock &LoopBlock) {
...
   if (!ParsedMinTripCount && !MyTripCount.has_value()) {
     LLVM_DEBUG(dbgs() << " PostPipeliner: No min tripcount\n");
     return false;
   }
 
-
-  if (ParsedMinTripCount){
+  if (ParsedMinTripCount) {
     MinTripCount = *ParsedMinTripCount;
+  } else {
```
martien-de-jong, line 130: This is a multi branch if (or if-ladder). You can make that clear by writing
```
if
else if
else
```
where all branches have the same indentation level.

llvm/test/CodeGen/AIE/aie2p/schedule/postpipeliner/TestMinTripCount.mir
martien-de-jong, line 1: It would look better to pre-squash this with the previous commit, since you've moved your test from .ll to .mir, which is one logical step. Feel free to ask me how to do that.

F-Stuckmann:
You may want to remove the merged aie-public commit. 
To update your branch to aie-public, you should rebase, because then the git commit history is not obfuscated once you merge your branch into aie-public.

PR #447, opened 2025-04-24
[AIE] Change the default of aie-addrspace-none-is-safe to true

martien-de-jong:
This enables scheduling loads in A and B together if either of them doesn't have a memory bank annotation.

We use the fact that the hardware serializes the memory accesses when a bank conflict arises, and buffer allocation does its best to avoid conflicts. Thus we always generate correct code, and have a statistical advantage.

Impact on many end-to-end and unit tests is rather big in actual diff volume, which is the natural result of cascading local scheduling diffs. I couldn't find any serious problems.

For QoR justification I refer to a comparison run that was posted in the aie-scc-qor-team slack channel last April 10.


andcarminati (dismissed):
LGTM.

additional nit: some updated tests could have the date also updated (super super nit).

andcarminati (approved):

PR #448, opened 2025-04-24
[AIEX] Allow HL for loops without minitercount information

andcarminati:
Please kindly note that some tests were kept without ZOL because they are testing the capabilities of the pre-pipeliner regarding the insertion of loop guards.

### DIFF
--- a/llvm/lib/Target/AIE/AIEBaseTargetTransformInfo.cpp
+++ b/llvm/lib/Target/AIE/AIEBaseTargetTransformInfo.cpp
@@ -23,6 +23,11 @@ static cl::opt<bool>
                            cl::desc("Enable hardware loops on AIE"),
                            cl::init(true), cl::Hidden);
 
+static cl::opt<bool> EnableZOLForLoopsWithoutMinIterCount(
+    "enable-aie-zol-without-minitercount",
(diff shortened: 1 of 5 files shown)


llvm/lib/Target/AIE/AIEBaseTargetTransformInfo.cpp
```diff
@@ -23,6 +23,11 @@ static cl::opt<bool>
                            cl::desc("Enable hardware loops on AIE"),
                            cl::init(true), cl::Hidden);
 
+static cl::opt<bool> EnableHLForLoopsWithoutMinIterCount(
+    "enable-aie-hl-without-minitercount",
```
martien-de-jong: nit: aie in front? (also the existing option?)
nit hl -> zol ?

andcarminati: I changed hl -> zol, but I kept aie in the same position (because of the other options). We could sanitize this in the future.

andcarminati:
Results:
image
image
By merging with https://github.com/Xilinx/llvm-aie/pull/459:
image

martien-de-jong (dismissed):

martien-de-jong (approved):

PR #449, opened 2025-04-25
[AIE] Combine REG_SEQ of copies of subregs into a copy of the superreg

martien-de-jong:
From a prototype of @konstantinschwarz
Rewrite a set of subreg copies into an equivalent superreg copy.

Some you win, some you lose. The story of near-optimality.

image
image
image


khallouh:
This looks good. We might have to do the same for calling convention copies where we also copy each subreg. For example:
`v64bfp16ebs8 test(v64bfp16ebs8 a) { return a; }`
compiles to 
```
_Z4test12v64bfp16ebs8:                  // @_Z4test12v64bfp16ebs8
	.p2align	4
// %bb.0:                               // %entry
	nopa	;		nopb	;		nops	;		ret	lr;		nopm	;		nopv	
	nop	                                //  Delay Slot 5
	mov	el0, el2                        //  Delay Slot 4
	mov	eh0, eh2                        //  Delay Slot 3
	vmov	x0, x2                          //  Delay Slot 2
.Lfunc_end4:
```
instead of simply:
```
_Z4test12v64bfp16ebs8:                  // @_Z4test12v64bfp16ebs8
	.p2align	4
// %bb.0:                               // %entry
	nopa	;		nopb	;		nops	;		ret	lr;		nopm	;		nopv	
	nop	                             
	nop                                
	nop                       
	vmov	ex0, ex2                  
	nop	                               
.Lfunc_end4:
```

llvm/lib/Target/AIE/AIEPostSelectOptimize.cpp
```diff
@@ -617,6 +617,70 @@ bool modifyStoreFlush(MachineBasicBlock &MBB, MachineRegisterInfo &MRI) {
...
+static bool combineSubRegCopyToSuperregCopy(MachineBasicBlock &MBB,
+                                            MachineRegisterInfo &MRI) {
+  const TargetInstrInfo *TII = MBB.getParent()->getSubtarget().getInstrInfo();
+  bool Changed = false;
+
+  for (MachineInstr &MI : make_early_inc_range(reverse(MBB))) {
+    if (!MI.isRegSequence())
+      continue;
+
+    LLVM_DEBUG(llvm::dbgs() << "Visiting ");
```
niwinanto: Nit: May be bit more context to the debug print.

llvm/lib/Target/AIE/AIEPostSelectOptimize.cpp
```diff
@@ -617,6 +617,70 @@ bool modifyStoreFlush(MachineBasicBlock &MBB, MachineRegisterInfo &MRI) {
...
+        break;
+      }
+
+      if (!SubRegDef->getOperand(1).getSubReg()) {
+        LLVM_DEBUG(llvm::dbgs() << "Src COPY is not a subreg COPY. Skip\n");
+        CommonSrcReg = {};
+        break;
+      }
+
+      unsigned SrcIdx = SubRegDef->getOperand(1).getSubReg();
```
niwinanto: Nit: SrcSubRegIdx may be

llvm/lib/Target/AIE/AIEPostSelectOptimize.cpp
```diff
@@ -617,6 +617,70 @@ bool modifyStoreFlush(MachineBasicBlock &MBB, MachineRegisterInfo &MRI) {
...
+
+      if (SubRegIdx != SrcIdx) {
+        LLVM_DEBUG(llvm::dbgs() << "SubReg indexes are not the same. Skip\n");
+        CommonSrcReg = {};
+        break;
+      }
+
+      Register SubRegCopySrc = SubRegDef->getOperand(1).getReg();
+      if (CommonSrcReg && SubRegCopySrc != CommonSrcReg) {
```
niwinanto, line 659: I am wondering, does this algorithm hold when we have partial copies or reg sequence with same subregister indices?

khallouh (dismissed):
LGTM

llvm/test/CodeGen/AIE/aie2p/GlobalIsel/post-select-combine.mir
```diff
@@ -0,0 +1,158 @@
...
+    ; CHECK-NEXT: [[COPY:%[0-9]+]]:vec512 = COPY $x0
+    ; CHECK-NEXT: [[COPY1:%[0-9]+]]:vec256 = COPY [[COPY]].sub_256_lo
+    ; CHECK-NEXT: [[COPY2:%[0-9]+]]:vec256 = COPY [[COPY]].sub_256_hi
+    ; CHECK-NEXT: [[REG_SEQUENCE:%[0-9]+]]:vec512 = REG_SEQUENCE [[COPY2]], %subreg.sub_512_lo, [[COPY1]], %subreg.sub_512_hi
+    ; CHECK-NEXT: $x0 = COPY [[REG_SEQUENCE]]
+    ; CHECK-NEXT: PseudoRET implicit $lr, implicit $x0
+    %0:vec512 = COPY $x0
+    %2:vec256 = COPY %0.sub_256_lo
+    %3:vec256 = COPY %0.sub_256_hi
+    %4:vec512 = REG_SEQUENCE %3, %subreg.sub_512_lo, %2, %subreg.sub_512_hi
```
niwinanto: `sub_256` in the copy and `sub_512` in the REG_SEQUENCE

llvm/lib/Target/AIE/AIEPostSelectOptimize.cpp
niwinanto: Discussed offline.

llvm/lib/Target/AIE/AIEPostSelectOptimize.cpp
```diff
@@ -617,6 +617,70 @@ bool modifyStoreFlush(MachineBasicBlock &MBB, MachineRegisterInfo &MRI) {
...
+    Register CommonSrcReg;
+    for (unsigned I = 1; I < MI.getNumExplicitOperands(); I += 2) {
+
+      Register SrcReg = MI.getOperand(I).getReg();
+      unsigned SubRegIdx = MI.getOperand(I + 1).getImm();
+
+      auto *SubRegDef = MRI.getVRegDef(SrcReg);
+      if (!SubRegDef->isCopy()) {
+        LLVM_DEBUG(llvm::dbgs() << "Src is not a COPY. Skip\n");
+        CommonSrcReg = {};
```
niwinanto, line 640: Can we use `NoRegister` over `{}`

llvm/lib/Target/AIE/AIEPostSelectOptimize.cpp
martien-de-jong: As discussed offline. The REGSEQ fully defines the result. If the list is incomplete, the missing parts of the register are undefined in the result. We choose to copy them from the original source.
If this partial definition is an idiomatic way of implying undef subregisters it may be the cause of the regression I see. I will check.

llvm/lib/Target/AIE/AIEPostSelectOptimize.cpp
```diff
@@ -617,6 +617,70 @@ bool modifyStoreFlush(MachineBasicBlock &MBB, MachineRegisterInfo &MRI) {
...
+  const TargetInstrInfo *TII = MBB.getParent()->getSubtarget().getInstrInfo();
+  bool Changed = false;
+
+  for (MachineInstr &MI : make_early_inc_range(reverse(MBB))) {
+    if (!MI.isRegSequence())
+      continue;
+
+    LLVM_DEBUG(llvm::dbgs() << "Visiting ");
+    LLVM_DEBUG(MI.dump());
+    Register CommonSrcReg;
```
niwinanto, line 628: Since we use this variable throughout, it would be nice to initialize.

llvm/test/CodeGen/AIE/aie2p/GlobalIsel/post-select-combine.mir
martien-de-jong: Good catch, thank you! I think I got a sanity check on using the wrong index on the COPY. RE_SEQUENCE seems to be more relaxed.

llvm/lib/Target/AIE/AIEPostSelectOptimize.cpp
martien-de-jong: We can, but why?

llvm/lib/Target/AIE/AIEPostSelectOptimize.cpp
martien-de-jong: It is default initialized.

llvm/lib/Target/AIE/AIEPostSelectOptimize.cpp
niwinanto: I just wanted to make it bit explicit.

andcarminati:
For the future: We could extend `PeepholeOptimizer` to handle some cases that are handled here.

llvm/lib/Target/AIE/AIEPostSelectOptimize.cpp
andcarminati: const?

martien-de-jong:
Regressions are diffuse. The net effect of the change is observed through the entirety of register coalescing and register allocation. One point of uncertainty I have is REG_SEQ that doesn't fill out the entire register, since then it may be cheaper to split into subregisters.

llvm/lib/Target/AIE/AIEPostSelectOptimize.cpp
```diff
@@ -617,6 +617,73 @@ bool modifyStoreFlush(MachineBasicBlock &MBB, MachineRegisterInfo &MRI) {
...
+  for (MachineInstr &MI : make_early_inc_range(reverse(MBB))) {
+    if (!MI.isRegSequence())
+      continue;
+
+    LLVM_DEBUG(llvm::dbgs() << "SubregCopyToSuperRegCopy visiting ");
+    LLVM_DEBUG(MI.dump());
+    Register CommonSrcReg;
+    // TODO: We may want to check that the list covers the whole superregister
+    // lest we run the risk of replacing an undefined register with the defined
+    // source subreg.
```
martien-de-jong, line 631: Perhaps compose the lanemask from the list and compare it to the superreg's lanemask?

niwinanto (approved):

PR #450, opened 2025-04-28
[AIEX] Remove duplicated imm. assignments to Simplifiable regs

andcarminati:
This aims to help LICM to hoist configuration registers assignments using immediate moves (generated by aie2p instruction selector)


llvm/lib/Target/AIE/AIEPostSelectOptimize.cpp
```diff
@@ -72,10 +72,17 @@ struct TrackableCopyOperands {
...
+struct TrackableMoveImmOperands {
+  MCRegister PhysReg;
+  int64_t ImmValue;
+};
+
 /// Track SSA copies of simple physical registers.
 /// Simple means they don't have sub or super registers.
 class PhysRegCopyTracker {
   DenseMap<MCRegister, SmallSet<Register, 4>> Copies;
+  DenseMap<MCRegister, int64_t> ImmAssign;
```
martien-de-jong: nit: I don't think that these fall under 'SSA copies', so the comment is inaccurate now

llvm/lib/Target/AIE/AIEPostSelectOptimize.cpp
```diff
@@ -155,11 +184,46 @@ getTrackableCopyOperands(const MachineInstr &MI,
...
-  auto [DstReg, SrcReg] = MI.getFirst2Regs();
+
+  Register DstReg = MI.getOperand(0).getReg();
+  MachineOperand &SrcMO = MI.getOperand(1);
+
+  // Fast skip relocations.
+  if (!SrcMO.isReg() && !SrcMO.isImm())
+    return false;
+
+  if (SrcMO.isImm()) {
```
martien-de-jong: nit: We are testing isImm() twice in succession. We can make the early return based on isReg() after dealing with isImm()

llvm/lib/Target/AIE/AIEPostSelectOptimize.cpp
```diff
@@ -155,11 +184,46 @@ getTrackableCopyOperands(const MachineInstr &MI,
   return {};
 }
 
+/// Returns whether the operands of \p MI can be tracked by PhysRegCopyTracker
+std::optional<TrackableMoveImmOperands>
```
martien-de-jong: I think TrackableMoveImmOperands is only used to implement this interface. Perhaps define it here?

martien-de-jong (commented):
It looks correct, but a bit convoluted.
Taking a step back, we were tracking virtregs copying the value of a physreg, which required tracking definitions of physregs for invalidation.
Now we have added tracking physregs copying immediate values, and we've built a separate tracking of physregs.  I think it would make sense to integrate the two, e.g. handlePhysRegsDefs(). That would check whether it was a physreg ref, invalidate the copy, and register a physreg value if it was an immediate def.
Also note that Register unifies PhysRegs and VirtRegs, and that VirtRegs can also be used in non-SSA form. Hence, we can generalize and track immediate values for VirtRegs as well by making a joint map keyed by Register.

llvm/test/CodeGen/AIE/aie2p/GlobalIsel/post-select-optimize-erase-physregs.mir
```diff
@@ -894,3 +894,113 @@ body:             |
...
+    %12:mbmm, %13:ep, %14:edc =  VLDA_2D_UPS_2x_dmw_lda_ups_w2b_upsSign0 $s3, %0:ep, $d0, implicit-def $srups_of, implicit $crsat, implicit $crupsmode, implicit $upssign0
+    $crupsmode = MOV_scalar_imm11_pseudo 1
+    %15:mbmm, %16:ep, %17:edc =  VLDA_2D_UPS_2x_dmw_lda_ups_w2b_upsSign0 $s3, %0:ep, $d0, implicit-def $srups_of, implicit $crsat, implicit $crupsmode, implicit $upssign0
+    $crupsmode = MOV_scalar_imm11_pseudo 1
+    %18:mbmm, %19:ep, %20:edc =  VLDA_2D_UPS_2x_dmw_lda_ups_w2b_upsSign0 $s3, %0:ep, $d0, implicit-def $srups_of, implicit $crsat, implicit $crupsmode, implicit $upssign0
+...
+
+# Test 15 - Remove immediate copies and normal copies that are duplicated.
+# Transform -
+# $crupsmode = MOV_scalar_imm11_pseudo 1  |   |    |   $crupsmode = MOV_scalar_imm11_pseudo 1
```
khallouh, line 953: I don't see any normal copies in this example. Looks like it was copied over from the previous test's comment?

andcarminati:
Results:

image

`Log_bf16_0` under investigation.

llvm/lib/Target/AIE/AIEPostSelectOptimize.cpp
```diff
@@ -62,17 +63,18 @@ static cl::opt<bool>
 
 namespace {
 
-/// Information about a COPY that can be tracked by PhysRegCopyTracker.
-struct TrackableCopyOperands {
-  Register VirtReg;
+/// Information about definitions of physical registers that can be tracked by
+/// PhysRegCopyTracker.
+struct TrackablePhysRegRefinitions {
```
martien-de-jong: Refinitions?

llvm/lib/Target/AIE/AIEPostSelectOptimize.cpp
```diff
@@ -62,17 +63,18 @@ static cl::opt<bool>
...
   MCRegister PhysReg;
   bool IsPhysRegDef; // Whether PhysReg is being redefined.
 };
 
-/// Track SSA copies of simple physical registers.
+/// Track SSA copies and constant initializations of simple physical registers.
 /// Simple means they don't have sub or super registers.
 class PhysRegCopyTracker {
-  DenseMap<MCRegister, SmallSet<Register, 4>> Copies;
+  DenseMap<MCRegister, std::variant<Register, int64_t>> LastAssigns;
```
martien-de-jong, line 77: Nice!

llvm/lib/Target/AIE/AIEPostSelectOptimize.cpp
```diff
@@ -93,25 +95,56 @@ class PhysRegCopyTracker {
...
     assert(range_size(TRI.regunits(PhysReg)) == 1 && "Phys reg has aliases.");
-    Copies[PhysReg].insert(VRegCopy);
+    if (std::holds_alternative<Register>(VirtRegOrConstant)) {
+      const Register VRegCopy = std::get<Register>(VirtRegOrConstant);
+      assert(VRegCopy.isVirtual());
+      LastAssigns[PhysReg] = VRegCopy;
+    } else {
+      const int64_t ImmValue = std::get<int64_t>(VirtRegOrConstant);
+      LastAssigns[PhysReg] = ImmValue;
+    }
```
martien-de-jong: It looks like an expensive way to copy VirtRegOrConstant

llvm/lib/Target/AIE/AIEPostSelectOptimize.cpp
```diff
@@ -93,25 +95,56 @@ class PhysRegCopyTracker {
...
-  /// Track that \p VRegCopy is a copy of \p PhysReg
-  void trackCopy(MCRegister PhysReg, Register VRegCopy) {
-    assert(VRegCopy.isVirtual());
+  /// Track that \p VirtRegOrConstant represents the current value of
+  /// \p PhysReg.
+  void
+  trackPhysRegCurrentValue(MCRegister PhysReg,
+                           std::variant<Register, int64_t> VirtRegOrConstant) {
     assert(range_size(TRI.regunits(PhysReg)) == 1 && "Phys reg has aliases.");
-    Copies[PhysReg].insert(VRegCopy);
```
martien-de-jong, line 100: We used to add to a set, now we keep only one?

llvm/lib/Target/AIE/AIEPostSelectOptimize.cpp
```diff
@@ -93,25 +95,56 @@ class PhysRegCopyTracker {
...
+
+  /// Return true if \p PhysReg currently contains \p ImmValue
+  bool isDefinedByConstantCopy(MCRegister PhysReg, int64_t ImmValue) const {
+    if (auto It = LastAssigns.find(PhysReg); It != LastAssigns.end()) {
+      // We cannot replace an Immediate with a Register copy.
+      if (!std::holds_alternative<int64_t>(It->second))
+        return false;
+
+      const int64_t CurrentImmValue = std::get<int64_t>(It->second);
+      return CurrentImmValue == ImmValue;
```
martien-de-jong: I'm wondering, can you not create a variant Search from ImmValue, and check whether `It->second == Search` ?

llvm/lib/Target/AIE/AIEPostSelectOptimize.cpp
```diff
@@ -127,9 +160,20 @@ class AIEPostSelectOptimize : public MachineFunctionPass {
...
-                         const TargetRegisterInfo &TRI) {
+std::optional<TrackablePhysRegRefinitions>
+getTrackablePhysRegDefsAndCopies(const MachineInstr &MI,
+                                 const TargetRegisterInfo &TRI) {
+
+  if (MI.isMoveImmediate()) {
+    Register DstReg = MI.getOperand(0).getReg();
+    if (TRI.isSimplifiableReservedReg(DstReg)) {
+      const int64_t ImmValue = MI.getOperand(1).getImm();
+      return TrackablePhysRegRefinitions{ImmValue, DstReg,
```
martien-de-jong: vscode doesn't check the names you type, it will happily autocomplete from earlier mistakes :-D

llvm/lib/Target/AIE/AIEPostSelectOptimize.cpp
```diff
@@ -142,33 +186,55 @@ getTrackableCopyOperands(const MachineInstr &MI,
...
+  if (MI.isMoveImmediate()) {
+    Register DstReg = MI.getOperand(0).getReg();
+    MachineOperand &SrcMO = MI.getOperand(1);
+    // Skip relocations.
+    if (!SrcMO.isImm())
+      return false;
+
+    const int64_t ImmValue = SrcMO.getImm();
+    if (DstReg.isPhysical() && CT.isDefinedByConstantCopy(DstReg, ImmValue)) {
+      // Assigning to DstReg a previously assigned value. Just remove MI.
```
martien-de-jong, line 203: Note: if the immediates are big and close, we may be able to replace it with and ADD.NC r1, r1, diff

llvm/lib/Target/AIE/AIEPostSelectOptimize.cpp
```diff
@@ -93,25 +95,56 @@ class PhysRegCopyTracker {
...
 
-  /// Track that \p VRegCopy is a copy of \p PhysReg
-  void trackCopy(MCRegister PhysReg, Register VRegCopy) {
-    assert(VRegCopy.isVirtual());
+  /// Track that \p VirtRegOrConstant represents the current value of
+  /// \p PhysReg.
+  void
+  trackPhysRegCurrentValue(MCRegister PhysReg,
+                           std::variant<Register, int64_t> VirtRegOrConstant) {
     assert(range_size(TRI.regunits(PhysReg)) == 1 && "Phys reg has aliases.");
```
martien-de-jong, line 103: I'm not sure whether a true alias, an alternative name for the same physical register, would not slip through this net.

martien-de-jong (commented):
Looks good. Just a few nits.

llvm/lib/Target/AIE/AIEPostSelectOptimize.cpp
```diff
@@ -93,25 +95,56 @@ class PhysRegCopyTracker {
...
   }
 
   /// If it exists, return a virtual register that holds a copy of \p PhysReg.
   std::optional<Register> getVirtualCopy(MCRegister PhysReg) const {
-    if (auto It = Copies.find(PhysReg);
-        It != Copies.end() && !It->second.empty())
-      return *It->second.begin();
+    if (auto It = LastAssigns.find(PhysReg); It != LastAssigns.end()) {
+      if (!std::holds_alternative<Register>(It->second))
+        return false;
```
niwinanto: Is it simply null{}, false will translate to 0 right?

llvm/lib/Target/AIE/AIEPostSelectOptimize.cpp
```diff
@@ -142,33 +186,55 @@ getTrackableCopyOperands(const MachineInstr &MI,
...
   }
   return {};
 }
 
 /// Returns true if \p MI is a COPY that was eliminated.
 bool trySimplifyCopy(MachineInstr &MI, const PhysRegCopyTracker &CT) {
-  if (!MI.isCopy())
+
+  if (MI.isCopy()) {
+    Register DstReg = MI.getOperand(0).getReg();
```
niwinanto: Nit: I feel `auto [DstReg, SrcReg] = MI.getFirst2Regs();` was good

llvm/lib/Target/AIE/AIEPostSelectOptimize.cpp
```diff
@@ -93,25 +95,56 @@ class PhysRegCopyTracker {
...
-  void trackCopy(MCRegister PhysReg, Register VRegCopy) {
-    assert(VRegCopy.isVirtual());
+  /// Track that \p VirtRegOrConstant represents the current value of
+  /// \p PhysReg.
+  void
+  trackPhysRegCurrentValue(MCRegister PhysReg,
+                           std::variant<Register, int64_t> VirtRegOrConstant) {
     assert(range_size(TRI.regunits(PhysReg)) == 1 && "Phys reg has aliases.");
-    Copies[PhysReg].insert(VRegCopy);
+    if (std::holds_alternative<Register>(VirtRegOrConstant)) {
```
niwinanto: check: `if (std::holds_alternative<Register>(VirtRegOrConstant))` means it hold the immediate right? Does the if else to be reversed?

llvm/lib/Target/AIE/AIEPostSelectOptimize.cpp
```diff
@@ -127,9 +160,20 @@ class AIEPostSelectOptimize : public MachineFunctionPass {
...
 
 /// Returns whether the operands of \p MI can be tracked by PhysRegCopyTracker
-std::optional<TrackableCopyOperands>
-getTrackableCopyOperands(const MachineInstr &MI,
-                         const TargetRegisterInfo &TRI) {
+std::optional<TrackablePhysRegRefinitions>
+getTrackablePhysRegDefsAndCopies(const MachineInstr &MI,
+                                 const TargetRegisterInfo &TRI) {
+
+  if (MI.isMoveImmediate()) {
```
niwinanto, line 140: Yes, `isMoveImm` always comes with 2 operands and second one with the imm value. Still I would like to see an assert for number of operands. Since, we can simply add this flag to any instruction.

llvm/lib/Target/AIE/AIEPostSelectOptimize.cpp
```diff
@@ -142,33 +186,55 @@ getTrackableCopyOperands(const MachineInstr &MI,
...
+
+  if (MI.isCopy()) {
+    Register DstReg = MI.getOperand(0).getReg();
+    Register SrcReg = MI.getOperand(1).getReg();
+    if (DstReg.isPhysical() && CT.isDefinedByRegisterCopy(DstReg, SrcReg)) {
+      // Assigning to DstReg a copy of itself. Just remove MI.
+      MI.eraseFromParent();
+      return true;
+    }
+    if (std::optional<Register> EquivVirtSrc;
```
niwinanto, line 184: This pattern looks so trivial, wondering why other generic passes does not do this for us.

niwinanto (commented):
The change looks good to me, especially the use of std::variant is really cool. Few nits and checks.

llvm/lib/Target/AIE/AIEPostSelectOptimize.cpp
andcarminati: We used to clean this every time before insertion, in practice it was holding just one register.

llvm/lib/Target/AIE/AIEPostSelectOptimize.cpp
andcarminati: Yeah, unfortunately there is no way to take the available data without knowing the type.

llvm/lib/Target/AIE/AIEPostSelectOptimize.cpp
andcarminati: Unfortunately it follows the same previous problem: we need type information to correct get the stored object.

llvm/lib/Target/AIE/AIEPostSelectOptimize.cpp
andcarminati: It could be a nice alternative way. The complication is another extension to retrieve the ADD.NC from TII that should be marinaded across different AIE targets.

llvm/lib/Target/AIE/AIEPostSelectOptimize.cpp
andcarminati: It could be, as this assert was here before I think we can keep.

llvm/lib/Target/AIE/AIEPostSelectOptimize.cpp
andcarminati: Good catch! Fixed it.

llvm/lib/Target/AIE/AIEPostSelectOptimize.cpp
andcarminati: It means that it holds a Register.

llvm/lib/Target/AIE/AIEPostSelectOptimize.cpp
andcarminati: Generically speaking no. Assert could be triggered if the target start to define and use conditional moves. Is not our case right now.

llvm/lib/Target/AIE/AIEPostSelectOptimize.cpp
andcarminati: Indeed, some cleaning can be performed by `MachineLateInstrsCleanup` however it runs very after our LICM. It is better to not reorder or run twice those Passes.

llvm/lib/Target/AIE/AIEPostSelectOptimize.cpp
martien-de-jong: But can't you just write `LastAssigns[PhysReg] = VirtRegOrConstant;` ?

andcarminati: I got it!

llvm/lib/Target/AIE/AIEPostSelectOptimize.cpp
martien-de-jong: I would like the tablegen backend to check the number of operands to which we add the IsMovImm attribute.... or actually the entire signature.

llvm/lib/Target/AIE/AIEPostSelectOptimize.cpp
martien-de-jong: It is a reserved register, the generic MachinePasses don't know whether a write to it doesn't have any nasty sideeffects. AFAIK, Gaetan added a target callback that says that it acts like a normal register. in LICM and DCE.
It doesn't mean I don't agree with you.

llvm/lib/Target/AIE/AIEPostSelectOptimize.cpp
martien-de-jong: I'm not suggesting to throw it away, just that it may not catch all dangerous situations.

martien-de-jong (approved):

Issue #352, opened 2025-02-13
Objdump assertion failure


jtuyls:
I am seeing the following assertion error with `llvm-objdump` on the attached object file:

[mm_npu4.o.txt](https://github.com/user-attachments/files/18782152/mm_npu4.o.txt)

```
mm_npu4.o:      file format elf32-aie

Disassembly of section .text:

00000000 <matmul_i8_i8_i32_32x32x32_8x8x8>:
       0: e1 00 00 10 00 84 00 00 00 5b 01 20 00 f0 2c 00       nopa    ;               nopb    ;               nops    ;               movxm   m1, #0x800;             nopv
      10: 7e 60 2b 00 0e 00 00 40 00 20 00 00 43 00     mova    r3, #0x2;               nopb    ;               paddxm   [sp], #0x40;           nops
      1e: 98 3d a0 10   lshl     r16, r2, r3
      22: f8 20 08 18   mov     m0, r16
      26: d4 41 00 f0 0c 41     padda    [p2], m0;              mov     m0, r0
      2c: ba 72 50 00 80 1d c7 f7 0c 01 padda    [p0], m0;              st       p6, [sp, #-60];                mov     m0, r1
      36: ba 78 60 8a ca 40 00 00 05 08 mova    r5, #0x40;              movx    r4, #0x6;               mov     r20, p2
      40: ba 78 60 48 0a 60 08 00 07 61 mova    r7, #0x308;             movx    r6, #0x100;             mov     r18, p0
      4a: 76 10 00 02 00 00 80 9d c3 f7 0c 21   padda    [p1], m0;              st       p7, [sp, #-64];                movxm   m0, #0x400
      56: ba 78 60 69 0a 21 00 00 01 00 mova    r1, #0x0;               movx    r2, #0x8;               mov     r19, p1

00000060 <.LBB0_1>:
      60: 02 70 90 34 00 60 2b 00       nops    ;               mov     p0, r18
      68: f8 20 6a 1c   mov     p4, r20
      6c: f8 20 6a 1a   mov     p2, r20
      70: e4 41 21 b0 49 08     lshl     r1, r1, r4;            mov     r0, r1
      76: e4 41 d2 b2 0a 0c     or       r16, r1, r5;           mov     p1, r18
      7c: e4 41 01 b1 47 80     lshl     r1, r16, r3;           mov     dj0, r1
[...]

PR #452, opened 2025-04-28
[AIE] Make the disassembler work for all cases.

martien-de-jong:
Effectively, we force incomplete decoder on every register operand

This fixes disassembly of paddb in its immediate form, which also surfaced in https://github.com/Xilinx/llvm-aie/issues/352

### DIFF
--- a/llvm/lib/Target/AIE/AIE2PRegOperandDef.td
+++ b/llvm/lib/Target/AIE/AIE2PRegOperandDef.td
@@ -11,6 +11,7 @@
 
 class AIE2PRegisterOperand<RegisterClass c> : RegisterOperand<c> {
   let EncoderMethod = "get" # c # OpValue;
+  let hasCompleteDecoder = false;
 }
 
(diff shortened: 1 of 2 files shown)


khallouh (approved):
Nice!

llvm/lib/Target/AIE/AIE2PRegOperandDef.td
```diff
@@ -11,6 +11,7 @@
 
 class AIE2PRegisterOperand<RegisterClass c> : RegisterOperand<c> {
   let EncoderMethod = "get" # c # OpValue;
+  let hasCompleteDecoder = false;
```
martien-de-jong, line 14: The effect of this is that every instruction with a register operand will lead to a TryDecode. In practice this TryDecode succeeds in most cases, not causing any extra overhead. In the cases where it fails it continues the search. The price that we pay is in the extra 'bytes to skip' that are added to the decoder table in order to be able to continue the search at the next entry.

PR #453, opened 2025-04-30
[AIE] Add loopsize + ZOLSetupSize checks for NumOfLoopBundles < 7

abnikant:
This adds checks for (LoopSize + ZOLSetupSize) and return if we already have elongated bundles size in ZOLSetup + LoopSize >= 112 bytes. This checks is added when NumOfLoopBundles < 7, for >=7 checks already exist. 

There is an improvement of 32-bytes in ReduceSumAxis kernels. 
Core_PMSize


martien-de-jong (commented):
Can we have a simple test please?

abnikant:
> Can we have a simple test please?

sure, added a test. please check.

andcarminati (dismissed):
LGTM.

llvm/test/CodeGen/AIE/aie2p/elongate/zol_phdr_loopbundles_112bytes.mir
```diff
@@ -0,0 +1,138 @@
...
+    renamable $r0 = ADD_add_r_ri killed renamable $r0, -1, implicit-def dead $srcarry
+    $lc = ADD_NC_mv_add_ri $r0, 0
+    $ls = MOVXM %bb.1
+    $le = MOVXM <mcsymbol .L_LEnd0>
+    renamable $x6, renamable $p0 = VLD_x_pstm_nrm_imm_pseudo killed renamable $p0, 0 :: (load (<32 x s16>) from %ir.p.addr.03)
+    NOPB
+    NOPA
+    NOPS
+    NOPXM
+    NOPV
```
martien-de-jong, line 117: Why do we start before postmisched and rely on pipelining to fill the prologue? If we provide bundled MIR we can just run and test the alignment pass, and drop the LLVM IR. It then also becomes relatively ease to build a few examples around the boundary cases.
As it is, the test is sensitive to details of pipelining, details of edges latency and basic details of scheduling (the nops have no data dependences, so might float anywhere), whereas the code change only has influence in the alignment pass.

abnikant: yes, I can add tests to just run alignment pass. This test checks creating edge-latency and running alignment pass after that. Let me add some tests.

martien-de-jong (approved):

PR #454, opened 2025-05-02
[AIE2p] Use multi-slot pseudo for const COPY with unique def

krishnamtibrewala:
This will help in Liner code. Since reg-to-reg copy goes only on one slot by using multi-slot in case of const copy we can bundle them or use different slot for better packing.
PS : Very small optimization

### DIFF
--- a/llvm/lib/Target/AIE/aie2p/AIE2PInstrInfo.cpp
+++ b/llvm/lib/Target/AIE/aie2p/AIE2PInstrInfo.cpp
@@ -557,10 +557,19 @@ void AIE2PInstrInfo::copyPhysReg(MachineBasicBlock &MBB,
 
   if (AIE2P::mMvSclSrcRegClass.contains(SrcReg) &&
       AIE2P::mMvSclDstRegClass.contains(DstReg)) {
-    // Build MultiSlotPseudo in preference
-    const unsigned MOVSclOpcode = getScalarMovOpcode(DstReg, SrcReg);
-    BuildMI(MBB, MBBI, DL, get(MOVSclOpcode), DstReg)
-        .addReg(SrcReg, getKillRegState(KillSrc));
+    if (MachineInstr *MI = MRI.getUniqueVRegDef(SrcReg);
+        MI && MI->isMoveImmediate()) {
+      // Try modifying scalar move to pseudo immediate move.
+      const int64_t Imm = MI->getOperand(1).getImm();
+      APInt ImmVal = APInt(64, Imm);
+      auto OpCode = getConstantMovOpcode(MRI, DstReg, ImmVal);
+      BuildMI(MBB, MBBI, DL, get(OpCode), DstReg).addImm(Imm);
+    } else {
+      // Build MultiSlotPseudo in preference
+      const unsigned MOVSclOpcode = getScalarMovOpcode(DstReg, SrcReg);
+      BuildMI(MBB, MBBI, DL, get(MOVSclOpcode), DstReg)
+          .addReg(SrcReg, getKillRegState(KillSrc));
+    }
   } else if ((AIE2P::eLRegClass.contains(SrcReg)) &&
              (AIE2P::eLRegClass.contains(DstReg))) {
     BuildMI(MBB, MBBI, DL, get(AIE2P::MOV_alu_mv_mv_mv_scl),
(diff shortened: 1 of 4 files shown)


krishnamtibrewala:
Note : The best place to do this would be after register coalescing, that is where const COPY are created, and we should do this before RA

llvm/lib/Target/AIE/aie2p/AIE2PInstrInfo.cpp
```diff
@@ -1130,6 +1138,19 @@ AIE2PInstrInfo::getRegOffsetSpillInstrInfoFromImmOffset(
   }
 }
 
+bool AIE2PInstrInfo::isConstatntMoveOpcode(unsigned Opcode) const {
+  switch (Opcode) {
+  case AIE2P::MOV_RLC_imm11_pseudo:
+  case AIE2P::MOV_PD_imm11_pseudo:
+  case AIE2P::MOV_S_imm11_pseudo:
+  case AIE2P::MOV_scalar_imm11_pseudo:
+  case AIE2P::MOVXM:
```
krishnamtibrewala: All these opcode should be sufficient since we are selecting one of these during instruction selection.

llvm/lib/Target/AIE/aie2p/AIE2PInstrInfo.cpp
```diff
@@ -1130,6 +1138,19 @@ AIE2PInstrInfo::getRegOffsetSpillInstrInfoFromImmOffset(
   }
 }
 
+bool AIE2PInstrInfo::isConstatntMoveOpcode(unsigned Opcode) const {
```
martien-de-jong: nit: misspelled name

llvm/lib/Target/AIE/aie2p/AIE2PInstrInfo.cpp
```diff
@@ -557,10 +557,18 @@ void AIE2PInstrInfo::copyPhysReg(MachineBasicBlock &MBB,
 
   if (AIE2P::mMvSclSrcRegClass.contains(SrcReg) &&
       AIE2P::mMvSclDstRegClass.contains(DstReg)) {
-    // Build MultiSlotPseudo in preference
-    const unsigned MOVSclOpcode = getScalarMovOpcode(DstReg, SrcReg);
-    BuildMI(MBB, MBBI, DL, get(MOVSclOpcode), DstReg)
-        .addReg(SrcReg, getKillRegState(KillSrc));
+    if (MachineInstr *MI = MRI.getUniqueVRegDef(SrcReg);
+        MI && isConstatntMoveOpcode(MI->getOpcode())) {
```
martien-de-jong: I'm pretty sure there's and existing isMoveImmediate() or suchlike running on an attribute

martien-de-jong:
How is this different from constant rematerialization by RA?

krishnamtibrewala:
> How is this different from constant rematerialization by RA?

Hi @martien-de-jong, i do not see that happening in RA, additionally the idea here is to covert the Scalar COPY instruction into PseudoImm move so that we can be packed in a same VLIW bundle.

martien-de-jong:
What is stopping you from implementing this before RA? Also, I think that register coalescing removes copies rather than creating them. I think that PHI elimination is the biggest creator of copies.

krishnamtibrewala:
> What is stopping you from implementing this before RA? Also, I think that register coalescing removes copies rather than creating them. I think that PHI elimination is the biggest creator of copies.


Hi @martien-de-jong you are right PHI is the biggest creator of COPY, the register coalescing pass helps to clean up these copies to a certain extend leading to IR like. (Note: When it comes to COPY from a unmatching sub-reg to sub-reg, coalescing pass does not do a great job for us)

bb.0
%1  = mov_imm_pseudo 0
%2 = COPY %1
%3 = ADD %1, %x

bb.1 
%4 = COPY %1

The only motivation to implement it before RA is the live range of %1 might reduce, aiding it in RA (both are big IFs)
The current implementation was more from ease of implementation & show a working PoC by using copyPhysReg(...) to pick a mov_imm_pseudo rather than mov_scl when possible, helping scheduler to do better bundling. 

I saw this helpful in Conv2D_bfp16_* test cases.

martien-de-jong:
@krishnamtibrewala yes, everything is related. There are more liverange considerations around REQ_SEQ and subreg use, especially across PHI nodes. I have a feeling that a combined PHI-elimination + constant materialization + register coalescing might be quite powerful. (although rematerialization might be reserved as a repair mechanism in core RA. It might influence coalescing decisions though.)

PR #455, opened 2025-05-02
[AIE2P] Fix hang during legalization of unmerge of v8s8

khallouh:
The generic legalization creates extracts which are custom legalized using the same unmerge resulting in an infinite loop. This fixes this by implementing custom legalization for unmerge.


llvm/lib/Target/AIE/AIELegalizerHelper.cpp
```diff
@@ -125,19 +125,20 @@ bool AIELegalizerHelper::pack32BitVector(LegalizerHelper &Helper,
   return true;
 }
 
-bool AIELegalizerHelper::unpack32BitVector(LegalizerHelper &Helper,
-                                           MachineInstr &MI,
-                                           Register SourceReg) const {
+// Unpack 32 bit or 64 bit vectors
+bool AIELegalizerHelper::unpackVector(LegalizerHelper &Helper, MachineInstr &MI,
+                                      Register SourceReg, bool Wide) const {
```
niwinanto: I would suggest derive `wide` inside the function, no need to update the function interface.

llvm/test/CodeGen/AIE/GlobalISel/legalize-shuffle-vector.mir
```diff
@@ -896,3 +896,333 @@ body: |
...
+    ; CHECK-NEXT: [[SHL4:%[0-9]+]]:_(s32) = G_SHL [[AND2]], [[C3]](s32)
+    ; CHECK-NEXT: [[OR5:%[0-9]+]]:_(s32) = G_OR [[OR4]], [[SHL4]]
+    ; CHECK-NEXT: [[AND3:%[0-9]+]]:_(s32) = G_AND [[ASSERT_SEXT3]], [[C7]]
+    ; CHECK-NEXT: [[SHL5:%[0-9]+]]:_(s32) = G_SHL [[AND3]], [[C2]](s32)
+    ; CHECK-NEXT: [[OR6:%[0-9]+]]:_(s32) = G_OR [[OR5]], [[SHL5]]
+    ; CHECK-NEXT: [[BITCAST1:%[0-9]+]]:_(<4 x s8>) = G_BITCAST [[OR6]](s32)
+    ; CHECK-NEXT: PseudoRET implicit $lr, implicit [[BITCAST1]](<4 x s8>)
+    %0:_(<8 x s8>) = COPY $l0
+    %1:_(<8 x s8>) = COPY $l0
+    %2:_(<4 x s8>) = G_SHUFFLE_VECTOR %0:_(<8 x s8>), %1:_, shufflemask(0, 1, 2, 3)
```
niwinanto, line 1224: The generated code is really sub-optimal, difficult to verify it is correct.

khallouh: Yes, there is no way around this I think because we are testing the legalization of shuffle which is suboptimal being the default (all combines taking place before). This shuffle test just showcases the bug. To verify the correctness of the feature you can refer to  the unmerge test `llvm/test/CodeGen/AIE/aie2p/GlobalIsel/legalize-unmerge-values.mir`

llvm/lib/Target/AIE/AIELegalizerHelper.cpp
khallouh: Done

llvm/lib/Target/AIE/AIELegalizerHelper.cpp
```diff
@@ -125,19 +125,21 @@ bool AIELegalizerHelper::pack32BitVector(LegalizerHelper &Helper,
...
   const LLT SourceRegTy = MRI.getType(SourceReg);
-  assert(SourceRegTy.getSizeInBits() == 32 &&
-         "cannot unpack vectors larger or smaller than 32-bit");
+  unsigned SourceRegSize = SourceRegTy.getSizeInBits();
+  assert((SourceRegSize == 32 || SourceRegSize == 64) &&
+         "cannot unpack vectors other than 32-bit or 64-bit");
 
-  const LLT S32 = LLT::scalar(32);
+  bool Wide = (SourceRegSize == 64) ? 1 : 0;
+  const LLT Type = Wide ? LLT::scalar(64) : LLT::scalar(32);
```
martien-de-jong: `LLT::scalar(SourceRegSize)` ?

khallouh: Done :)

llvm/test/CodeGen/AIE/GlobalISel/legalize-shuffle-vector.mir
niwinanto: Can we do some custom legalization for shuffle. May be `unmerge + merge ` operation .

khallouh: We might be able to do it but we don't need it since this pattern is handled in the middle-end. Moreover, if we want to add a similar combine in the backend (to optimize O0 because why not : ) ) then I think it is better to handle it in the prelegalizer combiner, the shuffle legalization using .lower() is only meant to be a fallback.

khallouh:
Note: the hang only occurs in O0 because the shuffle pattern leading to the unmerge is combined in IR at higher optimization levels and we don't come to the legalizer.

llvm/test/CodeGen/AIE/GlobalISel/legalize-shuffle-vector.mir
niwinanto: I was in an impression, we always gets this hang when we build the compiler and with this fix leading to increased memory footprint of the compiler. Yes, may be with `inline` we never hit this issue.

And regarding the legalization, idea was not to have the same middle end combiner in the legalizer, irrespective of what ever shuffle mask patterns, do `unmerge + merge` for vector of size less than 128-bit. As I can see, current generic lower rule for shuffle vector does `extract_vector + build_vector` which we support using real vectors. As you mentioned, since we dont hit this issue, may be we can create a ticket and move away.

llvm/lib/Target/AIE/AIELegalizerHelper.h
```diff
@@ -27,12 +27,14 @@ class AIELegalizerHelper {
   const AIEBaseSubtarget &ST;
 
   const LLT S32 = LLT::scalar(32);
+  const LLT S64 = LLT::scalar(64);
```
niwinanto: Nit: If they are not used, may be you can remove them.

niwinanto (dismissed):

niwinanto (approved):
LGTM!

llvm/lib/Target/AIE/AIELegalizerHelper.cpp
```diff
@@ -125,19 +125,20 @@ bool AIELegalizerHelper::pack32BitVector(LegalizerHelper &Helper,
...
 
   const LLT SourceRegTy = MRI.getType(SourceReg);
-  assert(SourceRegTy.getSizeInBits() == 32 &&
-         "cannot unpack vectors larger or smaller than 32-bit");
+  unsigned SourceRegSize = SourceRegTy.getSizeInBits();
+  assert((SourceRegSize == 32 || SourceRegSize == 64) &&
+         "cannot unpack vectors other than 32-bit or 64-bit");
 
-  const LLT S32 = LLT::scalar(32);
+  const LLT Type = LLT::scalar(SourceRegSize);
```
krishnamtibrewala, line 139: nit: we could have reused `SourceRegTy`

khallouh: We could not because `SourceRegTy` will be a vector type and `Type` a scalar (S64 or S32)

PR #451, opened 2025-04-28
[AIE2P] Allow more address chaining and add a profitability check heuristic

khallouh:
This allows chaining when the base pointer is used in other basic blocks but only when it is considered profitable: 
- This shouldn't happen in a loop as the resulting copy will be more costly 
- The cost of chaining is incremented for each offset falling outside the load/store immediate ranges. 
- An experimental threshold is used to determine if chaining is profitable based on the computed cost (e.g half the number of pointer adds to be chained)

Below QoR results for affected kernels, the rest are unchanged:
    
    
``` 
|---------------------------------------------|------------------------------------------|--------------|--------------|------------------------|--------------|------------------------|--------------|--------------------------------|-------------------------|------------------------------|-----------------------|--------------------------------|---------------|-----------------------|-------------------------|-------------------------|--------------|--------------|------------------------------|--------------------|-------------------------|------------------------------|---------------|---------------------------|---------------------------|---------------|-------------------------|---------------|----------------|---------------|----------------------|----------------|-------------------------------|-----------------------------------|------------------|----------------------------|------------------------------|-------------------------------|------------------------------|------------------------|-------------------------------|----------------------------|------------------|--------------------|---------------------------|-----------------------|-------------------------------|------------------------|--------------|---------------------------|---------------------------|------------------|-----------------------|------------------|------------------|------------------|------------------|---------------------------|---------------------------|---------------------------|----------------------------|------------------|------------------|-----------------------------------|------------------------|--------------|------------------|------------------|------------------------|------------------|------------------|--------------|------------|-------------|-------------|-------------|
[...]

PR #405, opened 2025-03-26
[AIEX] Global Combiner - Pointer Modifier

F-Stuckmann:
This implements a global combiner for determining post increments or offset Memory Instructions. 
The following is still work in progress: 

- There are 3 CombineResult classes with similarities that should be combined to one. This also includes adding the new CombineResult Class to the AIECombiners.
- The gain function of the combiners should be more general and moved to AIEGlobalCombiner. Depending on the QoR results, there may be some finetuning in the cost calculation.
-  Matching rules should be provided by .td, instead of being implemented as classes

That being said, the structure in `AIEPtrModPass` , `AIEGlobalCombiner` and  `BaseInstrInfo` should be stable and close to the final version.

PR #460, opened 2025-05-09
[AIEX][NFC] refactor Load/Store Imm check

F-Stuckmann:
This is a Non Functional Change to Move Immediate Checks from Instruction Selection to TII. 
The global combiner PR (https://github.com/Xilinx/llvm-aie/pull/405 ) needs this refactor, as well as https://github.com/Xilinx/llvm-aie/pull/451 .

### DIFF
--- a/llvm/lib/Target/AIE/AIE2InstrInfo.cpp
+++ b/llvm/lib/Target/AIE/AIE2InstrInfo.cpp
@@ -1434,6 +1434,75 @@ AIE2InstrInfo::getZOLSupport() const {
   return Result;
 }
 
+bool AIE2InstrInfo::isOffsetInImmediateRange(
+    unsigned Opcode, unsigned LoadStoreSize,
+    std::optional<APInt> Offset) const {
+  if (!Offset)
+    return false;
+
+  switch (Opcode) {
+  case AIE2::G_AIE_OFFSET_LOAD:
+  case AIE2::G_AIE_OFFSET_STORE: {
+    switch (LoadStoreSize) {
+    case 8:
+    case 16:
+      return checkSignedImmediateRange<3, 1>(Offset);
+    case 20:
+    case 32:
+      return checkSignedImmediateRange<6, 4>(Offset);
+    case 128:
+      return checkSignedImmediateRange<6, 16>(Offset);
+    case 256:
+      return checkSignedImmediateRange<6, 32>(Offset);
+    case 512:
+      return checkSignedImmediateRangeSplitting<6, 32, 32>(Offset);
+    default:
+      return false;
+    }
+  }
+  case AIE2::G_AIE_POSTINC_LOAD:
+  case AIE2::G_AIE_POSTINC_STORE: {
+    switch (LoadStoreSize) {
+    case 8:
+    case 16:
+      return checkSignedImmediateRange<4, 1>(Offset);
+    case 20:
+    case 32:
+      return checkSignedImmediateRange<7, 4>(Offset);
+    case 128:
+      return checkSignedImmediateRange<7, 16>(Offset);
+    case 256:
+    case 512:
+      return checkSignedImmediateRange<7, 32>(Offset);
+    default:
+      return false;
+    }
+  }
+  case AIE2::G_AIE_OFFSET_ZEXTLOAD:
+  case AIE2::G_AIE_OFFSET_SEXTLOAD: {
+    switch (LoadStoreSize) {
+    case 8:
+    case 16:
+      return checkSignedImmediateRange<3, 1>(Offset);
+    default:
+      return false;
+    }
+  }
+  case AIE2::G_AIE_POSTINC_SEXTLOAD:
+  case AIE2::G_AIE_POSTINC_ZEXTLOAD: {
+    switch (LoadStoreSize) {
+    case 8:
+    case 16:
+      return checkSignedImmediateRange<4, 1>(Offset);
+    default:
+      return false;
+    }
+  }
+  default:
+    return false;
+  }
+}
+
 unsigned AIE2InstrInfo::getPseudoJNZDOpcode() const { return AIE2::PseudoJNZD; }
 
 unsigned AIE2InstrInfo::getNumBypassedCycles(const InstrItineraryData *ItinData,
--- a/llvm/lib/Target/AIE/AIE2InstructionSelector.cpp
+++ b/llvm/lib/Target/AIE/AIE2InstructionSelector.cpp
@@ -1050,14 +1050,14 @@ AIE2InstructionSelector::getCombinedOpcodeSRSUPS(const MachineInstr &MemOp,
         switch (cast<GIntrinsic>(CombOp).getIntrinsicID()) {
         case Intrinsic::aie2_I512_v32_acc32_srs:
           FitsImmediateRange =
-              checkImmediateRangeSplitting<3, 32, 32>(Immediate);
+              checkSignedImmediateRangeSplitting<3, 32, 32>(Immediate);
           return LoadStoreOpcodes{
               /*ISelOpcode=*/AIE2::VST_SRS_S16_S32_ag_idx_imm,
               FitsImmediateRange,
               /*OffsetOpcode=*/AIE2::VST_SRS_S16_S32_ag_idx_imm};
         case Intrinsic::aie2_I512_v16_acc64_srs:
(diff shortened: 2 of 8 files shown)


llvm/lib/Target/AIE/AIE2InstrInfo.cpp
```diff
@@ -1434,6 +1434,89 @@ AIE2InstrInfo::getZOLSupport() const {
   return Result;
 }
 
+bool AIE2InstrInfo::isOffsetInImmediateRange(
+    unsigned Opcode, unsigned LoadStoreSize,
+    std::optional<APInt> Offset) const {
+  bool FitsImmediateRange = false;
```
andcarminati: I was thinking if we can have a data structure/class that can hold this information in a way that we can use lookup to find the information. Something like this:

```
OffsetRanges Ranges[] = {
    {G_AIE_OFFSET_LOAD, {
        {8, 3, 1},
        {16, 3, 1},
        ....
     }},
    {G_AIE_OFFSET_STORE, ...
    }}
}
};
``` 
Then we can use two nested fors in the base class to find the information, the derived class can give just the object.

martien-de-jong: I'm not against the switches, although it looks as we could eliminate a lot of breaks by directly returning.

llvm/lib/Target/AIE/AIE2InstructionSelector.cpp
```diff
@@ -2469,13 +2470,11 @@ LoadStoreOpcodes AIE2InstructionSelector::getLoadStoreOpcode(
     if (getLoadStoreSize(I) == 512) {
       unsigned RBID = deriveRegBankID(I.getOperand(0).getReg(), MRI, RBI);
       if (RBID == AIE2::AccRegBankID) {
-        FitsImmediateRange = checkImmediateRangeSplitting<6, 32, 32>(Offset);
```
martien-de-jong, line 2472: micronit: we now do the entire case analysis twice, once for the immediate range and once for the instruction selection.

llvm/lib/Target/AIE/AIE2InstrInfo.cpp
```diff
@@ -1434,6 +1434,89 @@ AIE2InstrInfo::getZOLSupport() const {
...
+  switch (Opcode) {
+  case AIE2::G_AIE_OFFSET_LOAD:
+  case AIE2::G_AIE_OFFSET_STORE: {
+    switch (LoadStoreSize) {
+    case 8:
+    case 16:
+      FitsImmediateRange = checkImmediateRange<3, 1>(Offset);
```
martien-de-jong: Could we make this check{Signed,Unsigned}ImmediateRange just to be clear?

llvm/lib/Target/AIE/AIEBaseInstrInfo.h
```diff
@@ -748,6 +752,22 @@ struct AIEBaseInstrInfo : public TargetInstrInfo {
   const MIRFormatter *getMIRFormatter() const override;
   mutable std::unique_ptr<AIEMIRFormatter> Formatter;
 };
+
+template <unsigned NumEncodingBits, unsigned Step>
+bool checkImmediateRange(std::optional<APInt> Immediate) {
+  unsigned MaxPow2 = NumEncodingBits + llvm::Log2_64(Step);
```
martien-de-jong: const

martien-de-jong: How about passing Log2_64(Step), so that both parameters have the same unit? I guess that computing pow2(StepBits) is cheaper than computing Log2(Step).

llvm/lib/Target/AIE/AIEBaseInstrInfo.h
```diff
@@ -748,6 +752,22 @@ struct AIEBaseInstrInfo : public TargetInstrInfo {
...
+  if (Immediate && isIntN(MaxPow2, Immediate->getSExtValue()) &&
+      Immediate->getSExtValue() % Step == 0) {
+    return true;
+  }
+  return false;
+}
+
+template <unsigned NumEncodingBits, unsigned Step, unsigned SplitOffset>
+bool checkImmediateRangeSplitting(std::optional<APInt> Immediate) {
+  return Immediate && checkImmediateRange<NumEncodingBits, Step>(Immediate) &&
```
martien-de-jong: nit: checkImmediateRange implies Immediate

llvm/lib/Target/AIE/AIEBaseInstructionSelector.h
```diff
@@ -198,24 +198,6 @@ inline unsigned getLoadStoreSize(const MachineInstr &MI) {
   return (*MI.memoperands_begin())->getSizeInBits().getValue();
 }
 
-template <unsigned NumEncodingBits, unsigned Step>
```
martien-de-jong, line 201: sorry, not your code. But still.

llvm/lib/Target/AIE/aie2p/AIE2PInstrInfo.cpp
```diff
@@ -1733,6 +1733,97 @@ AIE2PInstrInfo::getZOLSupport() const {
   return Result;
 }
 
+bool AIE2PInstrInfo::isOffsetInImmediateRange(
+    unsigned Opcode, unsigned LoadStoreSize,
+    std::optional<APInt> Offset) const {
```
martien-de-jong, line 1738: Early return false on !Offset?

llvm/lib/Target/AIE/AIEBaseInstrInfo.h
F-Stuckmann: These functions are used without checking the Immediate beforehand, if an Immediate is coming in, therefore removing the optional here is a rather large refactor

llvm/lib/Target/AIE/AIEBaseInstrInfo.h
F-Stuckmann: it is not my code and this change would be nicely suited for a proper cleanup. Because we now have multiple switch cases that checkImmediateRanges

llvm/lib/Target/AIE/AIEBaseInstrInfo.h
```diff
@@ -59,6 +59,10 @@ struct AIEBaseInstrInfo : public TargetInstrInfo {
     // of bundles.
     unsigned LoopSetupDistance;
   };
+  virtual bool isOffsetInImmediateRange(unsigned Opcode, unsigned LoadStoreSize,
+                                        std::optional<APInt> Offset) const {
+    llvm_unreachable("Target didn't implement OffsetFitImmRange");
```
martien-de-jong, line 64: nit: message out of sync with function name

llvm/lib/Target/AIE/AIEBaseInstrInfo.h
```diff
@@ -59,6 +59,10 @@ struct AIEBaseInstrInfo : public TargetInstrInfo {
     // of bundles.
     unsigned LoopSetupDistance;
   };
+  virtual bool isOffsetInImmediateRange(unsigned Opcode, unsigned LoadStoreSize,
```
martien-de-jong, line 62: I have a feeling that this may be a best guess for these generic opcodes. Perhaps document whether it should return true or false if there's uncertainty.

llvm/lib/Target/AIE/aie2p/AIE2PInstrInfo.cpp
```diff
@@ -1733,6 +1733,77 @@ AIE2PInstrInfo::getZOLSupport() const {
...
+    case 128:
+      return checkSignedImmediateRange<4, 16>(Offset);
+    case 256:
+      return checkSignedImmediateRange<4, 32>(Offset);
+    case 512:
+      return checkSignedImmediateRange<4, 64>(Offset);
+    case 1024:
+      return checkSignedImmediateRangeSplitting<4, 64, 64>(Offset);
+    case 2048:
+      return checkSignedImmediateRangeSplitting<4, 64, 192>(Offset);
```
martien-de-jong, line 1762: nit: I see that this checks 0 and 192, not the intermediate 64 and 128. I guess the implementation assumes that step divides splitoffset. Perhaps static_assert that?

martien-de-jong (approved):
There will always be nits

PR #461, opened 2025-05-09
[AIE] Embarrassing error set right.

martien-de-jong:
When we changed the setting of Latest to a setter, we also changed the setting of Earliest, for symmetry. However, in that chage we updated Earliest of the scheduled node, not that of its successors.


andcarminati (approved):
LGTM. I checked QoR and it is fine now. Thank you for the quick fix @martien-de-jong!

PR #462, opened 2025-05-09
[AIE2P] BFP16 update extract intrinsics

khallouh:


clang/lib/Headers/aie2p_upd_ext.h
```diff
@@ -1920,4 +1920,177 @@ INTRINSIC(v16acc64) set_v16acc64(int idx, v4acc64 b) {
...
+}
+
+INTRINSIC(v64int8) extract_v64int8(v64bfp16ebs16 v) { return v.mantissa; }
+
+INTRINSIC(v64int8) extract_data(v64bfp16ebs16 v) { return extract_v64int8(v); }
+
+INTRINSIC(int)
+extract_exponent(v64bfp16ebs16 v, int idx) {
+  v8int8 exp = v.exponent;
+  if (idx == 0)
```
niwinanto: check: does the idx is just 0 and 1? Or can this be 0-7?

Also, can we do something like this?
```
  v2int32 exp = v.exponent;
  return exp[idx ];
```

llvm/lib/Target/AIE/aie2p/AIE2PPreLegalizerCombiner.cpp
```diff
@@ -128,8 +128,10 @@ bool AIE2PPreLegalizerCombinerImpl::tryCombineAll(MachineInstr &MI) const {
   case TargetOpcode::G_SHUFFLE_VECTOR: {
     Register Src = MI.getOperand(1).getReg();
     LLT SrcType = MRI.getType(Src);
-    // We support concat for vector sizes greater than 128 bits.
-    if (SrcType.getSizeInBits() >= 128) {
+    // We support concat for vector sizes greater than 128 bits as well as those
+    // of 32 bits.
+    unsigned SrcSize = SrcType.getSizeInBits();
```
andcarminati: nit: `const SrcSize..``

clang/lib/Headers/aie2p_upd_ext.h
```diff
@@ -1920,4 +1920,177 @@ INTRINSIC(v16acc64) set_v16acc64(int idx, v4acc64 b) {
...
+    return (int)__builtin_shufflevector(exp, exp, 0, 1, 2, 3);
+  return (int)__builtin_shufflevector(exp, exp, 4, 5, 6, 7);
+}
+
+INTRINSIC(v64bfp16ebs16) insert(v64bfp16ebs16 v, int idx, int exp32) {
```
niwinanto, line 1938: Same as above, may be
```
  v2int32 exp64 = v.exponent;
  exp64[idx] = exp32;
  return {v.mantissa, exp64};
```

clang/lib/Headers/aie2p_upd_ext.h
khallouh: it is just 0 or 1 (e.g. either el0 or eh0) and yes we could do what you suggested. It is even nicer, I will update.

clang/lib/Headers/aie2p_upd_ext.h
```diff
@@ -1920,4 +1920,177 @@ INTRINSIC(v16acc64) set_v16acc64(int idx, v4acc64 b) {
...
+}
+
+INTRINSIC(int) extract_exponent(v128bfp16ebs8 v, int idx) {
+  v8int8 exp0 = v.exponentE0;
+  v8int8 exp1 = v.exponentE1;
+  if (idx == 0)
+    return (int)__builtin_shufflevector(exp0, exp0, 0, 1, 2, 3);
+  else if (idx == 1)
+    return (int)__builtin_shufflevector(exp0, exp0, 4, 5, 6, 7);
+  if (idx == 2)
```
niwinanto: nit: else if

clang/lib/Headers/aie2p_upd_ext.h
```diff
@@ -1920,4 +1920,177 @@ INTRINSIC(v16acc64) set_v16acc64(int idx, v4acc64 b) {
...
+
+INTRINSIC(v128bfp16ebs16)
+insert(v128bfp16ebs16 v, int idx, v64bfp16ebs16 vsub) {
+  if (idx == 0)
+    return {vsub.mantissa, v.mantissaX1, vsub.exponent, v.exponentE1};
+  return {v.mantissaX0, vsub.mantissa, v.exponentE0, vsub.exponent};
+}
+
+INTRINSIC(v128bfp16ebs16) set_v128bfp16ebs16(int idx, v64bfp16ebs16 vsub) {
+  v64bfp16ebs16 a;
```
niwinanto: nit: undefValue

llvm/lib/Target/AIE/aie2p/AIE2PInstrInfo.cpp
```diff
@@ -751,6 +751,16 @@ void AIE2PInstrInfo::copyPhysReg(MachineBasicBlock &MBB,
             TRI.getSubReg(DstReg, AIE2P::sub_l_odd))
         .addReg(TRI.getSubReg(SrcReg, AIE2P::sub_hi_exp),
                 getKillRegState(KillSrc));
+  } else if ((AIE2P::EXPVEC64RegClass.contains(SrcReg)) &&
+             (AIE2P::EXPVEC64RegClass.contains(DstReg))) {
+    BuildMI(MBB, MBBI, DL, get(AIE2P::MOV_alu_mv_mv_mv_e_mv_el_to_el),
```
niwinanto, line 756: This is really strange, why we need separate instructions for all combinations of copy between `el` and `eh`

llvm/test/CodeGen/AIE/aie2p/upd_ext_bfp16.ll
```diff
@@ -0,0 +1,883 @@
...
+;
+; (c) Copyright 2025 Advanced Micro Devices, Inc. or its affiliates
+; RUN: llc < %s -verify-machineinstrs -mtriple=aie2p | FileCheck %s
+
+%struct.v64bfp16ebs16 = type <{ <64 x i8>, <8 x i8> }>
+%struct.v64bfp16ebs8 = type <{ <64 x i8>, <8 x i8> }>
+%struct.v128bfp16ebs8 = type <{ <64 x i8>, <64 x i8>, <8 x i8>, <8 x i8> }>
+%struct.v128bfp16ebs16 = type <{ <64 x i8>, <64 x i8>, <8 x i8>, <8 x i8> }>
+
+define dso_local %struct.v64bfp16ebs16 @_Z11test_insert13v64bfp16ebs16Dv64_a(%struct.v64bfp16ebs16 %v.coerce, <64 x i8> noundef %m) local_unnamed_addr #0 {
```
niwinanto, line 15: Back-end is doing a great job optimizing the ll, few are too good to believe.

clang/test/CodeGen/aie/aie2p/aie2p-upd-ext-intrinsic.cpp
niwinanto, line 1: There are many back-end fails in this test file from one reason to other. But wont be a good idea to fix them along with this PR. How ever, there is a legalizer issue with some newly added intrinsics. It would be nice to fix them.
```
v128bfp16ebs8 test_insert(v128bfp16ebs8 v , int idx , int exp ) {return insert(v,idx,exp);}

v64bfp16ebs16 test_extract_v64bfp16ebs16(v128bfp16ebs16 m, int idx) {
  return extract_v64bfp16ebs16(m, idx);
}

v64bfp16ebs8 test_extract_v64bfp16ebs8(v128bfp16ebs8 m, int idx) {
  return extract_v64bfp16ebs8(m, idx);
}
```
And the legalizer error is
```
fatal error: error in backend: unable to legalize instruction: %41:_(<64 x s8>) = G_CONCAT_VECTORS %6:_(<8 x s8>), %40:_(<8 x s8>), %40:_(<8 x s8>), %40:_(<8 x s8>), %40:_(<8 x s8>), %40:_(<8 x s8>), %40:_(<8 x s8>), %40:_(<8 x s8>) (in function: _Z11test_insert13v128bfp16ebs8ii)
```

khallouh: Yes I missed that. That is created when legalizing G_SELECT. We need custom legalization for 64-bit vector concats. Alternatively we can add a separate custom legalization for 64-bit vector selects using 2 scalar SEL instructions instead of VSEL. I prefer the latter approach. I will add the custom legalizer for G_SELECT.

niwinanto (approved):

PR #463, opened 2025-05-09
Fix Ubuntu 22.04 distro build

jtuyls:
The distro build updated to Ubuntu 22.04 is still failing: https://github.com/Xilinx/llvm-aie/actions/runs/14928516968/job/41938811983

### DIFF
--- a/.github/workflows/amd-aie-distro.yml
+++ b/.github/workflows/amd-aie-distro.yml
@@ -74,7 +74,7 @@ jobs:
       - name: Checkout actions
         uses: actions/checkout@v3
         with:
-          ref: ${{ github.head_ref || github.ref_name }}
+          ref: ${{ github.event.pull_request.head.sha }}
           # check out anything
           sparse-checkout: .github/workflows/amd_aie_releases
           sparse-checkout-cone-mode: false
--- a/.github/workflows/amd-aie-distro.yml
+++ b/.github/workflows/amd-aie-distro.yml
@@ -1,6 +1,7 @@
 name: AMD AIE Distro
 
 on:
+  pull_request:
   workflow_dispatch:
     inputs:
       DEBUG_ENABLED:
@@ -73,7 +74,7 @@ jobs:
       - name: Checkout actions
         uses: actions/checkout@v3
         with:
-          ref: ${{ github.head_ref || github.ref_name }}
+          ref: ${{ github.event.pull_request.head.sha }}
           # check out anything
           sparse-checkout: .github/workflows/amd_aie_releases
           sparse-checkout-cone-mode: false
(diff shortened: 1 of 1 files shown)


jtuyls:
@philippjh @konstantinschwarz Could you help review? I think this should fix the remaining issues with the Ubuntu 22.04 distro build. See the build passing when I adjusted it to run on this PR earlier: https://github.com/Xilinx/llvm-aie/actions/runs/15045348634/job/42286423442

konstantinschwarz:
I see the current build is failing with
```
Deleting the contents of '/scratch/ghe-runners/1/_work/llvm-aie/llvm-aie'
Error: File was unable to be removed Error: EACCES: permission denied, unlink '/scratch/ghe-runners/1/_work/llvm-aie/llvm-aie/.ci/generate-buildkite-pipeline-premerge'
```

Do you know if this is explicitly fixed by upgrading to checkout@v4?

philippjh:
> Error: File was unable to be removed Error: EACCES: permission denied, unlink '/scratch/ghe-runners/1/_work/llvm-aie/llvm-aie/.ci/generate-buildkite-pipeline-premerge'

I'll later check with the runner host as we're building on different machines, which then may exhibit different issues.

jtuyls:
> Error: File was unable to be removed Error: EACCES: permission denied



> I see the current build is failing with
> 
> ```
> Deleting the contents of '/scratch/ghe-runners/1/_work/llvm-aie/llvm-aie'
> Error: File was unable to be removed Error: EACCES: permission denied, unlink '/scratch/ghe-runners/1/_work/llvm-aie/llvm-aie/.ci/generate-buildkite-pipeline-premerge'
> ```
> 
> Do you know if this is explicitly fixed by upgrading to checkout@v4?

When I ran it on the PR it did: https://github.com/Xilinx/llvm-aie/actions/runs/15045348634/job/42286423442. I am not a 100% sure it will work as well on main 'push', but I don't see what the difference would be. Based on the successful build on the PR, I think it would be a good try and in any case, it would be good to bump the checkout versions. Also, it doesn't seem to be some flaky machine issue as the main commits have been failing for a while now.

jtuyls:
I reverted to changes to test whether running the distro build on the PR with 'checkout@v3' results in the same issue as on main.

jtuyls:
> I reverted to changes to test whether running the distro build on the PR with 'checkout@v3' results in the same issue as on main.

Ok, it seems like it doesn't show the same issue as on main so this is likely not a fix..

philippjh:
> Ok, it seems like it doesn't show the same issue as on main so this is likely not a fix..

`Permission denied` was the issue you were initially seeing, correct?

philippjh:
I've repaired the permissions on the runner machines, but I suspect that the problem will come up again when a workflow is started that mounts the workspace into his container altering directory and file ownerships. A permanent solution would be to streamline all workflows to operate in a container (like the other build workflow do).

philippjh:
After checking the workflows, we will move the distro build to public runner infrastructure for now (see #470). Updating that particular workflow to support our private runner infra is on my personal backlog.

PR #465, opened 2025-05-13
[AIE2P] Support Wide BFP16 load/store intrinsics

khallouh:


clang/lib/Headers/aie2p_ldst.h
```diff
@@ -225,6 +225,70 @@ INTRINSIC(v128uint8) unpack(v128uint4 v) { return unpack(v, __SIGN_UNSIGNED); }
   FIFO_ST_FLUSH_BARE(T##_unaligned, _bare, DM_BANK, RESTRICT)                  \
   FIFO_ST_FLUSH_CONV(T##_unaligned, DM_BANK, RESTRICT)
 
+#define FIFO_ST_PUSH_BFP16_WIDE(T1, T2, DM_BANK, RESTRICT)                     \
+  INTRINSIC(void)                                                              \
+  fifo_st_reset(T1##_unaligned DM_BANK *RESTRICT &p, T1 v, fifo_state_t &s) {  \
+    T2##_unaligned DM_BANK *RESTRICT &q =                                      \
```
niwinanto, line 231: check: why we do `&q = p`, cant we use p directly?

khallouh: We have to cast it from T1 to T2 because we call the intrinsic using T2 (the narrow type)

clang/lib/Headers/aie2p_ldst.h
```diff
@@ -225,6 +225,70 @@ INTRINSIC(v128uint8) unpack(v128uint4 v) { return unpack(v, __SIGN_UNSIGNED); }
...
+  }                                                                            \
+  INTRINSIC(void)                                                              \
+  fifo_st_flush##VAR##_3d_byte(T1##_unaligned DM_BANK *RESTRICT &p,            \
+                               fifo_state_t &s, int off, int size1,            \
+                               addr_t &count1, int inc1, int size2,            \
+                               addr_t &count2, int inc2) {                     \
+    T2##_unaligned DM_BANK *RESTRICT &q =                                      \
+        (T2##_unaligned DM_BANK * RESTRICT &)p;                                \
+    fifo_st_flush(q, s);                                                       \
+    fifo_st_flush_3d_byte(q, s, off, size1, count1, inc1, size2, count2,       \
```
niwinanto, line 282: Why we just need to use `fifo_st_flush_3d_byte`  once?

clang/lib/Headers/aie2p_ldst.h
niwinanto: sounds good, I thought there will be an implicit cast.

niwinanto: One more check: Do we need to copy back q to p, since they are references.

clang/lib/Headers/aie2p_ldst.h
khallouh: Because the 3d post-increment needs to happen after the last access (the second one) I guess

niwinanto: As we discussed, it is clear to me the pointer updates with flush and pop instructions. Would be nice if we could reason pointer updates with fill.

clang/lib/Headers/aie2p_ldst.h
khallouh: You're right. We can remove that copy.

niwinanto (approved):
LGTM! Nice work

PR #466, opened 2025-05-13
[AIE] Set the bar: add a 14 stage, II=1 pipeline schedule

martien-de-jong:
Something we don't want to lose


andcarminati:
nit: change commit message to [AIE2P] ... or even [AIEX]

andcarminati (approved):
LGTM.

PR #467, opened 2025-05-13
[AIE2P][AIE2PS] Support all 64-bit-destination G_BUILD_VECTOR in legalizer

khallouh:


llvm/test/CodeGen/AIE/aie2p/GlobalIsel/legalize-shuffle-vector-64.mir
```diff
@@ -24,3 +24,72 @@ body: |
...
+    ; CHECK-NEXT: [[AIE_ADD_VECTOR_ELT_HI6:%[0-9]+]]:_(<64 x s8>) = G_AIE_ADD_VECTOR_ELT_HI [[AIE_ADD_VECTOR_ELT_HI5]], [[ASSERT_SEXT6]](s32)
+    ; CHECK-NEXT: [[AIE_ADD_VECTOR_ELT_HI7:%[0-9]+]]:_(<64 x s8>) = G_AIE_ADD_VECTOR_ELT_HI [[AIE_ADD_VECTOR_ELT_HI6]], [[ASSERT_SEXT7]](s32)
+    ; CHECK-NEXT: [[C8:%[0-9]+]]:_(s32) = G_CONSTANT i32 0
+    ; CHECK-NEXT: [[BITCAST:%[0-9]+]]:_(<8 x s64>) = G_BITCAST [[AIE_ADD_VECTOR_ELT_HI7]](<64 x s8>)
+    ; CHECK-NEXT: [[AIE_SEXT_EXTRACT_VECTOR_ELT8:%[0-9]+]]:_(s64) = G_AIE_SEXT_EXTRACT_VECTOR_ELT [[BITCAST]](<8 x s64>), [[C8]](s32)
+    ; CHECK-NEXT: [[BITCAST1:%[0-9]+]]:_(<8 x s8>) = G_BITCAST [[AIE_SEXT_EXTRACT_VECTOR_ELT8]](s64)
+    ; CHECK-NEXT: PseudoRET implicit $lr, implicit [[BITCAST1]](<8 x s8>)
+    %0:_(<32 x s8>) = COPY $wh0
+    %143:_(<32 x s8>) = COPY $wh0
+    %2:_(<8 x s8>) = G_SHUFFLE_VECTOR %143:_(<32 x s8>), %143:_, shufflemask(24, 25, 26, 27, 28, 29, 30, 31)
```
andcarminati: nit: maybe use a mask that cannot be optimized by the pre-legalizer combiner in a real scenario. For example `0, 25, 1, 27, 2, 29, 3, 31`.

llvm/test/CodeGen/AIE/aie2p/buildvector.ll
```diff
@@ -83,3 +83,23 @@ entry:
...
+define void @test_buildvector_64bits() {
+; CHECK-LABEL: test_buildvector_64bits:
+; CHECK:         .p2align 4
+; CHECK-NEXT:  // %bb.0: // %entry
+; CHECK-NEXT:    mova r0, #777; nopx
+; CHECK-NEXT:    vpush.hi.16 x0, x0, r0
+; CHECK-NEXT:    vpush.hi.16 x0, x0, r0
+; CHECK-NEXT:    vpush.hi.16 x0, x0, r0
+; CHECK-NEXT:    vpush.hi.16 x0, x0, r0
+; CHECK-NEXT:    vextract.64 r1:r0, x0, #0, vaddsign1
```
andcarminati, line 89: Super follow-up: can we combine this as a VBCST.16 + VEXTRACT.64?

khallouh: Yes, I used that pattern on purpose

llvm/test/CodeGen/AIE/aie2p/GlobalIsel/legalize-shuffle-vector-64.mir
khallouh: Yes and we can keep both. We should be able to combine this later.

llvm/test/CodeGen/AIE/aie2p/GlobalIsel/legalize-build-vector.mir
```diff
@@ -180,3 +180,40 @@ body:             |
     $bmll0 = COPY %1(<8 x s64>)
     PseudoRET implicit $lr, implicit $bmll0
 ...
+
+
+---
+name:            test_build_vector_v8s8
```
niwinanto, line 187: Please include legalizer tests for 16 and 32 bit element type cases.

llvm/lib/Target/AIE/AIELegalizerHelper.cpp
```diff
@@ -263,6 +263,16 @@ bool AIELegalizerHelper::legalizeG_BUILD_VECTOR(LegalizerHelper &Helper,
                       : MIRBuilder.buildBitcast(VecTy, Vec512Reg).getReg(0);
     MIRBuilder.buildInstr(II->getGenericUnpadVectorOpcode(), {DstReg},
                           {NewSrc2});
+  } else if (DstVecSize == 64) {
```
niwinanto, line 266: I started to feel like we are over using vector registers, may be we could consider keep it in the scalar registers itself.

llvm/lib/Target/AIE/aie2p/AIE2PLegalizerInfo.cpp
```diff
@@ -65,11 +65,11 @@ isValidVectorMergeUnmergeOp(const unsigned BigVectorId,
   };
 }
 
-static LegalityPredicate isValidVectorAIEP(const unsigned TypeIdx) {
+static LegalityPredicate isValidVectorAIE2P(const unsigned TypeIdx) {
```
niwinanto: It is used for other opcodes as well, we should add tests.

llvm/test/CodeGen/AIE/aie2p/GlobalIsel/legalize-build-vector.mir
khallouh: We should already have the one for v232 since it is already legal before my change. I already added the test for 4x16 in the end-2-end test buildvector.ll but I can add it here as well if needed.

llvm/lib/Target/AIE/AIELegalizerHelper.cpp
khallouh: Yes, I already tried the pack approach using shifts but there were more instructions than passing through the 512-bit vector. I don't mind either approach, both are pretty expensive without any pattern to combine.

llvm/lib/Target/AIE/aie2p/AIE2PLegalizerInfo.cpp
khallouh: I don't like the name actually. We were excluding 64-bits as invalid AIE2P vectors when it was just a question of missing support. For instance, we have legal vector types for all those sizes from 32-bits and upwards.
I would rather keep this as is and use another predicate for the other opcodes, or simply include the previous predicate directly since there is only one other action using that (for G_UNMERGE_VALUES)

llvm/lib/Target/AIE/AIELegalizerHelper.cpp
```diff
@@ -193,8 +193,8 @@ bool AIELegalizerHelper::legalizeG_BUILD_VECTOR(LegalizerHelper &Helper,
 
   assert((EltSize == 8 || EltSize == 16 || EltSize == 32 || EltSize == 64) &&
          "non-existent integer size");
-  assert(DstVecSize == 32 || (DstVecSize > 64 && DstVecSize <= 1024 &&
-                              "non-native vectors are not supported"));
+  assert(DstVecSize >= 32 && DstVecSize <= 1024 &&
+         "non-native vectors are not supported");
   assert(DstVecSize < 1024 && "vadd takes a 512-bit argument");
```
martien-de-jong: nit: vadd vs BUILD_VECTOR?
nit: both DstVecSize <= 1024 and DstVecSize < 1024? can they be combined?

llvm/lib/Target/AIE/aie2p/AIE2PLegalizerInfo.cpp
khallouh: I've split them and renamed the old one to unmerge since it's the only user left

llvm/lib/Target/AIE/AIELegalizerHelper.cpp
khallouh: I'm not sure where the vadd comes from in that other assert. That is misleading since we never lower to vadd but rather to `G_AIE_ADD_VECTOR_ELT_HI` which is different. I've combined the 2 asserts and made the message clearer

khallouh:
Closing in favor for https://github.com/Xilinx/llvm-aie/pull/1184

PR #468, opened 2025-05-14
[AIEX] Add proper deps between event instructions fixed SUs

andcarminati:
This PR prevents event instructions from being incorrectly placed in prologues or epilogues of pipelined loops.


llvm/lib/Target/AIE/AIEBaseSubtarget.cpp
```diff
@@ -30,6 +30,7 @@
 #include "llvm/CodeGen/ScheduleDAGMutation.h"
 #include "llvm/MC/MCRegisterInfo.h"
 #include "llvm/Support/Debug.h"
+#include <algorithm>
```
martien-de-jong: nit: llvm::any_of doesn't come from this file

andcarminati: Updated!

llvm/lib/Target/AIE/AIEBaseSubtarget.cpp
```diff
@@ -392,8 +392,46 @@ class EmitFixedSUnits : public ScheduleDAGMutation {
...
+
     for (SUnit &FreeSU : make_filter_range(DAG->SUnits, IsFreeSU)) {
       const MachineInstr &MI = *FreeSU.getInstr();
+
+      auto UseOrDefReg = [](const MachineInstr &MI) {
+        return llvm::any_of(
+            MI.operands(), [](const MachineOperand &MO) { return MO.isReg(); });
+      };
+
+      if (MI.hasUnmodeledSideEffects() && !MI.mayLoadOrStore() &&
```
niwinanto, line 406: check: like `isLock`, is it good to have, maybe `isEvent` more easy and readable?

llvm/lib/Target/AIE/AIEBaseSubtarget.cpp
```diff
@@ -392,8 +392,46 @@ class EmitFixedSUnits : public ScheduleDAGMutation {
...
+        if (!TopFixedBundles.empty()) {
+          SUnit *FixedDepSU = DAG->getSUnit(&*getBundleStart(
+              TopFixedBundles.back().getInstrs().front()->getIterator()));
+          SDep Dep(FixedDepSU, SDep::Artificial);
+          Dep.setLatency(1);
+          FreeSU.addPred(Dep, /*Required=*/true);
+        }
+        continue;
+      }
+
```
niwinanto, line 434: Since you are fixing this, could you update the latency for lock in the epilog merge case.

llvm/lib/Target/AIE/AIEBaseSubtarget.cpp
```diff
@@ -392,8 +392,46 @@ class EmitFixedSUnits : public ScheduleDAGMutation {
...
     for (SUnit &FreeSU : make_filter_range(DAG->SUnits, IsFreeSU)) {
       const MachineInstr &MI = *FreeSU.getInstr();
+
+      auto UseOrDefReg = [](const MachineInstr &MI) {
+        return llvm::any_of(
+            MI.operands(), [](const MachineOperand &MO) { return MO.isReg(); });
+      };
+
+      if (MI.hasUnmodeledSideEffects() && !MI.mayLoadOrStore() &&
+          !TII->isLock(MI.getOpcode()) && !UseOrDefReg(MI)) {
```
martien-de-jong, line 407: nit: I think we can justify the edge on the first condition alone. We don't know whether the other edges induce the correct latency. @niwinanto I think that any unmodeled side effect should have at least an edge. isEvent is just an example instance of it. What's generally missing is a method to find the proper latency. Note for instance that we may still see an event without a preceding store having taken effect -- If that is important you should have used a lock. This one edge just makes sure we don't drift too far, and the store will happen "around this time".

llvm/test/CodeGen/AIE/aie2p/schedule/events-and-pos-swp.mir
martien-de-jong, line 1: postswp?

llvm/test/CodeGen/AIE/aie2p/schedule/events-and-pos-swp.mir
```diff
@@ -14,44 +14,50 @@
...
-  ; CHECK-NEXT:    vst.srs.2x cml0, s0, srssign1, [p1], #64; event #1
-  ; CHECK-NEXT:    vst.srs.2x cml0, s0, srssign1, [p1], #64; event #0
-  ; CHECK-NEXT:    vst.srs.2x cml0, s0, srssign1, [p1], #64; event #1
   ; CHECK-NEXT:    vst.srs.2x cml0, s0, srssign1, [p1], #64
   ; CHECK-NEXT:    vst.srs.2x cml0, s0, srssign1, [p1], #64
+  ; CHECK-NEXT:    vst.srs.2x cml0, s0, srssign1, [p1], #64
+  ; CHECK-NEXT:    vst.srs.2x cml0, s0, srssign1, [p1], #64
+  ; CHECK-NEXT:    vst.srs.2x cml0, s0, srssign1, [p1], #64
+  ; CHECK-NEXT:    vst.srs.2x cml0, s0, srssign1, [p1], #64
+  ; CHECK-NEXT:    event #1
```
martien-de-jong: Why doesn't it fill the first delay slot? Is it included in the top-down pass?

llvm/lib/Target/AIE/AIEBaseSubtarget.cpp
martien-de-jong: In contrast, IsLock carries very specific information about the interaction with memory events.

martien-de-jong (dismissed):
Just a few nits

llvm/lib/Target/AIE/AIEBaseSubtarget.cpp
andcarminati: This is more general, we are interested in instructions with side-effects and without memory cycles or data cycles. We are implementing this for events, but I would prefer to not another target hook that we could forget about.

llvm/lib/Target/AIE/AIEBaseSubtarget.cpp
andcarminati: I think this could be implemented as a separated PR, as here we are mainly interested in event-like instructions.

llvm/test/CodeGen/AIE/aie2p/schedule/events-and-pos-swp.mir
andcarminati: Yes, it was picked up by top-down before switching to bottom-up.

martien-de-jong (approved):

PR #469, opened 2025-05-14
[AIEX] Add FuncArgCopyEdges dag mutator in pre-scheduler

khallouh:
Some instructions take a ~~singleton register~~ class in their operands. If a copy of that same physical register from the function arguments is moved closer to its user by the pre-scheduler and it happens to move past an instruction requiring that singleton register class, a live range interference is created which the register allocator cannot resolve by evicting the function argument physical register, leading to a "ran out of registers" error. 
To solve this, this introduces DAG mutator in the pre-scheduler with zero latency edge between such copies and such instructions, thus preventing the scheduler from creating the undesired interference.

**Edit**: it's not just singleton classes but also other small classes that overlap with the first registers used in the function arguments. For example ePSRFLdf which only has 2 registers with $p0 and $p1 as subregisters, which also happen to be the first registers used to pass pointers. Most other "scarce/small" registers classes like mReF2a ($r24) or `mReS16m` ($r16) either cannot be used to pass arguments or would require a lot of arguments to be used, so this issue only manifests with pointers and fifo loading instructions.


llvm/lib/Target/AIE/AIEBaseSubtarget.cpp
```diff
@@ -251,6 +251,47 @@ class BiasDepth : public ScheduleDAGMutation {
...
+    const MachineRegisterInfo &MRI = DAG->MRI;
+    MachineBasicBlock *MBB = DAG->getBB();
+
+    for (SUnit &SU : DAG->SUnits) {
+      MachineInstr &MI = *SU.getInstr();
+      if (!MI.isCopy())
+        continue;
+
+      auto MO = MI.getOperand(1);
+      if (!(MO.isReg() && MO.getReg().isPhysical()))
```
andcarminati: nit:
```      
if (!MO.isReg() || !MO.getReg().isPhysical())
       continue;
```

llvm/lib/Target/AIE/AIEBaseSubtarget.cpp
```diff
@@ -251,6 +251,47 @@ class BiasDepth : public ScheduleDAGMutation {
...
+        continue;
+
+      auto PhysReg = MO.getReg();
+      if (!MBB->isLiveIn(PhysReg))
+        continue;
+
+      SDep SingletonCopyEdge(&SU, SDep::Artificial);
```
andcarminati: What about moving this into the scope of the `if` below?

llvm/lib/Target/AIE/AIEBaseSubtarget.cpp
```diff
@@ -251,6 +251,47 @@ class BiasDepth : public ScheduleDAGMutation {
...
+
+      for (SUnit &SU2 : DAG->SUnits) {
+        MachineInstr &MI = *SU2.getInstr();
+
```
andcarminati: Are we creating an Edge from SU to itself?

andcarminati:
Can we have a test that runs only premisched?

llvm/lib/Target/AIE/AIEBaseSubtarget.cpp
khallouh: The physical reg copy destination would have to be constrained to same singleton class which might happen. I should return early for the predecessor SU

khallouh:
> Can we have a test that runs only premisched?


Done

llvm/lib/Target/AIE/AIEBaseSubtarget.cpp
khallouh: Done

khallouh:
QoR report
image

llvm/lib/Target/AIE/AIEBaseSubtarget.cpp
```diff
@@ -256,6 +256,78 @@ class BiasDepth : public ScheduleDAGMutation {
...
+  /// of any register in RC.
+  static bool containsRegOrSubReg(const TargetRegisterClass *RC,
+                                  MCPhysReg PhysReg,
+                                  const TargetRegisterInfo *TRI) {
+    // Direct containment check
+    if (RC->contains(PhysReg))
+      return true;
+
+    // Check if PhysReg is a subregister of any register in RC
+    for (MCPhysReg SuperReg : *RC) {
```
F-Stuckmann: nit: we do not check if a subregister contains PhysReg 
( it may not be needed)

We could simplify this function to: 
```
 for (MCPhysReg ClassReg : *RC)
    if (TRI->regsOverlap(ClassReg, PhysReg))
      return true;
```

llvm/lib/Target/AIE/AIEBaseSubtarget.cpp
```diff
@@ -256,6 +256,78 @@ class BiasDepth : public ScheduleDAGMutation {
...
+
+      auto MO = MI.getOperand(1);
+      if (!MO.isReg() || !MO.getReg().isPhysical())
+        continue;
+
+      auto PhysReg = MO.getReg();
+      if (!MBB->isLiveIn(PhysReg))
+        continue;
+
+      for (SUnit &SU2 : DAG->SUnits) {
```
F-Stuckmann: nit: didn't we already compare every SU pair before SU ? 
I think we can tighten the for loop ranges

llvm/lib/Target/AIE/AIEBaseSubtarget.cpp
```diff
@@ -256,6 +256,78 @@ class BiasDepth : public ScheduleDAGMutation {
...
+        if (any_of(MI2.defs(),
+                   [&MRI, TRI, PhysReg](const MachineOperand &DefMO) {
+                     if (!DefMO.isReg() || DefMO.getReg().isPhysical())
+                       return false;
+                     auto RC = MRI.getRegClass(DefMO.getReg());
+                     // A def with a small register class where PhysReg is
+                     // contained either directly or as a subregister of a class
+                     // member. This handles composite register classes like
+                     // ePSRFLdF where p0 is a subregister of plfr0.
+                     if (RC->getNumRegs() <= 2 &&
```
F-Stuckmann: Why are we not checking if we are handling a Singleton Register?

llvm/lib/Target/AIE/AIEBaseSubtarget.cpp
```diff
@@ -256,6 +256,78 @@ class BiasDepth : public ScheduleDAGMutation {
...
+                         containsRegOrSubReg(RC, PhysReg, TRI)) {
+                       LLVM_DEBUG(dbgs() << "FuncArgCopyEdges: PhysReg "
+                                         << printReg(PhysReg, TRI)
+                                         << " conflicts with "
+                                         << TRI->getRegClassName(RC) << "\n");
+                       return true;
+                     }
+                     return false;
+                   })) {
+          if (!DAG->canAddEdge(&SU2, &SU))
```
F-Stuckmann: If we only ever go from SU_{low_index} to SU_{high_index} i think we are always able to add the Edge.
We could possibly convert this to an assert

llvm/lib/Target/AIE/AIEBaseSubtarget.cpp
F-Stuckmann: this loop is quadratic, you may want to really limit the range of the innerloop

F-Stuckmann: can we identify Singleton Registers and walk the defs and uses?

llvm/lib/Target/AIE/AIEBaseSubtarget.cpp
```diff
@@ -256,6 +256,78 @@ class BiasDepth : public ScheduleDAGMutation {
...
+    const TargetRegisterInfo *TRI = MRI.getTargetRegisterInfo();
+    MachineBasicBlock *MBB = DAG->getBB();
+
+    for (SUnit &SU : DAG->SUnits) {
+      MachineInstr &MI = *SU.getInstr();
+      if (!MI.isCopy())
+        continue;
+
+      auto MO = MI.getOperand(1);
+      if (!MO.isReg() || !MO.getReg().isPhysical())
```
andcarminati: Could we give a name to check -> lambda? This can be reused next.

andcarminati:
Maybe we could make this a bit more self-documented, something like:
```
class FuncArgCopyEdges : public ScheduleDAGMutation {
  /// Check if PhysReg is contained in RC, either directly or as a subregister
  /// of any register in RC.
  static bool containsRegOrSubReg(const TargetRegisterClass *RC,
                                  MCPhysReg PhysReg,
                                  const TargetRegisterInfo *TRI) {
    // Direct containment check
    if (RC->contains(PhysReg))
      return true;

    // Check if PhysReg is a subregister of any register in RC
    for (MCPhysReg SuperReg : *RC) {
      if (TRI->isSubRegisterEq(SuperReg, PhysReg))
        return true;
    }
    return false;
  }

  /// Return true if SU2 defines a virtual register in a small register class
  /// that contains PhysReg (directly or as a sub-register of a class member),
  /// and SU2 is not the reference COPY SUnit itself.
  static bool conflictsWithPhysReg(const MachineRegisterInfo &MRI,
                                   const TargetRegisterInfo *TRI,
                                   MCPhysReg PhysReg,
                                   const SUnit &ReferenceSU,
                                   const SUnit &SU2) {
    if (&SU2 == &ReferenceSU)
      return false;
    return any_of(SU2.getInstr()->defs(),
                  [&MRI, TRI, PhysReg](const MachineOperand &DefMO) {
                    if (!DefMO.isReg() || DefMO.getReg().isPhysical())
                      return false;
                    const TargetRegisterClass *RC =
                        MRI.getRegClass(DefMO.getReg());
                    // A def with a small register class where PhysReg is
                    // contained either directly or as a subregister of a class
                    // member. This handles composite register classes like
                    // ePSRFLdF where p0 is a subregister of plfr0.
                    if (RC->getNumRegs() <= 2 &&
                        containsRegOrSubReg(RC, PhysReg, TRI)) {
                      LLVM_DEBUG(dbgs() << "FuncArgCopyEdges: PhysReg "
                                        << printReg(PhysReg, TRI)
                                        << " conflicts with "
                                        << TRI->getRegClassName(RC) << "\n");
                      return true;
                    }
                    return false;
                  });
  }

    const MachineRegisterInfo &MRI = DAG->MRI;
    const TargetRegisterInfo *TRI = MRI.getTargetRegisterInfo();

    // Filter: only COPYs whose source is a live-in physical register.
    auto IsFuncArgCopy = [&MBB](const SUnit &SU) {
      const MachineInstr &MI = *SU.getInstr();
      if (!MI.isCopy())
        return false;
      const MachineOperand &MO = MI.getOperand(1);
      return MO.isReg() && MO.getReg().isPhysical() &&
             MBB->isLiveIn(MO.getReg());
    };

    for (SUnit &SU : make_filter_range(DAG->SUnits, IsFuncArgCopy)) {
      MCPhysReg PhysReg = SU.getInstr()->getOperand(1).getReg();

      for (SUnit &SU2 : make_filter_range(DAG->SUnits, [&](const SUnit &SU2) {
               return conflictsWithPhysReg(MRI, TRI, PhysReg, SU, SU2);
             })) {
        if (!DAG->canAddEdge(&SU2, &SU))
          continue;
        LLVM_DEBUG(dbgs() << "FuncArgCopyEdges: Adding edge from COPY of "
                          << printReg(PhysReg, TRI) << " to "
                          << *SU2.getInstr());
        SDep SingletonCopyEdge(&SU, SDep::Artificial);
        SingletonCopyEdge.setLatency(0);
        SU2.addPred(SingletonCopyEdge, /*Required=*/true);
      }
    }
  };
};

```

llvm/lib/Target/AIE/AIEBaseSubtarget.cpp
khallouh: Done

llvm/lib/Target/AIE/AIEBaseSubtarget.cpp
khallouh: I had already renamed the PR to remove mention of singletons, this `RC->getNumRegs() <= 2` also handles classes with 2 registers since we hit this recently. I made the comment clearer.

llvm/lib/Target/AIE/AIEBaseSubtarget.cpp
khallouh: Agreed

khallouh:
When two COPYs define in the same small class (e.g. ePS = {p0, p1}), the mutator adds edges in both directions creating a cycle (this was actually causing conv2d_bfp16_kernel_red.ll to hang). The fix: skip func-arg COPYs in the target loop

llvm/test/CodeGen/AIE/aie2ps/schedule/function-arg-copies-mut.mir
```diff
@@ -0,0 +1,53 @@
+# NOTE: Assertions have been autogenerated by utils/update_mir_test_checks.py UTC_ARGS: --version 5
+#
+# This file is licensed under the Apache License v2.0 with LLVM Exceptions.
+# See https://llvm.org/LICENSE.txt for license information.
+# SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+#
+# (c) Copyright 2026 Advanced Micro Devices, Inc. or its affiliates
+# RUN: llc -O2 -mtriple=aie2ps -run-pass=machine-scheduler %s -o - | FileCheck %s
+
+---
```
F-Stuckmann, line 15: nit: can we have a comment describing what the file tests?

llvm/lib/Target/AIE/AIEBaseSubtarget.cpp
```diff
@@ -268,6 +273,89 @@ class BiasDepth : public ScheduleDAGMutation {
...
+/// that the register allocator cannot resolve.
+class FuncArgCopyEdges : public ScheduleDAGMutation {
+  /// Return true if \p SU defines a virtual register in a small register class
+  /// (at most 2 registers) that overlaps with \p PhysReg.
+  static bool conflictsWithPhysReg(const MachineRegisterInfo &MRI,
+                                   const TargetRegisterInfo *TRI,
+                                   MCPhysReg PhysReg, const SUnit &SU) {
+    return any_of(SU.getInstr()->defs(),
+                  [&MRI, TRI, PhysReg](const MachineOperand &DefMO) {
+                    if (!DefMO.isReg() || DefMO.getReg().isPhysical())
```
F-Stuckmann, line 289: nit: why do we skip over physical register?

llvm/lib/Target/AIE/AIEBaseSubtarget.cpp
```diff
@@ -268,6 +273,89 @@ class BiasDepth : public ScheduleDAGMutation {
...
+      const MachineInstr &MI = *SU.getInstr();
+      if (!MI.isCopy())
+        continue;
+      const MachineOperand &SrcMO = MI.getOperand(1);
+      if (!SrcMO.isReg() || !SrcMO.getReg().isPhysical())
+        continue;
+      MCPhysReg PhysReg = SrcMO.getReg().asMCReg();
+      if (!MBB->isLiveIn(PhysReg))
+        continue;
+      FuncArgCopies.emplace_back(&SU, PhysReg);
```
F-Stuckmann, line 326: nit: Aren't we collecting all the Livein PhysReg Copies, i.e. we are not only checking Function Arguments? Since we are not limited to the entry MBB, I think you should rename FuncArgCopies to something meaningful.

llvm/lib/Target/AIE/AIEBaseSubtarget.cpp
```diff
@@ -268,6 +273,89 @@ class BiasDepth : public ScheduleDAGMutation {
...
+      FuncArgCopySUs.insert(&SU);
+    }
+
+    if (FuncArgCopies.empty())
+      return;
+
+    // Second pass: for each non-COPY SUnit, check if it has a def conflicting
+    // with any of the collected function argument registers. This is O(N * K)
```
F-Stuckmann, line 334: nit: since we are more broad than function arguments, could you please harmonize the comments and update the PR title?

llvm/lib/Target/AIE/AIEBaseSubtarget.cpp
```diff
@@ -268,6 +273,89 @@ class BiasDepth : public ScheduleDAGMutation {
...
+    // where K is the number of func-arg COPYs (typically very small).
+    // We skip other func-arg COPYs: two COPYs of function arguments into
+    // overlapping small register classes (e.g. ePS = {p0, p1}) would create
+    // edges in both directions, forming a cycle.
+    for (SUnit &SU : DAG->SUnits) {
```
F-Stuckmann, line 339: nit: If we either way skip over any SU that is not in `FuncArgCopySUs`, we can just iterate over `FuncArgCopySUs`

llvm/lib/Target/AIE/AIEBaseSubtarget.cpp
```diff
@@ -268,6 +273,89 @@ class BiasDepth : public ScheduleDAGMutation {
...
+      for (auto &[CopySU, PhysReg] : FuncArgCopies) {
+        if (!conflictsWithPhysReg(MRI, TRI, PhysReg, SU))
+          continue;
+        assert(DAG->canAddEdge(&SU, CopySU) &&
+               "Unexpected cycle: func-arg COPY must be schedulable before "
+               "the conflicting instruction");
+        LLVM_DEBUG(dbgs() << "FuncArgCopyEdges: Adding edge from COPY of "
+                          << printReg(PhysReg, TRI) << " to "
+                          << *SU.getInstr());
+        SDep Dep(CopySU, SDep::Artificial);
```
F-Stuckmann, line 351: CHECK: We add a loop back towards the same SU, right? 
Don't we now create a cycle?

Could you check if having the tight reg class copies twice inside the same MBB we do not crash?

llvm/lib/Target/AIE/AIEBaseSubtarget.cpp
khallouh: The LiveIn PhysReg are argument copies since we are adding this in the prescheduler so before RA

llvm/lib/Target/AIE/AIEBaseSubtarget.cpp
khallouh: Because we don't need to process them as they are not allocatable operands

llvm/test/CodeGen/AIE/aie2ps/schedule/function-arg-copies-mut.mir
khallouh: Done, the baseline test fails with "ran out of registers during allocation error". My mutator resolves the error

llvm/lib/Target/AIE/AIEBaseSubtarget.cpp
khallouh: We must visit all other SUs for conflict checks anyway, so what's the benefit?

llvm/lib/Target/AIE/AIEBaseSubtarget.cpp
khallouh: My test already has two func-arg copies into the same tight register class in the same MBB. 
Also, how could we ever create a cycle if we `assert(canAddEdge(&SU, CopySU))` right before?

llvm/lib/Target/AIE/AIEBaseSubtarget.cpp
khallouh: This still handles only func arg copies since we are pre-RA

llvm/lib/Target/AIE/AIEBaseSubtarget.cpp
khallouh: Only the updated test has 2 tight reg class copies. The previous test had only one, sorry.

F-Stuckmann (approved):
Nice work

PR #472, opened 2025-05-21
[AIEX] Outline GEPs

F-Stuckmann:
Currently the IRTranslator inserts G_PTR_ADDs of inlined GEPs at the top of the MBB, instead of right before the inlined-gep  instruction.
So a check has to be performed, if these G_PTR_ADDs can be moved down to the instruction using them first. 
To not have a full movability check in the global combiner PR (https://github.com/Xilinx/llvm-aie/pull/405 ), the behaviour of the IRTranslator is not desired.

This PR outlines GEP if they are in memory Instructions or call instructions as a standalone opt pass.

Note: A better way is to tell the IRTranslator where to insert GEPs.


llvm/lib/Passes/PassBuilderPipelines.cpp
```diff
@@ -312,6 +313,10 @@ static cl::opt<bool> EnableLoopIterCountToAssumptions(
     cl::desc(
         "Enable Conversion of Loop Iteration Count Metadata to Assumptions."));
 
+static cl::opt<bool>
+    EnableReverseGEPInline("enable-outline-gep", cl::Hidden, cl::init(false),
+                           cl::desc("Enable Outlining of GEPs."));
```
martien-de-jong, line 318: nit: The variable name shows its history

llvm/test/Transforms/OutlineGEPInline/outline-gep.ll
```diff
@@ -0,0 +1,42 @@
...
+entry:
+  %0 = load i32, ptr %lp, align 4
+  store i32 %0, ptr @param, align 32
+  %1 = load i32, ptr getelementptr inbounds (%struct.gemm_bfp16_params, ptr @param, i20 0, i32 2), align 4
+  store i32 %1, ptr getelementptr inbounds (%struct.gemm_bfp16_params, ptr @param, i20 0, i32 1), align 4
+  %arrayidx1 = getelementptr inbounds i8, ptr %lp, i20 4
+  %2 = load i32, ptr %arrayidx1, align 4
+  tail call void @llvm.memset.p0.i32(ptr noundef nonnull align 1 dereferenceable(7) getelementptr inbounds (i8, ptr getelementptr inbounds (%struct.gemm_bfp16_params, ptr @param, i20 0, i32 7), i20 13), i8 0, i32 7, i1 false)
+  ret void
+}
```
martien-de-jong, line 42: Also tests for !inbounds and loads and stores without getelemptr, loads and stores with existing getelementptr instruction.

PR #477, opened 2025-05-27
Allow to peel a speculative first stage.

martien-de-jong:
This provides the infrastructure to do speculative pipelines, aka fractional stage counts.

If we find that MinIterCount isn't high enough, we try to find an initial stage that is side-effect free which can run as a preamble to the normal epilogue stages.


llvm/test/CodeGen/AIE/aie2p/schedule/postpipeliner/speculative-first-stage.mir
```diff
@@ -0,0 +1,177 @@
...
+  ; CHECK-NEXT:    nopa ; nopb ; nops ; nopxm ; nopv
+  ; CHECK-NEXT:    nopa ; nopb ; nops ; nopx ; vbcst.32 x0, r0; nopv
+  ; CHECK-NEXT:    .p2align 4
+  ; CHECK-NEXT:  .LBB1_1: // %for.body
+  ; CHECK-NEXT:    // =>This Inner Loop Header: Depth=1
+  ; CHECK-NEXT:    vlda x1, [p0, #64]; nopb ; nops ; nopx ; vshuffle x3, x1, x2, r10; nopv
+  ; CHECK-NEXT:    vlda x2, [p0], #128; nopb ; nops ; nopx ; vmax_lt.32 x0, r16, x0, x3, vaddsign1; nopv
+  ; CHECK-NEXT:    nopa ; nopb ; nops ; nopxm ; nopv
+  ; CHECK-NEXT:  .L_LEnd1:
+  ; CHECK-NEXT:    nopa ; nopb ; nops ; nopxm ; nopv
```
martien-de-jong: we fall back to NS=3 x II=4

llvm/lib/Target/AIE/AIEPostPipeliner.cpp
```diff
@@ -1079,6 +1079,7 @@ bool PostPipeliner::checkStages() {
     return false;
   }
 
+  NPrologueStages = NStages - 1;
```
martien-de-jong, line 1176: This now defines the mode. a SEF peeled loop will have NPrologueStages == NStages

llvm/lib/Target/AIE/AIEInterBlockScheduling.cpp
```diff
@@ -322,9 +328,24 @@ class PipelineExtractor : public PipelineScheduleVisitor {
...
+      }
+      NonEmpty--;
+    }
+    // And copy them.
+    for (auto &B : TimedRegion) {
+      if (NonEmpty > 0) {
+        CopyTo.push_back(B);
+      }
+      NonEmpty--;
+    }
```
martien-de-jong, line 342: This was an existing flaw that was fixed in the epilogue scheduler. The rotated schedule implementing a speculative stage will have leading empty cycles, which are not handled by the prologue scheduler. Better fix the problem at the source rather than downstream.

llvm/test/CodeGen/AIE/aie2/schedule/postpipeliner/round-memdep.mir
```diff
@@ -29,10 +29,7 @@
   ; CHECK-NEXT:    nop // Delay Slot 2
   ; CHECK-NEXT:    nop // Delay Slot 1
   ; CHECK-NEXT:  // %bb.1: // %for.body.preheader
-  ; CHECK-NEXT:    nopx
-  ; CHECK-NEXT:    nop
-  ; CHECK-NEXT:    nop
-  ; CHECK-NEXT:    vlda.ups.s32.s8 cm0, s0, [p0], #32
```
martien-de-jong, line 35: I guess these changed schedules were created bottom up, which also leaves room at the start.

llvm/lib/Target/AIE/AIEPostPipeliner.cpp
```diff
@@ -123,11 +123,35 @@ bool PostPipeliner::isPostPipelineCandidate(MachineBasicBlock &LoopBlock) {
...
+// implicit defs, no stores, and it dooesn't touch anything that is
+// live in to the loop. We explicitly use the fact that out-of-bounds loads do
+// not cause an exception.
+bool isSideEffectFree(MachineInstr *MI) {
+  if (MI->getNumDefs() != 1 || MI->getNumImplicitOperands() != 0 ||
+      MI->mayStore()) {
+    return false;
+  }
+  MachineBasicBlock *MBB = MI->getParent();
+  auto LiveIns = MBB->getLiveIns();
```
martien-de-jong: TODO: this is unused.

llvm/lib/Target/AIE/AIEPostPipeliner.cpp
```diff
@@ -1058,15 +1082,95 @@ bool PostPipeliner::schedule(ScheduleDAGMI &TheDAG, int InitiationInterval,
   LLVM_DEBUG(dbgs() << "PostPipeliner: Success\n");
   return true;
 }
+bool PostPipeliner::hasSufficientMinTripCount() const {
+  return MinTripCount - (NStages - 1) > 0;
+}
+
+// This visitor counts the inital bundles without any side-effect,
```
martien-de-jong: nit: initial

llvm/lib/Target/AIE/AIEPostPipeliner.cpp
```diff
@@ -1058,15 +1082,95 @@ bool PostPipeliner::schedule(ScheduleDAGMI &TheDAG, int InitiationInterval,
...
+// typically only on the first stage.
+class SideEffectFreePeeler : public PipelineScheduleVisitor {
```
martien-de-jong: I guess the name shows its history. It's not peeling, just counting initial SEF bundles.

llvm/lib/Target/AIE/AIEPostPipeliner.cpp
```diff
@@ -1058,15 +1082,95 @@ bool PostPipeliner::schedule(ScheduleDAGMI &TheDAG, int InitiationInterval,
...
+  //
+  // Prolog:
+  //   S-1
+  //   S0 S-1
+  // Loop:
+  //   S1 S0 S-1
+  // Epilog:
+  //      S1
+  //
+  auto LiveIns = DAG->getBB()->getLiveIns();
```
martien-de-jong: TODO: again, not used....

llvm/lib/Target/AIE/AIEPostPipeliner.cpp
```diff
@@ -1058,15 +1082,95 @@ bool PostPipeliner::schedule(ScheduleDAGMI &TheDAG, int InitiationInterval,
...
+  }
+
+  return false;
+}
 
 bool PostPipeliner::checkStages() {
   // We compute the stage in which each representative instruction runs,
   // and take the maximum to decide on the stage count
   NStages = 0;
+  int MaxCycle = 0;
```
martien-de-jong: A later commit removes this again, since it gets inconsistent in the peeling correction. I will squash them.

llvm/lib/Target/AIE/AIEPostPipeliner.cpp
```diff
@@ -1081,51 +1185,54 @@ bool PostPipeliner::checkStages() {
 
   NPrologueStages = NStages - 1;
   // Check that we have a positive trip count after adjusting
-  if (MinTripCount - (NStages - 1) <= 0) {
+  if (!hasSufficientMinTripCount() && !peelSideEffectFree(MaxCycle)) {
```
martien-de-jong: We now have two chances to get it right.

llvm/lib/Target/AIE/AIEPostPipeliner.cpp
```diff
@@ -1081,51 +1185,54 @@ bool PostPipeliner::checkStages() {
...
-          for (int M = 0; M < II; M++) {
-            Visitor.startBundle();
-            for (int K = 0; K < NInstr; K++) {
-              auto &Node = Info[K];
-              if (Filter(Node, Stage, M)) {
-                Visitor.addToBundle(DAG->SUnits[K].getInstr());
-              }
-            }
-            Visitor.endBundle();
-          }
```
martien-de-jong, line 1110: We lifted this lambda to be able to run a visitor on just the first stage.

llvm/lib/Target/AIE/AIEPostPipeliner.h
```diff
@@ -279,11 +290,18 @@ class PostPipeliner {
   // in the Visitor interface object.
   // There are section delimitor methods for prologue, loop, and epilogue
   // end end-of-epilogue.
-  // Between those delimitors, it will call emit() with instructions that need
-  // to be cloned and placed in the appropriate sections. These calls are
-  // bracketed with start and end methods to indicate cycles.
+  // Between those delimitors, it will call addToBundle() with instructions
+  // that need to be cloned and placed in the appropriate sections. These calls
+  // are bracketed with startBundle and endBundle methods to indicate cycles.
```
martien-de-jong, line 318: This comment has been out of date for a while.

llvm/test/CodeGen/AIE/aie2p/schedule/postpipeliner/speculative-first-stage.mir
```diff
@@ -67,26 +67,25 @@
...
   ; CHECK-NEXT:  .LBB1_1: // %for.body
   ; CHECK-NEXT:    // =>This Inner Loop Header: Depth=1
-  ; CHECK-NEXT:    vlda x1, [p0, #64]; nopb ; nops ; nopx ; vshuffle x3, x1, x2, r10; nopv
-  ; CHECK-NEXT:    vlda x2, [p0], #128; nopb ; nops ; nopx ; vmax_lt.32 x0, r16, x0, x3, vaddsign1; nopv
-  ; CHECK-NEXT:    nopa ; nopb ; nops ; nopxm ; nopv
+  ; CHECK-NEXT:    vlda x2, [p0], #128; nopb ; nops ; nopxm ; nopv
+  ; CHECK-NEXT:    nopa ; nopb ; nops ; nopx ; vshuffle x3, x1, x2, r10; nopv
   ; CHECK-NEXT:  .L_LEnd1:
-  ; CHECK-NEXT:    nopa ; nopb ; nops ; nopxm ; nopv
+  ; CHECK-NEXT:    vlda x1, [p0, #64]; nopb ; nops ; nopx ; vmax_lt.32 x0, r16, x0, x3, vaddsign1; nopv
```
martien-de-jong, line 85: Et voila. We have the same lc update, but now run in II=3

llvm/lib/Target/AIE/AIEPostPipeliner.cpp
```diff
@@ -601,7 +601,7 @@ void dumpSchedule(const ScheduleInfo &Info, int MinLength, int II,
     std::string Head = "SU" + std::to_string(K);
     dbgs() << Head;
     for (int I = Head.length() - 6; I < MinLength; I++) {
-      if (I % II == 0) {
+      if (I >= 0 && I % II == 0) {
```
martien-de-jong, line 602: `I` starts out negative and -2 % 2 == 0. This would give a confusing ghost stage boundary.

llvm/lib/Target/AIE/AIEPostPipeliner.cpp
```diff
@@ -123,11 +123,34 @@ bool PostPipeliner::isPostPipelineCandidate(MachineBasicBlock &LoopBlock) {
...
+SlotCounts getSlotCounts(MachineInstr &MI, const AIEBaseInstrInfo *TII) {
+  return SlotCounts{getSlotSet(MI, TII)};
+}
+
+// Our definition of side-effect free. There is one output register, no
+// implicit defs, no stores, and it dooesn't touch anything that is
+// live in to the loop. We explicitly use the fact that out-of-bounds loads do
+// not cause an exception.
+bool isSideEffectFree(MachineInstr *MI) {
+  if (MI->getNumDefs() != 1 || MI->getNumImplicitOperands() != 0 ||
```
martien-de-jong: Multi-def instructions are mostly postincrements, but since we check live-in anyway, I think we can also use multi-def. This would allow things like VMAX_LT, that compute two results, one of which is usually scratch.

llvm/test/CodeGen/AIE/aie2p/schedule/postpipeliner/gemm-bfp16-v10.mir
```diff
@@ -0,0 +1,175 @@
+# NOTE: Assertions have been autogenerated by utils/update_llc_test_checks.py UTC_ARGS: --version 4
+# This file is licensed under the Apache License v2.0 with LLVM Exceptions.
+# See https://llvm.org/LICENSE.txt for license information.
+# SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+#
+# (c) Copyright 2025 Advanced Micro Devices, Inc. or its affiliates
+
+# RUN: llc --mtriple=aie2p -O2 --start-before=postmisched %s \
+# RUN:   --debug-only=postpipeliner-summary -o - | FileCheck %s
```
andcarminati, line 10: nit: machine verifier?

llvm/lib/Target/AIE/AIEInterBlockScheduling.cpp
andcarminati: We should also remove that feature from epilogue scheduling after this change.

llvm/lib/Target/AIE/AIEInterBlockScheduling.cpp
```diff
@@ -322,9 +328,24 @@ class PipelineExtractor : public PipelineScheduleVisitor {
...
   void finish() override {
-    assert(Epilogue->TopInsert.empty() &&
-           "Epilogue already has a timed region at Top.");
-    Epilogue->TopInsert = TimedRegion;
+    auto &CopyTo = Epilogue->TopInsert;
+    assert(CopyTo.empty() && "Epilogue already has a timed region at Top.");
+
+    // Establish the number of bundles to copy.
+    int NonEmpty = TimedRegion.size();
+    for (auto &B : reverse(TimedRegion)) {
```
andcarminati: nit:
```
    int NonEmpty =
        std::distance(std::find_if(TimedRegion.rbegin(), TimedRegion.rend(),
                                   [](auto &B) { return !B.empty(); }),
                      TimedRegion.rend());
```

llvm/lib/Target/AIE/AIEInterBlockScheduling.cpp
```diff
@@ -322,9 +328,24 @@ class PipelineExtractor : public PipelineScheduleVisitor {
...
+      if (!B.empty()) {
+        break;
+      }
+      NonEmpty--;
+    }
+    // And copy them.
+    for (auto &B : TimedRegion) {
```
andcarminati: this could be just:

```
    // And copy them.
    for (int I = 0; I < NonEmpty; I++){
      CopyTo.push_back(TimedRegion[I]);
    }
```

llvm/lib/Target/AIE/AIEPostPipeliner.cpp
```diff
@@ -123,11 +123,32 @@ bool PostPipeliner::isPostPipelineCandidate(MachineBasicBlock &LoopBlock) {
...
+SlotCounts getSlotCounts(MachineInstr &MI, const AIEBaseInstrInfo *TII) {
+  return SlotCounts{getSlotSet(MI, TII)};
+}
+
+// Our definition of side-effect free. There are no implicit defs, no stores
+// and it doesn't touch anything that is live in to the loop.
+// We explicitly use the fact that out-of-bounds loads do not cause an
+// exception.
+bool isSideEffectFree(MachineInstr *MI) {
+  if (MI->getNumImplicitOperands() != 0 || MI->mayStore()) {
```
andcarminati: Should we consider also `hasUnmodeledSideEffects` ?

llvm/lib/Target/AIE/AIEPostPipeliner.cpp
```diff
@@ -123,11 +123,32 @@ bool PostPipeliner::isPostPipelineCandidate(MachineBasicBlock &LoopBlock) {
...
+    return false;
+  }
+  return !any_of(MI->defs(), [MBB = MI->getParent()](MachineOperand &D) {
```
andcarminati: nit: maybe MO instead of D.

llvm/lib/Target/AIE/AIEPostPipeliner.cpp
```diff
@@ -1058,6 +1085,66 @@ bool PostPipeliner::schedule(ScheduleDAGMI &TheDAG, int InitiationInterval,
   LLVM_DEBUG(dbgs() << "PostPipeliner: Success\n");
   return true;
 }
+bool PostPipeliner::hasSufficientMinTripCount() const {
+  return MinTripCount - (NStages - 1) > 0;
+}
+
+// This visitor counts the initial bundles without any side-effect,
+// typically only on the first stage.
+class CountSideEffectFreeBundles : public PipelineScheduleVisitor {
```
andcarminati, line 1098: I like this code, very compact and without a single conditional expression ;-)

llvm/lib/Target/AIE/AIEPostPipeliner.cpp
```diff
@@ -1058,6 +1085,66 @@ bool PostPipeliner::schedule(ScheduleDAGMI &TheDAG, int InitiationInterval,
...
+  CountSideEffectFreeBundles SEFCounter;
+  visitPipelineSection(SEFCounter, 1, [](const NodeInfo &Node, int S, int M) {
+    return Node.Stage == 0 && Node.ModuloCycle == M;
+  });
+
+  const int NSEF = SEFCounter();
+  const int MaxCycle = Info.Length - 1;
+  LLVM_DEBUG(dbgs() << "SEFPeeler: MaxCycle=" << MaxCycle << " NSEF=" << NSEF
+                    << " NStages=" << NStages << "\n");
+  if (MaxCycle < NSEF + NStages * II) {
```
andcarminati: Maybe we can have a comment about this condition.

llvm/lib/Target/AIE/AIEPostPipeliner.cpp
```diff
@@ -1167,4 +1257,19 @@ void NodeInfo::reset(bool FullReset) {
...
 }
 
+void NodeInfo::update(int II) {
+  ModuloCycle = Cycle % II;
+  Stage = Cycle / II;
+}
+
+void ScheduleInfo::rotate(int Rotation, int II) {
+  for (int N = 0; N < NInstr; N++) {
+    int NewCycle = Nodes[N].Cycle + Rotation;
```
andcarminati: nit: `const int`

llvm/test/CodeGen/AIE/aie2p/schedule/postpipeliner/speculative-first-stage.mir
```diff
@@ -67,26 +67,25 @@
...
   ; CHECK-NEXT:  // %bb.2: // %for.cond.cleanup
-  ; CHECK-NEXT:    nopa ; nopb ; nops ; nopx ; vshuffle x3, x1, x2, r10; nopv
-  ; CHECK-NEXT:    nopx ; vmax_lt.32 x0, r16, x0, x3, vaddsign1
-  ; CHECK-NEXT:    nop
+  ; CHECK-NEXT:    vlda x2, [p0], #128; nopb ; nops ; nopxm ; nopv
```
andcarminati: Check: should this `vlda x2` be part of the epilogue? I feel that we are incrementing p0 once more in this version (in the epilogue), and `p0` in the live-in set of this block. Or did I miss some point?

llvm/lib/Target/AIE/AIEInterBlockScheduling.cpp
martien-de-jong: Don't forget `const int NonEmpty` . I was hoping that find_first(reverse(TimedRegion), predicate) would work, in combination with erase on a reversed range. There's too many many ugly ways in STL that do work and too few elegant ways that don't.

llvm/lib/Target/AIE/AIEInterBlockScheduling.cpp
martien-de-jong: Indeed, thanks!

llvm/lib/Target/AIE/AIEInterBlockScheduling.cpp
martien-de-jong: Although it is correct, and makes it work independently of the preceding stage. This is then just a footprint and runtime optimization

llvm/lib/Target/AIE/AIEPostPipeliner.cpp
martien-de-jong: That's a definite yes, especially since it's just a cheap attribute check. Although I don't think you should be able to pipeline anything across such an instruction.

llvm/lib/Target/AIE/AIEPostPipeliner.cpp
martien-de-jong: No, it's one of the Defs. The type already says that it is a machine operand. I really see no point in saying the same thing twice.  Note that gleaning he 'def' indication from the defs() in the caller is impeding refactoring into a reusable object. If you would have asked for Def I would have happily obliged.

andcarminati: Oh, right! Maybe `Def` is better.

llvm/lib/Target/AIE/AIEPostPipeliner.cpp
martien-de-jong: Thank you, yes. It shrunk from the original that did have the conditionals, or rather the if-statements.

llvm/test/CodeGen/AIE/aie2p/schedule/postpipeliner/speculative-first-stage.mir
martien-de-jong: This was sooo wrong. Nope, you didn't miss anything.

andcarminati: Is the rotation algorithm shifting this instruction out, and inducing this same instruction to be extracted as part of the epilogue?

martien-de-jong: The epilogue started one stage too early

andcarminati:
It looks very good! Do you have some QoR results?

martien-de-jong:
Core_Compute_Insn_Count
Core_StackSize

martien-de-jong:
Core_PMSize

martien-de-jong:
The 2 rejections were licence expiries. They have been confirmed afterwards to have no regressions
No fantastic improvements, but it is only preparatory for gemm.

andcarminati (approved):
LGTM.

PR #478, opened 2025-05-30
[AIEX] Introduces the new s20 narrowing combiner

andcarminati:
I aims to address some cases that were difficult to cover using the old design, according to the new tests. It adds:

* PHI narrowing combiner
* TRUNC combiner
* ZEXT combiner


llvm/lib/Target/AIE/AIECombinerHelper.cpp
```diff
@@ -3265,3 +2859,209 @@ bool llvm::matchPairedExtracts(MachineInstr &MI, MachineRegisterInfo &MRI,
...
+            MayNeedFullPrecision = true;
+            continue;
+          }
+        }
+        return UsedByS20;
+      };
+
+  SmallPtrSet<MachineInstr *, 4> VisitedInstrs;
+  bool MayNeedFullPrecision = false;
+  if (!IsUsedByS20Instruction(IsUsedByS20Instruction, MI, VisitedInstrs,
```
andcarminati: This recursive lambda pattern can be a bit odd.

llvm/test/CodeGen/AIE/GlobalISel/prelegalizercombiner-s20-narrowing.mir
```diff
@@ -105,14 +105,15 @@ body:             |
...
   ; CHECK-NEXT: {{  $}}
   ; CHECK-NEXT: bb.1:
   ; CHECK-NEXT:   successors: %bb.1(0x80000000)
   ; CHECK-NEXT: {{  $}}
-  ; CHECK-NEXT:   [[PTR_ADD:%[0-9]+]]:_(p0) = G_PTR_ADD [[COPY]], [[LOAD]](s20)
+  ; CHECK-NEXT:   [[TRUNC:%[0-9]+]]:_(s20) = G_TRUNC [[LOAD]](s32)
+  ; CHECK-NEXT:   [[PTR_ADD:%[0-9]+]]:_(p0) = G_PTR_ADD [[COPY]], [[TRUNC]](s20)
   ; CHECK-NEXT:   G_STORE [[C1]](s32), [[PTR_ADD]](p0) :: (store (s32))
-  ; CHECK-NEXT:   G_STORE [[LOAD]](s20), [[PTR_ADD]](p0) :: (store (s20), align 4)
+  ; CHECK-NEXT:   G_STORE [[LOAD]](s32), [[PTR_ADD]](p0) :: (store (s32))
```
andcarminati, line 116: For me this is a bug fix, not a regression.

llvm/lib/Target/AIE/AIECombinerHelper.cpp
```diff
@@ -3265,3 +2859,209 @@ bool llvm::matchPairedExtracts(MachineInstr &MI, MachineRegisterInfo &MRI,
...
+
+  // Do we have any s20 user of this PHI node? Also, do we have users that
+  // may require all 32 bits of the representation (extra users)?
+  // We don't need to explicitly know the end users of the s20 value, we
+  // can just look to the truncations.
+  auto IsUsedByS20Instruction =
+      [&](auto &Self, MachineInstr &Phi,
+          SmallPtrSet<MachineInstr *, 4> &VisitedInstrs,
+          bool &MayNeedFullPrecision) {
+        if (VisitedInstrs.contains(&Phi))
```
andcarminati, line 2915: We should false here!

llvm/lib/Target/AIE/AIECombinerHelper.cpp
```diff
@@ -3265,3 +2859,209 @@ bool llvm::matchPairedExtracts(MachineInstr &MI, MachineRegisterInfo &MRI,
...
+          return true;
```
andcarminati: return false here!

andcarminati:
We are colliding here with `phi_combines`, precisely with `extend_through_phis` that basically undo our job.

andcarminati:
QoR results:
image
image
image

llvm/lib/Target/AIE/AIECombinerHelper.cpp
```diff
@@ -3265,3 +2859,228 @@ bool llvm::matchPairedExtracts(MachineInstr &MI, MachineRegisterInfo &MRI,
...
+  // Ideally, we could allow more users, provided that they are all TRUNCs.
+  // However, if we have more users, the live range of this register could
+  // spread through more blocks, and this could lead to more register pressure
+  // on s20 registers.
+  if (!MRI.hasOneNonDBGUse(SrcReg) || !isUsedByLikelyLegalS20User(MRI, MI))
+    return false;
+
+  if (SrcMI.getOpcode() == TargetOpcode::G_LOAD) {
+    MatchInfo = [=, &MI, &SrcMI](MachineIRBuilder &B) {
+      const LLT S20 = LLT::scalar(20);
```
andcarminati: Refactor? Same code below.

llvm/lib/Target/AIE/AIECombinerHelper.cpp
```diff
@@ -3265,3 +2859,126 @@ bool llvm::matchPairedExtracts(MachineInstr &MI, MachineRegisterInfo &MRI,
 
   return true;
 }
+
+// Can we narrow all operands of a PHI node without loosing precision?
```
martien-de-jong: nit: losing

llvm/lib/Target/AIE/AIECombinerHelper.cpp
```diff
@@ -3265,3 +2859,126 @@ bool llvm::matchPairedExtracts(MachineInstr &MI, MachineRegisterInfo &MRI,
...
+// We can narrow loosing precision, provided that all further users of this
+// PHI node are s20 users (in this case we are not interested in the higher
+// bits).
+static bool
+canBeNarrowedWithoutLoose(MachineInstr &Phi,
```
martien-de-jong: WithoutLoss ?

llvm/lib/Target/AIE/AIECombinerHelper.cpp
```diff
@@ -3265,3 +2859,126 @@ bool llvm::matchPairedExtracts(MachineInstr &MI, MachineRegisterInfo &MRI,
...
+// Do we have any s20 user of this PHI node? Also, do we have users that
+// may require all 32 bits of the representation (extra users)?
+// We don't need to explicitly know the end users of the s20 value, we
+// can just look to the truncations.
+static bool
+isUsedByS20Instruction(MachineInstr &Phi,
+                       SmallPtrSet<MachineInstr *, 4> &VisitedInstrs,
+                       bool &MayNeedFullPrecision, MachineRegisterInfo &MRI) {
+  if (VisitedInstrs.contains(&Phi))
+    return false;
```
martien-de-jong: I think the function name could contain the 'any' from the comment. I guess we return false here since this branch of the search tree doesn't contribute any S20 use.

llvm/lib/Target/AIE/AIECombinerHelper.cpp
```diff
@@ -3265,3 +2859,126 @@ bool llvm::matchPairedExtracts(MachineInstr &MI, MachineRegisterInfo &MRI,
...
+  VisitedInstrs.insert(&Phi);
+  Register DefReg = Phi.getOperand(0).getReg();
+  bool UsedByS20 = false;
+  for (auto &User : MRI.use_instructions(DefReg)) {
+    const Register InputDefReg = User.getOperand(0).getReg();
```
martien-de-jong: nit: This may not be a DefReg

llvm/lib/Target/AIE/AIECombinerHelper.cpp
```diff
@@ -2980,3 +2980,99 @@ bool llvm::matchNarrowPhi(MachineInstr &MI, MachineRegisterInfo &MRI,
...
+static bool isUsedByLikelyLegalS20User(MachineRegisterInfo &MRI,
+                                       const MachineInstr &MI) {
+  return any_of(MRI.use_nodbg_instructions(MI.getOperand(0).getReg()),
+                [&](const MachineInstr &UseMI) {
+                  return UseMI.getOpcode() == TargetOpcode::G_PHI ||
+                         UseMI.getOpcode() == TargetOpcode::G_PTR_ADD ||
+                         UseMI.getOpcode() == TargetOpcode::G_STORE ||
+                         UseMI.getOpcode() == TargetOpcode::G_INTRINSIC ||
+                         UseMI.getOpcode() ==
+                             TargetOpcode::G_INTRINSIC_W_SIDE_EFFECTS;
```
martien-de-jong, line 3176: Should we narrow the intrinsics down a bit?

llvm/lib/Target/AIE/AIECombinerHelper.cpp
```diff
@@ -3265,3 +2859,124 @@ bool llvm::matchPairedExtracts(MachineInstr &MI, MachineRegisterInfo &MRI,
...
+      if (!isIntN(20, DefMI->getOperand(1).getCImm()->getSExtValue()))
+        return false;
+    } else if (Opcode == TargetOpcode::G_ZEXT) {
+      if (MRI.getType(DefMI->getOperand(1).getReg()) != LLT::scalar(20))
+        return false;
+    } else if (Opcode != TargetOpcode::G_IMPLICIT_DEF) {
+      return false;
+    }
+  }
+  return true;
```
martien-de-jong, line 3046: I think the logic would look more reliable if we would explicitly return true for the safe cases and return false on any other path.  I would also use a switch for the top level opcode check.

llvm/lib/Target/AIE/AIECombinerHelper.cpp
```diff
@@ -3265,3 +2859,124 @@ bool llvm::matchPairedExtracts(MachineInstr &MI, MachineRegisterInfo &MRI,
...
+// can just look to the truncations.
+static bool isUsedByAnyS20Instruction(
+    MachineInstr &Phi, SmallPtrSet<MachineInstr *, 4> &VisitedInstrs,
+    bool &MayNeedFullPrecision, MachineRegisterInfo &MRI) {
+  if (VisitedInstrs.contains(&Phi))
+    return false;
+  VisitedInstrs.insert(&Phi);
+  Register DefReg = Phi.getOperand(0).getReg();
+  bool UsedByS20 = false;
+  for (auto &User : MRI.use_instructions(DefReg)) {
```
martien-de-jong, line 3077: Perhaps add a comment that we can not break/return from this loop as soon as we have an S20 use because we need to collect MayNeedFullPrecision. Since there's no shortcut possible, you could also define an analysis result class with an accumulation operation. Perhaps you could even factor out a generic PHI traversal.

llvm/lib/Target/AIE/AIECombinerHelper.cpp
```diff
@@ -3265,3 +2859,124 @@ bool llvm::matchPairedExtracts(MachineInstr &MI, MachineRegisterInfo &MRI,
...
+  bool MayNeedFullPrecision = false;
+  if (!isUsedByAnyS20Instruction(MI, VisitedInstrs, MayNeedFullPrecision, MRI))
+    return false;
+
+  VisitedInstrs.clear();
+  const bool CanKeepPrecision =
+      canBeNarrowedWithoutLoss(MI, VisitedInstrs, MRI);
+
+  // We can only narrow in the following 2 situations:
+  // * We may loose precision by narrowing this PHI node, but all
```
martien-de-jong: nit: lose

llvm/lib/Target/AIE/AIECombinerHelper.cpp
```diff
@@ -3265,3 +2859,124 @@ bool llvm::matchPairedExtracts(MachineInstr &MI, MachineRegisterInfo &MRI,
...
+  // users are s20 bit users, so higher bits are not important.
+  // * We have users that may require full precision, however, we know
+  // that the narrowed inputs can be represented as s20 without loose
```
martien-de-jong: nit: loss

llvm/lib/Target/AIE/AIECombinerHelper.cpp
```diff
@@ -3265,3 +2859,124 @@ bool llvm::matchPairedExtracts(MachineInstr &MI, MachineRegisterInfo &MRI,
...
+    const LLT S20 = LLT::scalar(20);
+    const Register NewDefReg = MRI.createGenericVirtualRegister(S20);
+    const Register DefReg = MI.getOperand(0).getReg();
+
+    MachineBasicBlock *MBB = MI.getParent();
+    MachineBasicBlock::iterator InsertPt = MBB->getFirstNonPHI();
+    B.setInsertPt(*MI.getParent(), InsertPt);
+    // Extend the output.
+    // We use ZExt because it is the only safe, considering the
+    // previous analysis (IsUsedByS20Instruction - trunc case).
```
martien-de-jong: nit: this name changed.

llvm/lib/Target/AIE/AIECombinerHelper.cpp
```diff
@@ -3265,3 +2859,124 @@ bool llvm::matchPairedExtracts(MachineInstr &MI, MachineRegisterInfo &MRI,
...
+  // * We may loose precision by narrowing this PHI node, but all
+  // users are s20 bit users, so higher bits are not important.
+  // * We have users that may require full precision, however, we know
+  // that the narrowed inputs can be represented as s20 without loose
+  // of precision.
+  if (!CanKeepPrecision && MayNeedFullPrecision)
+    return false;
+
+  MatchInfo = [=, &MRI, &MI](MachineIRBuilder &B) {
+    // We change the PHI node instead of building a new one.
```
martien-de-jong, line 3128: Add a short summary comment, like PHI(A, B)  -> ZEXT(PHI(TRUNC(A), TRUNC(B)))

llvm/lib/Target/AIE/AIECombinerHelper.cpp
```diff
@@ -3265,3 +2859,124 @@ bool llvm::matchPairedExtracts(MachineInstr &MI, MachineRegisterInfo &MRI,
...
+    // We use ZExt because it is the only safe, considering the
+    // previous analysis (IsUsedByS20Instruction - trunc case).
+    B.buildZExt(DefReg, NewDefReg);
+    MI.getOperand(0).setReg(NewDefReg);
+
+    // Truncate the inputs.
+    for (unsigned OpNum = 1; OpNum < MI.getNumOperands(); OpNum += 2) {
+      const Register SrcReg = MI.getOperand(OpNum).getReg();
+      const Register NewSrcReg = MRI.createGenericVirtualRegister(S20);
+      MachineInstr *DefMI = MRI.getVRegDef(SrcReg);
```
martien-de-jong, line 3146: perhaps check for an (earlier created) ZEXT ?

llvm/test/CodeGen/AIE/GlobalISel/prelegalizercombiner-s20-narrowing.mir
```diff
@@ -178,3 +180,91 @@ body:             |
...
+legalized:         false
+tracksRegLiveness: true
+body:             |
+  ; CHECK-LABEL: name: neg_G_PHI_G_CONSTANT_s20
+  ; CHECK: bb.0.entry:
+  ; CHECK-NEXT:   successors: %bb.1(0x80000000)
+  ; CHECK-NEXT: {{  $}}
+  ; CHECK-NEXT:   [[C:%[0-9]+]]:_(p0) = G_CONSTANT i20 0
+  ; CHECK-NEXT:   [[C1:%[0-9]+]]:_(s20) = G_CONSTANT i20 0
+  ; CHECK-NEXT:   [[C2:%[0-9]+]]:_(s32) = G_CONSTANT i32 1048577
```
martien-de-jong, line 239: The s20 user is going to truncate it anyway, and I hope it will be loaded in an s20 register. Do we have a case against it?

llvm/test/CodeGen/AIE/aie2/addr_1d2d3d.ll
```diff
@@ -27,14 +27,14 @@ define dso_local ptr @test_add_2d_ptr(ptr %a, i32 noundef %off, i32 noundef %siz
...
+; CHECK-NEXT:    mova r4, #6
 ; CHECK-NEXT:    lda dc0, [p2, #0]
-; CHECK-NEXT:    mov p0, p1
-; CHECK-NEXT:    lshl r0, r0, r3
+; CHECK-NEXT:    mov dn0, r1
+; CHECK-NEXT:    lshl r0, r0, r4
 ; CHECK-NEXT:    mov m0, r0
 ; CHECK-NEXT:    ret lr
-; CHECK-NEXT:    lshl r0, r2, r3 // Delay Slot 5
+; CHECK-NEXT:    lshl r0, r2, r4 // Delay Slot 5
```
martien-de-jong: We keep seeing these unexplained register allocator changes. I think that's a suboptimal register coalescing that gets repaired by removing a trivial move later on.

llvm/lib/Target/AIE/AIEBaseInstrInfo.cpp
```diff
@@ -1288,3 +1288,14 @@ AIEBaseInstrInfo::getAlignmentBoundaries(MachineBasicBlock &MBB) const {
...
+
+bool AIEBaseInstrInfo::isExtendLikelyToBeFolded(
+    MachineInstr &ExtMI, MachineRegisterInfo &MRI) const {
+  assert(ExtMI.getOpcode() == TargetOpcode::G_SEXT ||
+         ExtMI.getOpcode() == TargetOpcode::G_ZEXT ||
+         ExtMI.getOpcode() == TargetOpcode::G_ANYEXT);
+
+  const LLT S20 = LLT::scalar(20);
+  const LLT ActualType = MRI.getType(ExtMI.getOperand(1).getReg());
+  return ExtMI.getOpcode() == TargetOpcode::G_ZEXT && ActualType == S20;
```
martien-de-jong: It sounds as if G_ANYEXT could be included?

llvm/lib/Target/AIE/AIEBaseInstrInfo.cpp
```diff
@@ -1288,3 +1288,14 @@ AIEBaseInstrInfo::getAlignmentBoundaries(MachineBasicBlock &MBB) const {
...
 }
+
+bool AIEBaseInstrInfo::isExtendLikelyToBeFolded(
+    MachineInstr &ExtMI, MachineRegisterInfo &MRI) const {
+  assert(ExtMI.getOpcode() == TargetOpcode::G_SEXT ||
+         ExtMI.getOpcode() == TargetOpcode::G_ZEXT ||
+         ExtMI.getOpcode() == TargetOpcode::G_ANYEXT);
+
+  const LLT S20 = LLT::scalar(20);
+  const LLT ActualType = MRI.getType(ExtMI.getOperand(1).getReg());
```
martien-de-jong, line 1401: Early return on opcode?

llvm/test/CodeGen/AIE/GlobalISel/prelegalizercombiner-disable-zext-s20-phi.mir
```diff
@@ -0,0 +1,120 @@
...
+#
+# This file is licensed under the Apache License v2.0 with LLVM Exceptions.
+# See https://llvm.org/LICENSE.txt for license information.
+# SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+#
+# (c) Copyright 2025 Advanced Micro Devices, Inc. or its affiliates
+# RUN: llc -mtriple aie2p -run-pass=aie2p-prelegalizer-combiner -global-isel -verify-machineinstrs %s -o - | FileCheck %s
+
+# This should test the implementation of AIEBaseInstrInfo::isExtendLikelyToBeFolded
+# Check that we block the propagation the G_ZEXT to the sources of the phi operand with s20 type.
```
martien-de-jong, line 11: nit: add it in an earlier separate commit?

llvm/lib/Target/AIE/AIECombinerHelper.cpp
```diff
@@ -3265,3 +2859,124 @@ bool llvm::matchPairedExtracts(MachineInstr &MI, MachineRegisterInfo &MRI,
...
+    }
+  }
+  return true;
+}
+
+// Do we have any s20 user of this PHI node? Also, do we have users that
+// may require all 32 bits of the representation (extra users)?
+// We don't need to explicitly know the end users of the s20 value, we
+// can just look to the truncations.
+static bool isUsedByAnyS20Instruction(
```
martien-de-jong: Super nit: we may be traversing the same partial PHI chains from any block that refers to it. Perhaps we could cache analysis in a function-wide map.

llvm/lib/Target/AIE/AIECombinerHelper.cpp
```diff
@@ -3265,3 +2859,220 @@ bool llvm::matchPairedExtracts(MachineInstr &MI, MachineRegisterInfo &MRI,
...
+    bool &MayNeedFullPrecision, MachineRegisterInfo &MRI) {
+  if (VisitedInstrs.contains(&Phi))
+    return false;
+  VisitedInstrs.insert(&Phi);
+  Register DefReg = Phi.getOperand(0).getReg();
+  bool UsedByS20 = false;
+  for (auto &User : MRI.use_instructions(DefReg)) {
+    const unsigned Opcode = User.getOpcode();
+    if (Opcode == TargetOpcode::G_TRUNC &&
+        MRI.getType(User.getOperand(0).getReg()).getScalarSizeInBits() <= 20) {
```
F-Stuckmann: why are we here allowing the result of the truncation to have less than 20 bits?

llvm/lib/Target/AIE/AIECombinerHelper.cpp
```diff
@@ -3265,3 +2859,220 @@ bool llvm::matchPairedExtracts(MachineInstr &MI, MachineRegisterInfo &MRI,
 
   return true;
 }
+
+// Can we narrow all operands of a PHI node without losing precision?
+// We can narrow losing precision, provided that all further users of this
+// PHI node are s20 users (in this case we are not interested in the higher
+// bits).
+static bool
+canBeNarrowedWithoutLoss(MachineInstr &Phi,
```
F-Stuckmann, line 3021: NIT: I don't think the function is doing what the name suggests.
Maybe something more descriptive like: isUsedByAnyNativePrecisionInstruction

llvm/lib/Target/AIE/AIECombinerHelper.cpp
```diff
@@ -3265,3 +2859,220 @@ bool llvm::matchPairedExtracts(MachineInstr &MI, MachineRegisterInfo &MRI,
...
+  // users are s20 bit users, so higher bits are not important.
+  // * We have users that may require full precision, however, we know
+  // that the narrowed inputs can be represented as s20 without loose
+  // of precision.
+  if (!CanKeepPrecision && MayNeedFullPrecision)
+    return false;
+
+  MatchInfo = [=, &MRI, &MI](MachineIRBuilder &B) {
+    // We change the PHI node instead of building a new one.
+    const LLT S20 = LLT::scalar(20);
```
F-Stuckmann, line 3129: Can the magic number be a global constant, so that (future) changes to the bit width are simplified?

llvm/lib/Target/AIE/AIECombinerHelper.cpp
andcarminati: Here, we are looking to the inputs instead of the users. We trying to prove that we can encode all the 32bit inputs using 20 bits.

llvm/lib/Target/AIE/AIECombinerHelper.cpp
andcarminati: That is the point, I look for the invalid cases, if I catch one, I return false. I cannot just return true for a safe case without looking to all other cases.

llvm/lib/Target/AIE/AIECombinerHelper.cpp
andcarminati: Nice idea, I will implement.

llvm/lib/Target/AIE/AIECombinerHelper.cpp
andcarminati: I will create a result class, nice idea. If we see other cases where this traversal could be used, we can factor it out. Now I will keep as is because the other traversal will not use the cache.

llvm/lib/Target/AIE/AIECombinerHelper.cpp
andcarminati: This is a very corner case, you are right. I will split in two separated cases: 
* TRUNC == 20 -> UsedByS20 = true
* TRUNC > 20 -> MayNeedFullPrecision = true.

llvm/lib/Target/AIE/AIECombinerHelper.cpp
andcarminati: Unfortunately, our codebase is heavily based on s20. If we start to have different sizes here, we should think in parametrize this directly from TII.

llvm/lib/Target/AIE/AIECombinerHelper.cpp
andcarminati: That is the nice part: we can rely on existing combiners to do this job.

llvm/lib/Target/AIE/AIECombinerHelper.cpp
andcarminati: We reach an intrinsic that uses an s20, this is enough. This was implemented because we reached some code in mllib where we shift some s20 values to calculate offsets and this will be further legalized as s32.

llvm/lib/Target/AIE/AIEBaseInstrInfo.cpp
andcarminati: This is just to not conflict with the ZEXTs that we add.

llvm/lib/Target/AIE/AIECombinerHelper.cpp
martien-de-jong: But it means you have to be sure that you cover all illegal cases. My main problem is that you will return true on any opcode that isn't listed.  You can invert the logic using return all_of(operands, safe).

llvm/lib/Target/AIE/AIECombinerHelper.cpp
```diff
@@ -3265,3 +2859,155 @@ bool llvm::matchPairedExtracts(MachineInstr &MI, MachineRegisterInfo &MRI,
...
+  static std::map<MachineInstr *, PhiUsageAnalysisResult> CachedPhiAnalysis;
+  auto CacheIt = CachedPhiAnalysis.find(&Phi);
+  if (CacheIt != CachedPhiAnalysis.end()) {
+    Result = CacheIt->second;
+    return Result;
+  }
+
+  if (VisitedInstrs.contains(&Phi))
+    return Result;
+  VisitedInstrs.insert(&Phi);
```
martien-de-jong, line 3072: Nit: You could eliminate the set by pushing a preliminary result in the Cache with an 'InProgress' and 'Finalized'  status.

llvm/lib/Target/AIE/AIECombinerHelper.cpp
```diff
@@ -3265,3 +2859,155 @@ bool llvm::matchPairedExtracts(MachineInstr &MI, MachineRegisterInfo &MRI,
...
+// can just look to the truncations.
+static PhiUsageAnalysisResult
+isUsedByAnyS20Instruction(MachineInstr &Phi,
+                          SmallPtrSet<MachineInstr *, 4> &VisitedInstrs,
+                          MachineRegisterInfo &MRI) {
+
+  PhiUsageAnalysisResult Result;
+  // We can cache the results for a fast next search through the same block
+  // sequence.
+  static std::map<MachineInstr *, PhiUsageAnalysisResult> CachedPhiAnalysis;
```
martien-de-jong: I think this map will get clogged with phi nodes from other functions. Do we have a function-wide structure that can own it?

martien-de-jong: Also, the map may be invalidated if any of the PHI nodes is freed. I'm not sure whether other combiners are interleaved with the lifetime of this map?

llvm/lib/Target/AIE/AIECombinerHelper.cpp
martien-de-jong: I'm worried that we may run those combiners repeatedly to propagate the narrowing one phi one step further each time.

llvm/lib/Target/AIE/AIECombinerHelper.cpp
andcarminati: Discussed offline: better to remove this cache.

llvm/lib/Target/AIE/AIECombinerHelper.cpp
andcarminati: Cache was removed ;-)

llvm/lib/Target/AIE/AIECombinerHelper.cpp
andcarminati: I this this can happen indeed, but in this way can keep the solution as simple as possible, and we can reuse other combiners to cleanup the code.

llvm/test/CodeGen/AIE/GlobalISel/prelegalizercombiner-s20-narrowing.mir
andcarminati: We are subtracting this value (32 bits), so if we truncate we will change the final result here.

llvm/test/CodeGen/AIE/aie2/addr_1d2d3d.ll
andcarminati: Yes, sometimes annoying!

llvm/test/CodeGen/AIE/GlobalISel/prelegalizercombiner-disable-zext-s20-phi.mir
andcarminati: Now this is related to the first commit.

llvm/lib/Target/AIE/AIECombinerHelper.cpp
```diff
@@ -3311,3 +2905,147 @@ bool llvm::matchBroadcastToShl(MachineInstr &MI, MachineRegisterInfo &MRI,
...
+  }
+
+  return Result;
+}
+
+/// Narrow Phi nodes to s20, when it is safe.
+bool llvm::matchNarrowPhi(MachineInstr &MI, MachineRegisterInfo &MRI,
+                          BuildFnTy &MatchInfo) {
+
+  assert(MI.getOpcode() == TargetOpcode::G_PHI);
```
martien-de-jong: Very generally, in this case where we assert that it is a phi, could we call it Phi? MI is not a good name.

llvm/lib/Target/AIE/AIECombinerHelper.cpp
martien-de-jong: When we go generalize, I think it should be seen as a general RequiredBits case.

llvm/lib/Target/AIE/AIECombinerHelper.cpp
```diff
@@ -3049,3 +3049,99 @@ bool llvm::matchNarrowPhi(MachineInstr &MI, MachineRegisterInfo &MRI,
...
+                         UseMI.getOpcode() == TargetOpcode::G_STORE ||
+                         UseMI.getOpcode() == TargetOpcode::G_INTRINSIC ||
+                         UseMI.getOpcode() ==
+                             TargetOpcode::G_INTRINSIC_W_SIDE_EFFECTS;
+                });
+}
+
+static void changeLoadStoreDataRegister(MachineInstr &MI, Register DataReg,
+                                        MachineRegisterInfo &MRI) {
+  GLoadStore &LoadInst = cast<GLoadStore>(MI);
```
martien-de-jong: nit: LdStInst?

llvm/lib/Target/AIE/AIECombinerHelper.cpp
```diff
@@ -3049,3 +3049,99 @@ bool llvm::matchNarrowPhi(MachineInstr &MI, MachineRegisterInfo &MRI,
...
+  LoadInst.getMMO().setType(MRI.getType(DataReg));
+  LoadInst.getOperand(0).setReg(DataReg);
+}
+
+/// Narrow operations that are feeding truncations to s20.
+/// Covers G_LOAD and G_CONSTANT.
```
martien-de-jong, line 3188: I only see G_CONSTANT being mentioned?

andcarminati: Load is a bit more down.

F-Stuckmann:
Did something go wrong with a rebase from aie-public? I see a lot of changes from the global combiners, which should not be in this PR

llvm/lib/Target/AIE/AIECombinerHelper.cpp
```diff
@@ -3410,3 +3017,243 @@ bool llvm::matchBroadcastToShl(MachineInstr &MI, MachineRegisterInfo &MRI,
...
+
+    // Truncate the inputs.
+    for (unsigned OpNum = 1; OpNum < Phi.getNumOperands(); OpNum += 2) {
+      const Register SrcReg = Phi.getOperand(OpNum).getReg();
+      const Register NewSrcReg = MRI.createGenericVirtualRegister(S20);
+      MachineInstr *DefMI = MRI.getVRegDef(SrcReg);
+
+      MachineBasicBlock *DefMIMBB = DefMI->getParent();
+      MachineBasicBlock::iterator InsertPt = ++DefMI->getIterator();
+      if (InsertPt != DefMIMBB->end() && InsertPt->isPHI())
```
F-Stuckmann, line 3150: why aren't we directly assigning 
`MachineBasicBlock::iterator InsertPt = DefMIMBB->getFirstNonPHI();` ? I don't think we need the if condition

llvm/lib/Target/AIE/AIECombinerHelper.cpp
```diff
@@ -3410,3 +3017,243 @@ bool llvm::matchBroadcastToShl(MachineInstr &MI, MachineRegisterInfo &MRI,
...
+                         UseMI.getOpcode() == TargetOpcode::G_STORE ||
+                         UseMI.getOpcode() == TargetOpcode::G_INTRINSIC ||
+                         UseMI.getOpcode() ==
+                             TargetOpcode::G_INTRINSIC_W_SIDE_EFFECTS;
+                });
+}
+
+static void changeLoadStoreDataRegister(MachineInstr &MI, Register DataReg,
+                                        MachineRegisterInfo &MRI) {
+  GLoadStore &LdInst = cast<GLoadStore>(MI);
```
F-Stuckmann: NIT: LdStInst

llvm/lib/Target/AIE/AIECombinerHelper.cpp
andcarminati: This is very subtle. If we will insert a TRUNC of a result of a non-phi instruction, we should insert after this instruction itself. If we insert a TRUNC of a result generated by a phi, we should insert after all phis. If we go for unconditionally `InsertPt = DefMIMBB->getFirstNonPHI()` we fatally will end up generating use-before def cases. See:

We want for non phis:
```
x = PHI ....
y = ADD ....
z = TRUNC y
```

And not:

```
x = PHI ....
z = TRUNC y
y = ADD ....
```


We want for phis:
```
x = PHI ....
y = PHI ....
z = TRUNC x
```

And not:

```
x = PHI ....
z = TRUNC x
y = PHI ....
```

llvm/lib/Target/AIE/AIECombinerHelper.cpp
andcarminati: Done!

llvm/lib/Target/AIE/AIECombinerHelper.cpp
martien-de-jong: But don't we always insert before the first non-phi after the defMI?

andcarminati: > But don't we always insert before the first non-phi after the defMI?

Yes, but the exact point will depend on the type of instruction. For non-phi, immediately after, for phi, the first non-phi.

llvm/lib/Target/AIE/AIECombinerHelper.cpp
```diff
@@ -3405,3 +3012,147 @@ bool llvm::matchBroadcastToShl(MachineInstr &MI, MachineRegisterInfo &MRI,
...
 }
+
+// Can we narrow all operands of a PHI node without losing precision?
+// We can narrow losing precision, provided that all further users of this
+// PHI node are s20 users (in this case we are not interested in the higher
+// bits).
+static bool
+canBeNarrowedWithoutLoss(MachineInstr &Phi,
+                         SmallPtrSet<MachineInstr *, 4> &VisitedInstrs,
+                         MachineRegisterInfo &MRI) {
```
krishnamtibrewala, line 3023: nit: assert/check the PHI node is of S32 type.

llvm/lib/Target/AIE/AIECombinerHelper.cpp
```diff
@@ -3405,3 +3012,147 @@ bool llvm::matchBroadcastToShl(MachineInstr &MI, MachineRegisterInfo &MRI,
...
+    MachineInstr *DefMI = MRI.getVRegDef(Phi.getOperand(OpNum).getReg());
+    const unsigned Opcode = DefMI->getOpcode();
+    if (Opcode == TargetOpcode::G_PHI) {
+      if (!canBeNarrowedWithoutLoss(*DefMI, VisitedInstrs, MRI))
+        return false;
+    } else if (Opcode == TargetOpcode::G_CONSTANT) {
+      if (!isIntN(20, DefMI->getOperand(1).getCImm()->getSExtValue()))
+        return false;
+    } else if (Opcode == TargetOpcode::G_ZEXT) {
+      if (MRI.getType(DefMI->getOperand(1).getReg()) != LLT::scalar(20))
```
krishnamtibrewala, line 3040: CHECK : should we allow even if the SRC operand of G_ZEXT is != S20, allow it for anything that is less that 20bit too, we can have a S8 that is ZEXT to S32, which will be a COPY down the line but if we can do S8 to S20 without loosing precision we should do that.

llvm/lib/Target/AIE/AIECombinerHelper.cpp
```diff
@@ -3405,3 +3012,147 @@ bool llvm::matchBroadcastToShl(MachineInstr &MI, MachineRegisterInfo &MRI,
...
+        return false;
+    } else if (Opcode != TargetOpcode::G_IMPLICIT_DEF) {
+      return false;
```
krishnamtibrewala, line 3043: CHECK : how do we handle following case of load

<S32> %1 = LOAD X
<S32> %2 = LOAD X=32

<S32> %3 = G_PHI %1, %2
<S20> %4 = G_TRUNC %3
......... %4 is used in addr_2d/3d

llvm/lib/Target/AIE/AIECombinerHelper.cpp
```diff
@@ -3405,3 +3012,147 @@ bool llvm::matchBroadcastToShl(MachineInstr &MI, MachineRegisterInfo &MRI,
...
+  // PHI(A, B) -> ZEXT(PHI(TRUNC(A), TRUNC(B)))
+  MatchInfo = [=, &MRI, &Phi](MachineIRBuilder &B) {
+    // We change the PHI node instead of building a new one.
+    const LLT S20 = LLT::scalar(20);
+    const Register NewDefReg = MRI.createGenericVirtualRegister(S20);
+    const Register DefReg = Phi.getOperand(0).getReg();
+
+    MachineBasicBlock *MBB = Phi.getParent();
+    MachineBasicBlock::iterator InsertPt = MBB->getFirstNonPHI();
+    B.setInsertPt(*Phi.getParent(), InsertPt);
```
krishnamtibrewala, line 3135: We should be updating GISelChangeObserver to inform we have added a new instruction in the function using "[createdInstr](https://llvm.org/doxygen/classllvm_1_1GISelChangeObserver.html#ac53bec73a7e61679a0aed216280a0fc7)[createdInstr](https://llvm.org/doxygen/classllvm_1_1GISelChangeObserver.html#ac53bec73a7e61679a0aed216280a0fc7)"

llvm/test/CodeGen/AIE/GlobalISel/prelegalizercombiner-s20-narrowing.mir
krishnamtibrewala: CHECK : 
Will this now lead into 
$r1 = LOAD x
$m0 = COPY $r1
$p0 = PTR_ADD $p0, $m0 

If yes, then we are adding an extra copy which otherwise would have been avoided.

llvm/lib/Target/AIE/AIECombinerHelper.cpp
```diff
@@ -3156,3 +3156,99 @@ bool llvm::matchNarrowPhi(MachineInstr &Phi, MachineRegisterInfo &MRI,
...
+static void changeLoadStoreDataRegister(MachineInstr &MI, Register DataReg,
+                                        MachineRegisterInfo &MRI) {
+  GLoadStore &LdStInst = cast<GLoadStore>(MI);
+  LdStInst.getMMO().setType(MRI.getType(DataReg));
+  LdStInst.getOperand(0).setReg(DataReg);
+}
+
+/// Narrow operations that are feeding truncations to s20.
+/// Covers G_LOAD and G_CONSTANT.
+bool llvm::matchNarrowTrunc(MachineInstr &MI, MachineRegisterInfo &MRI,
```
krishnamtibrewala, line 3189: Please update GISelChangeObserver.

llvm/test/CodeGen/AIE/GlobalISel/prelegalizercombiner-s20-narrowing.mir
krishnamtibrewala: Its handled by the next commit

llvm/lib/Target/AIE/AIECombinerHelper.cpp
```diff
@@ -3156,3 +3156,99 @@ bool llvm::matchNarrowPhi(MachineInstr &Phi, MachineRegisterInfo &MRI,
...
+    };
+    return true;
+  }
+
+  return false;
+}
+
+/// Narrow operations that are fed by zext from s20.
+/// Covers G_STORE.
+bool llvm::matchNarrowZext(MachineInstr &MI, MachineRegisterInfo &MRI,
```
krishnamtibrewala, line 3236: Please update GISelChangeObserver.

llvm/lib/Target/AIE/AIECombinerHelper.cpp
martien-de-jong: Yep. And example would be a value that is used by a broadcast or insert with a byte elemtype.

martien-de-jong: Yep. An example would be an insert or a broadcast with a byte element type

llvm/lib/Target/AIE/AIECombinerHelper.cpp
```diff
@@ -3405,3 +3012,147 @@ bool llvm::matchBroadcastToShl(MachineInstr &MI, MachineRegisterInfo &MRI,
...
+  // because we need to collect MayNeedFullPrecision.
+  for (auto &User : MRI.use_instructions(DefReg)) {
+    const unsigned Opcode = User.getOpcode();
+    if (Opcode == TargetOpcode::G_TRUNC) {
+      const unsigned TruncatedSize =
+          MRI.getType(User.getOperand(0).getReg()).getScalarSizeInBits();
+      if (TruncatedSize == 20)
+        Result.IsUsedByS20 = true;
+      else if (TruncatedSize > 20)
+        Result.MayNeedFullPrecision = true;
```
martien-de-jong, line 3085: We could generalize by collecting the max required size.

llvm/lib/Target/AIE/AIECombinerHelper.cpp
andcarminati: This is a common case, we will trunc and narrow the phi. Later on, the load will be narrowed by the next combiner as well.

llvm/lib/Target/AIE/AIECombinerHelper.cpp
andcarminati: This is checked by the single caller function: matchNarrowPhi.

llvm/lib/Target/AIE/AIECombinerHelper.cpp
andcarminati: It is not like this way. We can narrow for < 8 bits, provided that we have a s20 user. Keep in mind that all s8 types are legalized to s32, so, if we narrow only for s8/s16, we will end with a phi node carrying a s20 that will be always extended to s32. Side effects:
* We loose one s20 register.
* All uses of this value should be preceded by (S/Z)EXT - being the first one relatively expensive.

So, we are using a simple rule here: no s20 user, no narrowing.

llvm/lib/Target/AIE/AIECombinerHelper.cpp
andcarminati: Sounds interesting, but I think it can only be useful if we have use cases with arbitrary scalar types.

llvm/lib/Target/AIE/AIECombinerHelper.cpp
andcarminati: Done!

llvm/lib/Target/AIE/AIECombinerHelper.cpp
andcarminati: Done!

llvm/lib/Target/AIE/AIECombinerHelper.cpp
andcarminati: Done!

llvm/lib/Target/AIE/AIECombinerHelper.cpp
```diff
@@ -3405,3 +3012,251 @@ bool llvm::matchBroadcastToShl(MachineInstr &MI, MachineRegisterInfo &MRI,
...
+// bits).
+static bool
+canBeNarrowedWithoutLoss(MachineInstr &Phi,
+                         SmallPtrSet<MachineInstr *, 4> &VisitedInstrs,
+                         MachineRegisterInfo &MRI) {
+  if (VisitedInstrs.contains(&Phi))
+    return true;
+  VisitedInstrs.insert(&Phi);
+  for (unsigned OpNum = 1; OpNum < Phi.getNumOperands(); OpNum += 2) {
+    MachineInstr *DefMI = MRI.getVRegDef(Phi.getOperand(OpNum).getReg());
```
F-Stuckmann: NIT: maybe assert that the phi works on s32 types or check for `DefMI` for nullptr

F-Stuckmann (commented):
LGTM

llvm/lib/Target/AIE/AIECombinerHelper.cpp
andcarminati: I updated asserting that the size should be 32 bits.

llvm/lib/Target/AIE/AIECombinerHelper.cpp
```diff
@@ -3405,3 +3012,254 @@ bool llvm::matchBroadcastToShl(MachineInstr &MI, MachineRegisterInfo &MRI,
...
+static bool
+canBeNarrowedWithoutLoss(MachineInstr &Phi,
+                         SmallPtrSet<MachineInstr *, 4> &VisitedInstrs,
+                         MachineRegisterInfo &MRI) {
+  if (VisitedInstrs.contains(&Phi))
+    return true;
+  VisitedInstrs.insert(&Phi);
+  for (unsigned OpNum = 1; OpNum < Phi.getNumOperands(); OpNum += 2) {
+    const Register PhiRegIn = Phi.getOperand(OpNum).getReg();
+    assert(MRI.getType(PhiRegIn).getSizeInBits() == 32 && "Mixed Phi node");
```
F-Stuckmann, line 3029: I like the assert

F-Stuckmann (approved):

PR #229, opened 2024-11-01
[AIEX] Ignore bank conflict if in the next cycle we cannot schedule the instruction.

krishnamtibrewala:

PR #481, opened 2025-06-06
[AIEX] Add memory wait cycles avoidance in case of shared base values.

andcarminati:
This prevents the scheduler from packing instructions sharing the same MMO Value because, in this case, bank conflict will happen anyway. This is intended to avoid wait cycles.


llvm/test/CodeGen/AIE/schedule/pointer-hazard.mir
```diff
@@ -7,33 +7,36 @@
 # (c) Copyright 2025 Advanced Micro Devices, Inc. or its affiliates
 
 # RUN: llc --mtriple=aie2 --issue-limit=6 --start-before=postmisched \
-# RUN:     %s -o - | FileCheck %s
+# RUN:     --aie-recognize-pointer-hazards=true %s -o - | FileCheck %s
```
andcarminati: Nit, we can remove if we change the default value to true.

llvm/lib/Target/AIE/AIEHazardRecognizer.cpp
```diff
@@ -677,3 +705,47 @@ AIEHazardRecognizer::getMemoryBanks(const MachineInstr *MI) const {
...
+
+  if (!(MI->mayLoad() || MI->mayStore()))
+    return 0;
+
+  if (MI->memoperands_empty())
+    return 0; // optimistic: we will assume no conflict.
+
+  MemoryObjectsBits ObjectsToBitMap = 0;
+  for (auto &MMO : MI->memoperands()) {
+    const Value *BaseObject = MMO->getValue();
```
andcarminati, line 788: If we wish to pay the runtime cost, we can use `ValueTracking` to go to the root Value (stripping all pointer casts, geps  and 2d/3d ptradds). Maybe it is not too heavy, as we can control the number of hops.

andcarminati: In this case, we can cache the result because probably we will do this same step several times for each instruction. We move this logic into the enumerator.

llvm/lib/Target/AIE/AIEHazardRecognizer.cpp
```diff
@@ -677,3 +706,58 @@ AIEHazardRecognizer::getMemoryBanks(const MachineInstr *MI) const {
...
+}
+
+unsigned MemoryObjectEnumerator::getObjectNumber(const Value *Object) {
+  unsigned ObjectNumber;
+
+  auto ItParent = ParentObjectMap.find(Object);
+  if (ItParent != ParentObjectMap.end()) {
+    Object = ItParent->second;
+  } else {
+    const Value *ParentObject = getUnderlyingObject(Object);
```
andcarminati: We are caching here to softener the overhead.

krishnamtibrewala:
Hi @andcarminati 
Nice way of avoiding conflict when bank annotation are missing. 

Should we consider the offset too ? There can be a case where two memory access have the same base pointer but the added offset can effectively lead to change in Bank.

For example : 
LOAD $x0, $p1
LOAD $x0, [$p1, XYZ] 
Where in XYZ is big enough to change the bank. 

It could be tricky to be sure to be sure of what offset will cause the change in bank, given at compile time we are not aware of value of $p1 but we can have upper cap where in XYZ >= [SIZE of bank] will for sure lead to change in bank

llvm/lib/Target/AIE/AIEHazardRecognizer.cpp
```diff
@@ -87,6 +95,7 @@ void FuncUnitWrapper::clearResources() {
   Reserved = 0;
   Slots = 0;
   MemoryBanks = 0;
+  ObjectsBits = 0;
```
krishnamtibrewala: nit : having "Mem" or "Memory" in the variable name would be helpful

llvm/lib/Target/AIE/AIEHazardRecognizer.cpp
```diff
@@ -351,7 +362,8 @@ AIEHazardRecognizer::getHazardType(SUnit *SU, int DeltaCycles) {
   }
 
   return getHazardType(Scoreboard, MI->getDesc(), getMemoryBanks(MI),
-                       MI->operands(), MRI, DeltaCycles);
+                       getMemoryObjectsBits(MI), MI->operands(), MRI,
+                       DeltaCycles);
```
krishnamtibrewala: nit : not as part of this PR for a future refactoring we should consider having 

`  getHzardType(Scoreboard, MI, MRI, DeltaCycles)`

 to wrap

```
  getHazardType(const ResourceScoreboard<FuncUnitWrapper> &TheScoreboard,
                const MCInstrDesc &Desc, MemoryBankBits MemoryBanks,
                uint64_t ObjectsBits,
                iterator_range<const MachineOperand *> MIOperands,
                const MachineRegisterInfo &MRI, int DeltaCycles) const;
```

andcarminati: Nice suggestion, thank you! I included another commit with this change.

andcarminati:
> Hi @andcarminati Nice way of avoiding conflict when bank annotation are missing.
> 
> Should we consider the offset too ? There can be a case where two memory access have the same base pointer but the added offset can effectively lead to change in Bank.
> 
> For example : LOAD $x0, $p1 LOAD $x0, [$p1, XYZ] Where in XYZ is big enough to change the bank.
> 
> It could be tricky to be sure to be sure of what offset will cause the change in bank, given at compile time we are not aware of value of $p1 but we can have upper cap where in XYZ >= [SIZE of bank] will for sure lead to change in bank

This is an interesting point, but I believe that, if we reach this situation, it is safe enough to consider as a conflict. As you said, we should encode more knowledge about the bank.

krishnamtibrewala:
> > Hi @andcarminati Nice way of avoiding conflict when bank annotation are missing.
> > Should we consider the offset too ? There can be a case where two memory access have the same base pointer but the added offset can effectively lead to change in Bank.
> > For example : LOAD $x0, $p1 LOAD $x0, [$p1, XYZ] Where in XYZ is big enough to change the bank.
> > It could be tricky to be sure to be sure of what offset will cause the change in bank, given at compile time we are not aware of value of $p1 but we can have upper cap where in XYZ >= [SIZE of bank] will for sure lead to change in bank
> 
> This is an interesting point, but I believe that, if we reach this situation, it is safe enough to consider as a conflict. As you said, we should encode more knowledge about the bank.

Yes, it is surely safe. The possibility of the case I share is less but when it happens it will be safe to mark them as no conflict. 
How about we create a ticket with lowest priority as a form of future documentation ?

andcarminati:
> > > Hi @andcarminati Nice way of avoiding conflict when bank annotation are missing.
> > > Should we consider the offset too ? There can be a case where two memory access have the same base pointer but the added offset can effectively lead to change in Bank.
> > > For example : LOAD $x0, $p1 LOAD $x0, [$p1, XYZ] Where in XYZ is big enough to change the bank.
> > > It could be tricky to be sure to be sure of what offset will cause the change in bank, given at compile time we are not aware of value of $p1 but we can have upper cap where in XYZ >= [SIZE of bank] will for sure lead to change in bank
> > 
> > 
> > This is an interesting point, but I believe that, if we reach this situation, it is safe enough to consider as a conflict. As you said, we should encode more knowledge about the bank.
> 
> Yes, it is surely safe. The possibility of the case I share is less but when it happens it will be safe to mark them as no conflict. How about we create a ticket with lowest priority as a form of future documentation ?

Hi @krishnamtibrewala , thank you for the suggestion, I feel that it is more correlated to an extension to memory banks (see that here we only look to the IR/Value here). I need to think, because I see some practical difficulties to implement efficiently this approach, like:

- How to represent bank ranges in the scoreboard in a way that they can be checked efficiently.
- How to expose bank sizes (actually, I think that this is more ABI/System knowledge rather than compiler).
- How to take offsets:
       * by looking to the MMOs (they will relative to Values, no banks), or
       * by looking to encodings.
- How to scale the encoded offsets to the real offset (TII knowledge for every memory instruction or tablegen). 
- How to deal with instructions with complex offsets (registers/2d/3d). Just ignore them? They can touch anything that we don't know, so they can put the scoreboard in a state that anything could be expected.

I would like to have a feasible scope for this ticket, feel free to suggest, maybe there is some simplified model that we can describe.

llvm/lib/Target/AIE/AIEPostPipeliner.cpp
```diff
@@ -694,7 +695,7 @@ bool PostPipeliner::scheduleFirstIteration(PostPipelinerStrategy &Strategy) {
         return false;
       }
 
-      HR.emitInScoreboard(Scoreboard, MI->getDesc(), MemoryBanks,
+      HR.emitInScoreboard(Scoreboard, MI->getDesc(), MemoryBanks, ObjectBits,
```
andcarminati, line 725: I prefer to keep this verbosity, as:

```
    const MemoryBankBits MemoryBanks = HR.getMemoryBanks(MI);
    const MemoryObjectsBits ObjectBits = HR.getMemoryObjectsBits(MI);
```
Can be costly.

andcarminati:
QoR results:
image

Average is -0.36%

The regression is related to a load that was split in two halves (alignment?), and now cannot run in parallel.

krishnamtibrewala:
> The regression is related to a load that was split in two halves (alignment?), and now cannot run in parallel.

CHECK : 
1. Is this the split by the compiler or by HW (HW does has the capability in case addr are not aligned)
2. How would split be related to this work ? It would be happening with or w/o this PR

andcarminati:
> > The regression is related to a load that was split in two halves (alignment?), and now cannot run in parallel.
> 
> CHECK :
> 
> 1. Is this the split by the compiler or by HW (HW does has the capability in case addr are not aligned)
> 2. How would split be related to this work ? It would be happening with or w/o this PR

Good observation @krishnamtibrewala, in fact my comment was not so clear. In this case, the loads were written in the source code in this way, the original scheduler output was (without any wait cycles):

```
VLDA wl2, [p2], m1; VLDB wh2, [p1, #32]; NOPXM
VLDA wl4, [p0]; VLDB wh4, [p0, #32] <- compile time information says that the first load is 32byte aligned
```

Perhaps, as alignment was 64byte for `p0` during execution, the second two loads were handled as just `VLD x4, [p0]`.

With this PR, we have:

```
NOPA; VLDB wl2, [p2], m1; NOPX
VLDA wl4, [p0]; VLDB wh2, [p1, #32]
VLDB wh4, [p0, #32]
```

krishnamtibrewala:
> during execution, the sec



> > > The regression is related to a load that was split in two halves (alignment?), and now cannot run in parallel.
> > 
> > 
> > CHECK :
> > 
> > 1. Is this the split by the compiler or by HW (HW does has the capability in case addr are not aligned)
> > 2. How would split be related to this work ? It would be happening with or w/o this PR
> 
> Good observation @krishnamtibrewala, in fact my comment was not so clear. In this case, the loads were written in the source code in this way, the original scheduler output was (without any wait cycles):
> 
> ```
> VLDA wl2, [p2], m1; VLDB wh2, [p1, #32]; NOPXM
> VLDA wl4, [p0]; VLDB wh4, [p0, #32] <- compile time information says that the first load is 32byte aligned
> ```
> 
> Perhaps, as alignment was 64byte for `p0` during execution, the second two loads were handled as just `VLD x4, [p0]`.
> 
> With this PR, we have:
> 
> ```
> NOPA; VLDB wl2, [p2], m1; NOPX
> VLDA wl4, [p0]; VLDB wh2, [p1, #32]
> VLDB wh4, [p0, #32]
> ```


Thanks for sharing the generated ASM snippet @andcarminati 

If we were previously executing `VLDA wl4, [p0]; VLDB wh4, [p0, #32]` then the core would have stalled for 1 cycle by itself, you can see that in the instr. profile report wait cycle must be > 0. This would have happened because we are trying to access the same bank twice in the same cycle.

Now in this PR we detects that and split that  `VLDA wl4, [p0];`  & `VLDB wh4, [p0, #32]` in two different VLIW bundle we would avoid that cycle stall which HW induces by itself. 

And if that's the only change in ASM we should not have seen regression, (We might see increase in PM, but only if the VLIW bundling had to have an extra instruction)

If we are seeing regression am speculating that we end up having more instructions(VLIW bundles) then the HW stall cycles. (We had seen similar thing happening where we say allowing bank conflicts helps us : WIP #229, Refer : https://github.com/Xilinx/llvm-aie/pull/229#issuecomment-2462671818)

andcarminati:
> If we were previously executing VLDA wl4, [p0]; VLDB wh4, [p0, #32] then the core would have stalled for 1 cycle by itself, you can see that in the instr. profile report wait cycle must be > 0. This would have happened because we are trying to access the same bank twice in the same cycle.

That is the interesting point, there is no stall related to those instructions. I guess it is related to some HW behavior.

llvm/lib/Target/AIE/AIEHazardRecognizer.cpp
```diff
@@ -98,20 +107,23 @@ void FuncUnitWrapper::blockResources() {
...
+  MemObjectsBits |= Other.MemObjectsBits;
   return *this;
 }
 
 bool FuncUnitWrapper::conflict(const FuncUnitWrapper &Other) const {
   if ((Required & Other.Required) != 0 || (Slots & Other.Slots) != 0 ||
       (MemoryBanks & Other.MemoryBanks) != 0 ||
-      (Reserved & Other.Required) != 0 || (Required & Other.Reserved) != 0) {
+      (Reserved & Other.Required) != 0 || (Required & Other.Reserved) != 0 ||
+      (MemObjectsBits & Other.MemObjectsBits) != 0) {
```
F-Stuckmann, line 126: If we do not have an addressspace, the memoryBankBits are 0. Would we still conflict?

llvm/lib/Target/AIE/AIEHazardRecognizer.cpp
```diff
@@ -447,13 +457,22 @@ auto toHazardType(bool Conflict) {
 }
 } // namespace
 
+ScheduleHazardRecognizer::HazardType AIEHazardRecognizer::getHazardType(
```
F-Stuckmann, line 461: I like this abstraction

llvm/test/CodeGen/AIE/schedule/pointer-hazard.mir
```diff
@@ -0,0 +1,201 @@
...
+
+# This file is licensed under the Apache License v2.0 with LLVM Exceptions.
+# See https://llvm.org/LICENSE.txt for license information.
+# SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+#
+# (c) Copyright 2025 Advanced Micro Devices, Inc. or its affiliates
+
+# RUN: llc --mtriple=aie2 --issue-limit=6 --start-before=postmisched \
+# RUN:     %s -o - | FileCheck %s
+
```
F-Stuckmann, line 11: NIT: Maybe add the commit message describing this unit test into the unit test

llvm/lib/Target/AIE/AIEHazardRecognizer.cpp
```diff
@@ -466,39 +485,44 @@ ScheduleHazardRecognizer::HazardType AIEHazardRecognizer::getHazardType(
...
-      MemoryBanks, TII->getMemoryCycles(SchedClass), DeltaCycles,
-      FUDepthLimit));
+      MemoryBanks, MemObjectsBits, TII->getMemoryCycles(SchedClass),
+      DeltaCycles, FUDepthLimit));
 }
 
 bool AIEHazardRecognizer::checkConflict(
     const ResourceScoreboard<FuncUnitWrapper> &Scoreboard, MachineInstr &MI,
-    int DeltaCycles) const {
+    int DeltaCycles) {
```
F-Stuckmann: I don't really undestand why checkConflict would modify the Hazard Recognizer. 

The const chain breaks at `MemoryObjectEnumerator::getObjectNumber(const Value *Object)` because we may update our `ParentObjectMap`. However shouldn't the object count be static after an initial sweep?

F-Stuckmann: I think we can get away with consting
`MemoryObjectEnumerator::getNumberOfMappedObjects()`

llvm/lib/Target/AIE/AIEHazardRecognizer.h
```diff
@@ -109,13 +117,24 @@ class FuncUnitWrapper {
   bool conflict(const FuncUnitWrapper &Other) const;
 };
 
+struct MemoryObjectEnumerator {
+  std::map<const Value *, unsigned> ObjectNumberingMap;
+  std::map<const Value *, const Value *> ParentObjectMap;
+  unsigned ObjectCounter = 0;
+
+  unsigned getNumberOfMappedObjects();
```
F-Stuckmann: NIT: `getNumberOfMappedObjects` be const

llvm/lib/Target/AIE/AIEHazardRecognizer.cpp
```diff
@@ -677,3 +725,59 @@ AIEHazardRecognizer::getMemoryBanks(const MachineInstr *MI) const {
...
+    return 0;
+
+  if (MI->memoperands_empty())
+    return 0; // optimistic: we will assume no conflict.
+
+  MemoryObjectsBits ObjectsToBitMap = 0;
+  for (auto &MMO : MI->memoperands()) {
+    const Value *BaseObject = MMO->getValue();
+    if (!BaseObject || ObjectEnumerator.getNumberOfMappedObjects() ==
+                           sizeof(MemoryObjectsBits))
```
F-Stuckmann: Could we have a debug message when we reach the limit of our MemoryObject Tracking capabilities?

llvm/lib/Target/AIE/AIEHazardRecognizer.cpp
```diff
@@ -677,3 +707,59 @@ AIEHazardRecognizer::getMemoryBanks(const MachineInstr *MI) const {
...
```
martien-de-jong: sizeof is in bytes, not bits.

llvm/lib/Target/AIE/AIEHazardRecognizer.cpp
```diff
@@ -677,3 +707,59 @@ AIEHazardRecognizer::getMemoryBanks(const MachineInstr *MI) const {
...
+      return 0; // optimistic: we will assume no conflict.
```
martien-de-jong: Probably not important, but it seems strange to break the loop here, and it is stranger not to return the already collected objects.

llvm/lib/Target/AIE/AIEHazardRecognizer.h
```diff
@@ -109,13 +117,24 @@ class FuncUnitWrapper {
   bool conflict(const FuncUnitWrapper &Other) const;
 };
 
+struct MemoryObjectEnumerator {
+  std::map<const Value *, unsigned> ObjectNumberingMap;
+  std::map<const Value *, const Value *> ParentObjectMap;
+  unsigned ObjectCounter = 0;
```
martien-de-jong, line 123: Could these be private/protected please?

llvm/lib/Target/AIE/AIEHazardRecognizer.cpp
```diff
@@ -466,39 +481,44 @@ ScheduleHazardRecognizer::HazardType AIEHazardRecognizer::getHazardType(
...
   if (EmissionCycle.conflict(Scoreboard[DeltaCycles]))
     return true;
 
-  // Verify memory bank hazards
+  // Verify memory bank and shared object hazards
   if (!MemoryAccessCycles.empty()) {
     FuncUnitWrapper MemoryBankAccessCycle(/*Req=*/0, /*Res=*/0, /*SlotSet=*/0,
                                           MemoryBanks);
+    FuncUnitWrapper ObjectAccessCycle(/*Req=*/0, /*Res=*/0, /*SlotSet=*/0,
+                                      /*MemoryBanks=*/0, MemObjectsBits);
```
martien-de-jong: I wouldn't mind combining MemoryBanks and MemOjectBits in one FuncUnitWrapper

llvm/lib/Target/AIE/AIEHazardRecognizer.cpp
```diff
@@ -461,6 +457,14 @@ auto toHazardType(bool Conflict) {
 }
 } // namespace
 
+ScheduleHazardRecognizer::HazardType AIEHazardRecognizer::getHazardType(
+    const ResourceScoreboard<FuncUnitWrapper> &TheScoreboard,
+    const MachineInstr *MI, const MCInstrDesc &Desc, int DeltaCycles) {
```
martien-de-jong: Desc is either MI->getDesc() or SelectedDesciptors.getDesc(MI). I think we should be able to make the last one work for a non MultiSlotPseudo as well. Then we can drop Desc from the interface

llvm/test/CodeGen/AIE/schedule/pointer-hazard.mir
martien-de-jong: it doesn't hurt. Note that --issue-limit isn't necessary for over 2 years.

llvm/lib/Target/AIE/AIEHazardRecognizer.cpp
andcarminati: Updated with a mutable object.

llvm/lib/Target/AIE/AIEHazardRecognizer.cpp
andcarminati: Done!

llvm/lib/Target/AIE/AIEHazardRecognizer.cpp
andcarminati: I think in this way it can help with debugs.

martien-de-jong: Well, you dump the cycle, and you can enable the scoreboard dump to see the actual conflict.

llvm/lib/Target/AIE/AIEHazardRecognizer.cpp
```diff
@@ -677,3 +707,65 @@ AIEHazardRecognizer::getMemoryBanks(const MachineInstr *MI) const {
...
+  MemoryObjectsBits ObjectsToBitMap = 0;
+  for (auto &MMO : MI->memoperands()) {
+    const Value *BaseObject = MMO->getValue();
+    if (!BaseObject) {
+      LLVM_DEBUG(dbgs() << "No Base Object!\n");
+      continue;
+    }
+
+    if (auto ObjectNumber = ObjectEnumerator.getObjectNumber(BaseObject))
+      ObjectsToBitMap |= 1 << *ObjectNumber;
```
martien-de-jong: The RHS is integer, and only carries 32 bits. I usually carry a `const uint64_t One = 1;` around for these purposes.

llvm/lib/Target/AIE/AIEAlternateDescriptors.h
```diff
@@ -55,6 +55,11 @@ class AIEAlternateDescriptors {
     return getSelectedDescriptor(MI).value_or(&MI->getDesc());
   }
 
+  const MCInstrDesc *getDesc(const MachineInstr *MI) const {
+    return getSelectedDescriptor(const_cast<MachineInstr *>(MI))
+        .value_or(&MI->getDesc());
```
martien-de-jong, line 60: I was hoping we could squeeze getDesc and getSelectedDescriptor together. Do we use the latter anywhere else?

llvm/lib/Target/AIE/AIEHazardRecognizer.cpp
```diff
@@ -677,3 +707,65 @@ AIEHazardRecognizer::getMemoryBanks(const MachineInstr *MI) const {
   }
   return MemoryBankUsed;
 }
+
+bool MemoryObjectEnumerator::isFull() const {
+  return ObjectCounter == sizeof(MemoryObjectsBits) * 8;
```
martien-de-jong: supernit: CHAR_BIT from `<climits>`

llvm/lib/Target/AIE/AIEHazardRecognizer.cpp
```diff
@@ -677,3 +707,65 @@ AIEHazardRecognizer::getMemoryBanks(const MachineInstr *MI) const {
...
+
+  auto ItParent = ParentObjectMap.find(Object);
+  if (ItParent != ParentObjectMap.end()) {
+    Object = ItParent->second;
+  } else {
+    const Value *ParentObject = getUnderlyingObject(Object);
+    ParentObjectMap[Object] = ParentObject;
+    Object = ParentObject;
+  }
+
```
martien-de-jong: You could also try to find the incoming value directly in the object number map, provided that you add it there with the parent's id.  You would converge to a single map lookup rather than two.

llvm/lib/Target/AIE/AIEHazardRecognizer.cpp
andcarminati: Done!

llvm/lib/Target/AIE/AIEHazardRecognizer.cpp
andcarminati: Nice idea, I simplified to a single map.

llvm/lib/Target/AIE/AIEAlternateDescriptors.h
andcarminati: Just two users now, getDesc and getDesc (const variant.)

martien-de-jong (approved):

PR #483, opened 2025-06-10
[llvm][aie] Use LLVM_DEBUG when printing debug info

newling:
Without this, there is the following linker error when building with `-DCMAKE_BUILD_TYPE=RelWithDebInfo` :

```
llvm-aie/llvm/lib/Target/AIE/AIEPostPipeliner.cpp:1138: undefined reference to `llvm::MachineInstr::dump() const'
```

A grep in `llvm/lib/Target` suggests wrapping in LLVM_DEBUG is the standard approach when calling dump. 

See https://github.com/llvm/llvm-project/blob/main/llvm/include/llvm/Support/Debug.h for macro info (and where dbgs() goes).

### DIFF
--- a/llvm/lib/Target/AIE/AIEPostPipeliner.cpp
+++ b/llvm/lib/Target/AIE/AIEPostPipeliner.cpp
@@ -1135,7 +1135,7 @@ void PostPipeliner::dump() const {
     const NodeInfo &Node = Info[I];
     dbgs() << I << " @" << Node.Cycle << " %" << Node.ModuloCycle << " S"
            << Node.Stage << " : ";
-    DAG->SUnits[I].getInstr()->dump();
+    LLVM_DEBUG(DAG->SUnits[I].getInstr()->dump());
   }
   PostPipelineDumper Dump;
   visitPipelineSchedule(Dump);


llvm/lib/Target/AIE/AIEPostPipeliner.cpp
```diff
@@ -1135,7 +1135,7 @@ void PostPipeliner::dump() const {
     const NodeInfo &Node = Info[I];
     dbgs() << I << " @" << Node.Cycle << " %" << Node.ModuloCycle << " S"
            << Node.Stage << " : ";
-    DAG->SUnits[I].getInstr()->dump();
+    LLVM_DEBUG(DAG->SUnits[I].getInstr()->dump());
```
martien-de-jong, line 1138: I guess this solves it. I wonder whether we can employ the dump method that is used by the mir serializer, which I hope isn't subject to NDEBUG.
I'm also wondering why our nightly build didn't catch this?

martien-de-jong (approved):

llvm/lib/Target/AIE/AIEPostPipeliner.cpp
newling: It did seem to solve it. But I've experimented a bit more this morning, and now removing the `LLVM_DEBUG` doesn't create a linker error. Anyway, thanks for merging. 

I see the commit message in the log isn't taken from the PR summary, not sure if I need to be careful to make better commit messages if contributing to peano?

PR #484, opened 2025-06-10
[AIE2] Alias Analysis using AddrSpace Info

martien-de-jong:
This is a port of Krishnam's original PR https://github.com/Xilinx/llvm-aie/pull/173 which I will close.
I'll be running QoR today-ish.

QoR results are varying. Also, memory bank annotations may not always be respected by the buffer allocation. Although it can be argued that intent of the programmer will usually be to have different objects be designated by pointers to non-overlapping banks, the gain we see doesn't make it worth to take the risk. I converted to draft.

### DIFF
--- a/llvm/lib/Target/AIE/AIEBaseAliasAnalysis.h
+++ b/llvm/lib/Target/AIE/AIEBaseAliasAnalysis.h
@@ -16,6 +16,7 @@
 
 #include "AIE.h"
 #include "llvm/Analysis/AliasAnalysis.h"
+#include "llvm/Analysis/TargetTransformInfo.h"
 
 namespace llvm {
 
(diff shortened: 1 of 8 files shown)


krishnamtibrewala:
Thank you @martien-de-jong for porting the PR. 
Wanted to give heads up, last I remember(long time back) there were no QoR gain. Therefore Gaetan suggested to hold onto it.

andcarminati:
Can we extent to aie2p? I see that it covers tile memory for aie2. Maybe it is not important...

llvm/lib/Target/AIE/AIEBaseAliasAnalysis.h
```diff
@@ -82,7 +86,12 @@ class AIEBaseAAWrapperPass : public ImmutablePass {
   const AIEBaseAAResult &getResult() const { return *Result; }
 
   bool doInitialization(Module &M) override {
-    Result.reset(new AIEBaseAAResult(M.getDataLayout()));
+    if (!M.getFunctionList().empty()) {
```
F-Stuckmann, line 89: If we do not have a function list in the module, shouldn't we still reset previous AliasResults`?

martien-de-jong: Theoretically yes. Although no one is going to query it in a module without functions. It seems to need TTI, which is normally found in the function. I wouldn't know where to find it otherwise.

Issue #480, opened 2025-06-05
`error in backend: unable to legalize instruction`


andrej:
Compiling the following minimal example with `-O2` causes the compiler to crash ("error in backend: unable to legalize instruction") with the below attached stack trace. Annotating function `bar` with `__attribute__((noinline))` or compiling with `-O1` circumvents the issue.

```
$ cat mwe.cc 
#include <aie_api/aie.hpp>
#include <cstdint>

void bar(int8_t *in_ptr, int8_t *out_ptr) {
  aie::vector<int8_t, 16> zeroes = aie::zeros<int8_t, 16>();
  aie::store_v(out_ptr, zeroes);
}

void baz(int8_t *in_ptr, int8_t *out_ptr) {
  for (unsigned i = 0; i < 3; i++) {
    for (unsigned j = 0; j < 2; j++) {
      for (unsigned k = 0; k < 1; k++) {
        aie::vector<int8_t, 16> result = aie::add(aie::load_v<16>(in_ptr), aie::zeros<int8_t, 16>());
        aie::store_v(out_ptr, result);
      }
    }
  }
}

void foo(int8_t *in_ptr, int8_t *out_ptr) {
  bar(out_ptr, in_ptr);
  baz(in_ptr, out_ptr);
}
```

Error/backtrace:
```
$ ${PEANO_INSTALL_DIR}/bin/clang++ -O2 -std=c++20 --target=aie2-none-unknown-elf -Wno-parentheses -Wno-attributes -Wno-macro-redefined -Wno-empty-body -DNDEBUG -I /scratch/roesti/mlir-aie/install/include -c mwe.cc
fatal error: error in backend: unable to legalize instruction: %14:_(<64 x s8>) = G_SHL %4:_, %12:_(<64 x s8>) (in function: _Z3fooPaS_)
PLEASE submit a bug report to https://github.com/llvm/llvm-project/issues/ and include the crash backtrace, preprocessed source, and associated run script.
Stack dump:
0.      Program arguments: /scratch/roesti/buildenv/lib/python3.10/site-packages/llvm-aie/bin/clang++ -O2 -std=c++20 --target=aie2-none-unknown-elf -Wno-parentheses -Wno-attributes -Wno-macro-redefined -Wno-empty-body -DNDEBUG -I /scratch/roesti/mlir-aie/install/include -c mwe.cc
1.      <eof> parser at end of file
2.      Code generation
[...]

PR #485, opened 2025-06-10
[AIEX] Combine G_SHL (vector, amt) into G_ADD

andcarminati:
This combiner solves the issue https://github.com/Xilinx/llvm-aie/issues/480

This is intended to transform some G_SHL vector operations to a form that can be legalized after.


llvm/test/CodeGen/AIE/GlobalISel/combine-broadcast-shl-add.mir
```diff
@@ -0,0 +1,213 @@
...
+    %2:_(<64 x s8>) = G_AIE_BROADCAST_VECTOR %1(s32)
+    %3:_(<64 x s8>) = G_SHL %0, %2(<64 x s8>)
+    PseudoRET implicit $lr, implicit %3
+
+...
+
+...
+---
+name:            broadcast_shl_by_4
+body:             |
```
krishnamtibrewala, line 101: Hi @andcarminati, after a point using "MUL" instruction is a better option.

andcarminati: Actually, G_MUL is only legal for scalar types. We can only reach the VMUL through intrinsic call now. I honestly thing that this is a "safe net" against some very uncommon cases.

krishnamtibrewala: Oh, yes you are right!

llvm/test/CodeGen/AIE/GlobalISel/combine-broadcast-shl-add.mir
```diff
@@ -0,0 +1,213 @@
...
+name:            broadcast_shl_by_9_bitcast
+body:             |
+  bb.1.entry:
+    liveins: $x0
+
+    ; CHECK-LABEL: name: broadcast_shl_by_9_bitcast
+    ; CHECK: liveins: $x0
+    ; CHECK-NEXT: {{  $}}
+    ; CHECK-NEXT: [[COPY:%[0-9]+]]:_(<32 x s16>) = COPY $x0
+    ; CHECK-NEXT: [[ADD:%[0-9]+]]:_(<32 x s16>) = G_ADD [[COPY]], [[COPY]]
```
F-Stuckmann, line 215: could this be a loop so that the code does not blow up ?

andcarminati: I believe that this is not feasible here in the combiner. To do that we need a separated pass. Also, we need extensions  to generate ZOL, because now we just lower them in different steps.

llvm/lib/Target/AIE/AIECombinerHelper.cpp
```diff
@@ -3265,3 +3265,49 @@ bool llvm::matchPairedExtracts(MachineInstr &MI, MachineRegisterInfo &MRI,
...
+    return false;
+
+  auto CstSrc =
+      getIConstantVRegValWithLookThrough(DefAmtMI->getOperand(1).getReg(), MRI);
+
+  if (!CstSrc)
+    return false;
+
+  const int ShiftAmount = CstSrc->Value.getZExtValue();
+  assert(ShiftAmount > 0 && "Invalid shift amount");
```
martien-de-jong: >= ?

andcarminati: I changed and included another test.

martien-de-jong (approved):

PR #489, opened 2025-06-13
Stuckmann.loop.optimization.remarks

F-Stuckmann:
Adding Pipeline and Loop Information to optimization remarks for automated metric extractions.
Metrics include:
- NS
- II
- Number of Bundles in MBB
- Hardware Loop
- All MBBs in a Loop

### DIFF
--- a/llvm/lib/CodeGen/MachinePipeliner.cpp
+++ b/llvm/lib/CodeGen/MachinePipeliner.cpp
@@ -2527,7 +2527,9 @@ bool SwingSchedulerDAG::schedulePipeline(SMSchedule &Schedule) {
              << "Schedule found with Initiation Interval: "
              << ore::NV("II", Schedule.getInitiationInterval())
              << ", MaxStageCount: "
-             << ore::NV("MaxStageCount", Schedule.getMaxStageCount());
+             << ore::NV("MaxStageCount", Schedule.getMaxStageCount())
+             << ore::NV("NS", Schedule.getMaxStageCount() + 1)
+             << ore::NV("BasicBlock", BB->getName());
     });
   } else
     Schedule.reset();
--- a/llvm/lib/Target/AIE/AIE2AsmPrinter.cpp
+++ b/llvm/lib/Target/AIE/AIE2AsmPrinter.cpp
@@ -4,7 +4,7 @@
 // See https://llvm.org/LICENSE.txt for license information.
 // SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
 //
-// (c) Copyright 2023-2024 Advanced Micro Devices, Inc. or its affiliates
+// (c) Copyright 2023-2025 Advanced Micro Devices, Inc. or its affiliates
 //
 //===----------------------------------------------------------------------===//
 //
@@ -16,12 +16,16 @@
 #include "AIE2AsmPrinter.h"
 #include "AIE2TargetMachine.h"
 #include "InstPrinter/AIE2InstPrinter.h"
+#include "llvm/Analysis/OptimizationRemarkEmitter.h"
 #include "llvm/CodeGen/AsmPrinter.h"
 #include "llvm/CodeGen/MachineInstr.h"
+#include "llvm/CodeGen/MachineOptimizationRemarkEmitter.h"
 #include "llvm/MC/MCStreamer.h"
 #include "llvm/Support/raw_ostream.h"
 
 using namespace llvm;
+
+#define DEBUG_TYPE "aie-asm-printer"
 
 // Simple pseudo-instructions have their lowering (with expansion to real
 // instructions) auto-generated.
@@ -68,6 +72,35 @@ bool AIE2AsmPrinter::PrintAsmMemoryOperand(const MachineInstr *MI,
 bool AIE2AsmPrinter::lowerOperand(const MachineOperand &MO,
                                   MCOperand &MCOp) const {
   return LowerAIEMachineOperandToMCOperand(MO, MCOp, *this);
+}
+
+void AIE2AsmPrinter::emitBundleCount(const MachineBasicBlock &MBB) {
+  unsigned BundleCount = 0;
+  for (auto &MI : MBB) {
+    if (!MI.isBundle())
+      continue;
+
+    BundleCount++;
+  }
+
+  if (BundleCount == 0) {
+    // encountered empty MBB, no need to dump Bundle info, this could be an
+    // empty back-edge
+    return;
+  }
+
+  ORE->emit([&]() {
+    return MachineOptimizationRemarkAnalysis(DEBUG_TYPE, "analysis",
+                                             MBB.begin()->getDebugLoc(), &MBB)
+           << ore::NV("BasicBlock", MBB.getName())
+           << ore::NV("BundleCount", BundleCount);
+  });
+}
+
+void AIE2AsmPrinter::emitBasicBlockStart(const MachineBasicBlock &MBB) {
+  emitBundleCount(MBB);
+
+  AsmPrinter::emitBasicBlockStart(MBB);
 }
 
 AsmPrinter *
--- a/llvm/lib/Target/AIE/AIE2AsmPrinter.cpp
+++ b/llvm/lib/Target/AIE/AIE2AsmPrinter.cpp
@@ -76,11 +76,15 @@ bool AIE2AsmPrinter::lowerOperand(const MachineOperand &MO,
 
 void AIE2AsmPrinter::emitBundleCount(const MachineBasicBlock &MBB) {
   unsigned BundleCount = 0;
+  unsigned ByteCount = 0;
+  auto *TII = static_cast<const AIEBaseInstrInfo *>(
+      MBB.getParent()->getSubtarget().getInstrInfo());
   for (auto &MI : MBB) {
     if (!MI.isBundle())
       continue;
 
     BundleCount++;
+    ByteCount += TII->getAIEMachineBundleSize(MI);
   }
 
   if (BundleCount == 0) {
@@ -93,7 +97,8 @@ void AIE2AsmPrinter::emitBundleCount(const MachineBasicBlock &MBB) {
     return MachineOptimizationRemarkAnalysis(DEBUG_TYPE, "analysis",
                                              MBB.begin()->getDebugLoc(), &MBB)
            << ore::NV("BasicBlock", MBB.getName())
-           << ore::NV("BundleCount", BundleCount);
+           << ore::NV("BundleCount", BundleCount)
+           << ore::NV("ByteCount", ByteCount);
   });
 }
 
--- a/llvm/lib/Target/AIE/AIE2AsmPrinter.h
+++ b/llvm/lib/Target/AIE/AIE2AsmPrinter.h
@@ -21,6 +21,9 @@
 namespace llvm {
 
 class AIE2AsmPrinter : public AIEBaseAsmPrinter {
+  // Dump Bundle Count to Optimization Remarks
+  void emitBundleCount(const MachineBasicBlock &MBB);
+
 public:
   explicit AIE2AsmPrinter(TargetMachine &TM,
                           std::unique_ptr<MCStreamer> Streamer)
@@ -38,6 +41,8 @@ public:
 
   // Wrapper needed for tblgenned pseudo lowering.
   bool lowerOperand(const MachineOperand &MO, MCOperand &MCOp) const;
+
+  void emitBasicBlockStart(const MachineBasicBlock &MBB) override;
 };
 
 AsmPrinter *createAIE2AsmPrinterPass(TargetMachine &TM,
--- a/llvm/lib/Target/AIE/AIEBaseHardwareLoops.cpp
+++ b/llvm/lib/Target/AIE/AIEBaseHardwareLoops.cpp
@@ -4,7 +4,7 @@
 // See https://llvm.org/LICENSE.txt for license information.
 // SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
 //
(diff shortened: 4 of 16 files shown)


llvm/lib/Target/AIE/AIE2AsmPrinter.cpp
```diff
@@ -70,6 +74,34 @@ bool AIE2AsmPrinter::lowerOperand(const MachineOperand &MO,
   return LowerAIEMachineOperandToMCOperand(MO, MCOp, *this);
 }
 
+void AIE2AsmPrinter::emitBundleCount(const MachineBasicBlock &MBB) {
+  unsigned BundleCount = 0;
+  for (auto &MI : MBB) {
+    if (!MI.isBundle())
```
andcarminati, line 83: Is this a standalone instruction? Should it be accounted as a full bundle because it takes one cycle?

llvm/test/CodeGen/AIE/aie2p/end-to-end/ore-hardware-loops.ll
```diff
@@ -0,0 +1,149 @@
+;
+; This file is licensed under the Apache License v2.0 with LLVM Exceptions.
+; See https://llvm.org/LICENSE.txt for license information.
+; SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+;
+; (c) Copyright 2025 Advanced Micro Devices, Inc. or its affiliates
+; RUN: llc -mtriple=aie2p --aie-force-postpipeliner \
```
andcarminati, line 7: Do we really need `--aie-force-postpipeliner` here?

llvm/lib/Target/AIE/AIEBaseHardwareLoops.cpp
```diff
@@ -73,6 +125,9 @@ class LowOverheadLoop {
       Preheader = MBB;
   }
   ~LowOverheadLoop() {
+    // dump the loop ID here, otherwise hardware loop detection is wrong
+    dumpLoopID();
```
andcarminati, line 123: Are we dumping this always? I mean, this very important for us (QoR) could be optional to other compiler users because of compile time.

llvm/lib/Target/AIE/AIE2AsmPrinter.cpp
```diff
@@ -70,6 +74,34 @@ bool AIE2AsmPrinter::lowerOperand(const MachineOperand &MO,
...
+
+  ORE->emit([&]() {
+    return MachineOptimizationRemarkAnalysis(DEBUG_TYPE, "analysis",
+                                             MBB.begin()->getDebugLoc(), &MBB)
+           << ore::NV("BasicBlock", MBB.getName())
+           << ore::NV("INST_BUNDLE", BundleCount);
+  });
+}
+
+void AIE2AsmPrinter::emitBasicBlockStart(const MachineBasicBlock &MBB) {
```
andcarminati, line 105: Should we implement this in the base class? It can help some refactoring in the future.

llvm/lib/Target/AIE/AIE2AsmPrinter.h
```diff
@@ -21,6 +21,9 @@
 namespace llvm {
 
 class AIE2AsmPrinter : public AIEBaseAsmPrinter {
+  // Dump Bundle Count to Optimization Remarks
```
andcarminati, line 24: Maybe a test?

llvm/lib/Target/AIE/AIEBaseHardwareLoops.cpp
F-Stuckmann: if clang is not called with `-fsave-optimization-record` , we don't actually dump anything.
We always seem to dump results into optimization remarks, even though nobody may read it

llvm/lib/Target/AIE/AIE2AsmPrinter.cpp
F-Stuckmann: No the MI is not a standalone instruction.
The MI here is a bundle (which may contain multiple instructions) or non Bundle instructions, such as labels

llvm/lib/Target/AIE/AIE2AsmPrinter.cpp
F-Stuckmann: confusingly enough, the baseASMPrinter is for aie1, while aie2asmPrinter is for aie2 and aie2p. 

This smells like a nice refactor opportunity.

llvm/test/CodeGen/AIE/aie2p/end-to-end/ore-hardware-loops.ll
F-Stuckmann: I extend the unit-test to also do the bundle counting. Since it the unit-test is a copy from  `conv2d_bfp16_convert.ll `, I want to have the same behaviour (also pipeline)

llvm/lib/Target/AIE/AIE2AsmPrinter.h
F-Stuckmann: I added a unit-test in `ore-hardware-loops.ll`

llvm/lib/CodeGen/MachinePipeliner.cpp
```diff
@@ -2527,7 +2527,8 @@ bool SwingSchedulerDAG::schedulePipeline(SMSchedule &Schedule) {
              << "Schedule found with Initiation Interval: "
              << ore::NV("II", Schedule.getInitiationInterval())
              << ", MaxStageCount: "
-             << ore::NV("MaxStageCount", Schedule.getMaxStageCount());
+             << ore::NV("MaxStageCount", Schedule.getMaxStageCount())
```
andcarminati, line 2530: I think that this could be inaccurate to us, because we are way behind the final schedule/RA.

andcarminati: Also, `getMaxStageCount` will give us actually the prologue count, that is `StageCount - 1`.

martien-de-jong, line 2530: I wouldn't mind cooking our own MORE in shouldUseSchedule, e.g. under the tag "prepipeliner"

llvm/lib/Target/AIE/AIEBaseHardwareLoops.cpp
```diff
@@ -56,6 +58,56 @@ class LowOverheadLoop {
...
+    MachineOptimizationRemarkEmitter ORE(*ML.getHeader()->getParent(), nullptr);
+    ORE.emit([&]() {
+      auto RE = MachineOptimizationRemarkAnalysis(
+                    DEBUG_TYPE, "analysis",
+                    ML.getHeader()->begin()->getDebugLoc(), ML.getHeader())
+                << ore::NV("LoopID", LoopID);
+
+      for (auto *MBB : ML.blocks()) {
+        RE << ore::NV("BasicBlock", MBB->getName());
+      }
```
martien-de-jong, line 73: CHECK: I think this would list blocks multiple times, for all loops a block is nested in. That's fine as long as we are aware when totaling metrics.

llvm/lib/Target/AIE/AIEBaseHardwareLoops.cpp
```diff
@@ -56,6 +58,56 @@ class LowOverheadLoop {
...
+
+      RE << ore::NV("Hardware-Loop", foundAllComponents());
```
martien-de-jong: nit: Perhaps Zero-Overhead-Loop?

llvm/lib/Target/AIE/AIEBaseHardwareLoops.cpp
```diff
@@ -56,6 +58,56 @@ class LowOverheadLoop {
...
+
+      sort(MachineLoops, [](MachineLoop *ML1, MachineLoop *ML2) {
+        return ML1->getHeader()->getNumber() < ML2->getHeader()->getNumber();
+      });
+
+      LoopCount = std::distance(MachineLoops.begin(), find(MachineLoops, ML));
+      return std::to_string(LoopCount) + Suffix;
+    }
+
+    for (auto *L : ML->getParentLoop()->getSubLoops()) {
```
martien-de-jong: I'm surprised that these seem to be sorted whereas the top level loops aren't. I think both aren't. In both cases I think you can find the index by counting the number of loops at the parent level that have a number less than the one you're searching for.

llvm/test/CodeGen/AIE/aie2p/end-to-end/ore-nested-loops.ll
```diff
@@ -0,0 +1,315 @@
...
+; CHECK-NEXT: Pass:            aie-hardware-loops
+; CHECK-NEXT: Name:            analysis
+; CHECK-NEXT: Function:        nestedLoops
+; CHECK-NEXT: Args:
+; CHECK-NEXT:   - LoopID:          '0'
+; CHECK-NEXT:   - BasicBlock:              for.cond
+; CHECK-NEXT:   - BasicBlock:              for.body
+; CHECK-NEXT:   - BasicBlock:              for.inc
+; CHECK-NEXT:   - Hardware-Loop:   'false'
+; CHECK-NEXT: ...
```
martien-de-jong, line 112: I'd like to have one example where a nested loop is a hardware loop, so that we check that it gets attributed to the right loop.

llvm/lib/Target/AIE/AIE2AsmPrinter.cpp
```diff
@@ -70,6 +74,34 @@ bool AIE2AsmPrinter::lowerOperand(const MachineOperand &MO,
...
+  for (auto &MI : MBB) {
+    if (!MI.isBundle())
+      continue;
+
+    BundleCount++;
+  }
+
+  if (BundleCount == 0) {
+    // encountered empty MBB, no need to dump Bundle info
+    return;
```
martien-de-jong, line 93: I think this throws away information. I would like to know where we have empty blocks, since it should have been eliminated.

llvm/lib/Target/AIE/AIE2AsmPrinter.cpp
```diff
@@ -70,6 +74,34 @@ bool AIE2AsmPrinter::lowerOperand(const MachineOperand &MO,
...
+  }
+
+  ORE->emit([&]() {
+    return MachineOptimizationRemarkAnalysis(DEBUG_TYPE, "analysis",
+                                             MBB.begin()->getDebugLoc(), &MBB)
+           << ore::NV("BasicBlock", MBB.getName())
+           << ore::NV("INST_BUNDLE", BundleCount);
```
martien-de-jong: I think INST_BUNDLE isn't the most descriptive of names.

martien-de-jong: Can we find the true size of the block?

martien-de-jong (commented):
Mostly good.

llvm/lib/Target/AIE/AIEBaseHardwareLoops.cpp
F-Stuckmann: correct

llvm/lib/Target/AIE/AIEBaseHardwareLoops.cpp
F-Stuckmann: I now count all the MBBs, that have a smaller Header BasicBlock Number

llvm/lib/CodeGen/MachinePipeliner.cpp
F-Stuckmann: It is a lot more hassle to move it to shouldUseSchedule, than to leave it here.  The abstraction layers don't really align for a clean move.

llvm/lib/Target/AIE/AIE2AsmPrinter.cpp
F-Stuckmann: If we have an empty back-edge, we get this case. An example is `for.end27` in `ore-nested-loops.ll`. 
We have to abort, otherwise the Optimization Remark causes a segmentation fault

llvm/lib/Target/AIE/AIE2AsmPrinter.cpp
F-Stuckmann: yes I can add the block size

llvm/lib/Target/AIE/AIEPostPipeliner.cpp
```diff
@@ -1052,8 +1053,8 @@ bool PostPipeliner::schedule(ScheduleDAGMI &TheDAG, int InitiationInterval,
 
   More.emit([&]() {
     return MachineOptimizationRemark("postpipeliner", "schedule", DbgLoc, BB)
-           << "Schedule found: NS=" << ore::NV("NS", NStages)
-           << " II=" << ore::NV("II", II);
```
martien-de-jong, line 1056: I borrowed this idiom from somewhere else, I think that if you echo the diagnostics on stderr it will be more readable.

llvm/lib/Target/AIE/AIEBaseInstrInfo.cpp
```diff
@@ -472,10 +472,25 @@ AIEBaseInstrInfo::getAIEMachineBundle(MachineBasicBlock::iterator MII) const {
   return Bundle;
 }
 
+const AIE::ConstMachineBundle AIEBaseInstrInfo::getAIEMachineBundle(
+    const MachineBasicBlock::const_iterator MII) const {
+
+  AIE::ConstMachineBundle Bundle = (getFormatInterface());
```
martien-de-jong, line 478: ??? I didn't know this was legal. it seems to infer the rhs type downwards from the assignment operator? I would prefer a normal constructor call (drop the = sign)

martien-de-jong (approved):
Some nits, nothing important enough to block.

PR #490, opened 2025-06-13
[AIEX] Outline GEPs

F-Stuckmann:
Currently the IRTranslator inserts G_PTR_ADDs of inlined GEPs at the top of the MBB, instead of right before the inlined-gep instruction.
So a check has to be performed, if these G_PTR_ADDs can be moved down to the instruction using them first.

This is a solves a similar issue like the previous gep outline in the IRTranslator.


llvm/test/CodeGen/AIE/GlobalISel/irtranslator-outline-memory-gep.ll
```diff
@@ -0,0 +1,214 @@
...
+; See https://llvm.org/LICENSE.txt for license information.
+; SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+;
+; (c) Copyright 2025 Advanced Micro Devices, Inc. or its affiliates
+; unit test for the loop iteration metadata conversion to assumptions
+;
+; NOTE: Example File to reverse inlined GEP
+;
+
+; RUN: llc -mtriple aie2 --stop-after=irtranslator -verify-machineinstrs -o - %s | FileCheck %s
```
andcarminati, line 13: I think we can run just the new pass so we can se the expected behavior.

andcarminati:
I spotted one warning:
```
[187/312] Building CXX object lib/Target/AIE/CMakeFiles/LLVMAIECodeGen.dir/AIEOutlineMemoryGEP.cpp.o
../llvm/lib/Target/AIE/AIEOutlineMemoryGEP.cpp:120:2: warning: extra ';' outside of a function is incompatible with C++98 [-Wc++98-compat-extra-semi]
}; // namespace
 ^
```

llvm/lib/Target/AIE/AIEOutlineMemoryGEP.cpp
```diff
@@ -0,0 +1,139 @@
...
+
+  insertNewGEPInst(*PointerOperandIdx, I, *GEP);
+}
+
+void insertNewGEPInst(const unsigned OpIdx, Instruction &InsertionPoint,
+                      GEPOperator &GEP) {
+  const std::vector<Value *> Indices(GEP.idx_begin(), GEP.idx_end());
+
+  GetElementPtrInst *NewGEPInstr = GetElementPtrInst::Create(
+      GEP.getSourceElementType(), GEP.getPointerOperand(), Indices,
```
martien-de-jong, line 83: `GEP.idx_begin(), GEP.idx_end()` instead of `Indices` should work here.

martien-de-jong (dismissed):

martien-de-jong (approved):

PR #491, opened 2025-06-16
[AIEX] Insert MMOs on lookup loads for non-trackable pointers

andcarminati:
Under some circumstances:
* Loaded value.
* Copied from physical register.

### DIFF
--- a/llvm/lib/Target/AIE/AIEPostSelectOptimize.cpp
+++ b/llvm/lib/Target/AIE/AIEPostSelectOptimize.cpp
@@ -48,6 +48,7 @@
 #include "llvm/CodeGen/MachineRegisterInfo.h"
 #include "llvm/CodeGen/TargetInstrInfo.h"
 #include "llvm/CodeGen/TargetPassConfig.h"
+#include "llvm/IR/GlobalValue.h"
 #include "llvm/Support/Debug.h"
 #include "llvm/Support/ErrorHandling.h"
 #include <optional>
@@ -60,6 +61,10 @@ using namespace llvm;
 static cl::opt<bool>
     EnablePostSelectOptimize("aie-post-select-opt", cl::Hidden, cl::init(true),
                              cl::desc("Enable post select optimize."));
+
+static cl::opt<bool> Assume4xLoadReadOnlyData(
+    "assume-4xload-ro-data", cl::Hidden, cl::init(true),
+    cl::desc("Assumes that the 4x load instructions access read-only data"));
 
 namespace {
 
@@ -490,25 +495,54 @@ using Operation = AIEBaseInstrInfo::AbstractOp::OperationType;
 using AbstractOp = AIEBaseInstrInfo::AbstractOp;
 using OptionalOp = std::optional<AbstractOp>;
 
-bool getGlobalValue(const MachineInstr *MI,
-                    SmallPtrSet<const Value *, 4> &GVSet,
+// Verify if an instruction has a GlobalValue (GV) as operand. In case of
+// true result, add such GV to the GVSet. Under some circumstances, such
+// GV operand can also be virtually retrieved.
+bool hasGlobalValue(MachineInstr *MI, SmallPtrSet<const Value *, 4> &GVSet,
                     MachineRegisterInfo &MRI) {
 
   MI = getDefIgnoringCopiesAndBitcasts(MI->getOperand(1).getReg(), MRI);
 
   // We need an instruction that explicitly moves a global
   // to a register (move immediate).
-  if (MI->getNumOperands() != 2)
-    return false;
+  if (MI->isMoveImmediate()) {
+    const MachineOperand *MO = MI->uses().begin();
 
-  const MachineOperand *MO = MI->uses().begin();
-
-  if (MO->isGlobal()) {
-    GVSet.insert(MO->getGlobal());
-    return true;
+    if (MO->isGlobal()) {
+      GVSet.insert(MO->getGlobal());
+      return true;
+    }
   }
 
-  return false;
+  if (!Assume4xLoadReadOnlyData)
+    return false;
+
+  // In the context of memory access, we cannot definitively determine the
+  // original value beyond this point (it may reside in another CU, for
+  // example). However, since the affected instructions are intended for
+  // accessing read-only tables, we can optimistically assign a common
+  // artificial value that symbolizes a fictional global variable. Be mindful
+  // that if these instructions are used to access read/write data or exceed the
+  // bounds of the allocated table, the behavior will be undefined.
+  //
+  // Furthermore, we can address two additional scenarios:
+  // 1. Loading the pointer from an external source.
+  // 2. Copying the pointer from a physical register.
+  // By limiting our focus in this manner, we streamline the search space to a
+  // well-known structure.
+  if (!MI->mayLoad() && !MI->isCopy())
+    return false;
+
+  if (MI->isCopy() && !MI->getOperand(1).getReg().isPhysical())
+    return false;
+
+  // This is a transient variable that will be discarded in the end.
+  Module &M = *MI->getMF()->getFunction().getParent();
+  auto *GV = M.getOrInsertGlobal("__aie_common_lookup_table",
+                                 PointerType::getUnqual(M.getContext()));
+
+  GVSet.insert(GV);
+  return true;
 }
 
 // Recognize BROADCAST (GLOBAL)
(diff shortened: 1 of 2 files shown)


andcarminati:
This is intended to regain the pipelined loop in GeluTemplated:

image

llvm/test/CodeGen/AIE/aie2p/GlobalIsel/propagate-mmo-noptr.mir
```diff
@@ -434,3 +436,200 @@ body:             |
...
+    %148:vec256 = VLDB_4x16_lo %147
+    %150:vec256 = VLDB_4x16_hi %147
+    $wl0 = COPY %142
+    $wl1 = COPY %145
+    $wl2 = COPY %148
+    $wl3 = COPY %150
+    PseudoRET implicit $lr
+...
+
+# Case 8: here we have the common case (ADD (SEL (BROADCAST(G), BROADCAST(G)))) but with
```
F-Stuckmann, line 538: where is the difference between case 7 and 8? The comments seem to suggest, they are equal

andcarminati: Oh, the comments should be updated indeed.

llvm/lib/Target/AIE/AIEPostSelectOptimize.cpp
```diff
@@ -489,25 +495,50 @@ using Operation = AIEBaseInstrInfo::AbstractOp::OperationType;
...
+bool getGlobalValue(MachineInstr *MI, SmallPtrSet<const Value *, 4> &GVSet,
                     MachineRegisterInfo &MRI) {
 
   MI = getDefIgnoringCopiesAndBitcasts(MI->getOperand(1).getReg(), MRI);
 
   // We need an instruction that explicitly moves a global
   // to a register (move immediate).
-  if (MI->getNumOperands() != 2)
+  if (MI->getNumOperands() == 2) {
+    const MachineOperand *MO = MI->uses().begin();
```
F-Stuckmann, line 509: how do we guarantee, that the MI is a move?

llvm/lib/Target/AIE/AIEPostSelectOptimize.cpp
```diff
@@ -489,25 +495,50 @@ using Operation = AIEBaseInstrInfo::AbstractOp::OperationType;
 using AbstractOp = AIEBaseInstrInfo::AbstractOp;
 using OptionalOp = std::optional<AbstractOp>;
 
-bool getGlobalValue(const MachineInstr *MI,
-                    SmallPtrSet<const Value *, 4> &GVSet,
+bool getGlobalValue(MachineInstr *MI, SmallPtrSet<const Value *, 4> &GVSet,
```
F-Stuckmann: NIT: add a comment describing this function. Wouldn't `hasGlobalValue` make more sense for the function name?

llvm/lib/Target/AIE/AIEPostSelectOptimize.cpp
andcarminati: Basically if an instruction has two operands ad one is a global, it should be a move immediate, because the global will be handled by the linker.

llvm/lib/Target/AIE/AIEPostSelectOptimize.cpp
andcarminati: You are right, it makes sense! I will update.

llvm/lib/Target/AIE/AIEPostSelectOptimize.cpp
```diff
@@ -490,25 +496,53 @@ using Operation = AIEBaseInstrInfo::AbstractOp::OperationType;
...
+    return false;
 
-  if (MO->isGlobal()) {
-    GVSet.insert(MO->getGlobal());
-    return true;
-  }
+  assert((!MI->isCopy() || MI->getOperand(1).getReg().isPhysical()) &&
+         "Not a copy from physical register!");
+  // This is a transient variable that will be discarded in the end.
+  Module &M = *MI->getParent()->getParent()->getFunction().getParent();
```
martien-de-jong: nit: MI->getMF() instead of two getParents() ?

llvm/lib/Target/AIE/AIEPostSelectOptimize.cpp
martien-de-jong: I think we check that it is a copy before. But then, the implementation is a bit strange, since it checks rather then asserts the operand count, and it only checks the one use operand. In that sense it doesn't do what the comment says.

llvm/lib/Target/AIE/AIEPostSelectOptimize.cpp
```diff
@@ -61,6 +62,11 @@ static cl::opt<bool>
     EnablePostSelectOptimize("aie-post-select-opt", cl::Hidden, cl::init(true),
                              cl::desc("Enable post select optimize."));
 
+static cl::opt<bool> AllowDummyGlobalsInLookupInstr(
+    "aie-use-dummy-globals-lookup", cl::Hidden, cl::init(true),
+    cl::desc("Enable the usage of dummy globals in lookup loads (when pointers"
+             "cannot be tracked)."));
```
martien-de-jong: The description/name could be stronger. I think this effectively assumes that the 4x load instructions access read-only data. How we implement that (insert an 'anonymous' object  that can not be returned by a write designation) is less important.
We may also want to have a user annotation that makes that assumption explicit.

llvm/lib/Target/AIE/AIEPostSelectOptimize.cpp
```diff
@@ -490,25 +496,53 @@ using Operation = AIEBaseInstrInfo::AbstractOp::OperationType;
...
-  }
+  assert((!MI->isCopy() || MI->getOperand(1).getReg().isPhysical()) &&
+         "Not a copy from physical register!");
+  // This is a transient variable that will be discarded in the end.
+  Module &M = *MI->getParent()->getParent()->getFunction().getParent();
+  auto *GV = M.getOrInsertGlobal("__aie_common_lookup_table",
+                                 PointerType::getUnqual(M.getContext()));
 
-  return false;
+  GVSet.insert(GV);
```
martien-de-jong, line 544: Check: GVSet is local, and the insert above is an insertion in the module data.

llvm/lib/Target/AIE/AIEPostSelectOptimize.cpp
andcarminati: Explicitly checking for a move immediate now.

llvm/lib/Target/AIE/AIEPostSelectOptimize.cpp
```diff
@@ -61,6 +62,10 @@ static cl::opt<bool>
     EnablePostSelectOptimize("aie-post-select-opt", cl::Hidden, cl::init(true),
                              cl::desc("Enable post select optimize."));
 
+static cl::opt<bool> Assume4xLoadReadOnlyData(
+    "assume-4xload-ro-data", cl::Hidden, cl::init(true),
```
F-Stuckmann, line 66: nit: we are never actually checking, if we are loading 4 times. 
maybe: `assume-global-ro-data` ?

andcarminati: This is related to the type of instruction that we are targeting, that uses a vector containing 4 pointers to load 4 values. We always load 4 values.

llvm/lib/Target/AIE/AIEPostSelectOptimize.cpp
```diff
@@ -490,25 +495,53 @@ using Operation = AIEBaseInstrInfo::AbstractOp::OperationType;
...
+  // well-known structure.
+  if (!MI->mayLoad() && !MI->isCopy())
+    return false;
 
-  if (MO->isGlobal()) {
-    GVSet.insert(MO->getGlobal());
-    return true;
-  }
+  assert((!MI->isCopy() || MI->getOperand(1).getReg().isPhysical()) &&
+         "Not a copy from physical register!");
```
martien-de-jong: Eh, you're asserting that it is not a copy and the failure string says it's not a copy?

martien-de-jong (dismissed):
Looks good to me. I guess the only controversial thing is the setting of the default, but I guess we'll think of plan B once we see it. However, we may want to document this to the mllib team.

llvm/lib/Target/AIE/AIEPostSelectOptimize.cpp
andcarminati: We are asserting that we have a "copy from physical register".

andcarminati: I was thinking about converting this assert to a return false. We don't need to crash if we don't meet the requirements, what do you think?

martien-de-jong (approved):

Issue #482, opened 2025-06-09
Lowering of llvm trunc instruction


newling:
The following snippet of LLVM IR:

```llvm
; I think this is: load 512 bits, interpret as 16xi32, truncate to 16xi16, store 256 bits.
define void @func_0(){
  %1 = load <8 x i64>, ptr @buff_src, align 256
  %2 = tail call <16 x i16> @llvm.aie2.I256.v16.acc32.srs(<8 x i64> %1, i32 0, i32 0)
  store <16 x i16> %2, ptr @buff_dst, align 256
  ret void
}
```

Makes use of the SRS (shift-round-saturate) instruction for AIE2. with what I think is

shift = 0 (first i32 operand)
round = not relevant, as we're not doing floating point math
saturate = do not saturate (do we need to set another register for this?)

The above compiles fine with 

```
llc test.ll  -O2 --march=aie2 --filetype=asm  -o test.s
```

As far as I can see, the following generic LLVM IR in exactly equivalent. Would it possible to perform instruction selection to make it compile to the same thing? 

```llvm
define void @func_1() { 
  %1 = load <16 x i32>, ptr @buff_src, align 256
  %2 = trunc <16 x i32> %1 to <16 x i16>
  store <16 x i16> %2, ptr @buff_dst, align 256
  ret void
}
```

Currently code like the above fails to compile, errors with 

```
virtual const llvm::RegisterBankInfo::InstructionMapping&
llvm::AIE2RegisterBankInfo::getInstrMapping(const llvm::MachineInstr&)
const: Assertion `(DstTy == S16 || DstTy == S20 || DstTy == S32) && 
(SrcTy == S32 || SrcTy == S64) && "Expected to see 
only 32-or-64 to 16-or-20-or-32 bit truncations"' failed.
```

Note: there is some (oldish) public documentation for SRS the instruction here https://www.xilinx.com/htmldocs/xilinx2023_2/aiengine_ml_intrinsics/intrinsics/group__intr__gpvectorconv__srs.html -- if you're reading this and have a more up-to-date version please let me know and I'll update this link.

konstantinschwarz:
We implemented generic `trunc` support in AIE2p, but took a slightly different approach: we aren't using SRS, but use VSHUFFLE instead to select the low bits from each lane: https://github.com/Xilinx/llvm-aie/blob/aie-public/llvm/lib/Target/AIE/aie2p/AIE2PInstrPatterns.td#L1256

[...]

PR #492, opened 2025-06-17
[AIE][AIE2] Use pattern to convert (special case) shufflevector to subvector extract

newling:
commit 1: whitespace NFC
commit 2: refactor NFC
commit 3: add one of the 'combiners' to AIE2 list

#### Testing: 
Currently all logic is tested for AIE2P in `GlobalISel/prelegalizercombiner-shuffle-vector.mir`, I could add a new file specific for AIE2 if requested

#### Question:
Should I just add all the 'combiners' to both? Is there any ISA change from AIE2 to AIE2P which means the other shuffle 'combiners' can only be used for AIE2P? 

#### Motivation:
I'm experimenting with lowerings from generic arithmetic ops to intrinsics. Consider

```LLVM
@buff_src = external global [128 x i32]
@buff_dst = external global [128 x i32]

define void @alpha() { 
  %1 = load <16 x i32>, ptr @buff_src, align 256
  %2 = trunc <16 x i32> %1 to <16 x i16>
  store <16 x i16> %2, ptr @buff_dst, align 256
  ret void
}

define void @beta() {
  %1 = load <16 x i32>, ptr @buff_src, align 256
  %2 = tail call <16 x i32> @llvm.aie2.vshuffle(<16 x i32> %1, <16 x i32> %1, i32 2)
  %3 = shufflevector <16 x i32> %2, <16 x i32> poison, <8 x i32> <i32 0, i32 1, i32 2, i32 3, i32 4, i32 5, i32 6, i32 7>
  store <8 x i32> %3, ptr @buff_dst, align 256
  ret void
}
```

Currently compiling for aie2P (`llc -O2 -mtriple=aie2P test.ll -o -`)  `@alpha` and `@beta` lower to exactly the same asm (because of [this](https://github.com/Xilinx/llvm-aie/blob/aie-public/llvm/lib/Target/AIE/aie2p/AIE2PInstrPatterns.td#L1256) for trunc from @niwinanto and [this](https://github.com/Xilinx/llvm-aie/blob/f0bbcd15fe35370746aa0adfa7094b2511b65034/llvm/lib/Target/AIE/AIECombine.td#L125) for shufflevector from @katerynamuts). 

But for aie2 ... 

`@alpha` crashes (as already diagnosed in https://github.com/Xilinx/llvm-aie/issues/482) and for `@beta` 

before this PR) `shufflevector` is scalarized 
after this PR) `shufflevector` is handled correctly (`vst wl0` appears directly).

### DIFF
--- a/llvm/lib/Target/AIE/AIECombine.td
+++ b/llvm/lib/Target/AIE/AIECombine.td
@@ -9,7 +9,7 @@
 //===----------------------------------------------------------------------===//
 include "llvm/Target/GlobalISel/Combine.td"
 
-// Explicitly listing each generic combine here ensures direct visibility and 
+// Explicitly listing each generic combine here ensures direct visibility and
 // control over all functionalities.
 def aie_all_combines : GICombineGroup<[trivial_combines, vector_ops_combines,
   insert_vec_elt_combines, extract_vec_elt_combines, combines_for_extload,
@@ -175,7 +175,7 @@ def AIE2PreLegalizerCombiner
                                                                  aie_all_combines, combine_S20NarrowingOpt,
                                                                  combine_globalval_offset,
                                                                  combine_extract_vector_elt_and_zsa_ext,
-                                                                 combine_splat_vector, combine_concat_to_pad_vector, 
+                                                                 combine_splat_vector, combine_concat_to_pad_vector,
                                                                  combine_single_diff_build_vector,
                                                                  combine_symmetric_build_vector,
                                                                  combine_broadcast_shl_to_add]> {
@@ -189,10 +189,10 @@ def AIE2PPreLegalizerCombiner
                                                                  aie_all_combines, combine_S20NarrowingOpt,
                                                                  combine_globalval_offset,
                                                                  combine_extract_vector_elt_and_zsa_ext,
-                                                                 combine_splat_vector, 
+                                                                 combine_splat_vector,
                                                                  combine_vector_shuffle_bcst_to_copy,
                                                                  combine_vector_broadcast,
-                                                                 combine_concat_to_pad_vector, 
+                                                                 combine_concat_to_pad_vector,
                                                                  combine_vector_shuffle_vsel,
                                                                  combine_vector_shuffle_broadcast,
                                                                  combine_single_diff_build_vector,
@@ -294,14 +294,14 @@ def AIE2PostLegalizerCustomCombiner
                                                           combine_add_vector_elt_undef,
                                                           combine_extract_concat,
                                                           combine_unmerge_concat,
-                                                          combine_upd_to_concat 
+                                                          combine_upd_to_concat
                                                           ]> {
 }
 
 def AIE2PPostLegalizerCustomCombiner
     : GICombiner<"AIE2PPostLegalizerCustomCombinerImpl", [  combine_global_load_store_increment,
                                                             combine_load_store_increment,
-                                                            ptr_add_immed_chain, 
+                                                            ptr_add_immed_chain,
                                                             combine_offset_load_store_ptradd,
                                                             combine_offset_load_store_share_ptradd,
                                                             combine_add_vector_elt_undef ]> {
--- a/llvm/lib/Target/AIE/AIECombine.td
+++ b/llvm/lib/Target/AIE/AIECombine.td
@@ -4,7 +4,7 @@
 // See https://llvm.org/LICENSE.txt for license information.
 // SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
 //
-// (c) Copyright 2023-2024 Advanced Micro Devices, Inc. or its affiliates
+// (c) Copyright 2023-2025 Advanced Micro Devices, Inc. or its affiliates
 //
 //===----------------------------------------------------------------------===//
 include "llvm/Target/GlobalISel/Combine.td"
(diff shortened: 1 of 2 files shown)


martien-de-jong:
Hi @newling, There are a few differences in ISA that sometimes require different handling between aie2 and aie2p, We try to hide that in TargetInstrInfo (TII) virtual methods, so that we can reuse them on all architectures.

I estimate that the handling of these generic shuffles is mainly target independent. They are recent, and they usually get tested in the most recent target. Between aie2 and aie2p, the biggest difference is in the vector load/store bandwidth, 256 for aie2 and 512 for aie2p. The shuffles however work on the 512 bit xn vectors on both architectures. 

So, if the shuffle combines don't work on aie2, they should be fixed -- thanks a bundle for testing. 

As for aie2-specific tests, there are some considerations. Really generic combines just need a code generator environment, and it doesn't matter which specific arch. They usually just run an MIR pass. Some of these combines use target-dependent opcodes, which should be retrieved from TII. I think that a wrong opcode would have a high probability of being caught by the machine-verifier, so running the same test with a different target would suffice. One could argue that these tests should be in a 'common' directory, where they could just have two RUN lines for different architectures, but I'm not going to put down pedantic requirements.

Just a final remark, peano was developed very much to serve the defined API, which enters codegen in the form of a set of intrinsics. That has left some holes in the generic vector operation handling, and you will probably hit a few problems. Your effort is much appreciated, but you may want to reconsider if completing all of it is on the critical path of a project.

newling:
Hi @martien-de-jong, thank for these insights and useful pointers!

> There are a few differences in ISA that sometimes require different handling between aie2 and aie2p, We try to hide that in TargetInstrInfo (TII) virtual methods, so that we can reuse them on all architectures.

Got it, I see that for the 'basic vector size' for [AIE2](https://github.com/Xilinx/llvm-aie/blob/adcfd3b8b9f63fbb00bad8fa48ab9cb12e88d7dd/llvm/lib/Target/AIE/AIE2InstrInfo.cpp#L1648) vs [AIE2P](https://github.com/Xilinx/llvm-aie/blob/adcfd3b8b9f63fbb00bad8fa48ab9cb12e88d7dd/llvm/lib/Target/AIE/aie2p/AIE2PInstrInfo.cpp#L1819). Perhaps not the best example here as they're the same (512).  Is the load/store bandwidth difference encoded in one of the virtual methods of [AIEBaseInstrInfo](https://github.com/newling/llvm-aie/blob/a32614d262c21f4a13a6fb7607057d332885f514/llvm/lib/Target/AIE/AIEBaseInstrInfo.h#L37)?



> Just a final remark, peano was developed very much to serve the defined API, which enters codegen in the form of a set of intrinsics. That has left some holes in the generic vector operation handling, and you will probably hit a few problems. Your effort is much appreciated, but you may want to reconsider if completing all of it is on the critical path of a project.

My goal is to support the workloads we are running on the MLIR->LLVM path as simply as possible, with a bias towards having support directly in peano, so that other peano users can benefit. There are cases where it makes more sense for us to get onto the intrinsic path while still in MLIR, but other cases where it's easier for peano to handle generic LLVM directly. 

As for testing in this PR: I intend to add a test for the change introduced here. But I would like to keep the functional change as small in this PR as possible, i.e. I only want to add the one combine here (i.e. [this](https://github.com/Xilinx/llvm-aie/pull/492/commits/f0bbcd15fe35370746aa0adfa7094b2511b65034) commit). Happy to add more combines in this PR if you prefer, though.

martien-de-jong:
Hi @newling, the vector size is an add-hoc abstraction that replaces the many occurrences of the literal 512. mainly in these shuffle combines. It is probably not the best interface, e.g. future archs may have shuffle instruction support for more than one vector size. The load bandwidth is currently handled in instruction selection, which still is largely fully target-dependent with little reuse.

We hope to factor out more common code; the status quo is a balance of refactoring effort and quick bring-up of aie2p.

Handling more combines in one PR isn't a problem for us, but you will suffer more delay in getting it approved. Since we would like separate commits with updated tests for each of the combines anyway, putting those in separate PRs  will increase the likelihood of getting some of them merged earlier.


Many thanks for your efforts.

newling:
Hi @martien-de-jong, thanks for sharing this information about general design goals and directions, it's very useful.  

I think this PR is ready for review now, please let me know what changes you'd like made. 

Regarding testing: In this PR there is 1 'combine' added for AIE2 (reused from AIE2P). It has 2 new tests, and 1 modified test. There is some good refactoring that can be done to combine the files for testing shufflevector for [AIE2](https://github.com/Xilinx/llvm-aie/blob/aie-public/llvm/test/CodeGen/AIE/aie2/intrinsics-shufflevec.ll) and [AIE2P](https://github.com/Xilinx/llvm-aie/blob/aie-public/llvm/test/CodeGen/AIE/aie2p/shufflevec.ll) but I'd prefer not to tackle that in this PR. 

I have split the PR into 3 commits. Which by the way is not an approach I am used to. Most projects I have contributed to don't have stacked commits in a PR that are each considered individually. It seems like a tricky juggling act unless you know exactly what the final PR will look like when you start, and don't make mistakes along the way! What is the approach if you find a minor error in a commit in the middle of the stack, do you need to rewrite the history of the branch by fixing that middle commit?

martien-de-jong (changes requested):
I'll approve whenyou fix the copyright dates of the changed files to 2025

martien-de-jong (approved):

PR #494, opened 2025-06-17
Reg Sequence G_CONCAT

F-Stuckmann:
Follow G_CONCAT to determine if a Vector Register or Accumulator is needed. 

Waiting on QoR.


F-Stuckmann:
image
image

llvm/lib/Target/AIE/aie2p/AIE2PRegisterBankInfo.cpp
```diff
@@ -914,9 +914,9 @@ bool AIE2PRegisterBankInfo::registerBankLookAheadSearch(
   for (auto &UseMI : MRI.use_nodbg_instructions(Reg)) {
     const unsigned UseOpcode = UseMI.getOpcode();
 
-    // skip copies, bitcasts and phis
+    // look through copies, bitcasts, phis and concat_vectors
```
andcarminati: nit: maybe check all G opcodes first.

llvm/lib/Target/AIE/aie2p/AIE2PRegisterBankInfo.cpp
```diff
@@ -914,9 +914,9 @@ bool AIE2PRegisterBankInfo::registerBankLookAheadSearch(
...
     if (UseOpcode == TargetOpcode::G_BITCAST || IsCopyToVReg(UseMI) ||
-        UseMI.isPHI()) {
+        UseMI.isPHI() || UseOpcode == TargetOpcode::G_CONCAT_VECTORS) {
```
martien-de-jong: When we follow PHI nodes, we may run in cycles, only broken by the depth limit of the search. Instead, we may want to establish completeness within the search depth and get an optimistic answer.

martien-de-jong: Can't we put this somewhere in a base class, or as a free utility somewhere? AFAICS, it is fully target independent.

martien-de-jong: Could I ask you to document this function?

llvm/lib/Target/AIE/AIEBaseRegisterBankInfo.cpp
```diff
@@ -490,3 +490,51 @@ AIEBaseRegisterBankInfo::getInstrMappingFinal(
...
+    const MachineRegisterInfo &MRI, const TargetRegisterInfo &TRI, Register Reg,
+    std::unordered_set<MachineInstr *> &VisitedMIs) const {
+
+  auto IsCopyToVReg = [](const MachineInstr &MI) {
+    return (MI.isCopy() && MI.getOperand(0).getReg().isVirtual());
+  };
+
+  // Check all Uses of Reg for Register Bank Preference
+  for (auto &UseMI : MRI.use_nodbg_instructions(Reg)) {
+    if (VisitedMIs.find(&UseMI) != VisitedMIs.end()) {
```
andcarminati: can we use just `contains` here?

andcarminati:
Should we remove `MaxDepthBankSearch` now?

llvm/lib/Target/AIE/AIEBaseRegisterBankInfo.cpp
F-Stuckmann: for readability we can use `is_contained` wrapper

andcarminati (approved):
LGTM.

PR #496, opened 2025-06-20
[aie][aie2p] Global Isel : trunc vector when size(src)/size(dst) = 4

newling:
Add instruction selection tablegen pattern for

```llvm
 %1 = trunc <32 x i32> %0 to <32 x i8>
 ```
 
It is based on existing patterns for 

```llvm
 %1 = trunc <32 x i32> %0 to <32 x i16>
 ```
 
 and 
 
 ```llvm
 %1 = trunc <32 x i16> %0 to <32 x i8>
 ```
 

The final asm with this new lowering for ` trunc <32 x i32> %0 to <32 x i8>` is 

```asm 
     vlda     x0, [p0, #0];          vldb     x1, [p0, #64]
     [...]
     ret     lr;             vshuffle        x2, x0, x1, r0
     mova    r1, #0;         vshuffle        x0, x0, x1, r0  //  Delay Slot 5
     vshuffle        x0, x0, x2, r1                  //  Delay Slot 4
     [...]
```
 

#### Questions for reviewers:

- I couldn't find a way in tablegen to do this with just 2 shuffles, and I'm not sure if it's possible with tablegen. The first 2 `vshuffles` compute the same thing (question: is CSE after instruction selection possible?). Maybe if I wrote the pattern in C++ I could do it with 2 shuffles and avoid the 'rematerialization', but will the final code be more efficient? 
 
-  I see in the tests before this PR, tests of `accregbank` and `vecregbank`. Do I need to have the same in my tests? 

- Do I need end-to-end tests? Full llc run, some kind of numerical test?


llvm/lib/Target/AIE/aie2p/AIE2PInstrPatterns.td
```diff
@@ -1257,26 +1258,52 @@ defm : Extract_512<i64, v8i64, (i32 c6u:$idx), VEXTRACT_64_vec_extract_imm_vaddS
 // instruction. Modes 0, 2, and 4 correspond to the deinterleaved operation of
 // 1, 2, and 4 bytes respectively on the concatenated src0 and src1 input
 // vectors of the VSHUFFLE instruction.
-class Trunc1024Pat<ValueType DstTy, ValueType SrcTy, Instruction ShuffleInstOpc, int Mode> :
+
+// | 1024-bits -> 512-bits | 512-bits -> 256-bits  | 1024-bits -> 256-bits |
+// +=======================+=======================+=======================|
+// |   v16i64 -> v16i32    |   v8i32  -> v8i16     |   v16i64 -> v16i16    |
+// |   v32i32 -> v32i16    |   v16i16 -> v16i8     |   v32i32 -> v32i8     |
+// |   v64i16 -> v64i8     |   v32i8  -> v32i4     |                       |
```
konstantinschwarz: The 512-bits -> 256-bits column looks off? This is `256-bits -> 128-bit`..

llvm/lib/Target/AIE/aie2p/AIE2PLegalizerInfo.cpp
```diff
@@ -281,13 +281,24 @@ AIE2PLegalizerInfo::AIE2PLegalizerInfo(const AIE2PSubtarget &ST)
...
+        const auto SrcElmBits = SrcTy.getElementType().getSizeInBits();
+        if (SrcElmBits != 64 && SrcElmBits != 32 && SrcElmBits != 16)
+          return false;
+
+        const TypeSize SrcBits = SrcTy.getSizeInBits();
+        const TypeSize DstBits = DstTy.getSizeInBits();
+
+        return ((SrcBits == 1024 && DstBits == 256) ||
+                (SrcBits == 1024 && DstBits == 512) ||
+                (SrcBits == 512 && DstBits == 256));
```
konstantinschwarz: A legalizer test would be nice too, e.g. in `llvm/test/CodeGen/AIE/aie2p/GlobalIsel/legalize-trunc.mir`

konstantinschwarz (commented):
The approach looks fine to me.

Ideally we would be able to do this "legalization" in the Legalizer, then we don't have to include this complexity into every AIE instruction selector.
However multiple G_TRUNC steps are combined in the artifact combiner during legalization, creating an infinite loop in the legalizer :/

llvm/lib/Target/AIE/aie2p/AIE2PInstrPatterns.td
```diff
@@ -15,6 +15,7 @@
 
 include "AIEBaseInstrPatterns.td"
 
+
```
konstantinschwarz: super nit: unrelated whitespace change?

niwinanto:
> The approach looks fine to me.
> 
> Ideally we would be able to do this "legalization" in the Legalizer, then we don't have to include this complexity into every AIE instruction selector. However multiple G_TRUNC steps are combined in the artifact combiner during legalization, creating an infinite loop in the legalizer :/

I agree with @konstantinschwarz Moving this complex ISel pattern to legalizer makes much more sense. However, did you try vshuffle with mode 36, this might help you to trunc v*i32 to v*i8 directly.

llvm/lib/Target/AIE/aie2p/AIE2PLegalizerInfo.cpp
```diff
@@ -281,13 +281,24 @@ AIE2PLegalizerInfo::AIE2PLegalizerInfo(const AIE2PSubtarget &ST)
...
-                SrcTy.getSizeInBits() == 1024) &&
-               DstTy.getElementType().getSizeInBits() * 2 ==
-                   SrcTy.getElementType().getSizeInBits();
+
+        if (!SrcTy.isVector() || !DstTy.isVector())
+          return false;
+
+        const auto SrcElmBits = SrcTy.getElementType().getSizeInBits();
+        if (SrcElmBits != 64 && SrcElmBits != 32 && SrcElmBits != 16)
+          return false;
```
martien-de-jong: Perhaps assert, or add the assumptions on DstElemBits as a comment. I assume we know it is <= SrcElemBits because we have a G_TRUNC and all truncations are legal.

newling:
My initial approach was via legalizing (i32->i8) --> (i32->i16->i8) but as predicted @konstantinschwarz this does get undone by a combiner. Maybe it's possible to adjust the list of combines used [here](https://github.com/Xilinx/llvm-aie/pull/400/commits/889a675caafe112e2c8e87b384bafc5a5d32cea7) but that might prevent a good optimization. 

@niwinanto your mode 36 suggestion looks very promising, I'm investigating now!

newling:
@niwinanto the problem I am facing using shuffle's mode 36 truncation is that it can do  `v16i32 -> v16i8` but the result there (v16i8) is 128-bits and I can't find an equivalent of `EXTRACT_SUBREG [...] sub_256_lo` for 128-bits. Is that because the underlying HW registers are 256-bits?

niwinanto:
> @niwinanto the problem I am facing using shuffle's mode 36 truncation is that it can do `v16i32 -> v16i8` but the result there (v16i8) is 128-bits and I can't find an equivalent of `EXTRACT_SUBREG [...] sub_256_lo` for 128-bits. Is that because the underlying HW registers are 256-bits?

@newling Yes, you are right. Basic HW register for vector registers are 256-bit wide, so we cannot extract a subregister smaller than than 256-bit. However, we have 128-bit q registers and there is a vector move operation which moves from 256 bit register into 128 bit register with truncation of upper 128 bits. Please take a look at `VMOV_alu_mv_mv_w_to_q`, this instruction can help here.

llvm/lib/Target/AIE/aie2p/AIE2PInstrPatterns.td
newling: Yeah, I've removed this table now (using new approach suggested [here](https://github.com/Xilinx/llvm-aie/pull/496#issuecomment-2999218824))

llvm/test/CodeGen/AIE/aie2p/GlobalIsel/legalize-trunc.mir
```diff
@@ -133,3 +133,34 @@ body:             |
...
+    ; CHECK-NEXT: [[UV:%[0-9]+]]:_(<8 x s64>), [[UV1:%[0-9]+]]:_(<8 x s64>) = G_UNMERGE_VALUES [[DEF]](<16 x s64>)
+    ; CHECK-NEXT: [[TRUNC:%[0-9]+]]:_(<8 x s16>) = G_TRUNC [[UV]](<8 x s64>)
+    ; CHECK-NEXT: [[TRUNC1:%[0-9]+]]:_(<8 x s16>) = G_TRUNC [[UV1]](<8 x s64>)
+    ; CHECK-NEXT: [[CONCAT_VECTORS:%[0-9]+]]:_(<16 x s16>) = G_CONCAT_VECTORS [[TRUNC]](<8 x s16>), [[TRUNC1]](<8 x s16>)
+    ; CHECK-NEXT: PseudoRET implicit $lr, implicit [[CONCAT_VECTORS]](<16 x s16>)
+    %1:_(<16 x s64>) = G_IMPLICIT_DEF
+    %0:_(<16 x s16>) = G_TRUNC %1(<16 x s64>)
+    PseudoRET implicit $lr, implicit %0
+...
+
```
niwinanto: Can we include new legal cases as well `v16s8_trunc_v16s32_vec512` and `v8s16_trunc_v8s64_vec512`

niwinanto:
@newling change looks good to me. Just one feed back to include more tests in the legalizer. Also, would be nice if you can rephrase the commit message and squash the clang format commit to respective commit before the merge.

newling:
> Also, would be nice if you can rephrase the commit message and squash the clang format commit to respective commit before the merge.

Will do. I'm still unsure what the recommended workflow is here, I've only worked on projects where all commits get squashed into a single commit when the PR is merged into main, as outlined [here](https://llvm.org/docs/GitHub.html)/ I guess in this project you prefer large 'feature' PRs where each commit is a single step towards the feature. As opposed to a sequence of small PRs? Anyway, I'm happy to adapt whatever you all use as a workflow :)

niwinanto (dismissed):
Thanks for addressing the feedback. LGTM!

llvm/lib/Target/AIE/aie2p/AIE2PLegalizerInfo.cpp
newling: I've updated this logic with additional comments and asserts

andcarminati:
> > Also, would be nice if you can rephrase the commit message and squash the clang format commit to respective commit before the merge.
> 
> Will do. I'm still unsure what the recommended workflow is here, I've only worked on projects where all commits get squashed into a single commit when the PR is merged into main, as outlined [here](https://llvm.org/docs/GitHub.html)/ I guess in this project you prefer large 'feature' PRs where each commit is a single step towards the feature. As opposed to a sequence of small PRs? Anyway, I'm happy to adapt whatever you all use as a workflow :)

I think this PR is in a nice shape now, only a small suggestion: append [AIE2p] in front of each commit message.

newling:
Ping

konstantinschwarz (approved):
LGTM. It would be nice to add LLVM IR -> assembly tests for all types that we expect to work, similar to https://github.com/Xilinx/llvm-aie/blob/aie-public/llvm/test/CodeGen/AIE/extractelement.ll

newling:
TODO(newling) don't land before I've verified numerical correctness of this transform.

PR #497, opened 2025-06-23
NFC: move some aie1 stuff out of the way

martien-de-jong:
Some decluttering


F-Stuckmann (dismissed):
Good start on the refactoring of AIE1.

At some point we may also want to move AIE2 stuff into the AIEBase, since they are shared in AIE2 and AIE2P.

martien-de-jong:
> Good start on the refactoring of AIE1.
> 
> At some point we may also want to move AIE2 stuff into the AIEBase, since they are shared in AIE2 and AIE2P.

Yes indeed. These commits are preparatory in order not to disrupt ongoing developments, banking on the fact that aie1 should be quite stable. It may well be that some generic code still uses AIE1 specific include files.

llvm/lib/Target/AIE/AIEInstructionSelector.cpp
```diff
@@ -13,11 +13,11 @@
 /// \todo This should be generated by TableGen.
 //===----------------------------------------------------------------------===//
 
-#include "AIE1RegisterBankInfo.h"
-#include "AIERegisterInfo.h"
 #include "AIETargetMachine.h"
 #include "InstPrinter/AIEInstPrinter.h"
 #include "MCTargetDesc/AIEMCTargetDesc.h"
+#include "aie1/AIE1RegisterBankInfo.h"
```
andcarminati, line 19: This is curious, sometimes AIE means AIE1. I guess this is the AIE1 instruction selector, we could also rename and move. If this is not AIE1, it is also curious to have target specific includes (but no AIE2/P includes.)

llvm/lib/Target/AIE/AIESubtarget.cpp
```diff
@@ -14,11 +14,11 @@
 
 #include "AIESubtarget.h"
 #include "AIE.h"
-#include "AIE1LegalizerInfo.h"
-#include "AIE1RegisterBankInfo.h"
 #include "AIECallLowering.h"
 #include "AIEFrameLowering.h"
 #include "AIETargetMachine.h"
+#include "aie1/AIE1LegalizerInfo.h"
```
andcarminati, line 20: I think we can also move this as AIESubtarget.cpp -> aie1/AIE1Subtarget.cpp

andcarminati:
Nice cleaning work. I guess we have some AIE files that are true AIE1-specific files, that can be also moved/renamed.

llvm/lib/Target/AIE/AIEInstructionSelector.cpp
martien-de-jong: Yes, there's more to come. I just want to have a quick turn-around time, so I do  little chunks at the time.

llvm/lib/Target/AIE/AIESubtarget.cpp
martien-de-jong: Yeah, subtarget.h will get a big commit in a follow-up PR.

F-Stuckmann (approved):

PR #29, opened 2024-05-06
Loop aware

martien-de-jong:
Reopened with a non-fork branch.

PR #500, opened 2025-06-23
[AIEX] Concat/Unmerge PHI Combiner

F-Stuckmann:
Added a Concat/Unmerge PHI Combiner. 

QoR does not show any results, because the kernels, that use this pattern are currently failing. 

However, in ReduceL2C8Axis_1_aie2_bf16 the II can be reduced from 9 to 8.


llvm/lib/Target/AIE/AIECombine.td
```diff
@@ -304,5 +313,7 @@ def AIE2PPostLegalizerCustomCombiner
                                                             ptr_add_immed_chain, 
                                                             combine_offset_load_store_ptradd,
                                                             combine_offset_load_store_share_ptradd,
-                                                            combine_add_vector_elt_undef ]> {
+                                                            combine_add_vector_elt_undef,
+                                                            combine_concat_unmerge_phis
+                                                             ]> {
```
andcarminati, line 334: Why are we using post-legalizer and not pre-legalizer? We are touching types here, so we can generate not legal types as well.

llvm/lib/Target/AIE/AIECombinerHelper.cpp
```diff
@@ -563,6 +563,183 @@ static MachineInstr *findPostIncMatch(MachineInstr &MemI,
...
+/// 3 = G_CONCAT 1, 2
+/// ....
+/// 6 = ...
+/// 7, 8 = G_UNMERGE 6
+/// 9,10 = G_UNMERGE 6
+/// \return unmerge source (6) if the pattern is valid and the sub-vector index
+/// used [ 0 (7, 9) or 1 (8, 10)]. Start the pattern checking from \p ConcatI .
+/// Follow \p UseIdx (0 or 1) of G_CONCAT to identify the pattern. Collect all
+/// relevant Results in \p MatchData .
+static std::pair<MachineInstr *, unsigned>
```
andcarminati: Can we embed this return info into AIEConcatUnmergeCombineMatchData? In this way we can concentrate all information together.

llvm/lib/Target/AIE/AIECombine.td
F-Stuckmann: the only instruction we generate is a concat in a dominating MBB, that we know is already legal.

llvm/lib/Target/AIE/AIECombinerHelper.h
```diff
@@ -30,6 +30,23 @@ struct FrequentIndexResult {
...
 };
 
+struct AIEConcatUnmergeCombineMatchData {
+  // New Concat Instruction
+  // MBB where the G_CONCAT of ConcatSub0 and ConcatSub1
+  // should be created.
+  MachineBasicBlock *ConcatMBB = nullptr;
+  // Phi Src Operand 0: Concat of ConcatSub0 and ConcatSub1
+  MachineOperand *ConcatSub0 = nullptr;
+  // Phi Src Operand 0: Concat of ConcatSub0 and ConcatSub1
```
andcarminati: Operand 1?

llvm/lib/Target/AIE/AIECombinerHelper.cpp
```diff
@@ -563,6 +563,183 @@ static MachineInstr *findPostIncMatch(MachineInstr &MemI,
...
+/// bb.0
+/// concatIn0 = ...
+/// concatIn1 = ...
+/// bb.1
+/// 1 = phi 7, bb.1, concatIn0, bb.0
+/// 2 = phi 10, bb.1 concatIn1, bb.0
+/// 3 = G_CONCAT 1, 2
+/// ....
+/// 6 = ...
+/// 7, 8 = G_UNMERGE 6
```
andcarminati, line 570: Why do we have two unmerges here? Is it common on mllib?

llvm/test/CodeGen/AIE/GlobalISel/simplify-concat-unmerge-phi.mir
```diff
@@ -0,0 +1,652 @@
...
+# SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+#
+# (c) Copyright 2025 Advanced Micro Devices, Inc. or its affiliates
+
+
+# RUN: llc -mtriple aie2 -run-pass=aie2-postlegalizer-custom-combiner %s -verify-machineinstrs -o - | FileCheck --check-prefix=COMMON %s
+# RUN: llc -mtriple aie2p -run-pass=aie2p-postlegalizer-custom-combiner %s -verify-machineinstrs -o - | FileCheck --check-prefix=COMMON %s
+
+
+# Incoming Concat Values from the same MBB (bb.0), backedge is the same MBB as the PHI node
```
andcarminati, line 15: Maybe just label as Positive and Negative test to be easier to follow.

llvm/lib/Target/AIE/AIECombinerHelper.cpp
```diff
@@ -563,6 +563,183 @@ static MachineInstr *findPostIncMatch(MachineInstr &MemI,
...
+    return std::pair{UnMergeI, UnMergeDefMO};
+  };
+
+  // Find PHI node
+  const auto OperandIdx = UseIdx + 1;
+  assert(ConcatI.getOperand(OperandIdx).isReg());
+  const Register Reg = ConcatI.getOperand(OperandIdx).getReg();
+  auto *PhiInst = MRI.getVRegDef(Reg);
+  if (!PhiInst || !PhiInst->isPHI())
+    return {nullptr, /*UnmergeSubVecDefIdx=*/0};
```
andcarminati: Could we use optional here?

llvm/lib/Target/AIE/AIECombinerHelper.cpp
```diff
@@ -563,6 +563,183 @@ static MachineInstr *findPostIncMatch(MachineInstr &MemI,
...
+        return UnMergeDefIdx;
+      }
+      UnMergeDefIdx++;
+    }
+
+    llvm_unreachable("Lost PHIs MO in UnMerge");
+    return (unsigned)-1;
+  };
+
+  const auto SubVecDefIdx =
```
andcarminati: nit: `const unsigned SubVecDefIdx = UnMergeDefMO->getOperandNo();` (you can drop the previous lambda)

llvm/lib/Target/AIE/AIECombinerHelper.cpp
```diff
@@ -563,6 +563,183 @@ static MachineInstr *findPostIncMatch(MachineInstr &MemI,
...
+  assert(ConcatI.getOperand(OperandIdx).isReg());
+  const Register Reg = ConcatI.getOperand(OperandIdx).getReg();
+  auto *PhiInst = MRI.getVRegDef(Reg);
+  if (!PhiInst || !PhiInst->isPHI())
+    return {nullptr, /*UnmergeSubVecDefIdx=*/0};
+
+  auto [UnMergeInst, UnMergeDefMO] = FindPhiUnMergeComponents(PhiInst);
+
+  if (!UnMergeInst ||
+      UnMergeInst->getOpcode() != TargetOpcode::G_UNMERGE_VALUES ||
```
andcarminati: Unmerge can have more then 2 defs, we should check this too.

llvm/lib/Target/AIE/AIECombinerHelper.cpp
```diff
@@ -563,6 +563,183 @@ static MachineInstr *findPostIncMatch(MachineInstr &MemI,
...
+
+bool llvm::matchConcatUnmergePhis(MachineInstr &ConcatI,
+                                  MachineRegisterInfo &MRI,
+                                  CombinerHelper &Helper,
+                                  AIEConcatUnmergeCombineMatchData &MatchInfo) {
+  LLVM_DEBUG(dbgs() << "MF: " << ConcatI.getParent()->getParent()->getName()
+                    << "\n");
+
+  auto UnMergeOrigin0 =
+      findUnmergeOrigin(ConcatI, /*UseIdx=*/0, MRI, Helper, MatchInfo);
```
andcarminati: I was thinking if we can find indexes 0 and 1 in a single sweep, storing relevant data in MatchInfo.

llvm/lib/Target/AIE/AIECombinerHelper.cpp
```diff
@@ -563,6 +563,183 @@ static MachineInstr *findPostIncMatch(MachineInstr &MemI,
...
+      GetUnmergeDefSubVectorIndex(UnMergeInst, UnMergeDefMO->getReg());
+  return {UnMergeSrc, /*UnmergeSubVecDefIdx=*/SubVecDefIdx};
+}
+
+bool llvm::matchConcatUnmergePhis(MachineInstr &ConcatI,
+                                  MachineRegisterInfo &MRI,
+                                  CombinerHelper &Helper,
+                                  AIEConcatUnmergeCombineMatchData &MatchInfo) {
+  LLVM_DEBUG(dbgs() << "MF: " << ConcatI.getParent()->getParent()->getName()
+                    << "\n");
```
andcarminati: We can assert that `ConcatI` is indeed a `G_CONCAT_VECTORS`

llvm/lib/Target/AIE/AIECombinerHelper.cpp
```diff
@@ -563,6 +563,183 @@ static MachineInstr *findPostIncMatch(MachineInstr &MemI,
...
+
+      // Check if the basic block operand matches the given MBB
+      if (MBB.getMBB()->empty() ||
+          !Helper.dominates(*FirstMI, *MBB.getMBB()->begin())) {
+        // If MBB is empty, it can only be a fallthrough MBB, therefore, it
+        // dominates ConcatI's MBB
+
+        if (MatchData.ConcatMBB && MatchData.ConcatMBB != MBB.getMBB()) {
+          LLVM_DEBUG(dbgs()
+                     << "Both Parts of the Incoming Concat Values do not "
```
andcarminati: Future concat?

llvm/lib/Target/AIE/AIECombinerHelper.cpp
```diff
@@ -563,6 +563,183 @@ static MachineInstr *findPostIncMatch(MachineInstr &MemI,
...
+    B.setMBB(*MatchInfo.ConcatMBB);
+  else
+    B.setInstr(MatchInfo.ConcatMBB->instr_back());
+
+  // Create new Concat Instruction
+  auto NewConcat = B.buildInstr(TargetOpcode::G_CONCAT_VECTORS);
+  const auto TargetVecType = MRI.getType(MatchInfo.UnmergeSourceOp->getReg());
+  NewConcat.addDef(MRI.createGenericVirtualRegister(TargetVecType));
+
+  NewConcat->addOperand(*MatchInfo.ConcatSub0);
```
andcarminati: Be careful when copying MOs, you can drag undesired flags.

llvm/lib/Target/AIE/AIECombinerHelper.cpp
```diff
@@ -563,6 +563,183 @@ static MachineInstr *findPostIncMatch(MachineInstr &MemI,
...
+                                  AIEConcatUnmergeCombineMatchData &MatchInfo,
+                                  GISelChangeObserver &Observer) {
+  /// Set Insertion Point
+  if (MatchInfo.ConcatMBB->empty())
+    B.setMBB(*MatchInfo.ConcatMBB);
+  else
+    B.setInstr(MatchInfo.ConcatMBB->instr_back());
+
+  // Create new Concat Instruction
+  auto NewConcat = B.buildInstr(TargetOpcode::G_CONCAT_VECTORS);
```
andcarminati: I think you can save several lines here by using `B.buildMergeLikeInstr()`
```
Register ConcatReg = MRI.createGenericVirtualRegister(TargetVecType);
B.buildMergeLikeInstr(ConcatReg, {Sub0, Sub1});
```

llvm/lib/Target/AIE/AIECombine.td
andcarminati: We also generate a phi, whose legality could be not the same as concat. This type of optimization always fits better in pre-legalization.

llvm/lib/Target/AIE/AIECombinerHelper.cpp
```diff
@@ -563,6 +563,183 @@ static MachineInstr *findPostIncMatch(MachineInstr &MemI,
...
+  const auto SubVecDefIdx =
+      GetUnmergeDefSubVectorIndex(UnMergeInst, UnMergeDefMO->getReg());
+  return {UnMergeSrc, /*UnmergeSubVecDefIdx=*/SubVecDefIdx};
+}
+
+bool llvm::matchConcatUnmergePhis(MachineInstr &ConcatI,
+                                  MachineRegisterInfo &MRI,
+                                  CombinerHelper &Helper,
+                                  AIEConcatUnmergeCombineMatchData &MatchInfo) {
+  LLVM_DEBUG(dbgs() << "MF: " << ConcatI.getParent()->getParent()->getName()
```
martien-de-jong: tip: ConcatI->getMF();

llvm/lib/Target/AIE/AIECombinerHelper.cpp
```diff
@@ -563,6 +563,183 @@ static MachineInstr *findPostIncMatch(MachineInstr &MemI,
...
+  if (!UnMergeOrigin0.first)
+    return false;
+  auto UnMergeOrigin1 =
+      findUnmergeOrigin(ConcatI, /*UseIdx=*/1, MRI, Helper, MatchInfo);
+  if (!UnMergeOrigin1.first)
+    return false;
+
+  // If Unmerge Origins diverge, skip
+  if (UnMergeOrigin0.first != UnMergeOrigin1.first)
+    return false;
```
martien-de-jong: CHECK: AFAIK, G_CONCAT can also take more than two operands

llvm/lib/Target/AIE/AIECombinerHelper.cpp
F-Stuckmann: It does happen sometimes

llvm/lib/Target/AIE/AIECombinerHelper.h
```diff
@@ -30,6 +30,23 @@ struct FrequentIndexResult {
   unsigned NonMatchingCount;
 };
 
+struct AIEConcatUnmergeCombineMatchData {
+  // New Concat Instruction
```
martien-de-jong: `Concat instruction's MBB` and rewrap?

llvm/lib/Target/AIE/AIECombinerHelper.cpp
```diff
@@ -563,6 +563,183 @@ static MachineInstr *findPostIncMatch(MachineInstr &MemI,
...
+  const auto TargetVecType = MRI.getType(MatchInfo.UnmergeSourceOp->getReg());
+  NewConcat.addDef(MRI.createGenericVirtualRegister(TargetVecType));
+
+  NewConcat->addOperand(*MatchInfo.ConcatSub0);
+  NewConcat->addOperand(*MatchInfo.ConcatSub1);
+  LLVM_DEBUG(dbgs() << "In bb." << MatchInfo.ConcatMBB->getNumber()
+                    << " created " << *NewConcat.getInstr());
+
+  // set Insertion Point to top of phi-MBB
+  B.setInsertPt(*ConcatI.getParent(), ConcatI.getParent()->begin());
```
martien-de-jong, line 719: nit: funny that we use three different methods to set the insertion point. I guess the earlier ones are specializations of this one?

llvm/lib/Target/AIE/AIECombinerHelper.cpp
```diff
@@ -563,6 +563,183 @@ static MachineInstr *findPostIncMatch(MachineInstr &MemI,
...
+          LLVM_DEBUG(dbgs()
+                     << "Both Parts of the Incoming Concat Values do not "
+                        "originate in the same MBB.");
+          return std::pair<MachineInstr *, MachineOperand *>{nullptr, nullptr};
+        }
+
+        // Concat Components
+        MatchData.ConcatMBB = MBB.getMBB();
+        if (UseIdx == 0)
+          MatchData.ConcatSub0 = IncomingMO;
```
martien-de-jong: Looks as if ConcatSub could be an array of two elements

llvm/lib/Target/AIE/AIECombinerHelper.cpp
```diff
@@ -563,6 +563,183 @@ static MachineInstr *findPostIncMatch(MachineInstr &MemI,
...
+      findUnmergeOrigin(ConcatI, /*UseIdx=*/1, MRI, Helper, MatchInfo);
+  if (!UnMergeOrigin1.first)
+    return false;
+
+  // If Unmerge Origins diverge, skip
+  if (UnMergeOrigin0.first != UnMergeOrigin1.first)
+    return false;
+
+  // If Concat behaves like a vector broadcast, i.e. Concat uses the same
+  // Sub-vector in both source operands, skip
```
martien-de-jong: The (local) essence is that we only have one phi node, so the pattern doesn't match. If we want to leave this pattern to be canonicalized to broadcast and then be further optimized, I would say that we widen the combine to handle all CONCAT patterns and rewrite to broadcast here.
Alternatively, we recognize that rewriting CONCAT to broadcast should have a higher priority and we shouldn't expect occurrences here.

llvm/lib/Target/AIE/AIECombinerHelper.cpp
```diff
@@ -563,6 +563,183 @@ static MachineInstr *findPostIncMatch(MachineInstr &MemI,
...
+/// used [ 0 (7, 9) or 1 (8, 10)]. Start the pattern checking from \p ConcatI .
+/// Follow \p UseIdx (0 or 1) of G_CONCAT to identify the pattern. Collect all
+/// relevant Results in \p MatchData .
+static std::pair<MachineInstr *, unsigned>
+findUnmergeOrigin(MachineInstr &ConcatI, unsigned UseIdx,
+                  MachineRegisterInfo &MRI, CombinerHelper &Helper,
+                  AIEConcatUnmergeCombineMatchData &MatchData) {
+  auto *FirstMI = &*ConcatI.getParent()->begin();
+
+  auto FindPhiUnMergeComponents = [UseIdx, &FirstMI, &Helper, &MRI,
```
martien-de-jong: I think it makes sense to specify the return type explicitly. Note that it also saves typing the dummy return value's type.

llvm/lib/Target/AIE/AIECombinerHelper.cpp
```diff
@@ -563,6 +563,183 @@ static MachineInstr *findPostIncMatch(MachineInstr &MemI,
...
+/// 9,10 = G_UNMERGE 6
+/// \return unmerge source (6) if the pattern is valid and the sub-vector index
+/// used [ 0 (7, 9) or 1 (8, 10)]. Start the pattern checking from \p ConcatI .
+/// Follow \p UseIdx (0 or 1) of G_CONCAT to identify the pattern. Collect all
+/// relevant Results in \p MatchData .
+static std::pair<MachineInstr *, unsigned>
+findUnmergeOrigin(MachineInstr &ConcatI, unsigned UseIdx,
+                  MachineRegisterInfo &MRI, CombinerHelper &Helper,
+                  AIEConcatUnmergeCombineMatchData &MatchData) {
+  auto *FirstMI = &*ConcatI.getParent()->begin();
```
krishnamtibrewala: nit: avoid use of "auto" they are definitely useful for coding, but decreases readability of code significantly.

llvm/lib/Target/AIE/AIECombinerHelper.cpp
F-Stuckmann: Since I do not know the sweeps, I still like to go through all operandIdx individually, but I will use MatchInfo to store the relevant data.

llvm/lib/Target/AIE/AIECombinerHelper.cpp
F-Stuckmann: I am now checking all operands

llvm/lib/Target/AIE/AIECombinerHelper.cpp
F-Stuckmann: Funnily enough internally they are very similar. But they don't really share a lot of code (as would be the expectation)

llvm/lib/Target/AIE/AIECombinerHelper.cpp
```diff
@@ -562,6 +562,177 @@ static MachineInstr *findPostIncMatch(MachineInstr &MemI,
...
+}
+
+bool llvm::matchConcatUnmergePhis(MachineInstr &ConcatI,
+                                  MachineRegisterInfo &MRI,
+                                  CombinerHelper &Helper,
+                                  AIEConcatUnmergeCombineMatchData &MatchInfo) {
+  assert(ConcatI.getOpcode() == TargetOpcode::G_CONCAT_VECTORS);
+  LLVM_DEBUG(dbgs() << "MF: " << ConcatI.getMF()->getName() << "\n");
+
+  unsigned NumUse = ConcatI.getNumOperands() - ConcatI.getNumDefs();
```
andcarminati: nit: const.

llvm/lib/Target/AIE/AIECombinerHelper.cpp
```diff
@@ -562,6 +562,177 @@ static MachineInstr *findPostIncMatch(MachineInstr &MemI,
...
+
+  MatchInfo.ConcatSubVecs.clear();
+  MatchInfo.ConcatSubVecs.resize(NumUse);
+
+  for (unsigned UseOpIdx = 0; UseOpIdx < NumUse; UseOpIdx++) {
+    if (!findUnmergeOrigin(ConcatI, /*UseIdx=*/UseOpIdx, MRI, Helper,
```
andcarminati: Nice! Generic.

llvm/lib/Target/AIE/AIECombinerHelper.cpp
andcarminati: Also, NumUses

llvm/lib/Target/AIE/AIECombinerHelper.cpp
```diff
@@ -562,6 +562,177 @@ static MachineInstr *findPostIncMatch(MachineInstr &MemI,
...
+        return UnMergeDefIdx;
+      }
+      UnMergeDefIdx++;
+    }
+
+    llvm_unreachable("Lost PHIs MO in UnMerge");
+    return (unsigned)-1;
+  };
+
+  const auto UnmergeDefIdx = GetOperandNo(UnMergeInst, UnMergeDefMO->getReg());
```
andcarminati: I am still trying to understand why we can't just use `const unsigned UnmergeDefIdx = UnMergeDefMO->getOperandNo();` here.

andcarminati: Oh, now I got! This MO is related to the Phi node. I think we should clarify some naming here. `MachineOperand` always is tied to an instruction. It should be `PhUsefMO` - or the register itself, like `UnMergeDefReg`.

andcarminati: Instructions can always share registers in def-use chains, but never MOs. This should be *very* consistent here.

llvm/lib/Target/AIE/AIECombinerHelper.cpp
```diff
@@ -562,6 +562,177 @@ static MachineInstr *findPostIncMatch(MachineInstr &MemI,
...
+
+  if (!UnMergeInst ||
+      UnMergeInst->getOpcode() != TargetOpcode::G_UNMERGE_VALUES ||
+      !MatchData.ConcatMBB) {
+    // Could not find Unmerge or no non-dominating-MBB could be determined.
+    return false;
+  }
+
+  MachineInstr *UnMergeSrc =
+      MRI.getVRegDef(UnMergeInst->getOperand(2).getReg());
```
andcarminati: Now we handle all uses of a CONCAT, should we adjust this 2 to a dynamic last operand?

llvm/lib/Target/AIE/AIECombinerHelper.cpp
andcarminati: How about `MachineInstr::findRegisterDefOperandIdx` ?

F-Stuckmann: So One operand is from the phi and the other one from the unmerge. 
Unfortunately, they are not the same operand, so i have to manually do the search.

F-Stuckmann: that is why in `GetOperandNo `I check that the registers are equal, instead of using `MO::isIdenticalTo`

llvm/lib/Target/AIE/AIECombinerHelper.cpp
F-Stuckmann: good catch

llvm/lib/Target/AIE/AIECombinerHelper.cpp
andcarminati: Yes, you should easily get the same result with:
```
-  auto GetOperandNo = [](const MachineInstr *UnMergeInst,
-                         const auto UnmergeMOReg) {
-    unsigned UnMergeDefIdx = 0;
-    for (auto MO : UnMergeInst->operands()) {
-      if (MO.isReg() && MO.getReg() == UnmergeMOReg) {
-        return UnMergeDefIdx;
-      }
-      UnMergeDefIdx++;
-    }
-
-    llvm_unreachable("Lost PHIs MO in UnMerge");
-    return (unsigned)-1;
-  };
-
-  const auto UnmergeDefIdx = GetOperandNo(UnMergeInst, UnMergeDefMO->getReg());
+  const unsigned UnmergeDefIdx = UnMergeInst->findRegisterDefOperandIdx(UnMergeDefMO->getReg());
```

Also, we don't need to use MOs as result unless we need some more information apart of the register.

llvm/lib/Target/AIE/AIECombinerHelper.cpp
```diff
@@ -562,6 +562,169 @@ static MachineInstr *findPostIncMatch(MachineInstr &MemI,
...
+  auto NewConcat = B.buildMergeLikeInstr(ConcatReg, MatchInfo.ConcatSubVecs);
+  LLVM_DEBUG(dbgs() << "In bb." << MatchInfo.ConcatMBB->getNumber()
+                    << " created " << *NewConcat.getInstr());
+
+  // set Insertion Point to top of phi-MBB
+  B.setInsertPt(*ConcatI.getParent(), ConcatI.getParent()->begin());
+
+  // Create new PHI node
+  auto NewPHI = B.buildInstr(TargetOpcode::G_PHI);
+  NewPHI.addDef(ConcatI.getOperand(0).getReg());
```
krishnamtibrewala, line 723: update the globelIsel observer.

llvm/lib/Target/AIE/AIECombinerHelper.cpp
```diff
@@ -562,6 +562,169 @@ static MachineInstr *findPostIncMatch(MachineInstr &MemI,
...
+      MatchData.UnmergeMBB = MBB.getMBB();
+    }
+    return {UnMergeI, UnMergeDefMO};
+  };
+
+  // Find PHI node
+  const auto OperandIdx = UseOpIdx + 1;
+  assert(ConcatI.getOperand(OperandIdx).isReg());
+  const Register Reg = ConcatI.getOperand(OperandIdx).getReg();
+  MachineInstr *PhiInst = MRI.getVRegDef(Reg);
```
andcarminati, line 626: I think we should test if the result of the phi node is single use. If some other instruction depends on the original value (type), I think we can have trouble. We can also detect this case and insert a proper unmerge again for that specific use. If we spot this case, we can extend in the future, I don't think is important now.

llvm/lib/Target/AIE/AIECombinerHelper.cpp
```diff
@@ -562,6 +562,169 @@ static MachineInstr *findPostIncMatch(MachineInstr &MemI,
...
+      // Unmerge Components
+      UnMergeDefMO = IncomingMO;
+      UnMergeI = MRI.getVRegDef(IncomingMO->getReg());
+      MatchData.UnmergeMBB = MBB.getMBB();
+    }
+    return {UnMergeI, UnMergeDefMO};
+  };
+
+  // Find PHI node
+  const auto OperandIdx = UseOpIdx + 1;
```
andcarminati: unsigned is better here!

llvm/lib/Target/AIE/AIECombinerHelper.cpp
```diff
@@ -562,6 +562,169 @@ static MachineInstr *findPostIncMatch(MachineInstr &MemI,
...
+  assert(ConcatI.getOperand(OperandIdx).isReg());
+  const Register Reg = ConcatI.getOperand(OperandIdx).getReg();
+  MachineInstr *PhiInst = MRI.getVRegDef(Reg);
+  if (!PhiInst || !PhiInst->isPHI())
```
andcarminati: You can omit this null check here. If we use an undefined register, someone did a mistake early, and this would trigger some error before we reach this combiner.

llvm/lib/Target/AIE/AIECombinerHelper.cpp
```diff
@@ -562,6 +562,169 @@ static MachineInstr *findPostIncMatch(MachineInstr &MemI,
...
+    return false;
+
+  auto [UnMergeInst, UnMergeDefMO] = FindPhiUnMergeComponents(PhiInst);
```
andcarminati: Here, you can return directly the register, it is much more common, it is all you need after.

llvm/lib/Target/AIE/AIECombinerHelper.cpp
```diff
@@ -562,6 +562,169 @@ static MachineInstr *findPostIncMatch(MachineInstr &MemI,
...
+      !MatchData.ConcatMBB) {
+    // Could not find Unmerge or no non-dominating-MBB could be determined.
+    return false;
+  }
+
+  const auto UnmergeUseIdx = UnMergeInst->getNumDefs();
+  const auto UnSourceReg = UnMergeInst->getOperand(UnmergeUseIdx).getReg();
+  // no need to check consistency of MatchData.UnmergeSourceReg across different
+  // UseOpIdx, since UnMergeSrc consistency is checked
+  if (!MatchData.UnmergeSourceReg)
```
andcarminati: Perhaps we can normalize to everything as Unmerge or UnMerge to use a single convention.

llvm/lib/Target/AIE/AIECombinerHelper.h
```diff
@@ -77,6 +94,41 @@ struct AIESingleDiffLaneBuildVectorMatchData {
 
 void foundPattern(MachineInstr &MemI);
 
+/// This is the matching function for Concat-Unmerge-PHI pattern.
```
andcarminati: You can keep this comment only in the .cpp file.

llvm/test/CodeGen/AIE/GlobalISel/simplify-concat-unmerge-phi.mir
```diff
@@ -0,0 +1,809 @@
...
+    %29:_(<16 x s32>), %FirstUnMerge:_(<16 x s32>) = G_UNMERGE_VALUES %27(<32 x s32>)
+    %46:_(s32) = G_INTRINSIC_W_SIDE_EFFECTS intrinsic(@llvm.loop.decrement), %32(s32)
+    %47:_(s32) = G_ASSERT_ZEXT %46, 1
+    G_BRCOND %47(s32), %bb.2
+    G_BR %bb.3
+
+  bb.3:
+    PseudoRET implicit $lr
+
+...
```
andcarminati, line 603: Can we have a negative test, where the phi node has another user apart of the concat?

llvm/lib/Target/AIE/AIECombinerHelper.cpp
```diff
@@ -562,6 +562,169 @@ static MachineInstr *findPostIncMatch(MachineInstr &MemI,
...
+
+  auto FindPhiUnMergeComponents = [UseOpIdx, &FirstMI, &Helper, &MRI,
+                                   &MatchData](MachineInstr *PhiInst)
+      -> std::pair<MachineInstr *, MachineOperand *> {
+    // Find Unmerge and Concat Components of the PHI node
+    MachineInstr *UnMergeI = nullptr;
+    MachineOperand *UnMergeDefMO = nullptr;
+    for (unsigned I = 1; I < PhiInst->getNumOperands(); I += 2) {
+      MachineOperand *IncomingMO = &PhiInst->getOperand(I);
+      MachineOperand &MBB = PhiInst->getOperand(I + 1);
```
andcarminati: nit: `MBBOp`. Even better if we could have directly the `MachineBasicBlock` here.

llvm/lib/Target/AIE/AIECombinerHelper.cpp
```diff
@@ -562,6 +562,169 @@ static MachineInstr *findPostIncMatch(MachineInstr &MemI,
...
+
+      // Check if the basic block operand matches the given MBB
+      if (MBB.getMBB()->empty() ||
```
andcarminati: Can you comment a bit more about this whole check?

llvm/test/CodeGen/AIE/GlobalISel/simplify-concat-unmerge-phi.mir
```diff
@@ -0,0 +1,809 @@
...
+    %9:_(p0) = nuw G_PTR_ADD %4, %8(s20)
+    %10:_(<16 x s32>) = G_LOAD %9(p0) :: (load (<16 x s32>))
+    %11:_(s20) = G_CONSTANT i20 128
+    %12:_(p0) = nuw G_PTR_ADD %4, %11(s20)
+    %35:_(<32 x s32>) = G_CONCAT_VECTORS %6(<16 x s32>), %5(<16 x s32>)
+    %14:_(<16 x s64>) = G_BITCAST %35(<32 x s32>)
+    G_STORE %14(<16 x s64>), %2(p0) :: (store (<16 x s64>))
+    %15:_(<16 x s64>) = G_LOAD %3(p0) :: (load (<16 x s64>))
+    %27:_(<32 x s32>) = G_BITCAST %15(<16 x s64>)
+    %29:_(<16 x s32>), %43:_(<16 x s32>) = G_UNMERGE_VALUES %27(<32 x s32>)
```
andcarminati, line 90: Curious that Combiner DCE did not kick in.

llvm/test/CodeGen/AIE/GlobalISel/simplify-concat-unmerge-phi.mir
```diff
@@ -0,0 +1,809 @@
...
+    G_INTRINSIC_W_SIDE_EFFECTS intrinsic(@llvm.set.loop.iterations), %1(s32)
+
+  bb.2:
+    %4:_(p0) = G_PHI %0(p0), %bb.1, %12(p0), %bb.3
+    %5:_(<16 x s32>) = G_PHI %33(<16 x s32>), %bb.1, %30(<16 x s32>), %bb.3
+    %6:_(<16 x s32>) = G_PHI %33(<16 x s32>), %bb.1, %29(<16 x s32>), %bb.3
+    %7:_(<16 x s32>) = G_LOAD %4(p0) :: (load (<16 x s32>) )
+    %8:_(s20) = G_CONSTANT i20 64
+    %9:_(p0) = nuw G_PTR_ADD %4, %8(s20)
+    %10:_(<16 x s32>) = G_LOAD %9(p0) :: (load (<16 x s32>))
```
andcarminati: Is this a dead load? It is removed after...

llvm/test/CodeGen/AIE/GlobalISel/simplify-concat-unmerge-phi.mir
```diff
@@ -0,0 +1,809 @@
...
+  ; COMMON-NEXT: {{  $}}
+  ; COMMON-NEXT:   [[PHI:%[0-9]+]]:_(<32 x s32>) = G_PHI [[CONCAT_VECTORS]](<32 x s32>), %bb.0, %23(<32 x s32>), %bb.1
+  ; COMMON-NEXT:   [[PHI1:%[0-9]+]]:_(p0) = G_PHI [[COPY]](p0), %bb.0, %10(p0), %bb.1
+  ; COMMON-NEXT:   [[C2:%[0-9]+]]:_(s20) = G_CONSTANT i20 128
+  ; COMMON-NEXT:   [[PTR_ADD:%[0-9]+]]:_(p0) = nuw G_PTR_ADD [[PHI1]], [[C2]](s20)
+  ; COMMON-NEXT:   [[BITCAST:%[0-9]+]]:_(<16 x s64>) = G_BITCAST [[PHI]](<32 x s32>)
+  ; COMMON-NEXT:   G_STORE [[BITCAST]](<16 x s64>), [[COPY2]](p0) :: (store (<16 x s64>))
+  ; COMMON-NEXT:   [[LOAD:%[0-9]+]]:_(<16 x s64>) = G_LOAD [[COPY3]](p0) :: (load (<16 x s64>))
+  ; COMMON-NEXT:   [[BITCAST1:%[0-9]+]]:_(<32 x s32>) = G_BITCAST [[LOAD]](<16 x s64>)
+  ; COMMON-NEXT:   [[UV:%[0-9]+]]:_(<16 x s32>), [[UV1:%[0-9]+]]:_(<16 x s32>) = G_UNMERGE_VALUES [[BITCAST1]](<32 x s32>)
```
andcarminati: Curious that these unmerges are still here. Usually, we have DCE while combining.

andcarminati:
Oh boy, we hit another bug in the tests (if you try to legalize the code):
```
llc: ../llvm/lib/CodeGen/GlobalISel/MachineIRBuilder.cpp:1150: void llvm::MachineIRBuilder::validateTruncExt(const llvm::LLT, const llvm::LLT, bool): Assertion `TypeSize::isKnownLT(DstTy.getSizeInBits(), SrcTy.getSizeInBits()) && "invalid widening trunc"' failed.
PLEASE submit a bug report to https://github.com/llvm/llvm-project/issues/ and include the crash backtrace.
Stack dump:
0.      Program arguments: ./bin/llc -mtriple aie2p -start-before=aie2p-prelegalizer-combiner ../llvm/test/CodeGen/AIE/GlobalISel/simplify-concat-unmerge-phi.mir -o concat.mir
1.      Running pass 'Function Pass Manager' on module '../llvm/test/CodeGen/AIE/GlobalISel/simplify-concat-unmerge-phi.mir'.
2.      Running pass 'Legalizer' on function '@ReferenceConcatUnMergePattern'
 #0 0x000055de48592727 llvm::sys::PrintStackTrace(llvm::raw_ostream&, int) ./../llvm/lib/Support/Unix/Signals.inc:723:13
 #1 0x000055de48590820 llvm::sys::RunSignalHandlers() ./../llvm/lib/Support/Signals.cpp:106:18
 #2 0x000055de48592dea SignalHandler(int) ./../llvm/lib/Support/Unix/Signals.inc:413:1
 #3 0x00007f108189f520 (/lib/x86_64-linux-gnu/libc.so.6+0x42520)
 #4 0x00007f10818f39fc __pthread_kill_implementation ./nptl/pthread_kill.c:44:76
 #5 0x00007f10818f39fc __pthread_kill_internal ./nptl/pthread_kill.c:78:10
 #6 0x00007f10818f39fc pthread_kill ./nptl/pthread_kill.c:89:10
 #7 0x00007f108189f476 gsignal ./signal/../sysdeps/posix/raise.c:27:6
 #8 0x00007f10818857f3 abort ./stdlib/abort.c:81:7
 #9 0x00007f108188571b _nl_load_domain ./intl/loadmsgcat.c:1177:9
#10 0x00007f1081896e96 (/lib/x86_64-linux-gnu/libc.so.6+0x39e96)
#11 0x000055de489737b8 (./bin/llc+0x37237b8)
#12 0x000055de48974855 llvm::MachineIRBuilder::buildInstr(unsigned int, llvm::ArrayRef<llvm::DstOp>, llvm::ArrayRef<llvm::SrcOp>, std::optional<unsigned int>) ./../llvm/lib/CodeGen/GlobalISel/MachineIRBuilder.cpp:0:5
#13 0x000055de488b2bc8 llvm::MachineIRBuilder::getCSEInfo() const ./../llvm/include/llvm/CodeGen/GlobalISel/MachineIRBuilder.h:317:57
#14 0x000055de488b2bc8 llvm::CSEMIRBuilder::canPerformCSEForOpc(unsigned int) const ./../llvm/lib/CodeGen/GlobalISel/CSEMIRBuilder.cpp:62:33
#15 0x000055de488b2bc8 llvm::CSEMIRBuilder::memoizeMI(llvm::MachineInstrBuilder, void*) ./../llvm/lib/CodeGen/GlobalISel/CSEMIRBuilder.cpp:125:3
#16 0x000055de488b2bc8 llvm::CSEMIRBuilder::buildInstr(unsigned int, llvm::ArrayRef<llvm::DstOp>, llvm::ArrayRef<llvm::SrcOp>, std::optional<unsigned int>) ./../llvm/lib/CodeGen/GlobalISel/CSEMIRBuilder.cpp:320:10
#17 0x000055de48972465 llvm::MachineIRBuilder::buildTrunc(llvm::DstOp const&, llvm::SrcOp const&) ./../llvm/lib/CodeGen/GlobalISel/MachineIRBuilder.cpp:872:3
#18 0x000055de46d13b8e llvm::AIELegalizerHelper::legalizeLoopDecrement(llvm::LegalizerHelper&, llvm::MachineInstr&) const ./../llvm/lib/Target/AIE/AIELegalizerHelper.cpp:1393:3
#19 0x000055de48921a4e llvm::LegalizerHelper::legalizeInstrStep(llvm::MachineInstr&, llvm::LostDebugLocObserver&) ./../llvm/lib/CodeGen/GlobalISel/LegalizerHelper.cpp:0:0
#20 0x000055de48917c2e llvm::Legalizer::legalizeMachineFunction(llvm::MachineFunction&, llvm::LegalizerInfo const&, llvm::ArrayRef<llvm::GISelChangeObserver*>, llvm::LostDebugLocObserver&, llvm::MachineIRBuilder&, llvm::GISelKnownBits*) ./../llvm/lib/CodeGen/GlobalISel/Legalizer.cpp:242:15
#21 0x000055de489196be llvm::Legalizer::runOnMachineFunction(llvm::MachineFunction&) ./../llvm/lib/CodeGen/GlobalISel/Legalizer.cpp:351:7
#22 0x000055de4779a940 llvm::MachineFunctionPass::runOnFunction(llvm::Function&) ./../llvm/lib/CodeGen/MachineFunctionPass.cpp:0:13
#23 0x000055de47c594b9 llvm::FPPassManager::runOnFunction(llvm::Function&) ./../llvm/lib/IR/LegacyPassManager.cpp:0:27
#24 0x000055de47c5fb81 llvm::FPPassManager::runOnModule(llvm::Module&) ./../llvm/lib/IR/LegacyPassManager.cpp:1487:13
#25 0x000055de47c59bd7 (anonymous namespace)::MPPassManager::runOnModule(llvm::Module&) ./../llvm/lib/IR/LegacyPassManager.cpp:0:27
#26 0x000055de47c59bd7 llvm::legacy::PassManagerImpl::run(llvm::Module&) ./../llvm/lib/IR/LegacyPassManager.cpp:541:44
#27 0x000055de46c35504 compileModule(char**, llvm::LLVMContext&) ./../llvm/tools/llc/llc.cpp:741:17
#28 0x000055de46c35504 main ./../llvm/tools/llc/llc.cpp:408:22
#29 0x00007f1081886d90 __libc_start_call_main ./csu/../sysdeps/nptl/libc_start_call_main.h:58:16
#30 0x00007f1081886e40 call_init ./csu/../csu/libc-start.c:128:20
#31 0x00007f1081886e40 __libc_start_main ./csu/../csu/libc-start.c:379:5
#32 0x000055de46c329e5 _start (./bin/llc+0x19e29e5)
Aborted (core dumped)
```

More context offline.

andcarminati:
> Oh boy, we hit another bug in the tests (if you try to legalize the code):
> 
> ```
> llc: ../llvm/lib/CodeGen/GlobalISel/MachineIRBuilder.cpp:1150: void llvm::MachineIRBuilder::validateTruncExt(const llvm::LLT, const llvm::LLT, bool): Assertion `TypeSize::isKnownLT(DstTy.getSizeInBits(), SrcTy.getSizeInBits()) && "invalid widening trunc"' failed.
> PLEASE submit a bug report to https://github.com/llvm/llvm-project/issues/ and include the crash backtrace.
> Stack dump:
> 0.      Program arguments: ./bin/llc -mtriple aie2p -start-before=aie2p-prelegalizer-combiner ../llvm/test/CodeGen/AIE/GlobalISel/simplify-concat-unmerge-phi.mir -o concat.mir
> 1.      Running pass 'Function Pass Manager' on module '../llvm/test/CodeGen/AIE/GlobalISel/simplify-concat-unmerge-phi.mir'.
> 2.      Running pass 'Legalizer' on function '@ReferenceConcatUnMergePattern'
>  #0 0x000055de48592727 llvm::sys::PrintStackTrace(llvm::raw_ostream&, int) ./../llvm/lib/Support/Unix/Signals.inc:723:13
>  #1 0x000055de48590820 llvm::sys::RunSignalHandlers() ./../llvm/lib/Support/Signals.cpp:106:18
>  #2 0x000055de48592dea SignalHandler(int) ./../llvm/lib/Support/Unix/Signals.inc:413:1
>  #3 0x00007f108189f520 (/lib/x86_64-linux-gnu/libc.so.6+0x42520)
>  #4 0x00007f10818f39fc __pthread_kill_implementation ./nptl/pthread_kill.c:44:76
>  #5 0x00007f10818f39fc __pthread_kill_internal ./nptl/pthread_kill.c:78:10
>  #6 0x00007f10818f39fc pthread_kill ./nptl/pthread_kill.c:89:10
>  #7 0x00007f108189f476 gsignal ./signal/../sysdeps/posix/raise.c:27:6
>  #8 0x00007f10818857f3 abort ./stdlib/abort.c:81:7
>  #9 0x00007f108188571b _nl_load_domain ./intl/loadmsgcat.c:1177:9
> #10 0x00007f1081896e96 (/lib/x86_64-linux-gnu/libc.so.6+0x39e96)
> #11 0x000055de489737b8 (./bin/llc+0x37237b8)
> #12 0x000055de48974855 llvm::MachineIRBuilder::buildInstr(unsigned int, llvm::ArrayRef<llvm::DstOp>, llvm::ArrayRef<llvm::SrcOp>, std::optional<unsigned int>) ./../llvm/lib/CodeGen/GlobalISel/MachineIRBuilder.cpp:0:5
> #13 0x000055de488b2bc8 llvm::MachineIRBuilder::getCSEInfo() const ./../llvm/include/llvm/CodeGen/GlobalISel/MachineIRBuilder.h:317:57
> #14 0x000055de488b2bc8 llvm::CSEMIRBuilder::canPerformCSEForOpc(unsigned int) const ./../llvm/lib/CodeGen/GlobalISel/CSEMIRBuilder.cpp:62:33
> #15 0x000055de488b2bc8 llvm::CSEMIRBuilder::memoizeMI(llvm::MachineInstrBuilder, void*) ./../llvm/lib/CodeGen/GlobalISel/CSEMIRBuilder.cpp:125:3
> #16 0x000055de488b2bc8 llvm::CSEMIRBuilder::buildInstr(unsigned int, llvm::ArrayRef<llvm::DstOp>, llvm::ArrayRef<llvm::SrcOp>, std::optional<unsigned int>) ./../llvm/lib/CodeGen/GlobalISel/CSEMIRBuilder.cpp:320:10
> #17 0x000055de48972465 llvm::MachineIRBuilder::buildTrunc(llvm::DstOp const&, llvm::SrcOp const&) ./../llvm/lib/CodeGen/GlobalISel/MachineIRBuilder.cpp:872:3
> #18 0x000055de46d13b8e llvm::AIELegalizerHelper::legalizeLoopDecrement(llvm::LegalizerHelper&, llvm::MachineInstr&) const ./../llvm/lib/Target/AIE/AIELegalizerHelper.cpp:1393:3
> #19 0x000055de48921a4e llvm::LegalizerHelper::legalizeInstrStep(llvm::MachineInstr&, llvm::LostDebugLocObserver&) ./../llvm/lib/CodeGen/GlobalISel/LegalizerHelper.cpp:0:0
> #20 0x000055de48917c2e llvm::Legalizer::legalizeMachineFunction(llvm::MachineFunction&, llvm::LegalizerInfo const&, llvm::ArrayRef<llvm::GISelChangeObserver*>, llvm::LostDebugLocObserver&, llvm::MachineIRBuilder&, llvm::GISelKnownBits*) ./../llvm/lib/CodeGen/GlobalISel/Legalizer.cpp:242:15
> #21 0x000055de489196be llvm::Legalizer::runOnMachineFunction(llvm::MachineFunction&) ./../llvm/lib/CodeGen/GlobalISel/Legalizer.cpp:351:7
> #22 0x000055de4779a940 llvm::MachineFunctionPass::runOnFunction(llvm::Function&) ./../llvm/lib/CodeGen/MachineFunctionPass.cpp:0:13
> #23 0x000055de47c594b9 llvm::FPPassManager::runOnFunction(llvm::Function&) ./../llvm/lib/IR/LegacyPassManager.cpp:0:27
> #24 0x000055de47c5fb81 llvm::FPPassManager::runOnModule(llvm::Module&) ./../llvm/lib/IR/LegacyPassManager.cpp:1487:13
> #25 0x000055de47c59bd7 (anonymous namespace)::MPPassManager::runOnModule(llvm::Module&) ./../llvm/lib/IR/LegacyPassManager.cpp:0:27
> #26 0x000055de47c59bd7 llvm::legacy::PassManagerImpl::run(llvm::Module&) ./../llvm/lib/IR/LegacyPassManager.cpp:541:44
> #27 0x000055de46c35504 compileModule(char**, llvm::LLVMContext&) ./../llvm/tools/llc/llc.cpp:741:17
> #28 0x000055de46c35504 main ./../llvm/tools/llc/llc.cpp:408:22
> #29 0x00007f1081886d90 __libc_start_call_main ./csu/../sysdeps/nptl/libc_start_call_main.h:58:16
> #30 0x00007f1081886e40 call_init ./csu/../csu/libc-start.c:128:20
> #31 0x00007f1081886e40 __libc_start_main ./csu/../csu/libc-start.c:379:5
> #32 0x000055de46c329e5 _start (./bin/llc+0x19e29e5)
> Aborted (core dumped)
> ```
> 
> More context offline.

Better to use s1 as BRCOND input, as we are before the legalization.

llvm/test/CodeGen/AIE/GlobalISel/simplify-concat-unmerge-phi.mir
F-Stuckmann: yeah DCE only happens during legalization

llvm/lib/Target/AIE/AIECombinerHelper.cpp
F-Stuckmann: added a unit-test for this behavior

llvm/lib/Target/AIE/AIECombinerHelper.cpp
F-Stuckmann: when I create an Instruction with `MachineIRBuilder`, I already register the new instructions with the globalisel Observer

llvm/lib/Target/AIE/AIECombinerHelper.cpp
```diff
@@ -562,6 +562,195 @@ static MachineInstr *findPostIncMatch(MachineInstr &MemI,
...
+  MachineInstr *FirstMI = &*ConcatI.getParent()->begin();
+
+  auto FindPhiUnmergeComponents =
+      [UseOpIdx, &FirstMI, &Helper, &MRI, &MatchData](
+          MachineInstr *PhiInst) -> std::pair<MachineInstr *, Register> {
+    // Find Unmerge and Concat Components of the PHI node
+    MachineInstr *UnmergeI = nullptr;
+    Register UnmergeDefReg;
+    for (unsigned I = 1; I < PhiInst->getNumOperands(); I += 2) {
+      Register IncomingReg = PhiInst->getOperand(I).getReg();
```
andcarminati, line 588: Great!

llvm/lib/Target/AIE/AIECombinerHelper.cpp
```diff
@@ -562,6 +562,195 @@ static MachineInstr *findPostIncMatch(MachineInstr &MemI,
...
+  };
+
+  // Find PHI node
+  const unsigned OperandIdx = UseOpIdx + 1;
+  assert(ConcatI.getOperand(OperandIdx).isReg());
+  const Register Reg = ConcatI.getOperand(OperandIdx).getReg();
+  MachineInstr *PhiInst = MRI.getVRegDef(Reg);
+  const unsigned NumUses =
+      std::distance(MRI.use_instr_nodbg_begin(Reg), MRI.use_instr_nodbg_end());
+  if (!PhiInst->isPHI() || NumUses > 1)
```
andcarminati: We could have this even smaller:
```
-  const unsigned NumUses =
-      std::distance(MRI.use_instr_nodbg_begin(Reg), MRI.use_instr_nodbg_end());
-  if (!PhiInst->isPHI() || NumUses > 1)
+  if (!PhiInst->isPHI() || !MRI.hasOneNonDBGUse(Reg))
     return false;
```

llvm/lib/Target/AIE/AIECombinerHelper.cpp
```diff
@@ -562,6 +562,195 @@ static MachineInstr *findPostIncMatch(MachineInstr &MemI,
...
+    // Find Unmerge and Concat Components of the PHI node
+    MachineInstr *UnmergeI = nullptr;
+    Register UnmergeDefReg;
+    for (unsigned I = 1; I < PhiInst->getNumOperands(); I += 2) {
+      Register IncomingReg = PhiInst->getOperand(I).getReg();
+      MachineBasicBlock *MBB = PhiInst->getOperand(I + 1).getMBB();
+
+      // Checking if MBB belongs to the loop body or the header.
+      // In the case of the header, MBB dominates the phi node
+      if (MBB->empty() || !Helper.dominates(*FirstMI, *MBB->begin())) {
```
andcarminati: I feel this dominance check a bit convoluted. We are trying to emulate BB dominance here using an inverted interface that is based on instructions. Is it possible to use something like:
```
+      UnmergeI = MRI.getVRegDef(IncomingReg);
       // Checking if MBB belongs to the loop body or the header.
       // In the case of the header, MBB dominates the phi node
-      if (MBB->empty() || !Helper.dominates(*FirstMI, *MBB->begin())) {
+      if (MBB->empty() || Helper.dominates(*UnmergeI, *PhiInst)) {
```

?

F-Stuckmann: I agree checking dominance is clearer than checking the inverse

llvm/lib/Target/AIE/AIECombinerHelper.cpp
```diff
@@ -562,6 +562,192 @@ static MachineInstr *findPostIncMatch(MachineInstr &MemI,
...
+    MachineInstr *UnmergeI = nullptr;
+    MachineInstr *IncomingMI = nullptr;
+    Register UnmergeDefReg;
+    for (unsigned I = 1; I < PhiInst->getNumOperands(); I += 2) {
+      Register IncomingReg = PhiInst->getOperand(I).getReg();
+      MachineBasicBlock *MBB = PhiInst->getOperand(I + 1).getMBB();
+      IncomingMI = MRI.getVRegDef(IncomingReg);
+
+      // Checking if MBB belongs to the loop body or the header.
+      // In the case of the header, MBB dominates the phi node
```
andcarminati: super nit: update the comment (MBB -> IncomingMI)

llvm/lib/Target/AIE/AIECombinerHelper.cpp
```diff
@@ -562,6 +562,192 @@ static MachineInstr *findPostIncMatch(MachineInstr &MemI,
...
+  }
+
+  const auto UnmergeUseIdx = UnmergeInst->getNumDefs();
+  const auto UnmergeSourceReg = UnmergeInst->getOperand(UnmergeUseIdx).getReg();
+  // no need to check consistency of MatchData.UnmergeSourceReg across different
+  // UseOpIdx, since UnMergeSrc consistency is checked
+  if (!MatchData.UnmergeSourceReg)
+    MatchData.UnmergeSourceReg = UnmergeSourceReg;
+
+  MachineInstr *UnmergeSrc = MRI.getVRegDef(*MatchData.UnmergeSourceReg);
```
andcarminati: All the information that you need is UnmergeSourceReg. Also we always set UnmergeSourceReg once (first run) and when we execute:

```
MachineInstr *UnmergeSrc = MRI.getVRegDef(*MatchData.UnmergeSourceReg);
```

Getting the same instruction, that will be always `MatchData.UnmergeSource == UnmergeSrc`. In this way, I feel that the code below never rejects a case.

We could simplify:
```
-  if (!MatchData.UnmergeSourceReg)
-    MatchData.UnmergeSourceReg = UnmergeSourceReg;
+  if (MatchData.UnmergeSourceReg && *MatchData.UnmergeSourceReg != UnmergeSourceReg)
+    return false; // what matters here is the register.
 
-  MachineInstr *UnmergeSrc = MRI.getVRegDef(*MatchData.UnmergeSourceReg);
-  if (!UnmergeSrc) {
-    // could not find source of UnMerge
-    return false;
-  }
-
-  if (MatchData.UnmergeSource && MatchData.UnmergeSource != UnmergeSrc) {
-    // Diverging UnMerge Sources
-    return false;
-  }
-
-  if (!MatchData.UnmergeSource)
-    MatchData.UnmergeSource = UnmergeSrc;
+  MatchData.UnmergeSourceReg = UnmergeSourceReg;
```

removing completely `UnmergeSource`.

F-Stuckmann: I don't even need `MatchData.UnmergeSource`. It is already in `UnmergeSourceReg`

llvm/lib/Target/AIE/AIECombinerHelper.cpp
```diff
@@ -562,6 +562,181 @@ static MachineInstr *findPostIncMatch(MachineInstr &MemI,
...
+  if (!UnmergeInst ||
+      UnmergeInst->getOpcode() != TargetOpcode::G_UNMERGE_VALUES ||
+      !MatchData.ConcatMBB) {
+    // Could not find Unmerge or no non-dominating-MBB could be determined.
+    return false;
+  }
+
+  const auto UnmergeUseIdx = UnmergeInst->getNumDefs();
+  const auto UnmergeSourceReg = UnmergeInst->getOperand(UnmergeUseIdx).getReg();
+  // check consistency of MatchData.UnmergeSourceReg across different
```
andcarminati: Maybe you can start all comments in upper case and and with dot.

llvm/lib/Target/AIE/AIECombinerHelper.cpp
```diff
@@ -562,6 +562,181 @@ static MachineInstr *findPostIncMatch(MachineInstr &MemI,
...
+  MachineInstr *PhiInst = MRI.getVRegDef(Reg);
+  if (!PhiInst->isPHI() || !MRI.hasOneNonDBGUse(Reg))
+    return false;
+
+  auto [UnmergeInst, UnmergeDefReg] = FindPhiUnmergeComponents(PhiInst);
+
+  if (!UnmergeInst ||
+      UnmergeInst->getOpcode() != TargetOpcode::G_UNMERGE_VALUES ||
+      !MatchData.ConcatMBB) {
+    // Could not find Unmerge or no non-dominating-MBB could be determined.
```
andcarminati: Could you update this comment, we are checking instruction dominance now, just to be clear.

F-Stuckmann:
I updated the unit-tests for PHI nodes with 3 incoming MBBs

llvm/lib/Target/AIE/AIECombinerHelper.cpp
```diff
@@ -557,6 +557,185 @@ static MachineInstr *findPostIncMatch(MachineInstr &MemI,
...
+      [UseOpIdx, &Helper, &MRI, &MatchData](
+          MachineInstr *PhiInst) -> std::pair<MachineInstr *, Register> {
+    // Find Unmerge and Concat Components of the PHI node.
+    MachineInstr *UnmergeI = nullptr;
+    MachineInstr *IncomingMI = nullptr;
+    Register UnmergeDefReg;
+    for (unsigned I = 1; I < PhiInst->getNumOperands(); I += 2) {
+      Register IncomingReg = PhiInst->getOperand(I).getReg();
+      MachineBasicBlock *MBB = PhiInst->getOperand(I + 1).getMBB();
+      IncomingMI = MRI.getVRegDef(IncomingReg);
```
andcarminati: nit: you can move the full declaration of `IncomingMI` down. You need just `UnmergeI` and  `UnmergeDefReg` in the outer scope.

llvm/lib/Target/AIE/AIECombinerHelper.cpp
```diff
@@ -557,6 +557,185 @@ static MachineInstr *findPostIncMatch(MachineInstr &MemI,
...
+
+  // Find PHI node.
+  const unsigned OperandIdx = UseOpIdx + 1;
+  assert(ConcatI.getOperand(OperandIdx).isReg());
+  const Register Reg = ConcatI.getOperand(OperandIdx).getReg();
+  MachineInstr *PhiInst = MRI.getVRegDef(Reg);
+  if (!PhiInst->isPHI() || !MRI.hasOneNonDBGUse(Reg))
+    return false;
+
+  auto [UnmergeInst, UnmergeDefReg] = FindPhiUnmergeComponents(PhiInst);
```
andcarminati, line 630: super nit: it looks a good case for optional, but I agree that optional pai is too verbose ;-).

llvm/lib/Target/AIE/AIECombinerHelper.cpp
```diff
@@ -557,6 +557,185 @@ static MachineInstr *findPostIncMatch(MachineInstr &MemI,
...
+    MachineInstr *UnmergeI = nullptr;
+    MachineInstr *IncomingMI = nullptr;
+    Register UnmergeDefReg;
+    for (unsigned I = 1; I < PhiInst->getNumOperands(); I += 2) {
+      Register IncomingReg = PhiInst->getOperand(I).getReg();
+      MachineBasicBlock *MBB = PhiInst->getOperand(I + 1).getMBB();
+      IncomingMI = MRI.getVRegDef(IncomingReg);
+
+      // Checking if IncomingMI belongs to the loop body or the header.
+      // In the case of the header, MBB dominates the phi node.
```
andcarminati: nit: better is `IncomingMI dominates the phi node`

llvm/lib/Target/AIE/AIECombinerHelper.cpp
```diff
@@ -557,6 +557,185 @@ static MachineInstr *findPostIncMatch(MachineInstr &MemI,
...
+        continue;
+      }
+
+      // Unmerge Components.
+      if (UnmergeI) {
+        LLVM_DEBUG(dbgs() << "Encountered too many Unmerge Instructions\n");
+        return {};
+      }
+      UnmergeDefReg = IncomingReg;
+      UnmergeI = IncomingMI;
```
andcarminati, line 617: The name of the lambda is `FindPhiUnmergeComponents`, so it make sense to move the check against `TargetOpcode::G_UNMERGE_VALUES` to here. If it is not UNMERGE -> `return {};`. In this case, you can avoid checking this later on.

llvm/lib/Target/AIE/AIECombinerHelper.h
```diff
@@ -30,6 +30,18 @@ struct FrequentIndexResult {
...
 
+struct AIEConcatUnmergeCombineMatchData {
+  // Concat Instruction's MBB where the G_CONCAT of ConcatSub0 and ConcatSub1
+  // should be created.
+  MachineBasicBlock *ConcatMBB = nullptr;
+  // PHI Src Operands
+  std::vector<Register> ConcatSubVecs;
+
+  // PHI components of the backedge
+  // Register for the PHI Source operand 2 sitting in the backedge path
```
andcarminati: nit: not 2 anymore.

llvm/lib/Target/AIE/AIECombinerHelper.h
```diff
@@ -30,6 +30,18 @@ struct FrequentIndexResult {
   unsigned NonMatchingCount;
 };
 
+struct AIEConcatUnmergeCombineMatchData {
+  // Concat Instruction's MBB where the G_CONCAT of ConcatSub0 and ConcatSub1
```
andcarminati: nit: `ConcatSub0 to ConcatSub[n-1] where n is the number of subvectors`

llvm/test/CodeGen/AIE/GlobalISel/simplify-concat-unmerge-phi.mir
```diff
@@ -0,0 +1,820 @@
...
+    G_BR %bb.3
+
+  bb.3:
+    PseudoRET implicit $lr
+
+
+...
+
+# Negative Test:
+# PHI node with 3 MBBs with Multiple dominating Sources
```
andcarminati, line 744: Nice!

llvm/test/CodeGen/AIE/GlobalISel/simplify-concat-unmerge-phi.mir
```diff
@@ -0,0 +1,820 @@
...
+---
+name:            ThreeMBBPHIMultipleDominatingMBBs
+alignment:       16
+legalized:       true
```
andcarminati: Please, remove the `legalized` flag in all tests. But please keep `tracksRegLiveness`, it is useful to use during review process.

llvm/lib/Target/AIE/AIECombinerHelper.h
F-Stuckmann: The new phi node only has two MBBs with the following structure:
- newly generated concat from the dominating MBB
-  UnmergeSource Register from the loop body

llvm/lib/Target/AIE/AIECombinerHelper.cpp
F-Stuckmann: Usually I like optional, but if the default constructor of a container already provides a nullptr, I don't really want to wrap it around optional again

llvm/lib/Target/AIE/AIECombinerHelper.cpp
```diff
@@ -557,6 +557,185 @@ static MachineInstr *findPostIncMatch(MachineInstr &MemI,
...
+  // UseOpIdx.
+  if (MatchData.UnmergeSourceReg &&
+      MatchData.UnmergeSourceReg != UnmergeSourceReg) {
+    // Diverging Unmerge Source Registers.
+    return false;
+  }
+  MatchData.UnmergeSourceReg = UnmergeSourceReg;
+
+  const auto UnmergeDefIdx =
+      UnmergeInst->findRegisterDefOperand(UnmergeDefReg)->getOperandNo();
```
andcarminati: super nit: If we use `findRegisterDefOperandIdx` we don't need `getOperandNo`.

llvm/lib/Target/AIE/AIECombinerHelper.cpp
```diff
@@ -557,6 +557,185 @@ static MachineInstr *findPostIncMatch(MachineInstr &MemI,
...
+bool llvm::matchConcatUnmergePhis(MachineInstr &ConcatI,
+                                  MachineRegisterInfo &MRI,
+                                  CombinerHelper &Helper,
+                                  AIEConcatUnmergeCombineMatchData &MatchInfo) {
+  assert(ConcatI.getOpcode() == TargetOpcode::G_CONCAT_VECTORS);
+  LLVM_DEBUG(dbgs() << "MF: " << ConcatI.getMF()->getName() << "\n");
+
+  const unsigned NumUses = ConcatI.getNumOperands() - ConcatI.getNumDefs();
+
+  MatchInfo.ConcatSubVecs.clear();
```
andcarminati: MatchInfo is not reused, you can skip this if you want.

llvm/lib/Target/AIE/AIECombinerHelper.h
```diff
@@ -30,6 +30,18 @@ struct FrequentIndexResult {
   unsigned NonMatchingCount;
 };
 
+struct AIEConcatUnmergeCombineMatchData {
+  // Concat Instruction's MBB where the G_CONCAT of ConcatSub0 and
+  // ConcatSub[n-1] should be created.
+  MachineBasicBlock *ConcatMBB = nullptr;
+  // PHI Src Operands.
+  std::vector<Register> ConcatSubVecs;
```
andcarminati: What about the usage of a ` SmallVector<Register, 4> ConcatSubVecs;` here? I think it fits our purposes.....

F-Stuckmann: I guess it would be more efficient for the majority of usecases.

llvm/lib/Target/AIE/AIECombinerHelper.cpp
```diff
@@ -557,6 +557,185 @@ static MachineInstr *findPostIncMatch(MachineInstr &MemI,
...
+      Register IncomingReg = PhiInst->getOperand(I).getReg();
+      MachineBasicBlock *MBB = PhiInst->getOperand(I + 1).getMBB();
+      MachineInstr *IncomingMI = MRI.getVRegDef(IncomingReg);
+
+      // Checking if IncomingMI dominates phi node.
+      if (Helper.dominates(*IncomingMI, *PhiInst)) {
+
+        if (MatchData.ConcatMBB && MatchData.ConcatMBB != MBB) {
+          LLVM_DEBUG(dbgs() << "Parts of the Future Concat do not "
+                               "originate in the same MBB.\n");
```
martien-de-jong, line 597: Not dominating the phi node doesn't say much. In an if-then-else before the loop where both branches jump to the header, both blocks do not dominate the phi nodes.

llvm/lib/Target/AIE/AIECombinerHelper.cpp
```diff
@@ -557,6 +557,185 @@ static MachineInstr *findPostIncMatch(MachineInstr &MemI,
...
+          return {};
+        }
+
+        // Concat Components.
+        MatchData.ConcatMBB = MBB;
```
martien-de-jong: Since this is called ConcatMBB, why isn't it set from the start with the MBB of the CONCAT?

llvm/lib/Target/AIE/AIECombinerHelper.cpp
F-Stuckmann: that is correct and in that case the combiner does nothing. There is a unit-test for this use case

llvm/lib/Target/AIE/AIECombinerHelper.cpp
F-Stuckmann: the name is poorly chosen, it is the MBB for the to-be generated Concat instruction . I clarified this.

llvm/lib/Target/AIE/AIECombinerHelper.cpp
martien-de-jong: But why do we test for domination? I'm not saying its incorrect, I'm saying it's superfluous. If was expecting the normal backedge check, which is that the PHI node dominates its UNMERGE use, not its negated reverse

llvm/lib/Target/AIE/AIECombinerHelper.cpp
```diff
@@ -557,6 +557,185 @@ static MachineInstr *findPostIncMatch(MachineInstr &MemI,
...
+
+  auto [UnmergeInst, UnmergeDefReg] = FindPhiUnmergeComponents(PhiInst);
+
+  if (!UnmergeInst || !MatchData.NewConcatMBB) {
+    // Could not find Unmerge or Concat Components.
+    return false;
+  }
+
+  const auto UnmergeUseIdx = UnmergeInst->getNumDefs();
+  const auto UnmergeSourceReg = UnmergeInst->getOperand(UnmergeUseIdx).getReg();
```
martien-de-jong: *UnmergeInst.uses.begin() ?

llvm/lib/Target/AIE/AIECombinerHelper.cpp
```diff
@@ -557,6 +557,185 @@ static MachineInstr *findPostIncMatch(MachineInstr &MemI,
...
+  if (MatchData.UnmergeSourceReg &&
+      MatchData.UnmergeSourceReg != UnmergeSourceReg) {
+    // Diverging Unmerge Source Registers.
+    return false;
+  }
+  MatchData.UnmergeSourceReg = UnmergeSourceReg;
+
+  const unsigned UnmergeDefIdx =
+      UnmergeInst->findRegisterDefOperandIdx(UnmergeDefReg);
+  assert(UnmergeDefIdx != (unsigned)-1);
```
martien-de-jong: I think it would be better to make UnmergeDefIdx int, copying the type of findRegisterDefOperandIdx, You can then assert for >= 0 and do a conversion with no chance on overflow to unsigned for the final return value.

llvm/lib/Target/AIE/AIECombinerHelper.h
```diff
@@ -30,6 +30,18 @@ struct FrequentIndexResult {
...
 };
 
+struct AIEConcatUnmergeCombineMatchData {
+  // Concat Instruction's MBB where the G_CONCAT of ConcatSub0 and
+  // ConcatSub[n-1] should be created.
+  MachineBasicBlock *NewConcatMBB = nullptr;
+  // Concat Source Registers.
+  SmallVector<Register, 4> ConcatSubVecs;
+
+  // PHI components of the backedge
```
martien-de-jong: nit: `component`?

llvm/lib/Target/AIE/AIECombinerHelper.h
```diff
@@ -30,6 +30,18 @@ struct FrequentIndexResult {
   unsigned NonMatchingCount;
 };
 
+struct AIEConcatUnmergeCombineMatchData {
+  // Concat Instruction's MBB where the G_CONCAT of ConcatSub0 and
+  // ConcatSub[n-1] should be created.
```
martien-de-jong, line 35: Use [] on both? Is n always 2, otherwise use 'through' instead of 'and'?

llvm/lib/Target/AIE/AIECombinerHelper.cpp
```diff
@@ -557,6 +557,185 @@ static MachineInstr *findPostIncMatch(MachineInstr &MemI,
...
+
+      // Unmerge Components.
+      if (UnmergeI) {
+        LLVM_DEBUG(dbgs() << "Encountered too many Unmerge Instructions\n");
+        return {};
+      }
+      UnmergeDefReg = IncomingReg;
+      UnmergeI = IncomingMI;
+      if (UnmergeI->getOpcode() != TargetOpcode::G_UNMERGE_VALUES) {
+        LLVM_DEBUG(dbgs() << "Could not find Unmerge Instruction\n");
```
martien-de-jong: nit: an earlier message said that too many UNMERGEs were encountered. Only here do we have evidence of that. Instead we could do an opcode check first on IncomingMI, leaving any update of state variables to after the early returns.

llvm/lib/Target/AIE/AIECombinerHelper.h
F-Stuckmann: unmerge and concat allow more than two operands, so we could also have more than two elements. 
There is also a unit-test for this

martien-de-jong (approved):

andcarminati (approved):
LGTM. Nice work!

PR #503, opened 2025-06-24
[AIE2P] Fix broadcast intrinsics.

SagarMaheshwari99:


clang/lib/Headers/aie2p_scl2vec.h
```diff
@@ -1354,10 +1354,10 @@ INTRINSIC(void *) extract_address(v16int32 v, int idx) {
 }
 // broadcast from scalar (alternative syntax to broadcast to vector)
 INTRINSIC(v64int8)
-broadcast_s8(char b) { return b - v64int8{0}; }
+broadcast_s8(int b) { return (char)b - v64int8{0}; }
```
niwinanto, line 1357: Can we update the tests as well?

niwinanto: As I remember, we reused aie2 tests for aie2p as well. Since at it, can you please fix it for aie2 as well.

martien-de-jong, line 1357: v64int8{b} doesn't work?

SagarMaheshwari99: No, that will only initialize the first element of the vector with `b`. 
But the subtraction `(char)b - v64int8{0}` is performed element-wise due to how vector types work, effectively resulting in a vector where each element is `b`.

We don't even need to specify `0` when initializing a `v64int8` vector, since all elements of a vector are initialized to `0` by default. 
So `(char)b - v64int8{}` will also work.

clang/lib/Headers/aie2p_scl2vec.h
```diff
@@ -1354,10 +1354,10 @@ INTRINSIC(void *) extract_address(v16int32 v, int idx) {
...
 
 INTRINSIC(v32int16)
-broadcast_s16(short b) { return b - v32int16{0}; }
+broadcast_s16(int b) { return (short)b - v32int16{0}; }
```
F-Stuckmann, line 1360: why weren't we happy with implicit casting?

martien-de-jong: short b is promoted to int

PR #504, opened 2025-06-24
[AIE2P] Enable stack size section for start up code.

SagarMaheshwari99:


konstantinschwarz (approved):

libc/startup/baremetal/aie2p/CMakeLists.txt
```diff
@@ -3,7 +3,8 @@
 # See https://llvm.org/LICENSE.txt for license information.
 # SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
 #
-# (c) Copyright 2024 Advanced Micro Devices, Inc. or its affiliates
+# (c) Copyright 2024-2025 Advanced Micro Devices, Inc. or its affiliates
+
```
andcarminati, line 7: I see that we are reusing aie2 startup in aie2p (including asm files). They are similar but not the same, so do you think it is better to have a clear separation here?

libc/startup/baremetal/aie2/crt1.cc
```diff
@@ -3,13 +3,10 @@
...
-// (c) Copyright 2024 Advanced Micro Devices, Inc. or its affiliates
+// (c) Copyright 2024-2025 Advanced Micro Devices, Inc. or its affiliates
 
 extern "C" {
 extern int main(int, char **);
-void _Exit(int val) {
-  (void)val;
-  done();
-}
+void _Exit(int) { done(); }
```
martien-de-jong, line 10: This will now get name mangled. _Exit is a defined libc symbol,

libc/startup/baremetal/aie2p/CMakeLists.txt
martien-de-jong: crt1.cc is supposed to be portable. only crt0 contains the arch-dependent stuff of setting up the initial stack pointer and calling _main_init. Perhaps we can move crt1.cc to a 'common' aie directory.

libc/startup/baremetal/aie2/crt1.cc
SagarMaheshwari99: but it is inside `extern "C"`

libc/startup/baremetal/aie2p/CMakeLists.txt
SagarMaheshwari99: We can take this up when we finalize code for future archs.

PR #506, opened 2025-06-26
[AIE2] Add missing vector legalization rules for G_PHI

andcarminati:


konstantinschwarz (commented):
Can you also add instruction selection tests to make sure this is handled correctly?

andcarminati:
> Can you also add instruction selection tests to make sure this is handled correctly?

Hi @konstantinschwarz, now we have legalization, register bank and selection tests.

F-Stuckmann:
Could we have the GlobalISEL tests in `llvm/test/CodeGen/AIE/GlobalISel`

andcarminati:
> Could we have the GlobalISEL tests in `llvm/test/CodeGen/AIE/GlobalISel`

In this case, it is not possible. Phi node are legalization involves different target types, so it is aie2x-specific.

F-Stuckmann (approved):

PR #507, opened 2025-06-26
Allow for more than 64 scheduler resources

martien-de-jong:
More more latest news:

QoR is done. We have about 10% of the benchmarks that gain 1 unit (16 bytes) of PM, and a very slight blip for instruction count.

I have created https://jira.xilinx.com/browse/AIECC-1021 to deal with the regression afterwards.

More Latest news:

StaticBitSet is now implemented using std::bitset.

Latest news:
This is a functionally correct version but not the definitive one. It has fixed a pre-existing bug in the hazardrecognizer that crashed due to a more strict assert. I agreed with Andreu to change ResourceSet into an interface class on top of std::bitset, with a few name changes that better align with HazardRecognizer parlance and hiding a constructor which is ambiguous and awkward in our scheme of things. 
I added unit tests for ResourceScoreboard and StaticBitSet.
There is a slight regression due to a correction of the pipelinedepth for memory operations, which tends to add nop cycles to exit blocks.

I sneak in two NFC commits that I had in a working branch. The second one centralizes all itinerary queries, which is useful for this PR. Taking the first one prevented a cherry-pick conflict.

The core of the change is that tablegen now generates the initializer for the resource through a macro, and the file that includes the generated data can change that macro.
We employ this to just list the resource index, not the expanded bitvector. This index can be converted to the ADT that can now be stretched to an arbitrary size locally in AIEHazardRecognizer.h


llvm/lib/Target/AIE/AIEHazardRecognizer.h
```diff
@@ -82,10 +83,12 @@ class FuncUnitWrapper {
 
   FuncUnitWrapper(const InstrStage &IS, SlotBits Slots = 0,
                   MemoryBankBits MemoryBanks = 0, uint64_t MemObjectsBits = 0)
-      : Required(IS.getReservationKind() == InstrStage::Required ? IS.getUnits()
-                                                                 : 0),
-        Reserved(IS.getReservationKind() == InstrStage::Reserved ? IS.getUnits()
-                                                                 : 0),
+      : Required(IS.getReservationKind() == InstrStage::Required
+                     ? (One << IS.getUnits())
+                     : 0),
```
martien-de-jong: This effectively computes the set that was generated by tablegen before.

llvm/lib/Target/AIE/AIEHazardRecognizer.cpp
```diff
@@ -99,7 +99,7 @@ void FuncUnitWrapper::clearResources() {
   MemObjectsBits = 0;
 }
 bool FuncUnitWrapper::isEmpty() const {
-  return Required == 0 && Reserved == 0 && Slots == 0 && MemoryBanks == 0;
+  return !Required && !Reserved && Slots == 0 && MemoryBanks == 0;
```
martien-de-jong: Not stricly necessary (anymore). 0 was interpreted as a set containing resource 0. I guess I'll add an .empty() method

llvm/lib/Target/AIE/AIEHazardRecognizer.cpp
```diff
@@ -121,9 +121,9 @@ FuncUnitWrapper &FuncUnitWrapper::operator|=(const FuncUnitWrapper &Other) {
 }
 
 bool FuncUnitWrapper::conflict(const FuncUnitWrapper &Other) const {
-  if ((Required & Other.Required) != 0 || (Slots & Other.Slots) != 0 ||
-      (MemoryBanks & Other.MemoryBanks) != 0 ||
-      (Reserved & Other.Required) != 0 || (Required & Other.Reserved) != 0 ||
+  if ((Required & Other.Required) || (Slots & Other.Slots) != 0 ||
+      (MemoryBanks & Other.MemoryBanks) != 0 || (Reserved & Other.Required) ||
+      (Required & Other.Reserved) ||
```
martien-de-jong: We may reorder these a bit, since the Required test has become more expensive.

llvm/lib/Target/AIE/AIEHazardRecognizer.h
```diff
@@ -57,16 +57,83 @@ const_bundled_instrs(const MachineInstr &MI, bool IncludeRoot = false) {
   return make_range(Begin, getBundleEnd(MI.getIterator()));
 }
 
+class ResourceSet {
+  static const int NumContainers = 1;
+  static const int Shift = 6;
+  static const int Mask = 0x3f;
+  static const uint64_t One = 1;
+  uint64_t Bits[NumContainers];
```
martien-de-jong: For initial stress testing, I'll change this temporarily to multiple uint8_t, so that aie2 and aie2p will definitely hit the second container

llvm/include/llvm/CodeGen/ResourceScoreboard.h
```diff
@@ -70,27 +65,37 @@ template <typename RC> class ResourceScoreboard {
...
-    Depth = Pow2;
-    Size = 2 * Depth;
+    Size = Pow2;
     clear();
   }
-  bool isValidDelta(int DeltaCycles) const {
-    return DeltaCycles >= -Depth && DeltaCycles <= 0;
+
+  bool isInRange(int Index) const {
+    return Index >= Lowest && Index <= Highest;
```
martien-de-jong: The former applied just to the insertion point of an instruction. The latter checks each access.

llvm/include/llvm/CodeGen/ResourceScoreboard.h
```diff
@@ -179,16 +179,12 @@ template <typename RC> class ResourceScoreboard {
...
   }
 
-  // FIXME: These are relics from the original user of the scoreboard. mainly
-  // hazardrecognizer variants. They should be moved there. In that original
-  // use, D was determined by the pipeline depth of the itineraries. We would
+  // This is a relic from the original user of the scoreboard, mainly
+  // hazardrecognizer variants. In that original use, D was determined by
+  // the pipeline depth of the itineraries. We would
   // double the size and putting the origin in the middle, so that we could
   // insert in cycles < 0. isValidDelta represents this sanity check
```
martien-de-jong: nit: comment out of date.

llvm/utils/TableGen/SubtargetEmitter.cpp
```diff
@@ -446,12 +446,17 @@ void SubtargetEmitter::EmitStageAndOperandCycleData(
       continue;
 
     StringRef Name = ProcModel.ItinsDef->getName();
+    OS << "\n"
+          "#ifndef FUNCUNIT_REPRESENTATION\n"
```
andcarminati, line 452: Nice, this give us a clear and less intrusive way to extend.

llvm/include/llvm/CodeGen/ResourceScoreboard.h
```diff
@@ -70,27 +65,37 @@ template <typename RC> class ResourceScoreboard {
...
+    // Implementation relies on masking for wrap-around, so round up
     // to a power of two.
     int Pow2 = 1;
-    while (Pow2 < D) {
+    while (Pow2 < Req) {
       Pow2 += Pow2;
     }
-    Depth = Pow2;
-    Size = 2 * Depth;
+    Size = Pow2;
```
andcarminati, line 84: Curious about the 2 * Depth that was here before. Non explainable safety margin maybe?

llvm/include/llvm/CodeGen/ResourceScoreboard.h
```diff
@@ -35,25 +35,20 @@ template <typename RC> class ResourceScoreboard {
...
 
-  /// The scoreboard extends from Head - Depth to Head + Depth - 1
-  /// Depth is half of Size. Emitting an instruction is allowed in the
-  /// range [-Depth, 0], which guarantees that all of its resources are
-  /// recorded.
+  /// The scoreboard extends from Head - Lowest to Head + Highest
   /// When querying for conflicts, we have more liberty, with out-of-range
   /// cycles interpreted as being empty.
-  int Depth = 0;
+  int Lowest = 0;
```
andcarminati: nit: maybe `LowestCycle` and `HighestCycle` to have a unit here.

llvm/lib/Target/AIE/AIEHazardRecognizer.cpp
```diff
@@ -796,4 +802,4 @@ AIEHazardRecognizer::getMemoryObjectsBits(const MachineInstr *MI) const {
       ObjectsToBitMap |= One << *ObjectNumber;
   }
   return ObjectsToBitMap;
-}
+}
```
andcarminati: nit: new line here!

llvm/lib/Target/AIE/AIEHazardRecognizer.cpp
```diff
@@ -543,34 +568,21 @@ bool AIEHazardRecognizer::checkConflict(
...
-
-      if (ThisCycle.conflict(Scoreboard[StageCycle])) {
-        LLVM_DEBUG(dbgs() << "*** Hazard in cycle=" << StageCycle
-                          << " EC=" << StageCycle - DeltaCycles << ":\n";
-                   ThisCycle.dump(); dbgs() << "\n");
-        return true;
-      }
+  // Note that DeltaCycles will be negative for bottom-up scheduling.
+  auto CycleConflict = [&Scoreboard,
+                        DeltaCycles](int C, const FuncUnitWrapper &ThisCycle) {
```
andcarminati: Can we rename this `C` to a more meaningful name?

llvm/lib/Target/AIE/AIEHazardRecognizer.cpp
```diff
@@ -543,34 +568,21 @@ bool AIEHazardRecognizer::checkConflict(
...
+    int StageCycle = DeltaCycles + C;
```
andcarminati: nit: `const int StageCycle`...

llvm/lib/Target/AIE/AIEHazardRecognizer.cpp
```diff
@@ -632,24 +643,22 @@ void AIEHazardRecognizer::enterResources(
...
-      Scoreboard[Cycle + C] |= ThisCycle;
-    }
+  Scoreboard[DeltaCycles].IssueCount++;
 
-    // Advance the cycle to the next stage.
-    Cycle += IS.getNextCycles();
-  }
+  auto Insert = [&Scoreboard, DeltaCycles](int C,
+                                           const FuncUnitWrapper &ThisCycle) {
+    int ScoreboardCycle = C + DeltaCycles;
```
andcarminati: nit: const

llvm/include/llvm/CodeGen/ResourceScoreboard.h
martien-de-jong: No, that was an explicit change introduced for backward scheduling. Depth came in as ~pipeline depth, but backward scheduling wants to schedule the most urgent one first; that required a negative insertion cycle, otherwise we would need to wait until current cycle was bumped far enough. In that waiting time, shorter instructions would start eating away at the resources required to schedule the longer one.

llvm/lib/Target/AIE/AIEHazardRecognizer.h
```diff
@@ -57,16 +57,83 @@ const_bundled_instrs(const MachineInstr &MI, bool IncludeRoot = false) {
...
+  // Strong type for overloading a convenient constructor.
+  class Index {
+    int Value;
+
+  public:
+    Index(int Value) : Value(Value) {};
+    int get() const { return Value; }
+  };
+
+  ResourceSet() {
```
andcarminati: maybe just:
```
  ResourceSet() {
    for (auto &BitsPart : Bits) {
      BitsPart = 0;
    }
  }
```

martien-de-jong: If that works for a low-level array, fine.

llvm/lib/Target/AIE/AIEHazardRecognizer.h
```diff
@@ -57,16 +57,83 @@ const_bundled_instrs(const MachineInstr &MI, bool IncludeRoot = false) {
...
+
+  ResourceSet operator~() const {
+    ResourceSet Result;
+    for (int I = 0; I < NumContainers; I++) {
+      Result.Bits[I] = ~Bits[I];
+    }
+    return Result;
+  }
+
+  operator bool() const {
```
andcarminati: maybe just:
```
  operator bool() const {
    return llvm::any_of(Bits, [](auto BitsPart) { return BitsPart; });
  }
```

llvm/lib/Target/AIE/AIEHazardRecognizer.h
```diff
@@ -58,7 +58,9 @@ const_bundled_instrs(const MachineInstr &MI, bool IncludeRoot = false) {
 }
 
 class ResourceSet {
-  static const int NumContainers = 1;
+  // This number should be big enough to accommodate
+  // all AIE targets.
```
martien-de-jong: I will look into computing the max from the tablegen files. We already collect all target itinerary inc files in a single file.

llvm/lib/Target/AIE/AIEHazardRecognizer.h
```diff
@@ -57,16 +57,83 @@ const_bundled_instrs(const MachineInstr &MI, bool IncludeRoot = false) {
...
+    }
+    return *this;
+  }
+  ResourceSet operator&(const ResourceSet &Other) const {
+    ResourceSet Result(*this);
+    Result &= Other;
+    return Result;
+  }
+
+  ResourceSet operator~() const {
```
andcarminati: Is this way too disruptive?

```
  ResourceSet operator~() const {
    ResourceSet Result;

    std::transform(std::begin(Bits), std::end(Bits), std::begin(Result.Bits),
                   [&](auto Part) { return ~Part; });

    return Result;
  }
```

llvm/lib/Target/AIE/AIEHazardRecognizer.h
```diff
@@ -57,16 +57,83 @@ const_bundled_instrs(const MachineInstr &MI, bool IncludeRoot = false) {
...
+    }
+    return *this;
+  }
+
+  ResourceSet &operator&=(const ResourceSet &Other) {
+    for (int I = 0; I < NumContainers; I++) {
+      Bits[I] &= Other.Bits[I];
+    }
+    return *this;
+  }
```
andcarminati: New line here.

llvm/lib/Target/AIE/AIEHazardRecognizer.h
```diff
@@ -57,16 +57,83 @@ const_bundled_instrs(const MachineInstr &MI, bool IncludeRoot = false) {
...
+  ResourceSet(uint64_t Value) : ResourceSet() { Bits[0] = Value; }
+
+  ResourceSet &operator|=(const ResourceSet &Other) {
+    for (int I = 0; I < NumContainers; I++) {
+      Bits[I] |= Other.Bits[I];
+    }
+    return *this;
+  }
+
+  ResourceSet &operator&=(const ResourceSet &Other) {
```
andcarminati: How about this?

```
  ResourceSet &operator&=(const ResourceSet &Other) {

    std::transform(std::begin(Bits), std::end(Bits), std::begin(Other.Bits),
                   std::begin(Bits),
                   [&](auto Part, auto PartOther) { return Part & PartOther; });
    return *this;
  }
```

andcarminati:
Nice update @martien-de-jong on this infrastructure. I left some non-critical comments as a first round. Also, can we have some mock tests to stress the implementation?

llvm/lib/Target/AIE/AIEHazardRecognizer.h
martien-de-jong: I will not give you my honest opinion, it might hurt.

llvm/utils/TableGen/SubtargetEmitter.cpp
```diff
@@ -1507,6 +1509,32 @@ void SubtargetEmitter::EmitProcessorModels(raw_ostream &OS) {
   }
 }
 
+//
+// EmitNumResources
+// Emit a const expr reprenting the number of resources
```
martien-de-jong: nit: typo.

llvm/lib/Target/AIE/AIEHazardRecognizer.h
```diff
@@ -57,15 +58,20 @@ const_bundled_instrs(const MachineInstr &MI, bool IncludeRoot = false) {
...
 }
 
+namespace {
+using BitBlock = uint64_t;
+constexpr const int BitsPerContainer = sizeof(BitBlock) * CHAR_BIT;
+constexpr int numContainers(int NumResources) {
+  return (NumResources + BitsPerContainer - 1) / BitsPerContainer;
+}
+constexpr int NumContainers = numContainers(TotalNumResources);
+} // namespace
```
martien-de-jong: AN earlier version of this commit had this in AIEMaxNumResources.h.  I think it is better to have it here.

llvm/lib/Target/AIE/AIEHazardRecognizer.h
martien-de-jong: It's just not the way that I would like to convince the reader that this is a loop which will be unrolled to (on intel) two instructions.

llvm/lib/Target/AIE/AIEHazardRecognizer.h
```diff
@@ -57,15 +58,113 @@ const_bundled_instrs(const MachineInstr &MI, bool IncludeRoot = false) {
...
+  // This takes the index of a resource and construct a set with that element.
+  explicit ResourceSet(Index Wrapped) : ResourceSet() {
+    int Idx = Wrapped.get();
+    int BlockIdx = Idx >> Shift;
+    assert(BlockIdx < NumContainers && "Adjust NumContainers");
+    Bits[BlockIdx] |= (One << (Idx & Mask));
+  }
+
+  // Mainly for generating empty sets
+  ResourceSet(uint64_t Value) : ResourceSet() { Bits[0] = Value; }
```
andcarminati: Why do we have `Bits[0] = Value;` here? Can we just use `ResourceSet(uint64_t) : ResourceSet() {}`?

llvm/lib/Target/AIE/AIEHazardRecognizer.h
```diff
@@ -82,10 +181,12 @@ class FuncUnitWrapper {
 
   FuncUnitWrapper(const InstrStage &IS, SlotBits Slots = 0,
                   MemoryBankBits MemoryBanks = 0, uint64_t MemObjectsBits = 0)
```
andcarminati: Oh, this was left behind. Could you please also update `uint64_t` to `MemoryObjectsBits`?

llvm/lib/Target/AIE/AIEHazardRecognizer.h
martien-de-jong: This is a backward compatible constructor, which can initialize a 'new' bit set with an 'old' initializer  We don't really need it I guess. What do you think?

llvm/lib/Target/AIE/AIEHazardRecognizer.h
martien-de-jong: And change MemoryObjectBits to a larger set ? ;-)

andcarminati: Now we have the recipe, but I think we can stick to 64bit for now ;-)

llvm/lib/Target/AIE/AIEHazardRecognizer.h
andcarminati: Actually, we need this constructor, but it could be this simpler version I guess.

llvm/lib/Target/AIE/AIEMaxNumResources.h
```diff
@@ -0,0 +1,31 @@
...
+#include "AIEGenSubtargetInfo.inc"
+
+#define GET_NUM_RESOURCES
+#include "AIE2GenSubtargetInfo.inc"
+
+#define GET_NUM_RESOURCES
+#include "AIE2PGenSubtargetInfo.inc"
+
+constexpr const int TotalNumResources = std::max(
+    AIE2PItineraries::NumResources,
```
andcarminati: I like this flexibility, we don't need to guess this information. Maybe we should write a comment saying that this file should be updated for every new target.

llvm/lib/Target/AIE/AIEMaxNumResources.h
```diff
@@ -0,0 +1,31 @@
...
+    std::max(AIE2Itineraries::NumResources, AIEItineraries::NumResources));
```
andcarminati: nit (just one max):
```
constexpr const int TotalNumResources =
    std::max({AIE2PItineraries::NumResources, AIE2Itineraries::NumResources,
              AIEItineraries::NumResources});
```

llvm/lib/Target/AIE/AIEMaxNumResources.h
```diff
@@ -0,0 +1,31 @@
...
+//
+// (c) Copyright 2025 Advanced Micro Devices, Inc. or its affiliates
+//
+//===----------------------------------------------------------------------===//
+
+#ifndef LLVM_LIB_TARGET_AIE_AIEMAXNUMRESOURCES_H
+#define LLVM_LIB_TARGET_AIE_AIEMAXNUMRESOURCES_H
+
+#include <algorithm>
+#include <climits>
```
andcarminati: Unnecessary (climits and cstdint )?

llvm/lib/Target/AIE/AIEHazardRecognizer.cpp
```diff
@@ -634,24 +650,23 @@ void AIEHazardRecognizer::enterResources(
     FuncUnitWrapper MemoryBankAndObjectsAccessCycle(
         /*Req=*/0, /*Res=*/0, /*SlotSet=*/0, MemoryBanks, MemObjectsBits);
     for (auto Cycles : MemoryAccessCycles) {
+      assert(Scoreboard.isInRange(DeltaCycles + Cycles - 1));
```
martien-de-jong, line 655: nit: I guess we may want to put the assert in the `operator[]` method

llvm/lib/Target/AIE/StaticBitSet.h
```diff
@@ -0,0 +1,108 @@
...
+      (NumBits + BitsPerContainer - 1) / BitsPerContainer;
+  Container Bits[NumContainers];
+
+public:
+  StaticBitSet() { clear(); }
+  // This takes the index of a resource and constructs a set containing that
+  // element.
+  explicit StaticBitSet(int BitNo) : StaticBitSet() {
+    int BlockIdx = BitNo >> Shift;
+    assert(BlockIdx < NumContainers && "Adjust NumContainers");
```
martien-de-jong: nit: perhaps "Adjust NumBits"  or "BitNo out of Range"

llvm/lib/Target/AIE/StaticBitSet.h
```diff
@@ -0,0 +1,108 @@
...
+    for (int I = 0; I < NumContainers; I++) {
+      Bits[I] &= Other.Bits[I];
+    }
+    return *this;
+  }
+  StaticBitSet operator&(const StaticBitSet &Other) const {
+    StaticBitSet Result(*this);
+    Result &= Other;
+    return Result;
+  }
```
martien-de-jong, line 40: I actually don't use this yet, but it looks ugly to break symmetry between & and |

llvm/lib/Target/AIE/AIEMaxNumResources.h
martien-de-jong: git grep -l "^#include .*\.inc" in the AIE directory?

llvm/lib/Target/AIE/AIEMaxNumResources.h
martien-de-jong: Ah, that's the syntax, thanks! I was expecting some recursive forwarding construct

llvm/lib/Target/AIE/AIEHazardRecognizer.h
martien-de-jong: We only needed the constructor for the empty set. I made all of those more explicit.

llvm/lib/Target/AIE/StaticBitSet.h
martien-de-jong: Also, if we start using it elsewhere, we should have a  set difference operation, i.e. a & ~b
Perhaps also an iterator, which may require masking after set negation so that we don't list entries outside of NumBits

llvm/unittests/CodeGen/ResourceScoreboardTest.cpp
```diff
@@ -0,0 +1,143 @@
...
+
+  for (int K = 12; K < 32; K++) {
+    Scoreboard.advance();
+    Scoreboard[11] = RC(K);
+  }
+  // Check the range
+  for (int I = 0; I <= 11; I++) {
+    EXPECT_EQ(Scoreboard[11 - I], 31 - I);
+  }
+}
```
andcarminati, line 146: nit: new line here.

llvm/unittests/CodeGen/ResourceScoreboardTest.cpp
```diff
@@ -0,0 +1,143 @@
+//===- ResourceScoreboardTest.cpp --------------------------------*- C++
+//-*-===//
```
martien-de-jong: grmbl

llvm/unittests/Target/AIE/StaticBitSetTest.cpp
```diff
@@ -0,0 +1,123 @@
...
+//
+// (c) Copyright 2025 Advanced Micro Devices, Inc. or its affiliates
+//
+//===----------------------------------------------------------------------===//
+
+#include "gtest/gtest.h"
+
+#include "StaticBitSet.h"
+
+// I like primes.
```
andcarminati, line 15: ;-)

andcarminati:
As discussed offline, we can use `std::bitset` or even `llvm/ADT/Bitset.h`.

llvm/lib/Target/AIE/StaticBitSet.h
```diff
@@ -0,0 +1,108 @@
...
+// This is a very low level implementation of a bitset, designed to be fully
+// static
+//===----------------------------------------------------------------------===//
+
+#ifndef LLVM_LIB_TARGET_AIE_STATICBITSET_H
+#define LLVM_LIB_TARGET_AIE_STATICBITSET_H
+
+#include <cassert>
+#include <climits>
+
```
martien-de-jong: This is a least effort implementation of a bitset that meets our purposes. The functionality is very close to std::bitset, but we can configure the container size for stress testing, and we only have a non-default constructor that interprets its single operand as a bitindex, rather than as a bitset. This meshes better with the resource interpretation as we use it.

llvm/lib/Target/AIE/StaticBitSet.h
```diff
@@ -0,0 +1,55 @@
...
+// This is a very low level implementation of a bitset, designed to be fully
+// static
+//===----------------------------------------------------------------------===//
+
+#ifndef LLVM_LIB_TARGET_AIE_STATICBITSET_H
+#define LLVM_LIB_TARGET_AIE_STATICBITSET_H
+
+#include <bitset>
+
+template <const int NumBits> class StaticBitSet : public std::bitset<NumBits> {
```
andcarminati, line 20: Sounds good!

andcarminati:
> There is a slight regression due to a correction of the pipelinedepth for memory operations, which tends to add nop cycles to exit blocks.

Hi @martien-de-jong, does it mean that we were optimistic before? I am curious about the QoR effect.

llvm/lib/Target/AIE/StaticBitSet.h
```diff
@@ -0,0 +1,55 @@
+//===--- StaticBitSet.h - Fully static bitset. ----------------------------===//
+//
+// This file is licensed under the Apache License v2.0 with LLVM Exceptions.
+// See https://llvm.org/LICENSE.txt for license information.
+// SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+//
+// (c) Copyright 2025 Advanced Micro Devices, Inc. or its affiliates
+//
+//===----------------------------------------------------------------------===//
+// This is a very low level implementation of a bitset, designed to be fully
```
andcarminati: I think this is in a higher level now with `std::bitset`.

llvm/lib/Target/AIE/AIEHazardRecognizer.cpp
```diff
@@ -85,28 +85,28 @@ void FuncUnitWrapper::dump() const {
...
+  Reserved.clear();
   Slots = 0;
   MemoryBanks = 0;
   MemObjectsBits = 0;
 }
 
 bool FuncUnitWrapper::isEmpty() const {
-  return Required == 0 && Reserved == 0 && Slots == 0 && MemoryBanks == 0 &&
-         MemObjectsBits == 0;
+  return Required.empty() && Reserved.empty() && Slots == 0 &&
```
andcarminati, line 103: I expect this `.empty()` to be fast, but not faster than `== 0`. Should we change the order here?

llvm/unittests/CodeGen/ResourceScoreboardTest.cpp
```diff
@@ -0,0 +1,142 @@
+//===- ResourceScoreboardTest.cpp -----------------------------------------===//
+//
+// This file is licensed under the Apache License v2.0 with LLVM Exceptions.
+// See https://llvm.org/LICENSE.txt for license information.
+// SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+//
+// (c) Copyright 2025 Advanced Micro Devices, Inc. or its affiliates
+//
+//===----------------------------------------------------------------------===//
+
```
andcarminati: Perhaps a description of the goal of this test, for the future!

llvm/lib/Target/AIE/AIEHazardRecognizer.cpp
```diff
@@ -623,35 +640,33 @@ void AIEHazardRecognizer::enterResources(
...
-    for (unsigned int C = 0; C < IS.getCycles(); ++C) {
-      Scoreboard[Cycle + C] |= ThisCycle;
-    }
+  Scoreboard[DeltaCycles].IssueCount++;
 
-    // Advance the cycle to the next stage.
-    Cycle += IS.getNextCycles();
-  }
+  // Insert ThisCycle at position Cycle relative to the start of the itinerary.
+  auto Insert = [&Scoreboard, DeltaCycles](int Cycle,
```
F-Stuckmann, line 663: Is this the same as reserving a FuncUnit? Could we harmonize the names?

llvm/lib/Target/AIE/AIEHazardRecognizer.cpp
```diff
@@ -671,14 +686,13 @@ void AIEHazardRecognizer::computeMaxLatency() {
   assert(ItinData && !ItinData->isEmpty());
   unsigned FirstRW = std::numeric_limits<unsigned>().max();
   unsigned LastRW = 0;
-  unsigned ItinDepth = 0;
+  auto MaxDepth = [this](int C, const FuncUnitWrapper &) {
```
F-Stuckmann: we only access PipelineDepth , could we only Capture the reference PipelineDepth ?

llvm/lib/Target/AIE/AIEPostPipeliner.h
```diff
@@ -210,9 +210,8 @@ class PostPipeliner {
   /// The length of the longest circuit in the graph.
   int RecMII = 0;
 
-  // The scoreboard and its depth.
+  // The scoreboard
```
andcarminati: nit: `The scoreboard.`

llvm/test/CodeGen/AIE/aie2/schedule/postpipeliner/add-store.mir
```diff
@@ -40,8 +40,10 @@
   ; CHECK-NEXT:  .L_LEnd0:
   ; CHECK-NEXT:    nopb ; nopa ; st r1, [p0], #4; add r1, r1, #1; nopm ; nopv
   ; CHECK-NEXT:  // %bb.3: // %for.cond.cleanup
-  ; CHECK-NEXT:    nopb ; nopa ; st r1, [p0], #4; nopxm ; nopv
-  ; CHECK-NEXT:    nopx
+  ; CHECK-NEXT:    nopa ; nopb ; nopxm ; st r1, [p0], #4
```
andcarminati, line 43: I guess that 9 cycles here is too much!

llvm/unittests/Target/AIE/HazardRecognizerTest.cpp
```diff
@@ -24,24 +24,48 @@ using namespace llvm;
 const auto Req = InstrStage::ReservationKinds::Required;
 const auto Res = InstrStage::ReservationKinds::Reserved;
 
+// Note: This test has a history, and bears the scratches and tears of
+// a number of hazardrecognizer and scoreboard reorganizations
```
andcarminati: nit: `reorganizations.`

llvm/lib/Target/AIE/AIEHazardRecognizer.cpp
```diff
@@ -137,6 +138,31 @@ bool FuncUnitWrapper::conflict(const FuncUnitWrapper &Other) const {
          !FormatInterface->getPacketFormats().getFormat(Slots | Other.Slots);
 }
 
+namespace {
+// Central itinerary data interpreter
+// It will call Action with the cycle number and the FuncUnits,
+// returning true as soon as Action returns true.
+// A full traversal can be made by returning false.
+bool anyStage(ArrayRef<const InstrStage> Stages,
+              std::function<bool(int C, const FuncUnitWrapper &)> Action,
```
F-Stuckmann: If you define the callable, you can use them in the lambdas to clarify them. 

```
using FuncUnitAction = std::function<bool(int C, const FuncUnitWrapper &)>;
bool anyStage(ArrayRef<const InstrStage> Stages, FuncUnitAction Action,
              std::optional<int> FUDepthLimit = std::nullopt) {
```
Then in the lambdas:
```
  FuncUnitAction Insert = [&Scoreboard, DeltaCycles](int Cycle,
                                           const FuncUnitWrapper &ThisCycle) {
    const int ScoreboardCycle = Cycle + DeltaCycles;
    assert(Scoreboard.isInRange(ScoreboardCycle));
    Scoreboard[ScoreboardCycle] |= ThisCycle;
    return false;
  };
```

llvm/lib/Target/AIE/AIEHazardRecognizer.h
martien-de-jong: As discussed. By building it on top of std::bitset we can delegate to std::bitset testing.

llvm/lib/Target/AIE/StaticBitSet.h
martien-de-jong: true. I'll adjust.

llvm/lib/Target/AIE/AIEHazardRecognizer.cpp
martien-de-jong: I find zero references.... I'll try to remove it.

llvm/lib/Target/AIE/AIEHazardRecognizer.cpp
martien-de-jong: Thanks for trying that!

llvm/unittests/CodeGen/CMakeLists.txt
```diff
@@ -1,3 +1,10 @@
+#
+# This file is licensed under the Apache License v2.0 with LLVM Exceptions.
+# See https://llvm.org/LICENSE.txt for license information.
+# SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+#
+# (c) Copyright 2025 Advanced Micro Devices, Inc. or its affiliates
```
F-Stuckmann: NIT: Copyright should be 
// Modifications (c) Copyright 2025 Advanced Micro Devices, Inc. or its
// affiliates

llvm/lib/Target/AIE/AIEHazardRecognizer.cpp
martien-de-jong: But I can't capture class members....

martien-de-jong:
> > There is a slight regression due to a correction of the pipelinedepth for memory operations, which tends to add nop cycles to exit blocks.
> 
> Hi @martien-de-jong, does it mean that we were optimistic before? I am curious about the QoR effect.

As discussed, we were wrong, if we want to use a 'pure' definition of the pipeline depth. The memory cycles were recorded in the scoreboard, and the allocation of the scoreboard uses the pipelinedepth. Since we didn't take the memory cycles into account for the pipelinedepth, we were checking an unrelated cycle for the resource conflict, which can lead to both false negatives and false positives. The pipelinedepth is also used to define the safety margin when we jump into an unknown block, hence these new extra cycles. The latter use of the pipelinedepth  was safe, since there we implicitly used the fact that all instructions with memory cycles do that late in the pipeline, ~E4.
I think we need two notions of the pipelinedepth, one that defines the range the scoreboard should be able to access. the other to represent the safely margin.

llvm/lib/Target/AIE/AIEHazardRecognizer.cpp
martien-de-jong: Oh, differently unimportant. Only used from dumpers.

llvm/lib/Target/AIE/AIEHazardRecognizer.cpp
```diff
@@ -709,6 +711,7 @@ void AIEHazardRecognizer::computeMaxLatency() {
   FirstRW = TII->getMinFirstMemoryCycle();
   LastRW = TII->getMaxLastMemoryCycle();
   MaxLatency = std::max(MaxLatency, int(LastRW - FirstRW + 1));
+  PipelineDepth = std::max(PipelineDepth, int(LastRW));
```
andcarminati, line 714: Oh, here is the bug fix that was mentioned.

llvm/test/CodeGen/AIE/aie2/schedule/postpipeliner/add-store.mir
andcarminati: Maybe not ;-)

llvm/lib/Target/AIE/AIEHazardRecognizer.cpp
martien-de-jong: Discussed offline. It's inserting a full FuncUnitWrapper into the scoreboard. This represents one cycle of the itinerary, in this case the Resaerved and Required resources.

andcarminati (approved):
LGTM. Nice work!

PR #513, opened 2025-07-01
Update aiebase_chess.h

SagarMaheshwari99:
Fix C preprocessor argument expansion and boolean logic

### DIFF
--- a/clang/lib/Headers/aiebase_chess.h
+++ b/clang/lib/Headers/aiebase_chess.h
@@ -26,7 +26,7 @@
 #define chess_keep_sw_loop2
 #define chess_loop_count(x)
 #define chess_loop_range(...)
-#define chess_manifest(x) (__builtin_constant_p(x) && (x != 0))
+#define chess_manifest(x) (__builtin_constant_p(x) && (!!(x)))
 #define chess_modulo_scheduling_budget_ratio(x)
 #define chess_no_hw_loop
 #define chess_no_warn_pipelining


clang/lib/Headers/aiebase_chess.h
```diff
@@ -26,7 +26,7 @@
 #define chess_keep_sw_loop2
 #define chess_loop_count(x)
 #define chess_loop_range(...)
-#define chess_manifest(x) (__builtin_constant_p(x) && (x != 0))
+#define chess_manifest(x) (__builtin_constant_p(x) && (!!(x)))
```
F-Stuckmann, line 29: would this also work?
#define chess_manifest(x) (__builtin_constant_p(x) && x)?

martien-de-jong: I only see a syntactic difference. I guess `&& (x)` would suffice to get the precedence of operators with the actual 'x' right. && forces its operands to 0/1.

SagarMaheshwari99: I think `&&` doesn't force the operands to 0/1 but rather just evaluates the expression as if it were a bool. 
While `!!(x)` will coerce `x` into a boolean, which sounds more intuitive, than just keeping `x`

martien-de-jong: image
So not forcing to 0/1, but definitely doing the comparison against zero.

martien-de-jong (approved):

clang/lib/Headers/aiebase_chess.h
SagarMaheshwari99: yes, the context is boolean.

PR #514, opened 2025-07-02
[AIEX] legalize unaligned ptr-stores

F-Stuckmann:
Fix crashes when storing unaligned pointers, such as 

`G_STORE %1, %0 :: (store (p0), align 2)`

### DIFF
--- a/llvm/lib/Target/AIE/AIE2LegalizerInfo.cpp
+++ b/llvm/lib/Target/AIE/AIE2LegalizerInfo.cpp
@@ -376,6 +376,10 @@ AIE2LegalizerInfo::AIE2LegalizerInfo(const AIE2Subtarget &ST) : AIEHelper(ST) {
           {ACC512, P0, ACC512, 32}, {ACC1024, P0, ACC1024, 32},
           {S128, P0, S128, 16},
       })
+      // Storing a pointer to an un-aligned address.
+      .customIf([=](const LegalityQuery &Query) {
+        return AIEHelper.isUnaligned20BitStore(Query);
+      })
       .widenScalarToNextPow2(0)
       .lowerIfMemSizeNotPow2()
       .bitcastIf(
@@ -393,7 +397,6 @@ AIE2LegalizerInfo::AIE2LegalizerInfo(const AIE2Subtarget &ST) : AIEHelper(ST) {
       .clampScalar(0, S32, S32)
       .lower();
 
-  // FIXME: Storing a pointer to an un-aligned address isn't supported.
   getActionDefinitionsBuilder({G_ZEXTLOAD, G_SEXTLOAD})
       .legalForTypesWithMemDesc({{S32, P0, S8, 8}, {S32, P0, S16, 16}})
       .widenScalarToNextPow2(0)
@@ -530,6 +533,8 @@ bool AIE2LegalizerInfo::legalizeCustom(
   switch (MI.getOpcode()) {
   default:
     break;
+  case TargetOpcode::G_STORE:
+    return AIEHelper.legalize_G_STORE(Helper, cast<GStore>(MI));
   case TargetOpcode::G_VASTART:
     return AIEHelper.legalizeG_VASTART(Helper, MI);
   case TargetOpcode::G_VAARG:
--- a/llvm/lib/Target/AIE/AIELegalizerHelper.cpp
+++ b/llvm/lib/Target/AIE/AIELegalizerHelper.cpp
@@ -416,6 +416,69 @@ bool AIELegalizerHelper::legalizeG_VASTART(LegalizerHelper &Helper,
   return true;
 }
 
+bool AIELegalizerHelper::isUnaligned20BitStore(
+    const LegalityQuery &Query) const {
+  const LLT &DstTy = Query.Types[0];
+  if (DstTy.getSizeInBits() != 20)
+    return false;
+
+  if (Query.MMODescrs.empty())
+    return false;
+
+  if (Query.Opcode != TargetOpcode::G_STORE)
+    return false;
+
+  if (!DstTy.isPointer() && !DstTy.isScalar())
+    return false;
+
+  const bool Is32BitAligned = Query.MMODescrs.begin()->AlignInBits % 32 == 0;
+  if (Is32BitAligned)
+    return false;
+
+  return true;
+}
+
+bool AIELegalizerHelper::legalize_G_STORE(LegalizerHelper &Helper,
+                                          GStore &StoreI) const {
+  MachineRegisterInfo &MRI = StoreI.getMF()->getRegInfo();
+  Register DstReg = StoreI.getOperand(0).getReg();
+  LLT DestLLT = MRI.getType(DstReg);
+
+  const LLT S20 = LLT::scalar(20);
+
+  GISelObserverWrapper DummyObserver;
+  MachineIRBuilder HelperBuilder(StoreI, DummyObserver);
+  DummyObserver.changedInstr(StoreI);
+
+  // PtrtoInt cast.
+  Register InputReg = DstReg;
+  if (DestLLT.isPointer()) {
+    InputReg = MRI.createGenericVirtualRegister(S20);
+    HelperBuilder.buildPtrToInt(InputReg, DstReg);
+  }
+
+  // EXT new PtrtoInt-Cast.
+  Register ExtReg = MRI.createGenericVirtualRegister(S32);
+  if (DestLLT.isPointer())
+    HelperBuilder.buildZExt(ExtReg, InputReg);
+  else {
+    // FIXME: implement with G_SEXT once instruction legalization works
+    Register TempExtReg = MRI.createGenericVirtualRegister(S32);
+    HelperBuilder.buildZExt(TempExtReg, InputReg);
+    HelperBuilder.buildSExtInReg(ExtReg, TempExtReg, 20);
+  }
+
+  // Replace ptr with EXT Result.
+  StoreI.getOperand(0).setReg(ExtReg);
+
+  // Update MMO Type.
+  MachineMemOperand &MMO = **StoreI.memoperands_begin();
+  MMO.setType(S32);
+
+  Helper.lowerStore(StoreI);
+  return true;
+}
+
 bool AIELegalizerHelper::legalizeG_VAARG(LegalizerHelper &Helper,
                                          MachineInstr &MI) const {
   MachineIRBuilder &MIRBuilder = Helper.MIRBuilder;
--- a/llvm/lib/Target/AIE/AIELegalizerHelper.h
+++ b/llvm/lib/Target/AIE/AIELegalizerHelper.h
@@ -15,6 +15,7 @@
 #define LLVM_LIB_TARGET_AIE_AIELEGALIZERHELPER_H
 
 #include "llvm/CodeGen/GlobalISel/LegalizerHelper.h"
+#include "llvm/CodeGen/GlobalISel/LegalizerInfo.h"
 #include "llvm/IR/InstrTypes.h"
 
 namespace llvm {
@@ -91,6 +92,8 @@ public:
                                         MachineInstr &MI,
                                         const unsigned LegalVectorSize) const;
   bool legalizeG_TRUNC(LegalizerHelper &Helper, MachineInstr &MI) const;
+  bool isUnaligned20BitStore(const LegalityQuery &Query) const;
+  bool legalize_G_STORE(LegalizerHelper &Helper, GStore &StoreI) const;
 };
 
 } // namespace llvm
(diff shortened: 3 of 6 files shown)


llvm/lib/Target/AIE/AIE2LegalizerInfo.cpp
```diff
@@ -377,6 +377,12 @@ AIE2LegalizerInfo::AIE2LegalizerInfo(const AIE2Subtarget &ST) : AIEHelper(ST) {
           {S128, P0, S128, 16},
       })
       .widenScalarToNextPow2(0)
+      // Storing a pointer to an un-aligned address.
+      .customIf([=](const LegalityQuery &Query) {
+        const LLT &DstTy = Query.Types[0];
+        return Query.Opcode == G_STORE && DstTy.isPointer() &&
```
andcarminati: I think that, in general, every unaligned `DstTy.getSizeInBits() == 20` should be custom handled.

andcarminati: How about the alignment?

llvm/lib/Target/AIE/AIELegalizerHelper.cpp
```diff
@@ -416,6 +416,40 @@ bool AIELegalizerHelper::legalizeG_VASTART(LegalizerHelper &Helper,
   return true;
 }
 
+bool AIELegalizerHelper::legalize_G_STORE_PtrPtr(LegalizerHelper &Helper,
+                                                 MachineInstr &StoreI) const {
+  auto DstReg = StoreI.getOperand(0).getReg();
+  MachineRegisterInfo &MRI = StoreI.getMF()->getRegInfo();
+  LLT DestLLT = MRI.getType(DstReg);
+
+  assert(DestLLT.isPointer() && DestLLT.getSizeInBits() == 20);
```
andcarminati: If we have an unaligned `store s20 p0` we will not crash, but we will generate an wrong 24-bit store. So I recommend to handle also this case.

llvm/lib/Target/AIE/AIELegalizerHelper.h
```diff
@@ -91,6 +91,7 @@ class AIELegalizerHelper {
                                         MachineInstr &MI,
                                         const unsigned LegalVectorSize) const;
   bool legalizeG_TRUNC(LegalizerHelper &Helper, MachineInstr &MI) const;
+  bool legalize_G_STORE_PtrPtr(LegalizerHelper &Helper, MachineInstr &MI) const;
```
andcarminati: Maybe legalize `legalize_G_STORE`. In the future, we can discover more cases to be custom-legalized and this should be the unique dispatcher.

llvm/lib/Target/AIE/AIELegalizerHelper.cpp
```diff
@@ -416,6 +416,40 @@ bool AIELegalizerHelper::legalizeG_VASTART(LegalizerHelper &Helper,
   return true;
 }
 
+bool AIELegalizerHelper::legalize_G_STORE_PtrPtr(LegalizerHelper &Helper,
+                                                 MachineInstr &StoreI) const {
+  auto DstReg = StoreI.getOperand(0).getReg();
```
andcarminati: nit: `Register`.

llvm/lib/Target/AIE/AIELegalizerHelper.cpp
```diff
@@ -416,6 +416,40 @@ bool AIELegalizerHelper::legalizeG_VASTART(LegalizerHelper &Helper,
...
+
+  // ZEXT new PtrtoInt-Cast.
+  Register ZExtReg = MRI.createGenericVirtualRegister(S32);
+  HelperBuilder.buildZExt(ZExtReg, PtrToIntReg);
+
+  // Replace ptr with ZEXT Result.
+  StoreI.getOperand(0).setReg(ZExtReg);
+
+  // Update MMO Type.
+  auto &MMO = **StoreI.memoperands_begin();
```
andcarminati: nit: full type is better.

llvm/lib/Target/AIE/AIE2LegalizerInfo.cpp
```diff
@@ -530,6 +535,8 @@ bool AIE2LegalizerInfo::legalizeCustom(
   switch (MI.getOpcode()) {
   default:
     break;
+  case TargetOpcode::G_STORE:
+    return AIEHelper.legalize_G_STORE_PtrPtr(Helper, MI);
```
andcarminati: You can cast MI directly to GLoadStore.

llvm/lib/Target/AIE/AIELegalizerHelper.cpp
```diff
@@ -416,6 +416,40 @@ bool AIELegalizerHelper::legalizeG_VASTART(LegalizerHelper &Helper,
   return true;
 }
 
+bool AIELegalizerHelper::legalize_G_STORE_PtrPtr(LegalizerHelper &Helper,
+                                                 MachineInstr &StoreI) const {
```
andcarminati: nit: `MachineInstr` -> `GLoadStore`. Then you don't need any assert.

llvm/lib/Target/AIE/AIELegalizerHelper.h
F-Stuckmann: should the name clarify that we are not legalizing any G_STORE?

andcarminati: But this is the custom legalization entry point for G_STORE.

llvm/lib/Target/AIE/AIE2LegalizerInfo.cpp
F-Stuckmann: For s20 and pointer types, aligned stores are already handled correctly as word stores in the first legalizer rule (`legalForTypesWithMemDesc`).
In the `customif` rule, we only get the non-aligned cases.

llvm/lib/Target/AIE/AIELegalizerHelper.cpp
F-Stuckmann: good idea

llvm/lib/Target/AIE/AIE2LegalizerInfo.cpp
andcarminati: It is a good idea to make the alignment check explicit to not be completely tied to the rule ordering.

llvm/lib/Target/AIE/AIELegalizerHelper.h
andcarminati: If see the other cases, we always encode `legalize_G_[OPERATION]`, and custom is used (most of times) to legalize a subset of non-legal operations. In this case, we don't need to encode which non-legal operation we are trying to handle.

llvm/lib/Target/AIE/AIELegalizerHelper.cpp
F-Stuckmann: I am now casting to GStore. Otherwise i find the function name and the argument confusing

llvm/lib/Target/AIE/AIELegalizerHelper.cpp
```diff
@@ -416,6 +416,69 @@ bool AIELegalizerHelper::legalizeG_VASTART(LegalizerHelper &Helper,
   return true;
 }
 
+bool AIELegalizerHelper::isUnaligned20BitStore(
+    const LegalityQuery &Query) const {
+  if (Query.MMODescrs.begin() == Query.MMODescrs.end())
```
andcarminati: Better `.empty()`?

llvm/lib/Target/AIE/AIELegalizerHelper.cpp
```diff
@@ -416,6 +416,69 @@ bool AIELegalizerHelper::legalizeG_VASTART(LegalizerHelper &Helper,
...
+    return false;
+
+  if (!DstTy.isPointer() && !DstTy.isScalar())
+    return false;
+
+  const bool Is32BitAligned = Query.MMODescrs.begin()->AlignInBits % 32 == 0;
+  if (Is32BitAligned)
+    return false;
+
+  if (DstTy.getSizeInBits() != 20)
```
andcarminati: This could be the first test.

llvm/lib/Target/AIE/AIELegalizerHelper.cpp
```diff
@@ -416,6 +416,69 @@ bool AIELegalizerHelper::legalizeG_VASTART(LegalizerHelper &Helper,
...
+    return false;
+
+  const LLT &DstTy = Query.Types[0];
+  if (Query.Opcode != TargetOpcode::G_STORE)
+    return false;
+
+  if (!DstTy.isPointer() && !DstTy.isScalar())
+    return false;
+
+  const bool Is32BitAligned = Query.MMODescrs.begin()->AlignInBits % 32 == 0;
```
andcarminati, line 431: We say yes to unaligned 32bit store?

llvm/lib/Target/AIE/AIELegalizerHelper.cpp
```diff
@@ -416,6 +416,69 @@ bool AIELegalizerHelper::legalizeG_VASTART(LegalizerHelper &Helper,
...
+    InputReg = MRI.createGenericVirtualRegister(S20);
+    HelperBuilder.buildPtrToInt(InputReg, DstReg);
+  }
+
+  // EXT new PtrtoInt-Cast.
+  Register ExtReg = MRI.createGenericVirtualRegister(S32);
+  if (DestLLT.isPointer())
+    HelperBuilder.buildZExt(ExtReg, InputReg);
+  else {
+    // FIXME: implement with G_SEXT once instruction selection works
```
andcarminati: `once instruction legalization works`

The selection failure is just the final result of the missing correct legalization.

llvm/lib/Target/AIE/aie1/AIE1LegalizerInfo.cpp
```diff
@@ -221,6 +222,10 @@ AIE1LegalizerInfo::AIE1LegalizerInfo(const AIEBaseSubtarget &ST)
           {ACC512, P0, ACC512, 32}, {ACC1024, P0, ACC1024, 32},
           {S128, P0, S128, 16},
       })
+      // Storing a pointer to an un-aligned address.
+      .customIf([=](const LegalityQuery &Query) {
+        return AIEHelper.isUnaligned20BitStore(Query);
```
andcarminati, line 227: Nice code reuse!

llvm/test/CodeGen/AIE/GlobalISel/legalize-store.mir
```diff
@@ -118,6 +118,87 @@ body: |
...
+...
+
+---
+name: test_store_p0_unaligned
+body: |
+  bb.0:
+    liveins: $p0, $p1
+    ; CHECK-LABEL: name: test_store_p0_unaligned
+    ; CHECK: liveins: $p0, $p1
+    ; CHECK-NEXT: {{  $}}
```
andcarminati, line 144: It looks big, but this is a safeguard for a very rare case!

llvm/lib/Target/AIE/AIELegalizerHelper.cpp
```diff
@@ -416,6 +416,69 @@ bool AIELegalizerHelper::legalizeG_VASTART(LegalizerHelper &Helper,
...
+bool AIELegalizerHelper::isUnaligned20BitStore(
+    const LegalityQuery &Query) const {
+  if (Query.MMODescrs.begin() == Query.MMODescrs.end())
+    return false;
+
+  const LLT &DstTy = Query.Types[0];
+  if (Query.Opcode != TargetOpcode::G_STORE)
+    return false;
+
+  if (!DstTy.isPointer() && !DstTy.isScalar())
```
andcarminati, line 431: If we test the size, before, this test is redundant.

andcarminati:
This PR looks good and correct, I left some small comments, mostly related to the style and small optimizations!

llvm/lib/Target/AIE/AIELegalizerHelper.cpp
F-Stuckmann: that would be clearer

llvm/lib/Target/AIE/AIELegalizerHelper.cpp
F-Stuckmann: we do not worry for weird vectors?

F-Stuckmann: I would rather have weird shaped vectors fail consistently, that only failing if the vector size is different than 20 bits

andcarminati (approved):
LGTM.

PR #516, opened 2025-07-08
Upstream sync with 8f54ed2173cb

isoard-amd:
This also include a fixup for a tablegen change.


isoard-amd:
Note that I chose that commit because it is the first commit from upstream that was clean again.

In particular, 14e6f63 introduced a regression that is fixed by 8f54ed2.

PR #519, opened 2025-07-08
[AIE] Pass -ffunction-sections and -fdata-sections by default

konstantinschwarz:
AIE does not have small addressing modes which are typically pessimized by using these options.
Instead, using separate sections by default allows the linker to garbage collect unused sections,
ensuring we keep the memory usage down to a minimum.


clang/lib/Driver/ToolChains/CommonArgs.cpp
```diff
@@ -731,7 +731,7 @@ llvm::StringRef tools::getLTOParallelism(const ArgList &Args, const Driver &D) {
 
 // PS4/PS5 uses -ffunction-sections and -fdata-sections by default.
```
SagarMaheshwari99, line 732: Nit: comment update.

SagarMaheshwari99 (approved):

PR #526, opened 2025-07-10
[AIEX] Add EarlyIfConversion Suport

andcarminati:
This support covers scalar registers and is intended to help with PM size reduction.

Other register classes could also be supported, but no profitable cases were found to justify.

QoR results:

Core_PMSize
 Gain: -0.88%
Core_Insn_Count
Gain: -0.17%
Core_StackSize
Gain: -1.77%


llvm/lib/Target/AIE/AIE2TargetMachine.cpp
```diff
@@ -131,6 +132,12 @@ void AIE2PassConfig::addISelPrepare() {
   TargetPassConfig::addISelPrepare();
 }
 
+bool AIE2PassConfig::addILPOpts() {
```
F-Stuckmann, line 135: could this go into AIEBaseTargetMachine?

llvm/lib/Target/AIE/AIEBaseInstrInfo.cpp
```diff
@@ -1414,3 +1419,66 @@ bool AIEBaseInstrInfo::isExtendLikelyToBeFolded(
...
+  // Check register classes.
+  const MachineRegisterInfo &MRI = MBB.getParent()->getRegInfo();
+  const TargetRegisterClass *RCTrue = MRI.getRegClass(TrueReg);
+  const TargetRegisterClass *RCFalse = MRI.getRegClass(FalseReg);
+  const TargetRegisterClass *ScalarRegisterClass = Support->ScalarRegisterClass;
+
+  if (!RCTrue || !RCFalse)
+    return false;
+
+  if (ScalarRegisterClass->hasSubClassEq(RCTrue) &&
```
F-Stuckmann, line 1441: wouldn't we always expect that the `RCTrue `and `RCFAlse ` are valid registers? Maybe convert this if to an assert

llvm/lib/Target/AIE/AIEBaseInstrInfo.cpp
```diff
@@ -1414,3 +1419,66 @@ bool AIEBaseInstrInfo::isExtendLikelyToBeFolded(
...
+      ScalarRegisterClass->hasSubClassEq(RCFalse)) {
+    CondCycles = 1;
```
F-Stuckmann, line 1448: Are these Cycles used in a cost evaluation during If Conversion?

llvm/lib/Target/AIE/AIEBaseInstrInfo.cpp
```diff
@@ -1414,3 +1419,66 @@ bool AIEBaseInstrInfo::isExtendLikelyToBeFolded(
...
+                                    ArrayRef<MachineOperand> Cond,
+                                    Register TrueReg, Register FalseReg) const {
+
+  auto Support = getIfConvSupport();
+  assert(Support && "No If-Conversion support for this target\n");
+
+  const TargetRegisterClass *SelectRegisterClass = Support->SelectRegisterClass;
+
+  MachineRegisterInfo &MRI = MBB.getParent()->getRegInfo();
+  auto Op = Cond[0].getImm();
```
F-Stuckmann: nit: const int64_t ImmOp

llvm/lib/Target/AIE/AIEBaseInstrInfo.cpp
```diff
@@ -1414,3 +1419,66 @@ bool AIEBaseInstrInfo::isExtendLikelyToBeFolded(
...
+  Register Reg = Cond[1].getReg();
+
+  const unsigned Opcode = Support->BranchToSelectMap[Op];
+  Register SelReg = MRI.createVirtualRegister(SelectRegisterClass);
+
+  BuildMI(MBB, MI, dl, get(TargetOpcode::COPY), SelReg).addReg(Reg);
```
F-Stuckmann: Check: Are we coping Reg into a new Register class for Register Bank Selection?

llvm/lib/Target/AIE/AIEBaseInstrInfo.cpp
andcarminati: Yes! But, as I observed in practice, the trace metric and critical path length is more important (more sensible).

llvm/lib/Target/AIE/AIEBaseInstrInfo.cpp
andcarminati: This is opcode, always `unsigned`. Perhaps `unsigned` is better than `auto` indeed.

llvm/lib/Target/AIE/AIEBaseInstrInfo.cpp
andcarminati: Actually, we are very ahead of register bank selection, this is the proper class. What we solve with this copy is the problem that SEL always use r27 as a selector, we model this as a singleton class that only contains r27.

llvm/lib/Target/AIE/AIEBaseInstrInfo.cpp
andcarminati: They are valid, but I have no clue if in some situation they could be physical (probably no). For physical there is no more register class.

llvm/lib/Target/AIE/AIE2TargetMachine.cpp
andcarminati: We need to think about this, indeed is not the best, but we duplicate for everything apart of the module pass manager configuration.

llvm/lib/Target/AIE/AIEBaseInstrInfo.cpp
F-Stuckmann: so the if conversion does nothing in that case, right?

llvm/lib/Target/AIE/AIEBaseInstrInfo.cpp
F-Stuckmann: could you add this as a comment into the code?

llvm/lib/Target/AIE/AIEBaseInstrInfo.cpp
F-Stuckmann: could you add this a comment in the code?

andcarminati: Sure! I will update including a comment! This is not intuitive indeed.

llvm/lib/Target/AIE/AIEBaseInstrInfo.cpp
andcarminati: Yes, it can kick-int only if we return true! Even if we return true, there is also no guarantee, this is the first filter that EarlyIfConversion uses to decide when apply.

llvm/lib/Target/AIE/AIEBaseInstrInfo.cpp
andcarminati: Yes, I will add!

llvm/lib/Target/AIE/aie2p/AIE2PInstrInfo.cpp
```diff
@@ -1980,3 +1981,16 @@ AIE2PInstrInfo::parseAbstractOp(const MachineInstr &MI) const {
   }
   return std::nullopt;
 }
+
+std::optional<AIEBaseInstrInfo::IfConvSupport>
+AIE2PInstrInfo::getIfConvSupport() const {
+  AIEBaseInstrInfo::IfConvSupport Result;
+
+  Result.BranchToSelectMap[AIE2P::PseudoJNZ] = AIE2P::SEL_NEZ;
```
F-Stuckmann, line 1989: I wish we had an unified opcode space for the generic (gpr) stuff.

andcarminati: At least they are constrained in TII.

llvm/lib/Target/AIE/AIE2TargetMachine.cpp
andcarminati: Moved to AIE2PassConfig to follow the pattern.

llvm/lib/Target/AIE/AIEBaseInstrInfo.cpp
andcarminati: Done!

llvm/lib/Target/AIE/AIEBaseInstrInfo.cpp
```diff
@@ -1414,3 +1419,72 @@ bool AIEBaseInstrInfo::isExtendLikelyToBeFolded(
...
+  if (ScalarRegisterClass->hasSubClassEq(RCTrue) &&
+      ScalarRegisterClass->hasSubClassEq(RCFalse)) {
+    // Latencies from Cond+Branch, TrueReg, and FalseReg to DstReg.
+    // We force a low Cond+Branch to be permissive as much as
+    // possible, considering that is better to control the behavior
+    // using getCriticalPathLimit instead of an unrealistic miss prediction
+    // penalty (no predictor in AIE).
+    CondCycles = 1;
+    TrueCycles = 1;
+    FalseCycles = 1;
```
martien-de-jong, line 1450: If I understand correctly, these are just the latencies around the select. If either the true or the false operand would e.g. call a runtime div routine, it will be analysed later as too expensive.

llvm/lib/Target/AIE/AIEBaseInstrInfo.cpp
```diff
@@ -1414,3 +1419,72 @@ bool AIEBaseInstrInfo::isExtendLikelyToBeFolded(
...
+
+  // We need to respect the selection instruction register requirements (if any)
+  const TargetRegisterClass *SelectRegisterClass = Support->SelectRegisterClass;
+
+  MachineRegisterInfo &MRI = MBB.getParent()->getRegInfo();
+  unsigned BranchOpcode = Cond[0].getImm();
+  Register ConditionReg = Cond[1].getReg();
+
+  const unsigned SelectOpcode = Support->BranchToSelectMap[BranchOpcode];
+  Register SelReg = MRI.createVirtualRegister(SelectRegisterClass);
```
martien-de-jong: nit: SelReg could be either the result of the Select or the register to select upon. CondReg?

llvm/lib/Target/AIE/AIEBaseInstrInfo.cpp
```diff
@@ -1414,3 +1419,72 @@ bool AIEBaseInstrInfo::isExtendLikelyToBeFolded(
...
+
+  BuildMI(MBB, MI, DL, get(TargetOpcode::COPY), SelReg).addReg(ConditionReg);
+
+  BuildMI(MBB, MI, DL, get(SelectOpcode), DestReg)
+      .addReg(TrueReg)
+      .addReg(FalseReg)
+      .addReg(SelReg);
```
martien-de-jong: Should we put this order in the IfConvSupport structure? (Personnaly, I find it strange to have the condition last)

llvm/lib/Target/AIE/AIEBaseInstrInfo.cpp
```diff
@@ -1414,3 +1419,72 @@ bool AIEBaseInstrInfo::isExtendLikelyToBeFolded(
...
+  assert(Support && "No If-Conversion support for this target\n");
+
+  // We need to respect the selection instruction register requirements (if any)
+  const TargetRegisterClass *SelectRegisterClass = Support->SelectRegisterClass;
+
+  MachineRegisterInfo &MRI = MBB.getParent()->getRegInfo();
+  unsigned BranchOpcode = Cond[0].getImm();
+  Register ConditionReg = Cond[1].getReg();
+
+  const unsigned SelectOpcode = Support->BranchToSelectMap[BranchOpcode];
```
martien-de-jong: I wouldn't mind hiding this in a method BranchToSelectOpcode(unsigned BranchOpcode)

llvm/lib/CodeGen/EarlyIfConversion.cpp
```diff
@@ -913,7 +916,8 @@ bool EarlyIfConverter::shouldConvertIf() {
                               FBBTrace.getCriticalPath());
 
   // Set a somewhat arbitrary limit on the critical path extension we accept.
-  unsigned CritLimit = SchedModel.MispredictPenalty/2;
+  unsigned CritLimit = std::max((int)SchedModel.MispredictPenalty / 2,
+                                (int)TII->getCriticalPathLimit());
```
martien-de-jong: I think it would be clearer without the max. I would shove the MispredictPenalty in the default implementation of getCriticalPathLimit()

llvm/lib/Target/AIE/AIEBaseInstrInfo.cpp
andcarminati: Yes. And in case of calls, the optimization will give up because calls cannot be speculatively executed.

llvm/lib/CodeGen/EarlyIfConversion.cpp
martien-de-jong: As discussed offline, we don't want to decide on critical path limit at all. If we can make all decisions in that first selection round, it would be nicer.

llvm/lib/Target/AIE/AIEBaseInstrInfo.h
```diff
@@ -77,6 +78,48 @@ struct AIEBaseInstrInfo : public TargetInstrInfo {
...
+      FalseReg,
+    };
+
+    // Ordering information about the input operands.
+    OperandIndex InputOperandsOrder[3];
+
+    // Register operand order to correctly build the instruction
+    // in a aie-independent way.
+    void registerOperandIndex(OperandIndex Op, unsigned Index) {
+      InputOperandsOrder[Index] = Op;
```
andcarminati: FYI @martien-de-jong: the map was reversed, so we don't need the zip anymore!

llvm/lib/CodeGen/EarlyIfConversion.cpp
```diff
@@ -912,10 +915,11 @@ bool EarlyIfConverter::shouldConvertIf() {
   unsigned MinCrit = std::min(TBBTrace.getCriticalPath(),
                               FBBTrace.getCriticalPath());
 
+  MachineBasicBlock &MBB = *IfConv.Head;
+  const TargetSubtargetInfo &STI = MBB.getParent()->getSubtarget();
   // Set a somewhat arbitrary limit on the critical path extension we accept.
-  unsigned CritLimit = SchedModel.MispredictPenalty/2;
+  unsigned CritLimit = STI.getCriticalPathLimit();
```
F-Stuckmann, line 921: nice use of Hooks

llvm/lib/Target/AIE/AIEBaseSubtarget.cpp
```diff
@@ -66,6 +66,11 @@ static cl::opt<unsigned>
     UserLatencyMargin("aie-latency-margin", cl::Hidden, cl::init(0),
                       cl::desc("Define the latency on ExitSU edges"));
 
+static cl::opt<unsigned> IfConversionCritPathLimit(
+    "aie-if-conv-critical-path-limit",
+    cl::desc("Specify the critical path extension we accept for if conversion"),
+    cl::init(10), cl::Hidden);
```
F-Stuckmann, line 72: CHECK: If we increase the path limit, we increase the aggressiveness of If Conversion ?

llvm/test/CodeGen/AIE/opt/early-if-conversion-aie2p.mir
```diff
@@ -0,0 +1,819 @@
...
+                        '$r14', '$r15', '$p6', '$p7' ]
+body:             |
+  ; CHECK-LABEL: name: test2
+  ; CHECK: bb.0.entry:
+  ; CHECK-NEXT:   liveins: $p0, $r0, $r1
+  ; CHECK-NEXT: {{  $}}
+  ; CHECK-NEXT:   [[COPY:%[0-9]+]]:ep = COPY $p0
+  ; CHECK-NEXT:   [[COPY1:%[0-9]+]]:er = COPY $r0
+  ; CHECK-NEXT:   [[COPY2:%[0-9]+]]:er = COPY $r1
+  ; CHECK-NEXT:   [[MOV_RLC_imm11_pseudo:%[0-9]+]]:er = MOV_RLC_imm11_pseudo 16
```
F-Stuckmann, line 264: This reference code does not seem to contain Architecture specific Instructions. Could you merge the unit-tests for aie2 and aie2p? I assume they are the same.

llvm/test/CodeGen/AIE/opt/early-if-conversion-aie2p.mir
```diff
@@ -0,0 +1,819 @@
+# NOTE: Assertions have been autogenerated by utils/update_mir_test_checks.py
+#
+# This file is licensed under the Apache License v2.0 with LLVM Exceptions.
+# See https://llvm.org/LICENSE.txt for license information.
+# SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+#
+# (c) Copyright 2025 Advanced Micro Devices, Inc. or its affiliates
+# RUN: llc -mtriple aie2p -run-pass=early-ifcvt -aie-if-conv-critical-path-limit=5 %s -verify-machineinstrs -o - | FileCheck %s
```
F-Stuckmann, line 8: why is the `aie-if-conv-critical-path-limit` different than in AIE2TargetMachine?

F-Stuckmann: Do you have a unit-test that clarifies, why 10 is more profitable than the 5 used for the unit-tests?

llvm/lib/Target/AIE/AIEBaseInstrInfo.h
```diff
@@ -77,6 +78,48 @@ struct AIEBaseInstrInfo : public TargetInstrInfo {
...
+    // Scalar register class that can be used through select
+    // instructions i.e. values to be choosen depending on the
+    // condition.
+    const TargetRegisterClass *ScalarRegisterClass;
+    // Scalar register class that can be used through select
+    // instructions as a result of a comparison.
+    const TargetRegisterClass *SelectRegisterClass;
+
+    // Here we define the operand order for inputs.
+    enum OperandIndex {
```
F-Stuckmann, line 95: I think this can be private

llvm/lib/Target/AIE/aie2p/AIE2PInstrInfo.cpp
```diff
@@ -1980,3 +1981,20 @@ AIE2PInstrInfo::parseAbstractOp(const MachineInstr &MI) const {
...
+std::optional<IfConvSupport> AIE2PInstrInfo::getIfConvSupport() const {
+  IfConvSupport Result;
+
+  Result.BranchToSelectMap[AIE2P::PseudoJNZ] = AIE2P::SEL_NEZ;
+  Result.BranchToSelectMap[AIE2P::PseudoJZ] = AIE2P::SEL_EQZ;
+
+  Result.ScalarRegisterClass = &AIE2P::eRRegClass;
+  Result.SelectRegisterClass = &AIE2P::mR27_selectRegClass;
+
+  Result.registerOperandIndex(IfConvSupport::TrueReg, 0);
```
F-Stuckmann, line 1995: Can these indices be Constants provided from IfConvSupport?

llvm/lib/Target/AIE/AIEBaseSubtarget.cpp
andcarminati: Yes. This was tuned experimentally.

llvm/test/CodeGen/AIE/opt/early-if-conversion-aie2p.mir
andcarminati: Indeed it contains. Most of the instructions use different mnemonic names.

llvm/test/CodeGen/AIE/opt/early-if-conversion-aie2p.mir
andcarminati: This was tunned experimentally. I used 5 just to test that the implementation can filter off some big paths. By having a big test with a default value of 10 that prevents the optimization will not add a better coverage.

andcarminati: > why is the `aie-if-conv-critical-path-limit` different than in AIE2TargetMachine?

Sorry, I didn't get this point.

llvm/lib/Target/AIE/AIEBaseInstrInfo.h
andcarminati: But we use this outside...

llvm/lib/Target/AIE/aie2p/AIE2PInstrInfo.cpp
andcarminati: They just express order, I feel that numbers can express better than constants here.

F-Stuckmann: Could you add a comment, that the indices here are determined by the `SelectOpcode` ?

llvm/lib/Target/AIE/AIEBaseInstrInfo.h
```diff
@@ -77,6 +78,48 @@ struct AIEBaseInstrInfo : public TargetInstrInfo {
...
+      ConditionReg = 0,
+      TrueReg,
+      FalseReg,
+    };
+
+    // Ordering information about the input operands.
+    OperandIndex InputOperandsOrder[3];
```
F-Stuckmann: nit: This is a Mapping between AIE Select Operands and the generic Select Operand Ordering. 
Maybe clarify the name: MapAIESelectOperandsToGenericSelect

llvm/lib/Target/AIE/AIEBaseInstrInfo.h
```diff
@@ -77,6 +78,48 @@ struct AIEBaseInstrInfo : public TargetInstrInfo {
...
+    // Register operand order to correctly build the instruction
+    // in a aie-independent way.
+    void registerOperandIndex(OperandIndex Op, unsigned Index) {
+      InputOperandsOrder[Index] = Op;
+    }
+
+    // Retrieve a specific register operand that should be encoded
+    // at the Index-th position.
+    // Register array should follow the order: {CondReg, TrueReg, FalseReg}.
+    Register getOperand(ArrayRef<Register> Registers, unsigned Index) {
```
F-Stuckmann: Assert that Registers only has 3 elements. Or as many as OperandIndex is big

llvm/lib/Target/AIE/AIEBaseInstrInfo.h
F-Stuckmann: Yeah, this mapping is confusing the hell out of me

llvm/lib/Target/AIE/AIEBaseInstrInfo.h
F-Stuckmann: I would rename this to `getSelectOperand`.

llvm/lib/Target/AIE/AIEBaseInstrInfo.h
F-Stuckmann: Also good: AIESelectOperandIndices. I don't think you need the method `registerOperandIndex`. You could directly initialize the Array in `AIE2InstrInfo::getIfConvSupport()` , i.e. `InputOperandOrder={IfConvSupport::TrueReg, IfConvSupport::FalseReg, IfConvSupport::ConditionReg};`

llvm/lib/Target/AIE/AIEBaseInstrInfo.h
F-Stuckmann: add a comment to clarify where these indices come from.

llvm/lib/Target/AIE/AIEBaseInstrInfo.h
```diff
@@ -77,6 +78,48 @@ struct AIEBaseInstrInfo : public TargetInstrInfo {
...
+
+    // Register operand order to correctly build the instruction
+    // in a aie-independent way.
+    void registerOperandIndex(OperandIndex Op, unsigned Index) {
+      InputOperandsOrder[Index] = Op;
+    }
+
+    // Retrieve a specific register operand that should be encoded
+    // at the Index-th position.
+    // Register array should follow the order: {CondReg, TrueReg, FalseReg}.
```
F-Stuckmann, line 115: nit: Index begs to be called SelectOpIdx.

llvm/lib/Target/AIE/AIEBaseInstrInfo.h
andcarminati: I clarified the comments and changed the identifier name. Also, InputOperandsOrder is a pre-allocated array, I cannot assign directly and I would avoid any dynamic memory allocation here.

F-Stuckmann (approved):

PR #539, opened 2025-07-14
[AIEX] Combine invariant loads

andcarminati:
This optimize some loads left by memcpy expansion.


llvm/lib/Target/AIE/Utils/AIEIRUtils.cpp
```diff
@@ -0,0 +1,110 @@
...
+  const llvm::Type *Ty = CD->getType();
+
+  if (!Ty->getArrayElementType()->isIntegerTy() ||
+      Ty->getArrayElementType()->getScalarSizeInBits() != WordSize * 8) {
+    Result.Fail = true;
+    return Result;
+  }
+
+  const unsigned NumElements = CD->getType()->getArrayNumElements();
+  for (unsigned I = 0; I < NumElements; I++) {
```
andcarminati: No, we don't need this for.

llvm/lib/Target/AIE/AIECombinerHelper.cpp
```diff
@@ -3441,3 +3442,65 @@ bool llvm::matchNarrowZext(MachineInstr &MI, MachineRegisterInfo &MRI,
...
+    return false;
+
+  if (MRI.getType(DataReg) != S32) {
+    LLVM_DEBUG(dbgs() << "Non-optimizable invariant load: " << MI);
+    return false;
+  }
+
+  unsigned Offset = 0;
+  const MachineInstr *DefPtrReg = MRI.getVRegDef(PtrReg);
+  if (DefPtrReg->getOpcode() == TargetOpcode::G_PTR_ADD) {
```
F-Stuckmann, line 3508: Should we follow the pointer update-chain recursively ?

llvm/lib/Target/AIE/AIECombinerHelper.cpp
```diff
@@ -3441,3 +3442,65 @@ bool llvm::matchNarrowZext(MachineInstr &MI, MachineRegisterInfo &MRI,
...
+      return false;
+
+    Offset = Cst->Value.getZExtValue();
+    DefPtrReg = MRI.getVRegDef(DefPtrReg->getOperand(1).getReg());
+  }
+
+  if (DefPtrReg->getOpcode() != TargetOpcode::G_GLOBAL_VALUE)
+    return false;
+
+  auto *GV = DefPtrReg->getOperand(1).getGlobal();
```
F-Stuckmann: nit: expand auto

llvm/lib/Target/AIE/Utils/AIEIRUtils.cpp
```diff
@@ -0,0 +1,110 @@
...
+                                                 unsigned Offset,
+                                                 unsigned WordSize) {
+
+  const Constant *Const = dyn_cast<Constant>(Aggregate);
+
+  if ((Offset % WordSize) != 0)
+    return nullptr; // Unaligned.
+
+  unsigned Index = Offset / WordSize;
+  if (!Const)
```
F-Stuckmann: nit: move this check up

llvm/lib/Target/AIE/AIECombinerHelper.cpp
```diff
@@ -3441,3 +3442,65 @@ bool llvm::matchNarrowZext(MachineInstr &MI, MachineRegisterInfo &MRI,
...
+  const Constant *ConstPtr = dyn_cast<const Constant>(GV);
+  if (!ConstPtr)
+    return false;
+
+  const Constant *C = AIEIRUtils::extractWordFromAggregateConstant(
+      ConstPtr->getOperand(0), Offset, 4);
```
F-Stuckmann: nit: add comment what 4 stands for

llvm/lib/Target/AIE/AIECombinerHelper.cpp
```diff
@@ -3441,3 +3442,65 @@ bool llvm::matchNarrowZext(MachineInstr &MI, MachineRegisterInfo &MRI,
...
+    LLVM_DEBUG(dbgs() << "Non-optimizable invariant load using: " << ConstPtr);
+    return false;
+  }
+
+  const ConstantInt *CI = dyn_cast<ConstantInt>(C);
+
+  if (!CI)
+    return false;
+
+  const int32_t V = CI->getSExtValue();
```
F-Stuckmann: This is an implicit cast. Can we lose information?

andcarminati: As we are looking to s32 values, we can't loose.

llvm/lib/Target/AIE/AIECombinerHelper.cpp
andcarminati: As we are before the legalization, there is no pointer chains when loading constants.

llvm/lib/Target/AIE/AIECombinerHelper.cpp
```diff
@@ -3441,3 +3442,66 @@ bool llvm::matchNarrowZext(MachineInstr &MI, MachineRegisterInfo &MRI,
...
+
+  assert(MI.getOpcode() == TargetOpcode::G_LOAD && "Not G_LOAD");
+
+  Register DataReg = MI.getOperand(0).getReg();
+  Register PtrReg = MI.getOperand(1).getReg();
+  const LLT S32 = LLT::scalar(32);
+
+  MachineMemOperand *MMO = *MI.memoperands_begin();
+  if (!MMO || !MMO->isInvariant())
+    return false;
```
F-Stuckmann, line 3497: provide a nice debug message, why matching failed

llvm/lib/Target/AIE/AIECombinerHelper.cpp
F-Stuckmann: could we handle variable sized words?

llvm/lib/Target/AIE/Utils/AIEIRUtils.cpp
```diff
@@ -0,0 +1,122 @@
...
+    }
+  }
+  return Result;
+}
+
+namespace llvm::AIEIRUtils {
+// Here, we parse IR all constants that are aggregate constants.
+const Constant *extractWordFromAggregateConstant(const Value *Aggregate,
+                                                 unsigned Offset,
+                                                 unsigned WordSize) {
```
F-Stuckmann: nit: change argument name indicate that the you are looking bytes here. Maybe you could directly pass the number of bits, so it is more generic.

llvm/lib/Target/AIE/AIECombinerHelper.cpp
andcarminati: In theory yes, but due to alignments and DL the parsing could be much harder. We catch all exposed cases with the current approach.

llvm/lib/Target/AIE/Utils/AIEIRUtils.cpp
```diff
@@ -0,0 +1,122 @@
...
+
+  if (!Ty->getArrayElementType()->isIntegerTy() ||
+      Ty->getArrayElementType()->getScalarSizeInBits() != WordSize * 8) {
+    Result.Fail = true;
+    return Result;
+  }
+
+  const unsigned NumElements = CD->getType()->getArrayNumElements();
+  if (Index >= NumElements) {
+    // Not fount yet!
```
martien-de-jong: nit: found

llvm/lib/Target/AIE/Utils/AIEIRUtils.cpp
```diff
@@ -0,0 +1,122 @@
...
+#include "llvm/IR/Constants.h"
+#include "llvm/Support/Casting.h"
+#include "llvm/Support/Debug.h"
+
+#define DEBUG_TYPE "aieirutils"
+
+struct ExtractPartialResult {
+  const llvm::Constant *Result = nullptr;
+  bool Fail = false;
+};
```
martien-de-jong: optional<const llvm::Constant *> perhaps?

llvm/lib/Target/AIE/Utils/AIEIRUtils.h
```diff
@@ -0,0 +1,30 @@
...
+#define LLVM_LIB_TARGET_AIE_UTILS_AIEIRUTILS_H
+
+namespace llvm {
+class Constant;
+class Value;
+} // namespace llvm
+
+namespace llvm::AIEIRUtils {
+
+const Constant *extractBytesFromAggregateConstant(const Value *Aggregate,
```
martien-de-jong: description would be useful.

llvm/lib/Target/AIE/Utils/AIEIRUtils.cpp
```diff
@@ -0,0 +1,122 @@
...
+      Result.Fail = true;
+      return Result;
+    }
+  }
+  return Result;
+}
+
+namespace llvm::AIEIRUtils {
+// Here, we parse IR all constants that are aggregate constants.
+const Constant *extractBytesFromAggregateConstant(const Value *Aggregate,
```
martien-de-jong: I don't understand the 'Bytes' in the name

llvm/lib/Target/AIE/Utils/AIEIRUtils.cpp
```diff
@@ -0,0 +1,122 @@
...
+                                                  unsigned WordSize) {
+
+  const Constant *Const = dyn_cast<Constant>(Aggregate);
+  if (!Const)
+    return nullptr;
+
+  if ((Offset % WordSize) != 0)
+    return nullptr; // Unaligned.
+
+  unsigned Index = Offset / WordSize;
```
F-Stuckmann: Does this not assume, that the Elements in `Aggregate` that the same size? 
I do not think the Index is properly calculated for variable size Constants, i.e. `ConstantStruct`

llvm/lib/Target/AIE/Utils/AIEIRUtils.cpp
```diff
@@ -0,0 +1,122 @@
...
+      if (Int->getType()->getScalarSizeInBits() != WordSize * 8) {
+        // Non-uniform struct (mixed types).
+        Result.Fail = true;
+        return Result;
+      }
+      if (ReverseIndex == 0) {
+        Result.Result = Int;
+        return Result;
+      }
+      ReverseIndex--;
```
F-Stuckmann: I do not think we need to count down `ReverseIndex` since we skip if Data is not aligned.

andcarminati: We need. If this is element 1,  for example, next one will be the correct one.

llvm/lib/Target/AIE/Utils/AIEIRUtils.cpp
andcarminati: > Does this not assume, that the Elements in Aggregate that the same size?

Yes we assume.

llvm/lib/Target/AIE/Utils/AIEIRUtils.cpp
```diff
@@ -0,0 +1,122 @@
...
+  ExtractPartialResult Result;
+  for (auto &EI : Struct->operands()) {
+    auto *ElemConst = llvm::dyn_cast<llvm::Constant>(EI);
+    if (!ElemConst) { // Could this happen? Safe guard.
+      Result.Fail = true;
+      return Result;
+    }
+
+    if (auto Int = llvm::dyn_cast<llvm::ConstantInt>(&EI)) {
+      if (Int->getType()->getScalarSizeInBits() != WordSize * 8) {
```
andcarminati: We bail out here @F-Stuckmann for different types.

llvm/lib/Target/AIE/Utils/AIEIRUtils.cpp
```diff
@@ -0,0 +1,122 @@
...
+// constant [8 x i32] [i32 1053459, i32 0, i32
+// 320012288, i32 4883, i32 0, i32 0, i32 993394688, i32 0], align 4
+static ExtractPartialResult
+extractWordFromConstantData(const llvm::ConstantData *CD, unsigned &Index,
+                            unsigned WordSize) {
+  ExtractPartialResult Result;
+  const llvm::Type *Ty = CD->getType();
+
+  if (!Ty->getArrayElementType()->isIntegerTy() ||
+      Ty->getArrayElementType()->getScalarSizeInBits() != WordSize * 8) {
```
andcarminati: We also bail out here @F-Stuckmann for different types.

llvm/lib/Target/AIE/AIECombinerHelper.cpp
```diff
@@ -3470,6 +3471,76 @@ bool llvm::matchCombineExtAndTrunc(MachineInstr &MI, MachineRegisterInfo &MRI,
...
     };
+  return true;
+}
+
+bool llvm::matchConstLoad(MachineInstr &MI, MachineRegisterInfo &MRI,
+                          GISelChangeObserver &Observer, BuildFnTy &MatchInfo) {
+
+  assert(MI.getOpcode() == TargetOpcode::G_LOAD && "Not G_LOAD");
+
+  Register DataReg = MI.getOperand(0).getReg();
```
F-Stuckmann: I think we can cast `MI` to `GAnyLoad` . That way this also works for `G_ZEXTLOAD` and `G_SEXTLOAD`

llvm/lib/Target/AIE/AIECombinerHelper.cpp
```diff
@@ -3470,6 +3471,76 @@ bool llvm::matchCombineExtAndTrunc(MachineInstr &MI, MachineRegisterInfo &MRI,
...
+  Register PtrReg = MI.getOperand(1).getReg();
+
+  const MachineMemOperand *MMO = *MI.memoperands_begin();
+  if (!MMO || !MMO->isInvariant()) {
+    LLVM_DEBUG(dbgs() << "Non-invariant load: " << MI);
+    return false;
+  }
+
+  const LLT DataRegType = MRI.getType(DataReg);
+  if (!DataRegType.isScalar() || DataRegType.getSizeInBits() > 32) {
```
F-Stuckmann, line 3501: CHECK: The size constraint comes form `MOVXM`, right?

andcarminati: Kind of yes. We could go for 64 bit as maximum, but we will generate two `MOVXM` instead of two loads. This case was not observed as well.

llvm/lib/Target/AIE/AIECombinerHelper.cpp
andcarminati: This is dangerous because we need to generate a constant + a sign extension. It may not work as is.

llvm/lib/Target/AIE/Utils/AIEIRUtils.cpp
andcarminati: Indeed, not anymore!

llvm/lib/Target/AIE/AIECombinerHelper.cpp
```diff
@@ -3470,6 +3471,76 @@ bool llvm::matchCombineExtAndTrunc(MachineInstr &MI, MachineRegisterInfo &MRI,
...
+    return false;
+
+  const Constant *Const = dyn_cast<Constant>(ConstPtr->getOperand(0));
+  if (!Const)
+    return false;
+
+  Type *IntTy =
+      Type::getIntNTy(Const->getContext(), DataRegType.getSizeInBits());
+  const Constant *C = llvm::ConstantFoldLoadFromConst(
+      const_cast<Constant *>(Const), IntTy, APInt(20, Offset),
```
F-Stuckmann: why do we need the const_cast?
Could Overflow of `Offset` occur?

llvm/lib/Target/AIE/AIECombinerHelper.cpp
```diff
@@ -3470,6 +3471,76 @@ bool llvm::matchCombineExtAndTrunc(MachineInstr &MI, MachineRegisterInfo &MRI,
...
+      GV->getParent()->getDataLayout());
+
+  if (!C) {
+    LLVM_DEBUG(dbgs() << "Non-optimizable invariant load using: " << ConstPtr);
+    return false;
+  }
+
+  const ConstantInt *CI = dyn_cast<ConstantInt>(C);
+
```
F-Stuckmann: nit: You can delete the new line between dyn_cast and the if check

llvm/lib/Target/AIE/AIECombinerHelper.cpp
andcarminati: Not by ourselves. We take a s20 constant and generate other.

llvm/lib/Target/AIE/Utils/AIEIRUtils.cpp
andcarminati: Removed.

andcarminati:
QoR Results:
Core_Insn_Count
Avg. gain:-0.09%
Core_PMSize
Avg. gain: -0.54%

For `ConcatAdf_bf16_0` we actually reduce the number of loads, but probably end with a corner case.

llvm/lib/Target/AIE/AIECombinerHelper.cpp
```diff
@@ -3470,6 +3475,78 @@ bool llvm::matchCombineExtAndTrunc(MachineInstr &MI, MachineRegisterInfo &MRI,
...
+bool llvm::matchConstLoad(MachineInstr &MI, MachineRegisterInfo &MRI,
+                          GISelChangeObserver &Observer, BuildFnTy &MatchInfo) {
+
+  if (!FoldInvariantLoads)
+    return false;
+
+  assert(MI.getOpcode() == TargetOpcode::G_LOAD && "Not G_LOAD");
+
+  Register DataReg = MI.getOperand(0).getReg();
```
krishnamtibrewala: nit: const Register

llvm/lib/Target/AIE/AIECombinerHelper.cpp
```diff
@@ -3470,6 +3475,78 @@ bool llvm::matchCombineExtAndTrunc(MachineInstr &MI, MachineRegisterInfo &MRI,
     MatchInfo = [=](MachineIRBuilder &B) {
       B.buildTrunc(TruncDstReg, ExtSrcReg);
     };
+  return true;
+}
+
+bool llvm::matchConstLoad(MachineInstr &MI, MachineRegisterInfo &MRI,
+                          GISelChangeObserver &Observer, BuildFnTy &MatchInfo) {
```
krishnamtibrewala, line 3484: nit : Observer passes but not used.

llvm/test/CodeGen/AIE/GlobalISel/combine-invariant-loads.mir
```diff
@@ -0,0 +1,373 @@
...
+    %10:_(s20) = G_CONSTANT i20 1
+    %11:_(p0) = G_PTR_ADD %2, %10(s20)
+    %12:_(s8) = G_LOAD %11(p0) :: (dereferenceable invariant load (s8) from @constinit4 + 1)
+    %13:_(s20) = G_CONSTANT i20 2
+    %14:_(p0) = G_PTR_ADD %2, %13(s20)
+    %15:_(s8) = G_LOAD %14(p0) :: (dereferenceable invariant load (s8) from @constinit4 + 2)
+    %16:_(s20) = G_CONSTANT i20 4
+    %17:_(p0) = G_PTR_ADD %2, %16(s20)
+    %18:_(s20) = G_LOAD %17(p0) :: (dereferenceable invariant load (s20) from @constinit4 + 4)
+    PseudoRET implicit $lr, implicit %9, implicit %12, implicit %15, implicit %18
```
krishnamtibrewala, line 737: Nice optimization, quite interesting LLVM did not have this already.

Super super Nit : can we have CHECK labels with "--aie-try-fold-loads=false" ?

llvm/lib/Target/AIE/AIECombinerHelper.cpp
andcarminati: Good catch @krishnamtibrewala , I will include in the next push! Thank you!

llvm/test/CodeGen/AIE/GlobalISel/combine-invariant-loads.mir
andcarminati: This is usually a left-over from the memcpy intrinsic. If we have a scalar load, the middle end can optimize in some cases, but memcpy that are pushed to the backend cannot be optimized by default.

F-Stuckmann (approved):
nice reuse of llvm methods

PR #549, opened 2025-07-17
Isoard.upstream sync

isoard-amd:


isoard-amd:
Resolved merge conflict in CodeGenTarget.cpp as upstream now extract enum name from llvm/CodeGen/ValueTypes.td automatically

PR #552, opened 2025-07-17
Isoard.upstream sync

isoard-amd:
Needed to +1 the AIE types indices


clang/test/CodeGen/aie/aie2p/aie2p-abi-bfp16.cpp
```diff
@@ -15,10 +15,6 @@
 #include <stdint.h>
 
 extern "C" {
-// CHECK:   %struct.v64bfp16ebs16 = type <{ <64 x i8>, <8 x i8> }>
```
konstantinschwarz: Looks like these lines were added manually. Can you preserve them?

isoard-amd: Ah, I will update it

PR #554, opened 2025-07-18
[AIE2][AIE2P][NFC] remove unused unknown slot

mludevid:
Small PR to remove the unused `unknown_slot` in AIE2 and AIE2P, which is a relic from AIE1.


khallouh (approved):
LGTM

PR #548, opened 2025-07-16
Legalizer support for vectors in G_FPEXT

Sameeranjoshi:
`G_FPEXT` crashed for vector types when converting `<32xbfloat> -> <32xfloat>` as vectors were not handled in legalizer. This patch adds support for vectors with a custom legalizer rule using following pattern 

```
   VDst = G_FPEXT VSrc
```
   converts to something like below
```
   ZeroVec = G_AIE_BROADCAST_VECTOR VSrc
   VShuffleLow = G_AIE_SHUFFLE_VECTOR ZeroVec, VSrc, 2
   VShuffleHigh = G_AIE_SHUFFLE_VECTOR ZeroVec, VSrc, 3
   VShuffleLow = G_BITCAST VShuffleLow
   VShuffleHigh = G_BITCAST VShuffleHigh
   VDst = G_CONCAT_VECTORS VShuffleLow, VShuffleHigh
```

PR #557, opened 2025-07-19
Handle Vector types in G_FADD using G_FPEXT

Sameeranjoshi:
Depends on #548


llvm/test/CodeGen/AIE/aie2p/GlobalIsel/legalize-vector-fadd.ll
```diff
@@ -0,0 +1,56 @@
+; RUN: llc -mtriple=aie2p -O0 -stop-after=legalizer %s -o - 2>&1 | FileCheck %s
```
andcarminati, line 1: For the tests, it is better to run just the pass of interest. For example, you can create an MIR test including just the ilegal type operation and run llc with `-run-pass=legalizer`. In this way we can easily spot the specific legalization change in action.

andcarminati: The good part is that you can use `llvm/utils/update_mir_test_checks.py` to update the tests.

Sameeranjoshi:
Squashed into https://github.com/Xilinx/llvm-aie/pull/604

PR #564, opened 2025-07-23
[NFC][AIEX] Share immediate operand class definitions

mludevid:
NFC: I moved some immediate operand class definitions from an AIE2P target specific file to an AIE target-agnostic file, to be shared with other AIE targets.


andcarminati (approved):
LGTM.

khallouh (approved):

PR #565, opened 2025-07-23
[AIE2P] Add missing copyPhysReg for exponent registers

andcarminati:


llvm/lib/Target/AIE/aie2p/AIE2PInstrInfo.cpp
```diff
@@ -765,6 +765,38 @@ void AIE2PInstrInfo::copyPhysReg(MachineBasicBlock &MBB,
   } else if ((AIE2P::ePSRFLdFRegClass.contains(SrcReg)) &&
              (AIE2P::ePSRFLdFRegClass.contains(DstReg))) {
     copyThroughSubRegs(MBB, MBBI, DL, DstReg, SrcReg, KillSrc);
+  } else if ((AIE2P::mEhmRegClass.contains(SrcReg)) &&
```
F-Stuckmann, line 768: I know, this is probably the wrong PR, but I really don't like this chain of Ifs.

andcarminati: We really don't need to create a chain of nested ifs here disrupting the existing convention. This is a very simple piece of code that is also very easy to test, just intended to copy between different classes, there is no need to overengineering. I had this around that could be useful for the future.

F-Stuckmann (approved):

PR #567, opened 2025-07-23
Remove bundle alignment through assembler relaxation

konstantinschwarz:
AIE2/AIE2p already use a dedicated AIEMachineAlignment pass to compute the correct alignment for all special AIE alignment requirements.
This change removes the return address alignment handling through assembler relaxation that was only used by AIE1.
Note that this is a breaking change for correct AIE1 code generation, and the CodeGen/AIE/aie1/BinaryOutput/JAL.mir test has been marked XFAIL.

### DIFF
--- a/llvm/lib/Target/AIE/MCTargetDesc/aie2p/AIE2PAsmBackend.cpp
+++ b/llvm/lib/Target/AIE/MCTargetDesc/aie2p/AIE2PAsmBackend.cpp
@@ -10,39 +10,8 @@
 
 #include "AIE2PAsmBackend.h"
 #include "llvm/MC/MCObjectStreamer.h"
-#include "llvm/Support/ErrorHandling.h"
 
 using namespace llvm;
-
-bool AIE2PAsmBackend::isCall(unsigned Opcode) const {
-  switch (Opcode) {
-  // TODO Add Call Opcode e.g. JL
-  default:
-    break;
-  }
-  return false;
-}
-
-bool AIE2PAsmBackend::isDelaySlotInstr(unsigned Opcode) const {
-  switch (Opcode) {
-  default:
-    break;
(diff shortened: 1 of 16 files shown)


llvm/lib/Target/AIE/MCTargetDesc/aie2p/AIE2PAsmBackend.cpp
```diff
@@ -10,40 +10,9 @@
...
-bool AIE2PAsmBackend::isDelaySlotInstr(unsigned Opcode) const {
-  switch (Opcode) {
-  default:
-    break;
-  // TODO Add other opcode e.g. JL, JZ, etc.
-  case AIE2P::RET:
-    return true;
-  }
-  return false;
-}
```
stephenneuendorffer, line 35: I guess this API isn't used for AIE2 or AIE2P?

stephenneuendorffer (approved):

llvm/lib/Target/AIE/MCTargetDesc/aie2p/AIE2PAsmBackend.cpp
konstantinschwarz: That's right. The alternative API in AIEXXXInstrInfo.cpp is `getNumDelaySlots`, which queries the MachineInstr property `hasDelaySlot`, which we set in AIEXXXInstrInfo.td

martien-de-jong: MC layer doen't know anymore about delay slots. The postscheduler creates the empty bundles, the machine alignment pass fills them out to the right size. When lowering to MC layer, we turn machine instructions into operands of the format instruction. This is copied/extended from hexagon, that had the advantage of uniform instruction sizes.

PR #568, opened 2025-07-24
Isoard.upstream sync

isoard-amd:


llvm/lib/Target/AIE/MCTargetDesc/AIEMCExpr.h
```diff
@@ -38,7 +38,7 @@ class AIEMCExpr : public MCTargetExpr {
 
   int64_t evaluateAsInt64(int64_t Value) const;
 
-  bool evaluatePCRelLo(MCValue &Res, const MCAsmLayout *Layout,
+  bool evaluatePCRelLo(MCValue &Res, const MCAssembler *Layout,
```
konstantinschwarz: This function isn't used anywhere (there isn't even a definition). Let's remove it instead?

isoard-amd: Was it forgotten or is it new?

konstantinschwarz: I think this was a copy&pasta error from very early times during the initial bring-up. I just discovered that it was never used. Also, there are no pc-relative addressing modes on current AIEs, so it makes sense that this is unused

PR #573, opened 2025-07-25
[AIE2][AIE2P] Fix vadd intrinsic header

khallouh:
The comparison with zero result should be all zeros instead of all ones for unsigned inputs


clang/test/CodeGen/aie/aie-vadd-intrinsics.cpp
```diff
@@ -679,9 +679,13 @@ unsigned long long test_gt_v64uint8(v64uint8 a, v64uint8 b, bool sgn) {
   return gt(a, b, sgn);
 }
 
-// CHECK-COMMON-LABEL: @_Z17test_ltz_v64uint8Dv64_h(
-// CHECK-COMMON-NEXT:  entry:
-// CHECK-COMMON-NEXT:    ret i64 -1
+// AIE2-LABEL: @_Z17test_ltz_v64uint8Dv64_h(
+// AIE2-NEXT:  entry:
+// AIE2-NEXT:    ret i64 -1
```
F-Stuckmann: is AIE2 also wrong?

khallouh: Yes, I fix it for AIE2 and AIE2P

khallouh: Some intrinsics were not fixed, I updated them and took the opportunity to remove the unused leftover headers from a recent refactoring.

clang/lib/Headers/aiev2_vadd.h
```diff
@@ -107,14 +107,14 @@ INTRINSIC(v64uint8) bor(v64uint8 a, v64uint8 b) { return a | b; }
 
 INTRINSIC(v64uint8) bneg_ltz(v64uint8 a, unsigned long long &cmp) {
   v64uint8 r = __builtin_aiev2_vbneg_ltz8(a, cmp);
-  cmp = (unsigned long long)(-1);
```
konstantinschwarz, line 110: Nit: no need to update this header if the next one deletes it?

konstantinschwarz (approved):

clang/lib/Headers/aiev2_vadd.h
konstantinschwarz: Looks like the change is still present after your recent update?

PR #575, opened 2025-07-25
Isoard.upstream sync

isoard-amd:
Upstream changed the `GlobalISel`'s `Combiner` pattern `trunc_ext_fold` into `cast_combines` which in itself is a good clean-up.

But while doing so they also made it restricted to single use, see `llvm/lib/CodeGen/GlobalISel/CombinerHelperCasts.cpp`
```c++
  if (!MRI.hasOneNonDBGUse(Ext->getReg(0)))
    return false;
```


isoard-amd:
See a188bf51d8cd78fbc523866af0130ab6a4f62c9a for what effect it causes on
```
Failed Tests (3):
  LLVM :: CodeGen/AIE/GlobalISel/prelegalizercombiner-s20-narrowing.mir
  LLVM :: CodeGen/AIE/aie2/GlobalISel/prelegalizercombiner-s20-narrowing.mir
  LLVM :: CodeGen/AIE/aie2p/GlobalIsel/prelegalizercombiner-s20-narrowing.mir
```

isoard-amd:
We could change it to allow any number of users like d25138a0f49a6dfa6caf29e96dfd9740a7fece2c
```
    [GlobalISel] Allow trunc_of_{s,z,any}ext even on multi-use

 llvm/lib/CodeGen/GlobalISel/CombinerHelperCasts.cpp      |  3 ---
 llvm/test/CodeGen/AArch64/GlobalISel/arm64-atomic.ll     | 48 ++++++++++++++++++++++++------------------------
 llvm/test/CodeGen/AArch64/GlobalISel/arm64-pcsections.ll | 40 ++++++++++++++++++++--------------------
 llvm/test/CodeGen/AArch64/GlobalISel/combine-trunc.mir   | 25 +++++++++++++++++--------
 ```

andcarminati:
Hi @isoard-amd, thank you for pointing this detail that is very important for our compiler performance. I would suggest to keep the combiner as is to avoid conflicts with future bumps. We can include a combiner afterwards (target-specific) to retake those cases. Sometimes global decisions related to combiners may have negative impact on AIE, so we try to handle combiners in a selective way.

PR #592, opened 2025-08-04
[AIEX] fix llvm reduce crash

F-Stuckmann:
This PR fixes a llvm-reduce crash when running llvm-reduce on .mir files. 

Add missing `MachineFunctionInfo::clone` function to `AIEMachineFunctionInfo` (copied from x86)


andcarminati (approved):
LGTM.

PR #599, opened 2025-08-06
[AIEX] Prevent libcalls for int to fp conversions in ZOL

F-Stuckmann:
Do not allow ZOL when sitofp and uitofp appear in loop bodies.

### DIFF
--- a/llvm/lib/Target/AIE/AIEBaseTargetTransformInfo.cpp
+++ b/llvm/lib/Target/AIE/AIEBaseTargetTransformInfo.cpp
@@ -132,6 +132,8 @@ bool AIETTICommon::isAllowedInZOL(Instruction &I) {
   case Instruction::FDiv:
   case Instruction::FPToSI:
   case Instruction::FPToUI:
+  case Instruction::UIToFP:
+  case Instruction::SIToFP:
     return false;
   }
   return true;
--- a/llvm/test/CodeGen/AIE/aie2p/zol/no-rt-calls.ll
+++ b/llvm/test/CodeGen/AIE/aie2p/zol/no-rt-calls.ll
@@ -35,8 +35,8 @@ for.cond.cleanup:                                 ; preds = %for.body
 ; We have uitofp in the loop body, and we check whether ZOL loops
 ; were created by looking for the hallmark intrinsics
 
-; CHECK: llvm.set.loop.iterations
-; CHECK: llvm.loop.decrement
+; CHECK-NOT: llvm.set.loop.iterations
+; CHECK-NOT: llvm.loop.decrement
 
 define dso_local float @uitofpConv(i32 noundef %n) local_unnamed_addr {
 entry:
@@ -59,8 +59,8 @@ for.cond.cleanup:                                 ; preds = %for.body
 ; We have sitofp in the loop body, and we check whether ZOL loops
 ; were created by looking for the hallmark intrinsics
 
-; CHECK: llvm.set.loop.iterations
-; CHECK: llvm.loop.decrement
+; CHECK-NOT: llvm.set.loop.iterations
+; CHECK-NOT: llvm.loop.decrement
 
 define dso_local float @sitofpConv(i32 noundef %n) local_unnamed_addr {
 entry:
--- a/llvm/test/CodeGen/AIE/aie2p/zol/no-rt-calls.ll
+++ b/llvm/test/CodeGen/AIE/aie2p/zol/no-rt-calls.ll
@@ -32,6 +32,54 @@ for.cond.cleanup:                                 ; preds = %for.body
   ret float %mul
 }
 
+; We have uitofp in the loop body, and we check whether ZOL loops
+; were created by looking for the hallmark intrinsics
+
+; CHECK: llvm.set.loop.iterations
+; CHECK: llvm.loop.decrement
+
+define dso_local float @uitofpConv(i32 noundef %n) local_unnamed_addr {
+entry:
+  %s32 = add i32 undef, 0
+  br label %for.body
+
+for.body:                                         ; preds = %entry, %for.body
+  %i.05 = phi i32 [ 0, %entry ], [ %inc, %for.body ]
+  %20 = phi <16 x float> [ undef, %entry ], [ %vecins.i.i.i.i.i, %for.body ]
+  %conv = uitofp i32 %s32 to float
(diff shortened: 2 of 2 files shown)


konstantinschwarz (dismissed):
Sounds reasonable

llvm/test/CodeGen/AIE/aie2p/zol/no-rt-calls.ll
```diff
@@ -32,6 +32,54 @@ for.cond.cleanup:                                 ; preds = %for.body
   ret float %mul
 }
 
+; We have uitofp in the loop body, and we check whether ZOL loops
+; were created by looking for the hallmark intrinsics
+
+; CHECK-NOT: llvm.set.loop.iterations
+; CHECK-NOT: llvm.loop.decrement
```
martien-de-jong, line 39: could we check that it generates hwloops before the fix, to test the test?

F-Stuckmann: good idea

llvm/lib/Target/AIE/AIEMachineAlignment.cpp
```diff
@@ -268,6 +280,8 @@ bool AIEMachineAlignment::runOnMachineFunction(MachineFunction &MF) {
     LLVM_DEBUG(dbgs() << "bb." << MBB.getNumber() << "\n"; MBB.dump());
 
     if (TII.isZOLBody(MBB)) {
+      assert(isZOLBodyWellFormed(MBB, TII) &&
```
andcarminati: Nice!

llvm/test/CodeGen/AIE/aie2p/zol/no-rt-calls.ll
andcarminati: Yes, add a base test as a first commit!

llvm/test/CodeGen/AIE/aie2p/zol/no-rt-calls.ll
```diff
@@ -32,6 +32,54 @@ for.cond.cleanup:                                 ; preds = %for.body
   ret float %mul
 }
 
+; We have uitofp in the loop body, and we check whether ZOL loops
+; were created by looking for the hallmark intrinsics
```
andcarminati, line 36: In this case it is not an intrinsic, it is a proper IR instruction.

llvm/lib/Target/AIE/AIEMachineAlignment.cpp
```diff
@@ -268,6 +280,8 @@ bool AIEMachineAlignment::runOnMachineFunction(MachineFunction &MF) {
...
+             "Error: Detected Invalid Instruction in Hardware Loop.");
```
andcarminati: nit: `... in Hardware Loop Body.`

martien-de-jong:
As discussed offline: We are preventing a crash in the machine alignment by making use of a lucky coincidence. If a future architecture would allow JL in ZOL, AIEMachineAlignment would still be broken.
The crash was because we could not dereference an iterator that occurred in getMachineBundleSize to perform an isBundle() test; That suggests that that iterator was a special kind of iterator like end(), and we should make that bundle test robust for the case, making sure we return the right size. My hunch is that we are looking at a case where the return address and the last-bundle-of-ZOL coincide, giving an empty alignment region.
Although we cannot reach the situation from a normal compiler invocation (after this fix) we can reconstruct a minimal mir example that we can unit test to be correctly handled by AIEMachineAlignment. Only a PR targeting AIEMachineAlignment with that unit  test attached would properly solve the issue.

llvm/lib/Target/AIE/AIEBaseTargetTransformInfo.cpp
```diff
@@ -132,6 +132,8 @@ bool AIETTICommon::isAllowedInZOL(Instruction &I) {
   case Instruction::FDiv:
```
andcarminati, line 132: FRem?

andcarminati: You could check all Ops/types with `libcall` in LegalizerInfo. I see several ones that are not covered here.

andcarminati:
Another call inside a ZOL loop:
```
void foo (int * array, int n) {
    for (int i = 0; i < n; i++) {
        array[i] = array[i]/3;
    }
}

```

llvm/test/CodeGen/AIE/aie2p/zol/no-rt-calls.ll
martien-de-jong: llvm.set.loop.iterations is still an intrinsic?

llvm/lib/Target/AIE/AIEMachineAlignment.cpp
```diff
@@ -248,6 +248,18 @@ void padRegions(std::vector<MultiBlockRegion> &AllRegions) {
   }
 }
 
+/// \return whether ZOL body in \p MBB is well formed.
+bool isZOLBodyWellFormed(MachineBasicBlock &MBB, const AIEBaseInstrInfo &TII) {
+  assert(TII.isZOLBody(MBB));
+
+  StringRef ErrorInfo;
+  for (auto It = MBB.begin(); It != MBB.end(); ++It) {
+    if (!TII.verifyControlFlowConstraints(*It, ErrorInfo))
```
martien-de-jong: nit: why not `for (auto &MI  : MBB) `?

martien-de-jong: nit: Perhaps it would pay off to factor this into `verifyLegalInZOL(MI, ErorrInfo)`

martien-de-jong: I wonder why Andreu didn't suggest to use `return llvm::all_of(MBB, lambda)`

F-Stuckmann:
So I want to split this pr. 
The first part is just a hot-fix of the JL in the ZOL body. 
The second part is a removing of a ZOL in the legalizer, if a call is detected. 
Since part 2 may take a little while longer, I want to get the hotfix out first.

martien-de-jong (approved):
Probably enough to unblock, but still inherently fragile. As discussed, lowering ZOL in legalization would provide a stable solution.

PR #606, opened 2025-08-08
Isoard.upstream sync2

isoard-amd:


llvm/lib/Target/AVR/Disassembler/AVRDisassembler.cpp
```diff
@@ -13,6 +13,7 @@
 //
 //===----------------------------------------------------------------------===//
 
+#include "AVRRegisterInfo.h"
```
konstantinschwarz, line 16: Strange, why do we have to touch this backend? Did we pick a broken upstream commit?

clang/lib/Basic/Targets/AIE.h
```diff
@@ -84,6 +84,10 @@ class LLVM_LIBRARY_VISIBILITY AIETargetInfo : public TargetInfo {
     }
   }
 
+  bool isAddressSpaceSupersetOf(LangAS A, LangAS B) const override {
```
konstantinschwarz, line 87: Nice!

llvm/lib/Target/AVR/Disassembler/AVRDisassembler.cpp
isoard-amd: That's because we have changes in this to handle zero bitwidth thingies

isoard-amd: See f11e493baf16667ce6936fa018323728b7bb7cf3.

isoard-amd: They simply cleaned-up the unused headers, but we were using one of them in our change.

PR #608, opened 2025-08-08
[AIEX] Add memset optimizations

andcarminati:
The idea is to reduce the number of stores related to memsets, avoiding whenever possible byte stores.

Current optimizations:

1. Memset align: for big stack objects, we align them to use vector stores.
2. Memset peeling: peel byte  and/or short stores out from memsets until we reach alignment, so we can avoid full byte scalarization of memsets.
3. Store combination: we combine stores into memsets to reach optimal store sequences. We can also take the leftovers from a previous memset expansion to create smaller vector stores.


llvm/test/CodeGen/AIE/GlobalISel/combine-memset-stack-align.mir
```diff
@@ -419,47 +315,215 @@ body:             |
...
-    ; LEGALIZED-NEXT: [[PTR_ADD9:%[0-9]+]]:_(p0) = G_PTR_ADD [[FRAME_INDEX]], [[C21]](s20)
-    ; LEGALIZED-NEXT: G_STORE [[C20]](s32), [[PTR_ADD9]](p0) :: (store (s32) into %ir.object1 + 40, align 8)
-    ; LEGALIZED-NEXT: [[C22:%[0-9]+]]:_(s32) = G_CONSTANT i32 0
-    ; LEGALIZED-NEXT: [[C23:%[0-9]+]]:_(s20) = G_CONSTANT i20 44
-    ; LEGALIZED-NEXT: [[PTR_ADD10:%[0-9]+]]:_(p0) = G_PTR_ADD [[FRAME_INDEX]], [[C23]](s20)
-    ; LEGALIZED-NEXT: G_STORE [[C22]](s32), [[PTR_ADD10]](p0) :: (store (s32) into %ir.object1 + 44, basealign 8)
-    ; LEGALIZED-NEXT: [[C24:%[0-9]+]]:_(s20) = G_CONSTANT i20 48
-    ; LEGALIZED-NEXT: [[PTR_ADD11:%[0-9]+]]:_(p0) = G_PTR_ADD [[FRAME_INDEX]], [[C24]](s20)
-    ; LEGALIZED-NEXT: G_STORE [[C1]](s32), [[PTR_ADD11]](p0) :: (store (s32) into %ir.object1 + 48, align 8)
+    ; LEGALIZED-NEXT: [[COPY7:%[0-9]+]]:_(s32) = COPY [[AIE_SEXT_EXTRACT_VECTOR_ELT]](s32)
```
andcarminati: Verify this!

llvm/lib/Target/AIE/AIECombinerHelper.cpp
```diff
@@ -13,6 +13,7 @@
 #include "AIE2TargetMachine.h"
 #include "AIEBaseInstrInfo.h"
 #include "MCTargetDesc/AIE2MCTargetDesc.h"
+#include "llvm/ADT/STLExtras.h"
```
andcarminati: Check includes.

llvm/lib/Target/AIE/AIECombinerHelper.cpp
```diff
@@ -3711,3 +3712,210 @@ bool llvm::matchPhiBitcast(MachineInstr &Phi, MachineRegisterInfo &MRI,
 
   return true;
 }
+bool llvm::matchAlignMemset(MachineInstr &MI, MachineRegisterInfo &MRI,
```
andcarminati, line 3800: nit: new line.

andcarminati:
TODO: add tests with different initializers.

llvm/lib/Target/AIE/AIECombinerHelper.cpp
```diff
@@ -3711,3 +3712,210 @@ bool llvm::matchPhiBitcast(MachineInstr &Phi, MachineRegisterInfo &MRI,
...
+        B.buildStore(DataReg, BuildPADD(Offset),
+                     *BuildMMO(MMO->getSize(), Align(1)));
+        OutlinedStores++;
+      }
+
+      // Dont't peel two more if we are aligned.
+      if ((Offset + 1) % 4) {
+        // Store the next two bytes to align to the next word boundary.
+        Register DataRegAdjustedToS16 =
+            B.buildConstant(LLT::scalar(16), (Initializer << 8) | 1).getReg(0);
```
andcarminati: Wrong: should be `(Initializer << 8)  |  Initializer `.

llvm/test/CodeGen/AIE/GlobalISel/combine-memset-peeling.mir
```diff
@@ -0,0 +1,614 @@
...
+    ; CHECK: liveins: $p0
+    ; CHECK-NEXT: {{  $}}
+    ; CHECK-NEXT: [[COPY:%[0-9]+]]:_(p0) = COPY $p0
+    ; CHECK-NEXT: [[C:%[0-9]+]]:_(s32) = G_CONSTANT i32 0
+    ; CHECK-NEXT: [[C1:%[0-9]+]]:_(s8) = G_CONSTANT i8 0
+    ; CHECK-NEXT: G_STORE [[C]](s32), [[COPY]](p0) :: (store (s32))
+    ; CHECK-NEXT: [[C2:%[0-9]+]]:_(s20) = G_CONSTANT i20 45
+    ; CHECK-NEXT: [[PTR_ADD:%[0-9]+]]:_(p0) = G_PTR_ADD [[COPY]], [[C2]](s20)
+    ; CHECK-NEXT: G_STORE [[C1]](s8), [[PTR_ADD]](p0) :: (store (s8))
+    ; CHECK-NEXT: [[C3:%[0-9]+]]:_(s16) = G_CONSTANT i16 1
```
andcarminati: Wrong constant here!

llvm/lib/Target/AIE/AIECombinerHelper.cpp
```diff
@@ -3711,3 +3712,210 @@ bool llvm::matchPhiBitcast(MachineInstr &Phi, MachineRegisterInfo &MRI,
...
+
+  if (!MMO)
+    return false;
+
+  MachineFunction *MF = MI.getMF();
+  const MachineFrameInfo &MFI = MF->getFrameInfo();
+  const TargetFrameLowering *TFI = MF->getSubtarget().getFrameLowering();
+  const int FrameIndex = DefDataInst->getOperand(1).getIndex();
+  int64_t BiggestAlignment = TFI->getStackAlignment();
+  const int64_t Offset = MFI.getObjectOffset(FrameIndex);
```
andcarminati: This information may not be accurate, can be zero!

llvm/lib/Target/AIE/AIECombinerHelper.cpp
andcarminati: Maybe force alignment here?

llvm/lib/Target/AIE/AIECombinerHelper.cpp
```diff
@@ -3711,3 +3712,210 @@ bool llvm::matchPhiBitcast(MachineInstr &Phi, MachineRegisterInfo &MRI,
...
+
+  // Convert to a power-of-two alignment.
+  while ((Offset % BiggestAlignment) != 0)
```
andcarminati: Change the approach to work directly on fixing the alignment on the stack object and memset.

llvm/test/CodeGen/AIE/GlobalISel/combine-memset-stack-align.mir
```diff
@@ -0,0 +1,376 @@
...
+
+...
+---
+name:            aligned_object_exact_size
+alignment:       16
+exposesReturnsTwice: false
+tracksRegLiveness: true
+stack:
+  - { id: 0, name: object0, type: default, offset: 0, size: 32,
+      alignment: 8, stack-id: default, callee-saved-register: '', callee-saved-restored: true,
```
andcarminati, line 96: Could we generate check lines for this alignment?

llvm/lib/Target/AIE/AIECombinerHelper.cpp
```diff
@@ -3711,3 +3716,62 @@ bool llvm::matchPhiBitcast(MachineInstr &Phi, MachineRegisterInfo &MRI,
...
+  assert(MI.getOpcode() == TargetOpcode::G_MEMSET && "Expected a G_MEMSET");
+
+  if (!MemsetOptimizations)
+    return false;
+
+  // Try to keep alignment increase as minimum.
+  const unsigned BasicVectorByteSize = TII.getBasicVecRegSize() / 8;
+
+  const Register CountReg = MI.getOperand(2).getReg();
+  auto Cst = getIConstantVRegValWithLookThrough(CountReg, MRI);
```
martien-de-jong, line 3815: expand auto?

llvm/lib/Target/AIE/AIECombinerHelper.cpp
```diff
@@ -3711,3 +3716,62 @@ bool llvm::matchPhiBitcast(MachineInstr &Phi, MachineRegisterInfo &MRI,
...
+  if (!Cst)
+    return false;
+  const unsigned ByteCount = Cst->Value.getZExtValue();
+
+  // Can we fill a basic vector?
+  if (ByteCount < BasicVectorByteSize)
+    return false;
+
+  const Register PtrReg = MI.getOperand(0).getReg();
+  const MachineInstr *DefDataInst = MRI.getUniqueVRegDef(PtrReg);
```
martien-de-jong, line 3825: const DefDataInst

llvm/lib/Target/AIE/AIECombinerHelper.cpp
```diff
@@ -3775,3 +3775,167 @@ bool llvm::matchAlignMemset(MachineInstr &MI, MachineRegisterInfo &MRI,
...
+        B.buildStore(DataReg, BuildPADD(Offset),
+                     *BuildMMO(MMO->getSize(), Align(1)));
+        OutlinedStores++;
+      }
+
+      // Dont't peel two more if we are aligned.
+      if ((Offset + 1) % 4) {
+        // Store the next two bytes to align to the next word boundary.
+        Register DataRegAdjustedToS16 =
+            B.buildConstant(LLT::scalar(16), (Initializer << 8) | Initializer)
```
martien-de-jong, line 4008: We've sign-extended Initializer. Could the sign bits  overflow into  the next byte  using this bitwise or, i.e. should we mask?

andcarminati: Good catch!

andcarminati:
QoR Results:

PM Size:
Core_PMSize

Note that important kernels were positively affected.
Avg. reduction: -1.68%

Instruction count:
Core_Insn_Count
Avg. reduction: -0.49%

Stack reduction:
Core_StackSize

The regressions are a side-effect of using one ne more vector. In case of function calls those vectors are saved. In absolute terms, we increase the stack in 64 bytes.

llvm/lib/Target/AIE/AIECombinerHelper.cpp
```diff
@@ -3783,3 +3791,395 @@ bool llvm::matchPhiOfUndef(MachineInstr &Phi, MachineRegisterInfo &MRI,
...
+    // Tricky part: we cannot erase the matched stores, make it dead.
+    for (GStore *ToDeleteMI : MatchedSeqStores) {
+      Observer.changingInstr(*ToDeleteMI);
+      makeStoreDead(ToDeleteMI, B.getTII(), MRI);
+      Observer.changedInstr(*ToDeleteMI);
+    }
+  };
+
+  return true;
+}
```
andcarminati: nit: new line.

llvm/lib/Target/AIE/AIECombinerHelper.cpp
```diff
@@ -3783,3 +3791,395 @@ bool llvm::matchPhiOfUndef(MachineInstr &Phi, MachineRegisterInfo &MRI,
...
+  const auto CstInit = getIConstantVRegValWithLookThrough(DataReg, MRI);
+  if (!CstInit)
+    return false;
+  const uint64_t Initializer = CstInit->Value.getZExtValue();
+
+  const auto RegAndOffset = getPtrAndConstantOffsetFromReg(PtrReg, MRI);
+  if (!RegAndOffset)
+    return false;
+
+  const unsigned Offset = RegAndOffset->second;
```
andcarminati: nit: signed type.

llvm/lib/Target/AIE/AIECombinerHelper.cpp
```diff
@@ -3783,3 +3791,395 @@ bool llvm::matchPhiOfUndef(MachineInstr &Phi, MachineRegisterInfo &MRI,
...
+
+  const Register DataReg = StMI.getValueReg();
+  const LLT DataType = MRI.getType(DataReg);
+
+  // We already have a vector store.
+  if (DataType.getSizeInBytes() >= MMOAlign.value())
+    return false;
+
+  // We can merge over aligned small types, provided that
+  // the root ot the memset (first store) is a byte or short
```
andcarminati: nit: of

llvm/test/CodeGen/AIE/GlobalISel/combine-seq-stores-memset.mir
```diff
@@ -0,0 +1,343 @@
...
+    G_STORE %82(s16), %90(p0) :: (store (s16), align 4)
+    %91:_(s20) = G_CONSTANT i20 74
+    %92:_(p0) = nuw G_PTR_ADD %84, %91(s20)
+    G_STORE %82(s16), %92(p0) :: (store (s16))
+    PseudoRET implicit $lr
+
+...
+
+---
+name:            combine_to_s32_mixed
```
F-Stuckmann, line 72: nice test

llvm/lib/Target/AIE/AIECombinerHelper.cpp
```diff
@@ -3783,3 +3791,395 @@ bool llvm::matchPhiOfUndef(MachineInstr &Phi, MachineRegisterInfo &MRI,
...
+// align this stack object to avoid store scalarization during
+// legalization.
+bool llvm::matchAlignMemset(MachineInstr &MI, MachineRegisterInfo &MRI,
+                            const AIEBaseInstrInfo &TII,
+                            GISelChangeObserver &Observer,
+                            BuildFnTy &MatchInfo) {
+  assert(MI.getOpcode() == TargetOpcode::G_MEMSET && "Expected a G_MEMSET");
+
+  if (!MemsetOptimizations)
+    return false;
```
F-Stuckmann, line 3807: nit: add debug message, that mem set optimizations are disabled and will be skipped

llvm/lib/Target/AIE/AIECombinerHelper.cpp
```diff
@@ -3783,3 +3791,395 @@ bool llvm::matchPhiOfUndef(MachineInstr &Phi, MachineRegisterInfo &MRI,
...
+
+  // Try to keep alignment increase as minimum.
+  const unsigned BasicVectorByteSize = TII.getBasicVecRegSize() / 8;
```
F-Stuckmann, line 3810: In AIE2P this return 256, but we could also perform 128 bit stores, right?

llvm/test/CodeGen/AIE/GlobalISel/combine-memset-stack-align.mir
```diff
@@ -0,0 +1,356 @@
...
+# RUN: llc -mtriple aie2p  -start-before=aie2p-prelegalizer-combiner \
+# RUN:      -stop-after=legalizer -verify-machineinstrs -o - %s  | FileCheck %s --check-prefix=LEGALIZED
+--- |
+  declare void @llvm.memset.p0.i64(ptr nocapture writeonly, i8, i64, i1 immarg) #0
+
+  declare void @prevent_dce(ptr nocapture writeonly) #0
+
+  define fastcc void @aligned_object_exact_size() personality ptr null {
+  entry:
+    %object0 = alloca [32 x i8], i32 0, align 8
```
F-Stuckmann, line 24: Could you add a test with 16xi8 size (128 bits)

llvm/test/CodeGen/AIE/GlobalISel/combine-seq-stores-memset.mir
```diff
@@ -0,0 +1,343 @@
...
+    %105:_(s20) = G_CONSTANT i20 97
+    %106:_(p0) = nuw G_PTR_ADD %84, %105(s20)
+    G_STORE %82(s8), %106(p0) :: (store (s8))
+    PseudoRET implicit $lr
+...
+
+# Negative test cases:
+
+---
+name:            no_combine_avoid_combiner_loop
```
F-Stuckmann, line 200: CHECK: we can't combine here, because we aren't byte aligned?

llvm/test/CodeGen/AIE/GlobalISel/combine-seq-stores-memset.mir
```diff
@@ -0,0 +1,343 @@
...
+    G_STORE %82(s8), %90(p0) :: (store (s8), align 4)
+    %91:_(s20) = G_CONSTANT i20 73
+    %92:_(p0) = nuw G_PTR_ADD %84, %91(s20)
+    G_STORE %13(s16), %92(p0) :: (store (s6))
+    PseudoRET implicit $lr
+
+...
+
+---
+name:            no_combine_to_s32_load
```
F-Stuckmann, line 235: CHECK: we can't combine, because of alias analysis in this case?

llvm/lib/Target/AIE/AIECombinerHelper.cpp
```diff
@@ -3783,3 +3791,395 @@ bool llvm::matchPhiOfUndef(MachineInstr &Phi, MachineRegisterInfo &MRI,
...
+                     MF->getMachineMemOperand(
+                         MachinePointerInfo::getFixedStack(*MF, FrameIndex),
+                         MachineMemOperand::MOStore, Size, OptimalAlign));
+  };
+
+  return true;
+}
+
+static std::optional<std::pair<Register, int64_t>>
+getPtrAndConstantOffset(const MachineInstr *MI, unsigned PointerIndex,
```
F-Stuckmann, line 3860: I feel like this could be also useful in`AIEClusterBaseAddress.cpp`, maybe move this into a util header

llvm/lib/Target/AIE/AIECombinerHelper.cpp
```diff
@@ -3783,3 +3791,395 @@ bool llvm::matchPhiOfUndef(MachineInstr &Phi, MachineRegisterInfo &MRI,
...
+
+  if (Cst)
+    return std::make_pair(MI->getOperand(PointerIndex).getReg(),
+                          Cst->Value.getSExtValue());
+
+  return std::nullopt;
+}
+
+static std::optional<std::pair<Register, int64_t>>
+getPtrAndConstantOffsetFromReg(Register PtrReg, MachineRegisterInfo &MRI) {
```
F-Stuckmann, line 3879: why isn't this merged with `getPtrAndConstantOffset` ? I think these two functions could be combined.

llvm/lib/Target/AIE/AIECombinerHelper.cpp
```diff
@@ -3783,3 +3791,395 @@ bool llvm::matchPhiOfUndef(MachineInstr &Phi, MachineRegisterInfo &MRI,
...
+  if (DefPtrReg->getOpcode() == TargetOpcode::G_PTR_ADD) {
+    return getPtrAndConstantOffset(DefPtrReg, 1, MRI);
+  }
+  return std::make_pair(PtrReg, 0);
+}
+
+static bool isBasePointerWordAligned(Register BasePtr,
+                                     MachineRegisterInfo &MRI) {
+
+  auto IsBasePointerAligned = [&](const MachineInstr *MI, int64_t Offset) {
```
F-Stuckmann: nit: This name could be updated, since we only care for stores.

llvm/lib/Target/AIE/AIECombinerHelper.cpp
```diff
@@ -3783,3 +3791,395 @@ bool llvm::matchPhiOfUndef(MachineInstr &Phi, MachineRegisterInfo &MRI,
...
+    if (!isa<GLoadStore>(MI))
```
F-Stuckmann, line 3892: If you extend this check to not-exit with `G_MEMSET` you could reuse this function in `matchAlignMemset`

llvm/lib/Target/AIE/AIECombinerHelper.cpp
```diff
@@ -3783,3 +3791,395 @@ bool llvm::matchPhiOfUndef(MachineInstr &Phi, MachineRegisterInfo &MRI,
...
+    const MachineMemOperand *MMO = MI->memoperands().front();
+    return MMO && (MMO->getAlign().value() % 4 == 0) && (Offset % 4 == 0);
+  };
+
+  for (const MachineInstr &MI : MRI.use_instructions(BasePtr)) {
+    if (MI.getOpcode() == TargetOpcode::G_PTR_ADD) {
+      auto RegAndOffset = getPtrAndConstantOffset(&MI, 0, MRI);
+      if (!RegAndOffset)
+        continue;
+      if (any_of(MRI.use_instructions(RegAndOffset->first),
```
F-Stuckmann, line 3906: what happens if a store is basePointerAligned and another is not basePointerAligned?
I don't think the method should return in that case

llvm/lib/Target/AIE/AIECombinerHelper.cpp
```diff
@@ -3783,3 +3791,395 @@ bool llvm::matchPhiOfUndef(MachineInstr &Phi, MachineRegisterInfo &MRI,
...
+                 [&](const MachineInstr &UseMI) {
+                   return IsBasePointerAligned(&UseMI, RegAndOffset->second);
+                 }))
+        return true;
+    } else if (IsBasePointerAligned(&MI, 0)) {
```
F-Stuckmann: Analogues here, wouldn't we want all stores to be base pointer aligned, not just the first?

llvm/lib/Target/AIE/AIECombinerHelper.cpp
```diff
@@ -3783,3 +3791,395 @@ bool llvm::matchPhiOfUndef(MachineInstr &Phi, MachineRegisterInfo &MRI,
...
+
+  MachineMemOperand *MMO = MI.memoperands().front();
+
+  if (!MMO)
+    return false;
+  const Align MMOAlign = MMO->getAlign();
+
+  // If it is already aligned we have nothing to do.
+  if (MMOAlign.value() % 4 == 0)
+    return false;
```
F-Stuckmann, line 3945: You could reuse an updated version of `IsBasePointerAligned` here ;)

llvm/lib/Target/AIE/AIECombinerHelper.cpp
```diff
@@ -3783,3 +3791,395 @@ bool llvm::matchPhiOfUndef(MachineInstr &Phi, MachineRegisterInfo &MRI,
...
+
+  const Register SizeReg = MI.getOperand(2).getReg();
+
+  const auto Cst = getIConstantVRegValWithLookThrough(SizeReg, MRI);
+  if (!Cst)
+    return false;
+  const int64_t Size = Cst->Value.getSExtValue();
```
F-Stuckmann, line 3952: With a more generic version of `getPtrAndConstantOffset`, we could also use it here.

llvm/lib/Target/AIE/AIECombinerHelper.cpp
```diff
@@ -3783,3 +3791,395 @@ bool llvm::matchPhiOfUndef(MachineInstr &Phi, MachineRegisterInfo &MRI,
...
+    return false;
+
+  // Try to keep alignment increase as minimum.
+  const unsigned BasicVectorByteSize = TII.getBasicVecRegSize() / 8;
+
+  const Register CountReg = MI.getOperand(2).getReg();
+  auto Cst = getIConstantVRegValWithLookThrough(CountReg, MRI);
+  if (!Cst)
+    return false;
+  const unsigned ByteCount = Cst->Value.getZExtValue();
```
F-Stuckmann, line 3818: With a more generic version of getPtrAndConstantOffset, we could also use it here.

llvm/lib/Target/AIE/AIECombinerHelper.cpp
```diff
@@ -3783,3 +3791,395 @@ bool llvm::matchPhiOfUndef(MachineInstr &Phi, MachineRegisterInfo &MRI,
...
+  //  - G_STORE s16
+  //  - G_MEMSET (... n - 3 ...)
+  // If we are short-aligned we can generate
+  //  - G_STORE s16
+  //  - G_MEMSET (... n - 2 ...)
+  if (Size < 3)
+    return false;
+
+  Register PtrReg = MI.getOperand(0).getReg();
+  const Register DataReg = MI.getOperand(1).getReg();
```
F-Stuckmann, line 3955: This could also use a generic variant of `getPtrAndConstantOffset`

llvm/lib/Target/AIE/AIECombinerHelper.cpp
```diff
@@ -3783,3 +3791,395 @@ bool llvm::matchPhiOfUndef(MachineInstr &Phi, MachineRegisterInfo &MRI,
...
+  // Next step is to prove that the base pointer is word-aligned.
+  // As we cannot assume, we can search for aligned uses of the base pointer.
+  if (!isBasePointerWordAligned(PtrReg, MRI))
+    return false;
+
+  MatchInfo = [=, &MI, &MRI, &Observer](MachineIRBuilder &B) {
+    auto &MF = B.getMF();
+    unsigned OutlinedStores = 0;
+
+    auto BuildPADD = [&](int64_t CurrentOffset) {
```
F-Stuckmann, line 3976: nit: we are not using `CurrentOffset`

llvm/lib/Target/AIE/AIECombinerHelper.cpp
```diff
@@ -3783,3 +3791,395 @@ bool llvm::matchPhiOfUndef(MachineInstr &Phi, MachineRegisterInfo &MRI,
...
+    if (Offset % 4) {
+      // If byte-aligned, align to the next short boundary.
+      if (Offset % 2) {
+        B.buildStore(DataReg, BuildPADD(Offset),
+                     *BuildMMO(MMO->getSize(), Align(1)));
+        OutlinedStores++;
+      }
+
+      // Dont't peel two more if we are aligned.
+      if ((Offset + 1) % 4) {
```
F-Stuckmann: CHECK: we enter this function, even if we already aligned the offset to a short? Don't we have to update the Offset in that case? 
Also in the case of Short alignment, isn't this check wrong? 

Also: I think we could reduce the nesting here for clarity, if we used something like `CurrentOffset`
`if (CurrerntOffset % 4 ==0 ) // skip peeling`
``
if (CurrentOffset %2) // handle byte allignment and update CurrentOffset
``
`if (CurrentOffset % 4) // handle short alignment `

llvm/lib/Target/AIE/AIECombinerHelper.cpp
```diff
@@ -3783,3 +3791,395 @@ bool llvm::matchPhiOfUndef(MachineInstr &Phi, MachineRegisterInfo &MRI,
...
+        Register DataRegAdjustedToS16 =
+            B.buildConstant(LLT::scalar(16), (Initializer << 8) | Initializer)
+                .getReg(0);
+        B.buildStore(DataRegAdjustedToS16, BuildPADD(Offset + 1),
+                     *BuildMMO(2, Align(2)));
+        OutlinedStores += 2;
+      }
+    }
+
+    const unsigned NewSize = Size - OutlinedStores;
```
F-Stuckmann: nit rename `OutlineStores` to `PeeledBytes`, since this is what you actually did and makes this arithmetic here clearer.

llvm/lib/Target/AIE/AIECombinerHelper.cpp
```diff
@@ -3783,3 +3791,395 @@ bool llvm::matchPhiOfUndef(MachineInstr &Phi, MachineRegisterInfo &MRI,
...
+    // No bytes left to memset.
+    if (NewSize == 0) {
+      MI.eraseFromParent();
+      return;
+    }
+
+    const int64_t MemsetOffset = Offset + OutlinedStores;
+    assert(MemsetOffset % 4 == 0 && "Memset still unaligned?");
+    // Now, what remains is aligned, we just need to fix Offset, Size and MMO.
+    MachineMemOperand *NewMMOMemSet = BuildMMO(MMO->getSize(), Align(4));
```
F-Stuckmann, line 4027: CHECK: if we already peeled stores, shouldn't the new MMO have a smaller size than the original one?

llvm/lib/Target/AIE/AIECombinerHelper.cpp
```diff
@@ -3783,3 +3791,395 @@ bool llvm::matchPhiOfUndef(MachineInstr &Phi, MachineRegisterInfo &MRI,
...
+    MI.dropMemRefs(MF); // Safe to drop the MMO now.
+    MI.addMemOperand(MF, NewMMOMemSet);
+    MI.getOperand(2).setReg(
+        B.buildConstant(LLT::scalar(20), NewSize).getReg(0));
```
F-Stuckmann, line 4031: Is the MMO still up to date, if the we use NewSize here?

llvm/lib/Target/AIE/AIECombinerHelper.cpp
```diff
@@ -3783,3 +3791,395 @@ bool llvm::matchPhiOfUndef(MachineInstr &Phi, MachineRegisterInfo &MRI,
...
+    MI.getOperand(0).setReg(BuildPADD(MemsetOffset));
+  };
+
+  return true;
+}
+
+static std::optional<std::pair<Register, int64_t>>
+getPtrAndConstantOffsetFromStore(const GStore *StMI, MachineRegisterInfo &MRI) {
```
F-Stuckmann, line 4039: nit: I think a generalized `getPtrAndConstantOffset` could also handle this case

llvm/lib/Target/AIE/AIECombinerHelper.cpp
andcarminati: As this is on by on by default, I see no point for this message.

llvm/lib/Target/AIE/AIECombinerHelper.cpp
andcarminati: Good catch!

llvm/lib/Target/AIE/AIECombinerHelper.cpp
```diff
@@ -3783,3 +3791,395 @@ bool llvm::matchPhiOfUndef(MachineInstr &Phi, MachineRegisterInfo &MRI,
...
+
+  MachineMemOperand *MMO = StMI.memoperands().front();
+
+  if (!MMO)
+    return false;
+  const Align MMOAlign = MMO->getAlign();
+
+  // Small alignments are not interesting, we are trying to match
+  // vector stores here.
+  const bool IsVectorAlignment = MMOAlign.value() >= 16;
```
F-Stuckmann: nit: move this Definition to where it is first used. Up here it adds confusion with the comment.

llvm/lib/Target/AIE/AIECombinerHelper.cpp
```diff
@@ -3783,3 +3791,395 @@ bool llvm::matchPhiOfUndef(MachineInstr &Phi, MachineRegisterInfo &MRI,
...
+  // all.
+  auto GetZeroStoreSizeInBytes = [&](GStore &CurrSt) -> unsigned {
+    const Register DataReg = CurrSt.getValueReg();
+    const LLT DataType = MRI.getType(DataReg);
+
+    // We already have a vector store, don't merge it.
+    if (DataType.getSizeInBytes() >= MMOAlign.value())
+      return 0;
+    auto Cst = getIConstantVRegValWithLookThrough(DataReg, MRI);
+    return (Cst && Cst->Value.getZExtValue() == 0) ? DataType.getSizeInBytes()
```
F-Stuckmann: Why are we limited to 0 here? I guess, this would still work with any constant, right?

llvm/lib/Target/AIE/AIECombinerHelper.cpp
```diff
@@ -3783,3 +3791,395 @@ bool llvm::matchPhiOfUndef(MachineInstr &Phi, MachineRegisterInfo &MRI,
...
+  // vector stores here.
+  const bool IsVectorAlignment = MMOAlign.value() >= 16;
+  if (MMOAlign.value() < 4)
+    return false;
+
+  const Register DataReg = StMI.getValueReg();
+  const LLT DataType = MRI.getType(DataReg);
+
+  // We already have a vector store.
+  if (DataType.getSizeInBytes() >= MMOAlign.value())
```
F-Stuckmann: I don't think this check here is necessary, since we also check it in line 4099

llvm/lib/Target/AIE/AIECombinerHelper.cpp
```diff
@@ -3783,3 +3791,395 @@ bool llvm::matchPhiOfUndef(MachineInstr &Phi, MachineRegisterInfo &MRI,
...
+
+    // We already have a vector store, don't merge it.
+    if (DataType.getSizeInBytes() >= MMOAlign.value())
+      return 0;
+    auto Cst = getIConstantVRegValWithLookThrough(DataReg, MRI);
+    return (Cst && Cst->Value.getZExtValue() == 0) ? DataType.getSizeInBytes()
+                                                   : 0;
+  };
+
+  auto PtrAndOffset = getPtrAndConstantOffsetFromStore(&StMI, MRI);
```
F-Stuckmann, line 4109: nit: For clarity I would move this right before the first use

llvm/lib/Target/AIE/AIECombinerHelper.cpp
andcarminati: The purpose of this function is just to prove that we have at least one aligned load/store related to the base pointer, we simply don't care about that the load or store is used for. We simply look to the others uses of the same pointer. For the matchAlignMemset we to the target instruction directly.

llvm/lib/Target/AIE/AIECombinerHelper.cpp
andcarminati: I one instruction is aligned, that is all we need.

llvm/lib/Target/AIE/AIECombinerHelper.cpp
andcarminati: If we see one aligned operation, it is safe to state the the pointer is aligned, otherwise we have wrong code from the beginning. Unaligned operations can be legalized by the legalizer, but wrong align information is uncoverable, so we can trust.

llvm/lib/Target/AIE/AIECombinerHelper.cpp
andcarminati: We care about loads and stores.

llvm/lib/Target/AIE/AIECombinerHelper.cpp
andcarminati: They have different use cases.

llvm/lib/Target/AIE/AIECombinerHelper.cpp
andcarminati: If we see the need, we can move for sure.

llvm/lib/Target/AIE/AIECombinerHelper.cpp
andcarminati: Here we care about one specific `memset`, `IsBasePointerAligned` is just to prove that we have an aligned operation anywhere else, that allow us to enable this current optimization.

llvm/lib/Target/AIE/AIECombinerHelper.cpp
```diff
@@ -3783,3 +3791,395 @@ bool llvm::matchPhiOfUndef(MachineInstr &Phi, MachineRegisterInfo &MRI,
...
+  // scalars to merge, go for it. The rationale for scalar merging is: If we
+  // have the first scalar store whose size is smaller then the alignment, by
+  // combining with the next, we have a chance of reducing the number of stores.
+  // For example:
+  //   STORE i8 0, [p0] (Align 16)
+  //   STORE i16 0, [p0+1]
+  //   STORE i8 0, [p0+3]
+  // Will be transformed to:
+  //   STORE i32 0, [p0] (Align 16)
+  if ((IsVectorAlignment && NumberOfBytes < 16) || MatchedSeqStores.empty())
```
F-Stuckmann: I feel like the minimum number of bytes should be something from TTA like `getMinVectorBitSize`.

llvm/lib/Target/AIE/AIECombinerHelper.cpp
andcarminati: This was great catch!

llvm/lib/Target/AIE/AIECombinerHelper.cpp
```diff
@@ -3783,3 +3791,395 @@ bool llvm::matchPhiOfUndef(MachineInstr &Phi, MachineRegisterInfo &MRI,
...
+  // In a pessimistic case, for example:
+  //   STORE i8 0, [p0] (Align 16)
+  //   STORE i16 0, [p0+1]
+  // The result will be just "rotated":
+  //   STORE i16 0, [p0] (Align 16)
+  //   STORE i8 0, [p0+2]
+  // This will cause a loop. We prevent by restricting
+  // combinations that will expand again to the same
+  // types: 3 bytes.
+  if (NumberOfBytes == 3)
```
F-Stuckmann, line 4173: Would we also have the same issue with `NumberOfBytes == 2`? Who is rotating the stores ?

llvm/lib/Target/AIE/AIECombinerHelper.cpp
andcarminati: This is related to the location size:
```
G_MEMSET %0(p0), %1(s8), %3(s20), 0 :: (store (s8) into %ir.object0, align 1)
```

Which is always `(s8)` for a memset.

llvm/lib/Target/AIE/AIECombinerHelper.cpp
andcarminati: This is not related to the pointer, this is the size register of the memset, which represents the number o bytes to be initialized.

llvm/lib/Target/AIE/AIECombinerHelper.cpp
andcarminati: This is the data register to be written, offset and base pointers are retrieved below.

llvm/test/CodeGen/AIE/GlobalISel/combine-memset-peeling.mir
```diff
@@ -0,0 +1,722 @@
...
+    ; LEGALIZED-NEXT: G_STORE [[C2]](s32), [[PTR_ADD5]](p0) :: (store (s8) into unknown-address + 5)
+    ; LEGALIZED-NEXT: [[C8:%[0-9]+]]:_(s20) = G_CONSTANT i20 6
+    ; LEGALIZED-NEXT: [[PTR_ADD6:%[0-9]+]]:_(p0) = G_PTR_ADD [[COPY]], [[C8]](s20)
+    ; LEGALIZED-NEXT: G_STORE [[C2]](s32), [[PTR_ADD6]](p0) :: (store (s8) into unknown-address + 6)
+    ; LEGALIZED-NEXT: PseudoRET implicit $lr
+    %0:_(p0) = COPY $p0
+    %1:_(s32) = G_CONSTANT i32 0
+    %2:_(s8) = G_CONSTANT i8 0
+    %3:_(s32) = G_CONSTANT i32 7
+    %5:_(s20) = G_CONSTANT i20 45
```
F-Stuckmann, line 318: Could we prove the alignment, if the offset here would maintain the alignment?

llvm/test/CodeGen/AIE/GlobalISel/combine-memset-peeling.mir
```diff
@@ -0,0 +1,722 @@
...
+    ; LEGALIZED-NEXT: [[PTR_ADD5:%[0-9]+]]:_(p0) = G_PTR_ADD [[COPY]], [[C8]](s20)
+    ; LEGALIZED-NEXT: G_STORE [[C3]](s32), [[PTR_ADD5]](p0) :: (store (s8) into unknown-address + 6)
+    ; LEGALIZED-NEXT: PseudoRET implicit $lr
+    %0:_(p0) = COPY $p0
+    %1:_(s32) = G_CONSTANT i32 0
+    %2:_(s8) = G_CONSTANT i8 0
+    %3:_(s32) = G_CONSTANT i32 7
+    G_STORE %1(s32), %0(p0) :: (store (s32), align 2)
+    %4:_(s20) = G_TRUNC %3(s32)
+    G_MEMSET %0(p0), %2(s8), %4(s20), 1 :: (store (s8))
```
F-Stuckmann, line 398: Wouldn't this memset be at least aligned by 2, since it uses the same base pointer as the aligned store?

llvm/test/CodeGen/AIE/GlobalISel/combine-memset-peeling.mir
F-Stuckmann: We do know the alignment of the memset here. Where is the difference between this and `peel_3_bytes`

llvm/lib/Target/AIE/AIECombinerHelper.cpp
andcarminati: Note that this function will always update the alignment of the memset to 4, if we can prove that the pointer is proper aligned, so next time we will skip because we test this at very beginning. If we just update alignment, Offset should be the same.

We cannot strip out the first case to be independent because we could have a realignment opportunity. The next version of this PR will include a more verbose alignment check.

llvm/lib/Target/AIE/AIECombinerHelper.cpp
andcarminati: Yes, see my previous comment about LocationSize.

llvm/lib/Target/AIE/AIECombinerHelper.cpp
andcarminati: I improved the comment.

llvm/lib/Target/AIE/AIECombinerHelper.cpp
andcarminati: Discussed offline, this leads to impractical cases.

llvm/lib/Target/AIE/AIECombinerHelper.cpp
andcarminati: No. This happens only with 3 bytes:

```
G_STORE s8
G_STORE s16

then:
G_STORE s16
G_STORE s8

then:
G_STORE s8
G_STORE s16

....
```

If we have two byte stores, they will be merged into a short store, and that is all.

llvm/test/CodeGen/AIE/GlobalISel/combine-memset-peeling.mir
andcarminati: This could be a further improvement if we found one case. In this case, we could compare the two offsets to see if the difference is aligned...

andcarminati: The difference is the uncommon odd offset.

llvm/test/CodeGen/AIE/GlobalISel/combine-memset-peeling.mir
andcarminati: The current implementation only considers word alignment. It could be a future improvement.

llvm/lib/Target/AIE/AIECombinerHelper.cpp
andcarminati: Yeah, currently it matches all the targets, but in the future we may need some change.

llvm/lib/Target/AIE/AIECombinerHelper.cpp
```diff
@@ -3787,3 +3792,64 @@ bool llvm::matchPhiOfUndef(MachineInstr &Phi, MachineRegisterInfo &MRI,
...
+  if (ByteCount < HalfVectorByteSize)
+    return false;
+
+  const Register PtrReg = MI.getOperand(0).getReg();
+  const MachineInstr *DefDataInst = MRI.getUniqueVRegDef(PtrReg);
+
+  if (DefDataInst->getOpcode() != TargetOpcode::G_FRAME_INDEX)
+    return false;
+
+  MachineMemOperand *MMO = MI.memoperands().front();
```
martien-de-jong: Can this fail if there are no memoperands?

llvm/lib/Target/AIE/AIECombinerHelper.cpp
```diff
@@ -3787,3 +3792,64 @@ bool llvm::matchPhiOfUndef(MachineInstr &Phi, MachineRegisterInfo &MRI,
...
+  MachineFrameInfo &MFI = MF->getFrameInfo();
+  const int FrameIndex = DefDataInst->getOperand(1).getIndex();
+
+  const Align OptimalAlign = Align(BasicVectorByteSize);
+  const Align MMOAlign = MMO->getAlign();
+
+  if (MMOAlign == OptimalAlign)
+    return false;
+
+  MatchInfo = [=, &MI, &MFI](MachineIRBuilder &B) {
```
martien-de-jong: nit: no need to capture MF or MFI. We can fetch them here from MI

llvm/lib/Target/AIE/AIECombinerHelper.cpp
```diff
@@ -3787,3 +3792,64 @@ bool llvm::matchPhiOfUndef(MachineInstr &Phi, MachineRegisterInfo &MRI,
...
+    MFI.setObjectAlignment(FrameIndex, OptimalAlign);
```
martien-de-jong, line 3847: I'm on unfamiliar grounds here, but should we check that this hasn't influenced some stack layout already?

llvm/lib/Target/AIE/AIECombinerHelper.cpp
```diff
@@ -3787,3 +3792,64 @@ bool llvm::matchPhiOfUndef(MachineInstr &Phi, MachineRegisterInfo &MRI,
...
+  MachineMemOperand *MMO = MI.memoperands().front();
+
+  if (!MMO)
+    return false;
+
+  MachineFunction *MF = MI.getMF();
+  MachineFrameInfo &MFI = MF->getFrameInfo();
+  const int FrameIndex = DefDataInst->getOperand(1).getIndex();
+
+  const Align OptimalAlign = Align(BasicVectorByteSize);
```
martien-de-jong: nit: if we only use half vectors, could we use halfvector alignment to save on alignment gaps?

llvm/lib/Target/AIE/AIECombinerHelper.cpp
```diff
@@ -3853,3 +3854,176 @@ bool llvm::matchAlignMemset(MachineInstr &MI, MachineRegisterInfo &MRI,
...
+
+static bool isBasePointerWordAligned(Register BasePtr,
+                                     MachineRegisterInfo &MRI) {
+
+  auto IsBasePointerAligned = [&](const MachineInstr *MI, int64_t Offset) {
+    if (!isa<GLoadStore>(MI))
+      return false;
+    const MachineMemOperand *MMO = MI->memoperands().front();
+    return MMO && matchAlignment<4>(MMO->getAlign().value()) &&
+           matchAlignment<4>(Offset);
```
martien-de-jong: I'm not sure whether we can always use front(), since that would fail on an empty memoperand range? And then, if we do have memoperands, can it return a nullptr, i.e. is the check for non-null MMO necessary?

llvm/lib/Target/AIE/AIECombinerHelper.cpp
martien-de-jong: Perhaps add a comment. The actual value must have the best alignment of all its uses.

llvm/lib/Target/AIE/AIECombinerHelper.cpp
martien-de-jong: Perhaps wrap in a getMemOperandOrNull() utility?

llvm/lib/Target/AIE/AIECombinerHelper.cpp
```diff
@@ -3853,3 +3854,176 @@ bool llvm::matchAlignMemset(MachineInstr &MI, MachineRegisterInfo &MRI,
...
+
+  // To be effective, we need at least 3 bytes here:
+  // If we are byte-aligned we can generate
+  //  - G_STORE s8
+  //  - G_STORE s16
+  //  - G_MEMSET (... n - 3 ...)
+  // If we are short-aligned we can generate
+  //  - G_STORE s16
+  //  - G_MEMSET (... n - 2 ...)
+  if (Size < 3)
```
martien-de-jong: Does it matter to bail out here? we are doing the alignment case analysis anyway. If we reach zero size by peeling  we remove the memset and we're done.

llvm/lib/Target/AIE/AIECombinerHelper.cpp
```diff
@@ -3853,3 +3854,176 @@ bool llvm::matchAlignMemset(MachineInstr &MI, MachineRegisterInfo &MRI,
...
+
+  if (Cst)
+    return std::make_pair(MI->getOperand(PointerIndex).getReg(),
+                          Cst->Value.getSExtValue());
+
+  return std::nullopt;
+}
+
+template <uint64_t TargetAlign> constexpr bool matchAlignment(uint64_t Value) {
+  return isAligned(Align(TargetAlign), Value);
```
martien-de-jong, line 3875: I find this a clumsy interface. I think the caller could usefully employ Value % TargetAlign, which should be brought down to zero, That could be done in a loop processing successive powers of two. The initial value is the shift in offset that you need to adjust the memset.

llvm/lib/Target/AIE/AIECombinerHelper.cpp
```diff
@@ -3853,3 +3854,176 @@ bool llvm::matchAlignMemset(MachineInstr &MI, MachineRegisterInfo &MRI,
...
+  };
+
+  for (const MachineInstr &MI : MRI.use_instructions(BasePtr)) {
+    if (MI.getOpcode() == TargetOpcode::G_PTR_ADD) {
+      auto RegAndOffset = getPtrAndConstantOffset(&MI, 0, MRI);
+      if (!RegAndOffset)
+        continue;
+      if (any_of(MRI.use_instructions(RegAndOffset->first),
+                 [&](const MachineInstr &UseMI) {
+                   return IsBasePointerAligned(&UseMI, RegAndOffset->second);
```
martien-de-jong: nit: the Offset part of this is invariant, and can be tested outside of the any_of()

llvm/lib/Target/AIE/AIECombinerHelper.cpp
```diff
@@ -4027,3 +4036,168 @@ bool llvm::matchPeelMemset(MachineInstr &MI, MachineRegisterInfo &MRI,
 
   return true;
 }
+
+static std::optional<std::pair<Register, int64_t>>
+getPtrAndConstantOffsetFromStore(const GStore *StMI, MachineRegisterInfo &MRI) {
+  return getPtrAndConstantOffsetFromReg(StMI->getPointerReg(), MRI);
+}
+
+// To make a store dead, we convert it to a dead ADDs and let DCE to do the
```
martien-de-jong: nit: a dead G_ADD

llvm/lib/Target/AIE/AIECombinerHelper.cpp
andcarminati: I think it is better if we can choose the alignment based on `ByteCount`.

llvm/lib/Target/AIE/AIECombinerHelper.cpp
```diff
@@ -4027,3 +4036,168 @@ bool llvm::matchPeelMemset(MachineInstr &MI, MachineRegisterInfo &MRI,
...
+                          MachineRegisterInfo &MRI) {
+  MachineFunction &MF = *StMI->getMF();
+  const Register DataReg = StMI->getValueReg();
+  StMI->dropMemRefs(MF);
+  for (int I = StMI->getNumOperands() - 1; I >= 0; I--)
+    StMI->removeOperand(I);
+  StMI->setDesc(TII.get(TargetOpcode::G_ADD));
+  for (unsigned I = 0; I < 3; I++) {
+    StMI->addOperand(MachineOperand::CreateReg(
+        I == 0 ? MRI.cloneVirtualRegister(DataReg) : DataReg, I == 0));
```
martien-de-jong: nit: just an addDef and two addUses without further conditionals?

llvm/lib/Target/AIE/AIECombinerHelper.cpp
```diff
@@ -4027,3 +4036,168 @@ bool llvm::matchPeelMemset(MachineInstr &MI, MachineRegisterInfo &MRI,
...
+  const uint64_t MinVectorStoreSize = 16;
+  MachineMemOperand *MMO = StMI.memoperands().front();
+
+  if (!MMO)
+    return false;
+  const Align MMOAlign = MMO->getAlign();
+
+  // Small alignments are less interesting, we are trying to match
+  // vector stores here. Even though we can match some byte/short
+  // stores to to word stores.
```
martien-de-jong: nit: to to

llvm/lib/Target/AIE/AIECombinerHelper.cpp
andcarminati: You are right, I will insert a check.

llvm/lib/Target/AIE/AIECombinerHelper.cpp
```diff
@@ -4027,3 +4036,168 @@ bool llvm::matchPeelMemset(MachineInstr &MI, MachineRegisterInfo &MRI,
...
+  const bool IsVectorAlignment = MMOAlign.value() >= MinVectorStoreSize;
```
martien-de-jong: declare closer to its first use?

llvm/lib/Target/AIE/AIECombinerHelper.cpp
andcarminati: The complete stack layout is finalized very late in the pipeline, basically it is safe to change. Some other combiners can also increase the alignment similarly, like the memset expander.

llvm/lib/Target/AIE/AIECombinerHelper.cpp
```diff
@@ -4027,3 +4036,168 @@ bool llvm::matchPeelMemset(MachineInstr &MI, MachineRegisterInfo &MRI,
...
+  const Register DataReg = StMI.getValueReg();
+  const LLT DataType = MRI.getType(DataReg);
+
+  // We can merge over aligned small types, provided that
+  // the root of the memset (first store) is a byte or short
+  // store. The goal is to fold this store with the next ones.
+  if (!IsVectorAlignment && DataType.getSizeInBytes() >= 4)
+    return false;
+
+  // Are we storing Zero? if yes, how many? Zero bytes means not a Zero store at
```
martien-de-jong: Overloading Zero here. ...A count of zero means....

llvm/lib/Target/AIE/AIECombinerHelper.cpp
```diff
@@ -4027,3 +4036,168 @@ bool llvm::matchPeelMemset(MachineInstr &MI, MachineRegisterInfo &MRI,
...
+
+    if (!CurrMMO)
+      return 0;
+    const Align CurrMMOAlign = MMO->getAlign();
+
+    // We already have a vector store, don't merge it.
+    if (DataType.getSizeInBytes() >= CurrMMOAlign.value())
+      return 0;
+    auto Cst = getIConstantVRegValWithLookThrough(DataReg, MRI);
+    return (Cst && Cst->Value.getZExtValue() == 0) ? DataType.getSizeInBytes()
```
martien-de-jong: There's APInt::IsZero()

llvm/lib/Target/AIE/AIECombinerHelper.cpp
andcarminati: I will include a test to see if it is empty.

llvm/lib/Target/AIE/AIECombinerHelper.cpp
```diff
@@ -4027,3 +4036,168 @@ bool llvm::matchPeelMemset(MachineInstr &MI, MachineRegisterInfo &MRI,
...
+  for (MachineInstr &MI : make_range(std::next(StMI.getIterator()),
+                                     StMI.getParent()->instr_end())) {
+    if (auto *CurrSt = dyn_cast<GStore>(&MI)) {
+      const unsigned ZeroBytes = GetZeroStoreSizeInBytes(*CurrSt);
+      if (!ZeroBytes) // Non-zero store.
+        break;
+
+      auto PtrAndOffset = getPtrAndConstantOffsetFromStore(CurrSt, MRI);
+      if (!PtrAndOffset) // Non-constant offset.
+        return false;
```
martien-de-jong: Aren't we wasting the accumulated MatchedSeqStores here?

llvm/lib/Target/AIE/AIECombinerHelper.cpp
```diff
@@ -4027,3 +4036,168 @@ bool llvm::matchPeelMemset(MachineInstr &MI, MachineRegisterInfo &MRI,
...
+  // have the first scalar store whose size is smaller then the alignment, by
+  // combining with the next, we have a chance of reducing the number of stores.
+  // For example:
+  //   STORE i8 0, [p0] (Align 16)
+  //   STORE i16 0, [p0+1]
+  //   STORE i8 0, [p0+3]
+  // Will be transformed to:
+  //   STORE i32 0, [p0] (Align 16)
+  if ((IsVectorAlignment && NumberOfBytes < MinVectorStoreSize) ||
+      MatchedSeqStores.empty())
```
martien-de-jong: nit: I think the empty() test should be hoisted to right after the collection loop. That separates the 'no opportunities found' from the 'not profitable' criteria.

llvm/lib/Target/AIE/AIECombinerHelper.cpp
andcarminati: Good catch!

llvm/lib/Target/AIE/AIECombinerHelper.cpp
martien-de-jong: I understand it, but I don't like it. The behaviour of the memset expander is encrypted here in a magic literal '3'. Also, I think we will not coalesce three byte stores under this condition?

llvm/lib/Target/AIE/AIECombinerHelper.cpp
```diff
@@ -4027,3 +4036,168 @@ bool llvm::matchPeelMemset(MachineInstr &MI, MachineRegisterInfo &MRI,
...
+
+      auto PtrAndOffset = getPtrAndConstantOffsetFromStore(CurrSt, MRI);
+      if (!PtrAndOffset) // Non-constant offset.
+        return false;
+
+      auto [CurrPtr, CurrOffset] = *PtrAndOffset;
+
+      // Pointers are different or we have non-linear store.
+      if ((CurrPtr != Ptr) || (ExpectedOffset != CurrOffset))
+        break;
```
martien-de-jong, line 4135: Would it be difficult to track a bitvector with written zero bytes on the same pointer so long as there's no overlap or other dangers and check for a consecutive range afterwards? That would take care of the strict ordering we now need.

llvm/lib/Target/AIE/AIECombinerHelper.cpp
andcarminati: Indeed I detest magic numbers, but we have a reason and a test for it. As far I remember, I tripped with this case in a real benchmark.

llvm/lib/Target/AIE/AIECombinerHelper.cpp
andcarminati: I would say that is this is feasible. We could have a follow-up work that handles gaps and non-zero values if we see practical cases.

llvm/lib/Target/AIE/AIECombinerHelper.cpp
andcarminati: Maybe it is not great, I tried to give a name to the alignment check and reuse `isAligned` function. About the rest of the implementation, but I don't thing that turning the caller code into a loop will help because, to align a memset, we should peel between one and two stores only, and each store is different (different MMO, type and constant to store).

llvm/lib/Target/AIE/AIECombinerHelper.cpp
```diff
@@ -4160,8 +4161,7 @@ bool llvm::matchSequentialStores(GStore &StMI, MachineRegisterInfo &MRI,
   //   STORE i8 0, [p0+3]
   // Will be transformed to:
   //   STORE i32 0, [p0] (Align 16)
-  if ((IsVectorAlignment && NumberOfBytes < MinVectorStoreSize) ||
-      MatchedSeqStores.empty())
+  if ((IsVectorAlignment && NumberOfBytes < MinVectorStoreSize))
```
martien-de-jong: nit: We can drop a pair of parentheses now.

llvm/lib/Target/AIE/AIECombinerHelper.cpp
martien-de-jong: I grant you the loop argument. But just have case 0, 1, 2, and 3 in a switch rather than repeated calls to matchAlignment would lead to simpler control flow. bottom line is that we don't want to know whether it's aligned, but in what way it is unaligned.

andcarminati: I would like to stick to the current approach, because I feel it simple and correct. Let's say, if we are not aligned, we just align on the fly base on the current offset. I also update that code with a fix and some comments.

llvm/lib/Target/AIE/AIECombinerHelper.cpp
```diff
@@ -3995,36 +3994,38 @@ bool llvm::matchPeelMemset(MachineInstr &MI, MachineRegisterInfo &MRI,
       return NewMMO;
     };
 
+    int64_t PeelOffset = Offset;
```
martien-de-jong, line 3993: Yes! PeelOffset represents an invariant of the process that fixes the alignment.

llvm/lib/Target/AIE/AIECombinerHelper.cpp
```diff
@@ -3995,36 +3994,38 @@ bool llvm::matchPeelMemset(MachineInstr &MI, MachineRegisterInfo &MRI,
...
+    const unsigned NewSize = Size - (PeelOffset - Offset);
 
     // No bytes left to memset.
     if (NewSize == 0) {
       MI.eraseFromParent();
       return;
     }
 
-    const int64_t MemsetOffset = Offset + PeeledStores;
+    const int64_t MemsetOffset = PeelOffset;
```
martien-de-jong, line 4024: Or even copy-propagate it to its use.

martien-de-jong (dismissed):
Good to go.

martien-de-jong (approved):

PR #613, opened 2025-08-12
[AIEX] Fully define s20 incoming Phi operands

andcarminati:
This combiner is necessary for AIESuperRegRewriter to work correctly.


llvm/lib/Target/AIE/AIECombinerHelper.cpp
```diff
@@ -3711,3 +3711,45 @@ bool llvm::matchPhiBitcast(MachineInstr &Phi, MachineRegisterInfo &MRI,
 
   return true;
 }
+
+/// This combiner is intended to fully define undef operands
+/// of s20 types in Phi instructions. The goal is to constrain
+/// the MIR to a fully well-defined shape that fits the requirements
```
konstantinschwarz, line 3717: Can you elaborate on the requirements here?
What's the issue that we are trying to solve here?

llvm/lib/Target/AIE/AIECombinerHelper.cpp
```diff
@@ -3711,3 +3711,45 @@ bool llvm::matchPhiBitcast(MachineInstr &Phi, MachineRegisterInfo &MRI,
...
+  MachineInstr *UndefMI = nullptr;
+  MachineOperand *UndefMO = nullptr;
+  for (unsigned I = 1; I < Phi.getNumOperands(); I += 2) {
+    MachineOperand &MO = Phi.getOperand(I);
+    const Register IncomingReg = MO.getReg();
+    MachineInstr *DefMI = MRI.getVRegDef(IncomingReg);
+    if (DefMI->getOpcode() == TargetOpcode::G_IMPLICIT_DEF) {
+      UndefMI = DefMI;
+      UndefMO = &MO;
+      break;
```
krishnamtibrewala, line 3746: CHECK : Can there be the case where more than one operand of PHI has G_IMPLICIT_DEF ?

llvm/lib/Target/AIE/AIECombinerHelper.cpp
```diff
@@ -3711,3 +3711,45 @@ bool llvm::matchPhiBitcast(MachineInstr &Phi, MachineRegisterInfo &MRI,
...
+    }
+  }
+
+  if (!UndefMI)
+    return false;
+
+  MatchInfo = [=, &Phi, &Observer](MachineIRBuilder &B) {
+    B.setInstrAndDebugLoc(*UndefMI);
+    const Register NewReg = B.buildConstant(S20, 0).getReg(0);
```
krishnamtibrewala, line 3755: CHECK : How do we know "0" is a safe/correct value to define the IMPLICIT_DEF reg with ?
Super Nit : can we have "`const int ImplicitDefInitValue = 0`" and use that with `buildConstant(...)`

llvm/lib/Target/AIE/AIECombinerHelper.cpp
andcarminati: More context offline.

llvm/lib/Target/AIE/AIECombinerHelper.cpp
andcarminati: Yes, and we don't need to care about this case because of the fixed-point nature of the combiner engine, so we can keep it simple.

llvm/lib/Target/AIE/AIECombinerHelper.cpp
andcarminati: Actually, we can use anything to represent IMPLICIT_DEF, it basically says that we don't care. Also, `0` is more likely to be CSEable because counters are also usually defined with this value.

llvm/lib/Target/AIE/AIECombinerHelper.cpp
martien-de-jong: compiler performance nit: It's not as if passing an array of undefs to MatchInfo is much of a complication. 
readability nit: the question shows that you need context knowledge. Dealing with all operands would make it 'locally complete'

llvm/lib/Target/AIE/AIECombinerHelper.cpp
martien-de-jong: Also, zero is statistically more likely to be the value of the register after power-up/reset. Although an impure argument, people never seem to willingly accept 'undefined behaviour' as a valid reason to 'break their code' without longwinded discussions.

martien-de-jong (dismissed):

martien-de-jong (dismissed):

martien-de-jong (approved):

PR #614, opened 2025-08-12
Isoard.upstream sync

isoard-amd:
@konstantinschwarz Note `tbaa` is now enabled by default. I do not know if it may cause issues for us (hardware people tend to abuse pointer casting).


isoard-amd:
@konstantinschwarz It is ready for merging.

PR #615, opened 2025-08-12
Isoard.upstream sync

isoard-amd:
Note that for the conflict resolution in `.github/workflows/build-ci-container.yml` I used upstream version, which leads to this file now being identical to upstream.


isoard-amd:
@konstantinschwarz seems the github workflow is still working, good!

konstantinschwarz:
> Note that for the conflict resolution in `.github/workflows/build-ci-container.yml` I used upstream version, which leads to this file now being identical to upstream.

FYI @philippjh. In https://github.com/Xilinx/llvm-aie/commit/9b6736eb5b6cf016fe41f3eec96cbc631194e483 we pinned the action version. Upstream modified the workflow file, and we now use the latest version of the action, same as upstream

PR #621, opened 2025-08-14
Isoard.upstream sync

isoard-amd:


llvm/lib/Target/AIE/AIEBasePipelinerLoopInfo.h
```diff
@@ -102,7 +103,7 @@ class AIEBasePipelinerLoopInfo : public TargetInstrInfo::PipelinerLoopInfo {
   ///
   /// Once this function is called, no other functions on this object are
   /// valid; the loop has been removed.
-  void disposed() override;
+  void disposed(LiveIntervals *LIS) override;
```
konstantinschwarz: Nit: forward declare `class LiverIntervals;` in the header and add the include in the `.cpp`?

isoard-amd: Actually they now provide a default empty implementation. I should just remove it.

isoard-amd:
@konstantinschwarz It is ready to be merged

PR #622, opened 2025-08-15
[AIE2P] Add support for Vector G_ICMP

andcarminati:
Starting from `G_BITCAST`, this PR adds support for vector comparison using directly generic opcodes. The main goal is to fix some compilation failures related to the following construct:
```
    %0:_(<32 x s16>) = COPY $x0
    %1:_(<32 x s16>) = COPY $x1
    %2:_(<32 x s1>) = G_ICMP intpred(ne), %0, %1
```
However, it looks a good moment (for completeness) to also implement G_ICMP support for all native vector integer types (FP ones are not covered in this PR, but they should follow similar approaches). I need to advise that this PR contains some points that can be spotted as quirks:

1. `%2:_(<64 x s1>) = G_ICMP intpred(ne), %0, %1` this case was simple to implement using our current GISel model, directing the output directly to a large scalar register.
2. This case:
```
    %0:_(<16 x s32>) = COPY $x0
    %1:_(<16 x s32>) = COPY $x1
    %2:_(<16 x s1>) = G_ICMP intpred(ne), %0, %1
```
Required a bit o engineering related to the `<16 x s1>` type. As as definition of the `G_ICMP` instruction, we should have a resulting bit for each vector element. In this way, comparing `<16 x s32>` vectors should always produce result as `<16 x s1>`, and, if we try to change this relation, `MachineVerifier` will complain. With this in mind, the only alternative is to push the related 16-by type to the legalizer and beyond, handling it in some parts of `RegBankSelect`/`RegInfo`.

The resulting `<16 x s32>` may need additional legalization for the surrounding 16-element boolean vector. We actually support vector concatenation be able to test the implementation. What is required for this additional implementation is not know because we don't have C++ code using it in this moment.

UPDATE: I changed the implementation to use a AIE-specific opcode `G_AIE_VECTOR_ICMP`, allowing 32-bit results for `<16 x s32>` comparisons.


llvm/lib/Target/AIE/aie2p/AIE2PLegalizerInfo.cpp
```diff
@@ -26,10 +26,12 @@ using namespace LegalityPredicates;
 static LegalityPredicate isLegalBitCastType(unsigned TypeIdx) {
   return [=](const LegalityQuery &Query) {
     LLT Ty = Query.Types[TypeIdx];
+    const unsigned TypeSize = Ty.getSizeInBits();
     if (Ty.isScalar())
-      return Ty.getSizeInBits() >= 32;
-    const int EltSize = Ty.isVector() ? Ty.getElementType().getSizeInBits() : 0;
-    return EltSize >= 8 && isPowerOf2_32(EltSize);
+      return TypeSize >= 32;
+    const unsigned EltSize = Ty.getElementType().getSizeInBits();
```
martien-de-jong, line 32: do all types that are not scalar have an elementtype?

llvm/lib/Target/AIE/aie2p/AIE2PLegalizerInfo.cpp
```diff
@@ -26,10 +26,12 @@ using namespace LegalityPredicates;
...
+    return TypeSize == 32 || TypeSize == 64 ||
+           (EltSize >= 8 && isPowerOf2_32(EltSize));
```
martien-de-jong, line 34: CHECK: we delay introduction of vnxs1 to a later commit.

llvm/lib/Target/AIE/aie2p/AIE2PRegisterInfo.td
```diff
@@ -113,7 +113,7 @@ class AIE2PRegisterClass <int size, int align, list<ValueType> regTypes, dag reg
 }
 
   class AIE2PScalarRegisterClass<dag reglist, RegAltNameIndex idx = NoRegAltName>
-      : AIE2PRegisterClass<32, 32, [v8i1, v16i1, bf16, i20, v2bf16, i32, f32, v4i8, v2i16], reglist,
+      : AIE2PRegisterClass<32, 32, [v8i1, v16i1, bf16, i20, v2bf16, i32, f32, v4i8, v2i16, v32i1, v16i1], reglist,
```
martien-de-jong: nit: v16i1 mentioned twice. Also, move vni1 together?

llvm/lib/Target/AIE/aie2p/AIE2PLegalizerInfo.cpp
```diff
@@ -105,17 +105,22 @@ static const LLT S256 = LLT::scalar(256);
...
 static const LLT S1024 = LLT::scalar(1024);
 
+// 16-bit vectors
+static const LLT V16S1 = LLT::fixed_vector(16, 1);
+
 // 32-bit vectors
 static const LLT V2S8 = LLT::fixed_vector(2, 8);
 
 // 32-bit vectors
+static const LLT V32S1 = LLT::fixed_vector(32, 1);
```
martien-de-jong, line 113: nit: I think we could do bitwise operations on V16S2, V8S4 too.

llvm/lib/Target/AIE/AIEBaseRegisterBankInfo.cpp
```diff
@@ -263,6 +263,8 @@ AIEBaseRegisterBankInfo::getPartialMappingIdx(const LLT &Ty) const {
   } else {
     unsigned Size = ElemTy.getSizeInBits();
     switch (Ty.getSizeInBits()) {
+    case 16:
+      return PMI_GPR;
```
martien-de-jong: Perhaps check Size == 1 for confidence?

llvm/lib/Target/AIE/aie2p/AIE2PInstructionSelector.cpp
```diff
@@ -563,6 +563,18 @@ bool AIE2PInstructionSelector::selectG_CONCAT_VECTORS(
...
+  // 16-bit to 32-bit vector concat
+  if (DstTy.getSizeInBits() == 32 && Src0Ty.getSizeInBits() == 16) {
+    auto ShiftAmt =
+        MIB.buildInstr(AIE2P::MOV_RLC_imm11_pseudo, {&AIE2P::eRRegClass}, {})
+            .addImm(-16);
+    auto Shift =
+        MIB.buildInstr(AIE2P::LSHL, {&AIE2P::eRRegClass}, {Src0, ShiftAmt});
+    auto Or = MIB.buildInstr(AIE2P::OR, {Dst}, {Shift, Src1});
+    I.eraseFromParent();
+    return constrainSelectedInstRegOperands(*Or, TII, TRI, RBI);
```
martien-de-jong: CHECK: CONCAT takes vectors, so we know that the upper bits are zero.

martien-de-jong: More generally, every 16 bit (or 8 for that matter) type that makes it through legalization must be a vector

llvm/test/CodeGen/AIE/aie2p/GlobalIsel/inst-select-concat-vectors.mir
```diff
@@ -124,3 +124,26 @@ body:             |
...
+    ; CHECK-NEXT: [[DEF:%[0-9]+]]:er = IMPLICIT_DEF
+    ; CHECK-NEXT: [[DEF1:%[0-9]+]]:er = IMPLICIT_DEF
+    ; CHECK-NEXT: [[MOV_RLC_imm11_pseudo:%[0-9]+]]:er = MOV_RLC_imm11_pseudo -16
+    ; CHECK-NEXT: [[LSHL:%[0-9]+]]:er = LSHL [[DEF]], [[MOV_RLC_imm11_pseudo]]
+    ; CHECK-NEXT: [[OR:%[0-9]+]]:er = OR [[LSHL]], [[DEF1]]
+    ; CHECK-NEXT: PseudoRET implicit $lr, implicit [[OR]]
+    %0:gprregbank(<16 x s1>) = G_IMPLICIT_DEF
+    %1:gprregbank(<16 x s1>) = G_IMPLICIT_DEF
+    %2:gprregbank(<32 x s1>) = G_CONCAT_VECTORS %0(<16 x s1>), %1(<16 x s1>)
+    PseudoRET implicit $lr, implicit %2
```
martien-de-jong: I wouldn't mind allowing all 'small power of two' vector types that would fit in gprs.

llvm/lib/Target/AIE/aie2p/AIE2PInstrPatterns.td
```diff
@@ -1085,6 +1085,81 @@ def : Pat<(int_aie2p_vltbf16 VEC512:$s1, VEC512:$s2),
 def : Pat<(int_aie2p_vgebf16 VEC512:$s1, VEC512:$s2),
             (VGE_bf16 VEC512:$s1, VEC512:$s2)>;
 
+// direct comparisons without instrinsics.
```
martien-de-jong: nit: intrinsics

llvm/lib/Target/AIE/aie2p/AIE2PInstrPatterns.td
```diff
@@ -1085,6 +1085,81 @@ def : Pat<(int_aie2p_vltbf16 VEC512:$s1, VEC512:$s2),
...
+// LT
+def : Pat<(v64i1 (setlt (v64i8 VEC512:$rs1), (v64i8 VEC512:$rs2))),
+	       (VLT_8_vaddSign1 VEC512:$rs1, VEC512:$rs2)>;
+def : Pat<(v32i1 (setlt (v32i16 VEC512:$rs1), (v32i16 VEC512:$rs2))),
+	       (VLT_16_vaddSign1 VEC512:$rs1, VEC512:$rs2)>;
+def : Pat<(v16i1 (setlt (v16i32 VEC512:$rs1), (v16i32 VEC512:$rs2))),
+	       (VLT_32_vaddSign1 VEC512:$rs1, VEC512:$rs2)>;
```
martien-de-jong, line 1124: do we also have setult etc for unsigned?

llvm/lib/Target/AIE/aie2p/AIE2PInstrPatterns.td
```diff
@@ -1085,6 +1085,81 @@ def : Pat<(int_aie2p_vltbf16 VEC512:$s1, VEC512:$s2),
...
+
+// NE
+def  : Pat<(v64i1 (setne (v64i8 VEC512:$rs1), (v64i8 VEC512:$rs2))),
+          (REG_SEQUENCE eL,
+             (XOR (EXTRACT_SUBREG
+                (VEQZ_8 (VSUB_8 VEC512:$rs1, VEC512:$rs2)), sub_l_even),
+                (MOV_RLC_imm11_pseudo (i32 -1))), sub_l_even,
+              (XOR (EXTRACT_SUBREG
+                (VEQZ_8 (VSUB_8 VEC512:$rs1, VEC512:$rs2)), sub_l_odd),
+                (MOV_RLC_imm11_pseudo (i32 -1))), sub_l_odd)>;
```
martien-de-jong, line 1232: Would there be a way to template those patterns, to make them reusable for other archs?

llvm/lib/Target/AIE/aie2p/AIE2PInstrPatterns.td
martien-de-jong: Could we start eliminating builtins by expressing them as operators? We may be freeloading on generic vector optimizers then.

martien-de-jong (commented):
Looks good. I would really want to expand the glue logic like bitcast and regsequence to all vector types that fit in GPRs,  You'll  get bitwise operators for free, and using some carry breaking schemes you could do add/sub as well.

konstantinschwarz:
Instead of making types smaller than our native register sizes legal, I'd prefer to custom legalize the `G_ICMP` to a AIE specific generic opcode that always produces a 32-bit result (`<32 x s1>`). This would be similar to our `G_AIE_[ZS]EXT_EXTRACT_VECTOR_ELT` opcode, which always extracts a 32-bit scalar, and then uses `G_TRUNC` to produce the original type (plus a `G_ASSERT_[ZS]EXT` to allow combining away redundant truncs+ext pairs).
What do you think?

llvm/lib/Target/AIE/AIEBaseRegisterBankInfo.cpp
andcarminati: Removed as part of the new version.

llvm/lib/Target/AIE/aie2p/AIE2PInstrPatterns.td
andcarminati: That is the goal, but now we have a G_AIE_VECTOR_ICMP.

andcarminati:
> Instead of making types smaller than our native register sizes legal, I'd prefer to custom legalize the `G_ICMP` to a AIE specific generic opcode that always produces a 32-bit result (`<32 x s1>`). This would be similar to our `G_AIE_[ZS]EXT_EXTRACT_VECTOR_ELT` opcode, which always extracts a 32-bit scalar, and then uses `G_TRUNC` to produce the original type (plus a `G_ASSERT_[ZS]EXT` to allow combining away redundant truncs+ext pairs). What do you think?

I updated to this approach, it was a bit tricky because of a special handling of `G_ICMP` related to table-gen's GlobalSel emitter.

llvm/lib/Target/AIE/AIEInstrGISel.td
```diff
@@ -140,3 +140,10 @@ def G_AIE_SHUFFLE_VECTOR : AIEGenericInstruction {
   let InOperandList = (ins type0:$src0, type0:$src1, type1:$mode);
   let hasSideEffects = false;
 }
+
+// AIE integer comparison.
+def G_AIE_VECTOR_ICMP : AIEGenericInstruction {
+  let OutOperandList = (outs type0:$dst);
+  let InOperandList = (ins unknown:$tst, type1:$src1, type1:$src2);
+  let hasSideEffects = false;
+}
```
andcarminati: nit: new line.

llvm/lib/Target/AIE/aie2p/AIE2PInstrPatterns.td
andcarminati: I really don't know, It would be nice!

llvm/lib/Target/AIE/aie2p/AIE2PLegalizerInfo.cpp
andcarminati: We have only vectors reaching this point. We cannot bitcast pointers to other types.

llvm/lib/Target/AIE/aie2p/AIE2PLegalizerInfo.cpp
martien-de-jong: We could put them under the common `32-bit vectors` comment

llvm/lib/Target/AIE/aie2p/AIE2PLegalizerInfo.cpp
```diff
@@ -254,9 +260,11 @@ AIE2PLegalizerInfo::AIE2PLegalizerInfo(const AIE2PSubtarget &ST)
   // Since the only integers smaller than 32 bits we produce are S20 (from
   // G_PTRTOINT), the only legal extension is S20 -> S32.
   // Extensions to types larger than 64 bits have to be broken down into
-  // multiple parts.
+  // multiple parts. We also explicitly mark S16 -> S32 as artificially
+  // legal because of G_ICMP, but this operation is removed by
+  // postlegalizer combiner.
   getActionDefinitionsBuilder({G_ANYEXT, G_SEXT, G_ZEXT})
-      .legalFor({{S32, S20}})
+      .legalFor({{S32, S20}, {S32, S16}})
```
martien-de-jong, line 265: CHECK: ANYEXT and SEXT don't preserve the number of bits that are set, but they do preserve anyof() and allof().
But since we are dedicating this to boolean vectors, could we only make it legal for ZEXT?

llvm/utils/TableGen/GlobalISelEmitter.cpp
```diff
@@ -865,10 +867,18 @@ Expected<InstructionMatcher &> GlobalISelEmitter::createAndImportSelDAGMatcher(
     // predicate operand needs to be swapped from the last operand to the first
     // source.
 
+    // This allows targets to build custom FCMP/ICMP variants to support
+    // distinguished type restrictions.
```
martien-de-jong: The text doesn't connect well. Perhaps explain that we interpret any operator ending in "FCMP" and "ICMP" as a comparison operator. Also, there's std::string::ends_with().

martien-de-jong: perhaps include the "_" into the match, i.e. match _ICMP?

martien-de-jong (approved):

PR #623, opened 2025-08-15
Isoard.upstream sync

isoard-amd:


isoard-amd:
@konstantinschwarz It is ready

PR #626, opened 2025-08-15
Isoard.upstream sync

isoard-amd:
Because upstream is now using `m` for FP8, we had to change the shorthand for AIE specific built-in types.

In particular, the mapping is now as follows:
- `__acc32` is `n` (as before)
- `__acc48` is `Ln` (was `e` before)
- `__acc64` is `LLn` (was `m` before)
- `__accfloat` is `g` (as before)

This means the `BuiltinsAIE.def` and `BuiltinsAIE2P.def` had to be updated.

Using the following `sed` expression should do the trick:

    sed ':e;/BUILTIN(.*)/s/\("[^"]*\)e\([^"]*"\)/\1Ln\2/g;te;:m;/BUILTIN(.*)/s/\("[^"]*\)m\([^"]*"\)/\1LLn\2/g;tm' -i clang/include/clang/Basic/BuiltinsAIE*.def


isoard-amd:
@konstantinschwarz It is ready for merging

PR #634, opened 2025-08-25
[AIEX] fix spillCalleeSavedRegisters crash

F-Stuckmann:
Fixes a crash `spillCalleeSavedRegisters` , when the `DebugLocation` cannot be found (i.e. due to an empty MBB)

### DIFF
--- a/llvm/lib/Target/AIE/AIEBaseFrameLowering.cpp
+++ b/llvm/lib/Target/AIE/AIEBaseFrameLowering.cpp
@@ -219,8 +219,10 @@ bool AIEBaseFrameLowering::spillCalleeSavedRegisters(
     ArrayRef<CalleeSavedInfo> CSI, const TargetRegisterInfo *TRI) const {
   MachineFunction *MF = MBB.getParent();
   auto *TII = static_cast<const AIEBaseInstrInfo *>(STI.getInstrInfo());
-  MachineBasicBlock::iterator MBBI = MBB.getLastNonDebugInstr();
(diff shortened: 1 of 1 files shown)


llvm/lib/Target/AIE/AIEBaseFrameLowering.cpp
```diff
@@ -219,8 +219,10 @@ bool AIEBaseFrameLowering::spillCalleeSavedRegisters(
     ArrayRef<CalleeSavedInfo> CSI, const TargetRegisterInfo *TRI) const {
   MachineFunction *MF = MBB.getParent();
   auto *TII = static_cast<const AIEBaseInstrInfo *>(STI.getInstrInfo());
-  MachineBasicBlock::iterator MBBI = MBB.getLastNonDebugInstr();
-  DebugLoc DL = MBBI->getDebugLoc();
+  DebugLoc DL;
+  if (auto MBBI = MBB.getLastNonDebugInstr(); MBBI != MBB.end())
+    DL = MBBI->getDebugLoc(); // MBBI is valid
+
```
SagarMaheshwari99, line 225: And I guess we are fine to proceed with empty DebugLoc?

SagarMaheshwari99 (approved):

llvm/lib/Target/AIE/AIEBaseFrameLowering.cpp
F-Stuckmann: Yes

PR #635, opened 2025-09-03
[AIE2P] Remove Size Constraints for PostInc Combiner

F-Stuckmann:
Remove Memory Size constraints for PostIncrement Selection in AIE2P. 

There are no regressions in performance or PM in QoR.

### DIFF
--- a/llvm/lib/Target/AIE/aie2p/AIE2PInstrInfo.cpp
+++ b/llvm/lib/Target/AIE/aie2p/AIE2PInstrInfo.cpp
@@ -297,8 +297,6 @@ std::optional<unsigned> AIE2PInstrInfo::getCombinedPostIncOpcode(
     MachineInstr &BaseMemI, MachineInstr &PostIncI, TypeSize Size) const {
   switch (PostIncI.getOpcode()) {
   case TargetOpcode::G_PTR_ADD:
-    if (Size >= 2048)
-      return {};
     switch (BaseMemI.getOpcode()) {
     case TargetOpcode::G_STORE:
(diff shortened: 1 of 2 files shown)


andcarminati:
Could we have a test?

martien-de-jong:
I'm curious as to the rationale. Does it block a useful optimization?
I can also imagine that it is a first step to share all G_AIE opcodes and bring more generic functionality to AIEBaseInstrInfo.cpp. Which I would applaud.

llvm/lib/Target/AIE/aie2p/AIE2PInstrInfo.cpp
```diff
@@ -313,8 +311,6 @@ std::optional<unsigned> AIE2PInstrInfo::getCombinedPostIncOpcode(
   case TargetOpcode::G_INTRINSIC:
     switch (cast<GIntrinsic>(PostIncI).getIntrinsicID()) {
     case Intrinsic::aie2p_add_2d:
-      if (Size >= 1024)
-        return {};
```
martien-de-jong, line 317: Also weird that this wasn't `Size > 512`

martien-de-jong (approved):

F-Stuckmann:
> I'm curious as to the rationale. Does it block a useful optimization? I can also imagine that it is a first step to share all G_AIE opcodes and bring more generic functionality to AIEBaseInstrInfo.cpp. Which I would applaud.

GEMM now performs vloads of 2048 bits and thus this optimization enables postincrements on these loads.

PR #636, opened 2025-09-03
[AIEX] Fix top-down scoreboard initialization

andcarminati:
Top-down scheduling is used only for Epilogue merging, and it takes place only in the first region of a MBB. This PR initializes the top-down scoreboard just when it is really needed, avoiding wrong schedules for multiregion in presence of and Post-SWP epilogue.


llvm/lib/Target/AIE/AIEMachineScheduler.cpp
```diff
@@ -355,6 +355,13 @@ void AIEPostRASchedStrategy::initializeBotScoreBoard(ScoreboardTrust Trust) {
 
 void AIEPostRASchedStrategy::initializeTopScoreBoard() {
 
+  // Top Scoreboard is only used for Post-SWP epilogue merging.
+  // If we have a multiregion MBB, the epilogue will be scheduled
+  // in the first topological region. In this case, ignore scoreboard
+  // initialization for other regions.
+  if (DAG->begin() != CurMBB->begin())
```
F-Stuckmann: nit: use a constant with a nice name to indicate, that we are checking, if we are not the first Region.

andcarminati: Done!

F-Stuckmann (approved):

PR #637, opened 2025-09-03
Rewrites reshapes that are adding 1-sized dims and are followed by a tile on the one-sized dim to a tile on the original shape followed by a reshape.

jorickert:
This is done to reduce the rank of tile ops to make downstream consumption easier

### DIFF
--- a/mlir/lib/Dialect/Tosa/IR/TosaCanonicalizations.cpp
+++ b/mlir/lib/Dialect/Tosa/IR/TosaCanonicalizations.cpp
@@ -109,6 +109,102 @@ void ConcatOp::getCanonicalizationPatterns(RewritePatternSet &results,
   results.add<SelfConcatToTile>(context);
 }
 
+/* Rewrites reshapes that are adding 1-sized dims and are followed by a tile on
+ * the one-sized dim to a tile on the original shape followed by a reshape. This
+ * is done to reduce the rank of tile ops. */
+struct TileOnOneSizedDim : public OpRewritePattern<tosa::TileOp> {
+  using OpRewritePattern<tosa::TileOp>::OpRewritePattern;
+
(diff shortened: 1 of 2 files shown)


mlir/lib/Dialect/Tosa/IR/TosaCanonicalizations.cpp
```diff
@@ -109,6 +109,102 @@ void ConcatOp::getCanonicalizationPatterns(RewritePatternSet &results,
   results.add<SelfConcatToTile>(context);
 }
 
+/* Rewrites reshapes that are adding 1-sized dims and are followed by a tile on
+ * the one-sized dim to a tile on the original shape followed by a reshape. This
+ * is done to reduce the rank of tile ops. */
```
ljfitz, line 114: Is the core canonicalization to minimize the dimensions needed to express the `tile` operation? If so it would be better to do so regardless of a directly connected reshape or not and instead insert `reshape` before and after as appropriate.

If this is not the core goal then I question having this as a canonicalization as matching tile will still have a guarantee that the tile op is in canonicalized form.

jorickert: I do not think its a good idea to always do the  tile rank reduction, as the inserted reshapes may prevent other patterns from matching.

I originally wrote this to target a specific case:

```
reshape 1x2x3 to 1x2x1x3
tile 1x2x1x3 to 1x2x4x3
reshape to 1x2x12

this could be canonicalized to :
tile 1x2x3 to 1x2x12
```
In this case the rank of the tile would be reduced and the reshapes completely canceled.

I marked this Pr as draft, as I think it generally needs more investigation when it makes sense to do this and when not

PR #643, opened 2025-09-09
Add constant folding for f16 erf and sqrt

jorickert:


ttjost (dismissed):
Are we doing this for only these ops? Or other ops already supported it?

jorickert:
> Are we doing this for only these ops? Or other ops already supported it?

Yes. There are a few places that use an optimization for types that match a native C++ types bitwidth, I will apply the same optimization we do for bf16 for f16 too

mlir/lib/Dialect/Tosa/Transforms/TosaFolders.cpp
```diff
@@ -308,7 +311,7 @@ DenseElementsAttr transpose(DenseElementsAttr attr, ShapedType inputType,
   if (baseType.isF64()) {
     return transposeTypeRaw<uint64_t>(attr, inputType, outputType, permValues);
   }
-  if (baseType.isBF16()) {
+  if (baseType.isBF16() || baseType.isBF16()) {
```
ttjost: 😆 ?
```suggestion
  if (baseType.isBF16() || baseType.isF16()) {
```

ttjost (dismissed):

konstantinschwarz (approved):

PR #577, opened 2025-07-28
[AIEX] Detach Loads from address chaining

F-Stuckmann:
Detach loads from address chain, so that they can be scheduled as early as possible and do not delay branch decision.

PR #646, opened 2025-09-10
[AIEX] Address chaining improvements: initial integration

andcarminati:
The goal of this PR is the integration of the first part of https://github.com/Xilinx/llvm-aie/pull/577 from @F-Stuckmann .

Parts:

- Code outlining.
- Initial tests.
- Unrestricted address chaining for code reduction.

This will also help the review of the previously mentioned PT.

The QoR numbers are under preparation.


andcarminati:
QoR (Absolute PM decrease > Absolute Stack increase):
Core_PMSize
Core_PMSize_absolute
Core_StackSize
Core_StackSize_absolute

llvm/lib/Target/AIE/AIEClusterBaseAddress.cpp
```diff
@@ -175,176 +149,235 @@ class AIEClusterBaseAddress : public MachineFunctionPass {
...
+  // be too pessimistic as to not consider uses in the current basic block.
+  ReachableMBBs.erase(MBB);
+  return ReachableMBBs;
+}
+
+bool AIEClusterBaseAddress::isRegUsedInSuccessiveMBBs(MachineBasicBlock *MBB,
+                                                      Register Reg) {
+  std::set<MachineBasicBlock *> ReachableMBBs = findReachableMBBs(MBB);
+  for (MachineInstr &Use : MRI->use_nodbg_instructions(Reg)) {
+    if (ReachableMBBs.count(Use.getParent()))
```
martien-de-jong, line 386: I think this is equivalent to Use.getParent() != MBB. Uses are dominated by the def, and therefore reachable.

martien-de-jong (approved):
I suggest to clean up my few nits in the follow up PR that implements the pointer restore.

llvm/lib/Target/AIE/AIEClusterBaseAddress.cpp
```diff
@@ -273,6 +277,10 @@ bool AIEClusterBaseAddress::shouldSkipChaining(
   if (Instrs.size() <= 1)
     return true;
 
+  // Chain MBB regardless.
+  if (EnableChainsAcrossMultiBlocks)
+    return false;
```
martien-de-jong, line 282: I would prefer only to have return true, except for the final one.

andcarminati:
FYI @F-Stuckmann.

PR #647, opened 2025-09-10
Martien.ghissue 641

martien-de-jong:
issue-641 had a runtime library call in a hardware loop, which very likely caused the output data mismatch. This PR fixes that error. 
The problem is that the TargetTransformInfo is a redundant coding with the legalizer in this respect, and that it is shared between AIE2 and AIE2P, whereas the two legalizers diverged on legalizing a scalar fmul.

### DIFF
--- a/llvm/lib/Target/AIE/AIE2TargetTransformInfo.cpp
+++ b/llvm/lib/Target/AIE/AIE2TargetTransformInfo.cpp
@@ -31,6 +31,14 @@ instCombineDemandedBits(InstCombiner &IC, IntrinsicInst &II, unsigned numBits) {
   return std::nullopt;
 }
 
+bool AIE2TTICommon::isAllowedInZOL(Instruction &I) {
+  // FMul is legalized to a VMUL for bfloat16 in other targets
+  if (I.getOpcode() == Instruction::FMul) {
+    return false;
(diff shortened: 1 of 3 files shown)


llvm/lib/Target/AIE/AIE2TargetTransformInfo.cpp
```diff
@@ -31,6 +31,14 @@ instCombineDemandedBits(InstCombiner &IC, IntrinsicInst &II, unsigned numBits) {
   return std::nullopt;
 }
 
+bool AIE2TTICommon::isAllowedInZOL(Instruction &I) {
+  // FMul is legalized to a VMUL for bfp16 in other targets
```
khallouh: bfp16 or bfloat16?

andcarminati (dismissed):
LGTM.

khallouh (dismissed):
LGTM

llvm/lib/Target/AIE/AIE2TargetTransformInfo.cpp
martien-de-jong: You're right. Probably bfloat16.

andcarminati (approved):
LGTM.

PR #648, opened 2025-09-10
[AIEX][AIE2P] Improve error handling in decoder methods

mludevid:
This PR changes two things:
- Some decoder tables have "holes" (zeros). If someone tries to decode an incorrect object file and the lookup table returns such a zero, the decoder method asserts. I updated them to fail properly.
- I updated the out-of-bound checks for the lookup tables. Since we know the number of bits that are read in the decoder methods we can check the bounds at compile time. A few decoder tables with non-power-of-two table sizes needed to be padded with zeroes to make this work. The previously introduced check catches these cases and fails properly as it did before these changes.

### DIFF
--- a/llvm/lib/Target/AIE/Disassembler/AIE2PGenDecoderMethods.inc
+++ b/llvm/lib/Target/AIE/Disassembler/AIE2PGenDecoderMethods.inc
@@ -15,40 +15,45 @@ static DecodeStatus DecodeDms_lda_2D(MCInst &MI, InsnType &Insn,
                                      uint64_t Address,
                                      const MCDisassembler *Decoder) {
   // ------------------- Outs ------------------------
-  unsigned dstRegRaw = fieldFromInstruction(Insn, 4, 7);
-  if (dstRegRaw >= std::size(mLdaSclDecoderTable))
-    return MCDisassembler::Fail;
+  const unsigned dstRegRawNumBits = 7;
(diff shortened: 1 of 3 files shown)


llvm/lib/Target/AIE/Disassembler/AIE2PGenDecoderMethods.inc
```diff
@@ -15,40 +15,45 @@ static DecodeStatus DecodeDms_lda_2D(MCInst &MI, InsnType &Insn,
                                      uint64_t Address,
                                      const MCDisassembler *Decoder) {
   // ------------------- Outs ------------------------
-  unsigned dstRegRaw = fieldFromInstruction(Insn, 4, 7);
-  if (dstRegRaw >= std::size(mLdaSclDecoderTable))
-    return MCDisassembler::Fail;
+  const unsigned dstRegRawNumBits = 7;
+  unsigned dstRegRaw = fieldFromInstruction(Insn, 4, dstRegRawNumBits);
+  static_assert(1 << dstRegRawNumBits == std::size(mLdaSclDecoderTable));
```
khallouh: Nit: enclose `1 << dstRegRawNumBits` in parentheses for clarity

khallouh (dismissed):
LGTM. It would have been nice to have a test where we fail properly

khallouh (approved):

PR #650, opened 2025-09-11
feat(tosa): Pad->Pad optimization

ttjost:


mlir/lib/Dialect/Tosa/IR/TosaCanonicalizations.cpp
```diff
@@ -551,9 +551,82 @@ struct MaterializePadValue : public OpRewritePattern<tosa::PadOp> {
...
+    ElementsAttr prevPaddingElems;
+    if (!matchPattern(prevPadOp.getPadding(), m_Constant(&prevPaddingElems))) {
+      return rewriter.notifyMatchFailure(
+          padOp, "padding must be a static shape value");
+    }
+    if (!paddingElems || !prevPaddingElems) {
+      return rewriter.notifyMatchFailure(
+          padOp,
+          "needs constant padding values to perform pad+pad optimization");
+    }
```
roberteg16, line 590: Isn't this one unnecessary? You already checked both before, right?

mlir/lib/Dialect/Tosa/IR/TosaCanonicalizations.cpp
```diff
@@ -551,9 +551,82 @@ struct MaterializePadValue : public OpRewritePattern<tosa::PadOp> {
...
+      paddingElems = DenseElementsAttr::get(
+          paddingElems.getShapedType(), paddingElems.getSplatValue<APInt>());
+    }
+    if (prevPaddingElems.isSplat()) {
+      prevPaddingElems =
+          DenseElementsAttr::get(prevPaddingElems.getShapedType(),
+                                 prevPaddingElems.getSplatValue<APInt>());
+    }
+
+    llvm::SmallVector<int64_t> paddingVals;
```
roberteg16, line 602: nit:
``` suggestion
    SmallVector<int64_t> paddingVals;
```

mlir/lib/Dialect/Tosa/IR/TosaCanonicalizations.cpp
```diff
@@ -551,9 +551,82 @@ struct MaterializePadValue : public OpRewritePattern<tosa::PadOp> {
...
+        FusedLoc::get(padOp->getContext(), {prevPadOp.getPadding().getLoc(),
+                                            padOp.getPadding().getLoc()}),
+        paddingVals);
+
+    auto newPadOp = rewriter.create<tosa::PadOp>(
+        FusedLoc::get(padOp->getContext(),
+                      {prevPadOp.getLoc(), padOp.getLoc()}),
+        padOp.getType(),
+        ValueRange{prevPadOp.getInput1(), padConst, padOp.getPadConst()},
+        padOp->getAttrs());
```
roberteg16, line 621: nit: should we merge both padOp->getAttrs() and prevPadOp->getAttrs()

roberteg16 (approved):
Very clean, thanks

mlir/lib/Dialect/Tosa/IR/TosaCanonicalizations.cpp
ttjost: I am thinking this can be a bit dangerous as we could have duplicated attributes and this would cause failures. usually, I think there shouldn't even be attributes attached to them, and I am only thinking about when we lower from onnx if we have `onnx_node_name`.

PR #657, opened 2025-09-18
[AIEX] Enable LoadStoreVectorizer 

andcarminati:
To merge part-word stores (to s32).


andcarminati:
QoR:
Core_Insn_Count
Core_PMSize
Core_PMSize_absolute

llvm/lib/Target/AIE/AIEBaseTargetTransformInfo.cpp
```diff
@@ -251,3 +251,30 @@ bool AIETTICommon::isProfitableOuterLSR(const Loop &L) const {
   return ConsiderLSROuterLoops.getNumOccurrences() > 0 ? ConsiderLSROuterLoops
                                                        : true;
 }
+
+unsigned AIETTICommon::getStoreVectorFactor(unsigned VF, unsigned StoreSize,
+                                            unsigned ChainSizeInBytes,
+                                            VectorType *VecTy) const {
+  // For StoreSize > 16, don't vectorize.
+  if (StoreSize > 16)
```
F-Stuckmann: shouldn't this also be dependent on alignment if StoreSize > 32

llvm/lib/Target/AIE/aie2p/AIE2PTargetTransformInfo.cpp
```diff
@@ -128,3 +128,22 @@ bool AIE2PTTIImpl::isHardwareLoopProfitable(Loop *L, ScalarEvolution &SE,
 bool AIE2PTTIImpl::isProfitableOuterLSR(const Loop &L) const {
   return Common.isProfitableOuterLSR(L);
 }
+
+unsigned AIE2PTTIImpl::getStoreVectorFactor(unsigned VF, unsigned StoreSize,
```
F-Stuckmann: shouldn't these hooks be implemented in `AIEBaseTargetTransformInfo.cpp`?

andcarminati: They are. See the Common class.

llvm/lib/Target/AIE/AIEBaseTargetTransformInfo.cpp
andcarminati: If StoreSize > 32 we avoid vectorization by returning 1. VectorFactor is the number of elements of StoreSize that can be placed in a single vector.

andcarminati: We are interested only in 8 and 16bit due to the scheduling memory cycles.

llvm/lib/Target/AIE/aie2p/AIE2PTargetTransformInfo.cpp
andcarminati: CRTP side effects.....

llvm/lib/Target/AIE/AIEBaseTargetTransformInfo.cpp
```diff
@@ -251,3 +251,30 @@ bool AIETTICommon::isProfitableOuterLSR(const Loop &L) const {
   return ConsiderLSROuterLoops.getNumOccurrences() > 0 ? ConsiderLSROuterLoops
                                                        : true;
 }
+
+unsigned AIETTICommon::getStoreVectorFactor(unsigned VF, unsigned StoreSize,
```
martien-de-jong: Any size not qualified by InBytes is in bits? I wouldn't mind a redundant InBits.

llvm/lib/Target/AIE/AIEBaseTargetTransformInfo.cpp
```diff
@@ -251,3 +251,30 @@ bool AIETTICommon::isProfitableOuterLSR(const Loop &L) const {
...
+                                            unsigned ChainSizeInBytes,
+                                            VectorType *VecTy) const {
+  // For StoreSize > 16, don't vectorize.
+  if (StoreSize > 16)
+    return 1;
```
martien-de-jong: nit: this check is redundant with the expression below.

llvm/lib/Target/AIE/AIEBaseTargetTransformInfo.cpp
```diff
@@ -251,3 +251,30 @@ bool AIETTICommon::isProfitableOuterLSR(const Loop &L) const {
...
+
+  // The cases of interest are 8 and 16-bit only.
+  return (StoreSize == 8) ? 4 : (StoreSize == 16) ? 2 : 1;
+}
+
+unsigned AIETTICommon::getLoadVectorFactor(unsigned VF, unsigned LoadSize,
+                                           unsigned ChainSizeInBytes,
+                                           VectorType *VecTy) const {
+  // Block load vectorization, it is costly to extract elements from vectors.
```
martien-de-jong, line 268: Although loads are also expensive, but that probably takes more context to compute gains.

llvm/lib/Target/AIE/AIEBaseTargetTransformInfo.cpp
```diff
@@ -251,3 +251,30 @@ bool AIETTICommon::isProfitableOuterLSR(const Loop &L) const {
...
+  return 1;
+}
+
+bool AIETTICommon::isLegalToVectorizeStoreChain(unsigned ChainSizeInBytes,
+                                                Align Alignment,
+                                                unsigned AddrSpace) const {
+  // Start from 4 byte sequences, to reach word stores. Alignment is
+  // is considered by default by the pass.
```
martien-de-jong: nit: double *is*

llvm/lib/Target/AIE/AIEBaseTargetTransformInfo.cpp
```diff
@@ -251,3 +251,30 @@ bool AIETTICommon::isProfitableOuterLSR(const Loop &L) const {
...
+  // Default return of allowsMisalignedMemoryAccesses is false.
+  return ChainSizeInBytes >= 4;
```
martien-de-jong, line 278: How would e.g. 5 be vectorized?

llvm/test/CodeGen/AIE/opt/store-vectorize-reduce.ll
martien-de-jong, line 1: could we have a separate test file with a few basic examples, i.e. just the stores that get combined?

llvm/lib/Target/AIE/AIEBaseTargetTransformInfo.cpp
```diff
@@ -251,3 +251,30 @@ bool AIETTICommon::isProfitableOuterLSR(const Loop &L) const {
...
+
+unsigned AIETTICommon::getStoreVectorFactor(unsigned VF, unsigned StoreSize,
+                                            unsigned ChainSizeInBytes,
+                                            VectorType *VecTy) const {
+  // For StoreSize > 16, don't vectorize.
+  if (StoreSize > 16)
+    return 1;
+
+  // The cases of interest are 8 and 16-bit only.
+  return (StoreSize == 8) ? 4 : (StoreSize == 16) ? 2 : 1;
```
martien-de-jong: I have a feeling that we could also allow 32x8 and 16x16 etc? I guess our BUILD_VECTOR patterns would be way more compact than separate stores. Even if it doesn't occur often it's almost for free.

martien-de-jong (commented):
I see no blockers, just a few nits.

llvm/lib/Target/AIE/AIEBaseTargetTransformInfo.cpp
andcarminati: According to the pass:
```
  // We use a simple greedy algorithm.
  //  - Given a chain of length N, find all prefixes that
  //    (a) are not longer than the max register length, and
  //    (b) are a power of 2.
  //  - Starting from the longest prefix, try to create a vector of that length.
  //  - If one of them works, great.  Repeat the algorithm on any remaining
  //    elements in the chain.
  //  - If none of them work, discard the first element and repeat on a chain
  //    of length N-1.
```

llvm/lib/Target/AIE/AIEBaseTargetTransformInfo.cpp
andcarminati: I see your point, for x8 case we will have a lot of shifts and ors that will be more expensive then a build vector. We could explore this in a follow-up work.

llvm/test/CodeGen/AIE/opt/store-vectorize-reduce.ll
andcarminati: Done.

llvm/lib/Target/AIE/AIEBaseTargetTransformInfo.cpp
andcarminati: Yes, maybe is cheaper to insert bytes on a vector than having shift + or sequences. We could explore in a follow-up work if we see benefit.

llvm/test/CodeGen/AIE/opt/vectorize-part-store.ll
```diff
@@ -0,0 +1,101 @@
...
+  %arrayidx4 = getelementptr inbounds nuw i8, ptr %B, i20 2
+  store i8 3, ptr %arrayidx4, align 2, !tbaa !2
+  %arrayidx6 = getelementptr inbounds nuw i8, ptr %B, i20 3
+  store i8 4, ptr %arrayidx6, align 1, !tbaa !2
+  %arrayidx8 = getelementptr inbounds nuw i8, ptr %B, i20 4
+  store i8 5, ptr %arrayidx8, align 4, !tbaa !2
+  ret void
+  }
+
+  ; This should be vectorized to two <2 x i16> stores.
```
martien-de-jong, line 48: Eh, this is one 4 x i16 already. Who is doing that?
Oh, right, you run the pass from opt with the wrong tuning.

llvm/lib/Target/AIE/AIEBaseTargetTransformInfo.cpp
martien-de-jong: So it would effectively round it down to four, the largest power of two. Ok.

martien-de-jong (approved):

PR #658, opened 2025-09-19
[AIEX] Split Constant and Load narrowing combiner

andcarminati:


llvm/lib/Target/AIE/AIECombinerHelper.cpp
```diff
@@ -3445,24 +3445,42 @@ bool llvm::matchNarrowTrunc(MachineInstr &MI, MachineRegisterInfo &MRI,
     return true;
   }
 
-  // Ideally, we could allow more users, provided that they are all TRUNCs.
-  // However, if we have more users, the live range of this register could
-  // spread through more blocks, and this could lead to more register pressure
-  // on s20 registers.
-  if (!MRI.hasOneNonDBGUse(SrcReg) || !isUsedByLikelyLegalS20User(MRI, MI))
+  return false;
```
F-Stuckmann: could you invert the return statements, so that all conditions lead to false and only if no condition triggered, you return a true?

So basically reducing nesting level of `if (SrcMI.getOpcode() == TargetOpcode::G_CONSTANT) {`

llvm/lib/Target/AIE/AIECombinerHelper.cpp
```diff
@@ -3445,24 +3445,42 @@ bool llvm::matchNarrowTrunc(MachineInstr &MI, MachineRegisterInfo &MRI,
...
+      return true;
+    const Register DstReg = MaybeTruncMI.getOperand(0).getReg();
+    if (MRI.getType(DstReg) != S20)
+      return true;
+    return !isUsedByLikelyLegalS20User(MRI, MaybeTruncMI);
+  };
+
+  // We should have a G_LOAD feeding interesting truncations.
+  const Register DstReg = MI.getOperand(0).getReg();
+  if (any_of(MRI.use_instructions(DstReg), IsNotProfitableTruncToS20))
```
F-Stuckmann: nice, that was also my implementation approach

llvm/lib/Target/AIE/AIECombinerHelper.cpp
```diff
@@ -3445,24 +3445,42 @@ bool llvm::matchNarrowTrunc(MachineInstr &MI, MachineRegisterInfo &MRI,
...
-  // However, if we have more users, the live range of this register could
-  // spread through more blocks, and this could lead to more register pressure
-  // on s20 registers.
-  if (!MRI.hasOneNonDBGUse(SrcReg) || !isUsedByLikelyLegalS20User(MRI, MI))
+  return false;
+}
+
+/// Narrow operations that are feeding truncations to s20.
+/// Covers G_LOAD.
+bool llvm::matchNarrowTruncLoad(MachineInstr &MI, MachineRegisterInfo &MRI,
```
F-Stuckmann, line 3453: nice work modularizing the combiners

llvm/lib/Target/AIE/AIECombinerHelper.cpp
```diff
@@ -3445,24 +3445,42 @@ bool llvm::matchNarrowTrunc(MachineInstr &MI, MachineRegisterInfo &MRI,
...
+                                CombinerHelper &Helper,
+                                GISelChangeObserver &Observer,
+                                BuildFnTy &MatchInfo) {
+
+  assert(MI.getOpcode() == TargetOpcode::G_LOAD);
+
+  const LLT S20 = LLT::scalar(20);
+  auto IsNotProfitableTruncToS20 = [&](const MachineInstr &MaybeTruncMI) {
```
F-Stuckmann: I think it is easier to understand if you invert the function here to `isProfitableTruncToS20`. 

Also don't we abort before can evaluate `isUsedByLikelyLegalS20User`?

llvm/test/CodeGen/AIE/GlobalISel/combine-trunc-to-s20.mir
```diff
@@ -164,3 +164,75 @@ body:             |
...
+  bb.1:
+    successors: %bb.4(0x40000000), %bb.3(0x40000000)
+    liveins: $r0
+
+    %1:_(s32) = COPY $r0
+    %0:_(s1) = G_TRUNC %1(s32)
+    %3:_(p0) = G_CONSTANT i20 0
+    %12:_(p0) = G_CONSTANT i20 0
+    %14:_(s32) = G_CONSTANT i32 0
+    %13:_(<32 x s32>) = G_BUILD_VECTOR %14(s32), %14(s32), %14(s32), %14(s32), %14(s32), %14(s32), %14(s32), %14(s32), %14(s32), %14(s32), %14(s32), %14(s32), %14(s32), %14(s32), %14(s32), %14(s32), %14(s32), %14(s32), %14(s32), %14(s32), %14(s32), %14(s32), %14(s32), %14(s32), %14(s32), %14(s32), %14(s32), %14(s32), %14(s32), %14(s32), %14(s32), %14(s32)
```
F-Stuckmann: Could you simplify the unit-test? 
Just have a load and two Truncs in different MBBs. You can catch the trunc values in the `PseudoRet` statement

llvm/test/CodeGen/AIE/GlobalISel/combine-trunc-to-s20.mir
```diff
@@ -164,3 +164,75 @@ body:             |
     %5:_(p0), %9:_(s20) = G_INTRINSIC intrinsic(@llvm.aie2.add.2d), %0:_(p0), %8:_(s20), %2:_(s20), %3:_(s20), %4:_(s20)
     $p0 = COPY %5
 ...
+
+# Here, we test multiple uses of the load. When all users are truncations to s20, which are used by
+# other s20 instructions, we should combine.
+
+---
+name:             G_LOAD_with_multiple_uses
```
F-Stuckmann, line 172: I would like to have a negative unit-test

llvm/test/CodeGen/AIE/GlobalISel/combine-trunc-to-s20.mir
andcarminati: This is will not catch the proper behavior. Please, see `isUsedByLikelyLegalS20User`.

andcarminati: I simplified the `G_BUILD_VECTOR`. Also, `G_INTRINSIC_W_SIDE_EFFECTS` is not removed by CSE, so we don't need phi nodes here.

F-Stuckmann (approved):
LGTM

andcarminati:
QoR: 
Core_Insn_Count
Core_PMSize_absolute
Core_StackSize_absolute

About the results, before we were relying on CSEed truncations to have single truncations related to loads. Now we handle all duplicated truncations and this means that we are now narrowing more aggressively. For some Conv2D* cases, we have more s20 loads, increasing pressure, and that pressure was smaller before because we had some loads standing on s32 register, that were moved to s20 on demand. I believe that the actual behavior is more desired, we should reduce stack using other means.

PR #660, opened 2025-09-22
[AIE2P] VextrBcst when extracting to integers instead of vectors

F-Stuckmann:


llvm/test/CodeGen/AIE/aie2p/shufflevec.ll
```diff
@@ -222,9 +222,9 @@ define <8 x i32> @test_shuffle_vector_to_extract_broadcast_vecElem(<8 x i32> nou
 ; CHECK:       // %bb.0: // %entry
 ; CHECK-NEXT:    nopa ; nopb ; nops ; ret lr; nopm ; nopv
 ; CHECK-NEXT:    nopx // Delay Slot 5
-; CHECK-NEXT:    vextract.32 r0, x2, #2, vaddsign1 // Delay Slot 4
-; CHECK-NEXT:    nop // Delay Slot 3
-; CHECK-NEXT:    vbcst.32 x0, r0 // Delay Slot 2
+; CHECK-NEXT:    nop // Delay Slot 4
+; CHECK-NEXT:    mova r0, #2 // Delay Slot 3
+; CHECK-NEXT:    vextbcst.32 x0, x2, r0 // Delay Slot 2
```
andcarminati: Great!

llvm/lib/Target/AIE/aie2p/AIE2PInstrPatterns.td
```diff
@@ -946,6 +946,13 @@ class BcstSubvecPat<AIE2PInst Inst, ValueType BcstDstTy,
                       (SubvecDstTy(extract_subvec_node(BcstDstTy VEC512:$s0), 
                                                       (i32 IdxTy:$idx))))), 
 (Inst VEC512:$s0, IdxTy:$idx)>;
+// Extract integer instead of a subvector
```
andcarminati: nit: new line.

llvm/lib/Target/AIE/aie2p/AIE2PInstrPatterns.td
```diff
@@ -946,6 +946,13 @@ class BcstSubvecPat<AIE2PInst Inst, ValueType BcstDstTy,
...
+class BcstIntPat<AIE2PInst Inst, ValueType BcstDstTy,
```
andcarminati: This class matches extract+broadcast. Maybe this could be reflected in the name, like `ExtractBcstIntPat`.

llvm/test/CodeGen/AIE/aie2p/shufflevec.ll
konstantinschwarz: VEXTBCST also supports immediate `idx` operands. Would be nice to support them as well, that would remove one more instruction here. See for example how we define `ExtractSubvecPat`

F-Stuckmann:
-0.24% Perf improvement
Core_Compute_Insn_Count

-3.6 Bytes saved
Core_PMSize_absolute

llvm/lib/Target/AIE/AIECombinerHelper.cpp
```diff
@@ -1048,33 +1048,56 @@ void llvm::applyGlobalValOffset(MachineInstr &MI, MachineRegisterInfo &MRI,
...
 
-bool llvm::matchExtractVecEltAndExt(
-    MachineInstr &MI, MachineRegisterInfo &MRI,
-    std::pair<MachineInstr *, bool> &MatchInfo) {
+namespace {
+bool feedsAnyExtBcstUse(MachineInstr &MI, MachineRegisterInfo &MRI,
+                        const AIEBaseInstrInfo &TII) {
+  assert(MI.getOpcode() == TargetOpcode::G_EXTRACT_VECTOR_ELT);
+
+  auto IsSingleNonDbgUse = [&MRI](MachineInstr &MI,
```
andcarminati: Maybe `IsSingleUsedByNonDbg`.

llvm/lib/Target/AIE/AIECombinerHelper.cpp
```diff
@@ -1048,33 +1048,56 @@ void llvm::applyGlobalValOffset(MachineInstr &MI, MachineRegisterInfo &MRI,
...
+    return UseMI;
+  };
+
+  auto *AnyExtMI = IsSingleNonDbgUse(MI, TargetOpcode::G_ANYEXT);
+  if (!AnyExtMI)
+    return false;
+
+  auto *BcstMI =
+      IsSingleNonDbgUse(*AnyExtMI, TII.getGenericBroadcastVectorOpcode());
+  return BcstMI;
```
andcarminati: Are we implicitly converting a pointer to a boolean?

llvm/lib/Target/AIE/AIECombinerHelper.cpp
andcarminati: This gives an impression that we are returning a boolean, but we are returning a pointer.

llvm/lib/Target/AIE/AIECombine.td
```diff
@@ -42,11 +42,11 @@ def combine_globalval_offset : GICombineRule<
   (apply [{  applyGlobalValOffset(*${root}, MRI, B, Observer, ${matchinfo});}])
 >;
 
-def combine_extract_vector_elt_and_zsa_ext_matchdata: GIDefMatchData<"std::pair<MachineInstr *, bool>">;
+def combine_extract_vector_elt_and_zsa_ext_matchdata: GIDefMatchData<"std::pair<MachineInstr *, std::pair<bool, bool>>">;
```
andcarminati: I would be happy if we could use `GIDefMatchData<"std::tuple<MachineInstr *, bool, bool>">` here.

llvm/lib/Target/AIE/AIECombinerHelper.cpp
```diff
@@ -1084,9 +1107,11 @@ bool llvm::matchExtractVecEltAndExt(
 
 void llvm::applyExtractVecEltAndExt(
     MachineInstr &MI, MachineRegisterInfo &MRI, MachineIRBuilder &B,
-    std::pair<MachineInstr *, bool> &MatchInfo) {
+    std::pair<MachineInstr *, std::pair<bool, bool>> &MatchInfo) {
   B.setInstrAndDebugLoc(MI);
-  auto [MatchMI, IsSignedExt] = MatchInfo;
+  auto [MatchMI, BoolPair] = MatchInfo;
```
andcarminati: With the tuple, you can use `std::tie(...) = MatchInfo;` to get the components.

llvm/lib/Target/AIE/aie2p/AIE2PInstrPatterns.td
```diff
@@ -940,41 +940,78 @@ def : Pat<(int_aie2p_vextract_broadcast128_I512 VEC512:$src, c6u:$idx),
...
-class BcstSubvecPat<AIE2PInst Inst, ValueType BcstDstTy, 
+class BcstExtSubvecPat<AIE2PInst Inst, ValueType BcstDstTy, 
                       ValueType SubvecDstTy, DAGOperand IdxTy> :
   Pat<(BcstDstTy (bcst_vector_node
                       (SubvecDstTy(extract_subvec_node(BcstDstTy VEC512:$s0), 
                                                       (i32 IdxTy:$idx))))), 
 (Inst VEC512:$s0, IdxTy:$idx)>;
 
-class BcstSubvecPair<AIE2PInst inst, DAGOperand idxTy>{
+// Extract scalar instead of a subvector
```
andcarminati, line 950: Nice. I like this use of class.

llvm/lib/Target/AIE/aie2p/AIE2PInstrPatterns.td
```diff
@@ -940,41 +940,78 @@ def : Pat<(int_aie2p_vextract_broadcast128_I512 VEC512:$src, c6u:$idx),
...
+foreach Pair = BcstExtSubvec32Pairs in {
+  def : BcstExtSubvecPat<Pair.Inst, v64i8, v4i8, Pair.IdxTy>;
+  def : BcstExtSubvecPat<Pair.Inst, v32i16, v2i16, Pair.IdxTy>;
 }
 
+foreach  Pair = BcstExtSubvec64Pairs in {
+  def : BcstExtSubvecPat<Pair.Inst, v64i8, v8i8, Pair.IdxTy>;
+  def : BcstExtSubvecPat<Pair.Inst, v32i16, v4i16, Pair.IdxTy>;
+  def : BcstExtSubvecPat<Pair.Inst, v16i32, v2i32, Pair.IdxTy>;
+  def : BcstExtSubvecPat<Pair.Inst, v8i64, v2i32, Pair.IdxTy>;
```
andcarminati: Is this case covered by a test?

F-Stuckmann: yes when they were introduced, they came with unit-tests.

I haven't changed the subvector ExtrBCST versions. I worked on the scalar ExtrBcst

llvm/lib/Target/AIE/AIECombinerHelper.cpp
F-Stuckmann: updated the naming, it was confusing

llvm/lib/Target/AIE/AIECombinerHelper.cpp
F-Stuckmann: yes, should i make this clearer?

llvm/lib/Target/AIE/AIECombinerHelper.cpp
F-Stuckmann: I can also directly access them 
`auto [MatchMI, IsSignedExt, BuildAssert] = MatchInfo;`

llvm/lib/Target/AIE/AIECombinerHelper.cpp
F-Stuckmann: I am making it into an explicit cast for clarity

llvm/lib/Target/AIE/AIECombinerHelper.cpp
```diff
@@ -1048,33 +1048,56 @@ void llvm::applyGlobalValOffset(MachineInstr &MI, MachineRegisterInfo &MRI,
...
 
-bool llvm::matchExtractVecEltAndExt(
-    MachineInstr &MI, MachineRegisterInfo &MRI,
-    std::pair<MachineInstr *, bool> &MatchInfo) {
+namespace {
+bool feedsAnyExtBcstUse(MachineInstr &MI, MachineRegisterInfo &MRI,
+                        const AIEBaseInstrInfo &TII) {
+  assert(MI.getOpcode() == TargetOpcode::G_EXTRACT_VECTOR_ELT);
+
+  auto GetSingleNonDbgUse = [&MRI](MachineInstr &MI,
```
andcarminati, line 1056: Side note: considering the llvm convention, Use usually refers to MOs and User to MIs. Most of the time they will give the same answer, but not always. See, for example `hasOneNonDBGUse` and `hasOneNonDBGUser`. As we are returning an instruction, `User` could be a better fit (for the lambda name), but maybe it is a too specific detail.

llvm/lib/Target/AIE/aie2p/AIE2PInstrPatterns.td
andcarminati: Oh, true!

andcarminati (approved):
LGTM.

PR #661, opened 2025-09-23
[AIEX] Improve AA for the PostSWP considering split 2D/3D post increment memory ops

andcarminati:
Now we handle can more complex cases, where a big 2D/3D memory operation is split in 2 (source code level) and we can now disambiguate the offset load, not just the postinc load.


andcarminati:
QoR:
Core_Insn_Count
Core_PMSize

llvm/lib/Target/AIE/AIEBaseAliasAnalysis.cpp
```diff
@@ -620,12 +620,41 @@ AliasResult AIEBaseAAResult::alias(const MemoryLocation &LocA,
   return AliasResult::MayAlias;
 }
 
+// Returns true if the pointer (Value) has an intrinsic padd as user.
+static bool hasInstrinsicPAddUser(const Value *V) {
```
F-Stuckmann: nit: `hasIntrinsicPAddUser`

llvm/lib/Target/AIE/AIEBaseAliasAnalysis.cpp
```diff
@@ -620,12 +620,41 @@ AliasResult AIEBaseAAResult::alias(const MemoryLocation &LocA,
...
+  for (const User *U : V->users()) {
+    if (const Instruction *Instr = dyn_cast<const Instruction>(U)) {
+      if (Instr->isCast())
+        return hasInstrinsicPAddUser(Instr);
+      if (const auto *MaybeIntrinsicCall = dyn_cast<CallBase>(Instr)) {
+        Intrinsic::ID ID = MaybeIntrinsicCall->getIntrinsicID();
+        unsigned PtrIdx;
+        if (isAIEPtrAddIntrinsic(ID, PtrIdx))
```
F-Stuckmann: nit: can you reduce the nesting levels?

llvm/lib/Target/AIE/AIEBaseAliasAnalysis.cpp
```diff
@@ -636,6 +665,29 @@ AliasResult AIE::aliasAcrossVirtualUnrolls(const MachineInstr *MIA,
     if (!Base)
       return std::nullopt;
 
+    if (auto ConstantOffset = getGEPConstantOffset(Val)) {
+      const int64_t Offset = *ConstantOffset;
+      const int64_t Size =
+          MMO->getSize().hasValue() ? MMO->getSize().getValue() : 0;
```
F-Stuckmann, line 673: if we do  not have a MMO Size, we might as return early

llvm/lib/Target/AIE/AIEBaseAliasAnalysis.cpp
```diff
@@ -636,6 +665,29 @@ AliasResult AIE::aliasAcrossVirtualUnrolls(const MachineInstr *MIA,
...
+      // store <16 x i32> %X, ptr %Ptr
+      // %OffsetPtr = getelementptr inbounds i8 %Ptr, i20 64
+      // store <16 x i32> %Y, ptr %OffsetPtr <- we are interested in this one.
+      // %65 = tail call { ptr, i20 } @llvm.aie2p.add.2d(ptr %Ptr, ....)
+      if (Size == Offset) {
+        // Let's try to find a padd intrinsic for this whole memory operation.
+        // If we have an intrinsic that uses the same base, we can assume that
+        // this memory operation is related to the memory operation targeted by
+        // the intrinsic.
+        if (hasInstrinsicPAddUser(Base)) {
```
F-Stuckmann: why are we limiting ourselves to only intrinsicPADDs?

llvm/test/CodeGen/AIE/aie2p/schedule/postpipeliner/reduce-mean-templated.ll
```diff
@@ -0,0 +1,142 @@
...
+entry:
+  call void @llvm.set.loop.iterations.i32(i32 1)
+  br label %for.body68
+
+for.cond.cleanup67:                               ; preds = %for.body68
+  ret void
+
+for.body68:                                       ; preds = %for.body68, %entry
+  %p_in_group.sroa.0.0.in103 = phi ptr [ %ifm, %entry ], [ %16, %for.body68 ]
+  %p_out_group.sroa.0.1102 = phi ptr addrspace(6) [ %ofm, %entry ], [ %8, %for.body68 ]
```
F-Stuckmann, line 73: I find it very difficult to follow this unit test. Could you create a slimmed down version that is based on the comment 

` store <16 x i32> %X, ptr %Ptr`
`%OffsetPtr = getelementptr inbounds i8 %Ptr, i20 64`
` store <16 x i32> %Y, ptr %OffsetPtr <- we are interested in this one.`
 `// %65 = tail call { ptr, i20 } @llvm.aie2p.add.2d(ptr %Ptr, ....)`
 
 I also assume using loads would massivly simplify the unit-test, you can make sure they do not get DCE is by adding them as implit in the return.
 
 Also a unit-test just running the scheduler would probably also be clearer and less fragile

llvm/lib/Target/AIE/AIEBaseAliasAnalysis.cpp
```diff
@@ -636,6 +665,29 @@ AliasResult AIE::aliasAcrossVirtualUnrolls(const MachineInstr *MIA,
...
+          MMO->getSize().hasValue() ? MMO->getSize().getValue() : 0;
+      // This could be part of a big 2D/3D load/store, where we have a first
+      // offset load/store and a second paddXd post inc load/store (any order).
+      // The offset from the offset load/store must match the amount of
+      // loaded/stored data. Example:
+
+      // store <16 x i32> %X, ptr %Ptr
+      // %OffsetPtr = getelementptr inbounds i8 %Ptr, i20 64
+      // store <16 x i32> %Y, ptr %OffsetPtr <- we are interested in this one.
+      // %65 = tail call { ptr, i20 } @llvm.aie2p.add.2d(ptr %Ptr, ....)
```
F-Stuckmann, line 682: we implicitly assume, that 2d is not a zero-move, right?
Would that make sense to elevate to basic AA as well?

llvm/lib/Target/AIE/AIEBaseAliasAnalysis.cpp
andcarminati: For the cases with 1D padds, the existing code can handle. See `isLockStepGEPChain`. This case is used in cases where a GEP is part of a big 2D/3D operation.

llvm/lib/Target/AIE/AIEBaseAliasAnalysis.cpp
andcarminati: Please note that we still need to return a `std::make_pair(Val, Base);`. I would like to keep a single return. With a test for the existence of size, we have two options:
* Early return.
* Insert another if scope.

I see the current form as the cleaner one.

llvm/test/CodeGen/AIE/aie2p/schedule/postpipeliner/reduce-mean-templated.ll
andcarminati: Indeed we are using loads in the test (representative reduction of the problem) and we are trying to disambiguate loads and stores. Please note that we are using a loop as a test with a `load -> operation -> store` pattern, and the disconnection loads and stores will lead to a more complicated (and less understandable) scenario. By running just the post-scheduler, I got no pipelined version, I think some metadata information is still not correctly serialized, sometimes it works, sometimes not, we need an investigation here.

llvm/lib/Target/AIE/AIEBaseAliasAnalysis.cpp
andcarminati: do you mean a non-zero increment? Yes, we assume this. For the basic AA, we look to this in `getUnderlyingObject`.

llvm/lib/Target/AIE/AIEBaseAliasAnalysis.cpp
andcarminati: Yes! done.

llvm/test/CodeGen/AIE/aie2p/schedule/postpipeliner/reduce-mean-templated.ll
F-Stuckmann: that is inconvenient, we should not be losing serialization data

llvm/test/CodeGen/AIE/aie2p/schedule/postpipeliner/reduce-mean-templated.ll
```diff
@@ -0,0 +1,142 @@
+; NOTE: Assertions have been autogenerated by utils/update_llc_test_checks.py
+
+; This file is licensed under the Apache License v2.0 with LLVM Exceptions.
+; See https://llvm.org/LICENSE.txt for license information.
+; SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+;
+; (c) Copyright 2025 Advanced Micro Devices, Inc. or its affiliates
+
+; This test checks postpipelining for aie2p considering AA analysis
```
F-Stuckmann, line 9: Could you mention, that the loop body contains `load -> vadd-> store` since the IR is very bloated. I think this would help in the understanding of the unittest

andcarminati: Done!

F-Stuckmann (approved):

PR #664, opened 2025-09-24
[AIE2P] Add alternative calling convention for library calls

konstantinschwarz:
In AIEs default calling convention, all vector registers are caller saved. However, most compiler-rt builtin library functions do not use vector registers. To avoid having to spill vectors to memory in the caller function, we can attach a different calling convention to these library functions.

I manually inspected each library function where we apply the alternative calling convention to make sure none of them actually use vector registers. To make this actually save and robust, we should also add a custom calling convention attribute to the frontend that ensures vector registers are actually saved by the callee.

### DIFF
--- a/llvm/lib/Target/AIE/aie2p/AIE2PCallingConv.td
+++ b/llvm/lib/Target/AIE/aie2p/AIE2PCallingConv.td
@@ -219,5 +219,12 @@ def CC_AIE2P : CallingConv<[
 def CSR_AIE2P
     : CalleeSavedRegs<(add lr, r8, r9, r10, r11, r12, r13, r14, r15, p6, p7)>;
 
+def CSR_AIE2P_Vec
+    : CalleeSavedRegs<(add lr, r8, r9, r10, r11, r12, r13, r14, r15, p6, p7,
+          wl0, wl2, wl4, wl6, wl8, wl10, wl1, wl3, wl5, wl7, wl9, wl11, wh0,
+          wh2, wh4, wh6, wh8, wh10, wh1, wh3, wh5, wh7, wh9, wh11, bmll0, bmll1,
+          bmll2, bmll3, bmll4, bmhl0, bmhl1, bmhl2, bmhl3, bmhl4, bmlh0, bmlh1,
+          bmlh2, bmlh3, bmlh4, bmhh0, bmhh1, bmhh2, bmhh3, bmhh4)>;
+
 // Needed for implementation of AIERegisterInfo::getNoPreservedMask()
 def CSR_NoRegs : CalleeSavedRegs<(add)>;
--- a/llvm/lib/Target/AIE/aie2p/AIE2PISelLowering.cpp
+++ b/llvm/lib/Target/AIE/aie2p/AIE2PISelLowering.cpp
@@ -17,14 +17,37 @@
 #include "MCTargetDesc/aie2p/AIE2PMCTargetDesc.h"
 #include "llvm/CodeGen/TargetLowering.h"
 #include "llvm/IR/IntrinsicsAIE2P.h"
+#include "llvm/IR/RuntimeLibcalls.h"
 
 using namespace llvm;
 
 #define DEBUG_TYPE "aie-lower"
 
+extern cl::opt<bool> VecCCLibcalls;
+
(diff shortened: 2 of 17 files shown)


konstantinschwarz:
This saves up to 512 bytes of stack (ArgMin/ArgMax), and some very minor code size savings.

Core_PMSize_absolute
Core_StackSize_absolute

llvm/lib/Target/AIE/aie2p/AIE2PISelLowering.cpp
```diff
@@ -21,9 +21,34 @@ using namespace llvm;
...
+    setLibcallCallingConv(RTLIB::SINTTOFP_I64_F64,
+                          CallingConv::AIE_PreserveAll_Vec);
+    setLibcallCallingConv(RTLIB::SDIV_I32, CallingConv::AIE_PreserveAll_Vec);
+    setLibcallCallingConv(RTLIB::SDIV_I64, CallingConv::AIE_PreserveAll_Vec);
+    setLibcallCallingConv(RTLIB::UDIV_I32, CallingConv::AIE_PreserveAll_Vec);
+    setLibcallCallingConv(RTLIB::UDIV_I64, CallingConv::AIE_PreserveAll_Vec);
+    setLibcallCallingConv(RTLIB::SREM_I32, CallingConv::AIE_PreserveAll_Vec);
+    setLibcallCallingConv(RTLIB::SREM_I64, CallingConv::AIE_PreserveAll_Vec);
+    setLibcallCallingConv(RTLIB::UREM_I32, CallingConv::AIE_PreserveAll_Vec);
+    setLibcallCallingConv(RTLIB::UREM_I64, CallingConv::AIE_PreserveAll_Vec);
```
andcarminati, line 48: I see some other libcalls in the LegalizerHelper. Are these only the verified ones?

konstantinschwarz: Yes, these are only the ones I verified. Others could likely be included as well, but I wanted to keep this limited for now

F-Stuckmann (dismissed):

llvm/lib/Target/AIE/aie2p/AIE2PCallingConv.td
```diff
@@ -219,5 +219,12 @@ def CC_AIE2P : CallingConv<[
 def CSR_AIE2P
     : CalleeSavedRegs<(add lr, r8, r9, r10, r11, r12, r13, r14, r15, p6, p7)>;
 
+def CSR_AIE2P_Vec
+    : CalleeSavedRegs<(add lr, r8, r9, r10, r11, r12, r13, r14, r15, p6, p7,
```
martien-de-jong, line 223: nit: can we write this as an extension to CSR_AIE2P?

llvm/lib/Target/AIE/aie2p/AIE2PISelLowering.cpp
```diff
@@ -22,9 +22,34 @@ using namespace llvm;
...
+    VecCCLibcalls("aie-libcalls-preserve-vectors", cl::init(true), cl::Hidden,
+                  cl::desc("Assume all vector registers are callee-saved by "
+                           "builtin library functions."));
+
 AIE2PTargetLowering::AIE2PTargetLowering(const TargetMachine &TM,
                                          const AIEBaseSubtarget &STI)
     : AIEBaseTargetLowering(TM, STI) {
+
+  if (VecCCLibcalls) {
+    setLibcallCallingConv(RTLIB::SINTTOFP_I32_F32,
```
martien-de-jong, line 33: nit: perhaps make a data-driven routine running on an array of function ids?

llvm/lib/Target/AIE/aie2p/AIE2PCallingConv.td
martien-de-jong: e.g. factoring using a named common set?

konstantinschwarz:
Rebased and added support for aie2ps. @F-Stuckmann can you approve again?

F-Stuckmann (dismissed):

F-Stuckmann (approved):

PR #665, opened 2025-09-25
[AIEX][NFC] Creation of LLT pointer type fails for bigger sizes(> 16bit)

niwinanto:


llvm/lib/Target/AIE/AIEEliminateDuplicatePHI.cpp
```diff
@@ -70,10 +70,9 @@ static void eliminateDuplicatePHIUses(MachineRegisterInfo &MRI,
 /// \param MI The instruction to check.
 static bool isPointerTypePHI(MachineRegisterInfo &MRI, const MachineInstr &MI) {
   assert(MI.getOpcode() == TargetOpcode::G_PHI);
-  const int DefaultAddrSpace = 0;
   const Register Reg = MI.getOperand(0).getReg();
   const LLT RegType = MRI.getType(Reg);
-  return RegType == LLT::pointer(DefaultAddrSpace, RegType.getSizeInBits());
+  return RegType.isPointer();
```
andcarminati, line 75: Do we trigger an assert somewhere?

andcarminati: This change is also a good simplification ;-)

andcarminati (approved):
LGTM.

PR #669, opened 2025-10-01
[AIEX] Add address chaining for fixed stack objects

andcarminati:


llvm/test/CodeGen/AIE/aie2p/shufflevec.ll
```diff
@@ -930,24 +930,24 @@ define <16 x i32> @shuffle_concat_extracted_subvectors_numSubvecs_not_powerOf2_n
...
 ; CHECK:       // %bb.0:
-; CHECK-NEXT:    vlda bmll0, [sp, #-64]
-; CHECK-NEXT:    vlda bmll1, [sp, #-128]
-; CHECK-NEXT:    nop
-; CHECK-NEXT:    nop
+; CHECK-NEXT:    vlda bmll0, [sp, #-64]; nopb ; nops ; nopxm ; nopv
 ; CHECK-NEXT:    nop
 ; CHECK-NEXT:    nop
 ; CHECK-NEXT:    nop
+; CHECK-NEXT:    mov p0, sp
```
andcarminati, line 937: Interesting case, I will take a look!

andcarminati: I looks like we combine frame index + load store to _spill variants. I tried to prevent this case, but it also prevents some good opportunities.

llvm/lib/Target/AIE/AIEClusterBaseAddress.cpp
```diff
@@ -347,6 +360,70 @@ bool AIEClusterBaseAddress::avoidPtrAdd(
...
+  for (MachineInstr &FIInstr : MBB) {
+    // Only consider G_FRAME_INDEX
+    if (FIInstr.getOpcode() != TargetOpcode::G_FRAME_INDEX)
+      continue;
+
+    const int FrameIdx = FIInstr.getOperand(1).getIndex();
+    if (!MFI.isFixedObjectIndex(FrameIdx))
+      continue;
+
+    if (IsUsedByLoad(FIInstr))
```
F-Stuckmann: should we worry about stores?

F-Stuckmann: why do we differentiate between loads and non-loads? shouldn't it only be related to the FrameIdx?

llvm/lib/Target/AIE/AIEClusterBaseAddress.cpp
```diff
@@ -347,6 +360,70 @@ bool AIEClusterBaseAddress::avoidPtrAdd(
...
+    for (MachineInstr *FI : make_range(next(FIs.begin()), FIs.end())) {
+      const int64_t Offset =
+          MFI.getObjectOffset(FI->getOperand(1).getIndex()) - FirstOffset;
+      MIB.setInstrAndDebugLoc(*FI);
+      Register NewOffsetReg =
+          MIB.buildConstant(LLT::scalar(20), Offset).getReg(0);
+
+      Observer.createdInstr(*MIB.buildInstr(TargetOpcode::G_PTR_ADD,
+                                            {FI->getOperand(0).getReg()},
+                                            {FirstPtr, NewOffsetReg}));
```
F-Stuckmann: Is there a reason we are create offsets based on the first FrameIndex ptr and not based on the previous one ? 
If we were using relative offsets, we could switch the ordering of the processing, right? 

```
  for (MachineBasicBlock &MBB : MF) {
    if (EnableStackChaining)
      Changed |= convertFIToPtrAdd(MBB, MIB, Observer);
    Changed |= processBasicBlock(MBB, MIB, Observer);
  }
```

llvm/test/CodeGen/AIE/aie2/end-to-end/Add2D-red.ll
```diff
@@ -34,48 +34,39 @@ declare { ptr, i20, i20 } @llvm.aie2.add.3d(ptr, i20, i20, i20, i20, i20, i20, i
...
+; ASM-NEXT:    lda r1, [p0], #-12
+; ASM-NEXT:    lda r3, [p0], #40
+; ASM-NEXT:    lda m1, [p0], #36
+; ASM-NEXT:    lda m0, [p0], #-8
+; ASM-NEXT:    lda dn0, [p0], #-8
+; ASM-NEXT:    lda dj0, [p0], #12
+; ASM-NEXT:    lda dn4, [p0], #-8; paddb [sp], #32
+; ASM-NEXT:    lda dj4, [p0], #-36; st p7, [sp, #-32] // 4-byte Folded Spill
+; ASM-NEXT:    lda r1, [p0, #0]; mov p7, sp
+; ASM-NEXT:    lda r5, [p0, #-36]; paddb [p7], #-36
```
F-Stuckmann, line 52: curious. I wonder if a shorter load chain would lead to offset loads?

llvm/test/CodeGen/AIE/aie2/end-to-end/Add2D-red.ll
```diff
@@ -34,48 +34,39 @@ declare { ptr, i20, i20 } @llvm.aie2.add.3d(ptr, i20, i20, i20, i20, i20, i20, i
...
-; ASM-NEXT:    st r6, [p7, #0]; paddb [p2], m4; and r8, r2, r0 // Delay Slot 2
+; ASM-NEXT:    lda m3, [p0], #-24
+; ASM-NEXT:    lda r0, [p0], #4
+; ASM-NEXT:    lda r1, [p0], #-12
+; ASM-NEXT:    lda r3, [p0], #40
+; ASM-NEXT:    lda m1, [p0], #36
+; ASM-NEXT:    lda m0, [p0], #-8
+; ASM-NEXT:    lda dn0, [p0], #-8
+; ASM-NEXT:    lda dj0, [p0], #12
+; ASM-NEXT:    lda dn4, [p0], #-8; paddb [sp], #32
```
F-Stuckmann, line 49: This paddb is unnecessary. Do you know why it is introduced and we are not immediatly going for a offset load?

F-Stuckmann: I think this comes from the arbitrary separating loads and non-load FrameIndex chains

llvm/lib/Target/AIE/AIEClusterBaseAddress.cpp
andcarminati: The reason is simplicity. We have an address chaining engine, we don't need to reimplement chaining.

llvm/lib/Target/AIE/AIEClusterBaseAddress.cpp
andcarminati: Maybe we could vanish this, because we break the chain using `hasMixedLoadStoreUse`. I will check this.

llvm/test/CodeGen/AIE/aie2/end-to-end/Add2D-red.ll
andcarminati: It was already here but scheduled in a different position.

andcarminati: See the left side `; ASM-NEXT: lda m3, [p0], #-24; paddb [sp], #32`

llvm/test/CodeGen/AIE/aie2/end-to-end/Add2D-red.ll
andcarminati: The first frame index is never eliminated by address chaining because we don't know the real offset before `PrologEpilogInserter`. We do only `relative` work here. This mov+padd is inserted very late here: https://github.com/Xilinx/llvm-aie/blob/8a04c6730ebabbe610e02dbf70b5cbaf9c96f543/llvm/lib/Target/AIE/aie2p/AIE2PRegisterInfo.cpp#L240

andcarminati: Basically after regalloc......

llvm/lib/Target/AIE/AIEClusterBaseAddress.cpp
```diff
@@ -347,6 +360,55 @@ bool AIEClusterBaseAddress::avoidPtrAdd(
...
+
+    const int FrameIdx = FIInstr.getOperand(1).getIndex();
+    if (!MFI.isFixedObjectIndex(FrameIdx))
+      continue;
+
+    FIs.push_back(&FIInstr);
+  }
+
+  bool Changed = false;
+  auto ReplaceFI = [&](std::vector<MachineInstr *> &FIs) {
```
F-Stuckmann: nit: with the simplification, do we need the lambda here?

F-Stuckmann:
LGTM

F-Stuckmann (approved):

PR #670, opened 2025-10-01
[DRAFT] [AIEX] simplify Debug types across AIE architectures

F-Stuckmann:


llvm/lib/Target/AIE/AIE2PostLegalizerCustomCombiner.cpp
```diff
@@ -25,7 +25,7 @@
 #include "llvm/CodeGen/MachineDominators.h"
 #include "llvm/InitializePasses.h"
 
-#define DEBUG_TYPE "aie2-postlegalizer-custom-combiner"
```
andcarminati, line 28: I guess this problematic because we use this as the pass name. If we decouple the debug from the pass name it could be fine.

andcarminati: See `INITIALIZE_PASS_BEGIN`. We register two passes with the same name.

F-Stuckmann: yeah, but i also don't like a different pass name and a different debug type.

llvm/lib/Target/AIE/AIE2PostLegalizerCustomCombiner.cpp
```diff
@@ -25,7 +25,7 @@
...
+#define DEBUG_TYPE "postlegalizer-custom-combiner"
```
niwinanto, line 28: Do you want to retain the `aie` keyword

F-Stuckmann:
properly fixed in https://github.com/Xilinx/llvm-aie/pull/828

PR #671, opened 2025-10-01
feat(tosa): sink tosa ops through tosa.concat

gerion-amd:
Currently supports elementwise and reduce operations, additional to matmul and reshape (within a limited set).
Could be extended to other operations later.


mlir/lib/Dialect/Tosa/Transforms/SinkInputOpsThroughConcat.cpp
```diff
@@ -0,0 +1,425 @@
...
+  SinkInputOpsThroughConcat(llvm::raw_ostream &os) : os(os) {};
+
+  void runOnOperation() override {
+    auto func = getOperation();
+    RewritePatternSet patterns(func.getContext());
+    MLIRContext *ctx = func.getContext();
+    // elementwise
+    patterns.add<
+        SinkSpecificOp<tosa::AbsOp>, SinkSpecificOp<tosa::BitwiseNotOp>,
+        SinkSpecificOp<tosa::CeilOp>, SinkSpecificOp<tosa::ClampOp>,
```
jorickert: I would not sink Clamp or Floor, as they are in the form of Relu often fused with the previous op, which may be prevented by the sinking.

mlir/lib/Dialect/Tosa/Transforms/SinkInputOpsThroughConcat.cpp
```diff
@@ -0,0 +1,425 @@
...
+        rewriter.create(concatOp->getLoc(), sample->getName().getIdentifier(),
+                        concatOps, concatOp.getType());
+    for (auto attr : sample->getAttrs()) {
+      switchedOp->setAttr(attr.getName(), attr.getValue());
+    }
+    rewriter.replaceOp(concatOp, switchedOp);
+    return success();
+  }
+
+protected:
```
jorickert, line 169: why protected and not private?

mlir/lib/Dialect/Tosa/Transforms/SinkInputOpsThroughConcat.cpp
```diff
@@ -0,0 +1,425 @@
...
+      if (!val.hasOneUse())
+        return GenericCheckResult<OpT>{
+            nullptr, "Operands must just connect to this concat."};
+
+      auto op = val.getDefiningOp<OpT>();
+      if (!op) {
+        std::string err;
+        llvm::raw_string_ostream errStream(err);
+        errStream << "Operand is not a " << OpT::getOperationName();
+        return GenericCheckResult<OpT>{nullptr, err.c_str()};
```
jorickert: Aren't you returning a pointer to function local memory here?

mlir/lib/Dialect/Tosa/Transforms/SinkInputOpsThroughConcat.cpp
```diff
@@ -0,0 +1,425 @@
...
+  llvm::raw_ostream &os;
+};
+
+struct CheckResult {
+  bool rewriteIsPossible;
+  const char *msg;
+};
+
+template <class OpT>
+struct GenericCheckResult {
```
jorickert: I am not a fan of passing around `char *`.

Instead you could change the return type to `FailureOr<OpT>` and do an  `return r.notifyMatchFailure("you msg")` in the functions that do the checks

mlir/lib/Dialect/Tosa/Transforms/SinkInputOpsThroughConcat.cpp
```diff
@@ -0,0 +1,425 @@
...
+  OpT operand;
+  const char *msg;
+};
+
+template <class OpT>
+bool checkForEquality(OpT sample, OpT op) {
```
jorickert: IMO this function is named suboptimal, as it does not check that two ops are equal, only that they have the same attributes and operand types. There can be other differences, for example a different body.

Personally i would split it in an `checkOpAttributesForEquality` and do the operand type check explicit

mlir/lib/Dialect/Tosa/Transforms/SinkInputOpsThroughConcat.cpp
```diff
@@ -0,0 +1,425 @@
...
+
+  LogicalResult matchAndRewrite(tosa::ConcatOp concatOp,
+                                PatternRewriter &rewriter) const override {
+    auto genChecks = doGenericChecks(concatOp);
+    auto sample = genChecks.operand;
+    if (!sample) {
+      return rewriter.notifyMatchFailure(concatOp, genChecks.msg);
+    }
+
+    const auto extraChecks = this->doExtraChecks(sample, concatOp);
```
jorickert: nit:
```suggestion
    const auto extraChecks = doExtraChecks(sample, concatOp);
```

mlir/lib/Dialect/Tosa/Transforms/SinkInputOpsThroughConcat.cpp
```diff
@@ -0,0 +1,425 @@
...
+    if (!extraChecks.rewriteIsPossible) {
+      return rewriter.notifyMatchFailure(concatOp, extraChecks.msg);
+    }
+
+    // rewriting, first collect all inputs and rearrange them for the concats
+    SmallVector<SmallVector<Value>> concatOperands(sample->getNumOperands());
+    for (auto val : concatOp.getOperands()) {
+      auto *rewriteOp = val.getDefiningOp();
```
jorickert:
```suggestion
      auto *producerOp= val.getDefiningOp();
```
This op is not getting rewritten?

mlir/lib/Dialect/Tosa/Transforms/SinkInputOpsThroughConcat.cpp
```diff
@@ -0,0 +1,425 @@
...
+      assert(rewriteOp != nullptr &&
+             "Previous check about null already happened.");
+      for (unsigned j = 0; j < rewriteOp->getNumOperands(); ++j) {
+        if (concatOperands.size() <= j) {
+          concatOperands.emplace_back(
+              SmallVector<Value>(concatOp->getNumOperands()));
+        }
```
jorickert: Do I understnad it correctly, that this branch will only be taken if an producer op has a variadic number of operands?
Also, why do you init the vector with the size of concat operands?

mlir/lib/Dialect/Tosa/Transforms/SinkInputOpsThroughConcat.cpp
```diff
@@ -0,0 +1,425 @@
...
+      concatOps.emplace_back(concatOpReplacement);
+    }
+    // leave it generic here to avoid a bunch of different constructors
+    auto *switchedOp =
+        rewriter.create(concatOp->getLoc(), sample->getName().getIdentifier(),
+                        concatOps, concatOp.getType());
+    for (auto attr : sample->getAttrs()) {
+      switchedOp->setAttr(attr.getName(), attr.getValue());
+    }
+    rewriter.replaceOp(concatOp, switchedOp);
```
jorickert: I think you can simplify this:
```suggestion
    rewriter.replaceOpWithNewOp<OpT>(concatOp, concatOp.getResultTypes(), concatOps, sample->getAttrs());
```

mlir/lib/Dialect/Tosa/Transforms/SinkInputOpsThroughConcat.cpp
```diff
@@ -0,0 +1,425 @@
...
+    assert(shapes.size() == op->getNumOperands() &&
+           "Something went wrong with the above loop.");
+    auto first = shapes[0];
+    for (size_t i = 0; i < first.size(); ++i) {
+      for (size_t j = 1; j < shapes.size(); ++j) {
+        if ((first[i] == 1 || shapes[j][i] == 1) && shapes[j][i] != first[i] &&
+            i == concat.getAxis()) {
+          return CheckResult{false, msg};
+        }
+      }
```
jorickert: If I understand it correctly all the shapes on the concat axis dim need  to be 1, or all need to  be not 1.

You do not need the nested loop here:

```suggestion
const auto axis = concat.getAxis();
const auto &first = shapes[0];
const bool firstIsOne = first[axis] == 1;

const bool mismatch =
  llvm::any_of(shapes, [&](const auto &s) {
    return (firstIsOne ^ (s[axis] == 1)); // xor as  we want to trigger the error if only the first or the other is one
  });

if (mismatch)
  return CheckResult{false, msg};
```

mlir/lib/Dialect/Tosa/Transforms/SinkInputOpsThroughConcat.cpp
jorickert: The same is also true for the LeakyRelu decomposition, but this is harder to prevents, as it consists out of multiple ops

mlir/lib/Dialect/Tosa/Transforms/SinkInputOpsThroughConcat.cpp
```diff
@@ -0,0 +1,425 @@
...
+    auto newShape = reshape.getNewShape();
+
+    if (oldShape.size() + concatOp.getAxis() < newShape.size())
+      return rewriter.notifyMatchFailure(
+          concatOp, "Sinking reshape not possible. Concat is on axis that is "
+                    "created by reshape.");
+    auto newAxis = oldShape.size() - (newShape.size() - concatOp.getAxis());
+
+    // we can sink the reshape if is differs only in adding or dropping
+    // dimensions of rank 1
```
jorickert: The comment here and the following loop seem to not have the same behaviour

mlir/test/Dialect/Tosa/sink-input-ops-through-concat.mlir
```diff
@@ -0,0 +1,991 @@
...
+func.func @reshape_no_prefix(%arg0: !in_type, %arg1: !in_type) -> !out_type {
+    %0 = tosa.reshape %arg0 {new_shape = array<i64: 1, 1, 2, 3>} : (!in_type) -> !re_type
+    %1 = tosa.reshape %arg1 {new_shape = array<i64: 1, 1, 2, 3>} : (!in_type) -> !re_type
+    %2 = tosa.concat %0, %1 {axis = 1 : i32} : (!re_type, !re_type) -> !out_type
+  return %2 : !out_type
+}
+
+// CHECK-LABEL:  func.func @reshape_no_prefix
+// CHECK-SAME:   ([[PARAM_0_:%.+]]: tensor<1x1x6xf32>, [[PARAM_1_:%.+]]: tensor<1x1x6xf32>) -> tensor<1x2x2x3xf32> {
+// CHECK-DAG:       [[VAR_0_:%.+]] = tosa.reshape [[PARAM_0_]] {new_shape = array<i64: 1, 1, 2, 3>} : (tensor<1x1x6xf32>) -> tensor<1x1x2x3xf32>
```
jorickert: Why is it not possible to sink the rehsape her?
concat (1x1x6, 1x1x6) to 1x2x6, reshape to 1x2x2x3

jorickert (commented):
Thanks, especially for adding so many tests.

It looks generally good to me, but I will need to think about the reshape again, the current implementation seems to only handle some cases

mlir/lib/Dialect/Tosa/Transforms/SinkInputOpsThroughConcat.cpp
gerion-amd: - For `doGenericChecks`: Subclasses should be able to use this function (`SinkReshapeOp` uses it for example).
- For `doExtraChecks`: Subclasses should/can override this function, if needed.

mlir/lib/Dialect/Tosa/Transforms/SinkInputOpsThroughConcat.cpp
gerion-amd: I think a much greater benefit of the sinking can happen, when the previous op can also be sunk through the concat but with the currect design that triggers only when the activation is sunk first.

mlir/lib/Dialect/Tosa/Transforms/SinkInputOpsThroughConcat.cpp
gerion-amd: Hmm, you're right, seems to not possible to solve with just a `char*`.

mlir/lib/Dialect/Tosa/Transforms/SinkInputOpsThroughConcat.cpp
gerion-amd: I don't completely get what you mean. It's possible to return a generic error like this:
```
if (!do_specific_check(...))
    return rewriter.notifyMatchFailure("Specific check failed")
````
Do you mean this? We would loose the exact failure reason then.

Comment (jorickert) [reply mlir/lib/Dialect/Tosa/Transforms/SinkInputOpsThroughConcat.cpp]: You could already do the `rewriter.notifyMatchFailure` in the specific check

File: mlir/lib/Dialect/Tosa/Transforms/SinkInputOpsThroughConcat.cpp (hunk 5 above)
Comment (gerion-amd) [reply mlir/lib/Dialect/Tosa/Transforms/SinkInputOpsThroughConcat.cpp]: Good idea, will do.

File: mlir/lib/Dialect/Tosa/Transforms/SinkInputOpsThroughConcat.cpp (hunk 8 above)
Comment (gerion-amd) [reply mlir/lib/Dialect/Tosa/Transforms/SinkInputOpsThroughConcat.cpp]: No, the loop reverses the order how the operands are saved. I did an additional comment in the code for clarification:
```
    // to visualize the process
    //
    // A -.
    // B -- (pOp1) --.
    // C -*          |- (concatOp)
    // D -.          |
    // E -- (pOp2) --*
    // F -*
    // 
    // The for loops traverse it in this order: ((A, B, C), (D, E, F))
    // It is assigned in reverse order in the vectors:
    // [[A, D], [B, E], [C, F]]
```

File: mlir/lib/Dialect/Tosa/Transforms/SinkInputOpsThroughConcat.cpp (hunk 9 above)
Comment (gerion-amd) [reply mlir/lib/Dialect/Tosa/Transforms/SinkInputOpsThroughConcat.cpp]: That is much better! Thanks.

File: mlir/lib/Dialect/Tosa/Transforms/SinkInputOpsThroughConcat.cpp (hunk 4 above)
Comment (gerion-amd) [reply mlir/lib/Dialect/Tosa/Transforms/SinkInputOpsThroughConcat.cpp]: The return value then would be `LogicalResult`, which is easily checkable for success. Good idea, will do!

File: mlir/lib/Dialect/Tosa/Transforms/SinkInputOpsThroughConcat.cpp (hunk 11 above)
Comment (gerion-amd) [reply mlir/lib/Dialect/Tosa/Transforms/SinkInputOpsThroughConcat.cpp]: Kind of: The loop checks if one of the shape is a suffix of the other one leaving behind a prefix with just dimensions of rank 1 otherwise the reshape would not be valid.
Do you think, the code is valid as is? Then I would just improve the comment.

File: mlir/test/Dialect/Tosa/sink-input-ops-through-concat.mlir (hunk 12 above)
Comment (gerion-amd) [reply mlir/test/Dialect/Tosa/sink-input-ops-through-concat.mlir]: You're right. The current transformation does not work because 1x1x6 and 1x1x2x3 have no common suffix. It is "too" restrictive currently.

**Comment by gerion-amd** [comment]
> It looks generally good to me, but I will need to think about the reshape again, the current implementation seems to only handle some cases

This is correct. Only reshapes that extend or lessen the shape with dimensions of rank 1 will be transformed. This seemed to be enough for real-world tests, so I left it that way, with the possibility to handle more advanced cases later. Maybe I can add a comment explaining this.

File: mlir/lib/Dialect/Tosa/Transforms/SinkInputOpsThroughConcat.cpp (hunk 10 above)
Comment (gerion-amd) [reply mlir/lib/Dialect/Tosa/Transforms/SinkInputOpsThroughConcat.cpp]: Good idea. This should be even simpler by counting the ones:
```
    const auto axis = concat.getAxis();
    // check that the ranks on the axis that the concat uses is not one for all
    // ops or one for all ops
    const auto oneDimensions = llvm::count(
        llvm::map_range(shapes, [&](const auto &s) { return s[axis] == 1; }),
        true);
    if (oneDimensions == shapes.size() || oneDimensions == 0) {
      return CheckResult{false, msg};
    }
```

Comment (jorickert) [reply mlir/lib/Dialect/Tosa/Transforms/SinkInputOpsThroughConcat.cpp]: Even simpler by using `count_if` instead of map_range

Comment (gerion-amd) [reply mlir/lib/Dialect/Tosa/Transforms/SinkInputOpsThroughConcat.cpp]: Will do!

File: mlir/lib/Dialect/Tosa/Transforms/SinkInputOpsThroughConcat.cpp (hunk 3 above)
Comment (gerion-amd) [reply mlir/lib/Dialect/Tosa/Transforms/SinkInputOpsThroughConcat.cpp]: BTW: The rewriter also want to have a `const char*` which would also point to local memory as long as I do formatting at runtime. However, there is absolutely no need for that (everything should be known at compile time). `std::string` is capable of constexpr formatting only from C++ 20 on. Do you know an easy way to the above formatting at compile time?

Comment (jorickert) [reply mlir/lib/Dialect/Tosa/Transforms/SinkInputOpsThroughConcat.cpp]: Mhh, the notify method and `Diagnostic` taking the arguments as char* is a bit unfortunate.
I still think its fine in your case, as the `notifyMatchFailure ` in the greedy rewriter does create a new `Diagnostic` for every call to `notifyMatchFailure`. And during the call the argument is valid

File: mlir/lib/Dialect/Tosa/Transforms/SinkInputOpsThroughConcat.cpp (hunk 4 above)
Comment (gerion-amd) [reply mlir/lib/Dialect/Tosa/Transforms/SinkInputOpsThroughConcat.cpp]: I use a `std::variant<LogicalResult, OpT>` now for the generic checks, and `LogicalResult` for the specific checks. Works well, I think.

**Comment by gerion-amd** [comment]
All comments should be addressed now.

File: mlir/lib/Dialect/Tosa/Transforms/SinkInputOpsThroughConcat.cpp (hunk 1 above)
Comment (jorickert) [reply mlir/lib/Dialect/Tosa/Transforms/SinkInputOpsThroughConcat.cpp]: Yeah, but we can not always sink the other ops.
I am fine with how it is right now, if it causes problems downstream we can still adjust it

File: mlir/lib/Dialect/Tosa/Transforms/SinkInputOpsThroughConcat.cpp (hunk 2 above)
Comment (jorickert) [reply mlir/lib/Dialect/Tosa/Transforms/SinkInputOpsThroughConcat.cpp:169]: Two notes: Its possible for subclasses to override functions, even if they are private (and IMO thats a quite useful pattern in C++).
These check functions do not access any class members, so would not need to members themselves. But these are more stylistic preferences, so I'm fine with how it is.

File: mlir/lib/Dialect/Tosa/Transforms/SinkInputOpsThroughConcat.cpp (hunk 8 above)
Comment (jorickert) [reply mlir/lib/Dialect/Tosa/Transforms/SinkInputOpsThroughConcat.cpp]: I still do not get it.
You have a nested vector. The outer vectors size is the number of operands of each op, and the inner vectors size is the number of operands of the original concat.

In the outer loop you iterate over all concat operands, and in the inner loop you iterate over their operands.
So I do not understand the condition: ` if (concatOperands.size() <= j) {` this can only happen if not all ops have the same number of operands, but this is not supported by the pass in general. I do not think this condition is needed.

File: mlir/lib/Dialect/Tosa/Transforms/SinkInputOpsThroughConcat.cpp (hunk 2 above)
Comment (gerion-amd) [reply mlir/lib/Dialect/Tosa/Transforms/SinkInputOpsThroughConcat.cpp:169]: The generic checks are called from the subclass, so it would not be possible to make it private. `doSpecificChecks` could be private but I see no advantage. It's whole purpose is to be specialized in a subclass.

Subclasses that implement `doExtraChecks` mainly just "encode" the types of checks that are needed. Therefore, implementing the checks as class method bound these checks to the class name (this is purely stylistic).

File: mlir/lib/Dialect/Tosa/Transforms/SinkInputOpsThroughConcat.cpp (hunk 8 above)
Comment (gerion-amd) [reply mlir/lib/Dialect/Tosa/Transforms/SinkInputOpsThroughConcat.cpp]: This is just to construct the initial vector with the correct size (similar to Python's `defaultdict`). So, `concatOperands` is constructed with the memory reserved for the number of operands of each op (at least, I hope so), but it has size 0 at the beginning. Therefore, only to construct the inner vector with a correct size, I use the if condition. Without the if condition, `operator[` would construct the vector, when it's used initially.

Maybe an array like structure would be better to save allocations since all dimensions are known beforehand, but it would need to be a dynamic (C-)array, so I went with SmallVector.

File: mlir/lib/Dialect/Tosa/Transforms/SinkInputOpsThroughConcat.cpp (hunk 11 above)
Comment (jorickert) [reply mlir/lib/Dialect/Tosa/Transforms/SinkInputOpsThroughConcat.cpp]: Regarding the logic of the code:
To sink the reshape through the concat, the following needs to hold:

1. The reshape does not modify the concatenated axis
2. The input and output shape of all reshapes needs to be the same, except for the concat axis, which cha have different sizes for the different reshapes.

I think directly checking for these conditions is clearer and more generic

Comment (gerion-amd) [reply mlir/lib/Dialect/Tosa/Transforms/SinkInputOpsThroughConcat.cpp]: I don't get what you mean exactly.

But I also used the work "rank" in a wrong meaning beforehand (as I just learned). I try to reformulate the current restrictions. Let's assume a `tensor<1x1x1x3>`. It has 4 dimensions for which the first three are of "size" 1 and the last one of size 3 (I used "rank" instead of "size" in my previous comment). Then my transformation would accept any reshape operation that reshapes this tensor to a shape that has more dimensions of size 1 as a prefix (e.g. `tensor<1x1x1x1x1x3>`) or that has fewer dimensions (e.g. `tensor<1x3>`) as long as every "deleted" dimension was of size 1.

There is one additional condition: The concat must happen on an axis that existed before the reshape (looking at the shape from the end). So the reshape is not sunk , if the concat is on axis 0 and the reshape modifies `tensor<1x5>` into `tensor<1x1x5>`.

For your first constraint: If this is for a generic reshape, I cannot say what the concatenated axis would be before the reshape. Is there a mapping between the shape before and after the reshape?

Comment (jorickert) [reply mlir/lib/Dialect/Tosa/Transforms/SinkInputOpsThroughConcat.cpp]: > For your first constraint: If this is for a generic reshape, I cannot say what the concatenated axis would be before the reshape. Is there a mapping between the shape before and after the reshape?

You can check it by doing the following algorithm:

On the result of the reshape: Calculate the product of dimensions that are more outermost/left then then concat dim.
On the input of the reshape: Multiply dimensions, starting from the outermost/left until you hit the product you calculated for the reshape result. If the next dim has the same size as the concat dim, you found the matching dim on the input. If the size is not the same, or the products do not match, the concat dim got modified by the reshape

File: mlir/lib/Dialect/Tosa/Transforms/SinkInputOpsThroughConcat.cpp (hunk 4 above)
Comment (jorickert) [reply mlir/lib/Dialect/Tosa/Transforms/SinkInputOpsThroughConcat.cpp]: Why not use `FailureOr<OpT>` ?

File: mlir/lib/Dialect/Tosa/Transforms/SinkInputOpsThroughConcat.cpp (hunk 11 above)
Comment (gerion-amd) [reply mlir/lib/Dialect/Tosa/Transforms/SinkInputOpsThroughConcat.cpp]: This makes sense. I'll implement it.
Edit: Done

File: mlir/lib/Dialect/Tosa/Transforms/SinkInputOpsThroughConcat.cpp (hunk 4 above)
Comment (gerion-amd) [reply mlir/lib/Dialect/Tosa/Transforms/SinkInputOpsThroughConcat.cpp]: I wasn't aware that this is an existing type. I thought that you introduced it as a suggestion for a class that could be returned. I changed it now to `FailureOr`.

File: mlir/lib/Dialect/Tosa/Transforms/SinkInputOpsThroughConcat.cpp [hunk 13]
```diff
@@ -0,0 +1,423 @@
...
+    RewritePatternSet patterns(func.getContext());
+    MLIRContext *ctx = func.getContext();
+    // elementwise
+    patterns.add<
+        SinkSpecificOp<tosa::AbsOp>, SinkSpecificOp<tosa::BitwiseNotOp>,
+        SinkSpecificOp<tosa::CeilOp>, SinkSpecificOp<tosa::ClampOp>,
+        SinkSpecificOp<tosa::ClzOp>, SinkSpecificOp<tosa::CosOp>,
+        SinkSpecificOp<tosa::EqualOp>, SinkSpecificOp<tosa::ErfOp>,
+        SinkSpecificOp<tosa::ExpOp>, SinkSpecificOp<tosa::FloorOp>,
+        SinkSpecificOp<tosa::GreaterOp>, SinkSpecificOp<tosa::GreaterEqualOp>,
```
jorickert: tosa.select and the comparison ops are broadcasting too

gerion-amd: I got the information about broadcast from the [LLVM documentation](https://mlir.llvm.org/docs/Dialects/TOSA), where it is explicitly mentioned in the description. Apparently, this is not complete. Can I use the trait `ResultsBroadcastableShape` (but this one is set also for unary operations)?

jorickert: I think using this trait is a good idea. That its also set for unary ops should not cause any problems

mlir/include/mlir/Dialect/Tosa/Transforms/Passes.td
```diff
@@ -142,4 +143,15 @@ def TosaReduceTransposes : Pass<"tosa-reduce-transposes", "func::FuncOp"> {
   }];
 }
 
+def SinkInputOpsThroughConcat : Pass<"sink-input-ops-through-concat", "mlir::ModuleOp"> {
+  let summary = "Sinks same operation through a Concat operation";
+  let description = [{
+    Pass that sinks the same operation through a concatenation, simplifying
+    later optimizations.
```
ljfitz, line 156: Please add a simple example to help people understand.

mlir/include/mlir/Dialect/Tosa/Transforms/Passes.td
```diff
@@ -142,4 +143,15 @@ def TosaReduceTransposes : Pass<"tosa-reduce-transposes", "func::FuncOp"> {
...
+
+    Only selected tosa operations that proove as "same" and "mergable" got sunk.
+    On a high level, that are all operations that do not interfere with the concat.
```
ljfitz: This is redundant as a transformation is not allowed to break the semantics of what it transforms.

It would be useful instead to say that it is greedy and doesn't do any cost-benefit assessment,

mlir/include/mlir/Dialect/Tosa/Transforms/Passes.td
gerion-amd: Good point. I've added the same example/picture than in the comments of the cpp file.

mlir/include/mlir/Dialect/Tosa/Transforms/Passes.td
gerion-amd: I have added it and also that the pass only works for a selected set of tosa operations.

mlir/lib/Dialect/Tosa/Transforms/SinkInputOpsThroughConcat.cpp
gerion-amd: I left it now that way.

mlir/test/Dialect/Tosa/sink-input-ops-through-concat.mlir
```diff
@@ -0,0 +1,991 @@
...
+// CHECK-DAG:       [[VAR_0_:%.+]] = tosa.add [[PARAM_0_]], [[PARAM_2_]] : (tensor<1x8x8xf32>, tensor<1x1x8xf32>) -> tensor<1x8x8xf32>
+// CHECK-DAG:       [[VAR_1_:%.+]] = tosa.add [[PARAM_1_]], [[PARAM_2_]] : (tensor<1x8x8xf32>, tensor<1x1x8xf32>) -> tensor<1x8x8xf32>
+// CHECK:           [[VAR_2_:%.+]] = tosa.concat [[VAR_0_]], [[VAR_1_]] {axis = 1 : i32} : (tensor<1x8x8xf32>, tensor<1x8x8xf32>) -> tensor<1x16x8xf32>
+// CHECK:           return [[VAR_2_]] : tensor<1x16x8xf32>
+// CHECK:         }
+// -----
+
+!in_type = tensor<1x8x8xf32>
+!out_type = tensor<2x8x8xf32>
+func.func @switch_op_concat_inmergable_broadcasting2(%arg0: !in_type, %arg1: !in_type, %broadcast: tensor<1x1x8xf32>) -> !out_type {
```
ehsan-toosi:
```suggestion
func.func @switch_op_concat_unmergable_broadcasting2(%arg0: !in_type, %arg1: !in_type, %broadcast: tensor<1x1x8xf32>) -> !out_type {
```

mlir/test/Dialect/Tosa/sink-input-ops-through-concat.mlir
```diff
@@ -0,0 +1,991 @@
...
+  %0 = tosa.add %arg0, %broadcast: (!in_type, tensor<1x1x8xf32>) -> !in_type
+  %1 = tosa.add %arg1, %arg1: (!in_type, !in_type) -> !in_type
+  %2 = tosa.concat %0, %1 {axis = 0 : i32} : (!in_type, !in_type) -> !out_type
+  return %2 : !out_type
+}
+
+// mlir2FileCheck.py
```
ehsan-toosi:
```suggestion
```

mlir/test/Dialect/Tosa/sink-input-ops-through-concat.mlir
ehsan-toosi: Isn't it better to move the sink down for reshape to the pass implemented by @jorickert ?

mlir/lib/Dialect/Tosa/Transforms/SinkInputOpsThroughConcat.cpp
gerion-amd: You were right, the vector is not only preallocated but also preconstructed. I removed the lines, now.

mlir/test/Dialect/Tosa/sink-input-ops-through-concat.mlir
```diff
@@ -0,0 +1,991 @@
...
+  return %2 : !out_type
+}
+
+// CHECK-LABEL:  func.func @reshape_wrong_axis
+// CHECK-SAME:   ([[PARAM_0_:%.+]]: tensor<1x1x5xf32>, [[PARAM_1_:%.+]]: tensor<1x1x5xf32>) -> tensor<2x1x1x5xf32> {
+// CHECK-DAG:       [[VAR_0_:%.+]] = tosa.reshape [[PARAM_0_]] {new_shape = array<i64: 1, 1, 1, 5>} : (tensor<1x1x5xf32>) -> tensor<1x1x1x5xf32>
+// CHECK-DAG:       [[VAR_1_:%.+]] = tosa.reshape [[PARAM_1_]] {new_shape = array<i64: 1, 1, 1, 5>} : (tensor<1x1x5xf32>) -> tensor<1x1x1x5xf32>
+// CHECK:           [[VAR_2_:%.+]] = tosa.concat [[VAR_0_]], [[VAR_1_]] {axis = 0 : i32} : (tensor<1x1x1x5xf32>, tensor<1x1x1x5xf32>) -> tensor<2x1x1x5xf32>
+// CHECK:           return [[VAR_2_]] : tensor<2x1x1x5xf32>
+// CHECK:         }
```
ehsan-toosi: Why this one isn't allowed? you should be able to sink it down.

mlir/test/Dialect/Tosa/sink-input-ops-through-concat.mlir
```diff
@@ -0,0 +1,991 @@
...
+// CHECK-DAG:       [[VAR_9_:%.+]] = tosa.transpose [[VAR_8_]], [[VAR_0_]] : (tensor<1x12x42xbf16>, tensor<3xi32>) -> tensor<1x42x12xbf16>
+// CHECK-DAG:       [[VAR_10_:%.+]] = tosa.concat [[VAR_3_]], [[VAR_7_]] {axis = 3 : i32} : (tensor<1x1x1x1x1x42xbf16>, tensor<1x1x1x1x1x42xbf16>) -> tensor<1x1x1x2x1x42xbf16>
+// CHECK-NOT: separator of consecutive DAGs
+// CHECK-DAG:       [[VAR_11_:%.+]] = tosa.reshape [[VAR_10_]] {new_shape = array<i64: 2, 1, 42>} : (tensor<1x1x1x2x1x42xbf16>) -> tensor<2x1x42xbf16>
+// CHECK-DAG:       [[VAR_12_:%.+]] = tosa.concat [[VAR_5_]], [[VAR_9_]] {axis = 0 : i32} : (tensor<1x42x12xbf16>, tensor<1x42x12xbf16>) -> tensor<2x42x12xbf16>
+// CHECK:           [[VAR_13_:%.+]] = tosa.matmul [[VAR_11_]], [[VAR_12_]] : (tensor<2x1x42xbf16>, tensor<2x42x12xbf16>) -> tensor<2x1x12xbf16>
+// CHECK:           [[VAR_14_:%.+]] = tosa.reshape [[VAR_13_]] {new_shape = array<i64: 1, 2, 1, 12>} : (tensor<2x1x12xbf16>) -> tensor<1x2x1x12xbf16>
+// CHECK:           return [[VAR_14_]] : tensor<1x2x1x12xbf16>
+// CHECK:         }
+// -----
```
ehsan-toosi, line 419: What is this test doing? could you write an explanation/comment please? it's hard to follow the IR.

mlir/test/Dialect/Tosa/sink-input-ops-through-concat.mlir
```diff
@@ -0,0 +1,991 @@
...
+}
+// CHECK-LABEL:  func.func @switch_op_concat
+// CHECK-SAME:   ([[PARAM_0_:%.+]]: tensor<1x8x8xi1>, [[PARAM_1_:%.+]]: tensor<1x8x8xf32>, [[PARAM_2_:%.+]]: tensor<1x8x8xf32>) -> tensor<2x8x8xf32> {
+// CHECK-DAG:       [[VAR_0_:%.+]] = tosa.concat [[PARAM_0_]], [[PARAM_0_]] {axis = 0 : i32} : (tensor<1x8x8xi1>, tensor<1x8x8xi1>) -> tensor<2x8x8xi1>
+// CHECK-DAG:       [[VAR_1_:%.+]] = tosa.concat [[PARAM_1_]], [[PARAM_2_]] {axis = 0 : i32} : (tensor<1x8x8xf32>, tensor<1x8x8xf32>) -> tensor<2x8x8xf32>
+// CHECK-DAG:       [[VAR_2_:%.+]] = tosa.concat [[PARAM_2_]], [[PARAM_1_]] {axis = 0 : i32} : (tensor<1x8x8xf32>, tensor<1x8x8xf32>) -> tensor<2x8x8xf32>
+// CHECK:           [[VAR_3_:%.+]] = tosa.select [[VAR_0_]], [[VAR_1_]], [[VAR_2_]] : (tensor<2x8x8xi1>, tensor<2x8x8xf32>, tensor<2x8x8xf32>) -> tensor<2x8x8xf32>
+// CHECK:           return [[VAR_3_]] : tensor<2x8x8xf32>
+// CHECK:         }
+// -----
```
ehsan-toosi, line 788: Can you add a test with select with different first operand?

mlir/lib/Dialect/Tosa/Transforms/SinkInputOpsThroughConcat.cpp
```diff
@@ -0,0 +1,419 @@
...
+    // reduce
+    patterns
+        .add<SinkReduceOp<tosa::ReduceAllOp>, SinkReduceOp<tosa::ReduceAnyOp>,
+             SinkReduceOp<tosa::ReduceMaxOp>, SinkReduceOp<tosa::ReduceMinOp>,
+             SinkReduceOp<tosa::ReduceProdOp>, SinkReduceOp<tosa::ReduceSumOp>>(
+            ctx, 2);
+    // others
+    patterns.add<SinkMatmulOp>(ctx, 2);
+    patterns.add<SinkReshapeOp>(ctx, 2);
+    patterns.add<SinkGenericOp>(ctx, 1, stats, os);
```
ehsan-toosi: Can you move these to `populateSinkInputOpsThroughConcatPatterns` function and expose it?

mlir/lib/Dialect/Tosa/Transforms/SinkInputOpsThroughConcat.cpp
```diff
@@ -0,0 +1,419 @@
...
+                 SinkElementwiseBroadcastableOp<tosa::BitwiseAndOp>,
+                 SinkElementwiseBroadcastableOp<tosa::BitwiseOrOp>,
+                 SinkElementwiseBroadcastableOp<tosa::BitwiseXorOp>,
+                 SinkElementwiseBroadcastableOp<tosa::IntDivOp>,
+                 SinkElementwiseBroadcastableOp<tosa::LogicalLeftShiftOp>,
+                 SinkElementwiseBroadcastableOp<tosa::LogicalRightShiftOp>,
+                 SinkElementwiseBroadcastableOp<tosa::LogicalXorOp>,
+                 SinkElementwiseBroadcastableOp<tosa::MaximumOp>,
+                 SinkElementwiseBroadcastableOp<tosa::MinimumOp>,
+                 SinkElementwiseBroadcastableOp<tosa::PowOp>,
```
ehsan-toosi: It would be really good if you define traits for the tosa ops and then rewrite them using
`OpTraitRewritePattern`

ehsan-toosi (commented):
I did an high-level review and checked the transformation. Just one major comment which was using trait as you and Jonas already discussed as well. 
I'll do the detailed code review if I find some time today.

mlir/test/Dialect/Tosa/sink-input-ops-through-concat.mlir
gerion-amd: Good catch! Will fix.

mlir/test/Dialect/Tosa/sink-input-ops-through-concat.mlir
gerion-amd: I implemented the suggested reshape sink down. This is now allowed.

mlir/test/Dialect/Tosa/sink-input-ops-through-concat.mlir
gerion-amd: This is implemented and working now.

mlir/test/Dialect/Tosa/sink-input-ops-through-concat.mlir
gerion-amd: You mean, not using `%arg0` in both cases? Since the transformation currently concatenates `%arg0` two times, it doesn't "know" or check that the operand is different. The outcome would be the same. Do I miss a transformation that should not be valid?

mlir/lib/Dialect/Tosa/Transforms/SinkInputOpsThroughConcat.cpp
jorickert: @ehsan-toosi What do you think about this?

mlir/lib/Dialect/Tosa/Transforms/SinkInputOpsThroughConcat.cpp
gerion-amd: Is this possible? I'm matching only `tosa::concatOp`. `OpTraitRewritePattern` seems to work for transformations that what to match a whole trait.

mlir/lib/Dialect/Tosa/Transforms/SinkInputOpsThroughConcat.cpp
gerion-amd: Do you mean a private class method? I have done it now.

mlir/lib/Dialect/Tosa/Transforms/SinkInputOpsThroughConcat.cpp
gerion-amd: Trait applied manually. I left unary operations without the check.

mlir/test/Dialect/Tosa/sink-input-ops-through-concat.mlir
gerion-amd: I added a comment now.

mlir/lib/Dialect/Tosa/Transforms/SinkInputOpsThroughConcat.cpp
ehsan-toosi: @jorickert This is actually a very solid point. This is gonna break all fusion when there is conv and relu/lrelu and concat situations. We should find a unit test that has conv->relu->concat and show that is better if we move relu after concat? but I'm suspicious that we are going underperform. I agree with @jorickert that we have to disable it for ReLU and LeakyReLU until we have the support to sinkdown for conv as well. After that we can enable these two ops as well again.

mlir/lib/Dialect/Tosa/Transforms/SinkInputOpsThroughConcat.cpp
```diff
@@ -0,0 +1,430 @@
...
+    auto amount = ++stats[opName];
+    if (amount == 1 || amount % 50 == 0)
+      os << "SinkInputOpsThroughConcat: Operation " << opName
+         << " -> Matched amount: " << amount << "\n";
+    return rewriter.notifyMatchFailure(concatOp,
+                                       "Only for statistics printing");
+  }
+
+private:
+  llvm::StringMap<unsigned> &stats;
```
roberteg16: nit:
``` suggestion
  llvm::StringMap<unsigned> &operationFrequency;
```

mlir/lib/Dialect/Tosa/Transforms/SinkInputOpsThroughConcat.cpp
```diff
@@ -0,0 +1,430 @@
...
+      if (!val.hasOneUse())
+        return rewriter.notifyMatchFailure(
+            concatOp, "Operands must just connect to this concat.");
+
+      auto op = val.getDefiningOp<OpT>();
+      if (!op) {
+        std::string err;
+        llvm::raw_string_ostream errStream(err);
+        errStream << "Operand is not a " << OpT::getOperationName();
+        return rewriter.notifyMatchFailure(concatOp, err.c_str());
```
roberteg16: `notifyMatchFailure` accepts a `llvm::Twine` which means you should be able to do this:

``` suggestion
        return rewriter.notifyMatchFailure(concatOp, Twine("Operand is not a ") + OpT::getOperationName());
```

mlir/lib/Dialect/Tosa/Transforms/SinkInputOpsThroughConcat.cpp
```diff
@@ -0,0 +1,430 @@
...
+
+  LogicalResult matchAndRewrite(tosa::ConcatOp concatOp,
+                                PatternRewriter &rewriter) const override {
+    auto reshapeOrError = doGenericChecks(concatOp, rewriter);
+    if (failed(reshapeOrError))
+      return reshapeOrError;
+    auto reshape = *reshapeOrError;
+
+    assert(reshape->getNumOperands() == 1 &&
+           "shape must have exactly one operand");
```
roberteg16: I think you can drop this, it is guaranteed by the operation

mlir/lib/Dialect/Tosa/Transforms/SinkInputOpsThroughConcat.cpp
```diff
@@ -0,0 +1,430 @@
...
+    const auto opType = reshape->getOperandTypes().front();
```
roberteg16: Does:
``` suggestion
    Type opType = reshape.getInput1().getType();
```

do the trick?

mlir/lib/Dialect/Tosa/Transforms/SinkInputOpsThroughConcat.cpp
```diff
@@ -0,0 +1,430 @@
...
+    const auto tenType = dyn_cast<ShapedType>(opType);
+    if (!tenType || !tenType.hasStaticShape())
+      return rewriter.notifyMatchFailure(
+          concatOp, "Dynamic shapes for reshapes are not supported.");
+    const auto oldShape = tenType.getShape();
+    const auto newShape = reshape.getNewShape();
```
roberteg16: If type is not obvious please do not use auto:  https://llvm.org/docs/CodingStandards.html#id30

``` suggestion
    ArrayRef<int64_t> oldShape = tenType.getShape();
    ArrayRef<int64_t> newShape = reshape.getNewShape();
```

mlir/lib/Dialect/Tosa/Transforms/SinkInputOpsThroughConcat.cpp
```diff
@@ -0,0 +1,430 @@
...
+    int64_t newPrefixProduct = 1;
+    for (size_t i = 0; i < newConcatAxis; ++i) {
+      newPrefixProduct *= newShape[i];
+    }
+
+    int64_t oldPrefixProduct = 1;
+    size_t oldConcatAxis = oldShape.size(); // default to invalid value
+    size_t i = 0;
+    assert(oldShape.size() > 0 &&
+           "Input sensor of rank 0 does not make sense here.");
```
roberteg16: Please, failure out, do not assert if you don't know when a scenario can happen

mlir/lib/Dialect/Tosa/Transforms/SinkInputOpsThroughConcat.cpp
```diff
@@ -0,0 +1,430 @@
...
+    // reduce
+    patterns
+        .add<SinkReduceOp<tosa::ReduceAllOp>, SinkReduceOp<tosa::ReduceAnyOp>,
+             SinkReduceOp<tosa::ReduceMaxOp>, SinkReduceOp<tosa::ReduceMinOp>,
+             SinkReduceOp<tosa::ReduceProdOp>, SinkReduceOp<tosa::ReduceSumOp>>(
+            ctx, 2);
+    // others
+    patterns.add<SinkMatmulOp>(ctx, 2);
+    patterns.add<SinkReshapeOp>(ctx, 2);
+    patterns.add<SinkGenericOp>(ctx, 1, stats, os);
```
roberteg16: can you add `/*benefit=*/` in front of the magic numbers? IDEs are really nice nowadays but when looking into the code from outside (github for instance) you don't get any hits

mlir/lib/Dialect/Tosa/Transforms/SinkInputOpsThroughConcat.cpp
```diff
@@ -0,0 +1,430 @@
...
+  }
+
+  llvm::StringMap<unsigned> stats;
```
roberteg16:
```suggestion
  llvm::StringMap<unsigned> operationFrequency;
```

roberteg16 (commented):
Looks very nice! I have yet to finish the review of the reshape, I'll come back ASAP

mlir/lib/Dialect/Tosa/Transforms/SinkInputOpsThroughConcat.cpp
gerion-amd: Sure. I thought to remember that there was a clang-format option for that to insert it automatically, but I cannot find it anymore.

mlir/lib/Dialect/Tosa/Transforms/SinkInputOpsThroughConcat.cpp
gerion-amd: I do that now. Just curious: Is it possible to do a reshape of a tensor of rank 0? I mean, it cannot be meaningful reshaped to something else.

mlir/lib/Dialect/Tosa/Transforms/SinkInputOpsThroughConcat.cpp
gerion-amd: This is why I inserted it as an assertion. Breaking the condition would directly result in a segfault the line afterwards and currently it's just a semantic constraint. The assertion converts it to a runtime constraint.

I currently see no scenario where this would be wrong, but maybe in the future there might be one, or maybe there is a possibility to misuse the APIs in a way that the assertion triggers (weird casts etc.).
As far as I know, assertions are noops in release builds, aren't they?

Edit: By using `reshape.getInput1()`, this is meaningless. I removed the assertion.

mlir/lib/Dialect/Tosa/Transforms/SinkInputOpsThroughConcat.cpp
roberteg16: I would check if `tensor<bf16>` counts as rank 0. If so then I guess you can reshape to `tensor<1xbf16>` or any other rank

mlir/lib/Dialect/Tosa/Transforms/SinkInputOpsThroughConcat.cpp
roberteg16: I see, perhaps by using `Type opType = reshape.getInput1().getType();` as the suggestion just below you can remove it since you are always getting the input.

mlir/lib/Dialect/Tosa/Transforms/SinkInputOpsThroughConcat.cpp
```diff
@@ -0,0 +1,430 @@
...
+      for (auto val : producerOp->getOperands()) {
+        concatOperands.emplace_back(val);
+      }
+    }
+    auto concatNew = rewriter.create<tosa::ConcatOp>(
+        concatOp.getLoc(), concatOperands, oldConcatAxis);
+    // calculate new shape for reshape by combining the shape of the concat with
+    // the remaining prefix of the old reshape
+    auto concatResult = concatNew->getResultTypes();
+    auto concatType = concatResult.front();
```
roberteg16:
```suggestion
    Type concatType = concatResult.getType();
```

mlir/lib/Dialect/Tosa/Transforms/SinkInputOpsThroughConcat.cpp
```diff
@@ -0,0 +1,430 @@
...
+    auto concatShapeT = dyn_cast<ShapedType>(concatType);
+    assert(concatShapeT && "op and thus concat must have a shape");
```
roberteg16: You can accomplish the same with:

```suggestion
    auto concatShapeT = cast<ShapedType>(concatType);
```

Cast asserts internally

mlir/lib/Dialect/Tosa/Transforms/SinkInputOpsThroughConcat.cpp
```diff
@@ -0,0 +1,430 @@
...
+           "shape must have exactly one operand");
+    const auto opType = reshape->getOperandTypes().front();
+    const auto tenType = dyn_cast<ShapedType>(opType);
+    if (!tenType || !tenType.hasStaticShape())
+      return rewriter.notifyMatchFailure(
+          concatOp, "Dynamic shapes for reshapes are not supported.");
+    const auto oldShape = tenType.getShape();
+    const auto newShape = reshape.getNewShape();
+
+    const auto newConcatAxis = concatOp.getAxis();
```
roberteg16: no auto:

```suggestion
    const int64_t newConcatAxis = concatOp.getAxis();
```

mlir/lib/Dialect/Tosa/Transforms/SinkInputOpsThroughConcat.cpp
```diff
@@ -0,0 +1,430 @@
...
+    int64_t newPrefixProduct = 1;
+    for (size_t i = 0; i < newConcatAxis; ++i) {
+      newPrefixProduct *= newShape[i];
+    }
+
+    int64_t oldPrefixProduct = 1;
+    size_t oldConcatAxis = oldShape.size(); // default to invalid value
```
roberteg16: I find choosing `oldShape.size()` as "invalid state" confusing. What do you thing of:
```suggestion
    std::optional<int64_t> oldConcatAxis;
```

mlir/lib/Dialect/Tosa/Transforms/SinkInputOpsThroughConcat.cpp
roberteg16: Also, I think it would be nicer if you rename it to `reshapeInType`. I believe `opType` is quite cryptic

mlir/test/Dialect/Tosa/sink-input-ops-through-concat.mlir
```diff
@@ -0,0 +1,991 @@
...
+
+!in_type = tensor<1x1x5xf32>
+!re_type = tensor<1x1x1x5xf32>
+!out_type = tensor<2x1x1x5xf32>
+func.func @reshape_other_axis(%arg0: !in_type, %arg1: !in_type) -> !out_type {
+    %0 = tosa.reshape %arg0 {new_shape = array<i64: 1, 1, 1, 5>} : (!in_type) -> !re_type
+    %1 = tosa.reshape %arg1 {new_shape = array<i64: 1, 1, 1, 5>} : (!in_type) -> !re_type
+    %2 = tosa.concat %0, %1 {axis = 0 : i32} : (!re_type, !re_type) -> !out_type
+  return %2 : !out_type
+}
```
roberteg16, line 247: Could you add a few more tests for negative cases for reshape? I would like see it failing in different scenarios

roberteg16 (commented):
LGTM, I would like just to have a few more reshape tests

mlir/lib/Dialect/Tosa/Transforms/SinkInputOpsThroughConcat.cpp
gerion-amd: It works and is way better. Thanks for the hint!

mlir/lib/Dialect/Tosa/Transforms/SinkInputOpsThroughConcat.cpp
gerion-amd: Yes, and shorter and safer. Thanks!

mlir/lib/Dialect/Tosa/Transforms/SinkInputOpsThroughConcat.cpp
gerion-amd: Yes, I do this now.

mlir/lib/Dialect/Tosa/Transforms/SinkInputOpsThroughConcat.cpp
gerion-amd: Applied, thanks!

mlir/lib/Dialect/Tosa/Transforms/SinkInputOpsThroughConcat.cpp
gerion-amd: Apparently, `getAxis()` returns an `uint32_t`. Did you chose another type on purpose?

mlir/lib/Dialect/Tosa/Transforms/SinkInputOpsThroughConcat.cpp
gerion-amd: I'm fine with it. Will change.

mlir/test/Dialect/Tosa/sink-input-ops-through-concat.mlir
gerion-amd: Good point. I actually forgot about them (the now "good" tests were the failing tests with the old algorithm prior in this PR).

mlir/lib/Dialect/Tosa/Transforms/SinkInputOpsThroughConcat.cpp
gerion-amd: It's possible to reshape a `tensor<bf16>` to `tensor<1xbf16>`. Sinking down should be always impossible, though, since there is no axis to concat before the reshape. I added a test case for this, too.

mlir/lib/Dialect/Tosa/Transforms/SinkInputOpsThroughConcat.cpp
roberteg16: Not really, I just thought it would be `int64_t`

gerion-amd: Ok, I have changed to `uint32_t` now.

mlir/lib/Dialect/Tosa/Transforms/SinkInputOpsThroughConcat.cpp
roberteg16: We can always control this by first doing matching of ops+activation and then latter doing the sinking

gerion-amd:
Changes from 0803705 to ba93cf8:
- Rename the variables with `old` and `new` prefix to `beforeReshape` and `afterReshape` to make it more specific that their position in the operation graph is meant
- Add documention of the reshape method
- Fix a bug for a reshape where the concat axis dimension is not of size 1. The approach is to also include the concat axis dimension size into the product and check for the same dimension before and afterwards.

mlir/lib/Dialect/Tosa/Transforms/SinkInputOpsThroughConcat.cpp
```diff
@@ -0,0 +1,450 @@
...
+
+      if (!llvm::equal(op->getOperandTypes(), sample->getOperandTypes()))
+        return rewriter.notifyMatchFailure(
+            concatOp, "Requires all operand types to be the same");
+
+      if (llvm::any_of(OpT::getAttributeNames(), [&](const auto &name) {
+            return sample->getAttr(name) != op->getAttr(name);
+          }))
+        return rewriter.notifyMatchFailure(
+            concatOp, "Requires all operand attributes to be the same");
```
mgehre-amd, line 195: To compare attributes and properties, you can use
```
lhs->getRawDictionaryAttrs() != rhs->getRawDictionaryAttrs()
||
!lhs->getName().compareOpProperties(lhs->getPropertiesStorage(),
                                          rhs->getPropertiesStorage())
```

mgehre-amd: But I think this check is too aggressive. We are introducing a `ResultNames` attributes on all ops to track their mapping into the original ONNX tensor, and this attribute doesn't imply any functional change. Could you please ignore the `ResultNames` attribute?

mgehre-amd (dismissed):

mlir/lib/Dialect/Tosa/Transforms/SinkInputOpsThroughConcat.cpp
gerion-amd: Thanks for the review! I compared all attributes in the initial version of this PR but later changed it explicitly to just the set returned by `OpT::getAttributeNames`. This should exclude any attributes that does not belong directly to the operation as specified within the dialect. So `ResultNames` should already be excluded.

mgehre-amd: Oh, that's clever. Nice!

ljfitz (dismissed):

jorickert (dismissed):

mlir/lib/Dialect/Tosa/Transforms/SinkInputOpsThroughConcat.cpp
gerion-amd: I'll leave it this way, then.

gerion-amd:
I added a second commit to have options to disable parts of the pass for better testing. This could be extended to optionally disable the relu operations mentioned above, if they get problematic.

jorickert (approved):

ehsan-toosi (approved):

mlir/lib/Dialect/Tosa/Transforms/SinkInputOpsThroughConcat.cpp
ehsan-toosi: not private. Like a function in the header that if somebody wants to use these patterns, they can access it. I'll resolve the comment for now.

ehsan-toosi:
Thanks.

ttjost (approved):

roberteg16 (approved):
Thanks!

PR #672, opened 2025-10-06
[AIEX] Add AA support for stack fixed objects

andcarminati:
This helps the scheduler to generate better code for cases where some parameters are passed by value through the stack.

This has a nice synergy with https://github.com/Xilinx/llvm-aie/pull/669

### DIFF
--- a/llvm/lib/Target/AIE/AIEBaseInstrInfo.cpp
+++ b/llvm/lib/Target/AIE/AIEBaseInstrInfo.cpp
@@ -1448,3 +1448,61 @@ void AIEBaseInstrInfo::insertSelect(MachineBasicBlock &MBB,
       .addReg(Support->getSelectSrcOperand(InputOperands, 1))
       .addReg(Support->getSelectSrcOperand(InputOperands, 2));
 }
+
+bool AIEBaseInstrInfo::areMemAccessesTriviallyDisjoint(
+    const MachineInstr &MIa, const MachineInstr &MIb) const {
+  if (!MIa.hasOneMemOperand() || !MIb.hasOneMemOperand())
+    return false;
+
+  // If mem-operands show that the same address Value is used by both
+  // instructions, check for non-overlapping offsets and widths.
+  const MachineMemOperand *MMOa = *MIa.memoperands_begin();
+  const MachineMemOperand *MMOb = *MIb.memoperands_begin();
+
+  auto CheckOverlapping = [=](int64_t OffsetA, int64_t OffsetB) {
+    const LocationSize WidthA = MMOa->getSize(), WidthB = MMOb->getSize();
+    const int64_t LowOffset = OffsetA < OffsetB ? OffsetA : OffsetB;
+    const int64_t HighOffset = OffsetA < OffsetB ? OffsetB : OffsetA;
+    const LocationSize LowWidth = (LowOffset == OffsetA) ? WidthA : WidthB;
+    return (LowWidth.hasValue() &&
+            LowOffset + (int64_t)LowWidth.getValue() <= HighOffset);
+  };
+
+  const int64_t MMOOffsetA = MMOa->getOffset();
+  const int64_t MMOOffsetB = MMOb->getOffset();
+
+  const Value *VALa = MMOa->getValue();
+  const Value *VALb = MMOb->getValue();
+  const bool SameValue = (VALa && VALb && (VALa == VALb));
+  if (SameValue)
+    return CheckOverlapping(MMOOffsetA, MMOOffsetB);
+
+  const PseudoSourceValue *PSVa = MMOa->getPseudoValue();
+  const PseudoSourceValue *PSVb = MMOb->getPseudoValue();
+
+  const bool ExistBothPseudoSources = PSVa && PSVb;
+  if (!ExistBothPseudoSources)
+    return false;
+
+  const bool SamePseudoSource = PSVa == PSVb;
+  if (SamePseudoSource)
+    return CheckOverlapping(MMOOffsetA, MMOOffsetB);
+
+  const FixedStackPseudoSourceValue *FixedStackA =
+      dyn_cast<FixedStackPseudoSourceValue>(PSVa);
+  const FixedStackPseudoSourceValue *FixedStackB =
+      dyn_cast<FixedStackPseudoSourceValue>(PSVb);
+
+  // If we have different fixed stack objects, we have disjoint access
+  // when offsets are different and we don't have any partial overlap
+  // (SameVal check).
+  const bool ExistsBothFixedStackObjs = FixedStackA && FixedStackB;
+  if (!ExistsBothFixedStackObjs)
+    return false;
+
+  const MachineFrameInfo &MFI = MIa.getMF()->getFrameInfo();
+  const int64_t ObjOffsetA = MFI.getObjectOffset(FixedStackA->getFrameIndex());
+  const int64_t ObjOffsetB = MFI.getObjectOffset(FixedStackB->getFrameIndex());
+
+  return CheckOverlapping(ObjOffsetA, ObjOffsetB);
+}
--- a/llvm/lib/Target/AIE/AIEBaseInstrInfo.h
+++ b/llvm/lib/Target/AIE/AIEBaseInstrInfo.h
@@ -869,6 +869,9 @@ protected:
                     const DebugLoc &dl, Register DestReg,
                     ArrayRef<MachineOperand> Cond, Register TrueReg,
                     Register FalseReg) const override;
+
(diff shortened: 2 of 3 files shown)


llvm/lib/Target/AIE/AIEBaseInstrInfo.cpp
```diff
@@ -1448,3 +1448,56 @@ void AIEBaseInstrInfo::insertSelect(MachineBasicBlock &MBB,
...
+        // If we have different fixed stack objects, we have disjoint access
+        // when offsets are different and we don't have any partial overlap
+        // (SameVal check).
+        if (FixedStackA && FixedStackB &&
+            FixedStackA->getFrameIndex() != FixedStackB->getFrameIndex()) {
+          const MachineFrameInfo &MFI = MIa.getMF()->getFrameInfo();
+          OffsetA = MFI.getObjectOffset(FixedStackA->getFrameIndex());
+          OffsetB = MFI.getObjectOffset(FixedStackB->getFrameIndex());
+          if (OffsetA != OffsetB) {
+            SameVal = true;
```
F-Stuckmann: CHECK: This is a conservative estimate that if the Frameindices are unequal and the offsets are equal they `could` overlap and to be safe we assign them the same Value?

llvm/lib/Target/AIE/AIEBaseInstrInfo.cpp
```diff
@@ -1448,3 +1448,56 @@ void AIEBaseInstrInfo::insertSelect(MachineBasicBlock &MBB,
...
+  int64_t OffsetA = MMOa->getOffset();
+  int64_t OffsetB = MMOb->getOffset();
+  const Value *VALa = MMOa->getValue();
+  const Value *VALb = MMOb->getValue();
+  bool SameVal = (VALa && VALb && (VALa == VALb));
+  if (!SameVal) {
+    const PseudoSourceValue *PSVa = MMOa->getPseudoValue();
+    const PseudoSourceValue *PSVb = MMOb->getPseudoValue();
+    if (PSVa && PSVb) {
+      if (PSVa == PSVb) {
```
F-Stuckmann: We have a lot of nesting levels here. Could we reduce them and add a couple of early exits?

llvm/lib/Target/AIE/AIEBaseInstrInfo.cpp
andcarminati: In this case, if FIs are different and also Offsets are different we use `SameVal` to check overlapping. We assign `SameVal` because they are in fact related to the same origin: stack.

llvm/lib/Target/AIE/AIEBaseInstrInfo.cpp
andcarminati: I rewrote the code accordingly.

llvm/lib/Target/AIE/AIEBaseInstrInfo.cpp
```diff
@@ -1448,3 +1448,61 @@ void AIEBaseInstrInfo::insertSelect(MachineBasicBlock &MBB,
...
+  const Value *VALa = MMOa->getValue();
+  const Value *VALb = MMOb->getValue();
+  const bool SameValue = (VALa && VALb && (VALa == VALb));
+  if (SameValue)
+    return CheckOverlapping(MMOOffsetA, MMOOffsetB);
+
+  const PseudoSourceValue *PSVa = MMOa->getPseudoValue();
+  const PseudoSourceValue *PSVb = MMOb->getPseudoValue();
+
+  const bool WithBothPseudoSources = PSVa && PSVb;
```
F-Stuckmann: nit: ExistBothPseudoSources?

llvm/lib/Target/AIE/AIEBaseInstrInfo.cpp
```diff
@@ -1448,3 +1448,61 @@ void AIEBaseInstrInfo::insertSelect(MachineBasicBlock &MBB,
...
+  if (!WithBothPseudoSources)
+    return false;
+
+  const bool SamePseudoSource = PSVa == PSVb;
+  if (SamePseudoSource)
+    return CheckOverlapping(MMOOffsetB, MMOOffsetB);
```
F-Stuckmann: There is a bug: return CheckOverlapping(MMOOffsetA, MMOOffsetB);

llvm/lib/Target/AIE/AIEBaseInstrInfo.cpp
```diff
@@ -1448,3 +1448,61 @@ void AIEBaseInstrInfo::insertSelect(MachineBasicBlock &MBB,
       .addReg(Support->getSelectSrcOperand(InputOperands, 1))
       .addReg(Support->getSelectSrcOperand(InputOperands, 2));
 }
+
+bool AIEBaseInstrInfo::areMemAccessesTriviallyDisjoint(
+    const MachineInstr &MIa, const MachineInstr &MIb) const {
+  if (!MIa.hasOneMemOperand() || !MIb.hasOneMemOperand())
```
F-Stuckmann, line 1454: I think we could make the first exit check more strict
`if (MIa.hasUnmodeledSideEffects() || MIb.hasUnmodeledSideEffects() || MIa.hasOrderedMemoryRef() || MIb.hasOrderedMemoryRef()) return false;
`

llvm/lib/Target/AIE/AIEBaseInstrInfo.cpp
```diff
@@ -1448,3 +1448,61 @@ void AIEBaseInstrInfo::insertSelect(MachineBasicBlock &MBB,
...
+    return CheckOverlapping(MMOOffsetA, MMOOffsetB);
+
+  const PseudoSourceValue *PSVa = MMOa->getPseudoValue();
+  const PseudoSourceValue *PSVb = MMOb->getPseudoValue();
+
+  const bool WithBothPseudoSources = PSVa && PSVb;
+  if (!WithBothPseudoSources)
+    return false;
+
+  const bool SamePseudoSource = PSVa == PSVb;
```
F-Stuckmann, line 1487: Should we also be checking `PSVa->isAliased()`

F-Stuckmann:
Overall I like the simpler, logical approach.

llvm/lib/Target/AIE/AIEBaseInstrInfo.cpp
andcarminati: True, happily it is not is not optimistic.

llvm/lib/Target/AIE/AIEBaseInstrInfo.cpp
andcarminati: `areMemAccessesTriviallyDisjoint` is related to the memory addresses, memory ordering is handled apart of AA.

llvm/lib/Target/AIE/AIEBaseInstrInfo.cpp
andcarminati: According to the sources:

```
    // If true, an LLVM IR value might point to this object.
    // Normally, spill slots and fixed-offset objects don't alias IR-accessible
    // objects, but there are exceptions (on PowerPC, for example, some byval
    // arguments have ABI-prescribed offsets).
```

Apart of that, I see that this is not relevant to us because we are not comparing against IR objects.

llvm/lib/Target/AIE/AIEBaseInstrInfo.cpp
```diff
@@ -1448,3 +1448,61 @@ void AIEBaseInstrInfo::insertSelect(MachineBasicBlock &MBB,
       .addReg(Support->getSelectSrcOperand(InputOperands, 1))
       .addReg(Support->getSelectSrcOperand(InputOperands, 2));
 }
+
+bool AIEBaseInstrInfo::areMemAccessesTriviallyDisjoint(
```
F-Stuckmann, line 1452: CHECK: this runs on any mayAlias Querry, correct? 
Shouldn't we update the PR name and the commit message?

andcarminati: Yes. But I did not get the point about the name/message. mayAlias is just for Alias analysis.

F-Stuckmann: The commit messages only talk about `fixed stack objects`, but `areMemAccessesTriviallyDisjoint` runs on any mayAlias querry, so also regular pointer disambiguation.

andcarminati: Done!

F-Stuckmann (approved):

PR #675, opened 2025-10-07
[AIE2P] split offsets before postincrment in large Load/Stores 

F-Stuckmann:
Put the zero-offset part last, since it doesn't need the pointer to survive.

Note: This is a preparation PR for GEMM.

Core_Compute_Insn_Count

Core_PMSize_absolute

Core_StackSize_absolute

### DIFF
--- a/llvm/lib/Target/AIE/aie2p/AIE2PInstructionSelector.cpp
+++ b/llvm/lib/Target/AIE/aie2p/AIE2PInstructionSelector.cpp
@@ -1985,7 +1985,9 @@ bool AIE2PInstructionSelector::selectWideG_AIE_LOAD_STORE(
   case AIE2P::G_AIE_POSTINC_STORE:
   case AIE2P::G_AIE_POSTINC_2D_STORE:
   case AIE2P::G_AIE_POSTINC_3D_STORE: {
-    for (unsigned SubRegIdx = 0; SubRegIdx < SplitFactor; ++SubRegIdx) {
+    // Split Offsets first and then perform the postinc.
+    // For consistency, also handle G_STORE in the same manner.
+    for (int SubRegIdx = SplitFactor - 1; SubRegIdx >= 0; SubRegIdx--) {
       const unsigned Offset = SubRegIdx * 64;
       auto Copy = MIB.buildInstr(TargetOpcode::COPY, {SubRegs[SubRegIdx]}, {})
                       .addReg(AMI.SrcDstOp.getReg(), 0,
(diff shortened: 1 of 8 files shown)


llvm/lib/Target/AIE/aie2p/AIE2PInstructionSelector.cpp
```diff
@@ -1985,7 +1985,10 @@ bool AIE2PInstructionSelector::selectWideG_AIE_LOAD_STORE(
   case AIE2P::G_AIE_POSTINC_STORE:
   case AIE2P::G_AIE_POSTINC_2D_STORE:
   case AIE2P::G_AIE_POSTINC_3D_STORE: {
-    for (unsigned SubRegIdx = 0; SubRegIdx < SplitFactor; ++SubRegIdx) {
+    for (unsigned Part = 0; Part < SplitFactor; ++Part) {
+      const unsigned SubRegIdx = AMI.MemI.getOpcode() == AIE2P::G_STORE
```
andcarminati: Why do we need to handle differently `S_STORE`? They will expand to offset stores anyway... If we have a strong objection we could also have a comment.

llvm/lib/Target/AIE/aie2p/AIE2PInstructionSelector.cpp
```diff
@@ -2057,7 +2060,9 @@ bool AIE2PInstructionSelector::selectWideG_AIE_LOAD_STORE(
   case AIE2P::G_AIE_POSTINC_LOAD:
   case AIE2P::G_AIE_POSTINC_2D_LOAD:
   case AIE2P::G_AIE_POSTINC_3D_LOAD: {
-    for (unsigned SubRegIdx = 0; SubRegIdx < SplitFactor; ++SubRegIdx) {
+    for (unsigned Part = 0; Part < SplitFactor; ++Part) {
```
andcarminati: Same comment as the `G_STORE` one.

llvm/lib/Target/AIE/aie2p/AIE2PInstructionSelector.cpp
F-Stuckmann: good catch, we can simplify it

andcarminati:
Do you think we could have another QoR run because of the G_LOAD/G_STORE change?

F-Stuckmann:
Before force pushing the change, I ran QoR to verify nothing degraded.

> Do you think we could have another QoR run because of the G_LOAD/G_STORE change?

andcarminati:
This looks good to me, I only recommend to fix the commit message before approval.

F-Stuckmann:
I updated the commit message

andcarminati (approved):
LGTM.

PR #677, opened 2025-10-07
[AIE2P] add missing Vextr Bcst Pattern and harmonize G_ASSERT_[S/Z]EXT removal. 

F-Stuckmann:
Minor Performance improvements: -0.03%

Core_Compute_Insn_Count

Core_PMSize_absolute

Core_StackSize_absolute


llvm/test/CodeGen/AIE/GlobalISel/postlegalizercombiner-vextr-assert-bcst.mir
```diff
@@ -0,0 +1,68 @@
...
+    ; CHECK-NEXT: [[ASSERT_SEXT:%[0-9]+]]:_(s32) = G_ASSERT_SEXT [[AIE_ZEXT_EXTRACT_VECTOR_ELT]], 16
+    ; CHECK-NEXT: [[ASSERT_ZEXT:%[0-9]+]]:_(s32) = G_ASSERT_ZEXT [[ASSERT_SEXT]], 16
+    ; CHECK-NEXT: [[AIE_BROADCAST_VECTOR:%[0-9]+]]:_(<16 x s32>) = G_AIE_BROADCAST_VECTOR [[ASSERT_ZEXT]](s32)
+    ; CHECK-NEXT: PseudoRET implicit $lr, implicit [[AIE_BROADCAST_VECTOR]](<16 x s32>)
+    %ptr:_(p0) = COPY $p0
+    %4:_(<16 x s32>) = G_LOAD %ptr(p0) :: (load (<16 x s32>))
+    %483:_(s32) = COPY $r6
+    %484:_(s32) = G_AIE_ZEXT_EXTRACT_VECTOR_ELT %4(<16 x s32>), %483(s32)
+    %485:_(s32) = G_ASSERT_SEXT %484, 16
+    %486:_(s32) = G_ASSERT_ZEXT %485, 16
```
konstantinschwarz: Are we really produces this IR? Is looks illegal: the `G_AIE_ZEXT_EXTRACT_VECTOR_ELT ` guarantees to zero-extend the vector element it extracts (or doesn't do any extension if the element type size matches the destination type size).
Having two asserts that claim the same value is both sign and zero extended looks wrong:
```
    %485:_(s32) = G_ASSERT_SEXT %484, 16
    %486:_(s32) = G_ASSERT_ZEXT %485, 16
```
Note that G_ASSERT_SEXT/G_ASSERT_ZEXT does not perform any operation. It simply tells the optimizers that they can remove redundant extensions

F-Stuckmann: Yes we produce it unfortunately. 
The second unit test (VEXTBCST_SEXT_CHAIN) is constructed to test the symmetry of the G_ASSERT_[S/Z]EXT removal.

konstantinschwarz: Can you investigate how it ends up in this situation? I don't think this is correct

llvm/lib/Target/AIE/AIECombinerHelper.cpp
```diff
@@ -1069,7 +1069,6 @@ bool llvm::matchExtractVecEltAndExt(
     return false;
   MachineInstr *ExtMI = &*MRI.use_instr_nodbg_begin(DstReg);
   switch (ExtMI->getOpcode()) {
-  case TargetOpcode::G_ANYEXT:
```
F-Stuckmann, line 1095: Hey @konstantinschwarz this is the case, where the G_ANYEXT inserts a G_ASSERT_SEXT, which is erroneous.

andcarminati:
Hi @F-Stuckmann, I think it is better to autosquash the fixup commits to be better to review.

F-Stuckmann:
I autosquashed the fixups, thanks @andcarminati

llvm/lib/Target/AIE/aie2p/AIE2PInstrPatterns.td
```diff
@@ -983,16 +990,19 @@ defset list<ExtBcstPair> ExtBcst64Pairs={
 
 foreach  Pair = ExtBcst8Pairs in {
   def : ExtBcstScalarPat<Pair.Inst, v64i8, i32, Pair.IdxTy>;
+  def : ExtZExtBcstScalarPat<Pair.Inst, v64i8, i32, Pair.IdxTy>;
```
andcarminati: Should be the other pattern called `ExtSExtBcstScalarPat`  to be symmetric?

andcarminati: Maybe you could unify with something like:

```
// Extract scalar instead of a subvector
class OpBcstScalarPat<SDPatternOperator OpNode, AIE2PInst Inst, ValueType BcstDstTy, 
                      ValueType ScalarDstTy, DAGOperand IdxTy> :
  Pat<(BcstDstTy (bcst_vector_node
                      (ScalarDstTy(OpNode(BcstDstTy VEC512:$s0), 
                                                      (i32 IdxTy:$idx))))), 
(Inst VEC512:$s0, IdxTy:$idx)>;
```
An then use like this:

```
def : OpBcstScalarPat<vextract_sext, Pair.Inst, v64i8, i32, Pair.IdxTy>;
def : OpBcstScalarPat<vextract_zext, Pair.Inst, v64i8, i32, Pair.IdxTy>;
```

I am not sure if this will work, but it shows the idea.

llvm/lib/Target/AIE/AIECombine.td
```diff
@@ -392,7 +398,9 @@ def AIE2PostLegalizerCustomCombiner
 }
 
 def AIE2PPostLegalizerCustomCombiner
-    : GICombiner<"AIE2PPostLegalizerCustomCombinerImpl", [  combine_global_load_store_increment,
+    : GICombiner<"AIE2PPostLegalizerCustomCombinerImpl", [
```
andcarminati, line 401: CHECK: we moved this from pre to post.

andcarminati: Could we have an MIR test for this combiner?

llvm/lib/Target/AIE/AIELegalizerHelper.cpp
```diff
@@ -1394,12 +1394,6 @@ bool AIELegalizerHelper::legalizeG_FMUL(LegalizerHelper &Helper,
 
   SrcLHS = MIRBuilder.buildAnyExt(S32, SrcLHS).getReg(0);
   SrcRHS = MIRBuilder.buildAnyExt(S32, SrcRHS).getReg(0);
-  SrcLHS = MIRBuilder
```
andcarminati, line 1397: Nice!

andcarminati: Nice work @F-Stuckmann , just small suggestions.

llvm/lib/Target/AIE/AIECombine.td
F-Stuckmann: yes we moved this from prelegalizer to postlegalizer

llvm/lib/Target/AIE/aie2p/AIE2PInstrPatterns.td
F-Stuckmann: nice idea.

F-Stuckmann:
the first force push contains the functional changes, while the second is a rebase on aie-public

andcarminati (approved):
LGTM.

PR #433, opened 2025-04-09
[Draft][wip] Support composite vector register class for register allocation.

niwinanto:
VEC/ACC/FIFO registers can be copied to each other and which can be used as an alternative to costly stack spill. However, the problem of having a composite register class is, sub-register indices corresponds to vector and accumulator does not cover the corresponding sized registers and leads to undefined uses if we use same sub-register indices for both.

The idea is to mimic the register composition hierarchy of vector for accumulator (only for smaller types, because 2048 does not matter) creating dummy 256-bit accumulators and reverting the separate sub-register indices. Now, vector and accumulator register has same lane masks.

PR #678, opened 2025-10-07
[AIE2P] Allow spilling vector/accumulator registers to registers instead of memory

konstantinschwarz:
Stripped down version of #433
FYI @andcarminati @F-Stuckmann


andcarminati:
> Stripped down version of #433 FYI @andcarminati @F-Stuckmann

Thank you @konstantinschwarz , we will move it forward.

F-Stuckmann:
There are still some unit-tests and some fine tuning missing.

andcarminati:
Could you please add some rationale about your change? I am a bit afraid about future upstream conflicts.

PR #682, opened 2025-10-17
[AIEX] Add a latency-aware heuristic component to AIEWAWRegRewriter

andcarminati:
This PR brings optimality in terms of pipelined loops for several important kernels.


llvm/lib/Target/AIE/AIEWawRegRewriter.cpp
```diff
@@ -609,6 +631,48 @@ AIEWawRegRewriter::getLastVRegDef(const MachineBasicBlock &MBB) const {
...
+    for (unsigned I = 0, E = MI.getNumOperands(); I != E; ++I) {
+      const MachineOperand &MO = MI.getOperand(I);
+      if (!MO.isReg() || !MO.isDef())
+        continue;
+
+      const Register Reg = MO.getReg();
+      if (!Reg.isVirtual())
+        continue;
+
+      auto OperandCycle = ItinData->getOperandCycle(SchedClass, I);
```
andcarminati: Move next to use?

llvm/lib/Target/AIE/AIEWawRegRewriter.cpp
```diff
@@ -609,6 +631,48 @@ AIEWawRegRewriter::getLastVRegDef(const MachineBasicBlock &MBB) const {
...
+        if (OperandCycle.value() >= MinRegisterLatency)
+          IsHighLat = true;
+      } else if (MI.mayLoad()) {
+        // If we have an instruction without OperandCycles, it is most probably
+        // a pseudo instruction (no itinerary). In this case, if it is a _split
+        // load, consider it as high latency.
+        IsHighLat = true;
+      }
+
+      if (IsHighLat) {
```
andcarminati: Early return?

andcarminati:
With the new cutoff commit:
Core_Insn_Count

Improvement: 1.11% avg.

llvm/lib/Target/AIE/AIEWawRegRewriter.cpp
```diff
@@ -609,6 +631,59 @@ AIEWawRegRewriter::getLastVRegDef(const MachineBasicBlock &MBB) const {
...
+      auto OperandCycle = ItinData->getOperandCycle(SchedClass, I);
+      const MCRegister AssignedPhysReg = VRM->getPhys(Reg);
+      bool IsHighLat = false;
+      if (OperandCycle.has_value()) {
+        if (OperandCycle.value() >= MinRegisterLatency)
+          IsHighLat = true;
+      } else if (MI.mayLoad()) {
+        // If we have an instruction without OperandCycles, it is most probably
+        // a pseudo instruction (no itinerary). In this case, if it is a _split
+        // load, consider it as high latency.
```
krishnamtibrewala, line 662: check : we should not have multi-slot pseudo here, they have a itinerary with it.

krishnamtibrewala:
> With the new cutoff commit: Core_Insn_Count
> 
> Improvement: 1.11% avg.

Curious to know how was the cut-off decided, and did it lead to reduction in perf, on benchmark where we were getting better performance without the cut-off logic

llvm/lib/Target/AIE/AIEWawRegRewriter.cpp
andcarminati: The multislot pseudos are ok here. they use one itinerary as default. The problem that I was facing here was related to the _split variants that are used just during RA, and they don't have itinerary. They also have a different number of operands.

andcarminati:
> > With the new cutoff commit: Core_Insn_Count
> > Improvement: 1.11% avg.
> 
> Curious to know how was the cut-off decided, and did it lead to reduction in perf, on benchmark where we were getting better performance without the cut-off logic

The original goal of this PR was Conv2D_DW_bf16, but I saw that it also affected Gemm. In this way, I decided to keep those kernels on the heuristic radar. This whole work was based on assembly code analysis (including the cutoff heuristic).

llvm/lib/Target/AIE/AIEWawRegRewriter.cpp
```diff
@@ -361,7 +376,14 @@ bool AIEWawRegRewriter::renameMBBPhysRegs(const MachineBasicBlock *MBB) {
       if (isWorthRenaming(Reg, VRegWithCopies)) {
         assert(VRM->hasPhys(Reg));
         MCRegister AssignedPhysReg = VRM->getPhys(Reg);
-        Candidates.emplace_back(&MO, AssignedPhysReg);
+
+        // Only consider to front physical registers that also
+        // used by high latency operands.
+        if (HighLatencyRegs.count(AssignedPhysReg))
```
andcarminati: If you see here, we reverse the MO ordering by pushing front. This is not by mistake, this helps a lot!

llvm/lib/Target/AIE/AIEWawRegRewriter.cpp
```diff
@@ -609,6 +631,59 @@ AIEWawRegRewriter::getLastVRegDef(const MachineBasicBlock &MBB) const {
...
+
+      if (IsHighLat) {
+        for (MCRegAliasIterator AI(AssignedPhysReg, TRI, true); AI.isValid();
+             ++AI)
+          HighLatRegisters.insert(*AI);
+      }
+    }
+  }
+
+  // This metric gives us an idea about the "demand" for high latency registers
```
andcarminati, line 676: Ratio regs/instructions.

llvm/lib/Target/AIE/AIEWawRegRewriter.cpp
```diff
@@ -361,7 +376,14 @@ bool AIEWawRegRewriter::renameMBBPhysRegs(const MachineBasicBlock *MBB) {
       if (isWorthRenaming(Reg, VRegWithCopies)) {
         assert(VRM->hasPhys(Reg));
         MCRegister AssignedPhysReg = VRM->getPhys(Reg);
-        Candidates.emplace_back(&MO, AssignedPhysReg);
+
+        // Only consider to front physical registers that also
```
martien-de-jong: nit: to put in front registers that are also used

llvm/lib/Target/AIE/AIEWawRegRewriter.cpp
martien-de-jong: perhaps even `return <closed expression>`

llvm/lib/Target/AIE/AIEWawRegRewriter.cpp
```diff
@@ -178,6 +188,9 @@ class AIEWawRegRewriter : public MachineFunctionPass {
   /// instruction.
   IndexedMap<const MachineInstr *, VirtReg2IndexFunctor>
   getLastVRegDef(const MachineBasicBlock &MBB) const;
+
+  std::set<MCRegister>
+  getHighOutputLatencyRegs(const MachineBasicBlock *MBB) const;
```
F-Stuckmann, line 196: nit: add a comment. 
Is this the total live range of a physical register or on a single Live Interval?

llvm/lib/Target/AIE/AIEWawRegRewriter.cpp
F-Stuckmann: add this as a source code comment

F-Stuckmann: CHECK: We push them to the front so that they get replaced first?

llvm/lib/Target/AIE/AIEWawRegRewriter.cpp
```diff
@@ -609,6 +631,48 @@ AIEWawRegRewriter::getLastVRegDef(const MachineBasicBlock &MBB) const {
...
+      const Register Reg = MO.getReg();
+      if (!Reg.isVirtual())
+        continue;
+
+      auto OperandCycle = ItinData->getOperandCycle(SchedClass, I);
+      const MCRegister AssignedPhysReg = VRM->getPhys(Reg);
+      bool IsHighLat = false;
+      if (OperandCycle.has_value()) {
+        if (OperandCycle.value() >= MinRegisterLatency)
+          IsHighLat = true;
```
F-Stuckmann: nit: add brackets here to differentiate between the `!OperandCycle.has_value()` case

llvm/lib/Target/AIE/AIEWawRegRewriter.cpp
```diff
@@ -609,6 +631,48 @@ AIEWawRegRewriter::getLastVRegDef(const MachineBasicBlock &MBB) const {
...
+      } else if (MI.mayLoad()) {
```
F-Stuckmann: nit: don't we always have IsHighLat==true when we encounter a load instruction? 
I think we could simplify the conditions to : 

```
auto SetIsHighLat = [&ItinData, &MinRegisterLatency](const MachineInstr &) {
  if (MI.mayLoad())
    return true;

  auto OperandCycle = ItinData->getOperandCycle(SchedClass, I);
  if (!OperandCycle)
    return false;

  return OperandCycle.value() >= MinRegisterLatency;
};
```

llvm/lib/Target/AIE/AIEWawRegRewriter.cpp
```diff
@@ -639,8 +639,10 @@ std::set<MCRegister> AIEWawRegRewriter::getHighOutputLatencyRegs(
 
   auto *ItinData = MF->getSubtarget().getInstrItineraryData();
   std::set<MCRegister> HighLatRegisters;
+  unsigned NInstrs = 0;
```
F-Stuckmann: nit: why don't we querry `MBB.size()` ?

llvm/lib/Target/AIE/AIEWawRegRewriter.cpp
```diff
@@ -670,6 +672,15 @@ std::set<MCRegister> AIEWawRegRewriter::getHighOutputLatencyRegs(
       }
     }
   }
+
+  // This metric gives us an idea about the "demand" for high latency registers
+  // in a specific loop. If the demand is low, clear HighLatRegisters to skip
+  // latency-aware heuristic.
+  // TODO: this should be replaced by more stable metrics related to SWP.
+  if (!HighLatRegisters.empty() &&
+      (((HighLatRegisters.size() * 100 / NInstrs)) < 250 /*calibrated value*/))
```
F-Stuckmann: nit: for readability add a nice name for `(HighLatRegisters.size() * 100 / NInstrs)`. Something like: `HighLatencyRegisterInstrRatio`

llvm/lib/Target/AIE/AIEWawRegRewriter.cpp
```diff
@@ -53,6 +53,16 @@ static cl::opt<bool> PreAlloc(
...
     cl::desc("Prime the LRU queue to make the allocations more loop-aware"));
 
+static cl::opt<unsigned>
+    MinRegisterLatency("aie-waw-reg-rewrite-min-lat", cl::Hidden,
+                       cl::desc("Minimum operand latency that should be "
+                                "considered for WAW rewriting"),
+                       cl::init(3));
+
+static cl::opt<bool>
+    LatencyAware("aie-realloc-latencyaware", cl::Hidden, cl::init(false),
```
F-Stuckmann: nit: in the last commit set the option to true, since with the cutoff it is now profitable.

llvm/lib/Target/AIE/AIEWawRegRewriter.cpp
andcarminati: Nice suggestion, I will just trust the load itinerary before, because we insert less registers in the set and this could bias the cutoff.

llvm/lib/Target/AIE/AIEWawRegRewriter.cpp
martien-de-jong: HighLatencyRegisterCountToInstrCountRatioInPercent

andcarminati: I changed the name according @F-Stuckmann and extended the comment to accommodate @martien-de-jong observation.

llvm/lib/Target/AIE/AIEWawRegRewriter.cpp
F-Stuckmann: are we looking here at the physical register mapped to a single virtual register def with a high output latency?

llvm/lib/Target/AIE/AIEWawRegRewriter.cpp
F-Stuckmann: nit: Rewrite the comment to 
`High latency registers should be renamed first, therefore insert them at the front`

F-Stuckmann: Maybe also rewrite for clarity: 

```
const auto InsertPoint = HighLatencyRegs.count(AssignedPhysReg) ? Candidates.begin() : Candidates.end(); 
Candidates.emplace(InsertPoint , &MO, AssignedPhysReg);
```

llvm/lib/Target/AIE/AIEWawRegRewriter.cpp
```diff
@@ -609,6 +634,46 @@ AIEWawRegRewriter::getLastVRegDef(const MachineBasicBlock &MBB) const {
...
+    for (unsigned I = 0, E = MI.getNumOperands(); I != E; ++I) {
+      const MachineOperand &MO = MI.getOperand(I);
+      if (!MO.isReg() || !MO.isDef())
+        continue;
+
+      const Register Reg = MO.getReg();
+      if (!Reg.isVirtual())
+        continue;
+
+      auto IsHighLatInstrOperand = [&]() {
```
F-Stuckmann, line 656: out of curiosity. Why is this lambda not outside of the for loops?

llvm/lib/Target/AIE/AIEWawRegRewriter.cpp
```diff
@@ -609,6 +634,46 @@ AIEWawRegRewriter::getLastVRegDef(const MachineBasicBlock &MBB) const {
...
+        auto OperandCycle = ItinData->getOperandCycle(SchedClass, I);
+        if (OperandCycle)
+          return OperandCycle.value() >= MinRegisterLatency;
+        // If we have an instruction without OperandCycles, it is most probably
+        // a pseudo instruction (no itinerary). In this case, if it is a _split
+        // load, consider it as high latency.
+        return MI.mayLoad();
+      };
+
+      if (IsHighLatInstrOperand()) {
```
F-Stuckmann: nit: reduce the nesting

```
if (!IsHighLatInstrOperand())
     continue;
     
const MCRegister AssignedPhysReg = VRM->getPhys(Reg);
for (MCRegAliasIterator AI(AssignedPhysReg, TRI, true); AI.isValid();
             ++AI)
          HighLatRegisters.insert(*AI);
```

llvm/lib/Target/AIE/AIEWawRegRewriter.cpp
```diff
@@ -671,6 +671,17 @@ std::set<MCRegister> AIEWawRegRewriter::getHighOutputLatencyRegs(
       }
     }
   }
+
+  // This metric gives us an idea about the "demand" for high latency registers
+  // in a specific loop. If the demand is low, clear HighLatRegisters to skip
+  // latency-aware heuristic. Basically, we evaluate high latency register count
+  // to instruction count ratio in percent.
+  // TODO: this should be replaced by more stable metrics related to SWP.
+  const int HighLatencyRegisterInstrRatio =
```
F-Stuckmann, line 681: could we have a unit case for this?

llvm/lib/Target/AIE/AIEWawRegRewriter.cpp
andcarminati: One `MCRegister` can also could be mapped to more than one VReg (defs), provided that at least one is high latency.

llvm/lib/Target/AIE/AIEWawRegRewriter.cpp
andcarminati: In this case, my goal was to have a `lazy evaluation boolean` where we can state the conditions very explicitly. It is inside so we can capture everything instead of using parameters (clean if statement).

llvm/test/CodeGen/AIE/aie2p/end-to-end/add-att-broadcasting.ll
```diff
@@ -31,35 +31,36 @@ define dso_local void @add_attribute_bcast(ptr noalias %ifm2, ptr noalias %ifm1,
...
-; CHECK-NEXT:    vlda.conv.fp32.bf16 cml4, [p0], m1; movx r1, #60; add.nc lc, r0, #-2
-; CHECK-NEXT:    vlda.conv.fp32.bf16 cml1, [p0], m1; movxm le, #.L_LEnd0; vadd.f dm2, dm0, dm4, r1
-; CHECK-NEXT:    vlda.conv.fp32.bf16 cml0, [p1], m0; nopb ; nops ; nopxm ; vadd.f dm0, dm3, dm1, r1
-; CHECK-NEXT:    vlda.conv.fp32.bf16 cml3, [p1], m0; nopb ; nops ; nopxm ; nopv
-; CHECK-NEXT:    nopa ; nopb ; nops ; nopxm ; nopv
+; CHECK-NEXT:    vlda.conv.fp32.bf16 cml1, [p0], m1
+; CHECK-NEXT:    vlda.conv.fp32.bf16 cml4, [p0], m1
+; CHECK-NEXT:    vlda.conv.fp32.bf16 cml2, [p1], m0
+; CHECK-NEXT:    vlda.conv.fp32.bf16 cml0, [p1], m0; movx r1, #60
+; CHECK-NEXT:    vlda.conv.fp32.bf16 cml1, [p0], m1; add.nc lc, r0, #-3; vadd.f dm3, dm0, dm1, r1
```
andcarminati, line 39: We can reach one more stage now.

llvm/lib/Target/AIE/AIEWawRegRewriter.cpp
andcarminati: We have one more test now.

andcarminati:
Complete QoR view:

Core_Insn_Count

Core_StackSize_absolute

Core_PMSize_absolute

(PM increase when we increase the number of SWP stages)

F-Stuckmann:
What happens with stack and PM ?

andcarminati:
> What happens with stack and PM ?

I added to the previous comment.

F-Stuckmann (commented):
LGTM

llvm/lib/Target/AIE/AIEWawRegRewriter.cpp
martien-de-jong: I think we can capture most by value.

martien-de-jong (approved):

llvm/lib/Target/AIE/AIEWawRegRewriter.cpp
andcarminati: We use `ItinData` and `MI`, one could be by value and other one by ref. I think I will keep as is to be less verbose.

PR #684, opened 2025-10-21
[AIEX][NFC] global combiners explicit default init

F-Stuckmann:
Use default initialization in global combiners.


konstantinschwarz (approved):
Thanks!

PR #685, opened 2025-10-21
[AIEX] Extend Staged 2D/3D regalloc to avoid spills

andcarminati:
This work is intended to avoid 2D/3D (when possible) register spills.

The idea and rationale behind this work is in a previous Draft PR: https://github.com/Xilinx/llvm-aie/pull/442.

To review, I recommend to follow this PR commit by commit.

Credits also for the co-author @krishnamtibrewala.


andcarminati:
QoR results:

Core_Insn_Count
Core_StackSize_absolute
Core_PMSize_absolute

krishnamtibrewala:
Thanks you @andcarminati, very much !!

llvm/lib/Target/AIE/aie2p/AIE2PTargetMachine.cpp
```diff
@@ -124,6 +124,8 @@ bool AIE2PPassConfig::addRegAssignAndRewriteOptimized() {
     addPass(createAIESuperRegRewriter());
     addPass(createGreedyRegisterAllocator(onlyAllocate3D2DRegisters));
     addPass(createAIESuperRegRewriter());
+    if (EnableFineGrainedStagedRA)
+      addPass(createAIEUnallocatedSuperRegRewriter());
```
krishnamtibrewala, line 118: Check : I think we should try the following, it should be helpful for `onlyAllocate3D2DRegisters` pass

```
    addPass(createGreedyRegisterAllocator(onlyAllocate3DRegisters));
    if (EnableFineGrainedStagedRA)
      addPass(createAIEUnallocatedSuperRegRewriter());
    addPass(createGreedyRegisterAllocator(onlyAllocate3D2DRegisters));
    if (EnableFineGrainedStagedRA)
      addPass(createAIEUnallocatedSuperRegRewriter());
```

krishnamtibrewala:
Also do you think following commit will help ? 
https://github.com/Xilinx/llvm-aie/pull/442/commits/a681b6ec5cfef563739a68382eb03970d452d487

llvm/lib/Target/AIE/aie2p/AIE2PTargetMachine.cpp
andcarminati: I would like to keeps as simple as possible now to setup a common ground first, and then we can test small changes to see the impact. I think it will better to decide.

andcarminati:
> Also do you think following commit will help ? [a681b6e](https://github.com/Xilinx/llvm-aie/commit/a681b6ec5cfef563739a68382eb03970d452d487)

Maybe yes! As mentioned before, I prefer to keep just the minimal necessary changes. We can test after, on top of this PR.

llvm/lib/Target/AIE/aie2p/AIE2PTargetMachine.cpp
krishnamtibrewala: Sounds good!

llvm/lib/Target/AIE/aie2p/AIE2PTargetMachine.cpp
```diff
@@ -70,17 +72,44 @@ void AIE2PPassConfig::addPreRegBankSelect() {
...
+
 static bool onlyAllocate3DRegisters(const TargetRegisterInfo &TRI,
                                     const MachineRegisterInfo &MRI,
                                     const Register &R) {
-  return AIE2P::eDSRegClass.hasSubClassEq(MRI.getRegClass(R));
+
+  const TargetRegisterClass *RegClass = MRI.getRegClass(R);
+  if (!AIE2P::eDSRegClass.hasSubClassEq(RegClass))
+    return false;
+  return EnableFineGrainedStagedRA ? isRegUsedBy2DOr3DInstruction(MRI, R)
```
F-Stuckmann: CHECK: is the function name still correct, since we also querry 2D registers?

F-Stuckmann: Why do we need need this extra check?

llvm/lib/Target/AIE/AIESuperRegRewriter.cpp
```diff
@@ -232,10 +236,32 @@ static void rewriteFullCopy(MachineInstr &MI, const std::set<int> &CopySubRegs,
...
+
+    if (!LIS.hasInterval(R))
+      continue;
+
+    if (VRM.hasPhys(R)) {
+      const MCRegister PhysReg = VRM.getPhys(R);
+      const LiveInterval &OldLI = LIS.getInterval(R);
+      LRM.unassign(OldLI);
+      LIS.removeInterval(R);
+      const LiveInterval &LI = LIS.getInterval(R);
```
F-Stuckmann: nit: clarify that you create a new Interval here by using: `createAndComputeVirtRegInterval`

andcarminati: I replaced by `createAndComputeVirtRegInterval` to be clear.

llvm/lib/Target/AIE/aie2p/AIE2PTargetMachine.cpp
```diff
@@ -70,17 +72,44 @@ void AIE2PPassConfig::addPreRegBankSelect() {
...
+                                         const Register &R) {
+
+  return std::any_of(
+      MRI.use_nodbg_instructions(R).begin(),
+      MRI.use_nodbg_instructions(R).end(), [&](const MachineInstr &MI) {
+        auto &TII = *static_cast<const AIEBaseInstrInfo *>(
+            MI.getMF()->getSubtarget().getInstrInfo());
+
+        // We should recognize both cases, with and without splitting. A 2D/3D
+        // instruction will always be split os splittable.
```
F-Stuckmann: nit: as

llvm/lib/Target/AIE/AIESuperRegRewriter.cpp
```diff
@@ -232,10 +236,32 @@ static void rewriteFullCopy(MachineInstr &MI, const std::set<int> &CopySubRegs,
...
     LIS.InsertMachineInstrInMaps(*PartCopy);
     LIS.getInterval(PartCopy->getOperand(0).getReg());
+    RegistersToRepair.insert(PartCopy->getOperand(1).getReg());
   }
 
   LIS.RemoveMachineInstrFromMaps(MI);
   MI.eraseFromParent();
+  // As we don't handle all registers now (selective LI filter),
+  // We should make sure that all LiveIntervals are correct.
+  // If we dont't repair, MI will compose the LIs of some registers,
```
F-Stuckmann: nit: don't

F-Stuckmann: What do you mean with compose? the dead MI will block some LI ?

llvm/lib/Target/AIE/AIESuperRegUtils.h
```diff
@@ -0,0 +1,72 @@
...
+namespace llvm::AIESuperRegUtils {
+
+/// Returns the subreg indices that can be used to rewrite \p Reg into smaller
+/// regs. Returns {} if the rewrite isn't possible.
+SmallSet<int, 8> getRewritableSubRegs(Register Reg,
+                                      const MachineRegisterInfo &MRI,
+                                      const AIEBaseRegisterInfo &TRI,
+                                      std::set<Register> &VisitedVRegs);
+
+SmallSet<int, 8> getRewritableSubRegs(Register Reg,
```
F-Stuckmann, line 53: nit: can we have a comment what this function does?

llvm/lib/Target/AIE/AIESuperRegUtils.cpp
```diff
@@ -0,0 +1,265 @@
...
+      if (!VisitedVRegs.count(DstReg) &&
+          getRewritableSubRegs(DstReg, MRI, TRI, VisitedVRegs).empty()) {
+        LLVM_DEBUG(dbgs() << "  Cannot rewrite "
+                          << printReg(DstReg, &TRI, 0, &MRI) << " in "
+                          << *RegOp.getParent());
+        return {};
+      }
+      Register SrcReg = RegOp.getParent()->getOperand(1).getReg();
+      if (!VisitedVRegs.count(SrcReg) &&
+          getRewritableSubRegs(SrcReg, MRI, TRI, VisitedVRegs).empty()) {
```
F-Stuckmann, line 70: Does it ever happen, that SrcReg has no SubRegs, but DstReg has them or vis versa? 
Can they also have different SubRegs?

llvm/lib/Target/AIE/AIESuperRegRewriter.cpp
andcarminati: A dead MI will have an invalid SlotIndex that will appear in some instruction`s LI, what is wrong. The original copy should not appear in any LI.

llvm/lib/Target/AIE/AIESuperRegUtils.cpp
```diff
@@ -0,0 +1,265 @@
...
+    LIS.InsertMachineInstrInMaps(*PartCopy);
+    LIS.getInterval(PartCopy->getOperand(0).getReg());
+    RegistersToRepair.insert(PartCopy->getOperand(1).getReg());
+  }
+
+  LIS.RemoveMachineInstrFromMaps(MI);
+  MI.eraseFromParent();
+  // As we don't handle all registers now (selective LI filter),
+  // We should make sure that all LiveIntervals are correct.
+  // If we dont't repair, MI will compose the LIs of some registers,
```
F-Stuckmann: nit: don't

llvm/lib/Target/AIE/AIESuperRegUtils.cpp
```diff
@@ -0,0 +1,265 @@
...
+  // what is not correct because MI was deleted.
+  updateLRMAndLIS(RegistersToRepair, VRM, LRM, LIS);
```
F-Stuckmann: nit: could we get a better name that does not have the abbreviations in it? 
`updateLiveIntervals`. Also since LIS is the most relevant Datastructure, have it first and then followed by helper structures `LRM` and `VRM`.

llvm/lib/Target/AIE/AIESuperRegUtils.cpp
andcarminati: This part was a refactor, but as we are handling a full copy here, we can expect subregs on both sides.

andcarminati: As a full copy, I mean a 2d to 2d copy or a 3d to 3d copy.

llvm/lib/Target/AIE/aie2p/AIE2PTargetMachine.cpp
andcarminati: Repaired as a fixup in the refactor commit.

llvm/lib/Target/AIE/AIESuperRegUtils.h
andcarminati: Yes, it is a refactor but it is always good to document.

llvm/lib/Target/AIE/AIESuperRegUtils.cpp
```diff
@@ -103,9 +103,9 @@ void rewriteFullCopy(MachineInstr &MI, const std::set<int> &CopySubRegs,
   Register SrcReg = MI.getOperand(1).getReg();
   LaneBitmask LiveSrcLanes = getLiveLanesAt(CopyIndex, SrcReg, LIS);
 
-  LIS.removeVRegDefAt(LIS.getInterval(DstReg), CopyIndex.getRegSlot());
-
+  MachineInstr *FirstMI = nullptr;
   SmallSet<Register, 8> RegistersToRepair;
+  bool IsFirstCopy = true;
```
andcarminati: nit: We can use FirstMI as a replacement for IsFirstCopy.

llvm/lib/Target/AIE/AIEUnallocatedSuperRegRewriter.cpp
```diff
@@ -0,0 +1,172 @@
+//===----- AIEUnallocatedSuperRegRewriter.cpp - Constrain tied sub-registers
```
andcarminati: Fix this.

llvm/test/CodeGen/AIE/aie2p/ra/staged-ra-check-undef.mir
```diff
@@ -0,0 +1,164 @@
...
+# This file is licensed under the Apache License v2.0 with LLVM Exceptions.
+# See https://llvm.org/LICENSE.txt for license information.
+# SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+#
+# (c) Copyright 2025 Advanced Micro Devices, Inc. or its affiliates
+#
+# RUN: llc -O2 -mtriple=aie2p -start-before=greedy \
+# RUN:     -stop-before=aie-unallocated-superreg-rewrite %s -o - | FileCheck %s
+
+# The goal of this test is to check if we properly insert undef flag on the def side
```
andcarminati, line 12: I could extend the verifier to cover partial definitions. It is far away from be trivial, but it could be done.

llvm/lib/Target/AIE/aie2p/AIE2PTargetMachine.cpp
```diff
@@ -60,17 +62,44 @@ void AIE2PPassConfig::addPreRegBankSelect() {
   }
 }
 
+static bool isRegUsedBy2DOr3DInstruction(const MachineRegisterInfo &MRI,
+                                         const Register &R) {
+
+  return std::any_of(
```
martien-de-jong: llvm::any_of on the range?

llvm/lib/Target/AIE/AIEUnallocatedSuperRegRewriter.cpp
```diff
@@ -125,6 +126,62 @@ void rewriteCandidates(RegRewriteInfo &RegistersToRewrite,
...
+                              MachineRegisterInfo &MRI, SlotIndexes &Indexes,
+                              LiveIntervals &LIS, VirtRegMap &VRM,
+                              LiveRegMatrix &LRM) {
+
+  SmallSet<Register, 8> RegistersToRepair;
+  for (auto [VReg, SubRegs] : RegistersToRewrite) {
+
+    for (MachineInstr &MI : MRI.def_instructions(VReg)) {
+
+      if (!MI.isBundledWithPred() || MI.isBundledWithSucc())
```
martien-de-jong, line 146: Can I have a comment explaining this directional asymmetry?

andcarminati:
There is no failures across all benchmarks.

llvm/lib/Target/AIE/AIESuperRegUtils.cpp
```diff
@@ -0,0 +1,321 @@
...
+  for (int SubRegIdx : CopySubRegs) {
+    if ((LiveSrcLanes & TRI.getSubRegIndexLaneMask(SubRegIdx)).none()) {
+      LLVM_DEBUG(dbgs() << "        Skip undefined subreg "
+                        << TRI.getSubRegIndexName(SubRegIdx) << "\n");
+      continue;
+    }
+
+    MachineInstr *PartCopy = BuildMI(*MI.getParent(), MI, MI.getDebugLoc(),
+                                     TII.get(TargetOpcode::COPY))
+                                 .addReg(DstReg, RegState::Define, SubRegIdx)
```
F-Stuckmann: nit: we do not need undefs here because we fully define the subregisters that we copy to?

llvm/lib/Target/AIE/AIESuperRegUtils.cpp
```diff
@@ -0,0 +1,321 @@
...
+                                 .addReg(SrcReg, 0, SubRegIdx)
+                                 .getInstr();
+
+    // Only set undefined on the first partial copy. The first copy doesn't read
+    // other lanes, but subsequent copies do read the previously written lanes.
+    // Setting undefined on all copies breaks live interval tracking and causes
```
F-Stuckmann, line 148: oh yeah i also found that out with subreg spilling XD

llvm/lib/Target/AIE/AIESuperRegUtils.cpp
```diff
@@ -0,0 +1,321 @@
...
+    // because it considers that every sub-lane copy will make the preceding
+    // one dead, what is not true for composite registers.
+    // TODO: investigate why subregister liveness is being ignored by LIS
+    // at this point.
+    RegistersToRepair.insert(PartCopy->getOperand(1).getReg());
+  }
+
+  // Replace the original copy by the first one, so we automatically repair
+  // DstReg's LI.
+  LIS.ReplaceMachineInstrInMaps(MI, *FirstMI);
```
F-Stuckmann: why aren't we inserting all the newly created machineinstructions into LIS?

llvm/lib/Target/AIE/AIESuperRegUtils.cpp
```diff
@@ -0,0 +1,321 @@
...
+
+SmallSet<int, 8> getRewritableSubRegs(Register Reg,
+                                      const MachineRegisterInfo &MRI,
+                                      const AIEBaseRegisterInfo &TRI) {
+  std::set<Register> VisitedVRegs;
+  return getRewritableSubRegs(Reg, MRI, TRI, VisitedVRegs);
+}
+
+/// Rewrite a full copy into multiple copies using the subregs in \p CopySubRegs
+void rewriteFullCopy(MachineInstr &MI, const std::set<int> &CopySubRegs,
```
F-Stuckmann: MI is a copy, right? Could you rename MI to CopyMI, so it becomes clearer?

llvm/lib/Target/AIE/AIESuperRegUtils.cpp
```diff
@@ -0,0 +1,321 @@
...
+            : TRI.getLargestLegalSuperClass(
+                  TRI.getSubRegisterClass(SuperRC, SubReg), MF);
+    SubRegToVReg[SubReg] = MRI.createVirtualRegister(SubRC);
+  }
+
+  // Rewrite full copies into multiple copies using subregs
+  for (MachineInstr &MI : make_early_inc_range(MRI.reg_instructions(Reg))) {
+    if (MI.isFullCopy())
+      AIESuperRegUtils::rewriteFullCopy(
+          MI, TRI.getSubRegSplit(MRI.getRegClass(Reg)->getID()), LIS, *TII, TRI,
```
F-Stuckmann: you are also passing TRI, I think you don't need to pass CopySubRegs here

You can get MRI like this: 
MachineRegisterInfo &MRI = MI.getMF()->getRegInfo();

llvm/lib/Target/AIE/AIESuperRegUtils.cpp
```diff
@@ -0,0 +1,321 @@
...
+
+  // Replace the original copy by the first one, so we automatically repair
+  // DstReg's LI.
+  LIS.ReplaceMachineInstrInMaps(MI, *FirstMI);
+  MI.eraseFromParent();
+  // As we don't handle all registers now (selective LI filter),
+  // We should make sure that all LiveIntervals are correct.
+  // If we don't repair, MI will compose the LIs of some registers,
+  // what is not correct because MI was deleted.
+  repairLiveIntervals(RegistersToRepair, VRM, LRM, LIS);
```
F-Stuckmann, line 176: Can't you use
LIS.InsertMachineInstrRangeInMaps(AllSubRegCopies)?

llvm/lib/Target/AIE/AIESuperRegUtils.cpp
```diff
@@ -0,0 +1,321 @@
...
+}
+
+/// Return a mask of all the lanes that are live at \p Index
+LaneBitmask getLiveLanesAt(SlotIndex Index, Register Reg,
```
F-Stuckmann, line 180: I love this function! Thank you very much

llvm/lib/Target/AIE/AIESuperRegUtils.cpp
```diff
@@ -0,0 +1,321 @@
...
+}
+
+void rewriteSuperReg(Register Reg, std::optional<Register> AssignedPhysReg,
+                     SmallSet<int, 8> &SubRegs, MachineRegisterInfo &MRI,
+                     const AIEBaseRegisterInfo &TRI, VirtRegMap &VRM,
+                     LiveRegMatrix &LRM, LiveIntervals &LIS,
+                     SlotIndexes &Indexes, LiveDebugVariables &DebugVars) {
+  LLVM_DEBUG(dbgs() << "Rewriting " << printReg(Reg, &TRI, 0, &MRI) << '\n');
+  MachineFunction &MF = VRM.getMachineFunction();
+  auto *TII =
```
F-Stuckmann, line 201: nit: const

llvm/lib/Target/AIE/AIEUnallocatedSuperRegRewriter.cpp
```diff
@@ -0,0 +1,235 @@
...
+  LLVM_DEBUG(dbgs() << "Found " << RegistersToRewrite.size()
+                    << " candidate register(s) for rewriting\n");
+
+  return RegistersToRewrite;
+}
+
+/// Split candidate registers into independent virtual registers for each
+/// subregister. Each composite register is rewritten using its subregister
+/// indices, with live intervals and debug information updated accordingly.
+void rewriteCandidates(RegRewriteInfo &RegistersToRewrite,
```
F-Stuckmann, line 114: Are these the candidates to rewrite? 
nit: rename for clarity

llvm/lib/Target/AIE/AIESuperRegUtils.cpp
```diff
@@ -0,0 +1,321 @@
...
+      static_cast<const AIEBaseInstrInfo *>(MF.getSubtarget().getInstrInfo());
+
+  // Collect all the subreg indices to rewrite as independent vregs.
+  SmallMapVector<int, Register, 8> SubRegToVReg;
+  const TargetRegisterClass *SuperRC = MRI.getRegClass(Reg);
+  assert(!SubRegs.empty());
+  for (int SubReg : SubRegs) {
+    const TargetRegisterClass *SubRC =
+        AssignedPhysReg.has_value()
```
F-Stuckmann, line 210: why do we care if the AssignedPhysReg is passed? 
If we go for largestLegalSuperclass, we aren't really going for subreg rewriting, but for a regular copy.

F-Stuckmann: Shouldn't the check be if a subregisterClass exists and if we fail to retrieve it we should go for largestlegalSuperclass? 
In the failure to get a subregClass should we still iterate through all the subregs? 
Don't group multiple together if we use a superclass?

llvm/lib/Target/AIE/AIESuperRegUtils.cpp
```diff
@@ -0,0 +1,321 @@
...
+            ? TRI.getSubRegisterClass(SuperRC, SubReg)
+            : TRI.getLargestLegalSuperClass(
+                  TRI.getSubRegisterClass(SuperRC, SubReg), MF);
+    SubRegToVReg[SubReg] = MRI.createVirtualRegister(SubRC);
+  }
+
+  // Rewrite full copies into multiple copies using subregs
+  for (MachineInstr &MI : make_early_inc_range(MRI.reg_instructions(Reg))) {
+    if (MI.isFullCopy())
```
F-Stuckmann, line 219: Does `reg_instructions(Reg)` also return subreg uses? 
If not at this point we can assume that Reg is a superregister and exit early if we aren't a superregister.

llvm/lib/Target/AIE/AIESuperRegUtils.cpp
```diff
@@ -0,0 +1,321 @@
...
+      AIESuperRegUtils::rewriteFullCopy(
+          MI, TRI.getSubRegSplit(MRI.getRegClass(Reg)->getID()), LIS, *TII, TRI,
+          VRM, LRM);
+  }
+
+  LLVM_DEBUG(dbgs() << "  Splitting range " << LIS.getInterval(Reg) << "\n");
+  for (MachineOperand &RegOp : make_early_inc_range(MRI.reg_operands(Reg))) {
+    LLVM_DEBUG(dbgs() << "  Changing " << *RegOp.getParent());
+    int SubReg = RegOp.getSubReg();
```
F-Stuckmann, line 226: How do we know, we don't have further subregs below us?

llvm/lib/Target/AIE/AIESuperRegUtils.cpp
```diff
@@ -0,0 +1,321 @@
...
+/// regs. Returns {} if the rewrite isn't possible.
+SmallSet<int, 8> getRewritableSubRegs(Register Reg,
+                                      const MachineRegisterInfo &MRI,
+                                      const AIEBaseRegisterInfo &TRI,
+                                      std::set<Register> &VisitedVRegs) {
+  if (Reg.isPhysical()) {
+    // TODO: One could use collectSubRegs() in AIEBaseInstrInfo.cpp
+    // But given that MOD registers are not part of the ABI, they should
+    // not appear as physical registers before RA.
+    LLVM_DEBUG(dbgs() << "  Cannot rewrite physreg " << printReg(Reg, &TRI)
```
F-Stuckmann, line 38: nit message should make it clear, that we cannot collect subregs to rewrite, since it already physical

llvm/lib/Target/AIE/AIESuperRegUtils.cpp
```diff
@@ -0,0 +1,321 @@
...
+    // Register does not have multiple subregs to be rewritten into.
+    LLVM_DEBUG(dbgs() << "  Cannot rewrite " << printReg(Reg, &TRI, 0, &MRI)
+                      << ": no sub-reg split\n");
+    return {};
+  }
+
+  VisitedVRegs.insert(Reg);
+  SmallSet<int, 8> UsedSubRegs;
+  for (MachineOperand &RegOp : MRI.reg_operands(Reg)) {
+    int SubReg = RegOp.getSubReg();
```
F-Stuckmann: nit: const

llvm/lib/Target/AIE/AIESuperRegUtils.cpp
```diff
@@ -0,0 +1,321 @@
...
+    if (SubReg && SubRegSplit.count(SubReg)) {
```
F-Stuckmann, line 55: if we are a subreg, shouldn't we automatically be in the SubRegSplit? 
I think we can assert it instead of checking it in the condition.

llvm/lib/Target/AIE/AIESuperRegUtils.cpp
```diff
@@ -0,0 +1,321 @@
...
+  if (SubRegSplit.size() <= 1) {
+    // Register does not have multiple subregs to be rewritten into.
+    LLVM_DEBUG(dbgs() << "  Cannot rewrite " << printReg(Reg, &TRI, 0, &MRI)
+                      << ": no sub-reg split\n");
+    return {};
+  }
+
+  VisitedVRegs.insert(Reg);
+  SmallSet<int, 8> UsedSubRegs;
+  for (MachineOperand &RegOp : MRI.reg_operands(Reg)) {
```
F-Stuckmann, line 53: Maybe add a comment that we are walking all defs and uses to find subreg uses and full copies.

llvm/lib/Target/AIE/AIESuperRegUtils.cpp
```diff
@@ -0,0 +1,321 @@
...
+    int SubReg = RegOp.getSubReg();
+    if (SubReg && SubRegSplit.count(SubReg)) {
+      UsedSubRegs.insert(SubReg);
+    } else if (RegOp.getParent()->isFullCopy()) {
```
F-Stuckmann, line 57: Why are we performing this check on uses? 
Isn't this a check we can perform once on the DefI?

llvm/lib/Target/AIE/AIESuperRegUtils.cpp
```diff
@@ -0,0 +1,321 @@
...
+      if (!VisitedVRegs.count(SrcReg) &&
+          getRewritableSubRegs(SrcReg, MRI, TRI, VisitedVRegs).empty()) {
+        LLVM_DEBUG(dbgs() << "  Cannot rewrite "
+                          << printReg(SrcReg, &TRI, 0, &MRI) << " in "
+                          << *RegOp.getParent());
+        return {};
+      }
+      UsedSubRegs.insert(SubRegSplit.begin(), SubRegSplit.end());
+    } else {
+      LLVM_DEBUG(dbgs() << "  Cannot rewrite " << RegOp << " in "
```
F-Stuckmann, line 80: Is this a regular use/Def of the full register?

llvm/lib/Target/AIE/AIESuperRegUtils.cpp
andcarminati: Because, later on, each new lane copy will make dead the previous one if we try to fix the LI. In this case, we say that the LR will start on the first copy and then it will simply continue though the next ones, until the next last use. I this way we prevent false dead flags everywhere.

llvm/lib/Target/AIE/AIESuperRegUtils.cpp
andcarminati: Yes, we start the definition on the first copy, and we consider that next copies will read the previous one. In practice, inserting undefs in all copies will not hurt in general, but as is, we have a more accurate model for incremental definitions.

llvm/lib/Target/AIE/AIESuperRegRewriter.cpp
```diff
@@ -232,10 +236,32 @@ static void rewriteFullCopy(MachineInstr &MI, const std::set<int> &CopySubRegs,
     LLVM_DEBUG(dbgs() << "        to " << *PartCopy);
     LIS.InsertMachineInstrInMaps(*PartCopy);
     LIS.getInterval(PartCopy->getOperand(0).getReg());
+    RegistersToRepair.insert(PartCopy->getOperand(1).getReg());
```
andcarminati: Comment about repair (only r)

llvm/lib/Target/AIE/AIESuperRegUtils.cpp
```diff
@@ -118,13 +136,31 @@ void rewriteFullCopy(MachineInstr &MI, const std::set<int> &CopySubRegs,
...
+    // because it considers that every sub-lane copy will make the preceding
+    // one dead, what is not true for composite registers.
+    // TODO: investigate why subregister liveness is being ignored by LIS
+    // at this point.
     RegistersToRepair.insert(PartCopy->getOperand(1).getReg());
   }
 
-  LIS.RemoveMachineInstrFromMaps(MI);
+  // Replace the original copy by the first one, so we automatically repair
+  // DstReg's LI.
```
andcarminati: Why?

llvm/lib/Target/AIE/AIEUnallocatedSuperRegRewriter.cpp
```diff
@@ -0,0 +1,171 @@
...
+/// registers.
+static RegRewriteInfo getRewriteCandidates(MachineRegisterInfo &MRI,
+                                           const AIEBaseRegisterInfo &TRI,
+                                           VirtRegMap &VRM) {
+  RegRewriteInfo RegistersToRewrite;
+  for (unsigned VRegIdx = 0, End = MRI.getNumVirtRegs(); VRegIdx != End;
+       ++VRegIdx) {
+    const Register Reg = Register::index2VirtReg(VRegIdx);
+
+    // Ignore un-used registers
```
andcarminati: Complete the comment.

llvm/lib/Target/AIE/AIEUnallocatedSuperRegRewriter.cpp
```diff
@@ -0,0 +1,171 @@
...
+    if (MRI.reg_nodbg_empty(Reg) || VRM.hasPhys(Reg))
+      continue;
+
+    SmallSet<int, 8> RewritableSubRegs =
```
andcarminati: const

llvm/lib/Target/AIE/AIEUnallocatedSuperRegRewriter.cpp
```diff
@@ -0,0 +1,171 @@
...
+                       SlotIndexes &Indexes, LiveDebugVariables &DebugVars) {
+
+  LLVM_DEBUG(dbgs() << "Rewriting " << RegistersToRewrite.size()
+                    << " candidate register(s)\n");
+
+  for (auto [VReg, SubRegs] : RegistersToRewrite) {
+    LLVM_DEBUG(dbgs() << "  Rewriting " << printReg(VReg, &TRI, 0, &MRI)
+                      << " into " << SubRegs.size() << " subregister(s)\n");
+    AIESuperRegUtils::rewriteSuperReg(
+        VReg, /*std::optional<Register> AssignedPhysReg = */ {}, SubRegs, MRI,
```
andcarminati: Local var for Assign...

llvm/lib/Target/AIE/AIESuperRegUtils.cpp
```diff
@@ -184,21 +184,26 @@ LaneBitmask getLiveLanesAt(SlotIndex Index, Register Reg,
...
   // Collect all the subreg indices to rewrite as independent vregs.
   SmallMapVector<int, Register, 8> SubRegToVReg;
   const TargetRegisterClass *SuperRC = MRI.getRegClass(Reg);
   assert(!SubRegs.empty());
   for (int SubReg : SubRegs) {
-    const TargetRegisterClass *SubRC = TRI.getSubRegisterClass(SuperRC, SubReg);
+    const TargetRegisterClass *SubRC =
+        AssignedPhysReg.has_value()
+            ? TRI.getSubRegisterClass(SuperRC, SubReg)
+            : TRI.getLargestLegalSuperClass(
```
andcarminati, line 212: Why?

llvm/test/CodeGen/AIE/aie2p/ra/staged-ra-spill.mir
```diff
@@ -6,42 +6,68 @@
...
+    ; RA-STAGED-NEXT: ST_D_SPILL killed renamable $d0, %stack.0, implicit $sp :: (store (s128) into %stack.0, align 4)
+    ; RA-STAGED-NEXT: renamable $d0 = LDA_D_SPILL %stack.1, implicit $sp :: (load (s128) from %stack.1, align 4)
+    ; RA-STAGED-NEXT: $p0, dead $dc0 = PADDA_2D_split killed $p0, $m0, $dn0, $dj0, $dc0
+    ; RA-STAGED-NEXT: renamable $d0 = LDA_D_SPILL %stack.0, implicit $sp :: (load (s128) from %stack.0, align 4)
+    ; RA-STAGED-NEXT: PseudoRET implicit $lr, implicit killed renamable $p0, implicit killed renamable $dc0, implicit $d1, implicit $d2, implicit $d3, implicit $d4, implicit $d5, implicit $d6, implicit $d7
+    ;
+    ; RA-STAGEG-FG-LABEL: name: test_spill_2d_last_stage
+    ; RA-STAGEG-FG: liveins: $d1, $d2, $d3, $d4, $d5, $d6, $d7, $p0, $p1
+    ; RA-STAGEG-FG-NEXT: {{  $}}
+    ; RA-STAGEG-FG-NEXT: renamable $dn0 = LDA_dms_lda_idx_imm renamable $p1, 0
```
andcarminati: STAGED

llvm/lib/Target/AIE/AIESuperRegUtils.cpp
```diff
@@ -257,10 +261,13 @@ void rewriteSuperReg(Register Reg, Register AssignedPhysReg,
     LIComponents.push_back(&SubRegLI);
     VRM.grow();
 
-    for (LiveInterval *LI : LIComponents) {
-      LRM.assign(*LI, SubPhysReg);
-      VRM.setRequiredPhys(LI->reg(), SubPhysReg);
-      LLVM_DEBUG(dbgs() << "  Assigned " << printReg(LI->reg()) << "\n");
+    if (AssignedPhysReg.has_value()) {
```
andcarminati: early return.

llvm/lib/Target/AIE/AIESuperRegUtils.cpp
andcarminati: I will keep as is, this code was just refactored.

llvm/lib/Target/AIE/AIESuperRegUtils.cpp
andcarminati: This was just a refactor.

llvm/lib/Target/AIE/AIESuperRegRewriter.cpp
andcarminati: The message is in a propriated commit.

llvm/lib/Target/AIE/AIESuperRegUtils.cpp
andcarminati: This was just a refactor.

llvm/lib/Target/AIE/AIESuperRegUtils.cpp
F-Stuckmann: Adding an assert crashes here because we cannot properly handle low and high parts of the 3d Registers, so a whole d0 or d4 register in the case of 3d_0.
They do not get returned from the `TRI.getSubRegSplit()`
Could you add a comment ? 
Also do we know, why we did not add it to `TRI.getSubRegSplit()`?

F-Stuckmann: These are not full copies but partial copies, that we currently don't handle

F-Stuckmann: I think we are leaving something on the table here. not sure if it is needed for this PR though

llvm/lib/Target/AIE/AIESuperRegUtils.cpp
F-Stuckmann: Is there a reason why we don't handle copies uniform in this method? 
we also encounter copies in this case here `if (SubReg && SubRegSplit.count(SubReg))`

llvm/lib/Target/AIE/AIESuperRegUtils.cpp
```diff
@@ -0,0 +1,267 @@
...
+
+SmallSet<int, 8> getRewritableSubRegs(Register Reg,
+                                      const MachineRegisterInfo &MRI,
+                                      const AIEBaseRegisterInfo &TRI) {
+  std::set<Register> VisitedVRegs;
+  return getRewritableSubRegs(Reg, MRI, TRI, VisitedVRegs);
+}
+
+/// Rewrite a full copy into multiple copies using the subregs in \p CopySubRegs
+void rewriteFullCopy(MachineInstr &CopyMI, LiveIntervals &LIS,
```
F-Stuckmann, line 97: note: in the current state we will never rewrite subreg copies, e.g. copies of d0 or d4 in the case of a 3d_0 register.
Do you know why they are not relevant for the Superregrewriter?

llvm/lib/Target/AIE/AIESuperRegUtils.cpp
F-Stuckmann: yeah it does.

llvm/lib/Target/AIE/AIESuperRegUtils.h
F-Stuckmann: Could we swap the documentation here? 
I think the high level function should get the documentation, the actual implementation function (the one above this) should get the slimmed documentation.

llvm/lib/Target/AIE/AIESuperRegUtils.cpp
```diff
@@ -0,0 +1,325 @@
...
+    const TargetRegisterClass *SubRC =
+        AssignedPhysReg.has_value()
+            ? TRI.getSubRegisterClass(SuperRC, SubReg)
+            : TRI.getLargestLegalSuperClass(
+                  TRI.getSubRegisterClass(SuperRC, SubReg), MF);
+    SubRegToVReg[SubReg] = MRI.createVirtualRegister(SubRC);
+  }
+
+  // Rewrite full copies into multiple copies using subregs
+  for (MachineInstr &MI : make_early_inc_range(MRI.reg_instructions(Reg))) {
```
F-Stuckmann, line 218: Do we rewrite Full copies again when we AssignedPhysReg is not set?

llvm/lib/Target/AIE/AIESuperRegUtils.cpp
```diff
@@ -0,0 +1,325 @@
...
+    if (MI.isFullCopy())
+      rewriteFullCopy(MI, LIS, *TII, TRI, VRM, LRM);
+  }
+
+  LLVM_DEBUG(dbgs() << "  Splitting range " << LIS.getInterval(Reg) << "\n");
```
F-Stuckmann, line 223: Do we split ranges again when we AssignedPhysReg is not set?

llvm/lib/Target/AIE/AIEUnallocatedSuperRegRewriter.cpp
```diff
@@ -0,0 +1,171 @@
...
+/// registers.
+static RegRewriteInfo getRewriteCandidates(MachineRegisterInfo &MRI,
+                                           const AIEBaseRegisterInfo &TRI,
+                                           VirtRegMap &VRM) {
+  RegRewriteInfo RegistersToRewrite;
+  for (unsigned VRegIdx = 0, End = MRI.getNumVirtRegs(); VRegIdx != End;
+       ++VRegIdx) {
+    const Register Reg = Register::index2VirtReg(VRegIdx);
+
+    // Ignore un-used od already allocated registers.
```
F-Stuckmann: nit: or

llvm/lib/Target/AIE/AIEUnallocatedSuperRegRewriter.cpp
```diff
@@ -0,0 +1,171 @@
+//===-- AIEUnallocatedSuperRegRewriter.cpp - Constrain tied sub-registers -===//
+//
+// This file is licensed under the Apache License v2.0 with LLVM Exceptions.
+// See https://llvm.org/LICENSE.txt for license information.
+// SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+//
+// (c) Copyright 2025 Advanced Micro Devices, Inc. or its affiliates
+//
+//===----------------------------------------------------------------------===//
```
F-Stuckmann, line 9: What registers aren't assigned after the previous 2d/3d reg allocs? 
Are we changing here the copies and moves not the padds? 
Could you Comment why this pass is necessary and the previous passes do not pick this up?

llvm/lib/Target/AIE/AIESuperRegUtils.cpp
andcarminati: I guess this will never occur because it will be prevented by. `getRewritableSubRegs`.

llvm/lib/Target/AIE/AIESuperRegUtils.cpp
andcarminati: I see that they may occur in case we have a `VReg0_unallocated = COPY VReg1_unallocated`. Considering Conv2D_bft16, we start with  5 3D virtual registers and end up creating 32 ones, so I this may be possible and I would like to keep this this possibility.

llvm/lib/Target/AIE/AIESuperRegUtils.cpp
andcarminati: Yes, in this part of the function we change the proper registers by others with `SubRegisterClass` (allocated) and `LargestLegalSuperClass` (unallocated). Also, we are not doing this again, each pass covesr one subset: allocated or unallocated.

llvm/lib/Target/AIE/AIEUnallocatedSuperRegRewriter.cpp
andcarminati: I will complement this part.

llvm/lib/Target/AIE/AIESuperRegUtils.cpp
andcarminati: > Also do we know, why we did not add it to TRI.getSubRegSplit()?
We always rewrite into the smallest lanes as possible.

Here we are concerned with the things that we can rewrite after. Any update here should be aligned with the rewrite after, and it is not clear the benefits afterwards. I really don't want to complicate this PR even more. Any improvement could be done with an established baseline as follow up.

llvm/lib/Target/AIE/AIESuperRegUtils.cpp
andcarminati: Because we should be able to full expand the copy after. At this moment we don't handle cases like `%7.sub_hi_dim:eds = COPY %6:ed`.

llvm/lib/Target/AIE/AIESuperRegUtils.cpp
andcarminati: This will catch things like `%7.sub_hi_dim:eds = COPY %6:ed`, not handled after.

llvm/lib/Target/AIE/AIESuperRegUtils.cpp
andcarminati: I included a TODO comment related to this investigation.

F-Stuckmann:
LGTM @martien-de-jong what do you say

llvm/lib/Target/AIE/AIESuperRegUtils.cpp
```diff
@@ -0,0 +1,269 @@
...
+  for (MachineOperand &RegOp : make_early_inc_range(MRI.reg_operands(Reg))) {
+    LLVM_DEBUG(dbgs() << "  Changing " << *RegOp.getParent());
+    int SubReg = RegOp.getSubReg();
+    assert(SubReg);
+    RegOp.setReg(SubRegToVReg[SubReg]);
+    RegOp.setSubReg(0);
```
F-Stuckmann, line 229: Note: 3D and 2D reg alloc will have already allocated the the 3D and 2D users except for the copies.
We spill them in line 178, so that we only work on subregs. 
The users we now encounter are the subreg copies from the `rewriteFullCopy`

andcarminati: Please note that `rewriteFullCopy` is to prevent failures during the rewrite. We can rewrite also operands in load and move (also immediate) instructions. See instructions covered by `ItinRegClassPair`.

F-Stuckmann (approved):

PR #686, opened 2025-10-21
[AIE2P] Fix the 128-bit vector register calling convention

konstantinschwarz:
128-bit vector types should be passed in the LSB of the 256-bit 'w' registers, similar to AIE2. However, we were promoting the 128-bit vector type to a 256-bit vector type at the call site and truncating back in the callee.

### DIFF
--- a/llvm/lib/Target/AIE/aie2p/AIE2PISelLowering.cpp
+++ b/llvm/lib/Target/AIE/aie2p/AIE2PISelLowering.cpp
@@ -15,6 +15,7 @@
 #include "AIE2PISelLowering.h"
 #include "AIEBaseSubtarget.h"
 #include "MCTargetDesc/aie2p/AIE2PMCTargetDesc.h"
+#include "llvm/CodeGen/TargetLowering.h"
 #include "llvm/IR/IntrinsicsAIE2P.h"
 
 using namespace llvm;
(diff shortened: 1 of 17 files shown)


llvm/lib/Target/AIE/aie2p/AIE2PISelLowering.cpp
```diff
@@ -128,3 +129,50 @@ bool AIE2PTargetLowering::getTgtMemIntrinsic(IntrinsicInfo &Info,
...
+MVT AIE2PTargetLowering::getRegisterTypeForCallingConv(LLVMContext &Context,
+                                                       CallingConv::ID CC,
+                                                       EVT VT) const {
+  // 128-bits registers aren't considered legal because AIE2P only has vector
+  // ops for 256+ bits vectors. However, for ABI compatibility reasons, we still
+  // want to keep those vectors as is, because they are passed differently
+  // compared to 256-bits vectors.
+  if (VT.isSimple() && VT.is128BitVector())
+    return VT.getSimpleVT();
+  // 128-bit integers aren't considered legal to prevent dafault legalization of
```
andcarminati: I am curious about the use case of this second part. Do we have some related test?

andcarminati:
Nice, now I see that we were crashing before in the legalizer (`G_ANYEXT`).

llvm/lib/Target/AIE/aie2p/AIE2PISelLowering.cpp
konstantinschwarz: We did not handle i128 types in our `AIE2PCallingConv.td` yet, and the tests were commented out.
I enabled the tests and added the missing ones, now we are correctly passing i128 in the `q` registers

llvm/lib/Target/AIE/aie2p/AIE2PISelLowering.cpp
```diff
@@ -128,3 +129,50 @@ bool AIE2PTargetLowering::getTgtMemIntrinsic(IntrinsicInfo &Info,
...
+  if (VT.isScalarInteger() && VT.getScalarSizeInBits() == 128)
+    return VT.getSimpleVT();
+
+  return AIEBaseTargetLowering::getRegisterTypeForCallingConv(Context, CC, VT);
+}
+
+unsigned AIE2PTargetLowering::getNumRegistersForCallingConv(
+    LLVMContext &Context, CallingConv::ID CC, EVT VT) const {
+  // See getRegisterTypeForCallingConv(): i128 is passed in a single register
+  if (VT.isScalarInteger() && VT.getScalarSizeInBits() == 128)
```
niwinanto: check: Why we have to really handle 128bit scalar case here. Why would we get more number of registers for this case?

niwinanto: Hmm, may be including `i128` should be enough
```
 class AIE2PVector128RegisterClass<dag reglist> :
      AIE2PRegisterClass<128, 128, [v16i8, v8i16, v8bf16, v4i32, v4f32], reglist> {
```

konstantinschwarz: If I remember correctly, that triggers the default legalization of i64 into i128 and breaks other things.
The handling here is a direct port of AIE2, I remember we had some issues with these types, but don't know all the details anymore

konstantinschwarz:
Thanks for making me go through 3 years old code and the hacks we implemented back then ;)
Turns out some of them are not required anymore, see more details in the commit message of b6c7daaf3a87f2e8f4f28cb3acc0078b721464fe.
The only hack needed now is for 128-bit vectors passed in 256-bit w registers.

andcarminati (approved):
LGTM.

PR #688, opened 2025-10-24
[AIE2P] Allow spilling vector/accumulator registers to registers instead of memory

F-Stuckmann:
Creating a combined register class for vector and accumulator registers changes the machine scheduling behavior because the combined register class is now used for register pressure calculation in the machine scheduler.
Therefore, vector and accumulator registers are considered as a single register pressure class and thus the machine scheduler ever so slightly changes the scheduling. This scheduling change changes register liveliness and causes spilling (sometimes even in the innermost loop). 
 
To mitigate this: 
1) Remove the combined register class for register pressure calculation during machine scheduling, so the machine scheduling does not change.
2) Updating WAW rewriter strategy to reach good IIs. 

Things to update in this PR: 
- unit tests
- fine tuning on the latest QoR
 
 P.S. For my sanity I continue https://github.com/Xilinx/llvm-aie/pull/678 here


F-Stuckmann:
outdated QoR

llvm/lib/Target/AIE/aie2p/AIE2PInstrInfo.cpp
```diff
@@ -1154,6 +1163,8 @@ AIE2PInstrInfo::getSpillPseudoExpandInfo(const MachineInstr &MI) const {
   case AIE2P::VST_EX_SPILL:
     return {{AIE2P::VST_dmx_sts_x_spill, AIE2P::sub_bfp16_x},
             {AIE2P::VST_E_SPILL, AIE2P::sub_bfp16_e}};
+  case AIE2P::VLDA_512_COMPOSED_REG_SPILL:
+    return {};
```
martien-de-jong, line 1188: This says 'nothing to expand' All other cases return something, so I'm wondering whether this should fall under 
`if (!MI.isPseudo())` in the top of the function

llvm/lib/CodeGen/RegisterPressure.cpp
```diff
@@ -999,6 +1009,12 @@ static void computeMaxPressureDelta(ArrayRef<unsigned> OldMaxPressureVec,
     if (PNew == POld) // No change in this set in the common case.
       continue;
 
+    // // Ignore Reg Pressure Set ID
+    const unsigned PSID = i;
+    if (std::find(IgnorePressureSetIDs.begin(), IgnorePressureSetIDs.end(),
```
andcarminati: Discussed offline, maybe we should try a single target hook, to which we pass the ID. All IDs a given by an enum, so it should be easy for the target to handle them.

llvm/lib/Target/AIE/aie2p/AIE2PInstrInfo.cpp
```diff
@@ -964,6 +965,8 @@ void AIE2PInstrInfo::storeRegToStackSlot(MachineBasicBlock &MBB,
     Opcode = AIE2P::ST_R_SPILL;
     SrcReg = ScratchReg;
     IsKill = true;
+  } else if (&AIE2P::spill_vec512_to_compositeRegClass == RC) {
```
martien-de-jong: I'm wondering why we don't use `regClassMatches()` here.

llvm/utils/TableGen/RegisterInfoEmitter.cpp
```diff
@@ -1060,6 +1060,23 @@ void RegisterInfoEmitter::runMCDesc(raw_ostream &OS) {
...
+  for (unsigned PSetID :
+       RegBank.getIgnoreRegPressureIndicesInMachineScheduler()) {
+    const RegUnitSet &Set = RegBank.getRegPressureSet(PSetID);
+    // Emit the ordered pressure set ID to match the IDs used by
+    // getRegClassPressureSets().
+    OS << Set.Order << ", // " << Set.Name << "\n";
+  }
+  // If no elements are in IgnoreRegPressureIndicesInMachineScheduler we need to
+  // add a no entry, so that tablegen does not throw an Exception.
+  OS << (uint16_t)-1 << ", // no reg pressure entry \n";
```
martien-de-jong: I think we should avoid C-casts. Moreover, I think you're casting at tablegen time, whereas I think you want to have the initialization do the cast.

martien-de-jong: Also, why would tablegen throw an exception?

llvm/lib/CodeGen/RegisterPressure.cpp
```diff
@@ -945,18 +948,25 @@ void RegPressureTracker::advance() {
...
   Delta.Excess = PressureChange();
   for (unsigned i = 0, e = OldPressureVec.size(); i < e; ++i) {
     unsigned POld = OldPressureVec[i];
     unsigned PNew = NewPressureVec[i];
     int PDiff = (int)PNew - (int)POld;
     if (!PDiff) // No change in this set in the common case.
       continue;
+
+    // // Ignore Reg Pressure Set ID
+    const unsigned PSID = i;
```
martien-de-jong, line 985: How about renaming the for loop's control variable to PSID?

llvm/include/llvm/CodeGen/TargetRegisterInfo.h
```diff
@@ -880,6 +887,20 @@ class TargetRegisterInfo : public MCRegisterInfo {
...
+  }
+  RegPressureIdxIgnoreMachineSched_iterator
+  regPressureSetIgnoreInMachineScheduler_end() const {
+    return RegPressureSetIdxIgnoreInMachineSchedulerEnd;
+  }
+  iterator_range<RegPressureIdxIgnoreMachineSched_iterator>
+  regPressureSetIndicesIgnoreInMachineScheduler() const {
+    return make_range(regPressureSetIgnoreInMachineScheduler_begin(),
+                      regPressureSetIgnoreInMachineScheduler_end());
+  }
```
martien-de-jong, line 889: Why don't we use ArrayRef for this interface?

llvm/utils/TableGen/Common/CodeGenRegisters.h
```diff
@@ -365,6 +367,8 @@ class CodeGenRegisterClass {
   /// Generate register pressure set for this register class and any class
   /// synthesized from it.
   bool GeneratePressureSet;
+  ///  Consider class's register pressure during Machine Scheduling
+  bool RegPressureInMachineScheduler;
```
martien-de-jong: nit: UseRegPressureInMachineScheduler?

llvm/lib/CodeGen/RegisterPressure.cpp
F-Stuckmann: We then have to modify multiple things at different stages. This sounds fragile. 
Would it be easier to have the Codegen changes in two commits? 
1) refactor for clang format 
2) actual change

This should minimize the merge conflicts.

llvm/utils/TableGen/RegisterInfoEmitter.cpp
F-Stuckmann: there are some sanity checks that expect an array to have at least an element.

llvm/lib/CodeGen/RegisterPressure.cpp
F-Stuckmann: would this lead to merge conflicts down the line?

llvm/utils/TableGen/Common/CodeGenRegisters.h
F-Stuckmann: I like it

llvm/include/llvm/CodeGen/TargetRegisterInfo.h
F-Stuckmann: Right, that is more efficient.

llvm/lib/Target/AIE/AIEWawRegRewriter.cpp
```diff
@@ -50,7 +50,7 @@ static cl::opt<bool> GPRRealloc("aie-gpr-realloc", cl::Hidden, cl::init(false),
                                 cl::desc("Re-allocate GPRs as well"));
 
 static cl::opt<bool> PreAlloc(
-    "aie-realloc-loopaware", cl::Hidden, cl::init(false),
+    "aie-realloc-loopaware", cl::Hidden, cl::init(true),
```
F-Stuckmann: make this a aie2p specific option?

llvm/lib/Target/AIE/aie2p/AIE2PInstrInfo.cpp
```diff
@@ -1047,7 +1050,13 @@ void AIE2PInstrInfo::loadRegFromStackSlot(MachineBasicBlock &MBB,
     BuildMI(MBB, I, DL, get(AIE2P::MOV_alu_mv_mv_mv_scl), DstReg)
         .addReg(Reg, getKillRegState(true));
     return;
+  } else if (&AIE2P::spill_vec512_to_compositeRegClass == RC) {
```
F-Stuckmann: oh here we are also missing reg class matching

llvm/lib/Target/AIE/aie2p/AIE2PTargetMachine.cpp
```diff
@@ -111,6 +112,7 @@ bool AIE2PPassConfig::addRegAssignAndRewriteOptimized() {
   }
   addPass(createGreedyRegisterAllocator());
   if (EnableWAWRegRewrite) {
+    PreAlloc = true;
```
F-Stuckmann: if (PreAlloc.getNumOccurrences() == 0) 
      PreAlloc = true;

This way the command line option is still being used and only when it is not overwritten it defaults to true in aie2p.

llvm/lib/CodeGen/RegisterPressure.cpp
andcarminati: > We then have to modify multiple things at different stages
 
I would like to minimize as much as possible CodeGen changes, merge conflicts are quite painful. We already need to specify the following steps to be able to break LR between two different classes:

1 - Create joint superclass.
2 - Implement TRI hook (LargestSuperClass).
3 - Handle pseudo spill code generation.
4 - Supply new TRI hook with the filtered pressure set IDs. (ie. `AIE2P::RegisterPressureSets::spill_vec512_to_composite`).

I can't see how this is fragile, specially if we constrain a good part of the implementation in `AIEBaseRegisterInfo`.

llvm/test/CodeGen/AIE/aie2p/end-to-end/ore-hardware-loops.ll
```diff
@@ -63,47 +63,51 @@ define weak_odr dso_local void @convert_bf16_to_bfp16(ptr noalias %in, ptr noali
...
+; CHECK-NEXT: 	vst.flush.512.conv	 [p2, sf, r26]
+; CHECK-NEXT: 	vst.flush.512.conv.2d	 [p2, sf, r26, d0]
+; CHECK-NEXT: 	ret	lr
+; CHECK-NEXT: 	nop	                                //  Delay Slot 5
+; CHECK-NEXT: 	nop	                                //  Delay Slot 4
+; CHECK-NEXT: 	nop	                                //  Delay Slot 3
+; CHECK-NEXT: 	nop	                                //  Delay Slot 2
+; CHECK-NEXT: 	nop	                                //  Delay Slot 1
 ; CHECK-NEXT: .Lfunc_end0:
+; CHECK-NEXT: 	.size	convert_bf16_to_bfp16, .Lfunc_end0-convert_bf16_to_bfp16
```
andcarminati: nit: we could drop everything after `CHECK-NEXT: .Lfunc_end0:`.

llvm/test/CodeGen/AIE/aie2p/ra/spill-vec-acc-Conv2D_DW_0-conv2d_dw.mir
```diff
@@ -0,0 +1,1704 @@
+# NOTE: Assertions have been autogenerated by utils/update_mir_test_checks.py UTC_ARGS: --version 4
```
andcarminati: Maybe check the name of this test.

llvm/test/CodeGen/AIE/aie2p/ra/spill-vec-acc-Conv2D_DW_0-conv2d_dw.mir
```diff
@@ -0,0 +1,1704 @@
...
+# See https://llvm.org/LICENSE.txt for license information.
+# SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+#
+# (c) Copyright 2025 Advanced Micro Devices, Inc. or its affiliates
+
+# RUN: llc -mtriple=aie2p -verify-machineinstrs --start-before=greedy --stop-after=virtregrewriter %s -o - | FileCheck %s
+
+# Example where spilling is reduced by the combined Vec/Acc Registerclass.
+--- |
+  source_filename = "Work/aie/0_0/src/0_0.cc"
```
andcarminati: Remove those information, for example, target triple is inferred from `-mtriple`

llvm/test/CodeGen/AIE/aie2p/ra/spill-vec-acc-Conv2D_DW_0-conv2d_dw.mir
```diff
@@ -0,0 +1,1704 @@
...
+  target datalayout = "e-m:e-p:20:32-i1:8:32-i8:8:32-i16:16:32-i32:32:32-f32:32:32-i64:32-f64:32-a:0:32-n32"
+  target triple = "aie2p"
+
+  $_Z9conv2d_dwILh1EaaaN3adf16io_buffer_configINS0_7extentsIJLj4294967295EEEENS0_7locking4syncENS0_10addressing6linearENS0_6marginILj0EEEEESA_SA_EvRNS0_9io_bufferIT0_NS0_9direction2inET3_EERNSB_IT1_SE_T4_EERNSB_IT2_NSD_3outET5_EER16conv2d_dw_params = comdat any
```
andcarminati: nit: simplify function name.

llvm/test/CodeGen/AIE/aie2p/ra/spill-vec-acc-Conv2D_DW_0-conv2d_dw.mir
```diff
@@ -0,0 +1,1704 @@
...
+  !2 = !{i32 1, !"wchar_size", i32 4}
+  !3 = !{!"clang version 20.0.0git (/scratch/work-ref/clang 4dcac713ab4f9d5b6562793a2a9eba31bfb14e52)"}
+  !4 = !{!5, !8, i64 32}
+  !5 = !{!"_ZTS16conv2d_dw_params", !6, i64 0, !6, i64 4, !8, i64 8, !8, i64 12, !6, i64 16, !6, i64 17, !9, i64 32}
+  !6 = !{!"omnipotent char", !7, i64 0}
+  !7 = !{!"Simple C++ TBAA"}
+  !8 = !{!"int", !6, i64 0}
+  !9 = !{!"_ZTS25conv2d_dw_internal_params", !8, i64 0, !8, i64 4, !8, i64 8, !8, i64 12, !8, i64 16, !8, i64 20, !8, i64 24, !8, i64 28, !8, i64 32, !8, i64 36, !8, i64 40, !8, i64 44, !8, i64 48, !8, i64 52, !8, i64 56, !8, i64 60, !8, i64 64, !8, i64 68, !8, i64 72, !8, i64 76, !8, i64 80, !8, i64 84, !8, i64 88, !8, i64 92, !8, i64 96, !8, i64 100, !8, i64 104, !8, i64 108, !8, i64 112, !8, i64 116, !8, i64 120, !8, i64 124, !8, i64 128, !8, i64 132}
+  !10 = !{!11, !12, i64 0}
+  !11 = !{!"_ZTSN3adf6detail14io_buffer_baseIaNS_7locking4syncEEE", !12, i64 0}
```
andcarminati: For each of these mangled names, you could give a simple one, like "A", "B" and so on.

llvm/test/CodeGen/AIE/aie2p/ra/spill-vec-acc-Conv2D_DW_0-conv2d_dw.mir
```diff
@@ -0,0 +1,1704 @@
...
+  !129 = distinct !{!129, !"_ZN3aie6load_vILj32EL15aie_dm_resource0ETkNS_21DecoratedElemBaseTypeEU3AS6aEENS_6vectorIN22aie_dm_resource_removeIT1_E4typeEXT_EEEPKS5_"}
+  !130 = distinct !{!130, !51, !52}
+
+...
+---
+name:            _Z9conv2d_dwILh1EaaaN3adf16io_buffer_configINS0_7extentsIJLj4294967295EEEENS0_7locking4syncENS0_10addressing6linearENS0_6marginILj0EEEEESA_SA_EvRNS0_9io_bufferIT0_NS0_9direction2inET3_EERNSB_IT1_SE_T4_EERNSB_IT2_NSD_3outET5_EER16conv2d_dw_params
+alignment:       16
+exposesReturnsTwice: false
+legalized:       true
+regBankSelected: true
```
andcarminati: I guess that much of this information can be removed.

llvm/test/CodeGen/AIE/aie2p/ra/spill-vec-acc-Conv2D_DW_0-conv2d_dw.mir
```diff
@@ -0,0 +1,1704 @@
...
+  declare void @llvm.aie2p.set.ctrl.reg(i32, i32) #0
+
+  define weak_odr dso_local void @_Z9conv2d_dwILh1EaaaN3adf16io_buffer_configINS0_7extentsIJLj4294967295EEEENS0_7locking4syncENS0_10addressing6linearENS0_6marginILj0EEEEESA_SA_EvRNS0_9io_bufferIT0_NS0_9direction2inET3_EERNSB_IT1_SE_T4_EERNSB_IT2_NSD_3outET5_EER16conv2d_dw_params(ptr noalias nonnull align 4 dereferenceable(8) %input, ptr noalias nonnull align 4 dereferenceable(8) %weights, ptr noalias nonnull align 4 dereferenceable(8) %output, ptr nonnull align 32 dereferenceable(192) %layer_param) local_unnamed_addr #1 comdat personality ptr @__gxx_personality_v0 {
+  entry:
+    %internal_param = getelementptr inbounds nuw i8, ptr %layer_param, i20 32
+    %0 = load i32, ptr %internal_param, align 32, !tbaa !4
+    %cmp414.not = icmp eq i32 %0, 0
+    br i1 %cmp414.not, label %for.cond.cleanup, label %for.body.lr.ph
+
+  for.body.lr.ph:                                   ; preds = %entry
```
andcarminati: Could we also reduce a bit this test? Sometimes the only way is manually =/.

For example, all those s32 can be simplified, because they are not increasing vec/acc pressure.

F-Stuckmann:
Register Pressure Ignoring is now chained through to the Machine Scheduler and AIEMachineScheduler, so that the Pre-RA scheduling remains constant. This solves the stack regressions. 
Updated QoR: 

Core_Compute_Insn_Count
Performance: -0.32%
Core_PMSize_perf_sorted_absolute
PM: -1.78 Bytes
Core_StackSize_perf_sorted_absolute
Stack: -24.4 Bytes

llvm/lib/CodeGen/RegisterPressure.cpp
```diff
@@ -274,6 +285,7 @@ void RegPressureTracker::init(const MachineFunction *mf,
   MBB = mbb;
   this->TrackUntiedDefs = TrackUntiedDefs;
   this->TrackLaneMasks = TrackLaneMasks;
+  IgnoreRegPressureSets = TRI->getIgnoreRegPressureSets();
```
andcarminati: I am a bit confused, why do we have IgnoreRegPressureSets sometimes as a field and sometimes as a parameter? Could we unify as a parameter?

F-Stuckmann: We have class wrapper methods that call static methods.  
We have to pass the IgnoreRegPressureSets to the static methods. 

Alternatively we could refactor the static methods into their wrapper methods, but that will increase our merge conflict probability, so I am not in favor of this approach.

F-Stuckmann:
updates: 
- reduced pre-scheduling unit-test
- single loopaware advantagous unit-test (end to end) 
- fixed ci failure 

I am still working on a reduced spill example.

llvm/utils/TableGen/RegisterInfoEmitter.cpp
```diff
@@ -1060,6 +1060,21 @@ void RegisterInfoEmitter::runMCDesc(raw_ostream &OS) {
 
   OS << "};\n\n";
 
+  const std::string IgnoreRegPressureIdxMachSchedName =
```
andcarminati: I still think that this is much more general than Machine Scheduler.

llvm/test/CodeGen/AIE/aie2p/ra/spill-vec-acc-Conv2D_DW_0-conv2d_dw.mir
F-Stuckmann: I now have all the tests reduced

llvm/lib/CodeGen/MachineScheduler.cpp
```diff
@@ -1399,13 +1399,16 @@ void ScheduleDAGMILive::initRegPressure() {
...
 
+  auto RegPressureIgnoreSets = TRI->getIgnoreRegPressureSets();
   // Cache the list of excess pressure sets in this region. This will also track
   // the max pressure in the scheduled code for these sets.
   RegionCriticalPSets.clear();
   const std::vector<unsigned> &RegionPressure =
     RPTracker.getPressure().MaxSetPressure;
   for (unsigned i = 0, e = RegionPressure.size(); i < e; ++i) {
     unsigned Limit = RegClassInfo->getRegPressureSetLimit(i);
+    if (find(RegPressureIgnoreSets, i) != RegPressureIgnoreSets.end())
```
andcarminati: nit: move this check just after the for.

llvm/lib/Target/AIE/aie2p/AIE2PInstrInfo.cpp
F-Stuckmann: nice catch!

llvm/utils/TableGen/RegisterInfoEmitter.cpp
F-Stuckmann: shoot, I missed that in my renaming spree

llvm/lib/CodeGen/RegisterPressure.cpp
```diff
@@ -48,28 +51,35 @@ using namespace llvm;
...
+                                LaneBitmask PrevMask, LaneBitmask NewMask,
+                                ArrayRef<uint16_t> IgnoreRegPressureSets) {
   assert((NewMask & ~PrevMask).none() && "Must not add bits");
   if (NewMask.any() || PrevMask.none())
     return;
 
   PSetIterator PSetI = MRI.getPressureSets(Reg);
   unsigned Weight = PSetI.getWeight();
   for (; PSetI.isValid(); ++PSetI) {
+    if (find(IgnoreRegPressureSets, *PSetI) != IgnoreRegPressureSets.end())
```
andcarminati: Could we have a helper function for this? This is a aesthetic change, this same line repeats several times.

Even better:
```
is_contained(R &&Range, const E &Element)
```

llvm/lib/Target/AIE/aie2p/AIE2PInstrInfo.cpp
```diff
@@ -1047,7 +1051,14 @@ void AIE2PInstrInfo::loadRegFromStackSlot(MachineBasicBlock &MBB,
     BuildMI(MBB, I, DL, get(AIE2P::MOV_alu_mv_mv_mv_scl), DstReg)
         .addReg(Reg, getKillRegState(true));
     return;
+  } else if (regClassMatches(AIE2P::spill_vec512_to_compositeRegClass, RC,
+                             DstReg)) {
+    Opcode = AIE2P::VLDA_512_COMPOSED_REG_SPILL;
   } else {
+    std::prev(I)->dump();
```
andcarminati: Please, remove debug lines.

llvm/test/CodeGen/AIE/aie2p/end-to-end/sigmoid_int8_1.ll
```diff
@@ -0,0 +1,112 @@
+; NOTE: Assertions have been autogenerated by utils/update_llc_test_checks.py
+;
+; This file is licensed under the Apache License v2.0 with LLVM Exceptions.
+; See https://llvm.org/LICENSE.txt for license information.
+; SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+;
+; (c) Copyright 2025 Advanced Micro Devices, Inc. or its affiliates
+
+; RUN: llc -mtriple=aie2p  %s -o - | FileCheck %s
```
andcarminati: Run the verifier?

llvm/include/llvm/CodeGen/RegisterPressure.h
```diff
@@ -394,6 +399,8 @@ class RegPressureTracker {
   /// Live-through pressure.
   std::vector<unsigned> LiveThruPressure;
 
+  ArrayRef<uint16_t> IgnoreRegPressureSets;
```
andcarminati: nit: add a comment for this field, according to the other fields.

andcarminati: Like you have in TRI:
```
// List of Register Pressure IDs to ignore
```

llvm/test/CodeGen/AIE/aie2p/end-to-end/sigmoid_int8_1.ll
F-Stuckmann: hey it does not crash. I'll take it.

llvm/lib/CodeGen/TargetRegisterInfo.cpp
```diff
@@ -51,12 +54,15 @@ static cl::opt<unsigned>
...
     const TargetRegisterInfoDesc *ID, regclass_iterator RCB,
-    regclass_iterator RCE, const char *const *SRINames,
+    regclass_iterator RCE, const uint16_t *IgnoreRegPressureSetsBegin,
+    unsigned IgnoreRegPressureSetsSize, const char *const *SRINames,
     const SubRegCoveredBits *SubIdxRanges, const LaneBitmask *SRILaneMasks,
     LaneBitmask SRICoveringLanes, const RegClassInfo *const RCIs,
     const MVT::SimpleValueType *const RCVTLists, unsigned Mode)
     : InfoDesc(ID), SubRegIndexNames(SRINames), SubRegIdxRanges(SubIdxRanges),
       SubRegIndexLaneMasks(SRILaneMasks), RegClassBegin(RCB), RegClassEnd(RCE),
+      IgnoreRegPressureSets(IgnoreRegPressureSetsBegin,
```
andcarminati: Why do we have begin and size? Why not just an ArrayRef as argument?

llvm/unittests/CodeGen/MFCommon.inc
```diff
@@ -32,12 +32,16 @@ protected:
 
 static TargetRegisterClass *const BogusRegisterClasses[] = {nullptr};
 
+static const uint16_t BogusIgnoreRegPressureSets[] = {
```
andcarminati, line 35: Dummy is more common.

F-Stuckmann: I think in this context it is clearer to keep the same naming scheme as the rest (i.e. BogusRegisterClasses)

llvm/lib/Target/AIE/aie2p/AIE2PTargetMachine.cpp
```diff
@@ -111,6 +112,9 @@ bool AIE2PPassConfig::addRegAssignAndRewriteOptimized() {
   }
   addPass(createGreedyRegisterAllocator());
   if (EnableWAWRegRewrite) {
+    // If PreAlloc isn't set by a command line argument, set it to True
```
andcarminati: Discussed offline, move this handling to `AIEWawRegRewriter`. It will help to set this flag in a per-loop basis in the future.

llvm/lib/CodeGen/TargetRegisterInfo.cpp
F-Stuckmann: nice!

F-Stuckmann: Passing an ArrayRef with size 0 is ambigous when generating the `TargetRegisterInfo.inc` files. So I would rather have C++ handle generating a zero size ArrayRef in `TargetRegisterInfo.cpp`.

llvm/lib/Target/AIE/aie2p/AIE2PRegisterInfo.td
```diff
@@ -983,6 +983,7 @@ def spill_eM_to_eR : AIE2PScalarRegisterClass<(add eM, eR, eDN)>;
 def spill_eDN_to_eR : AIE2PScalarRegisterClass<(add eDN, eR)>;
 def spill_eDJ_to_eR : AIE2PScalarRegisterClass<(add eDJ, eR, eDN)>;
 def spill_eDC_to_eR : AIE2PScalarRegisterClass<(add eDC, eR)>;
+def spill_vec512_to_composite : AIE2PVector512RegisterClass<(add mXm, mBMm, lfh0, lfh1, lfl0, lfl1, sfl, sfh, lfe)>;
```
niwinanto: Why we do only for 512-bit case?

llvm/lib/Target/AIE/aie2p/AIE2PRegisterInfo.cpp
```diff
@@ -482,6 +484,8 @@ AIE2PRegisterInfo::getLargestLegalSuperClass(const TargetRegisterClass *RC,
 
   if (AIE2P::eSRegClass.hasSubClassEq(RC))
     return &AIE2P::spill_eS_to_eRRegClass;
+  if (RC == &AIE2P::ACC512RegClass || RC == &AIE2P::VEC512RegClass)
```
niwinanto, line 481: check: I remember getting into some situation, where x register is used for bm and further down the line subvector of bm is trying to access, which does not exist. Do you get into some issues like that?

llvm/test/CodeGen/AIE/aie2p/load-store-unaligned.ll
```diff
@@ -301,66 +302,70 @@ define dso_local void @test_load_store_unaligned(<8 x i16> noundef %a, <4 x i32>
...
 ; CHECK-NEXT:    vextract.32 r3, x8, #3, vaddsign1
 ; CHECK-NEXT:    mova dj0, #20
 ; CHECK-NEXT:    st r2, [p0, #8]
 ; CHECK-NEXT:    st r3, [p0, #12]
 ; CHECK-NEXT:    st.s16 r18, [p3, dj0]
-; CHECK-NEXT:    vmov bmll0, x0
+; CHECK-NEXT:    mov p1, sp
 ; CHECK-NEXT:    padda [p1], #-448
-; CHECK-NEXT:    vmov x2, bmll0
+; CHECK-NEXT:    // implicit-def: $bmll0
```
niwinanto, line 322: What is `// implicit-def: $bmll0` at this stage?

llvm/lib/Target/AIE/aie2p/AIE2PRegisterInfo.cpp
F-Stuckmann: no we don't run into any issues.

llvm/lib/Target/AIE/aie2p/AIE2PRegisterInfo.td
andcarminati: I guess we avoid the need of the artificial register classes modeling common lanes.

llvm/lib/Target/AIE/aie2p/AIE2PRegisterInfo.cpp
niwinanto: Maybe `def spill_vec512_to_composite : AIE2PVector512RegisterClass<(add mXm, mBMm, lfh0, lfh1, lfl0, lfl1, sfl, sfh, lfe)>;` register allocator try to get the x registers first and we might not be getting into situations like that. Can we do an experiment to reorder the list, such that start with BM register? We will get more confidence.

llvm/utils/TableGen/RegisterInfoEmitter.cpp
```diff
@@ -1201,6 +1215,9 @@ void RegisterInfoEmitter::runTargetDesc(raw_ostream &OS) {
   // Start out by emitting each of the register classes.
   const auto &RegisterClasses = RegBank.getRegClasses();
   const auto &SubRegIndices = RegBank.getSubRegIndices();
+  const auto &IgnoreRegPressureSets = RegBank.getIgnoreRegPressureSets();
```
andcarminati: You take just the size, no?

llvm/lib/CodeGen/TargetRegisterInfo.cpp
andcarminati: You don't need the size, you can just use the other constructor to avoid 0 -> nullptr confusion:

```
ArrayRef<uint16_t>(/*begin*/ IgnoreRegPressureSetsName, /*end*/ IgnoreRegPressureSetsName),
```

You can still pass an `ArrayRef` directly :-).

llvm/include/llvm/CodeGen/RegisterPressure.h
```diff
@@ -157,7 +160,8 @@ class PressureDiff {
   const_iterator end() const { return &PressureChanges[MaxPSets]; }
 
   void addPressureChange(Register RegUnit, bool IsDec,
-                         const MachineRegisterInfo *MRI);
+                         const MachineRegisterInfo *MRI,
+                         ArrayRef<uint16_t> IgnoreRegPressureSets);
```
andcarminati: Is this redundant? You can retrieve the TRI from MRI:

```
MRI->getTargetRegisterInfo()
```

So you can ask for `IgnoreRegPressureSets` directly.

llvm/include/llvm/CodeGen/RegisterPressure.h
```diff
@@ -220,7 +224,8 @@ class PressureDiffs {
   /// Record pressure difference induced by the given operand list to
   /// node with index \p Idx.
   void addInstruction(unsigned Idx, const RegisterOperands &RegOpers,
-                      const MachineRegisterInfo &MRI);
+                      const MachineRegisterInfo &MRI,
+                      ArrayRef<uint16_t> IgnoreRegPressureSets);
```
andcarminati: Also redundant, we could reduce interface changes.

llvm/lib/CodeGen/RegisterPressure.cpp
```diff
@@ -48,28 +51,35 @@ using namespace llvm;
 /// Increase pressure for each pressure set provided by TargetRegisterInfo.
 static void increaseSetPressure(std::vector<unsigned> &CurrSetPressure,
                                 const MachineRegisterInfo &MRI, unsigned Reg,
-                                LaneBitmask PrevMask, LaneBitmask NewMask) {
+                                LaneBitmask PrevMask, LaneBitmask NewMask,
```
andcarminati: Also here, I guess you could simplify the interface.

llvm/include/llvm/CodeGen/RegisterPressure.h
```diff
@@ -394,6 +399,9 @@ class RegPressureTracker {
   /// Live-through pressure.
   std::vector<unsigned> LiveThruPressure;
 
+  /// List of Register Pressure IDs to ignore
```
andcarminati: Probably this is also redundant.

llvm/lib/Target/AIE/aie2p/AIE2PRegisterInfo.cpp
F-Stuckmann: did that and I still get no additional errors.

llvm/utils/TableGen/RegisterInfoEmitter.cpp
F-Stuckmann: yeah

llvm/test/CodeGen/AIE/aie2p/load-store-unaligned.ll
F-Stuckmann: bmll0 and bmll1 come in as livein registers into `bb.0.entry`.

F-Stuckmann: why bmll0 is implicit def but bmll1 isn't, isn't very clear to me.

llvm/lib/Target/AIE/AIEWawRegRewriter.cpp
```diff
@@ -205,6 +208,16 @@ MCPhysReg AIEWawRegRewriter::getAssignedPhysReg(const Register Reg) const {
   return Reg;
 }
 
+void AIEWawRegRewriter::setArchSpecificPassOptions() {
+  if (!MF->getTarget().getTargetTriple().isAIE2P())
+    return;
+
+  // NumOccurrences increments if it is set by a command line argument
+  bool UseDefaultPreAlloc = PreAlloc.getNumOccurrences() == 0;
```
andcarminati: nit: `const`.

F-Stuckmann: autsch, that one hurt

llvm/lib/Target/AIE/aie2p/AIE2PRegisterInfo.cpp
niwinanto: Thanks @F-Stuckmann for testing this.

llvm/lib/CodeGen/TargetRegisterInfo.cpp
andcarminati: Did you find some blocker for this approach? I mean, the pointer-pointer constructor?

llvm/lib/CodeGen/TargetRegisterInfo.cpp
```diff
@@ -51,12 +54,14 @@ static cl::opt<unsigned>
 
 TargetRegisterInfo::TargetRegisterInfo(
     const TargetRegisterInfoDesc *ID, regclass_iterator RCB,
-    regclass_iterator RCE, const char *const *SRINames,
-    const SubRegCoveredBits *SubIdxRanges, const LaneBitmask *SRILaneMasks,
-    LaneBitmask SRICoveringLanes, const RegClassInfo *const RCIs,
-    const MVT::SimpleValueType *const RCVTLists, unsigned Mode)
+    regclass_iterator RCE, ArrayRef<uint16_t> IgnoreRegPressureSets,
```
andcarminati, line 57: Oh, I like this way!

llvm/lib/Target/AIE/aie2p/AIE2PInstrInfo.cpp
```diff
@@ -1262,6 +1271,34 @@ bool AIE2PInstrInfo::expandPostRAPseudo(MachineInstr &MI) const {
     MI.eraseFromParent();
     return true;
   }
+  case AIE2P::VLDA_512_COMPOSED_REG_SPILL: {
+    unsigned int Opcode;
+    if (AIE2P::VEC512RegClass.contains(MI.getOperand(0).getReg())) {
```
andcarminati: nit: could we refactor `MI.getOperand(0).getReg()`?

F-Stuckmann: I also found a const opportunity.

llvm/lib/Target/AIE/aie2p/AIE2PInstrInfo.cpp
```diff
@@ -1287,11 +1288,12 @@ bool AIE2PInstrInfo::expandPostRAPseudo(MachineInstr &MI) const {
   }
   case AIE2P::VST_512_COMPOSED_REG_SPILL: {
     unsigned int Opcode;
-    if (AIE2P::VEC512RegClass.contains(MI.getOperand(0).getReg())) {
+    const Register Dst = MI.getOperand(0).getReg();
```
andcarminati: nit: Src?

andcarminati:
Maybe we could keep only the most recent QoR results to avoid confusion.

llvm/lib/Target/AIE/aie2p/AIE2PRegisterInfo.cpp
```diff
@@ -476,6 +478,8 @@ AIE2PRegisterInfo::getLargestLegalSuperClass(const TargetRegisterClass *RC,
...
```
F-Stuckmann, line 481: harmonizing this with `hasSubClassEq` will change register coalescing and thus degrade performance. I have added a `register-coalescer-regression.mir` test to capture this case.

andcarminati: I could be nice ho have a comment explaining this.

F-Stuckmann: next fixup ;)

llvm/lib/Target/AIE/aie2p/AIE2PInstrInfo.cpp
```diff
@@ -1082,7 +1091,7 @@ void AIE2PInstrInfo::loadRegFromStackSlot(
 SmallVector<AIEBaseInstrInfo::AIEPseudoExpandInfo, 4>
 AIE2PInstrInfo::getSpillPseudoExpandInfo(const TargetRegisterInfo &TRI,
                                          MachineInstr &MI) const {
-  if (!MI.isPseudo())
+  if (!MI.isPseudo() || MI.getOpcode() == AIE2P::VLDA_512_COMPOSED_REG_SPILL)
```
martien-de-jong: nit: this condition was crystal clear. If it's now a pseudo, there's nothing to expand. The switch below says how to deal with a particular pseudo. I would dedicate a case to this, explaining why we don't expand it and returning {} there.

llvm/lib/Target/AIE/aie2p/AIE2PInstrInfo.td
```diff
@@ -321,6 +321,7 @@ def VST_FIFO_SPILL : Pseudo<(outs ), (ins FIFO1024:$src, c16n_step64:$imm), "vst
 def VST_PLFR_SPILL : Pseudo<(outs ), (ins ePSRFLdF:$src, c16n_step64:$imm), "vst_plfr_spill", "${src}, [sp, $imm]">;
 def VST_EX_SPILL : Pseudo<(outs ), (ins VEC576:$src, c16n_step64:$imm), "vst_ex_spill", "${src}, [sp, $imm]">;
 def VST_E_SPILL : Pseudo<(outs ), (ins EXPVEC64:$src, c12n_step4:$imm), "vst_e_spill", "$src, [sp, $imm]">;
+def VST_512_COMPOSED_REG_SPILL : Pseudo<(outs ), (ins spill_vec512_to_composite:$src, c16n_step64:$imm), "vst_512_composed_reg_spill", "${src}, [sp, $imm]">;
```
martien-de-jong, line 324: CHECK: Perhaps look forward to AIE2PS, and check whether this generalizes nicely.

llvm/lib/Target/AIE/aie2p/AIE2PRegisterInfo.cpp
martien-de-jong, line 481: Q: In other places we have a connection to AIE2P::FIFO512RegClass. Why not here? Related question, could we generalize this 'shared spill class' mechanism and have these code snippets data-driven?

llvm/include/llvm/Target/Target.td
```diff
@@ -369,6 +369,9 @@ class RegisterClass<string namespace, list<ValueType> regTypes, int alignment,
   // classes will be resolved in the order defined by this value, with lower values
   // taking precedence over higher ones. Ties are resolved by enumeration order.
   int BaseClassOrder = ?;
+
+  // Consider Register Pressure during Machine Scheduling
```
martien-de-jong: nit:

- of this class
- Current use is in (PreRA) MachineScheduling, but we may want to have a more abstract description

llvm/lib/Target/AIE/aie2p/AIE2PRegisterInfo.td
```diff
@@ -983,7 +983,10 @@ def spill_eM_to_eR : AIE2PScalarRegisterClass<(add eM, eR, eDN)>;
 def spill_eDN_to_eR : AIE2PScalarRegisterClass<(add eDN, eR)>;
 def spill_eDJ_to_eR : AIE2PScalarRegisterClass<(add eDJ, eR, eDN)>;
 def spill_eDC_to_eR : AIE2PScalarRegisterClass<(add eDC, eR)>;
-def spill_vec512_to_composite : AIE2PVector512RegisterClass<(add mXm, mBMm, lfh0, lfh1, lfl0, lfl1, sfl, sfh, lfe)>;
+
+let ConsiderInMachineScheduling = false in {
```
martien-de-jong: nit: I always think of this `let ... in {}` construct to be a hack for lazy typers. For a single instance I would just do the regular constructor syntax, adding this single assignment in the body

llvm/unittests/CodeGen/MFCommon.inc
```diff
@@ -32,12 +32,18 @@ protected:
...
+    65535, // no reg pressure entry
```
martien-de-jong, line 36: CHECK: this value is just a placeholder to avoid zero-element arrays. The fact that it looks like USHORT_MAX is irrelevant, since its value will never be inspected. Hence, it could just be zero?

llvm/unittests/CodeGen/MFCommon.inc
```diff
@@ -32,12 +32,18 @@ protected:
...
 class BogusRegisterInfo : public TargetRegisterInfo {
 public:
   BogusRegisterInfo()
-      : TargetRegisterInfo(nullptr, BogusRegisterClasses, BogusRegisterClasses,
-                           nullptr, nullptr, nullptr, LaneBitmask(~0u), nullptr,
-                           nullptr) {
+      : TargetRegisterInfo(
+            nullptr, BogusRegisterClasses, BogusRegisterClasses,
+            ArrayRef<uint16_t>(BogusIgnoreRegPressureSets,
+                               BogusIgnoreRegPressureSets + size_t{0}),
```
martien-de-jong: I think this passes two pointers. I thought you wanted to use the (ptr, size_t) constructor.

llvm/utils/TableGen/Common/CodeGenRegisters.h
```diff
@@ -632,6 +636,10 @@ class CodeGenRegBank {
   // Give each register unit set an order based on sorting criteria.
   std::vector<unsigned> RegUnitSetOrder;
 
+  // Ignore Pressure Sets in Register Pressure Calculation in the Machine
+  // Scheduler.
```
martien-de-jong: A slightly more descriptive comment? I am guessing that it is a list of indices?

llvm/utils/TableGen/RegisterInfoEmitter.cpp
```diff
@@ -1060,6 +1060,20 @@ void RegisterInfoEmitter::runMCDesc(raw_ostream &OS) {
...
+  OS << "extern const uint16_t " << IgnoreRegPressureSetsName << "[] = {\n";
+  for (unsigned PSetID : RegBank.getIgnoreRegPressureSets()) {
+    const RegUnitSet &Set = RegBank.getRegPressureSet(PSetID);
+    // Emit the ordered pressure set ID to match the IDs used by
+    // getRegClassPressureSets().
+    OS << Set.Order << ", // " << Set.Name << "\n";
+  }
+  // If no elements are in IgnoreRegPressureSets we need to
+  // add a no entry, so that tablegen does not throw an Exception.
+  OS << RegBank.getNumRegPressureSets() << ", // no reg pressure entry \n";
```
martien-de-jong: Could you explain why we emit this particular number?

llvm/utils/TableGen/RegisterInfoEmitter.cpp
```diff
@@ -1201,6 +1215,10 @@ void RegisterInfoEmitter::runTargetDesc(raw_ostream &OS) {
   // Start out by emitting each of the register classes.
   const auto &RegisterClasses = RegBank.getRegClasses();
   const auto &SubRegIndices = RegBank.getSubRegIndices();
+  const auto &IgnoreRegPressureSetsSize =
```
martien-de-jong: Why a reference?

llvm/lib/CodeGen/RegisterPressure.cpp
```diff
@@ -999,6 +1029,11 @@ static void computeMaxPressureDelta(ArrayRef<unsigned> OldMaxPressureVec,
     if (PNew == POld) // No change in this set in the common case.
       continue;
 
+    // // Ignore Reg Pressure Set ID
```
martien-de-jong: nit: double //

llvm/lib/CodeGen/RegisterPressure.cpp
```diff
@@ -951,12 +972,20 @@ static void computeExcessPressureDelta(ArrayRef<unsigned> OldPressureVec,
...
   for (unsigned i = 0, e = OldPressureVec.size(); i < e; ++i) {
     unsigned POld = OldPressureVec[i];
     unsigned PNew = NewPressureVec[i];
     int PDiff = (int)PNew - (int)POld;
     if (!PDiff) // No change in this set in the common case.
       continue;
+
+    // // Ignore Reg Pressure Set ID
+    const unsigned PSID = i;
+    if (find(IgnorePressureSetIDs, PSID) != IgnorePressureSetIDs.end())
```
martien-de-jong: In other places you use is_contained(). I also wonder whether TRI->ignorePressureSet(PSID) would cover almost all use cases.

llvm/lib/Target/AIE/aie2p/AIE2PRegisterInfo.cpp
F-Stuckmann: Without this narrow RC definition, we do not allocate Vector or Accumulator registers during register coalescing, but the spill_vec512 register class. 
This then changes the register pressure and therefore also the machine scheduling to our disadvantage.

llvm/include/llvm/Target/Target.td
F-Stuckmann: This is not entirely true. Register pressure is also used in pre pipelining.

llvm/unittests/CodeGen/MFCommon.inc
F-Stuckmann: yes it could just be zero, but i wanted a nicely recognizeable value, so the User immediately sees that something is fishy with the last value.

llvm/lib/Target/AIE/AIEWawRegRewriter.cpp
```diff
@@ -205,6 +208,16 @@ MCPhysReg AIEWawRegRewriter::getAssignedPhysReg(const Register Reg) const {
   return Reg;
 }
 
+void AIEWawRegRewriter::setArchSpecificPassOptions() {
+  if (!MF->getTarget().getTargetTriple().isAIE2P())
```
martien-de-jong, line 214: Maybe add a justification for this AIE2P exception. I gather it's because we don't track performance on other targets, which is not related to any technical feature of the compiler. Also note that this will be replaced by loop-specific tuning.

llvm/unittests/CodeGen/MFCommon.inc
F-Stuckmann: I suppose we could also call ` ArrayRef<uint16_t>(BogusIgnoreRegPressureSets, size_t{0})`.
I don't really have a strong opinion on how to call the ArrayRef constructor.

llvm/lib/Target/AIE/aie2p/AIE2PTargetMachine.cpp
```diff
@@ -4,7 +4,7 @@
 // See https://llvm.org/LICENSE.txt for license information.
 // SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
 //
-// (c) Copyright 2024 Advanced Micro Devices, Inc. or its affiliates
+// (c) Copyright 2024-2025 Advanced Micro Devices, Inc. or its affiliates
```
martien-de-jong: Put it in the commit where it changed? (also the next one)

llvm/utils/TableGen/Common/CodeGenRegisters.h
F-Stuckmann: yes

llvm/utils/TableGen/RegisterInfoEmitter.cpp
F-Stuckmann: we could emit any number really. I could also go again with max uint16. There is no real preference here since the comment already notifies the user, that we ignore this entry.

llvm/lib/CodeGen/RegisterPressure.cpp
F-Stuckmann: damn I missed that one.

llvm/utils/TableGen/RegisterInfoEmitter.cpp
martien-de-jong: "we add a no entry" could be a special entry whose value makes it be ignored, or a value that is never read and hence ignored. When we generate a non-trivial value, it suggests the former. Also, we do not do it because tablegen would otherwise throw an exception (which? why? what's wrong), but because C forbids empty arrays.
Perhaps 'We need a place holder to avoid empty arrays' would say it all.

llvm/lib/Target/AIE/aie2p/AIE2PInstrInfo.cpp
F-Stuckmann: lol, that was my first approach XD

llvm/lib/Target/AIE/aie2p/AIE2PRegisterInfo.cpp
F-Stuckmann: I have a unit test to specifically not select a spill_vec reg class during register coalescing.

llvm/include/llvm/Target/Target.td
martien-de-jong: So is there a good reason why the prepipeliner should consider this in register pressure but the MachineScheduler not?

llvm/unittests/CodeGen/MFCommon.inc
martien-de-jong: But `BogusIgnoreRegPressureSets`, `BogusIgnoreRegPressureSets + 0` and `BogusIgnoreRegPressureSets + size_t{0}` all have the same type and value. It just looks awkward.

llvm/lib/Target/AIE/aie2p/AIE2PTargetMachine.cpp
F-Stuckmann: I missed that in cleaning up

llvm/unittests/CodeGen/MFCommon.inc
F-Stuckmann: oh i get this comment now. Going with pointers is indeed awkward when we don't do it in codegen the same way.

llvm/lib/Target/AIE/AIEWawRegRewriter.cpp
```diff
@@ -205,6 +208,18 @@ MCPhysReg AIEWawRegRewriter::getAssignedPhysReg(const Register Reg) const {
   return Reg;
 }
 
+void AIEWawRegRewriter::setArchSpecificPassOptions() {
+  // todo: use loop classes to enable WAW strategies.
```
andcarminati: nit: `TODO`

llvm/lib/CodeGen/MachineScheduler.cpp
```diff
@@ -1399,12 +1399,15 @@ void ScheduleDAGMILive::initRegPressure() {
            BotRPTracker.getPos() == priorNonDebug(RegionEnd, RegionBegin))) &&
          "Can't find the region bottom");
 
+  auto IgnoreRegPressureSets = TRI->getIgnoreRegPressureSets();
```
andcarminati: Sometimes we use auto, sometimes we expand the type (see below). We could harmonize.

F-Stuckmann:
Since my markdown is broken the raw csv file for the QoR. 

[Core_Compute_Insn_Count_perf_sorted.csv](https://github.com/user-attachments/files/23456179/Core_Compute_Insn_Count_perf_sorted.csv)

andcarminati:
Here we go:
```
|-----------------------------------------------------------|-----------------------|--------------------------|---------------|
| Core_Compute_Insn_Count                                   | mllib_reference_peano | mllib_PR-loopaware_peano | Total diff    |
|-----------------------------------------------------------|-----------------------|--------------------------|---------------|
| LayerNormC8Part2_aie2_int8_0                              |                  7858 |                     8434 | REGR(+7.33%)  |
|-----------------------------------------------------------|-----------------------|--------------------------|---------------|
| AvgPool2D_0                                               |                  1027 |                     1076 | REGR(+4.77%)  |
|-----------------------------------------------------------|-----------------------|--------------------------|---------------|
| AvgPool2D_int8_0                                          |                  1027 |                     1076 | REGR(+4.77%)  |
|-----------------------------------------------------------|-----------------------|--------------------------|---------------|
| AvgPool2D_1                                               |                   739 |                      772 | REGR(+4.47%)  |
|-----------------------------------------------------------|-----------------------|--------------------------|---------------|
| InterpolateLinear1D_AIE2_bfloat16                         |                 11438 |                    11822 | REGR(+3.36%)  |
|-----------------------------------------------------------|-----------------------|--------------------------|---------------|
| Transpose_aie2_bf16_021                                   |                  2057 |                     2121 | REGR(+3.11%)  |
|-----------------------------------------------------------|-----------------------|--------------------------|---------------|
| Transpose_aie2_bf16_201                                   |                  2073 |                     2137 | REGR(+3.09%)  |
|-----------------------------------------------------------|-----------------------|--------------------------|---------------|
| Transpose_aie2_bf16_210                                   |                  2076 |                     2140 | REGR(+3.08%)  |
|-----------------------------------------------------------|-----------------------|--------------------------|---------------|
| Transpose_aie2_bf16_120_pad                               |                  1950 |                     2010 | REGR(+3.08%)  |
|-----------------------------------------------------------|-----------------------|--------------------------|---------------|
| Transpose_aie2_bf16_201_pad                               |                  1957 |                     2017 | REGR(+3.07%)  |
|-----------------------------------------------------------|-----------------------|--------------------------|---------------|
| LeakyRelu_int8_0                                          |                   805 |                      829 | REGR(+2.98%)  |
|-----------------------------------------------------------|-----------------------|--------------------------|---------------|
| Transpose_aie2_bf16_021_pad                               |                  2518 |                     2582 | REGR(+2.54%)  |
|-----------------------------------------------------------|-----------------------|--------------------------|---------------|
| Elu_aie2_bf16_0                                           |                  2610 |                     2664 | REGR(+2.07%)  |
|-----------------------------------------------------------|-----------------------|--------------------------|---------------|
| Mod_aie2_int8_0                                           |                  3854 |                     3919 | REGR(+1.69%)  |
|-----------------------------------------------------------|-----------------------|--------------------------|---------------|
| Mod_aie2_int8_1                                           |                  3854 |                     3919 | REGR(+1.69%)  |
|-----------------------------------------------------------|-----------------------|--------------------------|---------------|
| Conv2D_bf16_1                                             |                202358 |                   204878 | REGR(+1.25%)  |
|-----------------------------------------------------------|-----------------------|--------------------------|---------------|
| Conv2D_bf16_OW8_0                                         |                202367 |                   204887 | REGR(+1.25%)  |
|-----------------------------------------------------------|-----------------------|--------------------------|---------------|
| Conv2D_bf16_OW8_1                                         |                101891 |                   103151 | REGR(+1.24%)  |
|-----------------------------------------------------------|-----------------------|--------------------------|---------------|
| DivAttributeBroadcasting_aie2_bf16_0                      |                  6039 |                     6103 | REGR(+1.06%)  |
|-----------------------------------------------------------|-----------------------|--------------------------|---------------|
| ElemDiv_aie2_1                                            |                  1549 |                     1565 | REGR(+1.03%)  |
|-----------------------------------------------------------|-----------------------|--------------------------|---------------|
| Conv2D_bf16_0                                             |                103046 |                   104070 | REGR(+0.99%)  |
|-----------------------------------------------------------|-----------------------|--------------------------|---------------|
| InstanceNormPart1_aie2_int8_0                             |                  8810 |                     8798 | IMPR(-0.14%)  |
|-----------------------------------------------------------|-----------------------|--------------------------|---------------|
| Erf_aie2_bf16_0                                           |                  1994 |                     1991 | IMPR(-0.15%)  |
|-----------------------------------------------------------|-----------------------|--------------------------|---------------|
| SoftMaxRowMajor_int8_0                                    |                 13259 |                    13235 | IMPR(-0.18%)  |
|-----------------------------------------------------------|-----------------------|--------------------------|---------------|
| CompareOpsAttributeBroadcasting_bf16_1                    |                  1201 |                     1196 | IMPR(-0.42%)  |
|-----------------------------------------------------------|-----------------------|--------------------------|---------------|
| CompareOpsAttributeBroadcasting_bf16                      |                  1195 |                     1190 | IMPR(-0.42%)  |
|-----------------------------------------------------------|-----------------------|--------------------------|---------------|
| CompareOps_K_EQ_GE_GT_LE_LT_CMP_EQ_bfloat16               |                  1179 |                     1174 | IMPR(-0.42%)  |
|-----------------------------------------------------------|-----------------------|--------------------------|---------------|
| CompareOpsBroadcasting_K_EQ_GE_GT_LE_LT_CMP_GE_bfloat16   |                  1174 |                     1169 | IMPR(-0.43%)  |
|-----------------------------------------------------------|-----------------------|--------------------------|---------------|
| CompareOps_K_EQ_GE_GT_LE_LT_CMP_GE_bfloat16               |                  1173 |                     1168 | IMPR(-0.43%)  |
|-----------------------------------------------------------|-----------------------|--------------------------|---------------|
| CompareOpsBroadcasting_K_EQ_GE_GT_LE_LT_CMP_GE_bfloat16_1 |                  1168 |                     1163 | IMPR(-0.43%)  |
|-----------------------------------------------------------|-----------------------|--------------------------|---------------|
| Rsqrt_bf16_0                                              |                  1711 |                     1699 | IMPR(-0.70%)  |
|-----------------------------------------------------------|-----------------------|--------------------------|---------------|
| Sqrt_int8_0                                               |                 17231 |                    17106 | IMPR(-0.73%)  |
|-----------------------------------------------------------|-----------------------|--------------------------|---------------|
| Sqrt_int8_1                                               |                 17231 |                    17106 | IMPR(-0.73%)  |
|-----------------------------------------------------------|-----------------------|--------------------------|---------------|
| Cumsum_AIE2_bfloat16                                      |                 12291 |                    12195 | IMPR(-0.78%)  |
|-----------------------------------------------------------|-----------------------|--------------------------|---------------|
| Conv2DA16W8_1                                             |                  2218 |                     2194 | IMPR(-1.08%)  |
|-----------------------------------------------------------|-----------------------|--------------------------|---------------|
| Cast_int8_0                                               |                   461 |                      456 | IMPR(-1.08%)  |
|-----------------------------------------------------------|-----------------------|--------------------------|---------------|
| Cast_int8_1                                               |                   461 |                      456 | IMPR(-1.08%)  |
|-----------------------------------------------------------|-----------------------|--------------------------|---------------|
| Cast_uint8_0                                              |                   461 |                      456 | IMPR(-1.08%)  |
|-----------------------------------------------------------|-----------------------|--------------------------|---------------|
| Cast_uint8_1                                              |                   461 |                      456 | IMPR(-1.08%)  |
|-----------------------------------------------------------|-----------------------|--------------------------|---------------|
| ReduceL2_aie2_int8_1                                      |                 17312 |                    17104 | IMPR(-1.20%)  |
|-----------------------------------------------------------|-----------------------|--------------------------|---------------|
| ReduceL2_aie2_int8_0                                      |                 38764 |                    38296 | IMPR(-1.21%)  |
|-----------------------------------------------------------|-----------------------|--------------------------|---------------|
| Conv2D_bf16xbfp16_0                                       |                  3708 |                     3660 | IMPR(-1.29%)  |
|-----------------------------------------------------------|-----------------------|--------------------------|---------------|
| ReduceL2C8Axis_1_aie2_int8                                |                 17947 |                    17655 | IMPR(-1.63%)  |
|-----------------------------------------------------------|-----------------------|--------------------------|---------------|
| Sin_aie2_bf16                                             |                  1583 |                     1556 | IMPR(-1.71%)  |
|-----------------------------------------------------------|-----------------------|--------------------------|---------------|
| Conv2DA16W8_0                                             |                 12332 |                    12092 | IMPR(-1.95%)  |
|-----------------------------------------------------------|-----------------------|--------------------------|---------------|
| SigmoidTemplated_int8_0                                   |                 11209 |                    10953 | IMPR(-2.28%)  |
|-----------------------------------------------------------|-----------------------|--------------------------|---------------|
| SigmoidTemplated_int8_1                                   |                 11209 |                    10953 | IMPR(-2.28%)  |
|-----------------------------------------------------------|-----------------------|--------------------------|---------------|
| BatchNorm1d_aie2p_int8                                    |                   217 |                      211 | IMPR(-2.76%)  |
|-----------------------------------------------------------|-----------------------|--------------------------|---------------|
| InstanceNormPart2_aie2_int8_0                             |                 13587 |                    13191 | IMPR(-2.91%)  |
|-----------------------------------------------------------|-----------------------|--------------------------|---------------|
| HardswishAsHardsigmoid_aie2_0                             |                  1757 |                     1705 | IMPR(-2.96%)  |
|-----------------------------------------------------------|-----------------------|--------------------------|---------------|
| Hardswish_aie2_0                                          |                  1757 |                     1705 | IMPR(-2.96%)  |
|-----------------------------------------------------------|-----------------------|--------------------------|---------------|
| BatchNorm2D_aie2p_int8                                    |                   303 |                      291 | IMPR(-3.96%)  |
|-----------------------------------------------------------|-----------------------|--------------------------|---------------|
| ThresholdedRelu_aie2_int8                                 |                   778 |                      747 | IMPR(-3.98%)  |
|-----------------------------------------------------------|-----------------------|--------------------------|---------------|
| ReduceL2_aie2_bf16_1                                      |                  7100 |                     6776 | IMPR(-4.56%)  |
|-----------------------------------------------------------|-----------------------|--------------------------|---------------|
| DegroupG8_aie2_int8_1                                     |                   484 |                      453 | IMPR(-6.40%)  |
|-----------------------------------------------------------|-----------------------|--------------------------|---------------|
| InstanceNormPart2_aie2_bf16_0                             |                  7549 |                     7053 | IMPR(-6.57%)  |
|-----------------------------------------------------------|-----------------------|--------------------------|---------------|
| DegroupG8_aie2_bf16_1                                     |                   838 |                      775 | IMPR(-7.52%)  |
|-----------------------------------------------------------|-----------------------|--------------------------|---------------|
| MulA16_0                                                  |                  3179 |                     2925 | IMPR(-7.99%)  |
|-----------------------------------------------------------|-----------------------|--------------------------|---------------|
| SoftSign_aie2_int8                                        |                  1013 |                      929 | IMPR(-8.29%)  |
|-----------------------------------------------------------|-----------------------|--------------------------|---------------|
| Ceil_bfloat16                                             |                  2187 |                     2000 | IMPR(-8.55%)  |
|-----------------------------------------------------------|-----------------------|--------------------------|---------------|
| Sigmoid_int8_1                                            |                   113 |                       82 | IMPR(-27.43%) |
|-----------------------------------------------------------|-----------------------|--------------------------|---------------|
| HardSigmoidTemplated_int8_0                               |                   298 |                      207 | IMPR(-30.54%) |
|-----------------------------------------------------------|-----------------------|--------------------------|---------------|
| Average diff                                              | +0.00%                | -0.32%                   | -0.32%        |
|-----------------------------------------------------------|-----------------------|--------------------------|---------------|
| Diff stdev                                                |                  0.00 |                     2.85 |          2.85 |
|-----------------------------------------------------------|-----------------------|--------------------------|---------------|
| Quantile #1                                               | +0.00%                | +0.00%                   | +0.00%        |
|-----------------------------------------------------------|-----------------------|--------------------------|---------------|
| Quantile #2                                               | +0.00%                | +0.00%                   | +0.00%        |
|-----------------------------------------------------------|-----------------------|--------------------------|---------------|
| Quantile #3                                               | +0.00%                | +0.00%                   | +0.00%        |
|-----------------------------------------------------------|-----------------------|--------------------------|---------------|

```

martien-de-jong (approved):
LGTM

PR #695, opened 2025-11-03
[AIE] widen RelocationPatch to 256 bits.

martien-de-jong:
This is necessary for relocation in packets > 128 bits. It also fixes an UB as found by @mgehre-amd.
Note that RelocationPatch is a rename of the original Patch, which was, albeit local to the file, a bit too generic  a name.


abnikant (approved):
LGTM!

PR #696, opened 2025-11-04
[AIEX][NFC] Allow passing separate opcodes for loop setup start and end in ZOLSupport

konstantinschwarz:
Also make adding the physical register def operand to the instruction optional.

### DIFF
--- a/llvm/lib/Target/AIE/AIE2InstrInfo.cpp
+++ b/llvm/lib/Target/AIE/AIE2InstrInfo.cpp
@@ -1446,7 +1446,8 @@ AIE2InstrInfo::getZOLSupport() const {
   Result.LoopStartOpcode = AIE2::LoopStart;
   Result.LoopEndOpcode = AIE2::PseudoLoopEnd;
   Result.SetLoopCountOpcode = AIE2::ADD_NC;
-  Result.SetAddressOpcode = AIE2::MOVXM_lng_cg;
+  Result.SetLoopStartOpcode = AIE2::MOVXM_lng_cg;
+  Result.SetLoopEndOpcode = AIE2::MOVXM_lng_cg;
   // We need at 112 bytes distance from the loop setup to the loop end label,
   // which requires 7 bundles of 16 bytes.
   Result.LoopSetupDistance = 7;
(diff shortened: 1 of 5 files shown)


llvm/lib/Target/AIE/AIEBaseHardwareLoops.cpp
```diff
@@ -413,18 +413,18 @@ void AIEBaseHardwareLoops::expandLoopStart(LowOverheadLoop &LoLoop) {
...
-      .addMBB(LoLoop.LoopEnd->getOperand(1).getMBB());
-  MachineInstrBuilder MIB;
-  MIB = BuildMI(*MBB, Start, Start->getDebugLoc(),
-                TII->get(LoweringData->SetAddressOpcode),
-                LoweringData->LERegister)
-            .addSym(LoLoop.LoopEnd->getOperand(0).getMCSymbol());
-  MachineInstr *MI = MIB.getInstr();
-  MI->getOperand(1).ChangeToMCSymbol(
-      LoLoop.LoopEnd->getOperand(0).getMCSymbol());
+  bool AddPhysRegOperands = LoweringData->AddPhysRegOperands;
```
andcarminati: nit: `const bool`

llvm/lib/Target/AIE/AIE2InstrInfo.cpp
```diff
@@ -1446,13 +1446,15 @@ AIE2InstrInfo::getZOLSupport() const {
...
-  Result.SetAddressOpcode = AIE2::MOVXM_lng_cg;
+  Result.SetLoopStartOpcode = AIE2::MOVXM_lng_cg;
+  Result.SetLoopEndOpcode = AIE2::MOVXM_lng_cg;
   // We need at 112 bytes distance from the loop setup to the loop end label,
   // which requires 7 bundles of 16 bytes.
   Result.LoopSetupDistance = 7;
   Result.LCRegister = AIE2::LC;
   Result.LSRegister = AIE2::LS;
   Result.LERegister = AIE2::LE;
+  Result.AddPhysRegOperands = true;
```
mludevid: Nit: Could we rename this to something like `AddExplicitRegOperands`? Alternatively, we could also make the two previous declarations optional and call them `AddLSRegister` and `AddLERegister`.

andcarminati: Alternatively, we could have LS and LE as `std::optional<Register>`. If we have them, we add, because we don't have other uses apart of current one.

konstantinschwarz: > Alternatively, we could have LS and LE as `std::optional<Register>`. If we have them, we add, because we don't have other uses apart of current one.

I like this idea. Eventually I could also see us adding a builder method to the ZOLSupport structure. For now, I think its fine to add those optionals

andcarminati (approved):
LGTM.

PR #699, opened 2025-11-07
SWP-aware WAWRegRewriter.

martien-de-jong:
This adds an alternative to the latency-aware heuristic in WAWRegisterRewriter and selects between the two based on LoopClass.
Focus kernel results here: (PMSize is flat)
image


llvm/lib/Target/AIE/AIEBaseSubtarget.cpp
```diff
@@ -195,6 +195,7 @@ bool updateSuccLatency(SDep &SuccEdge, SUnit &PredSU, int Latency) {
 // The initial graph will have ordering edges induced by hasSideEffects of the
 // locks/DONE.
 class LockDelays : public ScheduleDAGMutation {
+  bool ExactLatencies = true;
```
andcarminati, line 198: If we do this:
```
 class LockDelays : public ScheduleDAGMutation {
-  bool ExactLatencies = true;
+  bool ExactLatencies;
     const auto *TII = static_cast<const AIEBaseInstrInfo *>(DAG->TII);
     const int CoreStallCycle = TII->getCoreStallCycleAfterLock();
@@ -243,7 +243,7 @@ class LockDelays : public ScheduleDAGMutation {
   }
 
 public:
-  LockDelays(bool ExactLatencies) : ExactLatencies(ExactLatencies) {};
+  LockDelays(bool ExactLatencies = true) : ExactLatencies(ExactLatencies) {};
 };
```

We don't need any change in the current instantiation (getPostRAMutationsImpl) of the mutators.

llvm/lib/Target/AIE/AIEBaseSubtarget.cpp
```diff
@@ -677,16 +688,22 @@ class MemoryEdges : public ScheduleDAGMutation {
         // Get the correct latency from the Sched model.
         std::optional<int> MemLat = TII->getMemoryLatency(
             SrcMI.getDesc().getSchedClass(), MI.getDesc().getSchedClass());
-        if (!MemLat.has_value()) {
+        int Latency = 1;
```
andcarminati, line 691: CHECK: if we don't need exact latencies, and we don't have them, we use a very optimistic value instead.

llvm/lib/Target/AIE/AIEWawRegRewriter.cpp
```diff
@@ -366,6 +383,59 @@ RoundRobin AIEWawRegRewriter::computeLRURegisters(
...
+  std::map<const MachineInstr *, int> ModuloCycle;
+  int MaxDepth = 0;
+  for (auto &SU : DDG.SUnits) {
+    int D = SU.getDepth();
+    ModuloCycle.emplace(SU.getInstr(), D % MinII);
+    MaxDepth = std::max(MaxDepth, D);
+    LLVM_DEBUG(dbgs() << format("%4d D=%4d: ", SU.NodeNum, D)
+                      << *SU.getInstr());
+  }
+
```
andcarminati, line 466: Should we have `(void) MaxDepth;` to avoid warnings in strict release build mode?

llvm/lib/Target/AIE/AIEBaseSubtarget.cpp
martien-de-jong: I guess that's true. But I have bad feelings about default parameters in constructors. It restricts future constructors which would be ambiguous.

llvm/lib/Target/AIE/AIEBaseSubtarget.cpp
martien-de-jong: Not very optimistic. For a RAW ST -> LD pair, the latency is actually 1, since the LOAD reads memory late.. It's also the default that was assumed before Gaetan made it defined by a target hook.

llvm/lib/Target/AIE/AIEWawRegRewriter.cpp
martien-de-jong: Good catch. It's a relic from my first attempt.

llvm/lib/Target/AIE/AIEPostPipeliner.cpp
```diff
@@ -604,17 +606,31 @@ void dumpGraph(const ScheduleInfo &Info, ScheduleDAGInstrs *DAG) {
   dbgs() << "}\n";
 }
 
+int slotLetter(const SlotCounts &Slots) {
```
andcarminati: Should we use char here as well? Just for consistency, because we use:

```
char S = slotLetter(Info[K].Slots);
```

andcarminati:
nit: check this commit message: `[AIE][NFC] Use equivalant LoopUtils function`.

llvm/lib/Target/AIE/AIEPostPipeliner.cpp
martien-de-jong: Can do.

llvm/lib/Target/AIE/AIEWawRegRewriter.cpp
```diff
@@ -437,6 +437,16 @@ void AIEWawRegRewriter::sortSWPAware(OriginalAllocation &Candidates,
 }
 
 bool AIEWawRegRewriter::renameMBBPhysRegs(const MachineBasicBlock *MBB) {
+  // We do this mainly for the postpipeliner, and it will want to see ZOL
```
andcarminati: I guess we could have some verifier problems by accepting ZOL loops before MachineBlockPlacement, right? 

It is better to choose one direction here:
* Implement another hook, like `isPreLayoutZOLBody`.
* Extend isZOLBody and target machine verifier.
* Clearly state in the comment why we are not relying isZOLBody here.

andcarminati: Also, sometimes we can see some unexpected gains related to PreSWP + WAWRewriter. Are we preventing gains in some pipelined decrement/branch loop here? Could we guard this behavior with a command line option, to be able to run additional tests in the future?

llvm/lib/Target/AIE/AIEWawRegRewriter.cpp
```diff
@@ -66,14 +66,18 @@ static cl::opt<unsigned>
                                 "considered for WAW rewriting"),
                        cl::init(3));
 
-static cl::opt<bool>
-    LatencyAware("aie-realloc-latencyaware", cl::Hidden, cl::init(true),
+static cl::opt<int>
```
andcarminati: I am trying to figure out the rationale behind this change...

llvm/lib/Target/AIE/AIEWawRegRewriter.cpp
```diff
@@ -89,6 +93,63 @@ using OriginalAllocation =
...
+bool runLatencyAware(int LoopClass) {
+  switch (LatencyAware) {
+  case 0:
+    return false;
+  case 1:
+    return true;
+  default:
+    break;
+  }
+  switch (LoopClass) {
```
andcarminati, line 109: We have a symmetric relation with another switch below. Could we refactor? What about a data-driven approach?

llvm/lib/Target/AIE/AIEWawRegRewriter.cpp
martien-de-jong: Removed MaxDepth altogether

llvm/lib/Target/AIE/AIEWawRegRewriter.cpp
martien-de-jong: Hmm. Under these assumptions, we should perhaps not constrain it to single-block loops at all, because the only advantage we can have from PreSWP loops is the scheduler, and for the scheduler anything might work.
I don't think isPreLayoutZOLBody is very clear concept, it is so tied to the flow. 
If you have evidence of the combination of PreSWP, CountedLoops and WAWRegRewriter paying off, I would suggest to just remove the check. My main motivation was that I saw it kick in on loops with calls inside, which is concisely prevented by the ZOL check.

llvm/lib/Target/AIE/AIEWawRegRewriter.cpp
martien-de-jong: I'm overloading here, I agree. for tuning, I want to select false, true, or autoselect.

llvm/lib/Target/AIE/AIEWawRegRewriter.cpp
andcarminati: Or even `AIEBaseInstrInfo::isZOLBody(const MachineBasicBlock &MBB, bool OnlyCheckLastInstr = true)`

martien-de-jong: Ok, I found evidence in AvgPool. Removed the new check

llvm/lib/Target/AIE/AIEWawRegRewriter.cpp
martien-de-jong: I rewrote the option structure.

llvm/lib/Target/AIE/AIEWawRegRewriter.cpp
```diff
@@ -446,7 +494,18 @@ bool AIEWawRegRewriter::renameMBBPhysRegs(const MachineBasicBlock *MBB) {
   IndexedMap<const MachineInstr *, VirtReg2IndexFunctor> LastVRegDef =
       getLastVRegDef(*MBB);
 
-  std::set<MCRegister> HighLatencyRegs = getHighOutputLatencyRegs(MBB);
+  auto &NonConstMBB = *(const_cast<MachineBasicBlock *>(MBB));
+  AIE::SlotStatistics Statistics = AIE::computeSlotStatistics(NonConstMBB, TII);
```
andcarminati, line 495: I saw this, but I will not trigger any further discussions ;-)

llvm/lib/Target/AIE/AIEWawRegRewriter.cpp
```diff
@@ -523,9 +582,9 @@ bool AIEWawRegRewriter::renameMBBPhysRegs(const MachineBasicBlock *MBB) {
     }
   }
 
-  if (SWPAware) {
-    auto &NCMBB = *(const_cast<MachineBasicBlock *>(MBB));
-    sortSWPAware(Candidates, NCMBB);
+  int Bias = MinIIBias;
+  if (runSWPAware(Mode, LoopClass, Bias)) {
```
andcarminati: I feel that `runSWPAware` is not running the SWP aware. It just decides if we actually should run it or not. Could we address this in a different way, like:

```
if (auto Bias = getSWPAwareBias(Mode, LoopClass)) {
 ...
}
```

With an optional Bias.... ?

andcarminati (approved):
LGTM.

PR #700, opened 2025-11-07
[AutoBump] Merge with fixes of a58e774f (Jan 28, tosa.mul gets third operand) (1)

mgehre-amd:
- [x] https://github.com/Xilinx/torch-mlir/pull/633
- [x] https://github.com/Xilinx/onnx-mlir/pull/478
- [x] https://github.com/Xilinx/mlir-xten/pull/187

- **[mlir][tosa] Make TOSA MUL's Shift an Input (#121953)**
  The TOSA-v1.0 specification makes the shift attribute of the MUL
  (Hammard product) operator an input. Move the `shift` parameter of the
  MUL operator in the MILR TOSA dialect from an attribute to an input and
  update any lit tests appropriately.
  
  Expand the verifier of the `tosa::MulOp` operation to check the various
  constraints defined in the TOSA-v1.0 specification. Specifically, ensure
  that all input operands (excluding the optional shift) are of the same
  rank. This means that broadcasting tests which previously checked rank-0
  tensors would be broadcast are no longer valid and are removed.
  
  Signed-off-by: Jack Frankland <jack.frankland@arm.com>
  Co-authored-by: TatWai Chong <tatwai.chong@arm.com>

Also cherry-picked [llvm/llvm-project@20ae283](https://github.com/llvm/llvm-project/commit/20ae283d087224f6b82b7308054bd34a6764d926)

- **Solve merge conflicts and fixed tests after bump**
  Handle the upstream switch of tosa.mul to use an optional shift operand by allowing the constant folding path to accept and validate a constant zero shift tensor. Relax the shared binary folding helper to work with ops that gain extra operands so mul no longer asserts. Update the TOSA constant-folding test cases to spell out zero-shift operands, capture the new value numbering, and document these findings for future bump runs.

### DIFF
--- a/mlir/include/mlir/Dialect/Tosa/IR/TosaOpBase.td
+++ b/mlir/include/mlir/Dialect/Tosa/IR/TosaOpBase.td
@@ -239,7 +239,9 @@ class Tosa_ElementwiseOp<string mnemonic, list<Trait> traits = []> :
     Tosa_Op<mnemonic, !listconcat(traits, [
               DeclareOpInterfaceMethods<InferShapedTypeOpInterface,
                                         ["inferReturnTypeComponents"]>,
+              ResultsBroadcastableShape,
               TosaElementwiseOperator,
+              SameOperandsAndResultRank,
               Pure])> {
   let assemblyFormat =
       "operands attr-dict `:` functional-type(operands, results)";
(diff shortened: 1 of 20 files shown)


jorickert:
I think we should cherry-pick https://github.com/llvm/llvm-project/commit/20ae283d087224f6b82b7308054bd34a6764d926 in the same bump, otherwise we need to deal two times with changes to the mul shift

mgehre-amd:
> I think we should cherry-pick [llvm/llvm-project@20ae283](https://github.com/llvm/llvm-project/commit/20ae283d087224f6b82b7308054bd34a6764d926) in the same bump, otherwise we need to deal two times with changes to the mul shift

Done

jorickert (approved):

PR #704, opened 2025-11-10
[NFC][AIE] Streamline the format checks using Conflict bits.

martien-de-jong:
ConflictBits summarize conflicts between one slot and another, in particular between e.g. XM and X and M
Availability of a format with a particular slot occupation is summarized in a lookup table with the slot set as index. 
I have a hunch that the format check is actually not necessary anymore, but now that it is O(1), it doesn't matter much. The reasoning is that the 128 bit format can accommodate all combinations of real slots, and the conflict bits now catch the artificial composite slot constraints.


llvm/include/llvm/Target/CodeGenFormat.td
```diff
@@ -45,6 +45,10 @@ class InstSlot<string slotname, int Size, bit isDefault=false>  : Operand<OtherV
   // Namespace.
   bit isDefaultSlot = isDefault;
 
+  // Mark a slot as artificial. This will exclude it from any resource
+  // estimations.
+  bit Artificial = false;
```
martien-de-jong, line 50: This is used for e.g. nop slots that are just added to drive format selection, but don't occupy existing slots.

llvm/test/TableGen/CodeGenFormat/SlotLogicTests/slot_conflicts.td
```diff
@@ -0,0 +1,111 @@
...
+
+// This is a minimal check for the computations of the conflict bits.
+// We have the slots A. X, M and XM, which interfere like the
+// corresponding slots on typical AIE architectures.
+// We compute conflicts from the totality of formats, and we should find
+//  conflicts
+//  X -> {XM},
+//  M -> {XM},
+//  XM -> {X, M, XM}
+// and nothing more. In particular, ConflictBits of A should be equal to its SlotBits
```
martien-de-jong: I'll extend this test a bit.

llvm/test/TableGen/CodeGenFormat/SlotLogicTests/slot_conflicts.td
```diff
@@ -0,0 +1,123 @@
...
+//  X -> {X, XM},
+//  M -> {M, XM},
+//  XM -> {X, M, XM}
+//  A -> {A}
+// and nothing more.
+
+// CHECK-LABEL:   "X",
+// CHECK-NEXT:  8,
+// CHECK-NEXT:  1,
+// CHECK-NEXT:  5,
```
martien-de-jong, line 25: size=8, slot=0001, conflicts=0101

llvm/lib/Target/AIE/AIEBundle.h
```diff
@@ -91,10 +91,15 @@ template <class I> class Bundle {
...
-           FormatInterface->getPacketFormats().getFormat(OccupiedSlots |
-                                                         NewSlots);
+
+    auto *SlotInfo = FormatInterface->getSlotInfo(Slot);
+    // ConflictBits is a fast predictor of missing formats
+    SlotBits ConflictBits = SlotInfo->getConflictSet();
+    if (OccupiedSlots & ConflictBits) {
+      return false;
+    }
+    SlotBits NewSlots = OccupiedSlots | SlotInfo->getSlotSet();
```
andcarminati: If we are using `ConflictBits`, why do we still need to check the format (`getFormat`)?

andcarminati: Assert? and return true?

andcarminati: Or even a comment to clarify.

llvm/lib/Target/AIE/MCTargetDesc/AIEMCFormats.h
```diff
@@ -93,18 +93,27 @@ class MCSlotInfo {
...
+  /// e.g. XM implies X and M
+  const SlotBits ConflictBits;
+  /// Opcode of the NOP instruction attached to the slot
   const unsigned NopOpc;
 
 public:
   constexpr MCSlotInfo(const char *SlotName, unsigned Size, SlotBits Bits,
                        unsigned NopOpc)
-      : SlotName(SlotName), Size(Size), SlotOccupancy(Bits), NopOpc(NopOpc) {}
+      : SlotName(SlotName), Size(Size), SlotOccupancy(Bits), ConflictBits(Bits),
```
andcarminati: We can remove this constructor.

llvm/utils/TableGen/CodeGenFormat.cpp
```diff
@@ -244,6 +245,46 @@ void CodeGenFormat::run(raw_ostream &o) {
...
+  for (TGInstrLayout &Format : InstFormats) {
+    if (Format.isPacketFormat()) {
+      Format.computeSlotSet();
+    }
+  }
+
+  // The universe of all slots
+  uint64_t AllSlots = 0;
+  for (const auto &[Rec, Slot] : Slots) {
+    if (Slot.isDefaultSlot() || Slot.isArtificial()) {
```
andcarminati, line 298: We can iterate over the full pairs here and just use the second element, because `Rec` is not used. Or even better:

```
for (const auto &[_, Slot] : Slots) {
 ...
}
```

llvm/utils/TableGen/CodeGenFormat.cpp
```diff
@@ -244,6 +245,46 @@ void CodeGenFormat::run(raw_ostream &o) {
...
+  for (unsigned S = 0; S < Slots.size(); S++) {
+    // Effectively we check whether a slot cannot be combined with another
+    // slot in any format in which the first is accommodated. We start out
+    // with all slots excluded, and eliminate the exclusion when a format is
+    // encountered that allows both ThisSlot and the exclusion.
+    auto &Slot = Slots.getSlot(S).second;
+    const uint64_t ThisSlot = Slot.getSlotBits();
+    uint64_t Excluded = AllSlots;
+    for (TGInstrLayout &Format : InstFormats) {
+      if (Format.isPacketFormat()) {
```
andcarminati: nit:
```
if (!Format.isPacketFormat())
    continue;
```

(reduce indentation)

llvm/utils/TableGen/CodeGenFormat.cpp
```diff
@@ -244,6 +245,46 @@ void CodeGenFormat::run(raw_ostream &o) {
...
+        uint64_t Slots = Format.getSlotSet();
```
andcarminati: Is Slots redeclared multiple times?

martien-de-jong: Right, it hides the parameter. I'll rename.

llvm/lib/Target/AIE/AIEBundle.h
```diff
@@ -91,10 +91,15 @@ template <class I> class Bundle {
...
     assert(Slot != MCSlotKind());
-    SlotBits NewSlots = FormatInterface->getSlotInfo(Slot)->getSlotSet();
-    return (OccupiedSlots & NewSlots) == 0 &&
-           FormatInterface->getPacketFormats().getFormat(OccupiedSlots |
-                                                         NewSlots);
+
+    auto *SlotInfo = FormatInterface->getSlotInfo(Slot);
+    // ConflictBits is a fast predictor of missing formats
+    SlotBits ConflictBits = SlotInfo->getConflictSet();
+    if (OccupiedSlots & ConflictBits) {
```
F-Stuckmann, line 98: wouldn't the OccupiedSlots be the ConflictBits of a Bundle?

llvm/lib/Target/AIE/MCTargetDesc/AIEMCFormats.h
```diff
@@ -93,18 +93,23 @@ class MCSlotInfo {
   const char *SlotName;
   /// Size of the slot (in bits)
   const unsigned Size;
-  /// Bitset representing the occupancy of the slot
+  /// Bitset representing the occupancy of the slots
   const SlotBits SlotOccupancy;
-  /// Opcode of the NOP inst. attached to the slot
+  /// The closure of SlotOccupancy with the computed exclusions,
+  /// e.g. XM implies X and M
+  const SlotBits ConflictBits;
```
F-Stuckmann, line 100: why do we need a new `SlotBits`? 
Could we not have a unified `SlotOccupancy` where `XM` is properly mapped to `X` and `M` occupation?

llvm/lib/Target/AIE/AIEHazardRecognizer.cpp
```diff
@@ -61,7 +61,8 @@ void FuncUnitWrapper::setFormatInterface(const AIEBaseMCFormats *Formats) {
 
 bool FuncUnitWrapper::operator==(const FuncUnitWrapper &Other) const {
   return Required == Other.Required && Reserved == Other.Reserved &&
-         Slots == Other.Slots && MemoryBanks == Other.MemoryBanks &&
+         Slots == Other.Slots && Conflicts == Other.Conflicts &&
```
F-Stuckmann, line 64: why do we have to check slots and conflicts separately?

F-Stuckmann:
Overall I do not understand why we still carry around the regular `Slots`, when `ConflictBits` is the more accurate representation? 
Could we not reduce this to a single `OccupiedSlotbits`?

llvm/unittests/Target/AIE/HazardRecognizerTest.cpp
```diff
@@ -135,7 +135,7 @@ class MockHR : public AIEHazardRecognizer {
   bool hazard(unsigned SchedClass, int Delta, SlotBits SlotSet = 0,
               MemoryBankBits MemoryBanks = 0, MemoryObjectsBits ObjectsBits = 0,
               SmallVector<int, 2> MemoryAccessCycles = {}) {
-    return checkConflict(MockScoreboard, &Itins, SchedClass, SlotSet,
+    return checkConflict(MockScoreboard, &Itins, SchedClass, SlotSet, SlotSet,
```
F-Stuckmann, line 138: shouldn't we be passing `conflictbits` here?

llvm/lib/Target/AIE/AIEHazardRecognizer.cpp
```diff
@@ -139,7 +139,7 @@ bool FuncUnitWrapper::conflict(const FuncUnitWrapper &Other) const {
   // This allows representing a blocked cycle (Slots = ~0) without knowing
   // the slot and format details.
   return Slots && Other.Slots &&
-         !FormatInterface->getPacketFormats().getFormat(Slots | Other.Slots);
+         !FormatInterface->isFormatAvailable(Slots | Other.Slots);
```
F-Stuckmann, line 142: why aren't we checking conflict bits here?

llvm/unittests/Target/AIE/HazardRecognizerTest.cpp
martien-de-jong: This is a small existing mockup where they coincide.

llvm/lib/Target/AIE/AIEBundle.h
martien-de-jong: I think all your comments are covered by this one answer. We keep Slots since they represent the true base class from which the instructions derive. Each instruction has exactly one slot. These are also the slots that are listed in the ISA, in the formats, they define the operands of the format instructions, etc. 
The conflicts have no such direct interpretation. I use them for which they were designed, which is as an easy way to detect conflicts given the regular slot occupation bits without scanning the format table.
They end up in FuncUnitWrapper since that is at the heart of the conflict query interface. I cannot enter them in the ScoreBoard, since both X and M would set the XM bit, which would not allow X to be in the same bundle as M.

llvm/lib/Target/AIE/AIEHazardRecognizer.cpp
martien-de-jong: we are, at line 131, returning early if we find a conflict.

andcarminati (approved):
LGTM.

llvm/utils/TableGen/CodeGenFormat.h
```diff
@@ -349,12 +367,21 @@ class TGTargetSlot {
...
   bool isDefaultSlot() const { return IsDefaultSlot; }
+  bool isArtificial() const { return Artificial; }
+  void setConflictBits(uint64_t Bits) { ConflictBits = Bits; }
 
 private:
-  void setNumSlot(int SlotID) { NumSlot = SlotID; }
+  void setNumSlot(int SlotID) {
+    NumSlot = SlotID;
+    const uint64_t Base = 1;
+    SlotBits = Base << SlotID;
```
konstantinschwarz, line 382: I'm getting an undefined behavior sanitizer error when trying to build this commit:
```
llvm-public/build/bin/llvm-tblgen -gen-instr-format -I llvm-public/llvm/lib/Target/AIE -I llvm-public/build/include -I llvm-public/llvm/include -I llvm-public/llvm/lib/Target llvm-public/llvm/lib/Target/AIE/AIE2P.td --write-if-changed -o lib/Target/AIE/AIE2PGenFormats.inc -d lib/Target/AIE/AIE2PGenFormats.inc.d
llvm-public/llvm/utils/TableGen/CodeGenFormat.h:382:21: runtime error: shift exponent -1 is negative
SUMMARY: UndefinedBehaviorSanitizer: undefined-behavior llvm-public/llvm/utils/TableGen/CodeGenFormat.h:382:21
```

konstantinschwarz (commented):
Can you look into the sanitizer failure?

PR #705, opened 2025-11-11
[AIE2P] Enable encoding tests run lines

mludevid:
We had some run lines disabled in the encoding tests in AIE2P that are actually already passing. This PR simply activates them.


F-Stuckmann:
Do you know why they weren't enabled by default?

mludevid:
> Do you know why they weren't enabled by default?

I suppose at the time when we first merged them we were missing features in the asm printer/parser? There are a few test lines I could still not enable, maybe someone could look into adding those missing features.

F-Stuckmann (approved):
LGTM

PR #707, opened 2025-11-12
[AIE] Add a loopclass to pinpoint a postpipeliner candidate

martien-de-jong:
This improves SiLU by some 5%. We explicitly select the loopclass to reject it in prepipeliner.
image


llvm/lib/Target/AIE/AIELoopClass.cpp
```diff
@@ -78,6 +78,10 @@ static const KernelFeatures Kernels[] = {
     {45, {{900, 300, 0, 2520, 0, 0, 3360}, {45, 45, 0, 45, 45}}},
     {46, {{0, 0, 0, 0, 2160, 0, 120, 1080}, {0, 420, 420}}},
     {47, {{0, 0, 0, 0, 360, 0, 240, 360}, {0, 60, 60}}},
+    // These are pre-regalloc
+    {1000, {{0, 0, 0, 0, 480, 0, 480, 480}, {0, 120, 120}}},
```
F-Stuckmann, line 82: I wonder if we could build the final Kernel classification as a combination of ArrayRefs.
If we exceed 1k kernels for WAW rewriting, the updating will be a mess. (especially if this is at some point automated with a script, where we do not have to carefully extract the individual classes) 

In the case of separate ArrayRefs, we could simply check if the kernel key is in the corresponding ArrayRef.

llvm/lib/Target/AIE/AIEBasePipelinerLoopInfo.cpp
```diff
@@ -751,6 +753,19 @@ bool ZeroOverheadLoop::preferPostPipeliner(SMSchedule &SMS) {
     return false;
   }
 
+  AIE::SlotStatistics Stats = AIE::computeSlotStatistics(*LoopBlock, &TII);
+  int LoopClass = classifyLoop(Stats);
+  LLVM_DEBUG(dbgs() << "Stats="; Stats.dumpShort();
+             dbgs() << format("\nLoopClass=%d", LoopClass));
+  switch (LoopClass) {
+  case 1000:
```
F-Stuckmann: I would prefer to not have magic numbers across the codebase that kill us when we change the Loop Statistic dataset (see arrayref comment)

F-Stuckmann: A hotfix would be a #define RejectPrepipeline 1000

llvm/lib/Target/AIE/AIEBasePipelinerLoopInfo.cpp
```diff
@@ -751,6 +753,19 @@ bool ZeroOverheadLoop::preferPostPipeliner(SMSchedule &SMS) {
     return false;
   }
 
+  AIE::SlotStatistics Stats = AIE::computeSlotStatistics(*LoopBlock, &TII);
+  int LoopClass = classifyLoop(Stats);
```
F-Stuckmann: I wonder if we should move all the high level methods into SlotStatistics (`rejectPrepipeling` but also the WAW methods `getMinIIBias` `selectMode`). 
This fragmentation and magic numbers across the code base seems to be get us into a maintainance hell

llvm/lib/Target/AIE/AIELoopClass.cpp
martien-de-jong: Discussed offline. These loopclasses sample the total 'slotstatistics' space. The main reason to leave a gap is to avoid collisions with other additions from parallel developments. They are also not dedicated to a particular algorithm, and may generalize in an overlapping way to other algorithms. Ultimately we use the distance to each of those sampling points to make decisions, and they just define particular neighbourhood in that space.

llvm/lib/Target/AIE/AIEBasePipelinerLoopInfo.cpp
martien-de-jong: Also discussed offline. RejectPrePipeline could be a set of loopclasses. Other applications would not necessarily have this boolean nature, but want to compute a heuristic parameter based on the loop class. It may also look at the first N closest loopclasses to make a decision . In the long term, we will be tuning all these algorithm to jointly yield the best performance. That trained network will interface in some way with each individual algorithm. The challenge will be to separate the concerns of the algorithm from those of the network.

llvm/lib/Target/AIE/AIEBasePipelinerLoopInfo.cpp
martien-de-jong: Maintenance will not be less of a hell, since SlotStatistics is just a specific set of numbers. Whether we export the numbers or import the methods doesn't make much difference.  If I tune an algorithm, I think I'd like its source file to change. If this requires a new data point in the loopclass space, I think it's natural to change the LoopClass file, which will just add new lines, never change existing ones.

F-Stuckmann: maybe have a comment to clarify that once created loopclass statistics shouldn't be changed.

llvm/lib/Target/AIE/AIELoopClass.cpp
```diff
@@ -78,6 +78,10 @@ static const KernelFeatures Kernels[] = {
...
+    {1001, {{0, 0, 0, 0, 360, 0, 480, 720}, {0, 60, 60}}},
```
martien-de-jong, line 83: Needed to add this, since it caused a false positive match due to small distance

llvm/lib/Target/AIE/AIEBasePipelinerLoopInfo.cpp
martien-de-jong: There is one in AIELoopClass.cpp

F-Stuckmann (approved):

andcarminati (approved):

PR #709, opened 2025-11-12
MLIR: Add ReplaceOpAction into DialectConversion and PatternMatch

mgehre-amd:
Add action to enable listener for `RewriterBase::replaceOp`, `RewriterBase::replaceAllUsesWith` and friends.

### DIFF
--- a/mlir/include/mlir/IR/PatternMatch.h
+++ b/mlir/include/mlir/IR/PatternMatch.h
@@ -13,6 +13,7 @@
 #include "mlir/IR/BuiltinOps.h"
 #include "llvm/ADT/FunctionExtras.h"
 #include "llvm/Support/TypeName.h"
+#include <mlir/IR/PatternMatchAction.h>
 #include <optional>
 
 using llvm::SmallPtrSetImpl;
@@ -652,6 +653,9 @@ public:
   /// Find uses of `from` and replace them with `to`. Also notify the listener
   /// about every in-place op modification (for every use that was replaced).
   void replaceAllUsesWith(Value from, Value to) {
+    if (auto *fromOp = from.getDefiningOp())
+      getContext()->executeAction<ReplaceOpAction>(
+          []() {}, ArrayRef<IRUnit>{fromOp}, to);
     for (OpOperand &operand : llvm::make_early_inc_range(from.getUses())) {
       Operation *op = operand.getOwner();
       modifyOpInPlace(op, [&]() { operand.set(to); });
--- a/mlir/include/mlir/IR/PatternMatch.h
+++ b/mlir/include/mlir/IR/PatternMatch.h
@@ -11,9 +11,9 @@
 
 #include "mlir/IR/Builders.h"
 #include "mlir/IR/BuiltinOps.h"
+#include "mlir/IR/PatternMatchAction.h"
 #include "llvm/ADT/FunctionExtras.h"
 #include "llvm/Support/TypeName.h"
-#include <mlir/IR/PatternMatchAction.h>
 #include <optional>
 
 using llvm::SmallPtrSetImpl;
@@ -655,7 +655,7 @@ public:
   void replaceAllUsesWith(Value from, Value to) {
     if (auto *fromOp = from.getDefiningOp())
       getContext()->executeAction<ReplaceOpAction>(
-          []() {}, ArrayRef<IRUnit>{fromOp}, to);
+          /*actionFn=*/[]() {}, ArrayRef<IRUnit>{fromOp}, to);
     for (OpOperand &operand : llvm::make_early_inc_range(from.getUses())) {
       Operation *op = operand.getOwner();
       modifyOpInPlace(op, [&]() { operand.set(to); });
--- a/mlir/include/mlir/IR/PatternMatchAction.h
+++ b/mlir/include/mlir/IR/PatternMatchAction.h
@@ -1,0 +1,20 @@
+#ifndef MLIR_IR_PATTERNMATCHACTION_H
+#define MLIR_IR_PATTERNMATCHACTION_H
+
+#include "mlir/IR/Action.h"
+
+namespace mlir {
+struct ReplaceOpAction : public tracing::ActionImpl<ReplaceOpAction> {
+  using Base = tracing::ActionImpl<ReplaceOpAction>;
+  ReplaceOpAction(ArrayRef<IRUnit> irUnits, ValueRange replacement);
+  static constexpr StringLiteral tag = "op-replacement";
+  void print(raw_ostream &os) const override;
+
+  Operation *getOp() const;
+
+public:
+  ValueRange replacement;
+};
+} // namespace mlir
+
+#endif
--- a/mlir/lib/IR/PatternMatch.cpp
+++ b/mlir/lib/IR/PatternMatch.cpp
@@ -137,19 +137,15 @@ ReplaceOpAction::ReplaceOpAction(ArrayRef<IRUnit> irUnits,
 
 void ReplaceOpAction::print(raw_ostream &os) const {
   OpPrintingFlags flags;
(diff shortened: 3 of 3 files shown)


mlir/include/mlir/IR/PatternMatch.h
```diff
@@ -652,6 +653,9 @@ class RewriterBase : public OpBuilder {
   /// Find uses of `from` and replace them with `to`. Also notify the listener
   /// about every in-place op modification (for every use that was replaced).
   void replaceAllUsesWith(Value from, Value to) {
+    if (auto *fromOp = from.getDefiningOp())
```
cferry-AMD, line 656:
```
Also notify the listener
  /// about every in-place op modification (for every use that was replaced).
```

Was this not up-to-date? Or is an action something else than the listener?

mlir/include/mlir/IR/PatternMatchAction.h
```diff
@@ -0,0 +1,20 @@
+#ifndef MLIR_IR_PATTERNMATCHACTION_H
+#define MLIR_IR_PATTERNMATCHACTION_H
+
+#include "mlir/IR/Action.h"
+
+namespace mlir {
+struct ReplaceOpAction : public tracing::ActionImpl<ReplaceOpAction> {
+  using Base = tracing::ActionImpl<ReplaceOpAction>;
+  ReplaceOpAction(ArrayRef<IRUnit> irUnits, ValueRange replacement);
+  static constexpr StringLiteral tag = "op-replacement";
```
cferry-AMD, line 10:
```suggestion
  static constexpr StringLiteral tag = "ReplaceOpAction";
```

This suggestion applies only if this identifier is used for debugging purposes, so that filtering by the name of the action would work.

cferry-AMD (commented):
`replaceAllUsesWith(ValueRange, ValueRange)` should also be modified, shouldn't it?

mlir/include/mlir/IR/PatternMatch.h
jorickert: My main question is also whats the difference between an action and a listener

mgehre-amd: The RewriterBase supports attaching listeners, and the MLIRContext supports action handlers. In this PR, we use the action handlers on the MLIRContext, which allow us to observe across all passes without modifying each.

mlir/include/mlir/IR/PatternMatchAction.h
mgehre-amd: I was following the naming from the [documentation](https://github.com/Xilinx/llvm-aie/blob/a1da90b27083ff18773b3ac293a8c665137efa80/mlir/docs/ActionTracing.md#L65)

mgehre-amd:
> `replaceAllUsesWith(ValueRange, ValueRange)`
this calls the replaceAllUsesWith(Value from, Value to), which calls the action here
```
void replaceAllUsesWith(ValueRange from, ValueRange to) {
    assert(from.size() == to.size() && "incorrect number of replacements");
    for (auto it : llvm::zip(from, to))
      replaceAllUsesWith(std::get<0>(it), std::get<1>(it));
  }
```

cferry-AMD (dismissed):
Thanks for the claraifications, that looks good.

josel-amd (dismissed):

mlir/include/mlir/IR/PatternMatch.h
```diff
@@ -652,6 +653,9 @@ class RewriterBase : public OpBuilder {
...
+      getContext()->executeAction<ReplaceOpAction>(
+          []() {}, ArrayRef<IRUnit>{fromOp}, to);
```
roberteg16:
```suggestion
          /*actionFn=*/[]() {}, ArrayRef<IRUnit>{fromOp}, to);
```

mlir/lib/IR/PatternMatch.cpp
```diff
@@ -127,6 +127,35 @@ void RewriterBase::replaceAllOpUsesWith(Operation *from, Operation *to) {
...
+                                 ValueRange replacement)
+    : Base(irUnits), replacement(replacement) {
+  assert(irUnits.size() == 1);
+  assert(irUnits[0]);
+  assert(isa<Operation *>(irUnits[0]));
+}
+
+void ReplaceOpAction::print(raw_ostream &os) const {
+  OpPrintingFlags flags;
+  flags.elideLargeElementsAttrs(10);
```
jorickert: I assume we do not have access to the compiler flags at this point?

mlir/lib/IR/PatternMatch.cpp
```diff
@@ -127,6 +127,35 @@ void RewriterBase::replaceAllOpUsesWith(Operation *from, Operation *to) {
...
+  os << "` by ";
+  bool first = true;
+  for (auto r : replacement) {
+    if (!first)
+      os << ", ";
+    os << "`";
+    r.print(os, flags);
+    os << "`";
+    first = false;
+  }
```
roberteg16: Can you use `llvm:interleaveComma` here?

mlir/lib/IR/PatternMatch.cpp
```diff
@@ -265,8 +294,12 @@ void RewriterBase::replaceUsesWithIf(Value from, Value to,
   bool allReplaced = true;
   for (OpOperand &operand : llvm::make_early_inc_range(from.getUses())) {
     bool replace = functor(operand);
-    if (replace)
+    if (replace) {
+      if (auto *fromOp = from.getDefiningOp())
+        getContext()->executeAction<ReplaceOpAction>(
+            []() {}, ArrayRef<IRUnit>{fromOp}, to);
```
roberteg16:
```suggestion
            /*actionFn=*/[]() {}, ArrayRef<IRUnit>{fromOp}, to);
```

mlir/lib/IR/PatternMatch.cpp
jorickert: nit: This could have been written with `llvm::interleave`, but does not make a big difference

jorickert (dismissed):

mlir/include/mlir/IR/PatternMatch.h
```diff
@@ -13,6 +13,7 @@
 #include "mlir/IR/BuiltinOps.h"
 #include "llvm/ADT/FunctionExtras.h"
 #include "llvm/Support/TypeName.h"
+#include <mlir/IR/PatternMatchAction.h>
```
roberteg16: I believe mlir project prefers `""` for internal headers
```suggestion
#include "mlir/IR/PatternMatchAction.h"
```

roberteg16 (dismissed):
Thanks!

mlir/include/mlir/IR/PatternMatch.h
cferry-AMD: Thanks Robert, TIL :)

mlir/lib/IR/PatternMatch.cpp
mgehre-amd: This is just for debugging, so I think it's fine to not print large constants?

mlir/lib/IR/PatternMatch.cpp
mgehre-amd: done

mlir/lib/IR/PatternMatch.cpp
mgehre-amd: done

mlir/lib/IR/PatternMatch.cpp
jorickert: I agree, but if its not too hard to wire up i think it would be great if its controlled by the small element threshold.
But thats just a NIT

roberteg16 (approved):

PR #712, opened 2025-11-17
[AIEX][NFC] Refactor instruction selection to share common code across AIE targets

abhinay-anubola:
Refactor instruction selection by moving common logic from
AIE2PInstructionSelector into AIEBaseInstructionSelector for:
- Set/GetControlRegister
- UPS
- LOAD_UNPACK
- STORE_PACK
- STORE_SRS

Also separates status registers from control registers, adds
AIE2P support for get/set_status_reg, and refactors CGBuiltins
for FIFO LD/ST.

### DIFF
--- a/llvm/lib/Target/AIE/AIE2RegisterInfo.cpp
+++ b/llvm/lib/Target/AIE/AIE2RegisterInfo.cpp
@@ -566,3 +566,7 @@ AIE2RegisterInfo::matchControlRegisterBitwidth(Register CtrlReg,
     llvm_unreachable("Unknown control register.");
   }
 }
+
+Register AIE2RegisterInfo::getUnpackSignCtrlReg() const {
+  return AIE2::crUnpackSign;
+}
--- a/llvm/lib/Target/AIE/AIE2RegisterInfo.cpp
+++ b/llvm/lib/Target/AIE/AIE2RegisterInfo.cpp
@@ -539,3 +539,22 @@ bool AIE2RegisterInfo::isVecOrAccRegClass(const TargetRegisterClass &RC) const {
 
   return false;
 }
+
+unsigned
+AIE2RegisterInfo::matchControlRegisterBitwidth(Register CtrlReg,
+                                               unsigned SrcConstVal) const {
+  // Modulo by width of control regs.  To constrain the max possible value in
+  // the register according to register width.
+  switch (CtrlReg) {
+  case AIE2::crSat:
+    return SrcConstVal % (1 << 2);
+  case AIE2::crRnd:
+    return SrcConstVal % (1 << 4);
+  case AIE2::crF2FMask:
+  case AIE2::crF2IMask:
+  case AIE2::crFPMask:
+    return SrcConstVal % (1 << 5);
+  default:
+    llvm_unreachable("Unknown control register.");
+  }
+}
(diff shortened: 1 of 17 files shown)


llvm/lib/Target/AIE/AIE2RegisterInfo.cpp
```diff
@@ -539,3 +539,22 @@ bool AIE2RegisterInfo::isVecOrAccRegClass(const TargetRegisterClass &RC) const {
...
+  // the register according to register width.
+  switch (CtrlReg) {
+  case AIE2::crSat:
+    return SrcConstVal % (1 << 2);
+  case AIE2::crRnd:
+    return SrcConstVal % (1 << 4);
+  case AIE2::crF2FMask:
+  case AIE2::crF2IMask:
+  case AIE2::crFPMask:
+    return SrcConstVal % (1 << 5);
```
martien-de-jong, line 564: Perhaps we can generalize this using getRegSizeInBits(Reg, MRI)

llvm/lib/Target/AIE/AIE2RegisterInfo.cpp
```diff
@@ -539,3 +539,22 @@ bool AIE2RegisterInfo::isVecOrAccRegClass(const TargetRegisterClass &RC) const {
...
 }
+
+unsigned
+AIE2RegisterInfo::matchControlRegisterBitwidth(Register CtrlReg,
+                                               unsigned SrcConstVal) const {
+  // Modulo by width of control regs.  To constrain the max possible value in
+  // the register according to register width.
+  switch (CtrlReg) {
+  case AIE2::crSat:
+    return SrcConstVal % (1 << 2);
```
martien-de-jong, line 558: CHECK: this is mimicking the hardware when it would move it from a register source. To be frank, I would prefer a frontend error on writing an out of bound constant value.

llvm/lib/Target/AIE/AIEBaseInstructionSelector.cpp
```diff
@@ -1046,3 +1046,64 @@ bool AIEBaseInstructionSelector::selectWriteTM(MachineInstr &I,
...
+bool AIEBaseInstructionSelector::selectSetControlRegister(
+    MachineInstr &I, MachineRegisterInfo &MRI) {
+
+  const Register IdxReg = I.getOperand(1).getReg();
+  const Register SrcReg = I.getOperand(2).getReg();
+
+  // Check if the argument is constant for register map index.
+  const auto Idx = getIConstantVRegValWithLookThrough(IdxReg, MRI);
+  if (!Idx)
+    llvm_unreachable("Expected const value for control register map index.");
```
martien-de-jong: getIConstantVRegValWithLookThroughOrFail() could encapsulate the llvm_unreachable.

llvm/lib/Target/AIE/AIE2RegisterInfo.cpp
martien-de-jong: Hmm. It looks as if their size is 32 bit...

abhinay-anubola: Yes, you're right `getRegSizeInBits` would return 32. However, this function needs the semantic bit width.

llvm/lib/Target/AIE/AIE2RegisterInfo.cpp
abhinay-anubola: Where should the valid range constraints for control registers be enforced, at the intrinsic definition in headers?

mludevid:
Please sync with @F-Stuckmann on how the UNPACK refactors from this PR can be used instead of the ones from his PR: https://github.com/Xilinx/llvm-aie/pull/782. I have a feeling the commit from his PR might not be needed after all.

F-Stuckmann (approved):
Nice

PR #714, opened 2025-11-20
[Clang][AIEX] Disable Vector Combine from the frontend

andcarminati:
We can handle all transformations (legalization) made by VC, however, in some cases, those transformations can lead to very unoptimized code, specially when unaligned loads are created and CSEd afterwards by GVN. Also, all the missed opportunities are picked by our pre-legalizer combiner.

Also, toolchain tests were included for all pushed backend options.


andcarminati:
Full L1 QoR:
Core_Insn_Count
Core_PMSize_absolute
Core_StackSize_absolute

Stack regression related to SoftMaxRowMajor_int8_0 (not a focus kernel).

PR #716, opened 2025-11-20
[AIEX] Combine vector xor with MSB bitmask into vector add

konstantinschwarz:
AIEs don't have native vector xor support, turn this specific pattern into a vector add


llvm/lib/Target/AIE/AIECombinerHelper.cpp
```diff
@@ -4292,3 +4292,34 @@ bool llvm::matchExtractVecEltAssertBcst(MachineInstr &MI,
...
+  auto Cst = getIConstantVRegValWithLookThrough(ScalarReg, MRI);
+  if (!Cst)
+    return false;
+
+  // Get the element type from the broadcast vector destination
+  LLT BroadcastTy = MRI.getType(BroadcastReg);
+  if (!BroadcastTy.isVector())
+    return false;
+
+  unsigned ElemBitWidth = BroadcastTy.getElementType().getSizeInBits();
```
andcarminati, line 4311: nit: const

llvm/lib/Target/AIE/AIECombinerHelper.cpp
```diff
@@ -4292,3 +4292,34 @@ bool llvm::matchExtractVecEltAssertBcst(MachineInstr &MI,
...
+
+  // Get the value as unsigned to properly compare with MSB masks
+  uint64_t Value = Cst->Value.getZExtValue();
```
andcarminati: nit: const

llvm/lib/Target/AIE/AIECombinerHelper.cpp
```diff
@@ -4292,3 +4292,34 @@ bool llvm::matchExtractVecEltAssertBcst(MachineInstr &MI,
...
+
+  // Check if only MSB is set for the vector element type
+  if (ElemBitWidth == 8)
```
andcarminati, line 4317: Just as food for the brain, is it possible to support even more constants by changing the element time and applying bitcasts as a kind of generalization ;-)

andcarminati (approved):

PR #718, opened 2025-11-25
Martien.ubsanfix

martien-de-jong:
As reported by @konstantinschwarz. When setting the SlotID, we also computed the slotset, leading to a negative shift count for the default slot, which has ID=-1.

### DIFF
--- a/llvm/utils/TableGen/CodeGenFormat.cpp
+++ b/llvm/utils/TableGen/CodeGenFormat.cpp
@@ -1112,11 +1112,9 @@ void TGTargetSlots::finalizeSlots() {
 
   // Give an ID for each slot
   int SlotID = 0;
-  for (RecordSlot &Slot : Slots)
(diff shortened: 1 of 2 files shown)


llvm/utils/TableGen/CodeGenFormat.cpp
```diff
@@ -1112,11 +1112,9 @@ void TGTargetSlots::finalizeSlots() {
 
   // Give an ID for each slot
   int SlotID = 0;
-  for (RecordSlot &Slot : Slots)
-    if (Slot.second.isDefaultSlot())
-      Slot.second.setNumSlot(-1);
```
martien-de-jong, line 1117: We only need the ID, which is now default initialized

andcarminati (approved):
LGTM.

F-Stuckmann (approved):

PR #722, opened 2025-11-26
[AIEX] Disable hasCompleteDecoder for all instructions and registers

mludevid:
Enabling `hasCompleteDecoder` allows failing early when decoding an incorrect object file. In exchange you have to guarantee that the decoder methods never have to backtrack. That is not always the case for the registers and instructions in AIE targets. To avoid having to identify all the cases were backtracking is needed just to get a small speedup when decoding an incorrect object file this PR disables the `hasCompleteDecoder` flag for all instructions and registers in AIE targets.

### DIFF
--- a/llvm/lib/Target/AIE/aie2p/AIE2PRegisterInfo.td
+++ b/llvm/lib/Target/AIE/aie2p/AIE2PRegisterInfo.td
@@ -11,6 +11,8 @@
 //  Declarations that describe the AIE2p Register files
 //===----------------------------------------------------------------------===//
 
+include "AIEBaseRegisterInfo.td"
+
 let Namespace = "AIE2P" in {
 
 def sub_l_even : SubRegIndex<32,  0>;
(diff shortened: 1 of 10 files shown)


khallouh (dismissed):
LGTM!

llvm/lib/Target/AIE/aie2p/AIE2PRegisterInfo.td
```diff
@@ -110,6 +110,7 @@ class AIE2PRegisterClass <int size, int align, list<ValueType> regTypes, dag reg
       [DefaultMode],
       [RegInfo</*size*/size, /*spill size*/size, /*spill alignment*/align>]>;
   let DecodeZeroBitOperand = true;
+  let hasCompleteDecoder = false;
```
martien-de-jong: It's unclear to me where we would decode a registerclass and where we would decode a registeroperand.

martien-de-jong (dismissed):
I left a nit asking after an implementation detail, but for the rest it looks sane.

llvm/lib/Target/AIE/aie2p/AIE2PRegisterInfo.td
mludevid: I'm not sure I understand the question. This flag was needed in AIE2 both for register classes (the vector and accumulator ones) and for register operands. I double checked that the flag is actually needed (taking it out leads to decoding failures), but I didn't dig any deeper to understand to concrete instructions affected by this. I can have a look if you want me to.
In any case, the point of this PR is to disable `hasCompleteDecoder` everywhere to precisely avoid having to do these investigations into the decoding of instructions and their operands to find the conflicts that need to be resolved.

mludevid:
Discussed offline: I added a new base register class for AIE to factor out the common functionality. The new base register operand even replaces the existing target specific ones, since they are all identical.

martien-de-jong (approved):

PR #725, opened 2025-11-28
[AIEX] Fix instruction leaking and improve observer support in combines

mludevid:
We were using `removeFromParent` and `eraseFromParent` very inconsistently in the combiners. This PR cleans them up to always use `eraseFromParent` to prevent leaking memory. In the case of the `GlobalCombiners`, a new deferred deletion was needed, because the combiner creates a map of combined instructions to be used later in the pass. Immediately deleting the instructions would lead to dangling pointers.
Also, some missing observer calls were added and unnecessary ones removed (the MachineIRBuilder automatically calls `createdInstr` when building an instruction).
Note: AFAIK the new observer calls could have QoR implications by enabling more optimizations that rely on these updates from the observer (e.g. CSE). I'll run the benchmarks but I don't expect any changes in practice. The main goal of this PR is to get rid of the memory leaks.

### DIFF
--- a/llvm/lib/Target/AIE/AIECombinerHelper.cpp
+++ b/llvm/lib/Target/AIE/AIECombinerHelper.cpp
@@ -741,7 +741,8 @@ void llvm::applyConcatUnmergePhis(MachineInstr &ConcatI,
       MRI.getVRegDef(*MatchInfo.UnmergeSourceReg)->getParent();
   NewPHI->addOperand(MachineOperand::CreateMBB(UnmergeMBB));
   LLVM_DEBUG(dbgs() << "Created New Instruction " << *NewPHI.getInstr());
-  ConcatI.removeFromParent();
+  Observer.erasingInstr(ConcatI);
+  ConcatI.eraseFromParent();
 }
 
 bool llvm::matchGlobalPtrModOptimizer(MachineInstr &MemI,
@@ -811,7 +812,9 @@ void llvm::applyLdStInc(MachineInstr &MemI, MachineRegisterInfo &MRI,
     if (Helper.dominates(*Instr, *NewInstr))
       continue;
 
+    Observer.changingInstr(*Instr);
     Instr->moveBefore(NewInstr);
+    Observer.changedInstr(*Instr);
     LLVM_DEBUG(dbgs() << "Move Instr before " << *Instr);
   }
 
@@ -827,7 +830,9 @@ void llvm::applyLdStInc(MachineInstr &MemI, MachineRegisterInfo &MRI,
         continue;
 
       LLVM_DEBUG(dbgs() << "Delaying Instr " << *Instr);
+      Observer.changingInstr(*Instr);
       Instr->moveBefore(CombinedInsertionPoint);
+      Observer.changedInstr(*Instr);
     }
   }
 
@@ -866,9 +871,19 @@ void llvm::applyLdStInc(MachineInstr &MemI, MachineRegisterInfo &MRI,
     assert(RemoveMI->getParent() &&
            "RemoveMI was already deleted. This Combiner may have a conflict "
            "with the Combiner that already removed the MachineInstr.");
+    // Notify observer, then remove from BB, then defer deletion.
+    // We defer actual deallocation until end of combining pass because
+    // later combiners may reference this instruction via the remapping
+    // mechanism (createMapping/getMappedInstr).
+    Observer.erasingInstr(*RemoveMI);
     RemoveMI->removeFromParent();
+    GlobalCombinerPtr->deferDelete(RemoveMI);
   }
+
+  // Notify observer, remove from BB, then defer deletion for root instruction.
+  Observer.erasingInstr(MemI);
   MemI.removeFromParent();
+  GlobalCombinerPtr->deferDelete(&MemI);
 }
 
 // Match all equivalents of these:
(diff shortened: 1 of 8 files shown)


llvm/lib/Target/AIE/AIECombinerHelper.cpp
```diff
@@ -866,8 +871,18 @@ void llvm::applyLdStInc(MachineInstr &MemI, MachineRegisterInfo &MRI,
     assert(RemoveMI->getParent() &&
            "RemoveMI was already deleted. This Combiner may have a conflict "
            "with the Combiner that already removed the MachineInstr.");
+    // Notify observer, defer deletion, then remove from BB.
+    // We defer actual deallocation until end of combining pass because
+    // later combiners may reference this instruction via the remapping
+    // mechanism (createMapping/getMappedInstr).
+    Observer.erasingInstr(*RemoveMI);
+    GlobalCombinerPtr->deferDelete(RemoveMI);
```
mludevid: I think that ideally, we would modify the combine to not need the mapping of removed instructions. That would allow erasing the combined instructions directly. Maybe @F-Stuckmann can give some insights into whether that'd be possible and how much work would be involved.

martien-de-jong: I'm not sure whether this is relevant, but sometimes it's necessary to maintain todo lists, and the instruction's address is the only sensible identity you can use. Even if you're not derefencing the information it's dangerous to free, since the freed node can be reused, and will start to impersonate the old node.
I find that deferDelete is a robust approach in any transformation that needs some mappings to remain sane during the transformation.

mludevid: I hadn't thought about that, then I guess we'll be keeping this solution.

F-Stuckmann: The global combiners unfortunately have a lot of instructions that move up and down. 
We use the "deleted" instructions to recognize, if we have to move already combined instruction.
Therefore, @mludevid your approach is nice without causing too many headaches.

llvm/lib/Target/AIE/AIEPtrModOptimizer.h
```diff
@@ -71,6 +87,28 @@ class FoundCombiners {
 
   /// \return whether Analysis Pass generated this Object
   bool hasAnalysis() const { return GeneratedFromAnalysisPass; }
+
+  /// Add an instruction to be deleted at the end of the combining pass.
+  /// Has to be called *before* removeFromParent() to capture the
+  /// MachineFunction.
+  void deferDelete(MachineInstr *MI) {
```
niwinanto, line 84: Why do you need to constrain this API to be used before `removeFromParent()`? Anyway we are going to delete it only after removing from BB.

mludevid: The reason is that we need a reference to the machine function when deleting the instruction. And the easiest way I'm aware of getting a reference to it is by going through the basic block. If the instruction was already removed from the BB that's not possible anymore.

F-Stuckmann: A Global combiner consists of a CombineRoot (Instruction to match, i.e. Load/Store Instruction) and an InsertionPoint (where we want to insert the combined Instruction).
After combining (e.g. Ptr_ADD + Load Instruction)  we delete the CombineRoot (the Load Instruction).

Now to allow the global combiners to be more powerful, I allow it to move Instructions before or after the InsertionPoint.
These to-move Instructions are collected before any combiners are applied. 
But once we start applying combiners, the Load/Store and ptr_adds are transformed into a postinc/preinc loads/stores. And what we really want is to move the postinc/preinc load/stores. 

So before applying an individual combiner, I remap all the to-move instructions from the deleted instructions to the newly inserted combine Instructions (preinc/postinc load/store).
During this remapping i check out if I encounter any of the deleted instructions and then replace it in the to-move instructions with the combined Instruction.

niwinanto: We are running a machine function pass, so we are only dealing with one MF and you could pass the current MF to `finalizeDeferredDeletes`

mludevid: That is true for the calls to `finalizeDeferredDeletes` from the combiners. There we do know in which `MF` we are running. But I also call `finalizeDeferredDeletes` in the deconstructor of the `FoundCombiners` class. There we don't have a reference to the `MF`. What we could do is remove the call to `finalizeDeferredDeletes` in the deconstructor. It is only there as a safety measure, the combiners should be calling `finalizeDeferredDeletes` anwyays. Would that be okay for you?

niwinanto: We could make a MF member to the class, but as you mentioned, I dont think we really need to call `finalizeDeferredDeletes` from the destructor. One more suggestion I have is to use the assert when we have a BB to the MI. I mean, assert when we use deferred delete with out removing the instruction from BB. I feel that make this API more foolproof. Alternately, we could call `removeFromParent` inside `deferDelete` as well, which make it more easy to use.

mludevid: I had another idea: What if we pass the `MF` to `deferDelete`? At all call-sites we have the machine function available, we can either store it as we do it right now together with the instruction in a list of  machine functions and instructions or we can store it separately in a new `MF` member variable of `FoundCombiners`. This new member would need to be optional though, because I see that @F-Stuckmann's algorithm relies on the existence an default constructor. The advantages would be that we wouldn't have to update any of the calls to `finalizeDeferredDeletes` and we could keep the call in the deconstructor for additional safety.

niwinanto: Discussed offline and decided to pass MF through `finalizeDeferredDeletes` and remove deletion from destructor.

mludevid: Fixed.

niwinanto (approved):
LGTM!

PR #726, opened 2025-12-02
Guard cmake JOB_POOL property against duplicate entries if ...

jorickert:
…  HandleLLVMOptions gets included multiple times.

Otherwise including HandleLLVMOptions  multiple times and setting -DLLVM_PARALLEL_LINK_JOBS=SOME_NUMBER leads to the following ninja error : `duplicate pool 'link_job_pool'`

I created a PR upstream for the same issue https://github.com/llvm/llvm-project/pull/169052, but its a bit hard to get reviews there, so opening here too


julio-gorge (approved):

F-Stuckmann (approved):
nit: You may want to consider adding a unit-test.

PR #728, opened 2025-12-03
Fix MachineVerifier hotfix in SuperRegister Rewriter

F-Stuckmann:
LiveIntervals are a cached representation of when Registers are defined and used.
When we modify any Register, there is no automatic repair mechanism to update the LiveInterval cache.

**AIESuperRegUtils::rewriteFullCopy**
When we transform a superregister copy into its individual components this happens: 
`40      %18:ed = COPY %19:ed`
`100     Use %18`
to 
`40      undef %18:sub_mod= COPY %19.sub_mod`
`50      %18:sub_dim_size= COPY %19.sub_dim_size`
`100     Use %18`

The LiveInterval in the first case consists of 
`%18 [40r,100r:0) L0020 %18 [40r,100r:0) L0080 0@40r`
and the liveInterval also tracks that only 2 sublanes are alive (i.e. 2 and 8).
The other sublanes are undef.
The Register %18 and all the sublanes are attached to the `ValueNumber 0` at SlotIndex 40.

In the second example the live Intervals are now
`%18 [40r,100r:0) L0020 %18 [50r,100r:0) L0080 0@40`
Note: that Lane `L0080` now has a shorter LiveInterval and is **not** actually defined at 40r!

**Splitting range**
In the next step we remap the subregisters to individual Registers: 
`40      %20:sub_mod= COPY %19.sub_mod`
`50      %21:sub_dim_size= COPY %19.sub_dim_size`
`100     Use %20 & %21`

**Assigning Range**
Here we remap the LiveIntervals to the individual SubLanes. However, the definition point is not properly calculated.
`%20 [40r,100r:0) 0@100r `
`%21 [50r,100r:0) 0@100r`
Note: The Def point should be `0@40r` and `0@50r`.

**Bug**
In the *rewriteFullCopy*, we did not properly update the live intervals of the defs!
This causes %20 and %21 to not have live Segments, which ultimately leads to the verifier crashes.

As expected no changed in QoR.

P.S. I have been in hell and was not amused! Would not recommend a repeat. 
0/5 stars.

### DIFF
--- a/llvm/lib/Target/AIE/AIESuperRegUtils.cpp
+++ b/llvm/lib/Target/AIE/AIESuperRegUtils.cpp
@@ -108,26 +108,7 @@ void rewriteFullCopy(MachineInstr &CopyMI, LiveIntervals &LIS,
   const std::set<int> CopySubRegs =
       TRI.getSubRegSplit(MRI.getRegClass(DstReg)->getID());
 
-  if (!VRM.hasPhys(DstReg)) {
-    // FIXME: This pass may cause verification failures. The fix should
-    // be in the MachineVerifier. This is a very uncommon case where the
-    // destination register was not allocated yet.
-    // The machine verifier does not properly handle the semantics of:
-    // 1. **Partial register definitions with `undefined`**: When the first
-    // subregister is defined with `undefined`, it doesn't expect subsequent
-    // definitions to implicitly read that lane.
-    // 2. **Lane-based liveness for composite registers**: The verifier expects
-    // a continuous live range for the entire register, but with subregister
-    // definitions, different lanes have different live ranges that are being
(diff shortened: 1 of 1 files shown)


andcarminati (approved):
LGTM. Thank you for investigating this FIXME. I tested this path in the past, but maybe the solution was overtaken by some other transient change.

martien-de-jong (approved):

F-Stuckmann:
> LGTM. Thank you for investigating this FIXME. I tested this path in the past, but maybe the solution was overtaken by some other transient change.

I think `LIS.ReplaceMachineInstrInMaps(CopyMI, *FirstMI);` got you into quite a pickle.

PR #730, opened 2025-12-04
[MLIR][Quant] Introduce BlockFloatQuantizedType to represent block-float quantized types like BFP16

jorickert:


mlir/include/mlir/Dialect/Quant/IR/QuantTypes.h
```diff
@@ -224,6 +227,93 @@ class AnyQuantizedType
...
+/// data along a particular axis (e.g. BFP16). The concrete block format
+/// determines the implied storage characteristics and is not exposed in the IR.
+/// This class is experimental and may be subject to change.
+/// Design decisions:
+///  - The base class requires an integral storage type. For
+///  block-quantized/packed types, the required storage with depends on the
+///  number of elements. For example, a single BFP16 element requires 16 bits to
+///  be represented, but a block of 8 BFP16 elements can be packed into 9 bits
+///  per element on average (72 bits total). The storage type for
+///  BlockFloatQuantizedType is the "packed" type, so for BFP16 i9.
```
ljfitz:
```suggestion
///  per element on average (72 bits total). The storage type for
///  an element from BlockFloatQuantizedType is the "packed" type
///  divided by the number of packed elements, so for BFP16 i9.
```

ljfitz (dismissed):

ge28boj (dismissed):

mlir/include/mlir/Dialect/Quant/IR/QuantTypes.h
```diff
@@ -224,6 +227,93 @@ class AnyQuantizedType
...
+
+  static LogicalResult
+  verifyInvariants(function_ref<InFlightDiagnostic()> emitError,
+                   uint32_t blockModeRaw, int32_t axis, unsigned flags,
+                   Type storageType, Type expressedType, int64_t storageTypeMin,
+                   int64_t storageTypeMax);
+
+  Type getStorageType() const {
+    assert(false &&
+           "BlockFloatQuantizedType does not have a direct storage type");
```
ehsan-toosi, line 278: NIT: Is there no other way to use something else instead of assert?
llvm_unreachable("BlockFloatQuantizedType does not have a direct storage type");

mlir/lib/Dialect/Quant/IR/QuantOps.cpp
```diff
@@ -37,30 +39,62 @@ namespace {
...
                                         Type containerType) {
-  auto quantizedPerAxisType = dyn_cast<UniformQuantizedPerAxisType>(quantizedType);
+  auto quantizedPerAxisType =
+      dyn_cast<UniformQuantizedPerAxisType>(quantizedType);
   if (!quantizedPerAxisType)
     return success();
 
-  auto tensorType = dyn_cast<TensorType>(containerType);
-  if (!tensorType)
+  auto shapedType = dyn_cast<ShapedType>(containerType);
```
ehsan-toosi, line 47: nit: do we support other types other than tesnor that we needed this change?

ehsan-toosi (dismissed):

mlir/include/mlir/Dialect/Quant/IR/QuantTypes.h
jorickert: IMO llvm_unreachable is generally dangerous, at it compiles to undefined behaviour in release mode

mlir/lib/Dialect/Quant/IR/QuantOps.cpp
jorickert: No, but i generally think its an antipattern to check for TensortType when only being interested in the ShapedType properties. ShapedType can easily be added to new types, while TensorType can not be extended

ehsan-toosi (approved):

PR #731, opened 2025-12-04
Martien.solve xm

martien-de-jong:
WIP to correct the solver model for the XM slot. 
Also contains preparations to keep track of possible MSP materializations in the scoreboard without actually materializing.
All  coding in those commits was done by Cline in CNN mode (Continuous Nitpicking and Nagging). I added a set of our accepted coding rules for cline.


.clinerules
```diff
@@ -0,0 +1,130 @@
...
+(void)Success;  // Used in assert in debug builds
+assert(Success && "Operation must succeed");
+```
+
+## Build Verification
+
+After making changes to AIE target code, verify the changes by running the AIE CodeGen tests:
+
+```bash
+cd Release && ninja check-llvm-codegen-aie
```
andcarminati, line 204: Maybe ask for a user preference. For example, I follow other naming convention.

martien-de-jong: You can now include a user profile and inspect a variable to check from which codebase it comes.

PR #713, opened 2025-11-20
SubReg Spiller

F-Stuckmann:
still wip

PR #735, opened 2025-12-12
Revert Hotfixes in SubReg Coalescing and Intrinsic Attributes

F-Stuckmann:
Originally these changes were introduced to limit performance impacts, when partially defined SubRegisters are spilled. 
This is caused by the InlineSpiller.cpp only being able to spill full registers. 
So instead of spilling the defined SubRegisters, Inlinespiller will spill the whole Register (also the undefined components). 

Without Register Sequence generation only the defined Registers are spilled and only later register sequenced.
This later register sequencing however is suboptimal and may introduce moves.

Thus these hotifxes limited 
1) which intrinsics can be hoisted outside of a loop body
2) Register Sequence generation within a MBB.

With a proper SubRegister Spilling (https://github.com/Xilinx/llvm-aie/pull/713), these hotfixes can be removed.


F-Stuckmann:
some regressions creep up and the focus has shifted away from this simplification

PR #738, opened 2025-12-15
[NFC][AIE1] fix second Spill reload

F-Stuckmann:
The current state of AIE1 Stack reload has two undefs when we reload a mC Register (Reusing the same StackSlot here is also wrong but this is not my beer to fix)

`3852B	  dead undef %1661.sub_32_lo:mc = LDA_SPIL_GPR %stack.14, implicit $sp :: (load (s64) from %stack.14)`
`3860B	  undef %1661.sub_32_hi:mc = LDA_SPIL_GPR %stack.14, implicit $sp :: (load (s64) from %stack.14)`

This causes us to have multiple Defs for a single VReg if we recalculate the LiveIntervals. Also the first Def should not be dead!
Besides being wrong this also causes the MachineVerifier to crash.

`*** Bad machine code: Multiple connected components in live interval ***`
`- function:    conv2d`
`- interval:    %1661 [3852r,3852d:1)[3860r,3868r:0) 0@3860r 1@3852r  weight:INF`
`0: valnos 0`
`1: valnos 1`

This problem is solved by removing the undef flag for the second Spill-Load Instruction.

Note: This creeped up when looking at the interesting unit-test `llvm/test/CodeGen/AIE/vect/generic_vector_loop.ll` for https://github.com/Xilinx/llvm-aie/pull/713 . The Subreg spiller runs out of registers because an optimization is missing and I at least wanted to fix the obvious machine verify crashes. 

Note 2: `llvm/test/CodeGen/AIE/vect/generic_vector_loop.ll` crashes the Machine Verifier for the regular InlineSpiller XD

Note 3: It was previously hidden because we never actually properly recalculated the LiveIntervals for spill store/reload in InlineSpiller.

### DIFF
--- a/llvm/lib/Target/AIE/aie1/AIE1InstrInfo.cpp
+++ b/llvm/lib/Target/AIE/aie1/AIE1InstrInfo.cpp
@@ -568,7 +568,7 @@ void AIEInstrInfo::loadRegFromStackSlot(
         .addFrameIndex(FI)
         .addMemOperand(CreateMMO(FI));
     BuildMI(MBB, I, DL, get(AIE::LDA_SPIL_GPR))
-        .addReg(DstReg, RegState::DefineNoRead, AIE::sub_32_hi)
+        .addReg(DstReg, RegState::Define, AIE::sub_32_hi)
         .addFrameIndex(FI)
         .addMemOperand(CreateMMO(FI));
     return;


martien-de-jong (approved):
Rather blind approval. We shouldn't get stuck on flaky AIE1 code.

PR #739, opened 2025-12-15
Allow AnyInteger for quant.scast

ilango100:
- Allow any integer type to be used with storage cast
- Checking if the `integerType == quantType.storageType` is already present.
- Fix tests that checks if the storage type is integer


jorickert (approved):
Thanks

ljfitz (approved):

PR #740, opened 2025-12-16
[AIE][PostPipeliner] make PostPipeliner depend less on PostScheduler

martien-de-jong:
AIECC-1093

This makes the postpipeliner more flexible.

We no longer need to set the number of copies predicting the stage count. DAG and SchedInfo nodes are fixed at two copies.

The scoreboard is reduced to the lookahead the pipeliner needs to see the future conflicts with instructions scheduled earlier, which is a pure heuristic.
Correctness is checked separately by populating a  scoreboard with the steady state. This check can advance the scoreboard, since it only needs to check the loop body stage. Hence it can be of fixed size II + pipelinedepth.

The change also reduces runtime, especially for DAG construction

### DIFF
--- a/llvm/lib/Target/AIE/AIEMachineScheduler.cpp
+++ b/llvm/lib/Target/AIE/AIEMachineScheduler.cpp
@@ -4,7 +4,7 @@
 // See https://llvm.org/LICENSE.txt for license information.
 // SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
 //
-// (c) Copyright 2023-2024 Advanced Micro Devices, Inc. or its affiliates
+// (c) Copyright 2023-2025 Advanced Micro Devices, Inc. or its affiliates
 //
 //===----------------------------------------------------------------------===//
 
@@ -1457,14 +1457,16 @@ void llvm::AIEPostRASchedStrategy::buildGraph(ScheduleDAGMI &DAG, AAResults *AA,
   auto &BS = InterBlock.getBlockState(CurMBB);
   const auto &Region = BS.getCurrentRegion();
   int NCopies = 1;
-  if (int II = BS.FixPoint.II) {
+  if (BS.FixPoint.II) {
     assert(BS.Kind == BlockType::Loop);
     assert(BS.getRegions().size() == 1);
+    assert(Region.getBotFixedBundles().empty());
+    assert(Region.getTopFixedBundles().empty());
     // Try to wrap the linear schedule within II.
-    // We virtually unroll the body by the stagecount, computed from rounding
-    // up the length divided by II, adding one more stage to account for
-    // the added resource contention
-    NCopies = (BS.getScheduleLength() + II - 1) / II + 1;
+    // We create two copies of the loop body, which will make the loop-carried
+    // dependences appear as forward dependences between the first and the
+    // second iteration.
+    NCopies = 2;
   }
   DEBUG_BLOCKS(dbgs() << "    buildGraph, NCopies=" << NCopies << "\n");
   for (int S = 0; S < NCopies; S++) {
--- a/llvm/lib/Target/AIE/AIEPostPipeliner.cpp
+++ b/llvm/lib/Target/AIE/AIEPostPipeliner.cpp
@@ -19,6 +19,7 @@
 #include "llvm/Analysis/OptimizationRemarkEmitter.h"
 #include "llvm/CodeGen/MachineBasicBlock.h"
 #include "llvm/CodeGen/MachineOptimizationRemarkEmitter.h"
+#include "llvm/CodeGen/ResourceScoreboard.h"
 #include "llvm/CodeGen/ScheduleDAG.h"
 #include "llvm/CodeGen/ScheduleDAGInstrs.h"
 #include "llvm/Transforms/Utils/LoopUtils.h"
@@ -220,7 +221,7 @@ void PostPipeliner::scheduleNode(SUnit &SU, int Cycle,
   LLVM_DEBUG(dbgs() << "\n");
 
   int Next = SU.NodeNum + NInstr;
-  if (Next < NTotalInstrs) {
+  if (Next < int(Info.Nodes.size())) {
     Info[Next].Earliest = std::max(Info[Next].Earliest, Cycle + II);
   }
 }
@@ -443,7 +444,7 @@ void PostPipeliner::computeRecMII() {
 bool PostPipeliner::computeLoopCarriedParameters() {
 
   // Initialize slot counts.
-  for (int K = 0; K < NTotalInstrs; K++) {
+  for (int K = 0; K < NInstr; K++) {
     auto *MI = DAG->SUnits[K].getInstr();
     Info[K].Slots = getSlotCounts(MI->getOpcode(), TII);
   }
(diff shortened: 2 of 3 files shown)


llvm/lib/Target/AIE/AIEMachineScheduler.cpp
```diff
@@ -1457,14 +1457,16 @@ void llvm::AIEPostRASchedStrategy::buildGraph(ScheduleDAGMI &DAG, AAResults *AA,
...
+  if (BS.FixPoint.II) {
     assert(BS.Kind == BlockType::Loop);
     assert(BS.getRegions().size() == 1);
+    assert(Region.getBotFixedBundles().empty());
+    assert(Region.getTopFixedBundles().empty());
     // Try to wrap the linear schedule within II.
-    // We virtually unroll the body by the stagecount, computed from rounding
-    // up the length divided by II, adding one more stage to account for
-    // the added resource contention
-    NCopies = (BS.getScheduleLength() + II - 1) / II + 1;
```
martien-de-jong, line 1467: This was the last  of the remaining dependences that required the scheduler to run before the pipeliner

llvm/lib/Target/AIE/AIEPostPipeliner.cpp
```diff
@@ -743,16 +747,25 @@ bool PostPipeliner::scheduleFirstIteration(PostPipelinerStrategy &Strategy) {
...
+    // ModCycle + PipelineDepth
+    const int Horizon =
+        std::min(II + PipelineDepth, ScoreboardSize - PipelineDepth);
     LLVM_DEBUG(dbgs() << "  Emit in " << Cycle << "\n");
-    for (int N = 0; N < NCopies; N++) {
-      if (N > 0 && HR.checkConflict(Scoreboard, *MI, Cycle)) {
-        LLVM_DEBUG(dbgs() << "Conflict in iteration N=" << N << "\n");
+    int Iter = 0;
+    while (Cycle < Horizon) {
+      if (HR.checkConflict(Scoreboard, *MI, Cycle)) {
```
martien-de-jong, line 762: This scoreboard is no longer used for the correctness check, which we now do explicitly in scheduleOtherIterations. Here we just foreshadow conflicts with the first iteration.

llvm/lib/Target/AIE/AIEPostPipeliner.cpp
```diff
@@ -776,62 +789,82 @@ void dumpEarliestChain(const ScheduleInfo &Info, int N) {
...
-                   dumpEarliestChain(Info, N));
-        if (Strategy.mobility(ModuloSU) > 0) {
-          // The modulo Node can be delayed
-          ModuloNode.TweakedEarliest = ModuloNode.Earliest + 1;
+  // Make sure that copies can be placed at II from the previous one.
+  // We only check the second iteration. This may have had earliest
+  // pushed by LCDs from the first iteration. Since the dag doesn't change,
+  // the third iteration behaves the same.
+  for (int K = 0; K < NInstr; K++) {
+    const int N = K + NInstr;
```
martien-de-jong, line 803: we lost an outer loop, so we have a major diff on indentation. NInstr corresponds to the former outerloop control variable L.

llvm/lib/Target/AIE/AIEPostPipeliner.cpp
```diff
@@ -1253,37 +1286,38 @@ bool PostPipeliner::applySolver(const SolverData &Data, SWPSolver &Solver,
...
-  // in the unrolled dag.
-  const int Size = NCopies * II + HR.getPipelineDepth();
-  Scoreboard.config(0, Size - 1);
+  unsigned Depth = 0;
+  for (auto &SU : DAG->SUnits) {
+    if (int(SU.NodeNum) >= NInstr) {
+      break;
+    }
+    Depth = std::max(Depth, SU.getDepth());
+  }
```
martien-de-jong: This may have been a transient use of Depth. If so, I'll remove it.

llvm/lib/Target/AIE/AIEPostPipeliner.cpp
```diff
@@ -1389,16 +1423,6 @@ bool PostPipeliner::checkStages() {
...
   NStages = MaxStage + 1;
-
-  // Now check that we don't exceed the number of copies in the DAG. In that
-  // case we didn't reach steady state, and we may have missed conflicts.
-  // We expect this to be rare.
-  if (NStages > NCopies) {
-    DEBUG_SUMMARY(dbgs() << "PostPipeliner: Unsafe stage count, NCopies="
-                         << NCopies << "\n");
-    return false;
-  }
```
martien-de-jong, line 1400: We can't makes this check anymore, but we also 'reach steady state' by constructing it directly from NStages

llvm/lib/Target/AIE/AIEPostPipeliner.cpp
```diff
@@ -443,7 +444,7 @@ void PostPipeliner::computeRecMII() {
 bool PostPipeliner::computeLoopCarriedParameters() {
 
   // Initialize slot counts.
-  for (int K = 0; K < NTotalInstrs; K++) {
+  for (int K = 0; K < NInstr; K++) {
```
F-Stuckmann, line 447: Could we clarify if we are using NInstr of ScheduleInfo  or of PostPipeliner ? 
Are they the same? It seems as if PostPipeliner::NInstr is closer to ScheduleInfo::Nodes

llvm/lib/Target/AIE/AIEPostPipeliner.cpp
```diff
@@ -524,7 +527,7 @@ bool PostPipeliner::computeLoopCarriedParameters() {
   }
 
   LLVM_DEBUG(dbgs() << "Final Earliest - Latest:\n");
-  for (int K = 0; K < NTotalInstrs; K++) {
+  for (int K = 0; K < 2 * NInstr; K++) {
```
F-Stuckmann: nit: for consistency can we use `int(Info.Nodes.size())`  like above?

llvm/lib/Target/AIE/AIEPostPipeliner.cpp
martien-de-jong: Good point. NInstr was added to ScheduleInfo at some point; I guess the PostPipeliner one is redundant and should be eliminated. But whenever you see NInstr anywhere in postpipeliner code, it is the number  schedulable instructions in the basic block, i.e. the number of nodes in one copy of the dag.

llvm/lib/Target/AIE/AIEPostPipeliner.cpp
martien-de-jong: or better, run a range based for loop.

llvm/lib/Target/AIE/AIEPostPipeliner.cpp
martien-de-jong: On second thought, no. Ninstr is copied into ScheduleInfo to decouple an interface from the PostPipeliner itself, but most of the NInstr occurrences are to that of the PostPipeliner. Anyway, they both represent the same invariant number.

F-Stuckmann (approved):

PR #742, opened 2025-12-18
[SROA] Prevents conversions of pointer allocas to byte arrays 

andcarminati:
SROA may convert pointer-typed allocas to byte arrays on targets with non-byte-sized pointers, preventing mem2reg promotion.

On AIE2P (20-bit pointers, 3 bytes), when SROA processes an alloca of type `ptr addrspace(5)`, it converts it to `[3 x i8]` because:

1. The pointer size (24 bits) is not a legal integer type
2. SROA's type selection falls back to `ArrayType::get(Type::getInt8Ty(), 3)`

This creates a type mismatch:

- Alloca type: `[3 x i8]`
- Load/store types: `ptr addrspace(5)`

The mismatch causes `isAllocaPromotable()` to return false, preventing mem2reg from promoting the alloca to SSA form.


llvm/lib/Transforms/Scalar/SROA.cpp
```diff
@@ -4886,8 +4886,13 @@ AllocaInst *SROA::rewritePartition(AllocaInst &AI, AllocaSlices &AS,
         SliceTy = TypePartitionTy;
     }
 
-  if (!SliceTy)
-    SliceTy = ArrayType::get(Type::getInt8Ty(*C), P.size());
+  if (!SliceTy) {
```
konstantinschwarz, line 4889: Do we still need our original change in `AllocaSliceRewriter`?

andcarminati: I guess no, but I can double check and create a follow-up PR.

konstantinschwarz (approved):

PR #744, opened 2025-12-19
[AIE] Fix dereferencing an unengaged std::optional<int>.

SagarMaheshwari99:


llvm/lib/Target/AIE/AIEBaseSubtarget.cpp
```diff
@@ -459,16 +459,16 @@ class EmitFixedSUnits : public ScheduleDAGMutation {
       SDep Dep(&FreeSU, SDep::Artificial);
       auto Latency =
           AIE::maxLatency(&MI, *TII, *ItinData, /*IncludeStages=*/true);
-      if (TII->isLock(MI.getOpcode())) {
-        Dep.setLatency(std::max(
-            TII->getCoreResumeCycleAfterLock() -
-                *TII->getFirstMemoryCycle(FixedDepMI->getDesc().SchedClass) + 1,
-            Latency));
-      } else if (TII->isLock(FixedDepMI->getOpcode())) {
+      auto OptFirstMemCycle =
```
niwinanto, line 462: Please include a test.

llvm/lib/Target/AIE/AIEBaseSubtarget.cpp
```diff
@@ -459,16 +459,16 @@ class EmitFixedSUnits : public ScheduleDAGMutation {
...
+          TII->getFirstMemoryCycle(FixedDepMI->getDesc().SchedClass);
```
niwinanto, line 463: Also, would be nice to know which SchedClass does not get this info

PR #745, opened 2025-12-23
[AIE2P] Fix II_VLDB_POPX_512 II_VLDB_FILLX_512 itineraries

andcarminati:
Remove unnecessary port.


martien-de-jong (commented):
Looks convincing, I'm just not sure whether we should edit this generated file. @mludevid could you approve?

mludevid (dismissed):
Yes, I think we should merge this. I expect there will be more cases that we missed. I'll follow-up on this in the new year.

Nit: It would be nice to add a test for this special case, so we don't miss this in case we decide to regenerate the file again.

mludevid (approved):
LGTM! Thanks for adding the tests.

PR #746, opened 2025-12-23
[AIE2P] Add G_AIE_VSEL simplification combiner

andcarminati:
This is very beneficial for some new networks.


llvm/lib/Target/AIE/AIECombinerHelper.cpp
```diff
@@ -4725,3 +4725,86 @@ bool llvm::matchBroadcastExtractToCopy(MachineInstr &MI,
...
+  // Check if it's either half-and-half pattern
+  if (SelMask == UpperHalfMask || SelMask == LowerHalfMask) {
+    const bool LowerFromSrc1 = (SelMask == LowerHalfMask);
+
+    MatchInfo = [Src1Reg, Src2Reg, DstReg, DstTy, LowerFromSrc1,
+                 &MRI](MachineIRBuilder &B) {
+      const LLT HalfTy = DstTy.divide(2);
+
+      // Unmerge both sources into halves
+      Register Src1Lo = MRI.createGenericVirtualRegister(HalfTy);
```
andcarminati: nit const.

llvm/lib/Target/AIE/AIECombinerHelper.cpp
```diff
@@ -4725,3 +4725,86 @@ bool llvm::matchBroadcastExtractToCopy(MachineInstr &MI,
...
+  if (!DstTy.isVector())
+    return false;
+
+  // Get the selection mask constant
+  auto SelCst = getIConstantVRegValWithLookThrough(SelReg, MRI);
+  if (!SelCst)
+    return false;
+
+  const uint64_t SelMask = SelCst->Value.getZExtValue();
+  const unsigned NumElements = DstTy.getNumElements();
```
martien-de-jong, line 4755: nit: since we're in generic code, it would be nice to write this arithmetic in terms of APint

llvm/lib/Target/AIE/AIECombinerHelper.cpp
```diff
@@ -4725,3 +4725,86 @@ bool llvm::matchBroadcastExtractToCopy(MachineInstr &MI,
...
+    const bool LowerFromSrc1 = (SelMask == LowerHalfMask);
+
+    MatchInfo = [Src1Reg, Src2Reg, DstReg, DstTy, LowerFromSrc1,
+                 &MRI](MachineIRBuilder &B) {
+      const LLT HalfTy = DstTy.divide(2);
+
+      // Unmerge both sources into halves
+      const Register Src1Lo = MRI.createGenericVirtualRegister(HalfTy);
+      const Register Src1Hi = MRI.createGenericVirtualRegister(HalfTy);
+      B.buildUnmerge({Src1Lo, Src1Hi}, Src1Reg);
```
martien-de-jong, line 4794: Should we check the vector type to be unmergable, e.g. be large enough to span a composite register?

martien-de-jong (approved):
Take nits to taste. If these, especially the all-zero and all-one patterns are created from earlier legalization, I would like to prevent the clutter and do it right there.

llvm/lib/Target/AIE/AIECombinerHelper.cpp
andcarminati: We always build the VSEL for the legal full vector size, that is always a composite register.

llvm/lib/Target/AIE/AIECombinerHelper.cpp
andcarminati: This combiner is instantiated only for aie2p in `aie2p_additional_combines`, so it is fine to use 64bit.

llvm/lib/Target/AIE/AIECombinerHelper.cpp
andcarminati: As we always generate VSEL for full registers, we always have composite registers.

PR #748, opened 2026-01-05
[AIEX] Superreg Rewriter Machine Verifier Bug

F-Stuckmann:
This fixes a machine verifier bug where we insert Copies of dead subregisters (sublanes) and the newly inserted Copies are not marked as Dead defs.


llvm/test/CodeGen/AIE/aie2p/ra/staged-ra-cycle-in-bundle-no-fine-grained-alloc.ll
```diff
@@ -0,0 +1,344 @@
+; NOTE: Assertions have been autogenerated by utils/update_mir_test_checks.py UTC_ARGS: --version 5
+;
+; This file is licensed under the Apache License v2.0 with LLVM Exceptions.
+; See https://llvm.org/LICENSE.txt for license information.
+; SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+;
+; (c) Copyright 2025 Advanced Micro Devices, Inc. or its affiliates
+; RUN: llc -mtriple=aie2p --aie-staged-ra-fine-grained-alloc=false -verify-machineinstrs  --stop-before=aie-superreg-rewrite %s -o - | FileCheck %s
```
andcarminati: The commit message says that `fine-grain RA bug isolation` but we are disabling the fine grained option. Could you please clarify?

andcarminati: Could we use the `not --crash` option to assertively trigger the problem with this commit?

llvm/test/CodeGen/AIE/aie2p/ra/staged-ra-cycle-in-bundle.ll
```diff
@@ -9,6 +9,8 @@
 ; RUN: llc -mtriple=aie2p -verify-machineinstrs %s -o - | FileCheck %s --check-prefix=FINE-GRAINED
 ; RUN: llc -mtriple=aie2p --aie-staged-ra-fine-grained-alloc=false %s -o - | FileCheck %s --check-prefix=COARSE-GRAINED
 
+; Note: Fine-grained alloc is tested in staged-ra-cycle-in-bundle-no-fine-grained-alloc.ll
```
andcarminati: I am a bit confused with `no-fine-grained` part. In both cases we set --aie-staged-ra-fine-grained-alloc=false.

llvm/lib/Target/AIE/AIESuperRegUtils.cpp
```diff
@@ -210,6 +240,68 @@ static void rewriteOperandsToSubRegs(
...
+      LLVM_DEBUG(dbgs() << "  Subregister " << SubRegLI
+                        << " is not empty, skipping\n");
+      continue;
+    }
+
+    LLVM_DEBUG(dbgs() << "  Found empty subregister " << SubRegIdx
+                      << " with VReg " << printReg(VReg, &TRI, 0, &MRI)
+                      << " Range: " << SubRegLI << '\n');
+
+    // Find all copy instructions that define this empty subregister
```
andcarminati: empty subregister?

llvm/lib/Target/AIE/AIESuperRegUtils.cpp
```diff
@@ -210,6 +240,68 @@ static void rewriteOperandsToSubRegs(
...
+    // For COPY instructions, operand 0 is always the destination (def)
+    for (MachineInstr &MI : MRI.reg_nodbg_instructions(VReg)) {
+      if (!MI.isCopy())
```
andcarminati, line 274: if `isLiveIntervalEffectiveEmpty`, how it is possible to have non-copy instructions here? Alternatively, how is it possible to have a copy that reads the operand 0?

llvm/lib/Target/AIE/AIESuperRegUtils.cpp
```diff
@@ -210,6 +240,68 @@ static void rewriteOperandsToSubRegs(
...
+        continue;
+      if (!MI.getOperand(0).isDef())
+        continue;
+      if (MI.getOperand(0).getReg() != VReg)
```
andcarminati, line 278: If `if (MI.getOperand(0).getReg() != VReg)`, then `MI.getOperand(1).getReg() == VReg`? In this case how is it possible for VReg to have empty interval?

llvm/lib/Target/AIE/AIESuperRegUtils.cpp
F-Stuckmann: empty in the sense that the live interval is from R to D of the same slotindex

llvm/test/CodeGen/AIE/aie2p/ra/staged-ra-cycle-in-bundle.ll
F-Stuckmann: I should have added a comment. The staged-ra-cycle-in-bundle-no-fine-grained-alloc.ll specifically tests how far we can go before the machine verifier crashes. 
It now crashes after VirtRegRewriter, which isn't the case in this unit-test.

This PR moves the crash further back and was the main motivation for the PR.

llvm/lib/Target/AIE/AIESuperRegUtils.cpp
F-Stuckmann: yes we are looking at the src VReg here.
What happens is the following: 
- Inlinespiller spills a full 3d/2d register.
- Not all subregs are alive when the inlinespiller spills/reloads.
- During **reloading** we materialize a full register copy. 
- **Superregrewriter** materializes all the subreg copies and in doing so we also materialize dead subreg copies. Which typically would get a dead flag. 

However there is a bug in LiveIntervals, that dead flags aren't updated inside of bundles. 
This happens because Liveinterval isn't empty but goes from R to D of the same slotindex (and thus technically not dead). This is what i call an **effectiveDead LiveInterval**.
So we have to go and manually fix it.

This demonstrates the Error. 
%248 is never used but is missing the dead flag because it is inside of a bundle.
`%184:eds = LDA_DS_SPILL %stack.0, implicit $sp :: (load (s256) from %stack.0, align 4)`
`          dead %250:edcl = COPY %184.sub_dim_count:eds {`
`            internal %249:em_as_32bit = COPY %184.sub_mod:eds`
`            internal %248:edjh = COPY %184.sub_hi_dim_then_sub_dim_stride:eds`
`          }`

So we manually fix this up.

llvm/test/CodeGen/AIE/aie2p/ra/staged-ra-cycle-in-bundle-no-fine-grained-alloc.ll
```diff
@@ -5,9 +5,28 @@
 ; SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
 ;
 ; (c) Copyright 2025 Advanced Micro Devices, Inc. or its affiliates
-; RUN: llc -mtriple=aie2p --aie-staged-ra-fine-grained-alloc=false -verify-machineinstrs  --stop-before=aie-superreg-rewrite %s -o - | FileCheck %s
+; RUN: llc -mtriple=aie2p --aie-staged-ra-fine-grained-alloc=false -verify-machineinstrs --stop-before=prologepilog %s -o - | FileCheck %s
```
andcarminati: As discussed offline, the best here is:
* Stop after aie-superreg-rewrite in both cases, but without the verifier in the first commit.

F-Stuckmann:
Moved to a .mir test that just runs from greedy to superregrewriter.

andcarminati (approved):
LGTM. Thank you for investigating this issue.

PR #749, opened 2026-01-06
[AIE2P] Fix machine verifier complaining about mismatching sizes in pointer copies

khallouh:
We did not consider different address spaces in the type legality check used by the machine verifier to infer the register class of physical registers (e.g. pointers). This led to a machine verifier error complaining about mismatching sizes in copies.


andcarminati (approved):
LGTM.

PR #750, opened 2026-01-06
[NFC] Serialize/Deserialize Register Allocations during Allocation

F-Stuckmann:
This PR enables the stopping and restarting of the Register allocation within stages of the Register Allocation.

This PRs goal is to simplify the debugging and testing of our Register Allocation pipeline by making it modular like the rest of llvm.


llvm/test/tools/UpdateTestChecks/update_mir_test_checks/Inputs/print-registers-first.mir
```diff
@@ -0,0 +1,39 @@
+# NOTE: Assertions have been autogenerated by utils/update_mir_test_checks.py
```
andcarminati: If this file is not a test, why do we have this note about `update_mir_test_checks.py`?

F-Stuckmann:
Added Splits and Required Registers to serialization and deserialization

F-Stuckmann:
We are missing serdes of LiveStacks

PR #751, opened 2026-01-13
[AIE2P] Disallow s20 to GPR bank mapping in case of anyext users

andcarminati:
This can undo the s20 narrowing effect in some kernels, carrying the counter's using GPRs, hurting the postpipeliner.


llvm/lib/Target/AIE/AIEBaseRegisterBankInfo.cpp
```diff
@@ -315,7 +315,6 @@ bool AIEBaseRegisterBankInfo::requiresGPRRegBank(const MachineInstr &MI,
                                                  const MachineRegisterInfo &MRI,
                                                  unsigned Depth) const {
   switch (MI.getOpcode()) {
-  case TargetOpcode::G_ANYEXT:
   case TargetOpcode::G_SEXT:
```
F-Stuckmann, line 318: should we also disallow GPR regbank select for `G_SEXT` and `G_ZEXT`?

F-Stuckmann:
What are the performance numbers?

andcarminati:
Replaced by: https://github.com/Xilinx/llvm-aie/pull/771

PR #753, opened 2026-01-15
[AIEX] Fix SWP Epilogue Lock scheduling

mludevid:
Lock instructions require core stall cycles after memory operations, but this constraint was not enforced in the software pipeliner's epilogue scheduling.
This PR fixes epilogue scheduling by tracking memory access cycles in `AIERegMemEventTracker` and enforcing the required distance before locks.
Additionally, loops containing lock instructions are now rejected from software pipelining since the full scheduling constraints are not implemented for pipelined loop bodies.


llvm/lib/Target/AIE/AIEBaseInstrInfo.cpp
```diff
@@ -1141,6 +1141,12 @@ AIEBaseInstrInfo::isCopyInstrImpl(const MachineInstr &MI) const {
 
 std::unique_ptr<TargetInstrInfo::PipelinerLoopInfo>
 AIEBaseInstrInfo::analyzeLoopForPipelining(MachineBasicBlock *LoopBB) const {
+  // Lock instructions require special scheduling constraints (core stall/resume
+  // cycles) that are not implemented for software pipelined loops.
+  for (const MachineInstr &MI : *LoopBB) {
```
andcarminati: nit: `if (any_of( ... ) ....`

andcarminati: What do you think about creating a utility function is this file:
`llvm/lib/Target/AIE/Utils/AIEMachineBasicBlockUtils.h`

As: `bool hasLockInstructions(...)...`

So you can reuse.

llvm/test/CodeGen/AIE/aie2p/schedule/postpipeliner/epilogue-lock-after-store.mir
```diff
@@ -0,0 +1,87 @@
+# NOTE: Assertions have been autogenerated by utils/update_mir_test_checks.py UTC_ARGS: --version 5
+# This file is licensed under the Apache License v2.0 with LLVM Exceptions.
+# See https://llvm.org/LICENSE.txt for license information.
+# SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+#
+# (c) Copyright 2026 Advanced Micro Devices, Inc. or its affiliates
+
+# RUN: llc --mtriple=aie2p -O2 --run-pass=postmisched --aie-loop-min-tripcount=16 --aie-addrspace-none-is-safe %s -o - | FileCheck %s
```
andcarminati: nit: remove `--aie-addrspace-none-is-safe`

llvm/lib/Target/AIE/AIERegMemEventTracker.cpp
```diff
@@ -121,6 +133,16 @@ AIERegMemEventTracker::getSafeOperandsDistance(const MachineInstr &MI) const {
     }
   }
 
+  // Lock instructions stall the core. All preceding memory operations must
+  // complete before the core stalls.
+  if (TII->isLock(MI.getOpcode())) {
+    const int CoreStallCycle = TII->getCoreStallCycleAfterLock();
+    const int MIStallCycle = CoreStallCycle - 1;
```
andcarminati, line 140: `- then +` ? If this helps to understand you can write a comment.

andcarminati:
Nice and clean implementation. I left some nits.

llvm/lib/Target/AIE/AIEBaseInstrInfo.cpp
mludevid: Since we rely on `isLock` which is a function of `AIEBaseInstrInfo` I don't think it makes sense to move this to `AIEMachineBasicBlockUtils.h`.
Regarding `any_of`, I'm not sure it makes the code more readable in this case:
```
  for (const MachineInstr &MI : *LoopBB) {
    if (isLock(MI.getOpcode()))
      return nullptr;
  }
```
vs
```
  if (any_of(LoopBB, [this](const MachineInstr &MI) {
        return isLock(MI.getOpcode());
      })) {
    return nullptr;
  }
```
What is your preference?

andcarminati: Second one!

mludevid: I created `hasLockInstructions` in `AIEBaseInstrInfo`.

andcarminati (dismissed):
LGTM.

andcarminati (approved):

PR #754, opened 2026-01-15
[AIEX] Minimize Stackslots by Decomposition and only keeping defined parts of the Stack

F-Stuckmann:
This PR introduces a stack optimization pipeline.
The core idea is to not spill undef subregisters. 

- **AIESpillDecomposition** – Decomposes composite spill instructions (e.g., 256-byte DM spills) into smaller
    subreg spills (e.g., 128-byte CM spills) when only some subregs are live. Uses liveness analysis with
    LaneBitmask to determine which subregs actually need spilling.
 - **AIEStackSlotMinimization** – Analyzes which subregs of spilled registers are actually accessed. When only a
    subset is used, it recreates stack slots with smaller sizes. Also separates non-overlapping registers
    sharing the same frame index into independent slots.
 -  **Frame Object Ordering** - Added orderFrameObjects hook in AIEBaseFrameLowering to sort stack objects by alignment (highest first), reducing wasted padding.

Minor Bug fixes along the way: 
  - Moved isLoadFromStackSlot/isStoreToStackSlot implementations from AIE2/AIE2P to a generic AIEBaseInstrInfo
    using MachineMemOperand validation.
  - Added missing intrinsics to AIE2P frame index lowering.
  
Performance numbers: I am still looking why Frame Object ordering is screwing up.


llvm/lib/Target/AIE/AIEBaseInstrInfo.cpp
```diff
@@ -1529,3 +1526,80 @@ bool AIEBaseInstrInfo::areMemAccessesTriviallyDisjoint(
...
+  // Validate via MachineMemOperand (strongest check)
+  // This ensures we're actually accessing a FixedStackSlot, not just
+  // any instruction that happens to have a FI operand
+  for (const auto *MMO : MI.memoperands()) {
+    if (MMO->isLoad() &&
+        isa_and_nonnull<FixedStackPseudoSourceValue>(MMO->getPseudoValue())) {
+      FrameIndex = MI.getOperand(1).getIndex();
+      return MI.getOperand(0).getReg();
+    }
+  }
```
martien-de-jong, line 1576: I hope that stack loads have exactly one MMO which is a Load, and I would feel comforted if that were asserted.
I'm also not sure where these functions are used and whether it requires completeness in identifying all stack loads, whether stack loads from user stack objects always have their MMO's set correctly.
I'm also convinced that there are quintillion combinations of stack use, of which only a fraction is tested in practice and that we should hence be cautious. It is great to have this unified, granting us the possibility to concentrate all sanity checking in one place.

llvm/lib/Target/AIE/aie2p/AIE2PRegisterInfo.cpp
```diff
@@ -170,6 +170,8 @@ bool AIE2PRegisterInfo::eliminateFrameIndex(MachineBasicBlock::iterator II,
 
     // Note that LDB path does not support SPILL instructions
 
+  case AIE2P::LDA_dms_lda_spill:
+  case AIE2P::ST_dms_sts_spill:
```
martien-de-jong, line 174: nit: you call these intrinsics in your commit comment

llvm/test/CodeGen/AIE/aie2p/ra/spill-decomposition-3d.mir
```diff
@@ -0,0 +1,240 @@
...
+# See https://llvm.org/LICENSE.txt for license information.
+# SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+#
+# (c) Copyright 2026 Advanced Micro Devices, Inc. or its affiliates
+
+
+# RUN: llc -mtriple=aie2p --run-pass=none -o - -verify-machineinstrs %s | FileCheck %s
+
+---
+name:            _Z11conv2d_bf16ILh1EL5act_t0E8bfloat16S1_S1_Lb0ELb0ELb1ELb0ELb0EEvPT1_PT2_PT3_R18conv2d_bf16_paramsS7_S7_
```
martien-de-jong: I think we can use a readable name

martien-de-jong:
Just a convention, 'Adding baseline tests' would label a commit adding unit tests that you expect to change.
"Adding unit tests" would then add tests that you observed were missing, containing behaviour you want to keep. 
As a reviewer, I usually just skim commits of the first category, since I will see the changes in the functional commits.

llvm/lib/Target/AIE/AIE2TargetMachine.cpp
```diff
@@ -192,6 +196,12 @@ bool AIE2PassConfig::addRegAssignAndRewriteOptimized() {
 }
 
 void AIE2PassConfig::addPostRewrite() {
+  if (EnableStackMinimization) {
```
martien-de-jong, line 199: Probably not, but checking anyway, don't we have an AIEBasePassConfig where we can add this?

llvm/lib/Target/AIE/AIESpillDecomposition.cpp
```diff
@@ -0,0 +1,382 @@
...
+  bool HasLiveSubRegs = false;
+  /// Register state flags for store emission (e.g., Kill).
+  unsigned StoreFlags = 0;
+  /// Pre-computed info for each subreg of the composite register.
+  SmallVector<SubRegInfo, 8> SubRegs;
+
+  SpillPoint() = default;
+
+  /// Construct from a spill/reload instruction.
+  SpillPoint(MachineInstr &MI, int FI, bool IsStore,
```
martien-de-jong: nit: IsStore can easily be derived from MI

llvm/lib/Target/AIE/AIESpillDecomposition.cpp
```diff
@@ -0,0 +1,382 @@
...
+
+private:
+  const TargetRegisterInfo *TRI = nullptr;
+  const AIEBaseInstrInfo *TII = nullptr;
+  MachineFrameInfo *MFI = nullptr;
+  MachineRegisterInfo *MRI = nullptr;
+
+  DenseMap<int, FrameIndexSpillInfo> FIInfoMap;
+
+  /// Collect all spill/reload instructions, compute per-subreg liveness.
```
martien-de-jong: I think you invented "spill point" as a short hand for spill/reload instruction, and I like it.

llvm/lib/Target/AIE/AIESpillDecomposition.cpp
```diff
@@ -0,0 +1,382 @@
...
+  /// \return True if any instructions were transformed.
+  bool transformFunction();
+
+  /// Compute live registers at \p MI.
+  void computeLiveness(LivePhysRegs &LiveRegs, MachineBasicBlock &MBB,
+                       MachineInstr &MI, bool IsStore) const;
+
+  /// Emit a single subreg spill or reload instruction.
+  void emitDecomposedInstr(const SpillPoint &SP, const SubRegInfo &SI,
+                           bool IsStore) const;
```
martien-de-jong: Sounds like transformSpillPoint()

llvm/lib/Target/AIE/AIESpillDecomposition.cpp
```diff
@@ -0,0 +1,382 @@
...
+    if (I > 0)
+      OS << ", ";
+    OS << SP.SubRegs[I];
+  }
+  OS << "]";
+  return OS;
+}
+
+/// Build a SubRegInfo from expansion info.
+static SubRegInfo
```
martien-de-jong: SubRegInfo constructor?

llvm/lib/Target/AIE/AIESpillDecomposition.cpp
```diff
@@ -0,0 +1,382 @@
...
+  DenseMap<int, FrameIndexSpillInfo> FIInfoMap;
+
+  /// Collect all spill/reload instructions, compute per-subreg liveness.
+  void analyzeFunction(MachineFunction &MF);
+
+  /// Emit decomposed instructions for each collected spill/reload.
+  /// \return True if any instructions were transformed.
+  bool transformFunction();
+
+  /// Compute live registers at \p MI.
```
martien-de-jong: Please detail before or after MI.

llvm/lib/Target/AIE/AIESpillDecomposition.cpp
```diff
@@ -0,0 +1,382 @@
...
+
+} // end anonymous namespace
+
+SpillPoint::SpillPoint(MachineInstr &MI, int FI, bool IsStore,
+                       const LivePhysRegs &LiveRegs,
+                       const TargetRegisterInfo &TRI,
+                       const AIEBaseInstrInfo &TII,
+                       const MachineRegisterInfo &MRI)
+    : MI(&MI), FI(FI), IsStore(IsStore),
+      StoreFlags(MI.getOperand(0).isKill() ? RegState::Kill : 0) {
```
martien-de-jong: You seem to do your own liveness analysis in the right direction. Could we use that instead of isKill() ?

llvm/lib/Target/AIE/AIESpillDecomposition.cpp
```diff
@@ -0,0 +1,382 @@
...
+    LiveRegs.addLiveOuts(MBB);
+    auto End = MachineBasicBlock::iterator(MI).getReverse();
+    for (MachineInstr &LiveMI : make_range(MBB.rbegin(), End))
+      LiveRegs.stepBackward(LiveMI);
+  }
+}
+
+void AIESpillDecomposition::emitDecomposedInstr(const SpillPoint &SP,
+                                                const SubRegInfo &SI,
+                                                bool IsStore) const {
```
martien-de-jong: Isn't SP.IsStore sufficient?

llvm/lib/Target/AIE/AIESpillDecomposition.cpp
```diff
@@ -0,0 +1,382 @@
...
+
+      const int FI = IsStore ? StoreFI : LoadFI;
+      if (FI < 0 || FI >= MFI->getObjectIndexEnd() ||
+          MFI->isDeadObjectIndex(FI) || !MFI->isSpillSlotObjectIndex(FI))
+        continue;
+      if (TII->getSpillPseudoExpandInfo(*TRI, MI).empty())
+        continue;
+
+      LivePhysRegs LiveRegs(*TRI);
+      computeLiveness(LiveRegs, MBB, MI, IsStore);
```
martien-de-jong: This is quadratic && complicated. 
You could make two traversals, one for the stores in one direction, one for the loads in the other direction and
maintain the liveness information on these traversals that collect them. Then we don't need to do a liveness traversal for every spillpoint.

llvm/lib/Target/AIE/AIESpillDecomposition.cpp
```diff
@@ -0,0 +1,382 @@
...
+        emitDecomposedInstr(SP, SI, /*IsStore=*/true);
+      }
+      ToErase.push_back(SP.MI);
+    }
+
+    for (const SpillPoint &SP : Info.Reloads) {
+      bool Emitted = false;
+      for (const SubRegInfo &SI : SP.SubRegs) {
+        const auto &Stored = StoredSubRegsPerOffset.lookup(SI.Offset);
+        if (Stored.empty())
```
martien-de-jong: Check: this is the main reason why we want to deal with all Stores first, which is why we have them in separate lists.

llvm/lib/Target/AIE/AIESpillDecomposition.cpp
```diff
@@ -0,0 +1,382 @@
...
+
+  SpillPoint() = default;
+
+  /// Construct from a spill/reload instruction.
+  SpillPoint(MachineInstr &MI, int FI, bool IsStore,
+             const LivePhysRegs &LiveRegs, const TargetRegisterInfo &TRI,
+             const AIEBaseInstrInfo &TII, const MachineRegisterInfo &MRI);
+};
+
+/// Per-FI tracking of spills and reloads.
```
martien-de-jong: I think this is a perfect place to explain why we separate the spills and the reloads. 
Also, like we discussed, mentioning the fact that the algorithm collects all live lanes over all store instructions relating to a particular FrameIndex.

llvm/lib/Target/AIE/AIESpillDecomposition.cpp
```diff
@@ -0,0 +1,382 @@
...
+  });
+}
+
+bool AIESpillDecomposition::transformFunction() {
+  LLVM_DEBUG(dbgs() << "=== Phase 2: Transformation ===\n");
+
+  SmallVector<MachineInstr *, 16> ToErase;
+
+  for (auto &[FI, Info] : FIInfoMap) {
+    // Track which subregs were stored at each offset (local to transformation)
```
martien-de-jong: I don't know what 'local to transformation means' It could be 'for this FrameIndex?'

llvm/lib/Target/AIE/AIESpillDecomposition.cpp
```diff
@@ -0,0 +1,382 @@
...
+  if (!MF.getProperties().hasProperty(
+          MachineFunctionProperties::Property::TracksLiveness)) {
+    LLVM_DEBUG(dbgs() << "Liveness not tracked, skipping\n");
+    return false;
+  }
+
+  bool Changed = false;
+  do {
+    analyzeFunction(MF);
+  } while (transformFunction() && (Changed = true));
```
martien-de-jong: Please discuss rationale and convergence of this fixpoint loop. E.g. (just guessing) livein of successor blocks may change, maximum iteration count is loop depth + 1

llvm/lib/Target/AIE/AIEStackSlotMinimization.cpp
```diff
@@ -0,0 +1,349 @@
...
+  };
+  SmallVector<PendingLoad, 32> PendingLoads;
+
+  // Process all instructions: stores immediately, loads deferred
+  for (MachineBasicBlock &MBB : MF) {
+    for (MachineInstr &MI : MBB) {
+      int StoreFI = -1;
+      int LoadFI = -1;
+      const bool IsStore = TII->isStoreToStackSlot(MI, StoreFI) != 0;
+      const bool IsLoad = !IsStore && TII->isLoadFromStackSlot(MI, LoadFI) != 0;
```
martien-de-jong: I've seen this code before. Could we factor this by passing a lambda to a traversal function?

llvm/lib/Target/AIE/AIEBaseInstrInfo.cpp
```diff
@@ -1525,3 +1526,80 @@ bool AIEBaseInstrInfo::areMemAccessesTriviallyDisjoint(
 
   return CheckOverlapping(ObjOffsetA, ObjOffsetB);
 }
+
+Register AIEBaseInstrInfo::isLoadFromStackSlot(const MachineInstr &MI,
+                                               int &FrameIndex) const {
```
andcarminati: This is a nice refactor. However, I see that we can move even further. The two functions are almost identical. Could we have a common refactor called `isStackSlotMemoryAccess(bool IsLoad...)` ?

llvm/lib/Target/AIE/aie2p/AIE2PRegisterInfo.cpp
andcarminati: True, this should not be confused with intrinsics. Also, intrinsics will exist only until instruction selection.

andcarminati: Should this also be covered in `/scratch/llvm-aie/llvm/test/CodeGen/AIE/aie2p/eliminate-frame-index.mir` ?

llvm/test/CodeGen/AIE/aie2p/ra/spill-decomposition-3d.mir
```diff
@@ -0,0 +1,240 @@
...
+alignment:       16
+exposesReturnsTwice: false
```
andcarminati: Do we need this flag?

llvm/test/CodeGen/AIE/aie2p/ra/spill-decomposition-3d.mir
```diff
@@ -0,0 +1,240 @@
...
+#
+# This file is licensed under the Apache License v2.0 with LLVM Exceptions.
+# See https://llvm.org/LICENSE.txt for license information.
+# SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+#
+# (c) Copyright 2026 Advanced Micro Devices, Inc. or its affiliates
+
+
+# RUN: llc -mtriple=aie2p --run-pass=none -o - -verify-machineinstrs %s | FileCheck %s
+
```
andcarminati: Maybe a comment would be nice!

llvm/lib/Target/AIE/AIESpillDecomposition.cpp
```diff
@@ -0,0 +1,382 @@
...
+// smaller subreg spills (e.g., CM spills) when only some subregs are live.
+// This reduces stack usage by eliminating spills of undefined subregs.
+//
+// For example, if a DM register is spilled but only CML is live, we replace
+// the 256-byte DM spill with a 128-byte CM spill.
+//
+// The pass uses a two-phase approach:
+// 1. Analysis Phase: Collect all spill/reload info, compute liveness,
+//    pre-compute per-subreg decomposition details
+// 2. Transformation Phase: Emit decomposed instructions using cached info
```
andcarminati: Is cache a good name here?

llvm/lib/Target/AIE/AIESpillDecomposition.cpp
```diff
@@ -0,0 +1,382 @@
...
+#include "llvm/MC/MCRegister.h"
+#include "llvm/Support/Debug.h"
+
+using namespace llvm;
+
+#define DEBUG_TYPE "aie-spill-decomposition"
+
+namespace {
+
+constexpr unsigned BitsPerByte = 8;
```
andcarminati: `CHAR_BIT` ?

llvm/lib/Target/AIE/AIESpillDecomposition.cpp
```diff
@@ -0,0 +1,382 @@
...
+constexpr unsigned MinMemSizeBytes = 4;
+
+/// Pre-computed info for a single subreg within a spill/reload.
+struct SubRegInfo {
+  /// The physical subreg register.
+  MCRegister Reg;
+  /// Subreg index (0 for full register).
+  unsigned SubRegIndex = 0;
+  /// Opcode for the decomposed spill/reload instruction.
+  unsigned OpCode = 0;
```
andcarminati: nit `Opcode` (see comment)?

llvm/lib/Target/AIE/AIESpillDecomposition.cpp
```diff
@@ -0,0 +1,382 @@
...
+
+  /// Construct from a spill/reload instruction.
+  SpillPoint(MachineInstr &MI, int FI, bool IsStore,
+             const LivePhysRegs &LiveRegs, const TargetRegisterInfo &TRI,
+             const AIEBaseInstrInfo &TII, const MachineRegisterInfo &MRI);
+};
+
+/// Per-FI tracking of spills and reloads.
+struct FrameIndexSpillInfo {
+  SmallVector<SpillPoint, 4> Spills;
```
andcarminati: Check: In this first step, all `SpillPoint`'s will have the same `int FI;`, but this will going to change.

llvm/lib/Target/AIE/AIESpillDecomposition.cpp
```diff
@@ -0,0 +1,382 @@
...
+  SmallVector<SpillPoint, 4> Reloads;
```
andcarminati: The `bool IsStore` information is tracked 3 times, field, instruction and here. Should we simplify and derive it from a single place, let's say those `SmallVector`s?

llvm/lib/Target/AIE/AIESpillDecomposition.cpp
```diff
@@ -0,0 +1,382 @@
...
+char AIESpillDecomposition::ID = 0;
+char &llvm::AIESpillDecompositionID = AIESpillDecomposition::ID;
+
+INITIALIZE_PASS(AIESpillDecomposition, DEBUG_TYPE, "AIE Spill Decomposition",
+                false, false)
+
+void AIESpillDecomposition::computeLiveness(LivePhysRegs &LiveRegs,
+                                            MachineBasicBlock &MBB,
+                                            MachineInstr &MI,
+                                            bool IsStore) const {
```
andcarminati: IsStore is also redundant here.

llvm/lib/Target/AIE/AIESpillDecomposition.cpp
```diff
@@ -0,0 +1,382 @@
...
+      if (FI < 0 || FI >= MFI->getObjectIndexEnd() ||
+          MFI->isDeadObjectIndex(FI) || !MFI->isSpillSlotObjectIndex(FI))
+        continue;
+      if (TII->getSpillPseudoExpandInfo(*TRI, MI).empty())
+        continue;
+
+      LivePhysRegs LiveRegs(*TRI);
+      computeLiveness(LiveRegs, MBB, MI, IsStore);
+
+      SpillPoint SP(MI, FI, IsStore, LiveRegs, *TRI, *TII, *MRI);
```
andcarminati: SP looks like stack pointer ;-)

llvm/lib/Target/AIE/AIESpillDecomposition.cpp
```diff
@@ -0,0 +1,382 @@
...
+                        : 0);
+  SI.IsLive = !LiveRegs.available(MRI, SI.Reg);
+  return SI;
+}
+
+} // end anonymous namespace
+
+SpillPoint::SpillPoint(MachineInstr &MI, int FI, bool IsStore,
+                       const LivePhysRegs &LiveRegs,
+                       const TargetRegisterInfo &TRI,
```
andcarminati: Do you need TRI here? you can always get it from  `MRI.getTargetRegisterInfo()`.

llvm/lib/Target/AIE/AIESpillDecomposition.cpp
```diff
@@ -0,0 +1,382 @@
...
+      const bool IsLoad = !IsStore && TII->isLoadFromStackSlot(MI, LoadFI) != 0;
+
+      if (!IsStore && !IsLoad)
+        continue;
+
+      const int FI = IsStore ? StoreFI : LoadFI;
+      if (FI < 0 || FI >= MFI->getObjectIndexEnd() ||
+          MFI->isDeadObjectIndex(FI) || !MFI->isSpillSlotObjectIndex(FI))
+        continue;
+      if (TII->getSpillPseudoExpandInfo(*TRI, MI).empty())
```
andcarminati: Ok, this turns this pass target dependent, right?

llvm/lib/Target/AIE/AIESpillDecomposition.cpp
```diff
@@ -0,0 +1,382 @@
...
+        continue;
+
+      LivePhysRegs LiveRegs(*TRI);
+      computeLiveness(LiveRegs, MBB, MI, IsStore);
+
+      SpillPoint SP(MI, FI, IsStore, LiveRegs, *TRI, *TII, *MRI);
+      FrameIndexSpillInfo &Info = FIInfoMap[FI];
+
+      auto &List = IsStore ? Info.Spills : Info.Reloads;
```
andcarminati: Could we delegate to Info the proper list selection, like `Info.add()`? Also is this is a vector.

llvm/lib/Target/AIE/AIESpillDecomposition.cpp
```diff
@@ -0,0 +1,382 @@
...
+
+INITIALIZE_PASS(AIESpillDecomposition, DEBUG_TYPE, "AIE Spill Decomposition",
+                false, false)
+
+void AIESpillDecomposition::computeLiveness(LivePhysRegs &LiveRegs,
+                                            MachineBasicBlock &MBB,
+                                            MachineInstr &MI,
+                                            bool IsStore) const {
+  if (IsStore) {
+    LiveRegs.addLiveIns(MBB);
```
andcarminati: +1 to @martien-de-jong's comment about redoing this from scratch every time. We can build/update this incrementally.

llvm/lib/Target/AIE/AIESpillDecomposition.cpp
```diff
@@ -0,0 +1,382 @@
...
+        Emitted = true;
+        emitDecomposedInstr(SP, SI, /*IsStore=*/false);
+      }
+      if (Emitted)
+        ToErase.push_back(SP.MI);
+    }
+  }
+
+  LLVM_DEBUG(dbgs() << "  Erasing " << ToErase.size() << " instructions\n");
+  for (MachineInstr *MI : ToErase)
```
andcarminati: I would like to remove every instruction from this list before erasing.

llvm/lib/Target/AIE/AIESpillDecomposition.cpp
```diff
@@ -0,0 +1,382 @@
...
+bool AIESpillDecomposition::runOnMachineFunction(MachineFunction &MF) {
+  LLVM_DEBUG(dbgs() << "*** AIE Spill Decomposition: " << MF.getName()
+                    << " ***\n");
+
+  TRI = MF.getSubtarget().getRegisterInfo();
+  TII = static_cast<const AIEBaseInstrInfo *>(MF.getSubtarget().getInstrInfo());
+  MFI = &MF.getFrameInfo();
+  MRI = &MF.getRegInfo();
+
+  if (!MF.getProperties().hasProperty(
```
andcarminati: Could we also skip if the function has no fixed objects, like something:

```
 if(MF.getFrameInfo().MFI.getNumFixedObjects() == 0)
    return false;
```

So we don't need to walk through all blocks...

llvm/lib/Target/AIE/AIESpillDecomposition.cpp
```diff
@@ -0,0 +1,382 @@
...
+    }
+
+    for (const SpillPoint &SP : Info.Reloads) {
+      bool Emitted = false;
+      for (const SubRegInfo &SI : SP.SubRegs) {
+        const auto &Stored = StoredSubRegsPerOffset.lookup(SI.Offset);
+        if (Stored.empty())
+          continue;
+
+        // Check if any stored subreg at this offset overlaps with SI's lanes.
```
andcarminati: Could you please document a bit more this part?

llvm/lib/Target/AIE/AIEBaseFrameLowering.cpp
```diff
@@ -370,3 +370,207 @@ void AIEBaseFrameLowering::emitEpilogue(MachineFunction &MF,
...
+  Align Alignment;
+};
+
+/// Sort by alignment descending, then size descending.
+/// Used for the standard high-alignment-first ordering.
+struct AlignmentDescending {
+  bool operator()(const FrameObject &A, const FrameObject &B) const {
+    if (A.Alignment != B.Alignment)
+      return A.Alignment > B.Alignment;
+    return A.Size > B.Size;
```
martien-de-jong: Perhaps make it a stable ordering by ordering Index?

llvm/lib/Target/AIE/AIEBaseFrameLowering.cpp
```diff
@@ -370,3 +370,207 @@ void AIEBaseFrameLowering::emitEpilogue(MachineFunction &MF,
...
+  //
+  // High-alignment objects (alignment == MaxAlign):
+  //   These CANNOT be useful fillers because placing them requires padding
+  //   to MaxAlign first. The cost would be:
+  //     pad(CurrentOffset, MaxAlign) + 0 = InitialGap
+  //   This equals the baseline cost, so there is no benefit.
+  //   After we place fillers and reach MaxAlign, these objects can be placed
+  //   with zero additional padding since we are already MaxAlign-aligned.
+  SmallVector<FrameObject> FillerCandidates;
+  SmallVector<FrameObject> HighAlignObjects;
```
martien-de-jong: I'm not sure whether SmallVector is appropriate here. Why not just std::vector, pre-reserving as Objects.size() ?

llvm/lib/Target/AIE/AIEBaseFrameLowering.cpp
```diff
@@ -370,3 +370,207 @@ void AIEBaseFrameLowering::emitEpilogue(MachineFunction &MF,
...
+    const int64_t Size = MFI.getObjectSize(FI);
+    if (Size == 0)
+      return; // Variable-sized object; cannot reorder
+
+    const Align ObjAlign = MFI.getObjectAlign(FI);
+    Objects.push_back({FI, Size, ObjAlign});
+    MaxAlign = std::max(MaxAlign, ObjAlign);
+  }
+
+  // Compute starting offset (end of callee-saved area).
```
martien-de-jong: Discussed offline: Direction is not an issue, since we only need the next block in terms of stack alignment

llvm/lib/Target/AIE/AIEBaseFrameLowering.cpp
```diff
@@ -370,3 +370,207 @@ void AIEBaseFrameLowering::emitEpilogue(MachineFunction &MF,
...
+
+    for (size_t I = 0; I < FillerCandidates.size(); ++I) {
+      const FrameObject &Obj = FillerCandidates[I];
+
+      // Compute cost with this filler:
+      //
+      //   CurrentOffset                           MaxAlign boundary
+      //        v                                         v
+      //   [....|----PadToObj----|==Obj==|----GapAfter----|...]
+      //
```
martien-de-jong: Perhaps indicate that at the start, the [...| before CurrentOffset comprises an integral multiple of the stack alignment

llvm/lib/Target/AIE/AIEBaseFrameLowering.cpp
```diff
@@ -370,3 +370,207 @@ void AIEBaseFrameLowering::emitEpilogue(MachineFunction &MF,
...
+    // Find the best filler candidate:
+    // 1. Must reduce total cost (CostWithFiller < CostWithoutFiller)
+    // 2. Prefer minimal padding (object fits without wasting bytes)
+    const FrameObject *BestCandidate = nullptr;
+    int64_t BestPadding = INT64_MAX;
+    int64_t BestSize = 0;
+    size_t BestIdx = 0;
+
+    for (size_t I = 0; I < FillerCandidates.size(); ++I) {
+      const FrameObject &Obj = FillerCandidates[I];
```
martien-de-jong: You could Obj as an iterator, and replace BestIdx and I with Best iterator.

llvm/lib/Target/AIE/AIEStackSlotMinimization.cpp
```diff
@@ -0,0 +1,349 @@
...
+                      << R.Size << " align=" << R.Alignment.value() << "\n");
+  });
+}
+
+bool AIEStackSlotMinimization::replaceInstructions() {
+  LLVM_DEBUG(dbgs() << "Replacing instructions\n");
+
+  SmallVector<MachineInstr *, 32> ToErase;
+
+  forEachReplacement([&](int FI, ReplacementSlot &R) {
```
andcarminati: Edit MMO and FI operand?

llvm/lib/Target/AIE/AIEBaseInstrInfo.cpp
F-Stuckmann: Stack Users always have their MMOs correctly set. They originate in the Inlinespiller and lateron in the spill decomposition I ensure that we do not invalidate the MMOs.

llvm/lib/Target/AIE/AIEBaseInstrInfo.cpp
F-Stuckmann: that wasn't as bad as i was expecting

llvm/test/CodeGen/AIE/aie2p/ra/spill-decomposition-3d.mir
F-Stuckmann: I also cleaned up the flags of the other unit-tests.

llvm/lib/Target/AIE/AIESpillDecomposition.cpp
F-Stuckmann: I thought about it but i was getting confused about what the constructor was actually doing

llvm/lib/Target/AIE/AIEStackSlotMinimization.cpp
```diff
@@ -0,0 +1,349 @@
...
+//===----------------------------------------------------------------------===//
+
+/// Cached info about a spill/reload instruction to avoid recomputation.
+struct InstrInfo {
+  /// The original spill or reload instruction.
+  MachineInstr *MI;
+  /// The physical register being spilled/reloaded.
+  MCPhysReg PhysReg;
+  /// True if this is a store (spill), false if load (reload).
+  bool IsStore;
```
andcarminati: This could be a method that uses MI to infer if this is a store.

andcarminati: Also for the kill. We have a method `killsRegister`.

llvm/lib/Target/AIE/AIEStackSlotMinimization.cpp
```diff
@@ -0,0 +1,349 @@
...
+
+/// A minimized stack slot replacing part of an original spill slot.
+/// Groups instructions that access overlapping register lanes.
+struct ReplacementSlot {
+  /// MMO offset within the original spill slot.
+  int64_t Offset;
+  /// Combined lane mask of all registers stored to this slot.
+  LaneBitmask Mask;
+  /// Size in bytes of the replacement stack slot.
+  unsigned Size;
```
andcarminati: Do you have a reason to use 64 bits for Offset and 32 bits for Size? Is `size_t` better?

llvm/lib/Target/AIE/AIEStackSlotMinimization.cpp
```diff
@@ -0,0 +1,349 @@
...
+  SmallVector<ReplacementSlot, 2> Replacements;
+
+  ReplacementSlot *findForMask(int64_t Offset, LaneBitmask Mask) {
+    for (ReplacementSlot &R : Replacements)
+      if (R.Offset == Offset && (R.Mask & Mask).any())
+        return &R;
+    return nullptr;
+  }
+
+  ReplacementSlot &addStore(int64_t Offset, LaneBitmask Mask, unsigned Size,
```
andcarminati: Maybe a comment in this function. I can't see relation to a store.

llvm/lib/Target/AIE/AIEStackSlotMinimization.cpp
```diff
@@ -0,0 +1,349 @@
...
+  ReplacementSlot(int64_t Offset, LaneBitmask Mask, unsigned Size,
+                  Align Alignment, const TargetRegisterClass *RC)
+      : Offset(Offset), Mask(Mask), Size(Size), Alignment(Alignment), RC(RC) {}
+};
+
+/// Tracks all replacement slots for one original frame index.
+struct OriginalSlot {
+  SmallVector<ReplacementSlot, 2> Replacements;
+
+  ReplacementSlot *findForMask(int64_t Offset, LaneBitmask Mask) {
```
andcarminati: The name is unclear, could you please clarify?

llvm/lib/Target/AIE/AIESpillDecomposition.cpp
F-Stuckmann: yes, but I could introduce this hook to upstream.

llvm/test/CodeGen/AIE/disassembler.ll
```diff
@@ -1,3 +1,4 @@
+; NOTE: Assertions have been autogenerated by utils/update_llc_test_checks.py UTC_ARGS: --version 5
```
andcarminati: Why do we have this change?

llvm/test/CodeGen/AIE/aie2p/spill/dim_spill.mir
```diff
@@ -4,7 +4,7 @@
 # See https://llvm.org/LICENSE.txt for license information.
 # SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
 #
-# (c) Copyright 2024-2025 Advanced Micro Devices, Inc. or its affiliates
+# (c) Copyright 2024-2026 Advanced Micro Devices, Inc. or its affiliates
```
andcarminati, line 7: Why do we have this change?

llvm/test/CodeGen/AIE/aie2p/spill-prologepilog.mir
```diff
@@ -150,27 +150,27 @@ body:             |
     liveins: $p1, $dc5, $bmll0, $q0, $wl0, $e0
 
     ; CHECK-LABEL: name: test_mix
-    ; CHECK: stackSize: 320
+    ; CHECK: stackSize: 192
```
andcarminati: Nice =)

llvm/lib/Target/AIE/AIEBaseFrameLowering.cpp
```diff
@@ -370,3 +370,207 @@ void AIEBaseFrameLowering::emitEpilogue(MachineFunction &MF,
...
+namespace {
+
+/// Compute padding needed to align an offset to a given alignment.
+inline int64_t computePadding(int64_t Offset, Align Alignment) {
+  return alignTo(Offset, Alignment) - Offset;
+}
+
+struct FrameObject {
+  int Index;
+  int64_t Size;
```
andcarminati: Signed size?

llvm/lib/Target/AIE/AIEBaseFrameLowering.cpp
andcarminati: +1

llvm/lib/Target/AIE/AIESpillDecomposition.cpp
F-Stuckmann: we can have zero Fixed Stack objects but still have stack objects. 
This should do the trick:
```
if (MFI->getNumObjects() == 0) 
      continue;
```

llvm/lib/Target/AIE/AIE2TargetMachine.cpp
F-Stuckmann: We do not implement addPostRewrite for AIE1, so this is the lowest point to add it.

llvm/lib/Target/AIE/AIESpillDecomposition.cpp
F-Stuckmann: this and we want to know what Subregs we actually have to load.

llvm/lib/Target/AIE/AIESpillDecomposition.cpp
F-Stuckmann: In SpillDecomposition it is actually const!

llvm/lib/Target/AIE/AIESpillDecomposition.cpp
F-Stuckmann: I am now properly reusing all the flags of the original Spill Store/Load

llvm/lib/Target/AIE/AIESpillDecomposition.cpp
F-Stuckmann: We don't need it. I deleted it

llvm/lib/Target/AIE/AIEStackSlotMinimization.cpp
F-Stuckmann: yep it is cleaner!

llvm/test/CodeGen/AIE/aie2p/spill/dim_spill.mir
F-Stuckmann: artifact, it is removed now

llvm/test/CodeGen/AIE/disassembler.ll
F-Stuckmann: a screw up from my scripts, we should not be updating dissasembler

llvm/lib/Target/AIE/AIEStackSlotMinimization.cpp
F-Stuckmann: Afuera, This can be simplified

llvm/lib/Target/AIE/AIEBaseInstrInfo.cpp
```diff
@@ -1529,3 +1530,69 @@ bool AIEBaseInstrInfo::areMemAccessesTriviallyDisjoint(
...
+      return 0;
+  }
+
+  const TargetSubtargetInfo &ST = MI.getMF()->getSubtarget();
+  const auto &TRI =
+      *static_cast<const AIEBaseRegisterInfo *>(ST.getRegisterInfo());
+
+  const Register SPReg = TRI.getStackPointerRegister();
+  if (!MI.readsRegister(SPReg, &TRI))
+    return 0;
```
martien-de-jong, line 1554: I'm wondering whether we always have a physical SP read for e.g. spill pseudos.

llvm/lib/Target/AIE/AIESpillSlotOptimization.cpp
```diff
@@ -0,0 +1,514 @@
...
+                const TargetRegisterInfo &TRI, const LivePhysRegs &LiveRegs,
+                const MachineRegisterInfo &MRI) {
+  const MCRegister SubReg =
+      EI.SubRegIndex ? TRI.getSubReg(Reg, EI.SubRegIndex) : Reg.asMCReg();
+
+  // Compute memory size: use explicit size if provided, otherwise derive from
+  // subreg index size (converting bits to bytes).
+  unsigned MemSize = EI.MemSize;
+  if (!MemSize && EI.SubRegIndex)
+    MemSize = divideCeil(TRI.getSubRegIdxSize(EI.SubRegIndex), BitsPerByte);
```
martien-de-jong: 20 bits -> 3 bytes ok?

llvm/lib/Target/AIE/AIESpillSlotOptimization.cpp
```diff
@@ -0,0 +1,514 @@
...
+
+  // Compute offset within the spill slot (base + subreg offset in bytes).
+  // Note: BaseOffset is non-zero for dimension register spills that expand
+  // into multiple scalar accesses at different offsets within the same slot.
+  const int64_t Offset =
+      BaseOffset + (EI.SubRegIndex
+                        ? TRI.getSubRegIdxOffset(EI.SubRegIndex) / BitsPerByte
```
martien-de-jong: Why no ceil() here?

llvm/lib/Target/AIE/AIESpillSlotOptimization.cpp
```diff
@@ -0,0 +1,514 @@
...
+
+    // Pass 2 (Forward): Collect spills with pre-MI liveness.
+    LiveRegs.clear();
+    LiveRegs.addLiveIns(MBB);
+    SmallVector<std::pair<MCPhysReg, const MachineOperand *>, 2> Clobbers;
+
+    for (MachineInstr &MI : MBB) {
+      int FI = -1;
+      if (TII->isStoreToStackSlot(MI, FI) && isValidSpillSlot(FI, MI)) {
+        auto Spill = std::make_unique<SpillPoint>(
```
martien-de-jong, line 401: Align Spill <-> SP?
Also, with you new factorization, the bulk of both inner loops could live in one lambda with an IsLoad parameter.

llvm/lib/Target/AIE/AIEBaseInstrInfo.cpp
andcarminati: AFAIK yes!

llvm/lib/Target/AIE/AIEBaseInstrInfo.cpp
```diff
@@ -1529,3 +1530,69 @@ bool AIEBaseInstrInfo::areMemAccessesTriviallyDisjoint(
...
+      continue;
+    ++MatchingStackMMOs;
+  }
+
+  if (!MatchingStackMMOs)
+    return 0;
+
+  assert(MatchingStackMMOs == 1 &&
+         "Expected exactly one fixed-stack MachineMemOperand");
+  assert(MI.hasOneMemOperand() &&
```
andcarminati, line 1583: Could we simplify this? We count the matching MMOs and then we assert `MI.hasOneMemOperand()`.

llvm/lib/Target/AIE/AIESpillSlotOptimization.cpp
```diff
@@ -0,0 +1,514 @@
...
+
+  SubRegInfo(MCRegister Reg, unsigned SubRegIndex, unsigned MemSize,
+             int64_t Offset, bool IsLive);
+};
+
+/// Per-instruction spill/reload with pre-computed decomposition info.
+struct SpillPoint {
+  MachineInstr *MI = nullptr;
+  int FI = -1;
+  bool IsStore = false;
```
andcarminati, line 85: Can we get this information from MI?

llvm/lib/Target/AIE/AIESpillSlotOptimization.cpp
```diff
@@ -0,0 +1,514 @@
...
+  ReplacementSlot *findSlot(int64_t Offset) {
+    for (ReplacementSlot &R : Slots)
+      if (R.Offset == Offset)
+        return &R;
+    return nullptr;
+  }
+
+  /// Get or create a replacement slot for a given offset. Multiple subregs at
+  /// the same offset share a slot (StackSlotColoring ensures safety).
+  /// Collecting all slot metadata before committing to stack frame allocation.
```
andcarminati: How this -> `/// Collecting all slot metadata before committing to stack frame allocation.`

Connects to the function's behavior?

llvm/lib/Target/AIE/AIESpillSlotOptimization.cpp
```diff
@@ -0,0 +1,514 @@
...
+  }
+  OS << "]";
+  return OS;
+}
+
+//===----------------------------------------------------------------------===//
+// SpillPoint Construction
+//===----------------------------------------------------------------------===//
+
+SubRegInfo::SubRegInfo(MCRegister Reg, unsigned SubRegIndex, unsigned MemSize,
```
andcarminati: Can we move this constructor the the struct declaration? For example, in `FrameIndexInfo` I already see some code, so we can have the same standard.

llvm/lib/Target/AIE/AIESpillSlotOptimization.cpp
andcarminati: Should we get the minimum with 4?

llvm/lib/Target/AIE/AIESpillSlotOptimization.cpp
```diff
@@ -0,0 +1,514 @@
...
+
+//===----------------------------------------------------------------------===//
+// Data Structures
+//===----------------------------------------------------------------------===//
+
+/// Pre-computed info for a single subreg within a spill/reload.
+struct SubRegInfo {
+  MCRegister Reg;
+  unsigned SubRegIndex = 0;
+  unsigned MemSize = 0;
```
andcarminati: MemSize in bits or bytes?

llvm/lib/Target/AIE/AIESpillSlotOptimization.cpp
```diff
@@ -0,0 +1,514 @@
...
+
+  bool runOnMachineFunction(MachineFunction &MF) override;
+
+private:
+  const TargetRegisterInfo *TRI = nullptr;
+  const AIEBaseInstrInfo *TII = nullptr;
+  MachineFrameInfo *MFI = nullptr;
+  MachineRegisterInfo *MRI = nullptr;
+
+  MapVector<int, FrameIndexInfo> FIInfoMap;
```
martien-de-jong: Comment please. Mapping what to what?

llvm/lib/Target/AIE/AIESpillSlotOptimization.cpp
```diff
@@ -0,0 +1,514 @@
...
+
+  // Composite spill: decompose into subregs.
+  for (const auto &EI : ExpandInfos) {
+    SubRegInfo SI = buildSubRegInfo(Reg, BaseOffset, EI, TRI, LiveRegs, MRI);
+    DecomposeProfitable |= SI.IsLive;
+    SubRegs.push_back(SI);
+  }
+
+  // Only profitable if decomposition can reduce memory size.
+  const bool SizeReduction =
```
andcarminati, line 343: I guess we can simplify by merging those two loops. In a single pass through `ExpandInfos`  (previous loop) we can also get this `SizeReduction` information.

llvm/lib/Target/AIE/AIESpillSlotOptimization.cpp
```diff
@@ -0,0 +1,514 @@
...
+
+char AIESpillSlotOptimization::ID = 0;
+char &llvm::AIESpillSlotOptimizationID = AIESpillSlotOptimization::ID;
+
+INITIALIZE_PASS(AIESpillSlotOptimization, DEBUG_TYPE,
+                "AIE Spill Slot Optimization", false, false)
+
+bool AIESpillSlotOptimization::isValidSpillSlot(int FI,
+                                                MachineInstr &MI) const {
+  if (FI < 0 || FI >= MFI->getObjectIndexEnd() || MFI->isDeadObjectIndex(FI) ||
```
andcarminati: nit: turn this into a single return.

llvm/lib/Target/AIE/AIESpillSlotOptimization.cpp
```diff
@@ -0,0 +1,514 @@
...
+      }
+      if (Alignment > R->Alignment)
+        R->Alignment = Alignment;
+      return *R;
+    }
+    return Slots.emplace_back(Offset, Size, Alignment, RC);
+  }
+
+  /// Allocate a stack object for the given slot if not already created.
+  void allocateStackObject(ReplacementSlot &R, MachineFrameInfo &MFI) {
```
andcarminati: This idiom is could be more clear if could return explicitly the FI and the caller could assign it to NewFI.

llvm/lib/Target/AIE/AIESpillSlotOptimization.cpp
```diff
@@ -0,0 +1,514 @@
...
+    SubRegInfo SI = buildSubRegInfo(Reg, BaseOffset, EI, TRI, LiveRegs, MRI);
+    DecomposeProfitable |= SI.IsLive;
+    SubRegs.push_back(SI);
+  }
+
+  // Only profitable if decomposition can reduce memory size.
+  const bool SizeReduction =
+      llvm::any_of(SubRegs, [SlotSize](const SubRegInfo &SI) {
+        return SI.MemSize < SlotSize;
+      });
```
martien-de-jong, line 346: Isn't this always true if a register splits in subregisters?

llvm/lib/Target/AIE/AIESpillSlotOptimization.cpp
```diff
@@ -0,0 +1,514 @@
...
+  TII = static_cast<const AIEBaseInstrInfo *>(MF.getSubtarget().getInstrInfo());
+  MFI = &MF.getFrameInfo();
+  MRI = &MF.getRegInfo();
+
+  if (MFI->getNumObjects() == 0) {
+    LLVM_DEBUG(dbgs() << "No stack objects, skipping\n");
+    return false;
+  }
+
+  // Phase 1: Iterate until decomposition stabilizes (handles nested
```
andcarminati: nit: as we have a single phase here, we don't need to enumerate.

llvm/lib/Target/AIE/AIESpillSlotOptimization.cpp
andcarminati: Also, is `FI < 0 || FI >= MFI->getObjectIndexEnd() ||` a protection against out-of-bounds access in MFI?

llvm/lib/Target/AIE/AIESpillSlotOptimization.cpp
```diff
@@ -0,0 +1,514 @@
...
+  // Use TII methods to create spill/reload with correct pseudo opcodes.
+  if (SP.IsStore) {
+    TII->storeRegToStackSlot(MBB, MI.getIterator(), SI.Reg, IsKill, R.NewFI,
+                             R.RC, TRI, Register());
+  } else {
+    TII->loadRegFromStackSlot(MBB, MI.getIterator(), SI.Reg, R.NewFI, R.RC, TRI,
+                              Register());
+  }
+
+  MachineInstr &NewMI = *std::prev(MI.getIterator());
```
andcarminati, line 443: Oh, this is a workaround to locate the new instruction ;-)

llvm/test/CodeGen/AIE/aie2/set.ll
```diff
@@ -582,24 +579,24 @@ define dso_local noundef <16 x i64> @_Z17test_set_v16acc64iDv4_u7__acc64(i32 nou
...
 ; CHECK-NEXT:    nop
-; CHECK-NEXT:    lda r16, [sp, #-160] // 4-byte Folded Reload
+; CHECK-NEXT:    lda r16, [sp, #-64] // 4-byte Folded Reload
 ; CHECK-NEXT:    ret lr
 ; CHECK-NEXT:    nop // Delay Slot 5
 ; CHECK-NEXT:    nop // Delay Slot 4
 ; CHECK-NEXT:    nop // Delay Slot 3
 ; CHECK-NEXT:    nop // Delay Slot 2
-; CHECK-NEXT:    paddb [sp], #-160 // Delay Slot 1
+; CHECK-NEXT:    paddb [sp], #-64 // Delay Slot 1
```
andcarminati, line 599: This is amazing!

llvm/lib/Target/AIE/AIESpillSlotOptimization.cpp
```diff
@@ -0,0 +1,514 @@
...
+      // to check for matching stores and erase the reload if its store was
+      // decomposed, even if this reload has no live subregs to emit.
+
+      for (const SubRegInfo &SI : SP->SubRegs) {
+        if (!StoredOffsets.contains(SI.Offset))
+          continue;
+
+        HasMatchingStore = true;
+        // Only emit loads if decomposition is profitable and subreg is live
+        if (!SI.IsLive)
```
martien-de-jong: Remark that a reload's IsLive may differ from that of the corresponding Spill's.

llvm/lib/Target/AIE/AIESpillSlotOptimization.cpp
```diff
@@ -0,0 +1,514 @@
...
+  LLVM_DEBUG(dbgs() << "    Created spill for " << printReg(SI.Reg, TRI)
+                    << " -> %stack." << R.NewFI << "\n");
+}
+
+bool AIESpillSlotOptimization::rewriteInstructions(MachineFunction &MF) {
+  LLVM_DEBUG(dbgs() << "=== Phase 2: Rewrite Instructions ===\n");
+
+  SmallVector<MachineInstr *, 32> ToErase;
+
+  forEachFI([&](int FI, FrameIndexInfo &Info) {
```
martien-de-jong: I think I would find it more readable if this was a normal loop and the two main parts  (Spills and Reloads) were factored out as lambdas

llvm/lib/Target/AIE/AIESpillSlotOptimization.cpp
```diff
@@ -0,0 +1,514 @@
...
+
+  // Phase 1: Iterate until decomposition stabilizes (handles nested
+  // hierarchies)
+  bool Changed = false;
+  bool IterationChanged;
+  do {
+    analyzeFunction(MF);
+    IterationChanged = rewriteInstructions(MF);
+    if (IterationChanged) {
+      removeOriginalSlots();
```
martien-de-jong: I'm wondering whether we could do one sweep of removeOriginalSlots() after convergence

llvm/test/CodeGen/AIE/abi/int-ptr-structs.ll
```diff
@@ -4,7 +4,7 @@
 ; See https://llvm.org/LICENSE.txt for license information.
 ; SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
 ;
-; (c) Copyright 2023-2024 Advanced Micro Devices, Inc. or its affiliates
+; (c) Copyright 2023-2026 Advanced Micro Devices, Inc. or its affiliates
```
martien-de-jong, line 7: Only copyright change?

F-Stuckmann: this crept up in fixup commits

llvm/lib/Target/AIE/AIESpillSlotOptimization.cpp
F-Stuckmann: yes

llvm/lib/Target/AIE/AIESpillSlotOptimization.cpp
F-Stuckmann: yes anything smaller than 4 is problematic and we do not allow it.
The constructor increases the memory size to at least 4 bytes

llvm/lib/Target/AIE/AIESpillSlotOptimization.cpp
F-Stuckmann: No we may have a pseudo GPR Spill that does not reduce the Size of the spill.
This is a nice artifact of our inconsistent handling of pseudo Stack load/stores

llvm/lib/Target/AIE/AIESpillSlotOptimization.cpp
F-Stuckmann: yep, somebody did not properly think about that interface.

F-Stuckmann:
Latest Performance numbers:
Compute Instruction Count -0.03%

PM -5.28 Bytes

Stack -27.32 Bytes

andcarminati:
To avoid confusion, you could remove old graphs!

llvm/lib/Target/AIE/AIESpillSlotOptimization.cpp
```diff
@@ -0,0 +1,594 @@
...
+}
+
+bool AIESpillSlotOptimization::rewriteInstructions(MachineFunction &MF) {
+  LLVM_DEBUG(dbgs() << "=== Phase 2: Rewrite Instructions ===\n");
+
+  SmallVector<MachineInstr *, 32> ToErase;
+
+  // Process spills: create replacement slots and emit decomposed stores.
+  // Returns the set of offsets that have stores (for reload filtering).
+  auto ProcessSpills =
```
andcarminati: Maybe a function fits better for a big code.

llvm/lib/Target/AIE/AIESpillSlotOptimization.cpp
```diff
@@ -0,0 +1,594 @@
...
+  OrigSlotDecompositions.clear();
+
+  for (MachineBasicBlock &MBB : MF) {
+    // Pass 1 (Backward): Collect reloads with post-MI liveness.
+    LivePhysRegs LiveRegs(*TRI);
+    LiveRegs.addLiveOuts(MBB);
+
+    for (MachineInstr &MI : reverse(MBB)) {
+      int FI = -1;
+      if (TII->isLoadFromStackSlot(MI, FI) && isValidSpillSlot(FI, MI)) {
```
andcarminati, line 384: isValidSpillSlot may fail? Should it be an assert?

llvm/lib/CodeGen/PrologEpilogInserter.cpp
```diff
@@ -1046,11 +1051,11 @@ void PEI::calculateFrameObjectOffsets(MachineFunction &MF) {
   SmallVector<int, 8> ObjectsToAllocate;
 
   // Then prepare to assign frame offsets to stack objects that are not used to
-  // spill callee saved registers.
+  // spill callee saved registers (unless IncludeCSInOrdering is set).
   for (unsigned i = 0, e = MFI.getObjectIndexEnd(); i != e; ++i) {
     if (MFI.isObjectPreAllocated(i) && MFI.getUseLocalStackAllocationBlock())
       continue;
-    if (i >= MinCSFrameIndex && i <= MaxCSFrameIndex)
+    if (!IncludeCSInOrdering && i >= MinCSFrameIndex && i <= MaxCSFrameIndex)
```
martien-de-jong: nit: earlier these conditions were reversed...

llvm/lib/Target/AIE/aie1/AIE1FrameLowering.h
```diff
@@ -23,6 +23,9 @@ class AIEFrameLowering : public AIEBaseFrameLowering {
       : AIEBaseFrameLowering(STI,
                              /*StackAlignment=*/Align(32)) {}
 
+  /// AIE1 does not support including callee-saved in frame ordering.
+  bool orderFrameObjectsIncludesCalleeSaves() const override { return false; }
```
martien-de-jong: You have made the default such that we don't need to touch existing architectures, so please don't.

llvm/lib/Target/AIE/AIESpillSlotOptimization.cpp
```diff
@@ -0,0 +1,594 @@
...
+        auto SP = std::make_unique<SpillPoint>(MI, FI, MFI->getObjectSize(FI),
+                                               LiveRegs, *TII, *MRI);
+        LLVM_DEBUG(dbgs() << "  " << *SP << ": " << MI);
+        OrigSlotDecompositions[FI].Reloads.push_back(std::move(SP));
```
andcarminati: What about `emplace_back` for in place construction? Unique pointers are not so common here.

llvm/lib/Target/AIE/AIEBaseFrameLowering.cpp
```diff
@@ -370,3 +371,65 @@ void AIEBaseFrameLowering::emitEpilogue(MachineFunction &MF,
...
+
+  const MachineFrameInfo &MFI = MF.getFrameInfo();
+
+  SmallVector<FrameObject> Objects;
+  Objects.reserve(ObjectsToAllocate.size());
+
+  for (const int FI : ObjectsToAllocate) {
+    const int64_t ObjSize = MFI.getObjectSize(FI);
+    if (ObjSize <= 0)
+      return; // Variable-sized object; cannot reorder
```
martien-de-jong: I think you can just skip it. variable sized objects aren't mixed with static sized ones.

llvm/test/CodeGen/AIE/abi/int-ptr-structs.ll
```diff
@@ -17,7 +17,7 @@ define void @call_pass_2S3IP() {
 ; CHECK-NEXT:  // %bb.0: // %entry
 ; CHECK-NEXT:    padda [sp], #32
 ; CHECK-NEXT:    mov.u20 r12, #6
-; CHECK-NEXT:    st.spil lr, [sp, #-32]
+; CHECK-NEXT:    st.spil lr, [sp, #-32] // 4-byte Folded Spill
```
martien-de-jong: I think this change can be reverted. The trailing comment is redundant/

llvm/lib/Target/AIE/AIESpillSlotOptimization.cpp
```diff
@@ -0,0 +1,594 @@
...
+    const unsigned ComposedIndex =
+        ParentSubRegIndex
+            ? TRI.composeSubRegIndices(ParentSubRegIndex, EI.SubRegIndex)
+            : EI.SubRegIndex;
+
+    flattenExpandInfos(EI.ExpandedOpCode, ComposedIndex, TII, TRI, Result);
+
+    // Propagate explicit MemSize to leaf if specified at this level.
+    if (EI.MemSize && !Result.empty())
+      Result.back().SizeBytes = EI.MemSize;
```
andcarminati, line 266: I am a bit confused by this part.

llvm/lib/Target/AIE/AIEBaseFrameLowering.cpp
```diff
@@ -370,3 +371,65 @@ void AIEBaseFrameLowering::emitEpilogue(MachineFunction &MF,
...
+      return A.Alignment > B.Alignment;
+    if (A.Size != B.Size)
+      return A.Size > B.Size;
+    return A.Index < B.Index;
+  }
+};
+
+} // namespace
+
+/// Order stack objects by alignment descending to minimize padding.
```
andcarminati, line 399: This is a bit confusing. We align "Descending", but the comments says that "Objects with smaller alignment end up closer to SP". Which is the stack order? Or the stack is reversed or the alignment is ascending. Could you please clarify this with a comment?

llvm/lib/Target/AIE/AIEBaseFrameLowering.cpp
```diff
@@ -370,3 +371,65 @@ void AIEBaseFrameLowering::emitEpilogue(MachineFunction &MF,
   // Deallocate stack
   adjustSPReg(MBB, MBBI, DL, -StackSize, MachineInstr::FrameDestroy);
 }
+
+namespace {
+
+/// Compute padding needed to align an offset to a given alignment.
+inline int64_t computePadding(int64_t Offset, Align Alignment) {
```
andcarminati: I think this is not used.

llvm/lib/Target/AIE/AIEBaseFrameLowering.cpp
```diff
@@ -370,3 +371,65 @@ void AIEBaseFrameLowering::emitEpilogue(MachineFunction &MF,
...
+    return;
+
+  const MachineFrameInfo &MFI = MF.getFrameInfo();
+
+  SmallVector<FrameObject> Objects;
+  Objects.reserve(ObjectsToAllocate.size());
+
+  for (const int FI : ObjectsToAllocate) {
+    const int64_t ObjSize = MFI.getObjectSize(FI);
+    if (ObjSize <= 0)
```
andcarminati: You can use this to detect the end of the array and then sort in place, much simpler!

llvm/lib/Target/AIE/AIESpillSlotOptimization.cpp
F-Stuckmann: No we may encounter dead spill slots. This function can return false.

llvm/lib/Target/AIE/AIEBaseFrameLowering.cpp
F-Stuckmann: yes

llvm/lib/Target/AIE/AIESpillSlotOptimization.cpp
F-Stuckmann: updated

llvm/lib/CodeGen/PrologEpilogInserter.cpp
F-Stuckmann: harmonized it

andcarminati (approved):
LGTM.

PR #756, opened 2026-01-20
[AIEX] fix const half select G_AIE_VSEL

F-Stuckmann:
Fix Selection of VSEL when using halfSelect masks (00FF or FF00).

VSEL works in little endian mode: 
00FF would select src2 for the lower part and src1 for the higher part.


andcarminati (approved):
LGTM.

PR #760, opened 2026-01-27
[MIR] Update scalar LLT size check from 16 to 32

niwinanto:
See the LLT implementation for scalar types.
``` 
 /// * Non-pointer scalar (isPointer == 0 && isVector == 0):
  ///   SizeInBits: 32;
  static const constexpr BitFieldInfo ScalarSizeFieldInfo{32, 29};
```


mludevid (approved):
LGTM.

PR #763, opened 2026-01-27
WAW Register Renamer Crash Fix

F-Stuckmann:
When we have to revert Register Allocation during WAW we may have to revert a valid new Register allocation because it conflicts with a previously allocated register that we have to revert.


llvm/lib/Target/AIE/AIEWawRegRewriter.cpp
```diff
@@ -390,34 +391,74 @@ void AIEWawRegRewriter::revertAllocation(OriginalAllocation &Candidates,
...
-      }
+    // Case 2: current assignment conflicts with registers to restore
+    if (!VRM->hasPhys(VReg))
+      continue;
+    const MCPhysReg CurrentPhys = VRM->getPhys(VReg);
+    const bool CurrentConflictsWithRestore =
+        any_of(OrigRegsToRestore, [&](MCRegister OrigReg) {
+          return TRI->regsOverlap(CurrentPhys, OrigReg);
+        });
+    if (CurrentConflictsWithRestore) {
```
andcarminati: nit: nl here.

andcarminati: Also, to keep a symmetric code pattern and reduce identantion:
```
if (!CurrentConflictsWithRestore)
    continue;
```

andcarminati:
LGTM, just some nits.

llvm/lib/Target/AIE/AIEWawRegRewriter.cpp
```diff
@@ -390,34 +391,74 @@ void AIEWawRegRewriter::revertAllocation(OriginalAllocation &Candidates,
...
+  for (const auto &[MO, Org] : Candidates) {
+    if (!CandidatesToRevert.contains(MO))
+      continue;
+    const Register VReg = MO->getReg();
+    if (VRM->hasPhys(VReg))
+      unassignReg(VReg);
+  }
+
+  // Second pass: reassign all reverted candidates back to their original.
+  for (const auto &[MO, Org] : Candidates) {
```
F-Stuckmann, line 493: duplicated code

llvm/lib/Target/AIE/AIEWawRegRewriter.cpp
```diff
@@ -390,34 +391,74 @@ void AIEWawRegRewriter::revertAllocation(OriginalAllocation &Candidates,
...
-  // with a failed register class.
-  // The partial allocation may conflict with the original one in ugly ways,
-  // which means we may not be able to reassign in arbitrary order.
-  // To be safe, reset all allocations first.
-  for (auto [RC, Success] : RegClasses) {
-    LLVM_DEBUG(dbgs() << "RC=" << TRI->getRegClassName(RC)
-                      << (Success ? " Succeeded\n" : " Failed\n"));
-    if (Success) {
-      continue;
+  auto OrigOverlapsFailedRC = [&](MCRegister Org) {
```
andcarminati: If you populate your `RegClassSuccess &RegClasses` with the overlapping classes, your job is done!

F-Stuckmann: I only want to revert the reg  that has overlapping register conflict, not just any registerclass that my overlap but is legal.

llvm/lib/Target/AIE/AIEWawRegRewriter.cpp
```diff
@@ -383,43 +383,130 @@ bool AIEWawRegRewriter::reAllocate(OriginalAllocation &Candidates,
...
 
-  auto Overlaps = [this](MCRegister Reg, const TargetRegisterClass *RC) {
-    return any_of(RC->getRegisters(), [this, Reg](MCRegister RCReg) {
-      return TRI->regsOverlap(Reg, RCReg);
-    });
+  // ToRevert tracks which virtual registers need to revert to their originals.
+  unsigned MaxVReg = 0;
+  for (const auto &[MO, Org] : Candidates)
+    MaxVReg = std::max(MaxVReg, MO->getReg().virtRegIndex());
+  BitVector ToRevert(MaxVReg + 1);
```
andcarminati: Oh this could be a very sparse bitvector. What do you think about a `SparseSet<Register, VirtReg2IndexFunctor> ToRevert;`?

llvm/lib/Target/AIE/AIEWawRegRewriter.cpp
```diff
@@ -383,43 +383,130 @@ bool AIEWawRegRewriter::reAllocate(OriginalAllocation &Candidates,
...
+
+  // Helper: Get RegUnits for a physical register as a BitVector.
+  auto getRegUnits = [&](MCPhysReg Reg) {
```
andcarminati: Closures are objects/variables, so it is better to start with upper case.

llvm/lib/Target/AIE/AIEWawRegRewriter.cpp
```diff
@@ -383,43 +383,130 @@ bool AIEWawRegRewriter::reAllocate(OriginalAllocation &Candidates,
...
+    BitVector Units(NumRegUnits);
```
andcarminati, line 414: Same idea, we could optimize this to a set.

llvm/lib/Target/AIE/AIEWawRegRewriter.cpp
```diff
@@ -383,43 +383,130 @@ bool AIEWawRegRewriter::reAllocate(OriginalAllocation &Candidates,
...
+  // A successful candidate must revert if its NEW assignment overlaps with
+  // BlockedUnits. When it reverts, its original is added to BlockedUnits,
+  // potentially causing more candidates to revert.
+  // Convergence: Remaining shrinks each iteration, BlockedUnits only grows.
+  // Termination: Loop exits when no new conflicts are found.
+  LLVM_DEBUG(
+      dbgs() << "Phase 2: Fixed-point iteration for cascading conflicts\n");
+  unsigned Iteration = 0;
+  bool Changed = true;
+  while (Changed) {
```
andcarminati: As we will iterate at least one -> `do {} while ();`

llvm/lib/Target/AIE/AIEWawRegRewriter.cpp
F-Stuckmann: I still do not understand the clang format rule, this behaves more like a method than a Variable, but yeah i'll adhere to the clang format rules

llvm/lib/Target/AIE/AIEWawRegRewriter.cpp
andcarminati: If you decide to keep the bitvector, you can use `MRI->getNumVirtRegs()` directly.

llvm/lib/Target/AIE/AIEWawRegRewriter.cpp
```diff
@@ -383,43 +383,130 @@ bool AIEWawRegRewriter::reAllocate(OriginalAllocation &Candidates,
...
+    Changed = false;
+    ++Iteration;
+
+    // Swap-and-pop: O(1) removal by swapping with last element.
```
andcarminati: Why not just use a linked list, where removal from the front (`pop_front`) is O(1)? So you don't need to swap.

llvm/lib/Target/AIE/AIEWawRegRewriter.cpp
```diff
@@ -383,43 +383,130 @@ bool AIEWawRegRewriter::reAllocate(OriginalAllocation &Candidates,
...
+    for (size_t I = 0; I < Remaining.size();) {
+      const auto &[MO, Org] = Remaining[I];
+      const Register VReg = MO->getReg();
+
+      const BitVector NewUnits = getRegUnits(VRM->getPhys(VReg));
+      if (NewUnits.anyCommon(BlockedUnits)) {
```
andcarminati, line 465: nit: `if ((Changed = NewUnits.anyCommon(BlockedUnits)))`

llvm/lib/Target/AIE/AIEWawRegRewriter.cpp
```diff
@@ -383,43 +383,130 @@ bool AIEWawRegRewriter::reAllocate(OriginalAllocation &Candidates,
...
+  LLVM_DEBUG(
+      dbgs() << "Phase 2: Fixed-point iteration for cascading conflicts\n");
+  unsigned Iteration = 0;
+  bool Changed = true;
+  while (Changed) {
+    Changed = false;
+    ++Iteration;
+
+    // Swap-and-pop: O(1) removal by swapping with last element.
+    for (size_t I = 0; I < Remaining.size();) {
```
andcarminati: Why not `for (size_t I = 0; I < Remaining.size(); I++) `?

llvm/lib/Target/AIE/AIEWawRegRewriter.cpp
andcarminati: But, on the other hand, after the first iteration, `Remaining` will be empty and the for will not execute in the second iteration. I see that we always converge on the first iteration. Alternatively, the `for` makes sure to leave the `Remaining` vector empty (see condition).

llvm/lib/Target/AIE/AIEWawRegRewriter.cpp
```diff
@@ -383,43 +383,130 @@ bool AIEWawRegRewriter::reAllocate(OriginalAllocation &Candidates,
...
+    const Register VReg = MO->getReg();
+    if (ToRevert.test(VReg.virtRegIndex())) {
+      LLVM_DEBUG(dbgs() << "  Reverting " << printReg(VReg, TRI) << " to "
+                        << printReg(Org, TRI) << "\n");
+      assignReg(VReg, Org);
     }
   }
+  LLVM_DEBUG(dbgs() << "=== revertAllocation complete ===\n");
 }
-
```
andcarminati, line 422: We can keep this empty line.

llvm/lib/Target/AIE/AIEWawRegRewriter.cpp
F-Stuckmann: we do have a very barebones unit-test that does not test this cascading effect

llvm/lib/Target/AIE/AIEWawRegRewriter.cpp
F-Stuckmann: no we cannot. When swapping we have to keep I stable, otherwise we miss a conflict check! 
But i agree a linked list is propably easier to iterate and to understand

llvm/lib/Target/AIE/AIEWawRegRewriter.cpp
F-Stuckmann: I find this more difficult to read, also we could need `|=`, because we want to check if any `anyCommon` in the Remaining ever fired.

llvm/lib/Target/AIE/AIEWawRegRewriter.cpp
F-Stuckmann: I think overlap detection with bitvectors is more efficient than with sets.

llvm/lib/Target/AIE/AIEWawRegRewriter.cpp
F-Stuckmann: yeah that is cleaner

martien-de-jong (approved):
Approval for expediency. As discussed, I think the crash has proved register classes not to be independent, so partitioning the problem based on register classes has no real use. Hence we can drop the concept of a class failure, and just see whether the register could be assigned. reverting a single failure will now automatically catch the knock-on failures.

PR #765, opened 2026-01-30
[AIEX] Improve prologue scheduling by using AIERegMemEventTracker

andcarminati:
The goal is to reduce pessimism by precise register event tracking. Locks and events are handled in the same way as in the epilogue.


llvm/test/CodeGen/AIE/aie2p/elongate/zol_phdr_loopbundles_112bytes.mir
```diff
@@ -21,13 +21,8 @@
   ; CHECK-LABEL: zol_elongated_preheader_plus_NumOfLoopBundles_ge112bytes:
   ; CHECK:         .p2align 4
   ; CHECK-NEXT:  // %bb.0: // %entry
-  ; CHECK-NEXT:    vldb x6, [p0], #0
```
andcarminati, line 24: This is something relevant. The old approach detected a dependency through p0, but by using maxLatency, we extend the distance to 7 (load latency), what is quite pessimistic.

andcarminati:
Next step is to refactor a bit the code.

llvm/test/CodeGen/AIE/aie2p/end-to-end/add-att-broadcasting-aa.ll
```diff
@@ -145,8 +144,7 @@ for.cond.cleanup.ret.exitStub:                    ; preds = %for.cond.cleanup
 define dso_local void @add_attribute_bcast_prologue2(ptr %ifm2, ptr %ifm1, i32 %div16, ptr noalias %ofm, ptr noalias %targetptr) {
 ; CHECK-LABEL: add_attribute_bcast_prologue2:
 ; CHECK:       // %bb.0: // %newFuncRoot
-; CHECK-NEXT:    st.s16 r0, [p3, #0]; nopxm
```
andcarminati, line 148: Investigate this!

llvm/test/CodeGen/AIE/aie2p/end-to-end/add-att-broadcasting-aa.ll
```diff
@@ -27,8 +27,7 @@ declare <32 x bfloat> @llvm.aie2p.v32accfloat.to.v32bf16(<32 x float>) #1
 define dso_local void @add_attribute_bcast_prologue1(ptr %ifm2, ptr %ifm1, i32 %div16, ptr noalias %ofm, ptr noalias %targetptr, ptr %targetptrmayalias) {
 ; CHECK-LABEL: add_attribute_bcast_prologue1:
 ; CHECK:       // %bb.0: // %newFuncRoot
-; CHECK-NEXT:    st r0, [p4, #0]; nopx
```
andcarminati, line 30: Investigate this!

llvm/lib/Target/AIE/AIEBaseSubtarget.cpp
```diff
@@ -410,78 +410,60 @@ class EmitFixedSUnits : public ScheduleDAGMutation {
...
+    // Only handle prologues (preheader blocks), not epilogues here
+    if (!BotFixedBundles.empty() && BS.Kind != BlockType::Epilogue) {
+      AIERegMemEventTracker BackwardRET{ItinData, TRI, TII, AA};
+
+      // Track bot-fixed bundles backward (this sets BotFixedRegionSize)
+      BackwardRET.computeUseDefBackward(BotFixedBundles,
+                                        /*InSeparateRegion=*/false);
+
+      // Find the loop successor - must exist if we have bot-fixed bundles in a
+      // prologue
```
F-Stuckmann: What is the difference between regular interblock scheduling and the pipelining interblock scheduling?

llvm/lib/Target/AIE/AIEBaseSubtarget.cpp
```diff
@@ -410,78 +410,60 @@ class EmitFixedSUnits : public ScheduleDAGMutation {
...
         }
-        continue;
       }
+      assert(LoopSucc &&
+             "Bot-fixed bundles present but no pipelined loop successor found");
+
+      const BlockState &LoopBS =
+          Scheduler->getInterBlock().getBlockState(LoopSucc);
+      ArrayRef<AIE::MachineBundle> LoopTimedBundles = LoopBS.getTop().Bundles;
+      BackwardRET.computeUseDefBackward(LoopTimedBundles,
```
F-Stuckmann: How do we handle the wrap around nature of the loops?

llvm/lib/Target/AIE/AIERegMemEventTracker.cpp
```diff
@@ -40,10 +55,144 @@ void AIERegMemEventTracker::updateUseDefMaxCycle(Register Reg, unsigned Latency,
   }
 }
 
+void AIERegMemEventTracker::updateLastStoreCycle(int StoreCycle) {
+  LastStoreCycle = std::max(LastStoreCycle, StoreCycle);
```
F-Stuckmann, line 59: What is the difference between this and the scoreboard we use in the Postpipeliner? Can't we reuse it here?

llvm/lib/Target/AIE/AIEBaseSubtarget.cpp
andcarminati: They are different in the sense that without pipelining:
* We push a safety margin on the epilogue block based on what we see on the loop.
* We initialize the bot scoreboard of the prologue based on the loop state.

With a pipelined loop, we have the top and bot insert regions that are already scheduled. We carefully dump those regions to a place where the normal DAG construction will not see them (sequence of bundles), because we need full control to not rearrange/change timing properties. After this timing region placement, we carefully create DAG nodes for the bundles (Also, normal DAG construction is not prepared for bundles- but this is not the reason), tying them together to a proper in-sequence scheduling. After, we construct a timeline of events relative the EntrySU and ExitSU, so we can create dependencies for the "free instructions". Register dependencies are the easy part on this timeline construction, the problem arises when we consider also memory cycles and alias relations (most of the code). As result, for each "free" instruction, we have a single edge to Entry/Exit modeling all memory and register latencies. For example, a memory instruction in the prologue can execute a register write inside de loop, and this must be also tracked - we do this here using events with negative cycles.

Also, this is the starting point for prologue/epilogue merging as a whole, the most complicated part is done by the scheduler: release each fixed node in sequence before any other node. This involves, for example, the switching the from top to bottom-um scheduling and delta cycles in both directions.

llvm/lib/Target/AIE/AIEBaseSubtarget.cpp
andcarminati: Once we execute the first iteration of the loop (all instructions will be there), this first iteration will be responsible for the first events in the timeline that may influence the prologue. Once we iterate again, events will not be interesting anymore because we already saw them previously, closer to the prologue. Happily, pipelined loops execute at least once and they contain all instructions.

llvm/lib/Target/AIE/AIERegMemEventTracker.cpp
andcarminati: A scoreboard can track resources (functional units/slots etc.) in the pipeline without the notion of latency. An instruction may reserve/require resources the the scoreboard, but there in no notion of operand latency/data dependency there. Here, we are not interested any resources, only latencies that can accommodate data dependencies and also memory relations between "free" instructions and already scheduled instructions.

llvm/lib/Target/AIE/AIEBaseSubtarget.cpp
```diff
@@ -410,78 +410,60 @@ class EmitFixedSUnits : public ScheduleDAGMutation {
...
-        }
-        // This instruction shoud be also scheduled after the first top-fixed
-        // instruction.
-        if (!TopFixedBundles.empty()) {
-          SUnit *FixedDepSU = DAG->getSUnit(&*getBundleStart(
-              TopFixedBundles.back().getInstrs().front()->getIterator()));
-          SDep Dep(FixedDepSU, SDep::Artificial);
-          Dep.setLatency(1);
-          FreeSU.addPred(Dep, /*Required=*/true);
+    // Handle bot-fixed instructions (prologue) with backward tracking
```
mludevid: Looks really good. I think that we can make this function even more understandable by factoring out the intermediate steps into helper functions. Maybe it's even possible to share some code between the prologue and epilogue handling.

llvm/lib/Target/AIE/AIEBaseSubtarget.cpp
```diff
@@ -410,78 +410,60 @@ class EmitFixedSUnits : public ScheduleDAGMutation {
...
+    // Only handle prologues (preheader blocks), not epilogues here
+    if (!BotFixedBundles.empty() && BS.Kind != BlockType::Epilogue) {
```
mludevid: I think this check is redundant. I suggest moving the block type check into an assert (as we do in the prologue case) or removing it altogether.

llvm/lib/Target/AIE/AIEBaseSubtarget.cpp
mludevid: Typo: `iterate`

llvm/lib/Target/AIE/AIERegMemEventTracker.cpp
```diff
@@ -40,10 +55,144 @@ void AIERegMemEventTracker::updateUseDefMaxCycle(Register Reg, unsigned Latency,
...
+  }
+  return CurrentMax;
+}
+
 void AIERegMemEventTracker::computeUseDefForward(
     ArrayRef<AIE::MachineBundle> Bundles, bool InSeparateRegion) {
   int Cycle = 0;
   const int TotalCycles = Bundles.size();
+
+  // Track top-fixed region size (only for the first call, not separate region)
```
mludevid, line 201: Nit: This comment explains the what and not the why. Could you add an explanation of why this tracking is only needed in the first call? Also, why not track it in both cases and move the check to the place where the total cycles is actually used, i.e. if the size is only needed in a computation in certain scenarios, then I think it makes sense to keep that check closer to the computation rather than here.

llvm/lib/Target/AIE/AIERegMemEventTracker.cpp
```diff
@@ -101,109 +250,183 @@ void AIERegMemEventTracker::computeUseDefForward(
...
+  const int TotalCycles = Bundles.size();
+
+  // Track bot-fixed region size (only for the first call, not separate region)
+  if (!InSeparateRegion) {
+    BotFixedRegionSize = TotalCycles;
+  }
+
+  // Count progressively from 0 as we iterate backward through bundles
+  // Cycle represents the distance from ExitSU for each bundle
+  int Cycle = 0;
```
mludevid, line 264: Nit: AFAIK we usually call this the height. Please rename for consistency.

llvm/lib/Target/AIE/AIERegMemEventTracker.cpp
```diff
@@ -101,109 +250,183 @@ void AIERegMemEventTracker::computeUseDefForward(
...
+
+  for (const auto &Bundle : reverse(Bundles)) {
+    // Cycle is the distance from ExitSU for this bundle
```
mludevid: Nit: Repeated comment.

llvm/lib/Target/AIE/AIERegMemEventTracker.cpp
```diff
@@ -101,109 +250,183 @@ void AIERegMemEventTracker::computeUseDefForward(
...
+                                  : getFirstMemCycle(/*IsStore=*/true);
+    MaxLatency =
+        checkMemoryDependency(MaxLatency, LoadStoreCycle, MIMemoryCycle, true);
+  }
+  // Lock instructions: all subsequent memory operations must wait
+  if (TII->isLock(MI.getOpcode())) {
+    if (getFirstMemoryAccessCycle() > INT_MIN) {
+      const int CoreResumeCycle = TII->getCoreResumeCycleAfterLock();
+      const int MIResumeCycle = CoreResumeCycle - 1;
+      const int MemDep = getFirstMemoryAccessCycle() - MIResumeCycle + 2;
```
mludevid: Why `+2`?

llvm/lib/Target/AIE/AIERegMemEventTracker.cpp
```diff
@@ -15,23 +15,38 @@
...
+// Helper function to identify event-like instructions (side effects without
+// register/memory dependencies)
+bool isEventLikeInstruction(const MachineInstr &MI,
+                            const AIEBaseInstrInfo *TII) {
+  auto UseOrDefReg = [](const MachineInstr &MI) {
+    return llvm::any_of(MI.operands(),
+                        [](const MachineOperand &MO) { return MO.isReg(); });
+  };
+
+  return MI.hasUnmodeledSideEffects() && !MI.mayLoadOrStore() &&
```
F-Stuckmann, line 32: why do we treat unmodeled sideffects differently than locks or load/stores?

llvm/lib/Target/AIE/AIERegMemEventTracker.cpp
```diff
@@ -101,109 +250,183 @@ void AIERegMemEventTracker::computeUseDefForward(
...
-      const int MemoryDep = MaxAliasingStoreCycle - MIMemoryCycle + 1;
-      MaxLatency = std::max(MemoryDep, 0);
-    } else {
-      // Conservative: Use global LastStoreCycle
-      const int MemoryDep = getLastStoreCycle() - MIMemoryCycle + 1;
-      MaxLatency = std::max(MemoryDep, 0);
-    }
+    // Check stores in top-fixed (RAW for loads, WAR for stores)
+    const int StoreCycle =
+        AA ? getMaxAliasingMemCycle(MI, /*IsStore=*/true) : getLastStoreCycle();
```
F-Stuckmann: shouldn't this be `getLastMemoryAccessCycle` since we are unifying store and load?

andcarminati:
QoR:
image
image



This helps to improve the outermost loop of GEMM int8.

llvm/lib/Target/AIE/AIEBaseSubtarget.cpp
andcarminati: Done!

llvm/lib/Target/AIE/AIEBaseSubtarget.cpp
andcarminati: New checks available. We can have 1-stage pipelined loop with BotFixedBundles.

llvm/lib/Target/AIE/AIERegMemEventTracker.cpp
andcarminati: Definitely not. AIERegMemEventTracker is a register and memory event tracker so we register events in their respective cycles.

llvm/lib/Target/AIE/AIERegMemEventTracker.cpp
andcarminati: I rewrote this part for clarity.

llvm/lib/Target/AIE/AIERegMemEventTracker.cpp
andcarminati: They are opaque instructions: no operands, no memory cycles, do not load or store and they are not locks (with respective associated cycles). In essence, they are scheduling barriers.

llvm/lib/Target/AIE/AIERegMemEventTracker.cpp
andcarminati: This seems correct to me, we care about stores here.

F-Stuckmann: shouldn't we then only consider stores in the if condition?

llvm/lib/Target/AIE/AIERegMemEventTracker.cpp
```diff
@@ -40,10 +55,141 @@ void AIERegMemEventTracker::updateUseDefMaxCycle(Register Reg, unsigned Latency,
...
+    auto OptCycle =
+        InstrItins->getOperandCycle(MI.getDesc().getSchedClass(), OpNum);
+    unsigned OperandCycle = OptCycle ? *OptCycle : 0 /*implicit-def*/;
+
+    auto SafeDistance = [this, OperandCycle,
+                         IsBackward](const MachineOperand &MO, bool IsDef) {
+      const std::map<MCRegister, int> &RegToCycle =
+          IsDef ? RegisterToCycleDef : RegisterToCycleUse;
+      int CurrMaxLatency = 0;
+      for (MCRegAliasIterator Ali(MO.getReg(), TRI, true); Ali.isValid();
```
F-Stuckmann, line 161: we could also directly use RegUnits instead of going through all aliases.

llvm/lib/Target/AIE/AIERegMemEventTracker.cpp
```diff
@@ -40,10 +55,141 @@ void AIERegMemEventTracker::updateUseDefMaxCycle(Register Reg, unsigned Latency,
...
+           ++Ali) {
+        auto RegCycle = RegToCycle.find(*Ali);
+        if (RegCycle != RegToCycle.end()) {
+          const int RegCycleValue = RegCycle->second;
+          int ThisOperandLatency = IsBackward
+                                       ? RegCycleValue + OperandCycle + 1
```
F-Stuckmann: Could you add a comment why we have +1 here?
Is this the latency after which the register is free to be used again?

llvm/lib/Target/AIE/AIERegMemEventTracker.cpp
```diff
@@ -40,10 +55,141 @@ void AIERegMemEventTracker::updateUseDefMaxCycle(Register Reg, unsigned Latency,
...
+      }
+      return CurrMaxLatency;
+    };
+
+    const int DistFromLastWrite = SafeDistance(MO, /*IsDef*/ true);
+    const int DistFromLastRead =
+        MO.isDef() ? SafeDistance(MO, /*IsDef*/ false) : 0;
+
+    // Only use positive distances
+    if (DistFromLastWrite > 0)
```
F-Stuckmann: nit: std::max will filter out the case of DistFromLastWrite < 0

llvm/lib/Target/AIE/AIERegMemEventTracker.cpp
```diff
@@ -40,10 +55,141 @@ void AIERegMemEventTracker::updateUseDefMaxCycle(Register Reg, unsigned Latency,
...
+          int ThisOperandLatency = IsBackward
+                                       ? RegCycleValue + OperandCycle + 1
+                                       : RegCycleValue - OperandCycle + 1;
+          CurrMaxLatency = std::max(ThisOperandLatency, CurrMaxLatency);
+        }
+      }
+      return CurrMaxLatency;
+    };
+
+    const int DistFromLastWrite = SafeDistance(MO, /*IsDef*/ true);
```
F-Stuckmann, line 177: nit: If we unify RegisterToCycleDef and RegisterToCycleUse we only need to execute the lambda once, since the difference between this call and the one below is just a change in the datastructure we use throught the lambda.

llvm/lib/Target/AIE/AIERegMemEventTracker.cpp
F-Stuckmann: nit: we could also call checkMemoryDependency

llvm/lib/Target/AIE/AIERegMemEventTracker.cpp
F-Stuckmann: can we also refactor this to call `checkMemoryDependency` ? 
Can we harmonize the Cycles. Why dont we go to the next cycle that is free like in `checkMemoryDependency`  ?

F-Stuckmann: If i understand the code correctly, the ints in RegisterToCycleDef and MemoryCycleToStoreInstrs are complelty different, right? Could we harmonize this so that a cycle always means the same thing: 
a) Cycle before the last time it is blocked (as is used here and `MemoryCycleToStoreInstrs`/`MemoryCycleToLoadInstrs`)
b) Cycle the instruction Def/Use is last used (as in ` updateUseDefMaxCycle(MO.getReg(), EventCycle, IsDef);`)

llvm/lib/Target/AIE/AIERegMemEventTracker.cpp
```diff
@@ -73,7 +219,7 @@ void AIERegMemEventTracker::computeUseDefForward(
           int MemAccessCycle = Cycle + *OptLastMemCycle - 1;
```
F-Stuckmann: up until this point the code is identical to the `if (BundledMI->mayStore()) {` case. We could harmonize these two conditions.

llvm/lib/Target/AIE/AIERegMemEventTracker.cpp
F-Stuckmann, line 239: nit: we can hoist  ` const int EventCycle = Cycle + OperandCycle;` and use the variable in the conditional branches

llvm/lib/Target/AIE/AIERegMemEventTracker.cpp
```diff
@@ -101,109 +247,183 @@ void AIERegMemEventTracker::computeUseDefForward(
...
+
+    for (MachineInstr *BundledMI : Bundle.getInstrs()) {
+      const unsigned SchedClass = BundledMI->getDesc().getSchedClass();
+
+      // Track loads for AA-based dependencies
+      if (BundledMI->mayLoad()) {
+        auto OptFirstMemCycle = TII->getFirstMemoryCycle(SchedClass);
+        assert(OptFirstMemCycle && "Load instruction without MemoryCycles");
+        const int FirstMemCycle = *OptFirstMemCycle;
+        int FirstLoadFromEnd = Cycle - FirstMemCycle + 1;
```
F-Stuckmann: Why do I need a +1 here? 
Shouldn't the current variable be called `CycleAfterFirstLoadFromEnd`?

llvm/lib/Target/AIE/AIERegMemEventTracker.cpp
```diff
@@ -101,109 +247,183 @@ void AIERegMemEventTracker::computeUseDefForward(
...
+
+        // Track all loads, including those with negative cycles
+        updateFirstMemCycle(FirstLoadFromEnd, /*IsStore=*/false);
+        // Precise: track per-instruction for AA
+        addPerInstructionFirstMemCycle(FirstLoadFromEnd, BundledMI,
+                                       /*IsStore=*/false);
+      }
+
+      // Track stores for AA-based dependencies
+      if (BundledMI->mayStore()) {
```
F-Stuckmann: nit: unify with the `if (BundledMI->mayLoad()) {` block

llvm/lib/Target/AIE/AIERegMemEventTracker.cpp
```diff
@@ -101,109 +247,183 @@ void AIERegMemEventTracker::computeUseDefForward(
   }
 }
 
-unsigned
-AIERegMemEventTracker::getSafeOperandsDistance(const MachineInstr &MI) const {
+void AIERegMemEventTracker::computeUseDefBackward(
```
F-Stuckmann, line 253: I think we could harmonize this method and `computeUseDefForward`, since they compute similar results (except for a backward forward update of cycles)

llvm/lib/Target/AIE/AIERegMemEventTracker.cpp
```diff
@@ -101,109 +247,183 @@ void AIERegMemEventTracker::computeUseDefForward(
...
   if (MI.mayLoadOrStore()) {
     auto MemoryPipelineStage =
         MI.isBundle() ? TII->getMinFirstMemoryCycle()
                       : TII->getFirstMemoryCycle(MI.getDesc().getSchedClass());
     assert(MemoryPipelineStage.value() >= 1 && "Execution stages start at 0");
-    // Here, we calculate this initial latency.
-    // For example, if we have LastStoreCycle = 21 (absolute) and this
-    // instruction FirstMemoryCycle = 5 (relative to the start cycle of this
-    // instruction), then this instruction should start earliest at cycle 18,
-    // to be able to start the FirstMemoryCycle (absolute) at cycle 22. In
```
F-Stuckmann, line 119: nit: could you keep the comments that explain magic numbers?

llvm/lib/Target/AIE/AIERegMemEventTracker.cpp
```diff
@@ -101,109 +247,183 @@ void AIERegMemEventTracker::computeUseDefForward(
...
+    if (MI.mayStore()) {
+      // Free store: check loads in bot-fixed (WAR dependency)
+      const int LoadCycle = AA ? getMaxAliasingMemCycle(MI, /*IsStore=*/false)
+                               : getFirstMemCycle(/*IsStore=*/false);
+      MaxLatency =
+          checkMemoryDependency(MaxLatency, LoadCycle, MIMemoryCycle, true);
+    }
+
+    // Free load/store: check stores in bot-fixed (RAW dependency)
+    const int LoadStoreCycle = AA ? getMaxAliasingMemCycle(MI, /*IsStore=*/true)
```
F-Stuckmann, line 401: nit we could remove the above ` if (MI.mayStore()) {` code, if we use const bool IsStore= MI.mayStore() in these method calls

llvm/lib/Target/AIE/AIERegMemEventTracker.cpp
andcarminati: In this way it is simple, we just keep one register in the timeline. RegUnit can also work, but we need to create events for each RegUnit - add all reguinits and check all regunits after (I see a complexity increase.)

llvm/lib/Target/AIE/AIERegMemEventTracker.cpp
andcarminati: Good catch: I replaced by ` MaxLatency = std::max({MaxLatency, DistFromLastWrite, DistFromLastRead});`

llvm/lib/Target/AIE/AIERegMemEventTracker.cpp
andcarminati: But this flag is related to the events, not to MI.

llvm/lib/Target/AIE/AIERegMemEventTracker.cpp
andcarminati: We have some differences that could lead to hard to maintain conditional code.

llvm/lib/Target/AIE/AIERegMemEventTracker.cpp
andcarminati: Discussed offline.

llvm/lib/Target/AIE/AIERegMemEventTracker.cpp
andcarminati: You are right! I extended to cover the load case in a cleaner way!

andcarminati:
Hi @F-Stuckmann, thank you for the review and for catching some inconsistencies! I tried to improve the code with some refactor and harmonization, but I kept backward and forward mechanisms separated.

llvm/lib/Target/AIE/AIERegMemEventTracker.cpp
martien-de-jong: Perhaps isOpaqueScheduliingBarrier would be a better name?

llvm/lib/Target/AIE/AIERegMemEventTracker.cpp
```diff
@@ -40,10 +55,141 @@ void AIERegMemEventTracker::updateUseDefMaxCycle(Register Reg, unsigned Latency,
...
+}
+
+unsigned AIERegMemEventTracker::checkMemoryDependency(unsigned CurrentMax,
+                                                      int MemoryCycle,
+                                                      int MIMemoryCycle,
+                                                      bool IsBackward) const {
+  if (MemoryCycle > INT_MIN && MemoryCycle < INT_MAX) {
+    const int MemoryDep = IsBackward ? (MemoryCycle + MIMemoryCycle + 1)
+                                     : (MemoryCycle - MIMemoryCycle + 1);
+    return std::max(CurrentMax, (unsigned)std::max(0, MemoryDep));
```
martien-de-jong: static_cast<> ? Or, my preference, get rid of unsigned for anything that is not a size?

llvm/lib/Target/AIE/AIERegMemEventTracker.cpp
```diff
@@ -40,10 +55,141 @@ void AIERegMemEventTracker::updateUseDefMaxCycle(Register Reg, unsigned Latency,
...
+                         IsBackward](const MachineOperand &MO, bool IsDef) {
+      const std::map<MCRegister, int> &RegToCycle =
+          IsDef ? RegisterToCycleDef : RegisterToCycleUse;
+      int CurrMaxLatency = 0;
+      for (MCRegAliasIterator Ali(MO.getReg(), TRI, true); Ali.isValid();
+           ++Ali) {
+        auto RegCycle = RegToCycle.find(*Ali);
+        if (RegCycle != RegToCycle.end()) {
+          const int RegCycleValue = RegCycle->second;
+          int ThisOperandLatency = IsBackward
```
martien-de-jong: const

llvm/lib/Target/AIE/AIEBaseSubtarget.cpp
```diff
@@ -413,89 +403,105 @@ class EmitFixedSUnits : public ScheduleDAGMutation {
...
+  }
+
+  void establishSafeFreeSUToPrologueDistances(
+      ScheduleDAGInstrs *DAG, AIEPostRASchedStrategy *Scheduler,
+      const TargetRegisterInfo *TRI, const AIEBaseInstrInfo *TII,
+      const InstrItineraryData *ItinData) {
+
+    MachineBasicBlock *LoopSucc = nullptr;
+    for (MachineBasicBlock *Succ : DAG->getBB()->successors()) {
+      const BlockState &SuccBS = Scheduler->getInterBlock().getBlockState(Succ);
```
martien-de-jong, line 396: InterBlockScheduling::isPipelineLoop(Succ) ?

llvm/lib/Target/AIE/AIERegMemEventTracker.cpp
```diff
@@ -40,10 +55,141 @@ void AIERegMemEventTracker::updateUseDefMaxCycle(Register Reg, unsigned Latency,
...
+  }
+  return CurrentMax;
+}
+
 void AIERegMemEventTracker::computeUseDefForward(
     ArrayRef<AIE::MachineBundle> Bundles, bool InSeparateRegion) {
   int Cycle = 0;
   const int TotalCycles = Bundles.size();
+
+  // Track top-fixed region size (only for the first call, not separate region)
```
martien-de-jong, line 201: Why don't we call this parameter IsFirstCall, or TrackFixedSize?

llvm/lib/Target/AIE/AIERegMemEventTracker.cpp
```diff
@@ -112,21 +125,24 @@ int AIERegMemEventTracker::getMaxAliasingMemCycle(const MachineInstr &MI,
   return (MaxCycle == INT_MIN) ? 0 : MaxCycle;
 }
 
-unsigned AIERegMemEventTracker::checkMemoryDependency(unsigned CurrentMax,
-                                                      int MemoryCycle,
-                                                      int MIMemoryCycle,
-                                                      bool IsBackward) const {
+int AIERegMemEventTracker::checkMemoryDependency(int CurrentMax,
```
martien-de-jong, line 128: Yes, shun unsigned!

martien-de-jong (commented):
I would approve, but other people have had more input.

llvm/lib/Target/AIE/AIERegMemEventTracker.cpp
andcarminati: I guess this is related to before the fixup commit.

llvm/lib/Target/AIE/AIERegMemEventTracker.cpp
andcarminati: I think it is more clear, we have another users after.

llvm/lib/Target/AIE/AIEBaseSubtarget.cpp
andcarminati: I guess we don't have such function, right?

llvm/lib/Target/AIE/AIERegMemEventTracker.cpp
```diff
@@ -40,41 +55,175 @@ void AIERegMemEventTracker::updateUseDefMaxCycle(Register Reg, unsigned Latency,
...
         auto OptLastMemCycle = TII->getLastMemoryCycle(SchedClass);
-        if (OptLastMemCycle) {
-          int MemAccessCycle = Cycle + *OptLastMemCycle - 1;
-          if (InSeparateRegion)
-            MemAccessCycle = MemAccessCycle - TotalCycles;
-          updateLastMemoryAccessCycle(MemAccessCycle);
-        }
+        assert(OptLastMemCycle && "Memory instruction without MemoryCycles");
+        const int LastMemoryCycle = *OptLastMemCycle;
+        int LastMemCycle = Cycle + LastMemoryCycle - 1;
```
F-Stuckmann, line 215: CHECK: this -1 here is because the Count direction is from 0 to NInstrs ? 
and in the backward path the previous instruction is +1 because we count -NInstrs to 0 ? 

I still find the standalone +1 -1 magic numbers highly confusing. Could we wrap that in a `getFirstSafeDist(bool forward) ` ?

andcarminati: I don't think they can be considered magic numbers. -1 implies before the colliding cycle and +1 after.

llvm/lib/Target/AIE/AIERegMemEventTracker.cpp
```diff
@@ -40,41 +55,175 @@ void AIERegMemEventTracker::updateUseDefMaxCycle(Register Reg, unsigned Latency,
...
+          IsDef ? RegisterToCycleDef : RegisterToCycleUse;
+      int CurrMaxLatency = 0;
+      for (MCRegAliasIterator Ali(MO.getReg(), TRI, true); Ali.isValid();
+           ++Ali) {
+        auto RegCycle = RegToCycle.find(*Ali);
+        if (RegCycle != RegToCycle.end()) {
+          const int RegCycleValue = RegCycle->second;
+          // The +1 converts from "last cycle of overlap" to "first safe cycle".
+          // In instruction scheduling.
+          const int ThisOperandLatency = IsBackward
```
F-Stuckmann, line 168: nit: why can't we use `checkRegisterDependencies` here?

llvm/lib/Target/AIE/AIERegMemEventTracker.cpp
F-Stuckmann: I think a comment here explaining the singular -1 would already be sufficient or something like 
```
const int MemoryOperandLatency = OptLastMemCycle  - 1;
int LastMemCycle = Cycle + MemoryOperandLatency 
```

F-Stuckmann (approved):
Looks good, lets also get the refactor merged and it becomes clear what we are doing

PR #766, opened 2026-01-30
[AIEX]  Unified Opcodes across Subtargets

F-Stuckmann:
Unify Opcodes so that we now work on generic opcodes. 

@martien-de-jong I think we could also move the legalizer into tablegen.


llvm/include/llvm/Target/Target.td
```diff
@@ -651,6 +651,11 @@ class Instruction : InstructionEncoding {
   // legalized/regbankselected/selected.
   bit isPreISelOpcode = false;
 
+  // Indicates if this instruction is shared across AIE targets and should be
+  // placed in a stable enum layer (Layer 2) between target-defined generics
+  // and target-specific instructions.
+  bit isAIEShared = false;
```
martien-de-jong: You can't refer to AIE in this file.

llvm/lib/Target/AIE/AIEInstrGISel.td
```diff
@@ -60,6 +60,7 @@ let InOperandList = (ins type2:$val, type0:$base, type1:$modifier, type1:$size1,
 // after extracting the element from a vector. This gMIR is added
 // to avoid zero extending smaller size elements (S8, S16) to S32
 def G_AIE_ZEXT_EXTRACT_VECTOR_ELT : AIEGenericInstruction {
+  let AccessorMethod = "getGenericZExtExtractVectorEltOpcode";
```
martien-de-jong: Since we have unified the enumerations wrt generic opcodes, can't we just use the different namespaces interchangeably without ever using these accessors?  The existing getGenericZExtExtractVectorEltOpcode() is obsolescent, and will in due time be replaced with the equivalent AIE::G_ZEXT_EXTRACT_ELT (or whatever). 
Later we can even remove the generic opcodes from the specific enums.

llvm/lib/Target/AIE/AIEGenericOpcode.h
mludevid, line 1: Maybe move this to the next commit? Keep this as an aligning opcodes between targets commit and then in the next one add the checks that enforce this.

mludevid (commented):
Would it be possible to share the definitions of the instructions? As in instead of making sure that they are defined in all AIE targets centralize the definition and all targets include that central definition. We can still keep the check to make sure we didn't miss anything.

PR #768, opened 2026-02-02
[NFC][AIEX] Simplify fixed-point loop in WAW revertAllocation.

andcarminati:
A circular full processing of the queue without any addition to ToRevert is enough to detect the convergence.

### DIFF
--- a/llvm/lib/Target/AIE/AIEWawRegRewriter.cpp
+++ b/llvm/lib/Target/AIE/AIEWawRegRewriter.cpp
@@ -39,9 +39,9 @@
 #include "llvm/MC/MCRegister.h"
 #include "llvm/Support/Debug.h"
 #include "llvm/Support/raw_ostream.h"
-#include <list>
 #include <llvm/CodeGen/MachineBasicBlock.h>
 #include <map>
+#include <queue>
 
 using namespace llvm;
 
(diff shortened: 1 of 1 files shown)


llvm/lib/Target/AIE/AIEWawRegRewriter.cpp
```diff
@@ -451,32 +451,34 @@ void AIEWawRegRewriter::revertAllocation(OriginalAllocation &Candidates,
...
-        It = Remaining.erase(It);
-        Changed = true;
-      } else {
-        ++It;
-      }
+  unsigned SkipCounter = 0;
+  // If we add VReg to ToRevert and then we reach the same MO again without
+  // adding anything to ToRevert, we converged.
+  while (!Remaining.empty() && SkipCounter <= Remaining.size()) {
+    auto MORegPair = Remaining.front();
```
F-Stuckmann, line 460: CHECK: we copy Construct `MORegPair ` before deleting the first entry on the queue with `pop`, right?

andcarminati: Yes! It is cheap, just a pair.

F-Stuckmann (approved):

PR #769, opened 2026-02-02
[AIEX] Remove RegClassSuccess from WAW Rewriter

F-Stuckmann:
If Register Reallocation fails we precisly only revert the registers that conflict, therefore we can delete `RegClassSuccess` .


andcarminati (approved):
LGTM.

PR #772, opened 2026-02-02
[AIEX][AIE2P] Several new combiners to simplify vector operations

andcarminati:
The combiner eliminates a VSHIFT chain and replaces it with a simpler G_CONCAT_VECTORS operation with zero-padded upper half. The transformation is generalized to work with any vector size by calculating expected shift amounts based on source and destination types (e.g., for 512 bit expansion: shift1=16 bytes, shift2=48 bytes). The combiner includes usage-aware optimization: it analyzes how the result vector is actually used (via a new getMaxUsedVectorElement utility) and chooses between padding with UNDEF or zeros.

This PR also includes two redundant pad/unpad removal combiners.


andcarminati:
QoR:
image

llvm/test/CodeGen/AIE/GlobalISel/combine-vshift-chain-to-zero-pad.mir
F-Stuckmann, line 1: Can you move this unit-tests before the combiner implementation so that we have a before and after view?

llvm/lib/Target/AIE/AIECombinerHelper.cpp
```diff
@@ -1640,6 +1640,169 @@ void llvm::applyUnpadVector(MachineInstr &MI, MachineRegisterInfo &MRI,
...
+  unsigned MaxElement = 0;
+
+  for (const MachineInstr &User : MRI.use_nodbg_instructions(Reg)) {
+    const unsigned Opcode = User.getOpcode();
+
+    // VSEL: Check which elements are selected via the mask
+    if (Opcode == TII.getGenericVSelOpcode()) {
+      const Register MaskReg = User.getOperand(3).getReg();
+      const auto MaskVal = getIConstantVRegVal(MaskReg, MRI);
+      if (MaskVal) {
```
F-Stuckmann: we can reduce nesting if we directly reutn std::nullopt if !MaskVal

llvm/lib/Target/AIE/AIECombinerHelper.cpp
```diff
@@ -1640,6 +1640,169 @@ void llvm::applyUnpadVector(MachineInstr &MI, MachineRegisterInfo &MRI,
...
+llvm::getMaxUsedVectorElement(Register Reg, MachineRegisterInfo &MRI,
+                              const AIEBaseInstrInfo &TII) {
+
+  unsigned MaxElement = 0;
+
+  for (const MachineInstr &User : MRI.use_nodbg_instructions(Reg)) {
+    const unsigned Opcode = User.getOpcode();
+
+    // VSEL: Check which elements are selected via the mask
+    if (Opcode == TII.getGenericVSelOpcode()) {
```
F-Stuckmann: We may want to move these big if conditions into lambdas, so that the high level flow of the function is clearer

llvm/lib/Target/AIE/AIECombinerHelper.cpp
```diff
@@ -1640,6 +1640,169 @@ void llvm::applyUnpadVector(MachineInstr &MI, MachineRegisterInfo &MRI,
...
+      const Register MaskReg = User.getOperand(3).getReg();
+      const auto MaskVal = getIConstantVRegVal(MaskReg, MRI);
+      if (MaskVal) {
+        // Count the highest set bit in the mask
+        const uint64_t Mask = MaskVal->getZExtValue();
+        if (Mask != 0) {
+          const unsigned HighestBit = 63 - llvm::countl_zero(Mask);
+          MaxElement = std::max(MaxElement, HighestBit);
```
F-Stuckmann: do we expect multiple vsels in the register users?

llvm/test/CodeGen/AIE/GlobalISel/combine-vshift-chain-to-zero-pad.mir
andcarminati: Yes! Done!

llvm/lib/Target/AIE/AIECombinerHelper.cpp
```diff
@@ -1640,6 +1640,169 @@ void llvm::applyUnpadVector(MachineInstr &MI, MachineRegisterInfo &MRI,
...
+      continue;
+    }
+
+    // EXTRACT_VECTOR_ELT: Direct element access
+    if (Opcode == TargetOpcode::G_EXTRACT_VECTOR_ELT ||
+        isGenericExtractOpcode(Opcode, TII)) {
+      const Register IdxReg = User.getOperand(2).getReg();
+      const auto Idx = getIConstantVRegVal(IdxReg, MRI);
+      if (Idx) {
+        MaxElement =
```
F-Stuckmann: why don't we return here? why should we have multiple G_EXTRACT_VECTOR_ELT s coupled with G_UNMERGE_VALUES and return the maximum? aren't we then confusing combine patterns?

andcarminati: You can see the name of the function, this is just a helper function that we use to try prove that the upper part of a vector is not used, so we cannot just return max here because we may have other users accessing even higher elements. But, maybe you are talking about an early return in case if `!Idx` - this I agree, it will be part of the next version.

llvm/lib/Target/AIE/AIECombinerHelper.cpp
andcarminati: We look to all the users, it is too restrictive to assume just one user.

llvm/lib/Target/AIE/AIECombinerHelper.cpp
andcarminati: Yeah, definitely.

llvm/lib/Target/AIE/AIECombinerHelper.cpp
```diff
@@ -1640,6 +1640,181 @@ void llvm::applyUnpadVector(MachineInstr &MI, MachineRegisterInfo &MRI,
...
+/// results in padding with zeros.
+/// Transforms:
+///   %concat = G_CONCAT_VECTORS %src(<8 x s32>), %A(<8 x s32>)
+///   %shift1 = G_AIE_VSHIFT_RIGHT %concat, %B, 16
+///   %shift2 = G_AIE_VSHIFT_RIGHT %C, %shift1, 48
+/// Into:
+///   %zero_broadcast = G_AIE_BROADCAST_VECTOR %zero_scalar
+///   %zero_lo, %zero_hi = G_UNMERGE_VALUES %zero_broadcast
+///   %result = G_CONCAT_VECTORS %src, %zero_hi
+bool llvm::matchVShiftChainToZeroPad(MachineInstr &MI, MachineRegisterInfo &MRI,
```
F-Stuckmann: wouldn't it be more generic to first combine two vshifts into a single one and in the second stop combine shift to a unmerge?

llvm/lib/Target/AIE/AIECombinerHelper.cpp
```diff
@@ -1640,6 +1730,586 @@ void llvm::applyUnpadVector(MachineInstr &MI, MachineRegisterInfo &MRI,
...
+
+  // Calculate expected shift amounts based on vector sizes
+  const LLT DataSrcTy = MRI.getType(DataSrc);
+  const unsigned SrcSizeInBytes = DataSrcTy.getSizeInBits() / 8;
+
+  // Calculate expected shift amounts:
+  // First shift: adds padding (quarter of source size)
+  const unsigned ExpectedShift1 = SrcSizeInBytes / 4;
+  // Second shift: brings in zeros (1.5x source size)
+  const unsigned ExpectedShift2 = (SrcSizeInBytes + SrcSizeInBytes / 2) / 2;
```
F-Stuckmann: nit: I think it is sufficient to check the total shift size, not some predetermined shift amounts.

llvm/lib/Target/AIE/AIECombinerHelper.cpp
andcarminati: I know, this combiner is not intuitive ;-). The two shifts cannot be combined into one because:

1. They use different source vectors in different positions
2. The pattern computes a half-and-half split, not a simple shift
3. VSHIFT is a sliding window operation, not a selection operation
4. The actual semantic is "pad with zeros", which CONCAT expresses better

As an example of the operation of this combiner:

```
Original src:  [0,1,2,3,4,5,6,7,8,9,10,11,12,13,14,15]

After VSHIFT chain:
[0,1,2,3,4,5,6,7, 0,0,0,0,0,0,0,0]
 └─ from src ──┘  └─── zeros ────┘
```
This is exactly: CONCAT(UNPAD(src), zeros)

llvm/lib/Target/AIE/AIECombinerHelper.cpp
andcarminati: I am taking a look.

llvm/test/CodeGen/AIE/aie2p/shufflevec.ll
```diff
@@ -4,7 +4,7 @@
 ; See https://llvm.org/LICENSE.txt for license information.
 ; SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
 ;
-; (c) Copyright 2025 Advanced Micro Devices, Inc. or its affiliates
+; (c) Copyright 2025-2025 Advanced Micro Devices, Inc. or its affiliates
```
F-Stuckmann: nit: this line can remain unchanged for now

llvm/test/CodeGen/AIE/GlobalISel/combine-vshift-chain-to-zero-pad.mir
```diff
@@ -0,0 +1,409 @@
...
+# This file is licensed under the Apache License v2.0 with LLVM Exceptions.
+# See https://llvm.org/LICENSE.txt for license information.
+# SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+#
+# (c) Copyright 2026 Advanced Micro Devices, Inc. or its affiliates
+# RUN: llc -mtriple aie2p -run-pass=aie2p-prelegalizer-combiner %s \
+# RUN:         -verify-machineinstrs -o - | FileCheck %s
+
+# Tests for VSHIFT chain combiner that optimizes VSHIFT sequences into
+# CONCAT operations with usage-aware padding (UNDEF vs zeros).
```
F-Stuckmann: nit: also add a unit-test with different shift constants

llvm/lib/Target/AIE/AIECombinerHelper.cpp
```diff
@@ -1640,6 +1640,127 @@ void llvm::applyUnpadVector(MachineInstr &MI, MachineRegisterInfo &MRI,
...
+  } else if (DataSrcMI->getOpcode() == TII.getGenericPadVectorOpcode()) {
+    // Pattern: G_AIE_PAD_VECTOR_UNDEF %data(<8 x s32>)
+
+    // Get the actual data source from PAD
+    ActualDataSrc = DataSrcMI->getOperand(1).getReg();
+    const LLT ActualDataTy = MRI.getType(ActualDataSrc);
+
+    // Verify types match expected pattern: <8 x s32> → <16 x s32>
+    if (ActualDataTy.getSizeInBits() != DataSrcTy.getSizeInBits() / 2 ||
+        DataSrcTy.getSizeInBits() != DataSrcTy.getSizeInBits())
```
martien-de-jong: This looks strange. Typo?

llvm/lib/Target/AIE/AIECombinerHelper.cpp
```diff
@@ -1761,6 +1761,94 @@ bool llvm::matchVShiftChainToCopy(MachineInstr &MI, MachineRegisterInfo &MRI,
...
+///   %dst:_(<8 x s32>) = COPY %src(<8 x s32>)
+bool llvm::matchPadUnpadToCopy(MachineInstr &MI, MachineRegisterInfo &MRI,
+                               const AIEBaseInstrInfo &TII,
+                               BuildFnTy &MatchInfo) {
+  assert(MI.getOpcode() == TII.getGenericUnpadVectorOpcode() &&
+         "Expected G_AIE_UNPAD_VECTOR");
+
+  // Get the source of the unpad operation
+  Register UnpadDst = MI.getOperand(0).getReg();
+  Register UnpadSrc = MI.getOperand(1).getReg();
```
martien-de-jong, line 1933: lots of const?

llvm/lib/Target/AIE/AIECombinerHelper.cpp
```diff
@@ -1761,6 +1761,94 @@ bool llvm::matchVShiftChainToCopy(MachineInstr &MI, MachineRegisterInfo &MRI,
...
+  // Type validation: ensure input of PAD matches output of UNPAD
+  LLT PadSrcTy = MRI.getType(PadSrc);
+  LLT UnpadDstTy = MRI.getType(UnpadDst);
+
+  if (PadSrcTy != UnpadDstTy)
+    return false;
+
+  // Verify types are legal for these operations
+  LLT PadDstTy = MRI.getType(PadMI->getOperand(0).getReg());
+  if (!TII.isLegalTypeToPad(PadSrcTy) || !TII.isLegalTypeToUnpad(PadDstTy))
```
martien-de-jong: I'm not quite sure why we check legality of both operations. We're replacing one by a copy, only the other remains posing legality requirements.

llvm/lib/Target/AIE/AIECombinerHelper.cpp
```diff
@@ -1761,6 +1761,94 @@ bool llvm::matchVShiftChainToCopy(MachineInstr &MI, MachineRegisterInfo &MRI,
...
+
+/// Match G_AIE_PAD_VECTOR_UNDEF fed by G_AIE_UNPAD_VECTOR and combine to COPY.
+/// Transforms:
+///   %unpadded:_(<4 x s32>) = G_AIE_UNPAD_VECTOR %src(<16 x s32>)
+///   %dst:_(<16 x s32>) = G_AIE_PAD_VECTOR_UNDEF %unpadded(<4 x s32>)
+/// Into:
+///   %dst:_(<16 x s32>) = COPY %src(<16 x s32>)
+bool llvm::matchUnpadPadToCopy(MachineInstr &MI, MachineRegisterInfo &MRI,
+                               const AIEBaseInstrInfo &TII,
+                               BuildFnTy &MatchInfo) {
```
martien-de-jong, line 2240: I'm wondering whether we could generalize to matchCancellingUnaryOpsToCopy()

andcarminati: I would like to keep all combiners simple as much as possible. I would like to avoid the complexity that we have, for example, in the shuffle combiners.

llvm/lib/Target/AIE/AIECombinerHelper.cpp
```diff
@@ -419,6 +419,96 @@ bool isUseOf(const MachineInstr &MI, const MachineInstr &Def) {
   return false;
 }
 
+bool llvm::matchUnpadUnmerge(MachineInstr &UnpadMI, MachineRegisterInfo &MRI,
```
martien-de-jong, line 436: Can I have a short description of what we're trying to match?

llvm/lib/Target/AIE/AIECombinerHelper.cpp
```diff
@@ -419,6 +419,96 @@ bool isUseOf(const MachineInstr &MI, const MachineInstr &Def) {
...
+    return false;
+
+  // Collect all UNMERGE output registers
+  SmallVector<Register, 4> UnmergeDsts;
+  for (unsigned I = 0; I < NumUnmergeOutputs; ++I) {
+    UnmergeDsts.push_back(UnmergeMI->getOperand(I).getReg());
+  }
+
+  // 4. Check dominance (one must dominate the other)
+  const bool UnpadDominates = Helper.dominates(UnpadMI, *UnmergeMI);
```
martien-de-jong, line 476: I suggest to setup a dedicated transform lambda and return if it dominates. Otherwise, check the reverse dominance and setup the reverse transform lambda. We don't need to check things we don't need, and we don't need to capture unnecessary data in the lambda.

llvm/lib/Target/AIE/AIECombinerHelper.cpp
```diff
@@ -1950,6 +1950,67 @@ bool llvm::matchPadUnpadFusion(MachineInstr &MI, MachineRegisterInfo &MRI,
...
+  // This ensures the UNPAD output size is an exact multiple of each CONCAT
+  // operand size, so we can cleanly extract complete operands.
+  // Example: If UNPAD outputs 8 elements and each CONCAT operand has 4
+  //          elements, then 8 % 4 = 0 (aligned), and we need exactly 2
+  //          operands.
+  // Counter-example: If UNPAD outputs 8 elements but each CONCAT operand has
+  //                  3 elements, then 8 % 3 = 2 (NOT aligned), and we can't
+  //                  cleanly extract 8 elements using 3-element chunks.
+  if (UnpadElements % ConcatOpElements != 0)
+    return false;
```
martien-de-jong, line 2086: Do we have automatic guarantees about matching element types?

llvm/lib/Target/AIE/AIECombinerHelper.cpp
```diff
@@ -2012,6 +2021,126 @@ bool llvm::matchConcatUnpadFusion(MachineInstr &MI, MachineRegisterInfo &MRI,
...
+  // Process each operand of the outer CONCAT
+  for (unsigned I = 1; I < MI.getNumOperands(); ++I) {
+    const Register OpReg = MI.getOperand(I).getReg();
+    MachineInstr *OpMI = MRI.getVRegDef(OpReg);
+    const LLT OpTy = MRI.getType(OpReg);
+    const unsigned OpElements = OpTy.getNumElements();
+
+    if (OpMI->getOpcode() == TargetOpcode::G_CONCAT_VECTORS) {
+      // Nested CONCAT - flatten by extracting all its operands
+      for (unsigned J = 1; J < OpMI->getNumOperands(); ++J) {
```
martien-de-jong, line 2148: nit: I think we could match recursively, but there's probably very little return on investment.

llvm/lib/Target/AIE/AIECombinerHelper.cpp
```diff
@@ -2012,6 +2021,126 @@ bool llvm::matchConcatUnpadFusion(MachineInstr &MI, MachineRegisterInfo &MRI,
...
+      TotalElements += PadSrcElements;
+
+      // Calculate padding in terms of sub-vectors
+      // The sub-vector size is the PAD source size
+      const unsigned SubVecElements = PadSrcElements;
+      const unsigned NumSubVecs = OpElements / SubVecElements;
+      const unsigned PaddingSubVecs = NumSubVecs - 1;
+
+      for (unsigned K = 0; K < PaddingSubVecs; ++K) {
+        FlattenedOps.push_back(
```
martien-de-jong, line 2173: FlattenedOps.resize(size+PaddingSubVecs?)

llvm/lib/Target/AIE/AIECombinerHelper.cpp
```diff
@@ -2012,6 +2021,126 @@ bool llvm::matchConcatUnpadFusion(MachineInstr &MI, MachineRegisterInfo &MRI,
...
+    return false;
+
+  // Verify total elements match destination
+  if (TotalElements != DstElements)
+    return false;
+
+  // Verify all valid operands in FlattenedOps have the same type
+  // and ensure we have at least one valid operand
+  LLT SubVecTy;
+  bool FoundValidOp = false;
```
martien-de-jong: `optional<LLT>` ?

llvm/lib/Target/AIE/AIECombinerHelper.cpp
```diff
@@ -2450,6 +2450,83 @@ void llvm::applyUnmergeConcat(MachineInstr &MI, MachineRegisterInfo &MRI,
...
+      continue;
+
+    // Check if all input operands match (starting from operand 1)
+    bool AllMatch = true;
+    for (unsigned i = 1; i < NumOps; ++i) {
+      if (UserMI.getOperand(i).getReg() != MI.getOperand(i).getReg()) {
+        AllMatch = false;
+        break;
+      }
+    }
```
martien-de-jong: If you move the opcode check down, you could bundle this in a nice generic isEquivalentMI(I1, I2) helper.

martien-de-jong (dismissed):
I'm getting into real nitpicking territory. I see nothing really blocking, it's time it went in.

llvm/lib/Target/AIE/AIECombinerHelper.cpp
andcarminati: I will keep as is this part.

llvm/lib/Target/AIE/AIECombinerHelper.cpp
andcarminati: I guess yes, but I prefer to be defensive here.

llvm/lib/Target/AIE/AIECombinerHelper.cpp
andcarminati: I prefer to keep it explicitly to self document what we are doing here.

llvm/lib/Target/AIE/AIECombinerHelper.cpp
andcarminati: Nice suggestion.

llvm/lib/Target/AIE/AIECombinerHelper.cpp
andcarminati: Done!

martien-de-jong (dismissed):

martien-de-jong (approved):

PR #774, opened 2026-02-04
Revert "[ORC][LLJIT] Install eh-frame registration plugin during plat…

jorickert:
…form setup."

This reverts commit d0052ebbe2e2f691ec42cad3c8613ef387abc53f which was/is causing downstream test failure

### DIFF
--- a/llvm/lib/ExecutionEngine/Orc/LLJIT.cpp
+++ b/llvm/lib/ExecutionEngine/Orc/LLJIT.cpp
@@ -833,7 +833,14 @@ Error LLJITBuilderState::prepareForConstruction() {
       CreateObjectLinkingLayer =
           [](ExecutionSession &ES,
              const Triple &) -> Expected<std::unique_ptr<ObjectLayer>> {
-        return std::make_unique<ObjectLinkingLayer>(ES);
+        auto ObjLinkingLayer = std::make_unique<ObjectLinkingLayer>(ES);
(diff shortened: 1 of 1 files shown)


F-Stuckmann (dismissed):

mludevid:
Do we understand why this commit is breaking our downstream test? Is this our problem or is this a problem with the upstream commit? And in the second case: how was this fixed upstream? If there is a fixup commit upstream we should cherry-pick that rather than reverting.

jorickert:
> Do we understand why this commit is breaking our downstream test? Is this our problem or is this a problem with the upstream commit? And in the second case: how was this fixed upstream? If there is a fixup commit upstream we should cherry-pick that rather than reverting.

I agree that we should find out whats the root cause of the test failures. I tried to cherry-pick more commits around the problematic commit https://github.com/llvm/llvm-project/commits/main/llvm/lib/ExecutionEngine/Orc/LLJIT.cpp but did not find one that was solving the issue before hitting merge conflicts.
As i want to quickly unblock CI, i created this revert

mludevid:
If we cannot find the root cause of the test failure, I'd prefer marking the test as XFAIL on our end rather than reverting the entire commit. I feel like reverting is asking for trouble. Future LLVM bumps will bring in further changes to this file which will lead to merge conflicts if they modify the same part of the code as the commit we are reverting.

mludevid:
Discussed offline: we're going for the XFAIL solution. @F-Stuckmann is working on that PR. Closing this one.

PR #776, opened 2026-02-05
[ORC][LLI] Remove redundant eh-frame registration plugin construction…

F-Stuckmann:
… from lli.

As of d0052ebbe2e the setUpGenericLLVMIRPlatform function will automatically add an instance of the EHFrameRegistrationPlugin (for LLJIT instances whose object linking layers are ObjectLinkingLayers, not RTDyldObjectLinkingLayers).

This commit removes the redundant plugin creation in the object linking layer constructor function in lli.cpp to prevent duplicate registration of eh-frames, which is likely the cause of recent bot failures, e.g. https://lab.llvm.org/buildbot/#/builders/108/builds/8685.


andcarminati (approved):
LGTM.

PR #777, opened 2026-02-05
[AIE2P] Vshuffle scalar crash fix

F-Stuckmann:
fix crashing vshuffles when encountering scalars.


llvm/lib/Target/AIE/AIECombinerHelper.cpp
```diff
@@ -2178,8 +2194,13 @@ bool llvm::matchBroadcastElement(MachineInstr &MI, MachineRegisterInfo &MRI,
   if (!MaybeSplatIndex.has_value())
     return false;
 
-  const unsigned SrcNumElems =
-      MRI.getType(MI.getOperand(1).getReg()).getNumElements();
+  // This combine looks for G_BUILD_VECTOR sources, so bail out for scalar
+  // sources (e.g., from <1 x i64> lowered to s64).
+  const LLT SrcTy = MRI.getType(MI.getOperand(1).getReg());
+  if (!SrcTy.isVector())
+    return false;
```
khallouh: I opened this PR https://github.com/Xilinx/llvm-aie/pull/778 to extend it. If we bail out we hit the same issue in subsequent combines which you are handling below.

llvm/lib/Target/AIE/AIECombinerHelper.cpp
```diff
@@ -2170,6 +2170,22 @@ bool llvm::tryToCombineVectorShiftsByZero(MachineInstr &MI,
   return true;
 }
 
+/// Match a G_SHUFFLE_VECTOR that splats a single element from a G_BUILD_VECTOR
+/// source. This enables replacing the shuffle with a broadcast of that element.
+///
+/// Pattern matched:
+///   %src:_(<N x sM>) = G_BUILD_VECTOR %e0, %e1, ..., %eK
+///   %dst:_(<N x sM>) = G_SHUFFLE_VECTOR %src, %undefined, shufflemask(K, K, K,
```
khallouh: Nit: It doesn't have to be G_BUILD_VECTOR when the source is a scalar

llvm/test/CodeGen/AIE/GlobalISel/prelegalizercombiner-shuffle-scalar-src.mir
```diff
@@ -0,0 +1,72 @@
...
+# (c) Copyright 2025 Advanced Micro Devices, Inc. or its affiliates
+#
+# Test that G_SHUFFLE_VECTOR with a scalar source (from <1 x i64>) is handled
+# in the prelegalizer combiner. The <1 x i64> type is lowered to scalar s64 in
+# GlobalISel. Without proper handling, calling getNumElements() on a scalar
+# triggers an assertion failure.
+#
+# FIXME: These tests currently crash due to missing scalar source handling.
+# RUN: not --crash llc -mtriple aie2p -run-pass=aie2p-prelegalizer-combiner \
+# RUN:     -verify-machineinstrs %s -o /dev/null 2>&1 | FileCheck %s --check-prefix=CHECK-CRASH
```
khallouh: Do we still have the crash with my update?

llvm/lib/Target/AIE/AIECombinerHelper.cpp
```diff
@@ -2170,6 +2170,22 @@ bool llvm::tryToCombineVectorShiftsByZero(MachineInstr &MI,
   return true;
 }
```
khallouh, line 2172: Can you update the commit message? we don't bail out for scalar sources in matchBroadcastElement anymore

llvm/lib/Target/AIE/AIECombinerHelper.cpp
```diff
@@ -2341,12 +2357,25 @@ static bool matchShuffleToVecEltBroadcast(MachineInstr &MI,
...
-    const LLT DstElemTy = MRI.getType(SrcVecReg).getElementType();
+  const Register DstReg = MI.getOperand(0).getReg();
+  const Register SrcReg = MI.getOperand(1).getReg();
+  const LLT SrcTy = MRI.getType(SrcReg);
+
+  // Handle scalar sources (e.g., from <1 x i64> lowered to s64).
+  // For scalar, only element 0 exists, so broadcast it directly.
+  if (!SrcTy.isVector()) {
+    if (*UniqOpIdx != 0)
+      return false;
```
khallouh, line 2374: What if the second one is scalar? we could still combine in that case.

llvm/test/CodeGen/AIE/GlobalISel/prelegalizercombiner-shuffle-scalar-src.mir
khallouh: I checked out that commit and the test fails, could you take a look.

llvm/lib/Target/AIE/AIECombinerHelper.cpp
F-Stuckmann: we handle that case in matchShuffleToVecBroadcast

llvm/test/CodeGen/AIE/GlobalISel/prelegalizercombiner-shuffle-scalar-src.mir
F-Stuckmann: your PR fixed the crashes, I will update the first commit

llvm/lib/Target/AIE/AIECombinerHelper.cpp
F-Stuckmann: I also added a unit-test for that case

khallouh (approved):
LGTM

PR #779, opened 2026-02-05
[AIEX] FIX: commit MBB regions in top-down order 

andcarminati:
This fix a scheduler bug.


llvm/lib/Target/AIE/AIEMachineScheduler.cpp
```diff
@@ -762,22 +781,22 @@ void AIEPostRASchedStrategy::commitBlockSchedule(MachineBasicBlock *BB) {
...
-          TII->insertNoop(*BB, It);
-          continue;
-        }
-        It = getBundleEnd(Bundle.getInstrs().back()->getIterator());
-      }
+      // NOPs. That isn't a problem, since the callbacks to the
+      // HazardRecognizer were a bit flaky (e.g. when to call emitNoop vs
+      // advanceCycle). MachineScheduler just calls advanceCycle, and this is
+      // enough for us to insert NOPs because the sequence of Bundles gives us
+      // the full picture.
```
martien-de-jong, line 797: I would rephrase this, not referring to PRAS. Just say that empty bundles represent cycles that need to be materialized as a single nop.

martien-de-jong (approved):
Just a comment that's past its best-before date.

PR #780, opened 2026-02-06
[AIEX] Loop Strength Reduction

F-Stuckmann:
Add support for i20 Loop strength reduction.
Overall performance improvement of -0.03%.
PM: 2.26 Bytes
Stack: 0.8 Bytes

image

image

image


llvm/include/llvm/Analysis/IVUsers.h
```diff
@@ -106,11 +111,11 @@ class IVUsers {
 
 public:
   IVUsers(Loop *L, AssumptionCache *AC, LoopInfo *LI, DominatorTree *DT,
-          ScalarEvolution *SE);
+          ScalarEvolution *SE, const TargetTransformInfo *TTI = nullptr);
 
   IVUsers(IVUsers &&X)
       : L(std::move(X.L)), AC(std::move(X.AC)), DT(std::move(X.DT)),
-        SE(std::move(X.SE)), Processed(std::move(X.Processed)),
+        SE(std::move(X.SE)), TTI(X.TTI), Processed(std::move(X.Processed)),
```
martien-de-jong: All other members use std::move, TTI not. Is that deliberate?

llvm/include/llvm/Analysis/IVUsers.h
```diff
@@ -124,7 +129,7 @@ class IVUsers {
   /// AddUsersIfInteresting - Inspect the specified Instruction.  If it is a
   /// reducible SCEV, recursively add its users to the IVUsesByStride set and
   /// return true.  Otherwise, return false.
-  bool AddUsersIfInteresting(Instruction *I);
+  bool AddUsersIfInteresting(Instruction *I, bool BypassWidthCheck = false);
```
martien-de-jong, line 133: Document new parameter?

llvm/include/llvm/Analysis/TargetTransformInfo.h
```diff
@@ -798,6 +800,27 @@ class TargetTransformInfo {
   AddressingModeKind getPreferredAddressingMode(const Loop *L,
                                                 ScalarEvolution *SE) const;
 
+  /// Return true if IVUsers should look through the instruction to collect
+  /// its users instead. If true, populates GEPsToProcess with the GEP
+  /// instructions that should be processed as IV users.
+  /// This is useful for targets with non-native pointer sizes (e.g., targets
```
martien-de-jong: nit: I don't know what you mean by `non-native`. I think it would be clearer if you describe the difference between pointer and int bit sizes.

andcarminati:
I think it would be nice if you could explain the differences between this approach and the previous attempts.

llvm/lib/Analysis/IVUsers.cpp
```diff
@@ -153,9 +157,15 @@ bool IVUsers::AddUsersIfInteresting(Instruction *I) {
   // LSR is not APInt clean, do not touch integers bigger than 64-bits.
   // Also avoid creating IVs of non-native types. For example, we don't want a
   // 64-bit IV in 32-bit code just because the loop has one 64-bit cast.
```
martien-de-jong: Ah here is where non-native comes from. But this is the other way. 64 is non-native since it would need multi-word emulation on a 32 bit ISA. Our 20 bit pointers and offsets are extremely native, they just don't mesh with the standard integer sizes.

llvm/lib/Analysis/IVUsers.cpp
```diff
@@ -170,6 +180,19 @@ bool IVUsers::AddUsersIfInteresting(Instruction *I) {
...
+  // non-standard index sizes) to collect their users instead. This enables LSR
+  // to create pointer PHIs for targets with non-native pointer sizes.
+  SmallVector<GetElementPtrInst *, 4> GEPsToProcess;
+  if (TTI && TTI->shouldIVUsersLookThroughInst(I, GEPsToProcess)) {
+    LLVM_DEBUG(dbgs() << "Looking through instruction: " << *I << '\n');
+    bool AnyInteresting = false;
+    for (GetElementPtrInst *GEP : GEPsToProcess)
+      if (AddUsersIfInteresting(GEP, /*BypassWidthCheck=*/true))
+        AnyInteresting = true;
+    return AnyInteresting;
```
martien-de-jong: return llvm::any_of(GEPsToProcess, []....) ?

llvm/lib/Target/AIE/AIEBaseTargetTransformInfo.h
```diff
@@ -57,13 +57,37 @@ template <typename T> class AIEBaseTTIImpl : public BasicTTIImplBase<T> {
   /// Helper function to access this as a T.
   T *thisT() { return static_cast<T *>(this); }
 
+  //===--------------------------------------------------------------------===//
+  // Helpers for Loop Strength Reduction
+  //===--------------------------------------------------------------------===//
+
```
martien-de-jong: why not group them with LSR below?

llvm/lib/Target/AIE/AIEBaseTargetTransformInfo.h
```diff
@@ -114,6 +147,78 @@ template <typename T> class AIEBaseTTIImpl : public BasicTTIImplBase<T> {
...
   }
+
+  //===--------------------------------------------------------------------===//
+  // Loop Strength Reduction (LSR)
+  //
+  // AIE has 20-bit pointers but 32-bit integers, and post-increment load/store
+  // instructions (VLD_pstm, VST_pstm). These hooks enable LSR to generate
+  // pointer recurrences that the backend can combine with post-increment ops.
+  //===--------------------------------------------------------------------===//
+
```
martien-de-jong: Could we move the implementations to the cpp file? The helpers can live there too.

llvm/lib/Target/AIE/AIEBaseTargetTransformInfo.h
```diff
@@ -114,6 +147,78 @@ template <typename T> class AIEBaseTTIImpl : public BasicTTIImplBase<T> {
...
+  bool isIndexedStoreLegal(TTI::MemIndexedMode Mode, Type *Ty,
+                           const DataLayout &DL) const {
+    return Mode == TTI::MIM_PostInc && isIndexSizedInteger(Ty, DL);
+  }
+
+  /// Look through truncs to index size that feed GEP indices.
+  ///
+  /// Array indexing generates: %trunc = trunc i32 %idx to i20
+  /// Without this hook, IVUsers stops at the trunc (i20 not legal).
+  /// With this hook, IVUsers continues to the GEP, collecting pointer SCEVs.
```
martien-de-jong: nit: `IVUsers()` to make it singular for the benefit of grammatical congruence?

llvm/lib/Target/AIE/AIEBaseTargetTransformInfo.h
```diff
@@ -114,6 +147,78 @@ template <typename T> class AIEBaseTTIImpl : public BasicTTIImplBase<T> {
...
+  bool shouldIVUsersLookThroughInst(
+      Instruction *I,
+      SmallVectorImpl<GetElementPtrInst *> &GEPsToProcess) const {
+    auto *Trunc = dyn_cast<TruncInst>(I);
+    if (!Trunc)
+      return false;
+
+    const DataLayout &DL = Trunc->getModule()->getDataLayout();
+    const unsigned TruncWidth = Trunc->getType()->getIntegerBitWidth();
```
martien-de-jong: Check: TruncInst always has an integer type.

llvm/lib/Target/AIE/AIEBaseTargetTransformInfo.h
```diff
@@ -114,6 +147,78 @@ template <typename T> class AIEBaseTTIImpl : public BasicTTIImplBase<T> {
...
+    const unsigned IndexWidth = DL.getIndexSizeInBits(0);
```
martien-de-jong: Check: 0 is the addrspace, and we assume that all addrspaces have the same index width

llvm/lib/Target/AIE/AIEBaseTargetTransformInfo.h
```diff
@@ -114,6 +147,78 @@ template <typename T> class AIEBaseTTIImpl : public BasicTTIImplBase<T> {
...
+
+    if (TruncWidth != IndexWidth || DL.isLegalInteger(TruncWidth))
+      return false;
+
+    if (!allUsesAreGEPIndices(Trunc))
+      return false;
+
+    for (User *U : Trunc->users())
+      GEPsToProcess.push_back(cast<GetElementPtrInst>(U));
```
martien-de-jong: I wouldn't mind passing GEPsYoProcess to allUsesAreGEPIndices to collect in one traversal

llvm/lib/Transforms/Scalar/LoopStrengthReduce.cpp
```diff
@@ -3896,21 +3896,43 @@ static const SCEV *CollectSubexprs(const SCEV *S, const SCEVConstant *C,
...
-    return false;
+  // For Basic uses with preserved recurrence types, allow loop-invariant steps.
+  // For standard uses, require constant steps to preserve original behavior.
+  if (IsBasicUseWithPreservedType) {
+    if (!isa<SCEVConstant>(LoopStep) && !SE.isLoopInvariant(LoopStep, L))
+      return false;
+  } else {
+    if (!isa<SCEVConstant>(LoopStep))
+      return false;
+  }
```
martien-de-jong: nit: isa<SCEVConstant)() can be factored out, better maintaining the original code structure.

martien-de-jong (commented):
Nice! I'm curious whether anyone would notice if we didn't add the TTI hooks, but would mimic the current implementation by checking the trunc's width against the pointer width.

F-Stuckmann:
If we don't implement the hooks, we get offset ptr_adds inserted into the loops, because LSR does not like post incrementing SCEVs by default

llvm/lib/Transforms/Scalar/LoopStrengthReduce.cpp
F-Stuckmann: I complelty removed this part :)

llvm/lib/Target/AIE/AIEBaseTargetTransformInfo.h
F-Stuckmann: we now check it

F-Stuckmann:
We now have minimal changes upstream. 
We however have to tell ptr chaining to handle raw pointer load/stores and chained pointer loads/stores delicately, so that we minimize ptr copies and ptr_add with zeros.

llvm/lib/Transforms/Scalar/LoopStrengthReduce.cpp
```diff
@@ -4,7 +4,7 @@
 // See https://llvm.org/LICENSE.txt for license information.
 // SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
 //
-// Modifications (c) Copyright 2023-2024 Advanced Micro Devices, Inc. or its
+// Modifications (c) Copyright 2023-2026 Advanced Micro Devices, Inc. or its
```
andcarminati: Maybe we could restore this file.

llvm/lib/Target/AIE/AIEBaseTargetTransformInfo.h
F-Stuckmann: yes

llvm/include/llvm/Analysis/TargetTransformInfo.h
```diff
@@ -798,6 +800,22 @@ class TargetTransformInfo {
   AddressingModeKind getPreferredAddressingMode(const Loop *L,
                                                 ScalarEvolution *SE) const;
 
+  /// Return true if IVUsers() should look through the instruction to collect
+  /// its users instead. If true, populates GEPsToProcess with the GEP
+  /// instructions to process as IV users.
```
martien-de-jong: I think it's just collecting interesting GEPs. This can be reflected in the name and its return value can be replaced by !GEPsToProcess.empty();

llvm/lib/Analysis/IVUsers.cpp
```diff
@@ -153,9 +157,16 @@ bool IVUsers::AddUsersIfInteresting(Instruction *I) {
...
-  uint64_t Width = SE->getTypeSizeInBits(I->getType());
-  if (Width > 64 || !DL.isLegalInteger(Width))
-    return false;
+  // Use TTI hook if available to allow targets where pointer and integer bit
+  // sizes differ (e.g., 20-bit pointers with 32-bit integers) to enable IV
+  // user collection for index-sized types.
+  if (!BypassWidthCheck) {
+    const uint64_t Width = SE->getTypeSizeInBits(I->getType());
+    const bool IsValidType = TTI ? TTI->isValidIVUserType(I->getType())
+                                 : (Width <= 64 && DL.isLegalInteger(Width));
```
martien-de-jong: Do we need the BypassWidthCheck parameter? We override it with a TTI callback.

llvm/lib/Analysis/IVUsers.cpp
```diff
@@ -170,6 +181,18 @@ bool IVUsers::AddUsersIfInteresting(Instruction *I) {
...
 
+  // Allow targets to look through certain instructions (e.g., truncs to index
+  // size on targets where pointer and integer bit sizes differ) to collect
+  // their users instead. This enables LSR to create pointer PHIs.
+  SmallVector<GetElementPtrInst *, 4> GEPsToProcess;
+  if (TTI && TTI->shouldIVUsersLookThroughInst(I, GEPsToProcess)) {
+    LLVM_DEBUG(dbgs() << "Looking through instruction: " << *I << '\n');
+    bool AnyInteresting = false;
+    for (GetElementPtrInst *GEP : GEPsToProcess)
+      AnyInteresting |= AddUsersIfInteresting(GEP, /*BypassWidthCheck=*/true);
```
martien-de-jong: If we don't skip it, it will be our TTI call back doing the check. Doesn't that give sufficient tuning freedom?

llvm/lib/Target/AIE/AIEBaseTargetTransformInfo.h
```diff
@@ -114,6 +127,93 @@ template <typename T> class AIEBaseTTIImpl : public BasicTTIImplBase<T> {
...
+  }
+
+  /// Collect all GEP users of \p Trunc that use it as an index operand (not
+  /// the pointer operand). Returns false if any use is not a GEP index.
+  static bool collectGEPIndices(const TruncInst *Trunc,
+                                SmallVectorImpl<GetElementPtrInst *> &GEPs) {
+    for (const Use &U : Trunc->uses()) {
+      auto *GEP = dyn_cast<GetElementPtrInst>(U.getUser());
+      if (!GEP || U.getOperandNo() == 0)
+        return false;
```
martien-de-jong: Why do all users need to be GEPs for any to qualify?

llvm/lib/Target/AIE/AIEBaseTargetTransformInfo.h
```diff
@@ -114,6 +127,93 @@ template <typename T> class AIEBaseTTIImpl : public BasicTTIImplBase<T> {
...
+      GEPs.push_back(GEP);
```
martien-de-jong: Could we be pushing a GEP more than once, e.g. in case of A[i][i]

llvm/include/llvm/Analysis/IVUsers.h
```diff
@@ -124,7 +130,7 @@ class IVUsers {
   /// AddUsersIfInteresting - Inspect the specified Instruction.  If it is a
   /// reducible SCEV, recursively add its users to the IVUsesByStride set and
   /// return true.  Otherwise, return false.
```
martien-de-jong, line 132: Dcoument the new parameter?

llvm/include/llvm/Analysis/IVUsers.h
```diff
@@ -156,6 +162,11 @@ class IVUsers {
 
   /// dump - This method is used for debugging.
   void dump() const;
+
+private:
+  /// Look through \p I (e.g. a trunc to index size), adding its GEP users
+  /// instead. Returns the combined interesting result, or std::nullopt if not.
```
martien-de-jong: The comment is scanty. I think it says it's a wrapper around AddUsersIfInteresting. I'm not sure what and how it's combining the return value.

llvm/include/llvm/Analysis/TargetTransformInfo.h
```diff
@@ -808,6 +812,11 @@ class TargetTransformInfo {
   LLVM_ABI AddressingModeKind
   getPreferredAddressingMode(const Loop *L, ScalarEvolution *SE) const;
 
+  SmallVector<GetElementPtrInst *>
```
martien-de-jong, line 818: Documentation please

llvm/lib/Target/AIE/AIEBaseTargetTransformInfo.h
```diff
@@ -199,6 +204,11 @@ template <typename T> class AIEBaseTTIImpl : public BasicTTIImplBase<T> {
                                const PHINode *IV2) const override {
     return shouldMergeCongruentIVsImpl(IV1, IV2);
   }
+
+  bool isValidIVUserType(Type *Ty) const override {
+    return BaseT::isValidIVUserType(Ty) ||
+           isIndexSizedInteger(Ty, BaseT::getDataLayout());
+  }
```
martien-de-jong, line 255: CHECK: the name doesn't suggest a connection to pointer arithmetic. The implementation using getIndexSizeInBits() does. I fear this may lead to misuse in future, in the sense that we are promising that we can deal with any induction value of this type.

llvm/lib/Target/AIE/AIEBaseTargetTransformInfo.h
```diff
@@ -209,6 +226,20 @@ template <typename T> class AIEBaseTTIImpl : public BasicTTIImplBase<T> {
...
+  SmallVector<GetElementPtrInst *>
+  getIVUsersLookThroughCandidates(Instruction *I,
+                                  const Loop *L) const override {
+    auto *Trunc = dyn_cast<TruncInst>(I);
+    if (!Trunc)
+      return {};
+
+    const DataLayout &DL = BaseT::getDataLayout();
+    if (!isIndexSizedInteger(Trunc->getType(), DL) ||
+        DL.isLegalInteger(Trunc->getType()->getIntegerBitWidth()))
```
martien-de-jong: This conjunction looks strange. My expectation would be `if (!A && !B) reject() `here. If this expectation is wrong, that should be explained in a comment.

llvm/lib/Target/AIE/AIEBaseTargetTransformInfo.h
```diff
@@ -88,14 +88,36 @@ template <typename T> class AIEBaseTTIImpl : public BasicTTIImplBase<T> {
...
+           ShiftAmount->getValue().ult(IndexTy->getIntegerBitWidth());
+  }
+
   static SmallVector<GetElementPtrInst *>
   collectGEPIndicesThroughTrunc(const TruncInst *Trunc) {
     SmallVector<GetElementPtrInst *> GEPs;
     for (const Use &U : Trunc->uses()) {
-      GetElementPtrInst *GEP = getGEPIndexUser(U);
-      if (!GEP)
+      if (GetElementPtrInst *GEP = getGEPIndexUser(U)) {
```
martien-de-jong, line 109: I would group the selection in a lambda returning GEP or nullptr and have a simple loop doing the push or the return.

PR #782, opened 2026-02-06
[AIE2P] ups combiner

F-Stuckmann:
We regress because we replace two Instructions with small latencies with a single instruction with the combined latency and thus we remove scheduling freedom for the post pipeliner.
Overall performance +0.04%


andcarminati:
I can see this as a very positive change for GEEM.

llvm/test/CodeGen/AIE/aie2p/GlobalIsel/inst-select-vlda_ups-advance.mir
martien-de-jong, line 1: Do we want to add the test in this commit?

llvm/lib/Target/AIE/AIECombinerHelper.h
```diff
@@ -148,9 +148,20 @@ bool matchShuffleToExtractBroadcast(MachineInstr &MI, MachineRegisterInfo &MRI,
...
+///
+/// WARNING: Setting \p AllowSideEffects to true bypasses critical safety
+/// checks. Only use this in specialized combining flows (e.g., VLDA_UPS
+/// combining in instruction selection) where you have verified that moving
+/// \a Dest is safe despite its side effects. In almost all cases, this
+/// parameter should remain false.
 bool canAdvanceOp(MachineInstr &MemI, MachineInstr &Dest,
-                  const MachineRegisterInfo &MRI);
+                  const MachineRegisterInfo &MRI,
+                  bool AllowSideEffects = false);
```
martien-de-jong: Perhaps `SideEffectsAreChecked` describes the warning above better?

llvm/lib/Target/AIE/aie2p/AIE2PInstructionSelector.cpp
```diff
@@ -587,14 +587,22 @@ bool AIE2PInstructionSelector::selectG_AIE_LOAD_UPS(MachineInstr &UPSI,
   // First use is the G_INTRINSIC_W_SIDE_EFFECTS ID
   Register LoadResult = (std::next(UPSI.uses().begin()))->getReg();
```
martien-de-jong: Down below, we use `Register ShftReg = UPSI.getOperand(3).getReg();` etc.  Why not here?

llvm/lib/Target/AIE/aie2p/AIE2PInstructionSelector.cpp
```diff
@@ -587,14 +587,22 @@ bool AIE2PInstructionSelector::selectG_AIE_LOAD_UPS(MachineInstr &UPSI,
...
   MachineInstr *LoadOp = getDefIgnoringCopiesAndBitcasts(LoadResult, MRI);
+  // Should we build the instruction at load's position?
+  bool ShouldAdvanceOp = false;
```
mludevid: I'd prefer doing something similar as we do in the combiners (e.g. for the addressing modes): having an insertion point variable that stores a pointer to the machine instr where we want to insert the combined instruction. I find that clearer and it's also more flexible in case we improve the algorithm even further in the future by finding an insertion point between the two instructions.

llvm/lib/Target/AIE/AIEBaseInstructionSelector.cpp
```diff
@@ -786,15 +786,14 @@ bool AIEBaseInstructionSelector::selectG_AIE_LOAD_CONV(
...
+  MIB.setInstr(*InsertionPoint);
 
   std::optional<AddressingModeInfo> AMI =
       getOrDefineAddressingRegister(*LoadOp, MRI);
 
   if (!AMI) {
-    // Restore to the correct position.
-    MIB.setInstr(CONVI);
+    if (InsertionPoint != &CONVI)
+      MIB.setInstr(CONVI);
```
martien-de-jong, line 796: I'm wondering why we don't delay MIB.setInstr until we are fully committed for takeoff.

llvm/lib/Target/AIE/AIEBaseInstructionSelector.cpp
```diff
@@ -842,6 +842,95 @@ AIEBaseInstructionSelector::getCombinedOpcodeCONVLoad(
...
+    NewInstr.addDef(Def->getReg());
+  }
+
+  addAddressingMode(NewInstr, *AMI, LSO->FitsImmediateRange, false, MRI);
+
+  NewInstr.cloneMemRefs(*LoadOp);
+
+  if (!ConstantSign)
+    setUnsetCtrlRegister(MIB, *NewInstr, MRI, *getUnpackSignRegister(),
+                         SignReg);
```
martien-de-jong, line 926: Nit: It seems we don't need the UnpackSignRegister for ConstantSign However, we bailed out early if we couldn't find it.

llvm/lib/Target/AIE/AIEBaseInstructionSelector.cpp
```diff
@@ -842,6 +842,95 @@ AIEBaseInstructionSelector::getCombinedOpcodeCONVLoad(
...
+    return false;
+
+  const Register DstReg = UNPACKI.getOperand(0).getReg();
+  const Register SignReg = UNPACKI.getOperand(3).getReg();
+
+  const auto SignVal = getIConstantVRegValWithLookThrough(SignReg, MRI);
+  const bool ConstantSign = SignVal.has_value();
+  std::optional<LoadStoreOpcodes> LSO =
+      getCombinedOpcodeUNPACKLoad(*LoadOp, UNPACKI, AMI->ImmediateOffset,
+                                  ConstantSign && SignVal->Value == 0x1);
```
martien-de-jong, line 904: I tend to treat std::optional as a pointer. Hence I would write if SignVal && *SignVal == 0x1 and get rid of the ConstantSign boolean.

martien-de-jong (approved):
Nice refactoring! Just a few nits. I guess the title could tags AIE2PS as well.

PR #786, opened 2026-02-15
Preserve PDL symbol names 

SamNour:
This change fixes the absence of PDL symbol names (PDLL pattern names) on using `--debug`.

this PR include the testing for the pdl_interp symbol name propagation: taken from https://github.com/llvm/llvm-project/pull/149481/changes/bff70da233c5d2d2a597d7bc06710b5cae7416cf by https://github.com/jumerckx

Testing for the GreedyPatternRewriter logger (propagating Debugging names) is in _PR#32098_ in our project.

**For reviewers:**
Root cause and what this PR is fixing:
Not preserving symbold names when loweing: PDLL -> PDL dialect -> PDLInterp dialect -> ByteCode : utilised by Greedy Logger

At the moment the [Greedy Logger logs pattern meta data including the debugName](https://github.com/Xilinx/llvm-aie/blob/aie-public/mlir/lib/Transforms/Utils/GreedyPatternRewriteDriver.cpp#L565-L576) (which is not propagated from PDL symbols). The greedy logger [invokes the PatternApplicator](https://github.com/Xilinx/llvm-aie/blob/aie-public/mlir/lib/Transforms/Utils/GreedyPatternRewriteDriver.cpp#L614-L615) For native C++ rewrites, those are native RewritePattern instances, where debugName(s) are set, so the Logger works as expected.

For PDLL rewrites, those are PDLByteCodePattern instances (which inherit Pattern), [created in ByteCode.cpp](https://github.com/Xilinx/llvm-aie/blob/aie-public/mlir/lib/Rewrite/ByteCode.cpp#L39-L57), here we were not attaching the symbol names from PDLinterpret (fixed by this PR). To resolve the actual symbol name, the name should be initally preserved in PDLToPDLInterp.cpp, at the moment the leaf is just a generic [pdl_generated_rewriter_#\<num>](https://github.com/Xilinx/llvm-aie/blob/aie-public/mlir/lib/Conversion/PDLToPDLInterp/PDLToPDLInterp.cpp#L634-L640).
The PR resolves the correct symbol name to be propagated to the rest of the flow. 

Readings: [1](https://reviews.llvm.org/D89107), [2](https://discourse.llvm.org/t/whats-the-purpose-of-pdl-pattern/73369), [3](https://mlir.llvm.org/docs/Dialects/PDLInterpOps/)

### DIFF
--- a/mlir/lib/Conversion/PDLToPDLInterp/PDLToPDLInterp.cpp
+++ b/mlir/lib/Conversion/PDLToPDLInterp/PDLToPDLInterp.cpp
@@ -634,8 +634,11 @@ void PatternLowering::generate(SuccessNode *successNode, Block *&currentBlock) {
 SymbolRefAttr PatternLowering::generateRewriter(
     pdl::PatternOp pattern, SmallVectorImpl<Position *> &usedMatchValues) {
   builder.setInsertionPointToEnd(rewriterModule.getBody());
+  StringRef rewriterName = "pdl_generated_rewriter";
+  if (auto symName = pattern.getSymName())
+    rewriterName = symName.value();
   auto rewriterFunc = builder.create<pdl_interp::FuncOp>(
-      pattern.getLoc(), "pdl_generated_rewriter",
+      pattern.getLoc(), rewriterName,
       builder.getFunctionType(std::nullopt, std::nullopt));
   rewriterSymbolTable.insert(rewriterFunc);
 
--- a/mlir/lib/Rewrite/ByteCode.cpp
+++ b/mlir/lib/Rewrite/ByteCode.cpp
@@ -3,7 +3,7 @@
 // Part of the LLVM Project, under the Apache License v2.0 with LLVM Exceptions.
 // See https://llvm.org/LICENSE.txt for license information.
 // SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
-// Modifications (c) Copyright 2025 Advanced Micro Devices, Inc. or its
+// Modifications (c) Copyright 2025-2026 Advanced Micro Devices, Inc. or its
 // affiliates
 //
 //===----------------------------------------------------------------------===//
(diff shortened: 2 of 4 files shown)


mlir/lib/Rewrite/ByteCode.cpp
```diff
@@ -42,18 +42,24 @@ PDLByteCodePattern PDLByteCodePattern::create(pdl_interp::RecordMatchOp matchOp,
...
     generatedOps =
         llvm::to_vector<8>(generatedOpsAttr.getAsValueRange<StringAttr>());
 
   // Check to see if this is pattern matches a specific operation type.
-  if (std::optional<StringRef> rootKind = matchOp.getRootKind())
-    return PDLByteCodePattern(rewriterAddr, configSet, *rootKind, benefit, ctx,
-                              generatedOps);
-  return PDLByteCodePattern(rewriterAddr, configSet, MatchAnyOpTypeTag(),
-                            benefit, ctx, generatedOps);
+  PDLByteCodePattern pattern = [&]() -> PDLByteCodePattern {
```
gerion-amd: Calling a temporary function here works but looks quite complicated. If I see it right, the only difference is the `rootKind`. Is is possible to let the condition just set `rootKind`?
```cpp
StringRef rootKind = MatchAnyOpTypeTag();
if (std::optional<StringRef> maybeRootKind = matchOp.getRootKind())
    rootKind = *maybeRootKind;
pattern = PDLByteCodePattern(..., rootKind, ...);
```
Or with the one line if `<condition> ? <foo> : <bar>`.

mlir/test/Conversion/PDLToPDLInterp/pdl-to-pdl-interp-rewriter.mlir
```diff
@@ -276,3 +276,30 @@ module @create_empty_region {
     }
   }
 }
+
+// -----
+
+// CHECK-LABEL: module @named_pattern
+module @named_pattern {
```
gerion-amd, line 283: Does this test checks for a successfully set debug name? I don't get the connection currently.

gerion-amd (dismissed):
Thanks. This will be really helpful!

mlir/test/Conversion/PDLToPDLInterp/pdl-to-pdl-interp-rewriter.mlir
SamNour: This test check the pdl_interpt symbol name propagation, checking the debugging names from the GreedyPatternRewriter Logger are in PR#32098 in our project. Further details are in the description : )

mathmer-amd (dismissed):
Thank you!

mathmer-amd (dismissed):

gerion-amd (dismissed):

ehsan-toosi (dismissed):
LGTM. One question. Can't you add a test dummy lit test for PDLL with debug-only like the test you created in your other PR?

ehsan-toosi (approved):

mlir/lib/Conversion/PDLToPDLInterp/PDLToPDLInterp.cpp
```diff
@@ -634,8 +634,11 @@ void PatternLowering::generate(SuccessNode *successNode, Block *&currentBlock) {
 SymbolRefAttr PatternLowering::generateRewriter(
     pdl::PatternOp pattern, SmallVectorImpl<Position *> &usedMatchValues) {
   builder.setInsertionPointToEnd(rewriterModule.getBody());
+  StringRef rewriterName = "pdl_generated_rewriter";
+  if (auto symName = pattern.getSymName())
+    rewriterName = symName.value();
```
roberteg16, line 639: Is `symName.value()` guaranteed to live when its value is used later?

mlir/lib/Rewrite/ByteCode.cpp
```diff
@@ -42,18 +42,24 @@ PDLByteCodePattern PDLByteCodePattern::create(pdl_interp::RecordMatchOp matchOp,
   PatternBenefit benefit = matchOp.getBenefit();
   MLIRContext *ctx = matchOp.getContext();
 
+  StringRef debugName = matchOp.getRewriter().getLeafReference().getValue();
```
roberteg16, line 45: Same here

roberteg16 (approved):

mlir/lib/Conversion/PDLToPDLInterp/PDLToPDLInterp.cpp
SamNour: `symName.value()` returns a `StringRef` backed by `StringAttr` which is owned and uniquified by the `MLIRContext` through `StringAttrStorage`[[1]](https://github.com/Xilinx/llvm-aie/blob/aie-public/mlir/lib/IR/AttributeDetail.h#L324-L343). So the referenced bytes outlive the local variable

mlir/lib/Rewrite/ByteCode.cpp
SamNour: Same logic applies here with `getLeafReference()` as well.

roberteg16:
🎉

PR #787, opened 2026-02-17
Fix: avoid passing local variable to lambda by reference

gvartany-amd:
The variable is local, so it is freed after the end of the function. This tripped ASAN's use-after-free check when the lambda was called.


F-Stuckmann (approved):

mludevid (approved):
LGTM

PR #788, opened 2026-02-17
[AIEX] Fix SEF peeling scheduling bug

mludevid:
There was a problem in the software pipeliner that led to incorrect schedules when creating an SEF stage. Rotating the schedule to create that SEF stage was causing an inconsistent state, where the scheduled cycles were updated, but the earliest and latest cycles for unscheduled SUs were not updated accordingly. This inconsistent state led to incorrect schedules being accepted.

Commit 1 adds an assert to verify the state is consistent when scheduling the other iterations and commit 2 fixes the SEF creation to avoid this inconsistent state.

### DIFF
--- a/llvm/lib/Target/AIE/AIEPostPipeliner.cpp
+++ b/llvm/lib/Target/AIE/AIEPostPipeliner.cpp
@@ -820,6 +820,20 @@ void dumpEarliestChain(const ScheduleInfo &Info, int N) {
   dbgs() << "  --> SU" << N << " @" << Info[N].Cycle << "\n";
 }
 
+#ifndef NDEBUG
+/// Recompute Earliest from direct predecessors only.
+int computeEarliestFromPreds(const SUnit &SU, const ScheduleInfo &Info) {
+  int Earliest = 0;
+  for (const SDep &Dep : SU.Preds) {
+    const SUnit *Pred = Dep.getSUnit();
+    if (Pred->isBoundaryNode())
+      continue;
+    const NodeInfo &PredNode = Info[Pred->NodeNum];
+    Earliest = std::max(Earliest, PredNode.Cycle + Dep.getSignedLatency());
+  }
+  return Earliest;
+}
+#endif
 } // namespace
 
 bool PostPipeliner::scheduleOtherIterations(PostPipelinerStrategy &Strategy) {
@@ -833,6 +847,13 @@ bool PostPipeliner::scheduleOtherIterations(PostPipelinerStrategy &Strategy) {
     NodeInfo &Node = Info[N];
     const SUnit &ModuloSU = DAG->SUnits[N - NInstr];
     NodeInfo &ModuloNode = Info[N - NInstr];
+
+#ifndef NDEBUG
+    // Assert that Earliest is still consistent with scheduled predecessors.
+    const int RecomputedEarliest = computeEarliestFromPreds(SU, Info);
+    assert(Node.Earliest >= RecomputedEarliest &&
+           "Earliest is stale: predecessor pushes a later cycle");
+#endif
 
     // Earliest tracks the latencies of the loop carried deps
     const int Earliest = Node.Earliest;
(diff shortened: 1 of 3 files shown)


llvm/lib/Target/AIE/AIEPostPipeliner.cpp
```diff
@@ -813,6 +828,13 @@ bool PostPipeliner::scheduleOtherIterations(PostPipelinerStrategy &Strategy) {
     const SUnit &ModuloSU = DAG->SUnits[N - NInstr];
     NodeInfo &ModuloNode = Info[N - NInstr];
 
+#ifndef NDEBUG
+    // Assert that Earliest is still consistent with scheduled predecessors.
+    const int RecomputedEarliest = computeEarliestFromPreds(SU, Info);
+    assert(Node.Earliest >= RecomputedEarliest &&
+           "Earliest is stale: predecessor pushes a later cycle");
+#endif
```
martien-de-jong, line 856: setting NDEBUG will give you an unused function warning

llvm/lib/Target/AIE/AIEPostPipeliner.cpp
```diff
@@ -1549,13 +1555,18 @@ void NodeInfo::update(int II) {
...
+
   for (int N = 0; N < NInstr; N++) {
-    Nodes[N].Cycle = Nodes[N].Cycle + Rotation;
+    Nodes[N].Cycle = Nodes[N].Cycle + PendingRotation;
     Nodes[N].update(II);
 
     commitCycle(N);
   }
+
+  PendingRotation = 0;
```
martien-de-jong: I think we should also reset PendingRotation for new pipelining attempts. I think there's a function to reset the schedule somewhere.

llvm/lib/Target/AIE/AIEPostPipeliner.cpp
```diff
@@ -813,6 +827,13 @@ bool PostPipeliner::scheduleOtherIterations(PostPipelinerStrategy &Strategy) {
     const SUnit &ModuloSU = DAG->SUnits[N - NInstr];
     NodeInfo &ModuloNode = Info[N - NInstr];
 
+#ifndef NDEBUG
+    // Assert that Earliest is still consistent with scheduled predecessors.
+    const int RecomputedEarliest = computeEarliestFromPreds(SU, Info);
+    assert(Node.Earliest >= RecomputedEarliest &&
```
mludevid, line 854: Actually, could this check equality? I'm not sure, because we are not doing the transitive computation of earliest.

martien-de-jong: I think it could. All transitivity is reflected in Cycle of the predecessors. We push Earliest whenever we commit a cycle of a predecessor.

llvm/lib/Target/AIE/AIEPostPipeliner.cpp
mludevid: True! But that reset schedule function is on the PostPipeliner. Do we want to reset it there? Another option would be to reset it at the beginning of `scheduleWithStrategy`. I think that's the safest and clearest option. We don't care about the rotation anyways outside of this function. We can also reset it if schedule other iterations fails. That would ensure it's always 0 outside of `scheduleWithStrategy`.

llvm/lib/Target/AIE/AIEPostPipeliner.cpp
mludevid: I missed an important detail: the SUs of the second copy are also actually scheduled (by calling `scheduleNode` at the end). That in turn updates the earliest of the successor SUs. And since we are going through the SUs in order we are guaranteed that all predecessors of the current SU are already scheduled. For some reason I was under the impression that we only looped through all SUs in the second copy and verified that the insert cycle was after earliest. I think enforcing equality should work.

llvm/lib/Target/AIE/AIEPostPipeliner.cpp
mludevid: I've updated the code to reset the `PendingRotation` when exiting `scheduleWithStrategy`. That way `PendingRotation` is always `0` outside of `scheduleWithStrategy`.

llvm/lib/Target/AIE/AIEPostPipeliner.cpp
mludevid: I tried this and equality doesn't always hold. `Earliest` can be larger than the latency from the predecessors, e.g. because of the loop-coupling: `Info[K].Earliest = max(Info[K].Earliest, Info[K+NInstr].Earliest - II);`.

martien-de-jong (approved):

mludevid:
I rebased and squashed the fixup commits.

PR #789, opened 2026-02-17
[AIEX] fixed ordering for RegClasses in WAwRegRewriter

F-Stuckmann:
Using Pointers as keys in a map leads to inconsistent behaviour. 
Fix: Use RegClassIDs instead of pointers for Keys.


F-Stuckmann:
The problematic datastructures were already deleted in https://github.com/Xilinx/llvm-aie/pull/769

PR #790, opened 2026-02-18
[AIEX] Combine vector broadcast and extract.

niwinanto:


llvm/test/CodeGen/AIE/GlobalISel/combine-broadcast-extract.mir
```diff
@@ -338,9 +336,8 @@ body:             |
     ; CHECK-NEXT: [[COPY:%[0-9]+]]:_(<16 x s32>) = COPY $x0
     ; CHECK-NEXT: [[C:%[0-9]+]]:_(s32) = G_CONSTANT i32 0
     ; CHECK-NEXT: [[AIE_SEXT_EXTRACT_VECTOR_ELT:%[0-9]+]]:_(s32) = G_AIE_SEXT_EXTRACT_VECTOR_ELT [[COPY]](<16 x s32>), [[C]](s32)
-    ; CHECK-NEXT: [[AIE_BROADCAST_VECTOR:%[0-9]+]]:_(<32 x s16>) = G_AIE_BROADCAST_VECTOR [[AIE_SEXT_EXTRACT_VECTOR_ELT]](s32)
-    ; CHECK-NEXT: [[AIE_SEXT_EXTRACT_VECTOR_ELT1:%[0-9]+]]:_(s32) = G_AIE_SEXT_EXTRACT_VECTOR_ELT [[AIE_BROADCAST_VECTOR]](<32 x s16>), [[C]](s32)
-    ; CHECK-NEXT: $r0 = COPY [[AIE_SEXT_EXTRACT_VECTOR_ELT1]](s32)
+    ; CHECK-NEXT: [[ASSERT_SEXT:%[0-9]+]]:_(s32) = G_ASSERT_SEXT [[AIE_SEXT_EXTRACT_VECTOR_ELT]], 16
```
mludevid: I don't think this is correct. AFAIK `G_ASSERT_SEXT` is just a hint and not enforced. And in this example the input to it was not sign extended from 16 bits. This also leads to the combiner potentially producing incorrect results, since before the combine the 32 bit value in `%2` is truncated to 16 bits during the broadcasting and later sign-extended to 32 bits again. But after the combine we just get a copy.

niwinanto: Good catch, I was in an impression `G_ASSERT_SEXT` is lowered to extent.s8/s16. But you are right, it is just a hint. Also, I could see no instruction selection for SEXT and ZEXT just a copy. 

Here I can use extent.s8/s16 to fix the problem. We might have one more opportunity to remove extent.s8/s16 if they are properly extended, but that can be a different combine.

llvm/lib/Target/AIE/AIECombinerHelper.cpp
```diff
@@ -4652,6 +4652,64 @@ bool llvm::matchExtractVecEltAssertBcst(MachineInstr &MI,
   return true;
 }
 
+/// G_AIE_BROADCAST_VECTOR %vec, %scalar
+/// G_AIE_[S/Z]EXT_EXTRACT_VECTOR_ELT %dst, %vec, %idx
+/// -> COPY %dst, %scalar (with G_ASSERT_[S/Z]EXT if element size < 32)
```
mludevid: The comment is not up to date anymore.

llvm/test/CodeGen/AIE/GlobalISel/combine-broadcast-extract.mir
```diff
@@ -269,10 +269,8 @@ body:             |
     ; CHECK-NEXT: [[COPY:%[0-9]+]]:_(<32 x s16>) = COPY $x0
     ; CHECK-NEXT: [[C:%[0-9]+]]:_(s32) = G_CONSTANT i32 1
     ; CHECK-NEXT: [[AIE_SEXT_EXTRACT_VECTOR_ELT:%[0-9]+]]:_(s32) = G_AIE_SEXT_EXTRACT_VECTOR_ELT [[COPY]](<32 x s16>), [[C]](s32)
-    ; CHECK-NEXT: [[AIE_BROADCAST_VECTOR:%[0-9]+]]:_(<32 x s16>) = G_AIE_BROADCAST_VECTOR [[AIE_SEXT_EXTRACT_VECTOR_ELT]](s32)
-    ; CHECK-NEXT: [[C1:%[0-9]+]]:_(s32) = G_CONSTANT i32 0
-    ; CHECK-NEXT: [[AIE_SEXT_EXTRACT_VECTOR_ELT1:%[0-9]+]]:_(s32) = G_AIE_SEXT_EXTRACT_VECTOR_ELT [[AIE_BROADCAST_VECTOR]](<32 x s16>), [[C1]](s32)
-    ; CHECK-NEXT: $r0 = COPY [[AIE_SEXT_EXTRACT_VECTOR_ELT1]](s32)
+    ; CHECK-NEXT: [[SEXT_INREG:%[0-9]+]]:_(s32) = G_SEXT_INREG [[AIE_SEXT_EXTRACT_VECTOR_ELT]], 16
```
mludevid, line 272: Q: Just to be sure, this is lowered to an `EXTEND` which takes the 16 least significant bits and sign extends them to 32 bits? If that's the case I think this looks good now.

llvm/test/CodeGen/AIE/GlobalISel/combine-broadcast-extract.mir
mludevid, line 355: This comment needs to be updated as well.

mludevid, line 288: Same here.

llvm/lib/Target/AIE/AIECombinerHelper.cpp
```diff
@@ -4652,6 +4652,64 @@ bool llvm::matchExtractVecEltAssertBcst(MachineInstr &MI,
...
+
+  // Get the scalar that was broadcast
+  const Register BroadcastSrcReg = SrcMI->getOperand(1).getReg();
+  const Register DstReg = MI.getOperand(0).getReg();
+
+  // Types must match for a simple copy
+  const LLT SrcTy = MRI.getType(BroadcastSrcReg);
+  const LLT DstTy = MRI.getType(DstReg);
+  if (SrcTy != DstTy)
+    return false;
```
mludevid: I don't get this. 1. This combine does not only emit simple copies, so the comment is misleading. 2. Shouldn't we be making sure the source and the destination are `s32`? I'm not sure this combiner should also work on `s64`. If it does, we are missing tests for them.

llvm/lib/Target/AIE/AIECombinerHelper.cpp
```diff
@@ -4652,6 +4652,64 @@ bool llvm::matchExtractVecEltAssertBcst(MachineInstr &MI,
...
+               DstTy](MachineIRBuilder &B) {
+    if (ElemSize < 32) {
+      if (IsSext) {
+        // Use G_SEXT_INREG for sign extension
+        B.buildSExtInReg(DstReg, BroadcastSrcReg, ElemSize);
+      } else {
+        // Use G_AND with mask for zero extension
+        const uint64_t Mask = (1ULL << ElemSize) - 1;
+        auto MaskReg = B.buildConstant(DstTy, Mask);
+        B.buildAnd(DstReg, BroadcastSrcReg, MaskReg);
```
mludevid, line 4703: We do have an unsigned extend instruction on AIEs. Why not use it to save an additional instruction?

mludevid: Discussed offline. We actually combine the and to an unsigned extend during instruction selection.

llvm/test/CodeGen/AIE/GlobalISel/combine-broadcast-extract.mir
niwinanto: Yes, we select to `EXTEND`

llvm/test/CodeGen/AIE/GlobalISel/combine-broadcast-extract.mir
niwinanto: Actually, the combiner under test is not matched here. Yes it is confusing as the latest combine is just the instructions in reverse order.

llvm/lib/Target/AIE/AIECombinerHelper.cpp
niwinanto: Done.

llvm/lib/Target/AIE/AIECombinerHelper.cpp
niwinanto: Broadcast support `s64` , but extract support `s64` though `2xs32` so should be fine. But your comment make sense as we might deviate in future and would be nice if we make this explicit.

mludevid (approved):
Thanks for addressing my comments. LGTM

PR #791, opened 2026-02-18
Preliminiaries for postregalloc in postpipeliner

martien-de-jong:
A bunch of commits with small generalizations.
In particular, the relax assumptions about absence of physical registers in post scheduler logic, which is essential for postregalloc.

### DIFF
--- a/llvm/lib/Target/AIE/AIEInterBlockScheduling.cpp
+++ b/llvm/lib/Target/AIE/AIEInterBlockScheduling.cpp
@@ -375,7 +375,9 @@ bool InterBlockScheduling::leaveBlock() {
   // After scheduling a basic block, check convergence to determine which block
   // to schedule next and with what parameters
   auto &BS = *CurrentBlockState;
-  switch (updateFixPoint(BS)) {
+  const auto Stage = updateFixPoint(BS);
+  BS.FixPoint.Stage = Stage;
+  switch (Stage) {
   case SchedulingStage::GatheringRegions:
   case SchedulingStage::SchedulingNotConverged:
   case SchedulingStage::Scheduling:
@@ -516,7 +518,7 @@ MachineInstr *InterBlockScheduling::latencyConverged(BlockState &BS) const {
 
 SchedulingStage InterBlockScheduling::updateFixPoint(BlockState &BS) {
   if (BS.Kind != BlockType::Loop) {
-    return BS.FixPoint.Stage = SchedulingStage::SchedulingDone;
+    return SchedulingStage::SchedulingDone;
   }
 
   if (BS.FixPoint.Stage == SchedulingStage::GatheringRegions) {
@@ -525,7 +527,7 @@ SchedulingStage InterBlockScheduling::updateFixPoint(BlockState &BS) {
     // Now we can create the interblock edges between the top and the bottom
     // region
     BS.initInterBlock(*Context, *HR);
-    return BS.FixPoint.Stage = SchedulingStage::Scheduling;
+    return SchedulingStage::Scheduling;
   }
 
   BS.FixPoint.NumIters++;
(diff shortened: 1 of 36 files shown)


llvm/lib/Target/AIE/AIEPostPipeliner.cpp
```diff
@@ -48,6 +48,26 @@ static cl::opt<int> PresetII("aie-postpipeliner-target-ii",
 
 PipelineScheduleVisitor::~PipelineScheduleVisitor() {}
 
+std::optional<int> PostPipelinerStrategy::fitInInterval(
+    MachineInstr &MI, int First, int Last, int II,
+    const AIEHazardRecognizer &HR,
+    ResourceScoreboard<FuncUnitWrapper> &Scoreboard) {
+  const int Step = fromTop() ? 1 : -1;
+  if (Step < 0) {
+    std::swap(First, Last);
```
F-Stuckmann, line 58: nit: can we have an assert that First < Last?

F-Stuckmann:
Since these are all NFC, you could add [NFC] in each commit message

llvm/lib/Target/AIE/AIEInterBlockScheduling.cpp
```diff
@@ -67,6 +67,10 @@ static cl::opt<bool> EnableMultiSlotInstrMaterialization(
     cl::desc("Statically materialize Multi-Slot Pseudo Instructions in "
              "loops."));
 
+static cl::opt<bool>
+    MaterializeAll("aie-materialize-all", cl::Hidden, cl::init(false),
```
andcarminati, line 71: I guess we will reevaluate this option with all the upcoming features.

llvm/lib/Target/AIE/AIEMultiSlotInstrMaterializer.h
```diff
@@ -4,26 +4,27 @@
...
 // \file assigns an issue-slot to multi slot pseudo instructions within a single
 // block loop to help loop pipelining.
 //
 //===----------------------------------------------------------------------===//
-#include "AIEBaseInstrInfo.h"
 
 namespace llvm {
 class AIEHazardRecognizer;
-}
+class MachineBasicBlock;
```
andcarminati, line 18: Funny that `AIEHazardRecognizer` is in the llvm namespace, and not in the llvm::AIE below.

llvm/lib/Target/AIE/AIEPostPipeliner.cpp
```diff
@@ -48,6 +48,26 @@ static cl::opt<int> PresetII("aie-postpipeliner-target-ii",
 
 PipelineScheduleVisitor::~PipelineScheduleVisitor() {}
 
+std::optional<int> PostPipelinerStrategy::fitInInterval(
+    MachineInstr &MI, int First, int Last, int II,
+    const AIEHazardRecognizer &HR,
+    ResourceScoreboard<FuncUnitWrapper> &Scoreboard) {
+  const int Step = fromTop() ? 1 : -1;
+  if (Step < 0) {
```
andcarminati, line 57: nit: is more intuitive to use `!fromTop()` as condition.

andcarminati (dismissed):
LGTM, every cleanup is welcome! I left some some comments related to my own understanding.

llvm/lib/Target/AIE/AIEPostPipeliner.cpp
martien-de-jong: Before or after the swap? ;-)

martien-de-jong:
> Since these are all NFC, you could add [NFC] in each commit message
The one that is not labelled NFC may actually not be NFC, and it is in upstream code.

llvm/lib/Target/AIE/AIEPostPipeliner.cpp
martien-de-jong: That's a matter of opinion. Step represents the direction, and hence which bound initializes and which terminates the loop.

llvm/lib/Target/AIE/AIEMultiSlotInstrMaterializer.h
martien-de-jong: That's the way it is. At least it has an AIE prefix to avoid collisions.

llvm/lib/Target/AIE/AIEInterBlockScheduling.cpp
martien-de-jong: I guess it will remain a test/debug facility.

andcarminati (approved):

F-Stuckmann (approved):

PR #792, opened 2026-02-18
[AIEX] Add RemoveRedundantMemoryDeps mutator

andcarminati:
This adds a new scheduling mutator for software pipelining that removes redundant memory dependencies when data-dependency paths already enforce the required ordering.

Algorithm: Uses Floyd-Warshall to compute transitive closure of data (RAW) dependencies across all scheduling units. For each memory dependency edge, checks if a data/anti path exists between the same nodes. If found, the memory edge is redundant and removed.

Rationale: Memory dependencies are conservatively added by alias analysis and can over-constrain the scheduler. When a data or anti-dependency path already enforces the ordering between two memory operations, the explicit memory edge is unnecessary. Removing these redundant edges gives the software pipeliner more scheduling freedom, enabling better loop overlapping and improved instruction-level parallelism.


andcarminati:
More checks are needed.

PR #793, opened 2026-02-19
[AIEX] Fix register rewriting failed: cycle in copy bundle

andcarminati:
To fix the problem, we introduce infrastructure to distinguish between expandable and rewritable registers in the AIEUnallocatedSuperRegRewriter pass.
    
    Key changes:
    
    - Add `isExpandableRegister()` helper to check if a register has
    a subregister split, consolidating duplicate checks
    
    - Replace `RegRewriteInfo` typedef with a struct containing two lists:
    
      - `ExpandableRegs`: All registers with SubRegSplit (superset)
      - `RewritableRegs`: Subset that can be immediately rewritten
    
    - Rename `getRewriteCandidates()` to `getRewriteAndExpandCandidates()`
    to reflect the dual-list approach
    
    - Update `expandCopyBundles()` to process all expandable registers
    
    - Update `rewriteCandidates()` to process only rewritable registers


llvm/test/CodeGen/AIE/aie2p/ra/unallocated-supereg-all-bundles.mir
martien-de-jong, line 1: nit: superreg (double r)

llvm/lib/Target/AIE/AIEUnallocatedSuperRegRewriter.cpp
```diff
@@ -93,33 +100,44 @@ static RegRewriteInfo getRewriteCandidates(MachineRegisterInfo &MRI,
...
+      Info.ExpandableRegs.push_back({Reg, RewritableSubRegs});
+    } else if (AIESuperRegUtils::isExpandableRegister(Reg, MRI, TRI)) {
+      // Register is expandable but not immediately rewritable
+      // It needs bundle expansion first
+      auto &SubRegSplit = TRI.getSubRegSplit(MRI.getRegClass(Reg)->getID());
+      SmallSet<int, 8> AllSubRegs(SubRegSplit.begin(), SubRegSplit.end());
+      LLVM_DEBUG(dbgs() << "Expandable (not rewritable) "
+                        << printReg(Reg, &TRI, 0, &MRI) << ":"
+                        << printRegClassOrBank(Reg, MRI, &TRI) << '\n');
+      Info.ExpandableRegs.push_back({Reg, AllSubRegs});
```
martien-de-jong, line 117: Why don't we just store a pointer to SubRegSplit?  That's a const pointer.  Now we create three copies, one to populate it, one to create the pair, and one to push it into the vector.

llvm/lib/Target/AIE/AIESuperRegUtils.cpp
```diff
@@ -25,29 +25,30 @@
 
 namespace llvm::AIESuperRegUtils {
 
+bool isExpandableRegister(Register Reg, const MachineRegisterInfo &MRI,
+                          const AIEBaseRegisterInfo &TRI) {
+  if (Reg.isPhysical())
+    return false;
+
+  auto &SubRegSplit = TRI.getSubRegSplit(MRI.getRegClass(Reg)->getID());
```
F-Stuckmann, line 33: nit: we could return SubRegSplit as an optional

F-Stuckmann (approved):

PR #794, opened 2026-02-20
Fix QType with 'u' prefix to create unsigned storageType

ilango100:
- QuantTypes parser is parsing incorrectly that when the storage type has `u` prefix, it's storageType is created as signless integer. This PR fixes that to create storage type to be unsigned integer.
- Disable 2 tests in lower-quant-ops.mlir that work on unsigned QTypes
- Fix tests in quant canonicalize to use correct storage type for quant type

### DIFF
--- a/mlir/lib/Dialect/Quant/IR/TypeParser.cpp
+++ b/mlir/lib/Dialect/Quant/IR/TypeParser.cpp
@@ -47,7 +47,8 @@ static IntegerType parseStorageType(DialectAsmParser &parser, bool &isSigned) {
       return nullptr;
     }
     isSigned = false;
-    type = parser.getBuilder().getIntegerType(storageTypeWidth, false);
+    type = parser.getBuilder().getIntegerType(storageTypeWidth,
+                                              /*isSigned=*/isSigned);
   } else {
     return nullptr;
   }
--- a/mlir/lib/Dialect/Quant/IR/TypeParser.cpp
+++ b/mlir/lib/Dialect/Quant/IR/TypeParser.cpp
@@ -47,7 +47,7 @@ static IntegerType parseStorageType(DialectAsmParser &parser, bool &isSigned) {
       return nullptr;
     }
     isSigned = false;
-    type = parser.getBuilder().getIntegerType(storageTypeWidth);
+    type = parser.getBuilder().getIntegerType(storageTypeWidth, false);
   } else {
     return nullptr;
   }
--- a/mlir/test/Dialect/Quant/lower-quant-ops.mlir
+++ b/mlir/test/Dialect/Quant/lower-quant-ops.mlir
@@ -22,26 +22,26 @@ func.func @dcast_per_layer_scalar(%arg0: !qalias) -> f32 {
 
 // -----
 
-// CHECK-LABEL: @dcast_per_layer_scalar_unsigned
-// CHECK-SAME: %[[ARG_0:.*]]: !quant.uniform
+// COM: CHECK-LABEL: @dcast_per_layer_scalar_unsigned
+// COM: CHECK-SAME: %[[ARG_0:.*]]: !quant.uniform
 
-// CHECK: %[[STORED_INT:.*]] = quant.scast %[[ARG_0]] : !quant.uniform<u8:f32, 2.000000e+00:10> to i8
+// COM: CHECK: %[[STORED_INT:.*]] = quant.scast %[[ARG_0]] : !quant.uniform<u8:f32, 2.000000e+00:10> to i8
 
-// CHECK: %[[SCALE:.*]] = arith.constant 2.000000e+00 : f32
-// CHECK: %[[ZERO_POINT:.*]] = arith.constant 10 : i8
+// COM: CHECK: %[[SCALE:.*]] = arith.constant 2.000000e+00 : f32
+// COM: CHECK: %[[ZERO_POINT:.*]] = arith.constant 10 : i8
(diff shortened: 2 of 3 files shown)


mlir/lib/Dialect/Quant/IR/TypeParser.cpp
```diff
@@ -47,7 +47,7 @@ static IntegerType parseStorageType(DialectAsmParser &parser, bool &isSigned) {
       return nullptr;
     }
     isSigned = false;
-    type = parser.getBuilder().getIntegerType(storageTypeWidth);
+    type = parser.getBuilder().getIntegerType(storageTypeWidth, false);
```
roberteg16: Could you add a hit of what "false" means in this context?

``` C++
    type = parser.getBuilder().getIntegerType(storageTypeWidth, /*isSigned=*/ false);
```

I also wonder if the built-in parser cannot be used to parse the type instead of using this custom parsing.

ilango100: Updated to have the hint `/*isSigned=*/`.

I agree that we should be using the built-in type parser instead of custom parsing, which is what I started with, but reverted back as it needed much more changes in the test-cases and quant lowering pass as well. To keep changes minimal, I'm going with the existing approach.

roberteg16 (approved):

mlir/test/Dialect/Quant/lower-quant-ops.mlir
```diff
@@ -251,28 +251,28 @@ func.func @qcast_per_layer_scalar_bounds(%arg0: f32) -> !qalias {
...
-// CHECK: %[[STORED_CLAMPED:.*]] = arith.minui %[[STORED_CLAMPED_TEMP]], %[[C_10]] : i8
+// COM: CHECK-DAG: %[[C_2:.*]] = arith.constant 2 : i8
+// COM: CHECK-DAG: %[[C_10:.*]] = arith.constant 10 : i8
+// COM: CHECK: %[[STORED_CLAMPED_TEMP:.*]] = arith.maxui %[[STORED_INT]], %[[C_2]] : i8
+// COM: CHECK: %[[STORED_CLAMPED:.*]] = arith.minui %[[STORED_CLAMPED_TEMP]], %[[C_10]] : i8
 
-// CHECK: %[[STORED_QUANT:.*]] = quant.scast %[[STORED_CLAMPED]] : i8 to !quant.uniform<u8<2:10>:f32, 2.000000e+00>
-// CHECK: return %[[STORED_QUANT]] : !quant.uniform<u8<2:10>:f32, 2.000000e+00>
+// COM: CHECK: %[[STORED_QUANT:.*]] = quant.scast %[[STORED_CLAMPED]] : i8 to !quant.uniform<u8<2:10>:f32, 2.000000e+00>
+// COM: CHECK: return %[[STORED_QUANT]] : !quant.uniform<u8<2:10>:f32, 2.000000e+00>
```
roberteg16, line 269: Doesn't this disable the chekcs?

mlir/test/Dialect/Quant/lower-quant-ops.mlir
```diff
@@ -251,28 +251,28 @@ func.func @qcast_per_layer_scalar_bounds(%arg0: f32) -> !qalias {
...
-!qalias = !quant.uniform<u8<2:10>:f32, 2.0>
-func.func @qcast_per_layer_scalar_unsigned_bounds(%arg0: f32) -> !qalias {
-  %0 = quant.qcast %arg0 : f32 to !qalias
-  return %0 : !qalias
-}
+// !qalias = !quant.uniform<u8<2:10>:f32, 2.0>
+// func.func @qcast_per_layer_scalar_unsigned_bounds(%arg0: f32) -> !qalias {
+//   %0 = quant.qcast %arg0 : f32 to !qalias
+//   return %0 : !qalias
+// }
```
roberteg16, line 275: Why is the test commented out?

mlir/test/Dialect/Quant/lower-quant-ops.mlir
ilango100: Yes, this disables the checks. I've updated the main comment to reflect this. The reason is that the quant ops lowering code produces `arith.constant` op with the zero-point value (which can be signed, unsigned or signless). But `arith.constant` doesn't support signed or unsigned results, it only supports signless integer.

Because of that, the test is failing. I've disabled the tests only for unsigned quant types, existing tests are still present.

The solution to this is to handle the unsigned/signed types properly by inserting an `arith.bitcast` after the `arith.constant`. I was thinking we can take this up as a follow-up based on priority.

mlir/test/Dialect/Quant/lower-quant-ops.mlir
ilango100: Added the reason for commenting above.

roberteg16: I think it would be better if we change it for XFAIL

mlir/test/Dialect/Quant/lower-quant-ops.mlir
roberteg16: I think it would be better if we change it for XFAIL

ilango100: But the problem with XFAIL is that it will not check other tests in the same file. It will consider the whole file failed as soon as a single check fails.

PR #795, opened 2026-02-24
[AIEX] Shared post-legalizer custom combines list

mludevid:
Create a `aie_postlegalizer_custom_shared_combines` list with the combines used by all targets.

We already do this for the pre-legalization combines. I think this is not a completely NFC, because we are reordering some combines. We'll have to run some benchmarks to verify that QoR is not affected.


F-Stuckmann (approved):
No regressions whatsoever. Lets merge it

PR #802, opened 2026-02-26
[AIEX] enable VextrBcstShfl with copy-propagate G_IMPLICIT_DEF

F-Stuckmann:
Instruction matching cannot look through copies of Implicit def, if an implicit Def is needed. 
This can happen if an implicit Def is assigned to a Register Bank that is different to the one expected in the VextrBcstShfl instruction matcher(i.e. FIFOreg).

This PR replaces these copies.
Note: the big gains for this PR come from removing Copies when source and destination have the same regbank.


llvm/lib/Target/AIE/AIE2.h
```diff
@@ -4,7 +4,7 @@
 // See https://llvm.org/LICENSE.txt for license information.
 // SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
 //
-// (c) Copyright 2023-2024 Advanced Micro Devices, Inc. or its affiliates
+// (c) Copyright 2023-2026 Advanced Micro Devices, Inc. or its affiliates
```
martien-de-jong, line 7: nit: copyright-only change?

llvm/lib/Target/AIE/AIECombinerHelper.cpp
```diff
@@ -2111,6 +2111,40 @@ bool llvm::matchOffsetLoadStoreSharePtrAdd(MachineInstr &MI,
...
+}
+
+void llvm::applyCopyOfImplicitDef(MachineInstr &MI, MachineRegisterInfo &MRI,
+                                  MachineIRBuilder &B,
+                                  GISelChangeObserver &Observer) {
+  const Register DstReg = MI.getOperand(0).getReg();
+  const Register SrcReg = MI.getOperand(1).getReg();
+  const RegisterBank *DstRB = MRI.getRegBankOrNull(DstReg);
+  const RegisterBank *SrcRB = MRI.getRegBankOrNull(SrcReg);
+  if (DstRB == SrcRB) {
```
martien-de-jong, line 2147: CHECK: this includes nullptr == nullptr

llvm/lib/Target/AIE/AIEPreISelCombiner.cpp
```diff
@@ -0,0 +1,159 @@
...
+#include "llvm/CodeGen/GlobalISel/CSEInfo.h"
+#include "llvm/CodeGen/GlobalISel/Combiner.h"
+#include "llvm/CodeGen/GlobalISel/CombinerHelper.h"
+#include "llvm/CodeGen/GlobalISel/CombinerInfo.h"
+#include "llvm/CodeGen/GlobalISel/GIMatchTableExecutorImpl.h"
+#include "llvm/CodeGen/GlobalISel/GISelKnownBits.h"
+#include "llvm/CodeGen/MachineDominators.h"
+#include "llvm/InitializePasses.h"
+
+#define DEBUG_TYPE "aie-preisel-combiner"
```
martien-de-jong, line 30: nit: pre-isel for readability?

llvm/lib/Target/AIE/AIEPreISelCombiner.cpp
```diff
@@ -0,0 +1,159 @@
...
+
+public:
+  AIEPreISelCombinerImpl(MachineFunction &MF, CombinerInfo &CInfo,
+                         const TargetPassConfig *TPC, GISelKnownBits &KB,
+                         GISelCSEInfo *CSEInfo,
+                         const AIEPreISelCombinerImplRuleConfig &RuleConfig,
+                         const AIE2Subtarget &STI, MachineDominatorTree *MDT,
+                         const LegalizerInfo *LI);
+
+  static const char *getName() { return "AIEPreISelCombiner"; }
```
martien-de-jong, line 54: Any use for this function?

llvm/lib/Target/AIE/AIEPreISelCombiner.cpp
```diff
@@ -0,0 +1,159 @@
...
+namespace {
+
+#define GET_GICOMBINER_TYPES
+#include "AIEGenPreISelGICombiner.inc"
+#undef GET_GICOMBINER_TYPES
+
+class AIEPreISelCombinerImpl : public Combiner {
+  mutable CombinerHelper Helper;
+  const AIEPreISelCombinerImplRuleConfig &RuleConfig;
+  const AIE2Subtarget &STI;
```
martien-de-jong: why AIE2?

llvm/lib/Target/AIE/AIECombinerHelper.cpp
F-Stuckmann: oh yeah that should not happen. I now replace it with a implicit def, if the reg bank is nullptr

llvm/lib/Target/AIE/AIECombinerHelper.cpp
```diff
@@ -2125,6 +2125,40 @@ bool llvm::matchOffsetLoadStoreSharePtrAdd(MachineInstr &MI,
...
+bool llvm::matchCopyOfImplicitDef(MachineInstr &MI, MachineRegisterInfo &MRI) {
+  assert(MI.isCopy() && "Expected a COPY instruction");
+  const Register DstReg = MI.getOperand(0).getReg();
+  const Register SrcReg = MI.getOperand(1).getReg();
+  if (!DstReg.isVirtual() || !SrcReg.isVirtual())
+    return false;
+  const MachineInstr *SrcDef = MRI.getVRegDef(SrcReg);
+  if (!SrcDef || SrcDef->getOpcode() != TargetOpcode::G_IMPLICIT_DEF)
+    return false;
+  return MRI.getType(DstReg) == MRI.getType(SrcReg);
```
martien-de-jong, line 2137: nit: G_IMPLICIT_DEF doesn't have type-specific semantics, we can create it in the required type. Hence we don't need this condition.

llvm/lib/Target/AIE/AIECombinerHelper.cpp
```diff
@@ -2125,6 +2125,40 @@ bool llvm::matchOffsetLoadStoreSharePtrAdd(MachineInstr &MI,
...
+
+void llvm::applyCopyOfImplicitDef(MachineInstr &MI, MachineRegisterInfo &MRI,
+                                  MachineIRBuilder &B,
+                                  GISelChangeObserver &Observer) {
+  const Register DstReg = MI.getOperand(0).getReg();
+  const Register SrcReg = MI.getOperand(1).getReg();
+  const RegisterBank *DstRB = MRI.getRegBankOrNull(DstReg);
+  const RegisterBank *SrcRB = MRI.getRegBankOrNull(SrcReg);
+  if (DstRB == SrcRB) {
+    MRI.replaceRegWith(DstReg, SrcReg);
```
martien-de-jong, line 2148: nit: I don't there's merit in reusing IMPLICIT_DEF. I'd just always create a new one. It will blindly fulfill single-use constraints etc.

martien-de-jong (commented):
Please adjust the title to something like copy-propagate G_IMPLICIT_DEF.
I would simplify it to always create a pristine G_IMPLICIT_DEF with the proper type and regbank. Less conditions, less required  testing, no unnecessary multi-use nodes.

F-Stuckmann:
@martien-de-jong check out that the simplification breaks a CSE in the fixup commits.
 
Before we had a common Implicit def and CSE could delete common code, now with a fresh implicit def, we have to reproduce exaclty the same behaviour twice.

F-Stuckmann:
Removed the simplification because touching CSE is outside of the scope of this PR

martien-de-jong (approved):

PR #803, opened 2026-02-26
[MLIR][Quant] Add alias type printing for BlockFloat

jorickert:


mlir/test/Dialect/Quant/print-block-float-aliases.mlir
```diff
@@ -0,0 +1,46 @@
...
+// CHECK: tensor<4x!quant.block_float<mode=MX6, axis=2>>
+func.func @no_alias_mx6_other_axis(
+    %arg0: tensor<4x!quant.block_float<mode=MX6, axis=2>>)
+    -> tensor<4x!quant.block_float<mode=MX6, axis=2>> {
+  return %arg0 : tensor<4x!quant.block_float<mode=MX6, axis=2>>
+}
+
+// -----
+
+// CHECK: !bfp16_bf = !quant.block_float<mode=BFP16, axis=1>
```
jorickert: I am not very happy with the `_bf` suffix. As mlir does not allow an alias to end in a number, another approach would be to use `_` as suffix or adjust MLIR to allow aliases to end in a number

mlir/lib/Dialect/Quant/IR/TypeParser.cpp
```diff
@@ -472,3 +473,29 @@ void QuantDialect::printType(Type type, DialectAsmPrinter &os) const {
...
+namespace {
+struct QuantOpAsmDialectInterface : public OpAsmDialectInterface {
+  using OpAsmDialectInterface::OpAsmDialectInterface;
+
+  AliasResult getAlias(Type type, raw_ostream &os) const final {
+    const auto blockType = dyn_cast<BlockFloatQuantizedType>(type);
+    if (!blockType)
+      return AliasResult::NoAlias;
+    if (blockType.getAxis() != 1)
+      return AliasResult::NoAlias;
```
gerion-amd, line 722: Maybe we want this to encode the axis in the type (e.g. `mx6_1bf`) some time in the future, but I see no blocker for this with the current design.

mlir/test/Dialect/Quant/print-block-float-aliases.mlir
gerion-amd: Do you know if there is a hard requirement why this constraint is enforced? For types, it's quite common to end in numbers.

Otherwise, for me it's fine the way it is.
A few other suggestions just for fun:
- `mx6t` (t for "type")
- `mx6+`
- `mx6!`

gerion-amd (dismissed):
Thanks. I think, it's much easier to read with that.

mlir/test/Dialect/Quant/print-block-float-aliases.mlir
ttjost: I like `mx6!`, `bfp16!` !

ttjost (dismissed):
Could you check how `!` would look like as the last character? It looks prettier than what you have right now, IMO.

mlir/test/Dialect/Quant/print-block-float-aliases.mlir
jorickert: In commit b9f07f40ae3b074f42203491d22a83aa2670864c (i plan to upstream this one) i lifted the restriction that alias can not end in a number. Now it can simply be mx6 and bfp16

ttjost (approved):
I wonder if disambiguation is the only reason for not having the support for trailing digits in the first place. If it is, then LGTM, but I think it makes more sense to open a PR upstream to get an idea of why we didn't allow it in the first place. Then we can merge this safely.

jorickert:
> I wonder if disambiguation is the only reason for not having the support for trailing digits in the first place. If it is, then LGTM, but I think it makes more sense to open a PR upstream to get an idea of why we didn't allow it in the first place. Then we can merge this safely.

Upstream completely changed the implementation of the alias printing, its now part of the type not dialect. :-/

gerion-amd:
> > I wonder if disambiguation is the only reason for not having the support for trailing digits in the first place. If it is, then LGTM, but I think it makes more sense to open a PR upstream to get an idea of why we didn't allow it in the first place. Then we can merge this safely.
> 
> Upstream completely changed the implementation of the alias printing, its now part of the type not dialect. :-/

What is the consequence? Should we bump first or still merge? If bumping is not an option in the near future, I would still merge it, if we can achieve the same MLIR syntax after the bump.

PR #804, opened 2026-02-27
Add canonicalization pattern to fold reshape into concat if possible

jorickert:


mlir/lib/Dialect/Tosa/IR/TosaCanonicalizations.cpp
```diff
@@ -103,10 +106,114 @@ struct SelfConcatToTile : public OpRewritePattern<tosa::ConcatOp> {
...
+    auto reshapeOp =
+        dyn_cast<ReshapeOp>(*concatOp.getResult().getUsers().begin());
+    if (!reshapeOp)
+      return rewriter.notifyMatchFailure(
+          concatOp, "Sole user of concat is not a reshape");
+
+    const auto concatType = dyn_cast<ShapedType>(concatOp.getType());
+    if (!concatType || !concatType.hasStaticShape())
+      return rewriter.notifyMatchFailure(
+          concatOp, "Concat result must be a statically shaped");
```
ehsan-toosi:
```suggestion
          concatOp, "Concat result must be statically shaped");
```

mlir/lib/Dialect/Tosa/IR/TosaCanonicalizations.cpp
```diff
@@ -103,10 +106,114 @@ struct SelfConcatToTile : public OpRewritePattern<tosa::ConcatOp> {
...
+        return rewriter.notifyMatchFailure(
+            concatOp,
+            "Dimensions in [min(orig, new), max(orig, new)) are not all 1");
+    }
+
+    auto newConcatOp = rewriter.create<ConcatOp>(
+        rewriter.getFusedLoc({concatOp.getLoc(), reshapeOp.getLoc()}),
+        reshapeOp.getType(), concatOp.getInput1(), newConcatAxis);
+    rewriter.replaceOp(reshapeOp, newConcatOp.getResult());
+    rewriter.eraseOp(concatOp);
```
ehsan-toosi, line 217:
```suggestion
    // Concat is single user and it's safe to erase.
    rewriter.eraseOp(concatOp);
```

mlir/lib/Dialect/Tosa/IR/TosaCanonicalizations.cpp
```diff
@@ -103,10 +106,114 @@ struct SelfConcatToTile : public OpRewritePattern<tosa::ConcatOp> {
   }
 };
 
+// Folds reshape(concat(inputs, axis=a)) -> concat(inputs, axis=a') when the
+// reshape merely transfers a factor of N (the number of inputs) from one
+// dimension to another without reordering data.
+struct ConcatReshapeFusion : public OpRewritePattern<ConcatOp> {
+  using OpRewritePattern<ConcatOp>::OpRewritePattern;
```
ehsan-toosi, line 118: There is always going to be an argument between fold and canonicalization but this is more like a ::fold pattern to me for tosa reshape. Could you move it to ::fold?

mlir/lib/Dialect/Tosa/IR/TosaCanonicalizations.cpp
```diff
@@ -103,10 +106,114 @@ struct SelfConcatToTile : public OpRewritePattern<tosa::ConcatOp> {
...
+      return rewriter.notifyMatchFailure(concatOp,
+                                         "Did not find a new concat axis");
+
+    const int64_t numInputs = concatOp.getInput1().size();
+
+    if (concatOutputShape[origConcatAxis] !=
+        reshapeShape[origConcatAxis] * numInputs)
+      return rewriter.notifyMatchFailure(
+          concatOp, "Factor transferred from old to new axis does not match "
+                    "number of inputs");
```
ehsan-toosi, line 176: Shouldn't this be an assert considering the checks you have before?

mlir/test/Dialect/Tosa/canonicalize.mlir
```diff
@@ -1644,3 +1647,83 @@ func.func @canonicalize_select_to_clamp_i8_and_i64_pat2(%arg0: tensor<13x21x3xi8
...
+
+func.func @concat_reshape_fusion_rank_change(
+    %arg0: tensor<1x256x100xf32>, %arg1: tensor<1x256x100xf32>) -> tensor<512x100xf32> {
+  %0 = tosa.concat %arg0, %arg1 {axis = 0 : i32} : (tensor<1x256x100xf32>, tensor<1x256x100xf32>) -> tensor<2x256x100xf32>
+  %1 = tosa.reshape %0 {new_shape = array<i64: 512, 100>} : (tensor<2x256x100xf32>) -> tensor<512x100xf32>
+  return %1 : tensor<512x100xf32>
+}
+// CHECK-LABEL: @concat_reshape_fusion_rank_change
+// CHECK: tosa.concat
+// CHECK: tosa.reshape
```
ehsan-toosi, line 1729: Add a tiny test for multi-use concat to be complete on testing?

ehsan-toosi (dismissed):

mlir/lib/Dialect/Tosa/IR/TosaCanonicalizations.cpp
jorickert: I do not think this can be a fold, as in the fold i can only modify the op itself, but i need to modify both the concat and the use-def chain from the reshape to its user.

ehsan-toosi: Makes sense. Thanks

mlir/lib/Dialect/Tosa/IR/TosaCanonicalizations.cpp
```diff
@@ -103,10 +106,114 @@ struct SelfConcatToTile : public OpRewritePattern<tosa::ConcatOp> {
   }
 };
 
+// Folds reshape(concat(inputs, axis=a)) -> concat(inputs, axis=a') when the
+// reshape merely transfers a factor of N (the number of inputs) from one
+// dimension to another without reordering data.
```
gerion-amd, line 111: Can you make a small concrete example here?

mlir/lib/Dialect/Tosa/IR/TosaCanonicalizations.cpp
jorickert: No, the important part here is that the valid reshapes for fusing depend on the number of inputs, which we check here.
For example:
4 inputs 1x4x100, concat axis 0 → 4x4x100 -> Reshape to 2x8x100

Here all previous checks pass, but concatOutputShape[0] / reshapeShape[0] = 4 / 2 = 2  which is not equal  4 = numInputs .

This prevents us from trying to rewrite this concat to 1x4x100, concat axis 1 → 1x16x100 , which would not be correct.

But if we have 
2 inputs 1x4x100, concat axis 0 → 2x4x100 -> Reshape to 1x8x100
this would pass the check

mlir/lib/Dialect/Tosa/IR/TosaCanonicalizations.cpp
```diff
@@ -103,10 +106,115 @@ struct SelfConcatToTile : public OpRewritePattern<tosa::ConcatOp> {
...
+    int64_t newConcatAxis = -1;
+    for (size_t i = 0; i < rank; ++i) {
+      if (concatOutputShape[i] == reshapeShape[i])
+        continue;
+      if (i == origConcatAxis)
+        continue;
+      if (newConcatAxis != -1)
+        return rewriter.notifyMatchFailure(
+            concatOp, "Reshape changes more than two dimensions, cannot fuse");
+      newConcatAxis = i;
```
gerion-amd, line 159: While extremely unlikely, this can overflow.

mlir/test/Dialect/Tosa/canonicalize.mlir
```diff
@@ -1644,3 +1647,97 @@ func.func @canonicalize_select_to_clamp_i8_and_i64_pat2(%arg0: tensor<13x21x3xi8
...
+func.func @concat_reshape_fusion_multi_use(
+    %arg0: tensor<1x256x100xf32>,
+    %arg1: tensor<1x256x100xf32>) -> (tensor<1x512x100xf32>, tensor<2x256x100xf32>) {
+  %0 = tosa.concat %arg0, %arg1 {axis = 0 : i32} : (tensor<1x256x100xf32>, tensor<1x256x100xf32>) -> tensor<2x256x100xf32>
+  %1 = tosa.reshape %0 {new_shape = array<i64: 1, 512, 100>} : (tensor<2x256x100xf32>) -> tensor<1x512x100xf32>
+  return %1, %0 : tensor<1x512x100xf32>, tensor<2x256x100xf32>
+}
+// CHECK-LABEL: @concat_reshape_fusion_multi_use
+// CHECK: tosa.concat
+// CHECK: tosa.reshape
```
gerion-amd, line 1743: Can you also add a negative test for a multi-dim reshape?

mlir/lib/Dialect/Tosa/IR/TosaCanonicalizations.cpp
```diff
@@ -103,10 +106,115 @@ struct SelfConcatToTile : public OpRewritePattern<tosa::ConcatOp> {
...
+        return rewriter.notifyMatchFailure(
+            concatOp, "All concat inputs must be statically shaped");
+      if (inputType.getDimSize(origConcatAxis) != concatAxisSize)
+        return rewriter.notifyMatchFailure(
+            concatOp,
+            "Not all inputs have the same size along the concat axis");
+    }
+
+    // All input dimensions in [min(orig, new), max(orig, new)) must be 1 to
+    // ensure no data gets reordered.
```
gerion-amd: Can me make a small comment, why also the first dimension needs to be checked?
NIT: The outer brackets do not fit in `[min(orig, new), max(orig, new))`.

gerion-amd (dismissed):
Nice!

mlir/lib/Dialect/Tosa/IR/TosaCanonicalizations.cpp
jorickert: Lets hope nobody uses concats with such a high rank 😅

mlir/lib/Dialect/Tosa/IR/TosaCanonicalizations.cpp
jorickert: I extended the comment a bit. `[)` is intentional, its the notation for half-open intervals https://en.wikipedia.org/wiki/Interval_(mathematics)#Half-open_intervals

jorickert: I extended the comment a bit. `[)` is intentional, its the notation for half-open intervals https://en.wikipedia.org/wiki/Interval_(mathematics)#Half-open_intervals

ehsan-toosi (approved):

mlir/lib/Dialect/Tosa/IR/TosaCanonicalizations.cpp
```diff
@@ -103,10 +106,124 @@ struct SelfConcatToTile : public OpRewritePattern<tosa::ConcatOp> {
...
+
+    const int64_t numInputs = concatOp.getInput1().size();
+
+    if (concatOutputShape[origConcatAxis] !=
+        reshapeShape[origConcatAxis] * numInputs)
+      return rewriter.notifyMatchFailure(
+          concatOp, "Factor transferred from old to new axis does not match "
+                    "number of inputs");
+
+    assert(reshapeShape[newConcatAxis] ==
```
ttjost, line 178: this is irrelevant as we are already bailing on the `if` statement.
```suggestion
```

ttjost (approved):

PR #806, opened 2026-03-02
fixup! bitreverse non-power-of-2

abnikant:
Widened type to s32 before .lower() calls lowerBitreverse, so the bitreverse and masks all operate correctly at 32-bit width.


llvm/test/CodeGen/AIE/GlobalISel/legalize-bitswap.mir
```diff
@@ -43,8 +43,22 @@ body:             |
     ; CHECK-LABEL: name: test_bitswap_s16
     ; CHECK: liveins: $r0
     ; CHECK-NEXT: {{  $}}
-    ; CHECK-NEXT: [[C:%[0-9]+]]:_(s32) = G_CONSTANT i32 1280
-    ; CHECK-NEXT: [[TRUNC:%[0-9]+]]:_(s16) = G_TRUNC [[C]](s32)
+    ; CHECK-NEXT: [[C:%[0-9]+]]:_(s32) = G_CONSTANT i32 5
```
mludevid: Q: Why was the `G_BSWAP` of `5` correctly constant propagated to `1280` before and after the change that is no longer the case?

mludevid (dismissed):
Other than the lacking constant propagation, which is a pity, LGTM.

llvm/test/CodeGen/AIE/GlobalISel/legalize-bitswap.mir
abnikant: Updated the legalizer rules to only widen non-power-of-2 types (like s10), for constant folder is not able to optimize the expanded 4-byte swap sequence. With this change, the S16, S32 will avoid widen/shift/truncate overhead.

llvm/lib/Target/AIE/aie2p/AIE2PLegalizerInfo.cpp
```diff
@@ -291,8 +291,16 @@ AIE2PLegalizerInfo::AIE2PLegalizerInfo(const AIE2PSubtarget &ST)
       .clampScalar(0, S32, S32)
       .clampScalar(1, S32, S32);
 
-  getActionDefinitionsBuilder(G_BITREVERSE).lower();
-  getActionDefinitionsBuilder(G_BSWAP).lower();
+  getActionDefinitionsBuilder(G_BITREVERSE)
+      .lowerFor({S16, S32})
+      .widenScalarToNextPow2(0, /*Min=*/32)
+      .clampScalar(0, S32, S32)
+      .lower();
```
niwinanto: Do we need `.lower()`, I see `.lowerFor({S16, S32})` already present.

khallouh: Isn't `.clampScalar(0, S32, S32)` enough to force S32 type why do we need `widenScalarToNextPow2` as well?

khallouh: Q: Why is the Min Power of two st to 32 instead of 16 if .lower works fine for S16 already?

llvm/lib/Target/AIE/aie2p/AIE2PLegalizerInfo.cpp
```diff
@@ -291,8 +291,16 @@ AIE2PLegalizerInfo::AIE2PLegalizerInfo(const AIE2PSubtarget &ST)
...
+  getActionDefinitionsBuilder(G_BSWAP)
```
khallouh: Q: Why do we only see a diff in bitreverse and not in BSWAP?

mludevid: We did before reverting to the normal lowering for s16 and s32. @abnikant, could you add non-power of two tests?

llvm/lib/Target/AIE/aie2p/AIE2PLegalizerInfo.cpp
mludevid: There's special lowering and widening logic in generic LLVM code. So lowering s16 and s32 is handled automatically correctly and the widening to s32 is also done properly by LLVM. But to get the correct widening logic you need to call `widenScalarToNextPow2` or any other of the widening functions. This looks correct to me.

mludevid (commented):
Thanks for addressing my comment. Please add some more tests to show this new legalization logic.

llvm/lib/Target/AIE/aie2p/AIE2PLegalizerInfo.cpp
abnikant: for bitreverse test already exists. BSWAP minimum size should be 16-bits. We don't need widenScalarToNextPow2 for bswap. Updated the change and commit message.

llvm/lib/Target/AIE/aie2p/AIE2PLegalizerInfo.cpp
khallouh: I understand that it is correct. My 2 questions was whether  `.widenScalarToNextPow2(0, /*Min=*/16)` would be also correct (I guess so but to double check) and Why is `.clampScalar(0, S32, S32)` necessary?

khallouh: It seems .clampScalar(0, S32, S32) is necessary for larger widths which are not handled by Widen rule but there is no test covering this, would be nice to have it

abnikant: yes, .lowerFor({S16, S32}) handles byte-aligned types directly and .widenScalarToNextPow2(0, /*Min=*/32) widens everything else to s32, the .clampScalar guards against unexpected type widths.

abnikant: > It seems .clampScalar(0, S32, S32) is necessary for larger widths which are not handled by Widen rule but there is no test covering this, would be nice to have it

yes, added a test for s64.

abnikant: > Do we need `.lower()`, I see `.lowerFor({S16, S32})` already present.

yes, lowerFor is for S16 and S32, other types will go through lower.

khallouh (dismissed):
LGTM

llvm/lib/Target/AIE/aie2p/AIE2PLegalizerInfo.cpp
abnikant: We can simplify the code further - minScalar(0, S32) which already widens anything narrower than s32, so we can remove .widenScalarToNextPow2(0, /Min=/32) . I have pushed the change.

niwinanto (approved):
LGTM!

PR #807, opened 2026-03-02
[AIEX][NFC] SImplfy feature checks to exclude only AIE2

F-Stuckmann:
Combiner features are now only disabled for aie2, instead of explicitly allowing them for aie2p

### DIFF
--- a/llvm/lib/Target/AIE/AIECombinerHelper.cpp
+++ b/llvm/lib/Target/AIE/AIECombinerHelper.cpp
@@ -2902,14 +2902,14 @@ bool llvm::matchShuffleToExtractInsertEltToBroadcast(MachineInstr &MI,
     return false;
 
   unsigned MinFrequency;
-  if (MI.getMF()->getTarget().getTargetTriple().isAIE2P())
+  if (MI.getMF()->getTarget().getTargetTriple().isAIE2())
+    llvm_unreachable("MinFrequency unimplemented for target.");
+  else
     // The scalarization of G_SHUFFLE_VECTOR in the legalizer is more beneficial
     // if there are more exceptions than NumSrcElems / 2 as AIE2P's VINSERT
     // instrutions require a move to a register used for the index unlike VPUSH.
     MinFrequency = (ShuffleMaxNumInsertions != 0) ? ShuffleMaxNumInsertions
                                                   : NumSrcElems / 2;
-  else
-    llvm_unreachable("MinFrequency unimplemented for target.");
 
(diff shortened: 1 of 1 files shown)


llvm/lib/Target/AIE/AIECombinerHelper.cpp
```diff
@@ -3023,18 +3023,18 @@ bool llvm::matchMostlySequentialShuffleWithInsertions(MachineInstr &MI,
   const LLT ElemTy = Src1Ty.getElementType();
 
   unsigned MaxNumInsertions;
-  if (MI.getMF()->getTarget().getTargetTriple().isAIE2P())
+  if (MI.getMF()->getTarget().getTargetTriple().isAIE2())
```
khallouh, line 3026: Note: This assumes we know that every future target will have the same feature as AIE2P regarding VINSERT requiring a register. If not, the combine would be suboptimal.

khallouh: That unreachable was not meant to disable the combine for other targets, but rather to invite the developer to check if the max number of insertions is in line with the target architecture and adjust accordingly. If we don't care about any non AIE2 architecture, I would rather remove the if condition altogether and set the same Max number of insertions regardless of the target. We don't track AIE2 QoR after all.

khallouh: Also, that combine is not even added to the AIE2 list

mludevid:
Instead of replacing target checks in the combiner code with different target checks, I'd prefer moving this information to target hooks. That would also allow maintaining one list of combiners for all targets and deactivating specific combiners for specific targets through these target hooks. E.g. if a combiner requires a target to have immediate indices for the insert instruction for the combine to be beneficial, let's add a target hook that informs the combiner if the target we are currently compiling for has these immediate indices.

If efficiency is a concern we could even look into how to deactivate the entire combine by querying that target hook, to avoid the combine to match all possible candidates before bailing out due to the target restrictions.

F-Stuckmann:
@mludevid I agree this is also hacky and the correct way would be to have generic opcodes that we legalize in ISEL.
I will close this PR

mludevid:
@F-Stuckmann, how does having generic opcodes help deciding whether a certain combine is worth it in a specific architecture? Maybe I'm missing some details, to be honest I haven't looked into the details of these combines and what target instruction they require to be beneficial. Also, "legalize in ISEL", what does that mean?

PR #809, opened 2026-03-02
Windows: allow CI runner to find dev tools

thomthehound:
This small patch initializes the MSVC developer environment on the windows-2022 runner so the Windows wheel build can actually find the required compiler tools (`cl`, etc.).

Fresh Windows wheels do not appear to have been published for quite some time. Restoring them will directly support the downstream [Windows bring-up](https://github.com/Xilinx/mlir-aie/pull/2861) effort in mlir-aie.

This is probably not the only fix still needed in the Windows distribution CI, but it should remove the immediate blocker. I will be monitoring the CI log to identify other issues this surfaces.


thomthehound:
@konstantinschwarz and @philippjh

Dear Sirs,
I do not wish to be a bother about this, but it would be very helpful to me if I could understand why this is being ignored.

konstantinschwarz (approved):
This looks good to me, sorry for the delay @thomthehound !

PR #811, opened 2026-03-03
Enable Multi Slot Pseudo materialization by default

F-Stuckmann:
Enable Multi Slot Materialization by default. 

-0.63% Core Compute Insn Count. 
 

llvm/lib/Target/AIE/AIEMultiSlotInstrMaterializer.cpp
```diff
@@ -198,52 +199,53 @@ namespace {
...
-  // Reverse lookup of the alternative that matches BestSlot.
-  auto FindOpcode = [TII, Opcode = MSP->getOpcode()](int BestSlot) {
-    const AIEBaseMCFormats *Formats = TII->getFormatInterface();
-    auto *Alternatives = Formats->getAlternateInstsOpcode(Opcode);
-    for (auto Opcode : *Alternatives) {
-      if (Formats->getSlotKind(Opcode) == MCSlotKind(BestSlot)) {
-        return Opcode;
-      }
+
+  // Reverse lookup: find the alternative opcode for BestSlot.
```
martien-de-jong: I think you can pick up BestOpcode in the earlier traversal.

llvm/lib/Target/AIE/AIEPostPipeliner.cpp
```diff
@@ -191,7 +191,9 @@ bool isSideEffectFree(MachineInstr *MI) {
 } // namespace
 
 int PostPipeliner::getResMII(MachineBasicBlock &LoopBlock) {
-  // Add up all slot requirements and return the maximum slot count
+  // Add up per-slot instruction counts using primary slot sets (not conflict
+  // sets) so the lower bound reflects actual slot occupancy, not cross-slot
+  // interference.
```
martien-de-jong: I'd say that using conflict counts here would make perfect sense. If I have X, M and XM in the block, ResMII should be two, not one. and since X and M are always >= XM, the extra XM count will never dominate ResMII.

llvm/test/CodeGen/AIE/aie2/schedule/postpipeliner/conv2d_bf16-1.mir
```diff
@@ -38,12 +38,12 @@
   ; CHECK-NEXT:    vldb wh1, [p0, #32]
   ; CHECK-NEXT:    vldb wl1, [p0], m4
   ; CHECK-NEXT:    vldb wh3, [p0, #32]; add.nc lc, r0, #-1
-  ; CHECK-NEXT:    vlda.3d wl3, [p0], d1; movxm ls, #.LBB0_2
+  ; CHECK-NEXT:    vldb.3d wl3, [p0], d1; movxm ls, #.LBB0_2
```
martien-de-jong, line 41: It looks as if we have swapped A and B almost everywhere. I guess the postscheduler materialized first first alternative, here we do first fit slot?

llvm/lib/Target/AIE/aie2ps/AIE2PSInstrInfo.cpp
```diff
@@ -1853,3 +1853,7 @@ AIE2PSInstrInfo::parseAbstractOp(const MachineInstr &MI) const {
   }
   return std::nullopt;
 }
+
+bool AIE2PSInstrInfo::isPseudoFillLoad(unsigned Opc) const {
+  return Opc == AIE2PS::VLD_FILL_512_pseudo;
+}
```
martien-de-jong: It's a bit lonely here. Let's put it next to its other is... cousins, shall we?

llvm/lib/Target/AIE/AIEMultiSlotInstrMaterializer.cpp
```diff
@@ -196,6 +196,47 @@ bool assignSlots(SlotMapping &SlotToBanks, const MachineBasicBlock &MBB,
...
 
 namespace {
+
+/// Separate FILL pseudo instructions from different FIFO streams into
+/// different load slots using round-robin assignment. This ensures FILL
+/// operations from independent streams land on complementary slots so the
+/// pipeliner can pack them into the same VLIW bundle.
+///
+/// \return true if FILL instructions from multiple streams were separated.
+static bool separateFillStreams(MachineBasicBlock &MBB,
```
martien-de-jong: nit: You're in anonymous namespace. Leaving out static may just fit it on one line.

llvm/lib/Target/AIE/AIEMultiSlotInstrMaterializer.cpp
```diff
@@ -196,6 +196,47 @@ bool assignSlots(SlotMapping &SlotToBanks, const MachineBasicBlock &MBB,
...
+                                const AIEBaseInstrInfo *TII) {
+  const SmallVector<MCSlotKind, 2> LoadSlots =
+      TII->getFormatInterface()->getLoadSlotKinds();
+  assert(LoadSlots.size() >= 2 && "Need at least two load slots");
```
martien-de-jong: Why not just return false?

llvm/lib/Target/AIE/AIEMultiSlotInstrMaterializer.cpp
```diff
@@ -196,6 +196,47 @@ bool assignSlots(SlotMapping &SlotToBanks, const MachineBasicBlock &MBB,
...
+
+  LLVM_DEBUG(dbgs() << "Separating " << Fills.size()
+                    << " FILL streams into different load slots\n");
+
+  // Round-robin: assign each FILL to the next available load slot.
+  unsigned SlotIdx = 0;
+  for (auto *Fill : Fills) {
+    const MCSlotKind Slot = LoadSlots[SlotIdx++ % LoadSlots.size()];
+    const auto OpCode = TII->getSlotOpcode(Slot, *Fill);
+    if (!OpCode)
```
martien-de-jong: This says we can't materialize in the current slot.
A- we could still try in the other slot
B- we have construed this instruction to be materializable in one of the load slots. I guess each Fill instruction can use SlotIdx to index into its own alternatives. This would relax the assumptions.
I would at least document the `isPseudoFillLoad` interface with the expectations we have here, so at least that it should have more than one alternative.

llvm/lib/Target/AIE/AIEMultiSlotInstrMaterializer.cpp
```diff
@@ -324,6 +365,12 @@ void staticallyMaterializeMultiSlotInstructions(MachineBasicBlock &MBB,
     LLVM_DEBUG(
         dbgs()
         << "Could not find slot assignments, skipping bank materialization\n");
+    // Separate FILL instructions from different FIFO streams into different
+    // slots. Remaining POP pseudos are left for the pipeliner to assign
+    // dynamically, which produces better slot balance than the greedy
+    // fallback.
+    if (separateFillStreams(MBB, TII))
+      return;
```
martien-de-jong: You short-circuit MaterializeAll with this return.

llvm/lib/Target/AIE/AIEPostPipeliner.cpp
F-Stuckmann: offline: unfortunately X and M <= XM because the XM Conflict bits a X Conflicts + M Conflicts and so it dominates the RESMII and is too conservative for the postpipeliner

llvm/test/CodeGen/AIE/aie2p/end-to-end/ore-hardware-loops.ll
```diff
@@ -2,10 +2,12 @@
 ; See https://llvm.org/LICENSE.txt for license information.
 ; SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
 ;
-; (c) Copyright 2025 Advanced Micro Devices, Inc. or its affiliates
+; (c) Copyright 2025-2026 Advanced Micro Devices, Inc. or its affiliates
+; RUN: llc -mtriple=aie2p --aie-force-postpipeliner %s -o - | FileCheck %s --check-prefix asm
```
andcarminati: Could we use upper-case ASM and REMARKS?

andcarminati:
What is the effect on GEMM_int8? Also, as pointed by @martien-de-jong, we could try to restrict this materialization for swp candidates, as we already have on-the-fly materialization in our hazard recognizer for normal non-candidate cases.

llvm/lib/Target/AIE/AIEMultiSlotInstrMaterializer.cpp
```diff
@@ -198,52 +199,53 @@ namespace {
...
   };
-  int BestSlot = 0;
-  for (int S = 1; S < Counts.size(); S++) {
-    if (Better(S, BestSlot)) {
+
+  // Iterate only over the primary slots of the MSP's alternatives.
+  // With conflict-set-based statistics, secondary (conflict) indices
+  // would also appear non-zero in Counts, but they are not valid
+  // materialization targets.
+  int BestSlot = static_cast<int>(Formats->getSlotKind(Alternatives->front()));
```
andcarminati: nit: maybe a lambda to map opcode to slot, so you can use it on the lines 221 and 223.

F-Stuckmann:
the materialize all option is already just for the pipeline loops. I renamed the option name to clarify what we are actually doing

llvm/lib/Target/AIE/AIEMultiSlotInstrMaterializer.cpp
```diff
@@ -196,6 +196,44 @@ bool assignSlots(SlotMapping &SlotToBanks, const MachineBasicBlock &MBB,
...
+                    << " FILL streams into different load slots\n");
+
+  // Round-robin over each FILL's own alternatives so that consecutive
+  // FILLs land on complementary slots.
+  unsigned SlotIdx = 0;
+  for (auto *Fill : Fills) {
+    const auto *Alts = Formats->getAlternateInstsOpcode(Fill->getOpcode());
+    if (!Alts || Alts->size() < 2)
+      continue;
+    const unsigned AltOpcode = (*Alts)[SlotIdx++ % Alts->size()];
```
martien-de-jong: nit: SlotIdx has no fixed relations with the slots. It's an alternative index, and may not actually anti-correlate  neighbours. Better would be to collect the slots and select the alternative that leads to the slot not chosen by the previous assignment.

llvm/lib/Target/AIE/AIEMultiSlotInstrMaterializer.cpp
```diff
@@ -319,6 +357,11 @@ void staticallyMaterializeMultiSlotInstructions(MachineBasicBlock &MBB,
     LLVM_DEBUG(
         dbgs()
         << "Could not find slot assignments, skipping bank materialization\n");
+    // Separate FILL instructions from different FIFO streams into different
+    // slots. Remaining POP pseudos are left for the pipeliner to assign
+    // dynamically, which produces better slot balance than the greedy
+    // fallback.
```
martien-de-jong: I don't think we discriminate different FIFO streams. I think we leave any non-FILL (in stead of only POP) instruction unmaterialized at this point. The pipeliner doesn't assign dynamically, and it is not certain whether the materialization is done by MaterializeAll or by loop aware scheduling.
We could give the TII hook a more neutral name by describing what it does: apply round robin slot assignment to the instruction selected by the callback.
Even if this is a target-specific ad-hoc heuristic, we can engineer it in a generic way.

llvm/lib/Target/AIE/AIEMultiSlotInstrMaterializer.cpp
```diff
@@ -286,6 +287,33 @@ void materializeMSP(MachineInstr *MSP, SlotStatistics &Statistics,
...
+      const SlotStatistics PreheaderStats = computeSlotStatistics(
+          const_cast<MachineBasicBlock &>(*Preheader), TII);
+      // Only carry over the load slot pressure from the preheader.
+      // Non-load slots (ALU, Lng, etc.) should not bias MOV/other
+      // pseudo materialization decisions.
+      for (const MCSlotKind Slot : LoadSlots)
+        PreheaderFixed[static_cast<int>(Slot)] =
+            PreheaderStats.Fixed.at(static_cast<int>(Slot));
+      LLVM_DEBUG(dbgs() << "Preheader Fixed:\n  " << PreheaderFixed << "\n");
+      Statistics.Fixed += PreheaderFixed;
```
martien-de-jong: By adding it in this way you give equal weight to preheader and loop body. 
I think the loop should have absolute priority, only checking (separate) preheader counts if we have a choice.
I'm also not sure why the mechanism should be confined to the load slots. I also expect preheader interference in the move slot, which sets up loop parameters in the preheader and performs simple vector operations in the loop body.
In the same vein, the store slot is frequently busy in the epilogue, and could be avoided for pointer arithmetic. I don't see any contribution from the epilogue here, whereas the problem is symmetric. I thought we discussed to have a secondary decision based on prologue and epilogue, perhaps shifting from prologue to both to epilogue based on the stage in which an instruction runs.

llvm/lib/Target/AIE/AIEMultiSlotInstrMaterializer.cpp
```diff
@@ -284,20 +285,64 @@ void materializeMSP(MachineInstr *MSP, SlotStatistics &Statistics,
...
+                                      int LoadResMII) {
+  int MinFixed = INT_MAX;
+  int MaxFixed = 0;
+  for (const MCSlotKind Slot : LoadSlots) {
+    const int Fixed = Statistics.Fixed.at(static_cast<int>(Slot));
+    MinFixed = std::min(MinFixed, Fixed);
+    MaxFixed = std::max(MaxFixed, Fixed);
+  }
+  const bool Balanced = MinFixed >= MaxFixed;
+
```
martien-de-jong: You can do an early return here. Nit: MinFixed > MaxFixed never occurs, so >-= looks strange.

llvm/lib/Target/AIE/AIEMultiSlotInstrMaterializer.cpp
```diff
@@ -198,52 +199,48 @@ namespace {
...
+  auto SlotOf = [Formats](unsigned Opc) {
+    return static_cast<int>(Formats->getSlotKind(Opc));
+  };
+
+  // Compare conflict-aware slot pressure: prefer the slot whose primary
+  // index has lower Fixed utilization, breaking ties by lower Free.
+  // Fixed and Free already use conflict sets, so they reflect cross-slot
+  // interference (e.g. Lng blocking Alu and Mv).
+  auto Better = [&Statistics](int A, int B) {
+    if (Statistics.Fixed.at(A) == Statistics.Fixed.at(B))
```
martien-de-jong: why introduce noise by removing brackets?

llvm/lib/Target/AIE/AIEMultiSlotInstrMaterializer.cpp
martien-de-jong: I think we found that the numbers proved me wrong. Effectively we still add it with equal weight in the benefit comparison, but they are at least presented separately, without the need to subtract it out again in a later stage.

llvm/lib/Target/AIE/AIEMultiSlotInstrMaterializer.cpp
```diff
@@ -290,13 +291,93 @@ void materializeMSP(MachineInstr *MSP, SlotStatistics &Statistics,
...
+bool hasAlternativeReachedThreshold(const SlotStatistics &Statistics,
+                                    const SlotCounts &InitialTotal,
+                                    ArrayRef<unsigned> Alternatives,
+                                    const AIEBaseMCFormats *Formats) {
+  const int N = Alternatives.size();
+  for (const unsigned AltOpc : Alternatives) {
+    const int S = static_cast<int>(Formats->getSlotKind(AltOpc));
+    const int FairShare = computeFairShare(InitialTotal, S, N);
+    const bool HasLoopBodyPressure = FairShare > 0;
+    const bool ReachedFairShare = Statistics.Fixed.at(S) >= FairShare;
```
martien-de-jong: Nit: instead of dividing on the right hand side of the comparison we could multiply on the left.

llvm/lib/Target/AIE/AIEMultiSlotInstrMaterializer.cpp
```diff
@@ -290,13 +291,93 @@ void materializeMSP(MachineInstr *MSP, SlotStatistics &Statistics,
...
+    if (!HasBias || HasLoopBodyPressure)
+      continue;
+    LLVM_DEBUG(dbgs() << "Removing neighbor bias for slot " << S
+                      << " (no loop body pressure)"
+                      << " [Fixed=" << Statistics.Fixed.at(S) << "]\n");
+    NeighborFixed[S] = 0;
+  }
+
+  if (!hasAlternativeReachedThreshold(Statistics, InitialTotal, Alternatives,
+                                      Formats))
```
martien-de-jong: nit: can I have { bracketed syntax } here?

llvm/lib/Target/AIE/aie2p/AIE2PInstrInfo.cpp
```diff
@@ -307,7 +307,8 @@ bool AIE2PInstrInfo::isFifoStoreConvOpcode(unsigned Opcode) const {
           (Opcode == AIE2P::VST_PUSH_576_CONV_bfp16ebs8_fp32));
 }
 
-bool AIE2PInstrInfo::isRoundRobinSlotCandidate(unsigned Opc) const {
+bool AIE2PInstrInfo::isRoundRobinSlotCandidate(const MachineInstr &MI) const {
```
martien-de-jong: int, for future compatibility? People will need to retype their overrides.

llvm/lib/Target/AIE/AIEBaseInstrInfo.h
```diff
@@ -183,7 +183,9 @@ struct AIEBaseInstrInfo : public TargetInstrInfo {
   /// Selected instructions must have alternatives in at least two different
   /// slots so that the multi-slot materializer can separate consecutive
   /// candidates onto complementary slots.
-  virtual bool isRoundRobinSlotCandidate(unsigned Opc) const { return false; }
+  virtual bool isRoundRobinSlotCandidate(const MachineInstr &MI) const {
+    return false;
+  }
```
martien-de-jong, line 185: please, leave out the const on the MachineInstr. You block people from using fairly basic methods.

F-Stuckmann: but I can pass a non-const MI into `isRoundRobinSlotCandidate`.

F-Stuckmann: It just says, that MI is not being changed inside of `isRoundRobinSlotCandidate`, see our call of the method in `AIEMultiSlotInstrMaterializer.cpp`

llvm/lib/Target/AIE/aie2p/AIE2PInstrInfo.cpp
F-Stuckmann: good thinking

llvm/lib/Target/AIE/AIEMultiSlotInstrMaterializer.cpp
F-Stuckmann: that is nicer

martien-de-jong (approved):

PR #812, opened 2026-03-03
[AIE2PS] Port back https://github.com/Xilinx/llvm-aie/pull/785

martien-de-jong:
This allows storing s256, which is for some reason created by the llvm optimizers


F-Stuckmann (approved):
Nice

PR #814, opened 2026-03-04
[AIEX][AIE2PS] Refactor additional aie2p/aie2ps combiners

andcarminati:
This is also a preparation for the next batch of combiners (https://github.com/Xilinx/llvm-aie/pull/772).

### DIFF
--- a/llvm/lib/Target/AIE/AIECombine.td
+++ b/llvm/lib/Target/AIE/AIECombine.td
@@ -448,7 +448,8 @@ def aie2_postlegalizer_custom_combines : GICombineGroup<[
   combine_upd_to_concat
 ]>;
 
-def aie2p_postlegalizer_custom_combines : GICombineGroup<[
+// Post-legalizer custom combines for AIE2P and more recent targets (AIE2PS, etc.).
+def aie2p_plus_postlegalizer_custom_combines : GICombineGroup<[
   combine_extract_vector_assert_combine,
   combine_extract_broadcast_to_scalar
 ]>;
@@ -460,7 +461,7 @@ def AIE2PostLegalizerCustomCombiner
 
 def AIE2PPostLegalizerCustomCombiner
     : GICombiner<"AIE2PPostLegalizerCustomCombinerImpl",
-                 [aie_postlegalizer_custom_shared_combines, aie2p_postlegalizer_custom_combines]> {
+                 [aie_postlegalizer_custom_shared_combines, aie2p_plus_postlegalizer_custom_combines]> {
 }
 
 def AIE2PSPreLegalizerCombiner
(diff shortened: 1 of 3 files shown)


llvm/lib/Target/AIE/AIECombine.td
```diff
@@ -448,7 +448,7 @@ def aie2_postlegalizer_custom_combines : GICombineGroup<[
   combine_upd_to_concat
 ]>;
 
-def aie2p_postlegalizer_custom_combines : GICombineGroup<[
+def aie2pplus_postlegalizer_custom_combines : GICombineGroup<[
```
mludevid: Why `aie2pplus`?

mludevid (dismissed):
LGTM. Since this is adding new combines for AIE2PS, did you check the effect this has on QoR?

llvm/lib/Target/AIE/AIECombine.td
andcarminati: It means everything that start with aie2p, but not before. Naming is always the hard part.

andcarminati:
> LGTM. Since this is adding new combines for AIE2PS, did you check the effect this has on QoR?

Not yet, but simplifications are welcome and we may see some improvements on the next runs.

llvm/lib/Target/AIE/AIECombine.td
mludevid: Alright. Naming is indeed difficult. I also don't have a much better suggestion. Maybe separate it with an underscore (`aie2p_plus_postlegalizer_custom_combines`) for readability and add a comment?

mludevid:
> > LGTM. Since this is adding new combines for AIE2PS, did you check the effect this has on QoR?
> 
> Not yet, but simplifications are welcome and we may see some improvements on the next runs.

Okay. I'm fine with merging this anyways and tweaking any combine in a follow-up if we see regressions.

mludevid (approved):
Thanks! LGTM.

llvm/test/CodeGen/AIE/GlobalISel/combine-broadcast-extract.mir
```diff
@@ -6,6 +6,7 @@
 #
 # (c) Copyright 2025-2026 Advanced Micro Devices, Inc. or its affiliates
 # RUN: llc -mtriple=aie2p -run-pass=aie2p-postlegalizer-custom-combiner -global-isel %s -verify-machineinstrs -o - | FileCheck %s
+# RUN: llc -mtriple=aie2ps -run-pass=aie2ps-postlegalizer-custom-combiner -global-isel %s -verify-machineinstrs -o - | FileCheck %s
```
katerynamuts, line 9: There are tests with intrinsics for aie2p and no for aie2ps. For this combine we also need to update the whitelist in `mayMIShiftElements` in AIECombinerHelper.cpp for aie2ps.

PR #816, opened 2026-03-04
[AIE2PS] Implement getSMSMutations subtarget hook

andcarminati:
This helps PreSWP to reach better schedules and also discover better candidates for PostSWP.

### DIFF
--- a/llvm/lib/Target/AIE/aie2ps/AIE2PSSubtarget.h
+++ b/llvm/lib/Target/AIE/aie2ps/AIE2PSSubtarget.h
@@ -85,6 +85,11 @@ public:
         AIEBaseSubtarget::getPostRAMutationsImpl(getTargetTriple(), nullptr);
   }
 
+  void getSMSMutations(std::vector<std::unique_ptr<ScheduleDAGMutation>>
+                           &Mutations) const override {
+    Mutations = AIEBaseSubtarget::getSMSMutationsImpl(getTargetTriple());
+  }
+
   // Finer-grained live ranges, giving more accurate use-def chains.
   bool enableSubRegLiveness() const override { return true; }
 


llvm/lib/Target/AIE/aie2ps/AIE2PSSubtarget.h
```diff
@@ -85,6 +85,11 @@ class AIE2PSSubtarget : public AIE2PSGenSubtargetInfo, public AIEBaseSubtarget {
         AIEBaseSubtarget::getPostRAMutationsImpl(getTargetTriple(), nullptr);
   }
 
+  void getSMSMutations(std::vector<std::unique_ptr<ScheduleDAGMutation>>
+                           &Mutations) const override {
+    Mutations = AIEBaseSubtarget::getSMSMutationsImpl(getTargetTriple());
```
F-Stuckmann, line 90: since we either way go to the AIEBaseSubTarget, why do we have to override it?

andcarminati: This is a long story related to our inheritance model and one class cannot implement methods related to a different class hierarchy. See `class AIE2PSSubtarget : public AIE2PSGenSubtargetInfo, public AIEBaseSubtarget`, actually we cannot inject a AIEBaseSubtarget into AIE2PSGenSubtargetInfo and cannot make AIE2PSGenSubtargetInfo the AIEBaseSubtarget superclass.

F-Stuckmann:
I am working on a PR to fix this whole SubTarget Inheritance problem

F-Stuckmann:
[AIESubTarget Simplification ](https://github.com/Xilinx/llvm-aie/pull/818) should make this PR redundant

martien-de-jong:
I agree. This should be taken from BaseSubtarget directly. In particular, https://github.com/Xilinx/llvm-aie/pull/826 should be equivalent to this one, hence I close this.

PR #820, opened 2026-03-05
 [AIEX] Implement TTI hook to avoid creating IVs with load and store use

andcarminati:
AIE prefers to keep separate pointer IVs when one is used for loads and the other for stores, as this enables better instruction scheduling and utilization of independent address generation units. This is specially beneficial for PostSWP.


llvm/test/CodeGen/AIE/opt/congruent-iv-merge.ll
```diff
@@ -18,10 +18,12 @@ define void @separate_load_store(ptr %arr, i32 %n) {
 ; CHECK:       loop:
 ; CHECK-NEXT:    [[I:%.*]] = phi i32 [ 0, [[ENTRY:%.*]] ], [ [[I_NEXT:%.*]], [[LOOP]] ]
 ; CHECK-NEXT:    [[STORE_PTR:%.*]] = phi ptr [ [[ARR:%.*]], [[ENTRY]] ], [ [[STORE_PTR_NEXT:%.*]], [[LOOP]] ]
+; CHECK-NEXT:    [[STORE_PTR1:%.*]] = phi ptr [ [[ARR]], [[ENTRY]] ], [ [[STORE_PTR_NEXT1:%.*]], [[LOOP]] ]
 ; CHECK-NEXT:    [[VAL:%.*]] = load i32, ptr [[STORE_PTR]], align 4
 ; CHECK-NEXT:    [[RESULT:%.*]] = add i32 [[VAL]], 1
-; CHECK-NEXT:    store i32 [[RESULT]], ptr [[STORE_PTR]], align 4
+; CHECK-NEXT:    store i32 [[RESULT]], ptr [[STORE_PTR1]], align 4
```
F-Stuckmann, line 24: why would we want to increase register pressure? 
What structure is this PR fixing? Do we not do pointer chaining between load and store chains?

F-Stuckmann: have you run QoR? the last time i increased register pressure inside of loops I got a bloody nose

andcarminati: Target case shared offline. No QoR numbers available yet.

llvm/lib/Analysis/TargetTransformInfo.cpp
```diff
@@ -1442,6 +1442,11 @@ bool TargetTransformInfo::isProfitableToSinkOperands(
   return TTIImpl->isProfitableToSinkOperands(I, OpsToSink);
 }
 
+bool TargetTransformInfo::shouldMergeCongruentIVs(const PHINode *IV1,
+                                                  const PHINode *IV2) const {
```
F-Stuckmann, line 1446: I am wondering if moving this to the bottom of the file may lead to less merge conflicts?

llvm/lib/Target/AIE/AIEBaseTargetTransformInfo.cpp
```diff
@@ -64,6 +64,12 @@ static cl::opt<unsigned> PreferSwpOverUnroll(
     "aie-prefer-swp-over-unroll", cl::Hidden, cl::init(9),
     cl::desc("Aim for pipelining if MinIterCount is at least this value."));
 
+static cl::opt<bool> MergeCongruentIVs(
+    "aie-merge-congruent-ivs", cl::Hidden, cl::init(false),
```
F-Stuckmann, line 68: nit: can we have a positive name for the option? 
Somthing like `aie-separate-congruent-ivs` ? and setting it to true enables the behavior? 
maybe also mention load/store IV separation in the argument name?

llvm/lib/Target/AIE/AIEBaseTargetTransformInfo.cpp
```diff
@@ -296,3 +302,64 @@ InstructionCost AIETTICommon::getMemoryOpCost(unsigned Opcode, Type *Src,
...
+    Visited.insert(IV);
+
+    while (!Worklist.empty()) {
+      const Value *V = Worklist.pop_back_val();
+
+      for (const Use &U : V->uses()) {
+        const User *Usr = U.getUser();
+
+        if (auto *LI = dyn_cast<LoadInst>(Usr)) {
+          if (LI->getPointerOperand() == V)
```
F-Stuckmann: nit: HasLoads |= LI->getPointerOperand() == V;

F-Stuckmann (commented):
Nice

llvm/include/llvm/Analysis/TargetTransformInfoImpl.h
```diff
@@ -380,6 +380,10 @@ class TargetTransformInfoImplBase {
 
   bool isProfitableToHoist(Instruction *I) const { return true; }
 
+  bool shouldMergeCongruentIVs(const PHINode *IV1, const PHINode *IV2) const {
+    return true; // Default: always merge (preserves current behavior)
```
martien-de-jong, line 384: nit: I don't like trailing comments

llvm/lib/Transforms/Utils/ScalarEvolutionExpander.cpp
```diff
@@ -1786,6 +1789,14 @@ SCEVExpander::replaceCongruentIVs(Loop *L, const DominatorTree *DT,
     if (OrigPhiRef->getType()->isPointerTy() != Phi->getType()->isPointerTy())
       continue;
 
+    // Check target-specific preference
+    if (TTI && !TTI->shouldMergeCongruentIVs(Phi, OrigPhiRef)) {
+      SCEV_DEBUG_WITH_TYPE(
+          DebugType, dbgs() << "INDVARS: Skipping congruent iv merge due to "
+                            << "target preference: " << *Phi << '\n');
+      continue;
+    }
```
martien-de-jong, line 1798: This is a very local decision. should we also have an entry call that allows you to reset counts etc. per loop header?
Or perhaps split the loop in candidate selection  and rewrite with a user filter in between?

llvm/lib/Target/AIE/AIEBaseTargetTransformInfo.cpp
```diff
@@ -296,3 +302,64 @@ InstructionCost AIETTICommon::getMemoryOpCost(unsigned Opcode, Type *Src,
...
+          if (Visited.insert(Usr).second)
+            Worklist.push_back(Usr);
+        }
+      }
+    }
+
+    return {HasLoads, HasStores};
+  };
+
+  auto [IV1HasLoads, IV1HasStores] = AnalyzePointerUses(IV1);
```
martien-de-jong, line 351: I think you can return early

llvm/lib/Target/AIE/AIEBaseTargetTransformInfo.cpp
```diff
@@ -296,3 +302,64 @@ InstructionCost AIETTICommon::getMemoryOpCost(unsigned Opcode, Type *Src,
...
+  auto [IV2HasLoads, IV2HasStores] = AnalyzePointerUses(IV2);
+
+  // Check if currently separated (one IV for loads only, other for stores only)
+  bool CurrentlySeparated =
```
martien-de-jong: nit: const

llvm/lib/Target/AIE/AIEBaseTargetTransformInfo.cpp
```diff
@@ -296,3 +302,64 @@ InstructionCost AIETTICommon::getMemoryOpCost(unsigned Opcode, Type *Src,
...
+      (IV1HasLoads && !IV1HasStores && IV2HasStores && !IV2HasLoads) ||
+      (IV2HasLoads && !IV2HasStores && IV1HasStores && !IV1HasLoads);
```
martien-de-jong: I think it would help readability having named pairs IV1Class and IV2Class with IsDedicatedLoad(Class) and IsDedicatedStore(Class).  The early return(s) could run of IsMixed(Class) which covers two of the four combinations.

llvm/lib/Target/AIE/AIEBaseTargetTransformInfo.h
```diff
@@ -114,6 +121,14 @@ template <typename T> class AIEBaseTTIImpl : public BasicTTIImplBase<T> {
     // Default return of allowsMisalignedMemoryAccesses is false.
     return ChainSizeInBytes >= 4;
   }
+
```
martien-de-jong, line 124: I would like an precise definition of 'congruent' here. I have a feeling that it actually means 'equivalent' i.e. they have the same value on every iteration, which is different from 'make the same strep' on every iteration.

llvm/lib/Transforms/Utils/ScalarEvolutionExpander.cpp
andcarminati: This is `replaceCongruentIVs`, we skip using other conditions as well, like pointers with different types. I can't see a real benefit on this separation.

llvm/lib/Target/AIE/AIEBaseTargetTransformInfo.cpp
andcarminati: I feel that this is more confusing, as it uses a flipped meaning.

llvm/lib/Analysis/TargetTransformInfo.cpp
andcarminati: I guess there is no direct impact.

llvm/include/llvm/Analysis/TargetTransformInfoImpl.h
andcarminati: removed.

llvm/lib/Target/AIE/AIEBaseTargetTransformInfo.cpp
andcarminati: I rewrote this part.

andcarminati:
Hi @F-Stuckmann and @martien-de-jong, I addressed all the comments, thank you!

martien-de-jong (approved):

PR #824, opened 2026-03-06
[AIEX] fixup! matchNarrowTruncLoad

abnikant:
Fix an assertion failure ("invalid narrowing extend") in `matchNarrowTruncLoad`. Add an early return when the load's destination register is already s20.

### DIFF
--- a/llvm/test/CodeGen/AIE/GlobalISel/combine-trunc-load-s20-no-narrow.mir
+++ b/llvm/test/CodeGen/AIE/GlobalISel/combine-trunc-load-s20-no-narrow.mir
@@ -1,0 +1,31 @@
+#
+# This file is licensed under the Apache License v2.0 with LLVM Exceptions.
+# See https://llvm.org/LICENSE.txt for license information.
+# SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+#
+# (c) Copyright 2025-2026 Advanced Micro Devices, Inc. or its affiliates
+#
+# all_of on an empty use set returns true and the combiner narrows the load, 
+# then re-processes and crashes building G_ZEXT(s20, s20).
+#
+# RUN: llc -mtriple aie2 -run-pass=aie2-prelegalizer-combiner %s -o - | FileCheck %s
+# RUN: llc -mtriple aie2p -run-pass=aie2p-prelegalizer-combiner %s -o - | FileCheck %s
+# RUN: llc -mtriple aie2ps -run-pass=aie2ps-prelegalizer-combiner %s -o - | FileCheck %s
+
+---
+name:            G_LOAD_no_trunc_user
+legalized:       false
+tracksRegLiveness: true
+body:             |
+  bb.0:
+    liveins: $p0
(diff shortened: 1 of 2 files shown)


llvm/test/CodeGen/AIE/GlobalISel/combine-trunc-load-s20-no-narrow.mir
```diff
@@ -0,0 +1,30 @@
...
+    ; CHECK-LABEL: name: G_LOAD_s32_no_trunc_user
+    ; CHECK: [[C:%[0-9]+]]:_(s32) = G_CONSTANT i32 0
+    ; CHECK-NEXT: [[C1:%[0-9]+]]:_(p0) = G_CONSTANT i20 0
+    ; CHECK-NEXT: [[LOAD:%[0-9]+]]:_(s20) = G_LOAD [[C1]](p0) :: (volatile load (s20), align 4)
+    ; CHECK-NEXT: $r0 = COPY [[C]](s32)
+    ; CHECK-NEXT: PseudoRET implicit $lr, implicit $r0
+    %0:_(s32) = G_CONSTANT i32 0
+    %2:_(p0) = G_CONSTANT i20 0
+    %1:_(s32) = G_LOAD %2(p0) :: (volatile load (s32))
+    $r0 = COPY %0(s32)
```
konstantinschwarz: Can you simplify the test? Looks like we only need the volatile G_LOAD?
Wondering why we would even attempt truncating the s32 load to s20, if there is no s20 user?

abnikant: yes, the earlier fix was preventing it from reaching all_of  by returning false. The real issue is when a volatile G_LOAD has no users, all_of on the empty use set returns true, so the combiner narrows the load from s32 to s20 even though there is no user of it.

llvm/test/CodeGen/AIE/GlobalISel/combine-trunc-load-s20-no-narrow.mir
```diff
@@ -0,0 +1,31 @@
...
+legalized:       false
+tracksRegLiveness: true
+body:             |
+  bb.0:
+    liveins: $p0
+    ; CHECK-LABEL: name: G_LOAD_no_trunc_user
+    ; CHECK: liveins: $p0
+    ; CHECK-NEXT: {{  $}}
+    ; CHECK-NEXT: [[COPY:%[0-9]+]]:_(p0) = COPY $p0
+    ; CHECK-NEXT: [[LOAD:%[0-9]+]]:_(s32) = G_LOAD [[COPY]](p0) :: (volatile load (s32))
```
niwinanto, line 26: Not related to your PR, but do you know why we don't dead code eliminate this unused load?

abnikant: because it is marked volatile - G_LOAD [[COPY]](p0) :: (volatile load (s32))

niwinanto (approved):
LGTM!

PR #827, opened 2026-03-06
[MLIR][TOSA] guard against illegal rw with quant types

tr00:
The TOSA dialect allows building with quant types from the Quant dialect, but some rewrites such as abs(abs(x)) --> abs(x) are not semantic-preserving when x is e.g. tensor<...x!quant.uniform<...>>.
Added guards and LIT tests for abs and negate canonicalization.

### DIFF
--- a/mlir/include/mlir/Dialect/Tosa/Utils/QuantUtils.h
+++ b/mlir/include/mlir/Dialect/Tosa/Utils/QuantUtils.h
@@ -4,6 +4,9 @@
 // See https://llvm.org/LICENSE.txt for license information.
 // SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
 //
+// Modifications (c) Copyright 2026 Advanced Micro Devices, Inc. or its
+// affiliates
+//
 //===----------------------------------------------------------------------===//
 //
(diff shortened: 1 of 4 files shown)


mlir/include/mlir/Dialect/Tosa/Utils/QuantUtils.h
```diff
@@ -26,6 +29,18 @@ namespace tosa {
 // Utility functions to support quantization handling in Tosa.
 //===----------------------------------------------------------------------===//
 
+/// Returns true if the type is a quantized tensor.
+inline bool isQuantizedTensor(Type type) {
```
jorickert: Lets move this to the .cpp

mlir/include/mlir/Dialect/Tosa/Utils/QuantUtils.h
```diff
@@ -26,6 +29,18 @@ namespace tosa {
...
+  auto tensorType = mlir::dyn_cast<RankedTensorType>(type);
```
jorickert: cast<ShapedType> or getElementTypeOrSelf

PR #831, opened 2026-03-09
[AIEX] Legalize G_SEXT_INREG from boolean to G_SELECT

andcarminati:
Replace two shifts by a single select, as constants are hoisted anyway. This cleans a bit the MIR of some benchmarks.

### DIFF
--- a/llvm/lib/Target/AIE/AIELegalizerHelper.cpp
+++ b/llvm/lib/Target/AIE/AIELegalizerHelper.cpp
@@ -84,7 +84,6 @@ bool AIELegalizerHelper::pack32BitVector(LegalizerHelper &Helper,
   assert(SourceRegTy.getSizeInBits() == 32 &&
          "cannot pack vectors larger or smaller than 32-bit");
 
-  const LLT S32 = LLT::scalar(32);
   unsigned Offset = 0;
   Register DstCastReg = MRI.createGenericVirtualRegister(S32);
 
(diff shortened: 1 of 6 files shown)


llvm/lib/Target/AIE/AIELegalizerHelper.cpp
```diff
@@ -225,8 +224,7 @@ bool AIELegalizerHelper::legalizeG_BUILD_VECTOR(LegalizerHelper &Helper,
     // vpush takes 32/64-bit operands so we sign extend the input variable. This
     // is required here since we don't have 8 or 16-bit registers.
     if (DstVecEltTy.getSizeInBits() < 32 && EltRegTy.getSizeInBits() != 32) {
-      const Register EltReg32 =
-          MRI.createGenericVirtualRegister(LLT::scalar(32));
+      const Register EltReg32 = MRI.createGenericVirtualRegister(S32);
```
F-Stuckmann, line 227: where does th S32 type come from?

F-Stuckmann: AIELegalizerHelper defines common types.

llvm/lib/Target/AIE/AIELegalizerHelper.cpp
```diff
@@ -482,8 +500,6 @@ bool AIELegalizerHelper::legalize_G_STORE(LegalizerHelper &Helper,
   Register DstReg = StoreI.getOperand(0).getReg();
   LLT DestLLT = MRI.getType(DstReg);
 
-  const LLT S20 = LLT::scalar(20);
```
F-Stuckmann, line 485: I like it

F-Stuckmann (approved):
I like the Type cleanup

PR #833, opened 2026-03-09
Stuckmann.andreu.prologue.improve

F-Stuckmann:
@andcarminati how about these refactors?

### DIFF
--- a/llvm/lib/Target/AIE/AIERegMemEventTracker.cpp
+++ b/llvm/lib/Target/AIE/AIERegMemEventTracker.cpp
@@ -1,4 +1,4 @@
-//===- AIERegMemEventTracker.h - Register event tracker -..------*- C++ -*-===//
+//===- AIERegMemEventTracker.cpp - Register event tracker -------*- C++ -*-===//
 //
 // This file is licensed under the Apache License v2.0 with LLVM Exceptions.
 // See https://llvm.org/LICENSE.txt for license information.
@@ -68,6 +68,16 @@ int AIERegMemEventTracker::minLastMemoryDelay() const {
   return memoryCycleToDelay(TII->getMinLastMemoryCycle());
 }
 
+int AIERegMemEventTracker::memoryDelay(unsigned SchedClass,
+                                       bool IsBackward) const {
+  return IsBackward ? firstMemoryDelay(SchedClass)
+                    : lastMemoryDelay(SchedClass);
+}
+
+int AIERegMemEventTracker::minMemoryDelay(bool IsBackward) const {
+  return IsBackward ? minFirstMemoryDelay() : minLastMemoryDelay();
+}
+
 void AIERegMemEventTracker::updateUseDefMaxCycle(Register Reg, int EventCycle,
                                                  bool IsDef) {
   auto [It, Inserted] =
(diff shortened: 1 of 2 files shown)


andcarminati:
Could you please separate this commit in a different **PR: Fix 6 failing lit tests for macOS compatibility and updated codegen**

llvm/lib/Target/AIE/AIERegMemEventTracker.cpp
```diff
@@ -44,78 +45,80 @@ std::map<MCRegister, int> &AIERegMemEventTracker::getRegToCycleMap(bool IsDef) {
...
-void AIERegMemEventTracker::updateUseDefMaxCycle(Register Reg, int Latency,
-                                                 bool IsDef) {
-  std::map<MCRegister, int> &RegisterToCycle = getRegToCycleMap(IsDef);
-  MCRegister MCReg = Reg.asMCReg();
-  auto EmplacePair = RegisterToCycle.emplace(MCReg, Latency);
-  if (!EmplacePair.second) {
-    auto RegCycle = EmplacePair.first;
-    RegCycle->second = std::max(RegCycle->second, Latency);
-  }
+static int memoryCycleToOffset(std::optional<int> OneBasedMemoryCycle) {
```
andcarminati: I feel confusing the term offset here. Offset is mostly used in the context of address calculation.

llvm/lib/Target/AIE/AIERegMemEventTracker.cpp
```diff
@@ -44,78 +45,80 @@ std::map<MCRegister, int> &AIERegMemEventTracker::getRegToCycleMap(bool IsDef) {
...
-  }
+int AIERegMemEventTracker::getMemCycle(bool IsStore) const {
+  return IsStore ? StoreCycle : LoadCycle;
+}
+
+int AIERegMemEventTracker::getMemoryAccessCycle() const {
+  return std::max(StoreCycle, LoadCycle);
+}
+
+void AIERegMemEventTracker::addPerInstructionMemCycle(int MemCycle,
```
andcarminati, line 104: It is better to keep the ordering here.

llvm/lib/Target/AIE/AIERegMemEventTracker.cpp
```diff
@@ -125,18 +128,29 @@ int AIERegMemEventTracker::getMaxAliasingMemCycle(const MachineInstr &MI,
...
-  if (MemoryCycle > INT_MIN && MemoryCycle < INT_MAX) {
-    // The +1 converts from "last cycle of overlap" to "first safe cycle".
-    // In instruction scheduling.
-    const int MemoryDep = IsBackward ? (MemoryCycle + MIMemoryCycle + 1)
-                                     : (MemoryCycle - MIMemoryCycle + 1);
-    return std::max(CurrentMax, std::max(0, MemoryDep));
-  }
-  return CurrentMax;
+int AIERegMemEventTracker::getMinSafeDistance(int CurrentMax, int StoredCycle,
+                                              int OperandOffset,
```
andcarminati: OperandOffset is not always operand now. Sometimes is memory cycle.

F-Stuckmann: should we simplify it to offset?

llvm/lib/Target/AIE/AIERegMemEventTracker.cpp
```diff
@@ -148,104 +162,159 @@ int AIERegMemEventTracker::checkRegisterDependencies(int CurrentMax,
...
-          const int RegCycleValue = RegCycle->second;
-          // The +1 converts from "last cycle of overlap" to "first safe cycle".
-          // In instruction scheduling.
-          const int ThisOperandLatency = IsBackward
-                                             ? RegCycleValue + OperandCycle + 1
-                                             : RegCycleValue - OperandCycle + 1;
-          CurrMaxLatency = std::max(ThisOperandLatency, CurrMaxLatency);
-        }
+        if (RegCycle != RegToCycle.end())
+          MaxLatency = getMinSafeDistance(MaxLatency, RegCycle->second,
```
andcarminati: The lambda says MaxRegLatency, but we are calculating distances. I guess CurrMaxLatency  is not precise enough in the original code.

llvm/lib/Target/AIE/AIERegMemEventTracker.cpp
```diff
@@ -148,104 +162,159 @@ int AIERegMemEventTracker::checkRegisterDependencies(int CurrentMax,
...
     unsigned OperandCycle = OptCycle ? *OptCycle : 0 /*implicit-def*/;
 
-    auto SafeDistance = [this, OperandCycle,
-                         IsBackward](const MachineOperand &MO, bool IsDef) {
+    auto MaxRegLatency = [this, OperandCycle,
+                          IsBackward](const MachineOperand &MO, bool IsDef) {
       const std::map<MCRegister, int> &RegToCycle =
           IsDef ? RegisterToCycleDef : RegisterToCycleUse;
-      int CurrMaxLatency = 0;
+      int MaxLatency = 0;
```
andcarminati: maybe CurrSafeDistance.

llvm/lib/Target/AIE/AIERegMemEventTracker.cpp
```diff
@@ -148,104 +162,159 @@ int AIERegMemEventTracker::checkRegisterDependencies(int CurrentMax,
...
+  if (!MI.mayLoadOrStore())
+    return CurrentMax;
+
+  const int MIMemoryCycle = [&]() {
+    if (MI.isBundle())
+      return minMemoryOffset(IsBackward);
+    return memoryOffset(MI.getDesc().getSchedClass(), IsBackward);
+  }();
+
+  int MaxLatency = CurrentMax;
```
andcarminati: I start to rethink those names, maybe CurrSafeDistance as well.

andcarminati:
Generally looks good. Some remarks:
* Sometimes we are using Latency where the concept is distance, even in the original code.
* The term offset is confusing, could we replace?
* Can we also run the tests?
* The first commit is better in a different PR.

llvm/lib/Target/AIE/AIERegMemEventTracker.cpp
F-Stuckmann: what ordering?

llvm/lib/Target/AIE/AIERegMemEventTracker.cpp
```diff
@@ -80,29 +100,18 @@ int AIERegMemEventTracker::getFirstMemoryAccessCycle() const {
 }
 
 void AIERegMemEventTracker::updateFirstMemCycle(int Cycle, bool IsStore) {
-  if (IsStore) {
+  if (IsStore)
```
martien-de-jong, line 103: Objection, your honour.

PR #841, opened 2026-03-11
[AIEX] Support ZOL jump to LEND distance

khallouh:
This allows the target to specify a ZOL setup distance in either cycles or a concrete byte distance. Also, this allows the target to opt to keep jump instructions away from ZOL LEND at a particular distance in bytes.


mludevid (approved):
LGTM

PR #844, opened 2026-03-12
[MLIR][Quant] Added default to QuantizationFlags

tr00:
Added QuantizationFlags::None to be more robust than unsigned integer value 0.

### DIFF
--- a/mlir/include/mlir/Dialect/Quant/IR/QuantTypes.h
+++ b/mlir/include/mlir/Dialect/Quant/IR/QuantTypes.h
@@ -3,7 +3,7 @@
 // Part of the LLVM Project, under the Apache License v2.0 with LLVM Exceptions.
 // See https://llvm.org/LICENSE.txt for license information.
 // SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
-// Modifications (c) Copyright 2025 Advanced Micro Devices, Inc. or its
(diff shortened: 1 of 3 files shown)


mlir/include/mlir/Dialect/Quant/IR/QuantTypes.h
jorickert, line 43: IMO the even better approach would be to get rid of this flag completely, its unclear to me why it exists, it seems like it just duplicates information from the storage type

tr00: But what about signless integer types? Then the sign is not encoded in the type.

jorickert: I would prefer to just not use signless integer type as storage type if sign of it is relevant

ilango100: I'm of the same opinion, as it's creating confusion when the storage type is itself signed, but this flag is set to 0. Should we consider storageType or override with this flag?

tr00: Maybe as a first measure, we could add a check to verifyInvariants() in QuantTypes.cpp that ensures storageType and flag agree.

PR #846, opened 2026-03-13
[AIEX] Fix alignment of fall-through jump-table blocks

konstantinschwarz:


llvm/lib/Target/AIE/Utils/AIEMachineBasicBlockUtils.cpp
```diff
@@ -35,7 +36,21 @@ bool isBlockOnlyReachableByFallthrough(const MachineBasicBlock *MBB) {
   if (Pred->empty())
     return isBlockOnlyReachableByFallthrough(Pred);
 
+  //  On AIE, branches are not marked as terminators, as there will be regular
+  //  instructions scheduled in the branch delay slots. Thus, we need to check
+  //  whether a basic block is part of a jump table and conservatively not
+  //  consider it to be only reachable as a fall-through.
+  auto JumpsToBB = [MBB](const MachineJumpTableEntry &JTE) {
+    return find(JTE.MBBs, MBB) != JTE.MBBs.end();
```
mludevid: Nit: `is_contained`

llvm/lib/Target/AIE/Utils/AIEMachineBasicBlockUtils.cpp
```diff
@@ -35,7 +36,21 @@ bool isBlockOnlyReachableByFallthrough(const MachineBasicBlock *MBB) {
   if (Pred->empty())
     return isBlockOnlyReachableByFallthrough(Pred);
 
+  //  On AIE, branches are not marked as terminators, as there will be regular
+  //  instructions scheduled in the branch delay slots. Thus, we need to check
+  //  whether a basic block is part of a jump table and conservatively not
+  //  consider it to be only reachable as a fall-through.
+  auto JumpsToBB = [MBB](const MachineJumpTableEntry &JTE) {
```
mludevid, line 44: I just discovered that we have the same logic in the assembly printer in `isBlockOnlyReachableByFallthrough`. Can we share both functions? In any case the `isBBinJumpTable` could be shared for sure.

llvm/lib/Target/AIE/Utils/AIEMachineBasicBlockUtils.cpp
```diff
@@ -35,7 +36,21 @@ bool isBlockOnlyReachableByFallthrough(const MachineBasicBlock *MBB) {
...
+    return find(JTE.MBBs, MBB) != JTE.MBBs.end();
+  };
+  const MachineJumpTableInfo *JTI = MBB->getParent()->getJumpTableInfo();
+  if (JTI && any_of(JTI->getJumpTables(), JumpsToBB)) {
+    return false;
+  }
+
   // Check the terminators in the previous blocks
+  // NOTE: for targets that use delay slot filling (all except AIE1),
```
mludevid, line 53: Nit: Maybe we can assert this?

llvm/test/CodeGen/AIE/aie2p/align-fallthrough-mbb-jump-table.mir
```diff
@@ -0,0 +1,1520 @@
...
+# (c) Copyright 2026 Advanced Micro Devices, Inc. or its affiliates
+
+# RUN: llc -mtriple=aie2p %s --run-pass=aie-machine-alignment -o - | FileCheck %s
+
+# bb.2 is the layout successor of bb.1, but also part of a jump table.
+# bb.1 ends in an unconditional indirect branch, that happens to jump to bb.2
+# However, bb.2 still needs to be aligned, and cannot be merged with bb.1 for
+# alignment purposes.
+
+---
```
mludevid, line 16: Nit: A more concise test would be nice.

konstantinschwarz: I reduced the number of blocks in the jump table and simplified control flow.

llvm/lib/Target/AIE/Utils/AIEMachineBasicBlockUtils.cpp
konstantinschwarz: The one in AIEBaseAsmPrinter now dispatches to AIEMachineBasicBlockUtils.

I'm still wondering if we need to scan all instructions in our predecessor and look for undecidable indirect branches, since we don't have the `isTerminator` information.
That is independent of the fix in this PR though, and should be a follow-up.

mludevid (approved):
Thanks for addressing the comments. LGTM.

PR #848, opened 2026-03-15
[AIE2PS]  Enable getUnrollingPreferences target hook.

SagarMaheshwari99:


llvm/lib/Target/AIE/aie2ps/AIE2PSTargetTransformInfo.h
```diff
@@ -24,17 +24,27 @@
...
+  bool isVectorExtractIntrinsicID(Intrinsic::ID ID) const override;
+  bool isGetSSIntrinsicID(Intrinsic::ID ID) const override;
+};
+
 class AIE2PSTTIImpl : public AIEBaseTTIImpl<AIE2PSTTIImpl> {
   typedef AIEBaseTTIImpl<AIE2PSTTIImpl> BaseT;
   typedef TargetTransformInfo TTI;
   friend BaseT;
-  AIETTICommon Common;
+  AIE2PSTTICommon Common;
```
F-Stuckmann, line 38: why are we not keeping  `AIETTICommon` ?

F-Stuckmann: ah CRTP artifacts. 
We could merge AIETTICommon and AIEBaseTTIImpl<T> and then have a proper inheritance from BasicTTIImplBase<T> -> AIEBaseTTIImpl<T> -> AIE2TTIImpl

SagarMaheshwari99: yeah that seemed like refactoring for another day.

F-Stuckmann: But that's probably outside of the scope of this PR

llvm/lib/Target/AIE/AIEBaseTargetTransformInfo.cpp
```diff
@@ -239,11 +249,9 @@ bool AIETTICommon::isScalarizedVectorOpIdiomLoop(const Loop *L) const {
   for (auto &I : *LoopBlock) {
     if (isa<LoadInst>(I) || isa<StoreInst>(I))
       return false;
-    // Weak detection for vector loop idioms that we want to unroll
-    // to be able to further apply combiners.
-    // TODO: implement exact detection by looking into IV + insert/extract.
```
F-Stuckmann: nit: i think we should keep the comment, we are still not looking into the IV

llvm/lib/Target/AIE/AIEBaseTargetTransformInfo.cpp
```diff
@@ -229,6 +229,16 @@ bool AIETTICommon::isGetSSExtractInsertVectorSequence(
   return false;
 }
 
+// Detects element-wise vector copy/reshape patterns:
+// insertelement(vec, extractelement(other_vec, idx), idx2)
+// These loops collapse to shufflevector instructions after full unrolling.
+static bool isExtractInsertElementSequence(const Instruction &I) {
```
F-Stuckmann: isn't this check too generic? we would for example also match arithmetic heavy loops and I don't know if we want to unroll then

SagarMaheshwari99: But here 
`Insert->getOperand(1)` is the value being inserted. If it's directly an `ExtractElementInst`, there's zero computation between extract and insert. If there's arithmetic, the operand would be the arithmetic result, not the extract.
for ex.. 
%x = extractelement <8 x i32> %src, i32 %i
%ins = insertelement <16 x i32> %dst, i32 %x, i32 %i
 
this shows direct edge between extract and insert..

but I guess if the pattern itself is inside a large body.. then we might need to reconsider..

F-Stuckmann (approved):
We should consider the IV in the unrolling decision, but it can be a follow-up for this PR

PR #849, opened 2026-03-16
fixup! [AIEX] Support ZOL jump to LEND distance

khallouh:
We don't need to elongate the jump bundle since there we will be at least 4 full delay slot bundles, at least a full bundle in the preheader and at least a full one in the loop.


mludevid (approved):
LGTM.

PR #850, opened 2026-03-16
[AIE2PS] Fix Fp16 mangling.

SagarMaheshwari99:


konstantinschwarz (approved):
_Float16 is an official type and as such uses the correct name mangling.
A better change would probably be to introduce a wrapper struct `float16` that naturally uses to desired name mangling.

PR #851, opened 2026-03-16
[AIE2PS] Add remaining spill support

isoard-amd:
In AIE2PS this bring support for spilling of:
- E/F/G/EG/FEG/EW/FEW
- Their respective 2x and 4x sizes (not that this isn't the optimal version, but we are limited by the current spilling API capabilities)
- PSRFL register (note that this is currently "dirty" due to the FIFO register requiring a bounce register)

In all AIE1/2/2P/2PS:
- Fix offset computation (we respect register description's spill sizes more constently)
- As a side-effect, we no longer need to provide a spill size override in most situations
- The order of the spill instruction still matter (they must match between the store and load, as they affect the layout)


F-Stuckmann:
I like where this PR is going :)

llvm/lib/Target/AIE/aie2ps/AIE2PSInstrInfo.cpp
```diff
@@ -494,6 +526,88 @@ AIE2PSInstrInfo::getSpillPseudoExpandInfo(const TargetRegisterInfo &TRI,
...
+    return {{AIE2PS::ST_dml_sts_scalar_spill, AIE2PS::NoSubRegister, 8}};
+  case AIE2PS::VST_EG_SPILL:
+    return {{AIE2PS::ST_dml_sts_scalar_spill, AIE2PS::NoSubRegister, 8}};
+  case AIE2PS::VST_EG2_SPILL:
+    return {{AIE2PS::VST_128_dmv_sts_eg_spill, AIE2PS::NoSubRegister, 16}};
+  case AIE2PS::VST_FEG_SPILL:
+    return {{AIE2PS::VST_128_dmv_sts_feg_spill, AIE2PS::NoSubRegister, 16}};
+  case AIE2PS::VST_FEG2_SPILL:
+    return {{AIE2PS::VST_dmw_sts_feg2_spill, AIE2PS::NoSubRegister, 32}};
+  case AIE2PS::VST_EW_SPILL: // FIXME: Use VST_EG_SPILL
```
F-Stuckmann, line 540: why are we not using VST_EG_SPILL ?

llvm/lib/Target/AIE/aie2ps/AIE2PSInstrInfo.cpp
```diff
@@ -800,6 +800,67 @@ void AIE2PSInstrInfo::storeRegToStackSlot(MachineBasicBlock &MBB,
...
     return bounceViaRegClass(&AIE2PS::VEC1024RegClass);
+  } else if (regClassMatches(AIE2PS::ePSRFLdFRegClass, RC, SrcReg)) {
+    // For physical registers, resolve sub-register indices to actual registers.
+    // For virtual registers, use sub-register index directly.
+    MachineRegisterInfo &MRI = MBB.getParent()->getRegInfo();
+    MachineFunction &MF = *MBB.getParent();
+    MachineFrameInfo &MFI = MF.getFrameInfo();
+
+    // Helper to create store MMO with proper offset and size for a sub-register
+    unsigned ByteOffset = 0;
```
F-Stuckmann: This code seems to be duplicated. Could you refactor it into a helper function? Then you could also call it from line 974

llvm/lib/Target/AIE/aie2ps/AIE2PSRegisterInfo.td
```diff
@@ -1209,13 +1209,13 @@ def spill_eDC_to_eR : AIE2PSScalarRegisterClass<(add eDC, eR)>;
 
 class AIE2PVector1076FifoRegisterClass<dag reglist> :
     AIE2PSRegisterClass<1088, 512, [i32], reglist>;
-def sub_ptr : SubRegIndex<20,  0>;
-def sub_fifo : SubRegIndex<1024, 20>;
-def sub_avail : SubRegIndex<32, 1044>;
+def sub_fifo : SubRegIndex<1024, 0>;
+def sub_avail : SubRegIndex<32, 1024>;
```
F-Stuckmann, line 1213: CHECK: why are we flipping the ordering here?

llvm/lib/Target/AIE/aie1/AIE1RegisterInfo.td
```diff
@@ -284,9 +284,19 @@ def VEC512 : AIEVector512RegisterClass<(add xa, xb, xc, xd)>;
 def VEC1024 : AIEVector1024RegisterClass<(add ya, yd)>;
 
 class AIEVector384RegisterClass<dag reglist> :
-    AIERegisterClass<384, [v8i48], reglist>;
+    AIEBaseRegisterClass<"AIE", [v8i48], /*alignment*/384, reglist> {
+  dag Regs = reglist;
+  let RegInfos = RegInfoByHwMode<
+      [DefaultMode],
+      [RegInfo</*size*/384, /*spill size*/512, /*spill alignment*/256>]>;
```
F-Stuckmann, line 291: nit: We can drop aie1 support

llvm/lib/Target/AIE/aie2ps/AIE2PSInstrInfo.cpp
isoard-amd: Because there is no EG subregister in EW, it is split into two sub-registers (E and G), and as such, the current API cannot handle it. If we want to improve this, the API need to take a list of sub-register (E and G), and recombine them into a super-register (a single EG register), and then use the op on it.

llvm/lib/Target/AIE/aie2ps/AIE2PSRegisterInfo.td
isoard-amd: I tried to align it with the way we lay it out in memory. But technically we could keep the old order in the .td file (as we can reorder the layout without changing it in the .td description).

As for why this order is better, the main reason is that we want the spill layout to be in decreasing order of alignment constraint.

llvm/lib/Target/AIE/aie2p/AIE2PInstrInfo.cpp
```diff
@@ -1062,7 +1062,7 @@ SmallVector<AIEBaseInstrInfo::AIEPseudoExpandInfo, 4>
 AIE2PInstrInfo::getSpillPseudoExpandInfoByOpcode(unsigned Opcode) const {
   switch (Opcode) {
   case AIE2P::ST_R_SPILL:
-    return {{AIE2P::ST_dms_sts_spill, AIE2P::NoSubRegister, 4}};
+    return {{AIE2P::ST_dms_sts_spill}};
```
F-Stuckmann, line 1072: nice cleanup :)

llvm/lib/Target/AIE/AIEBaseInstrInfo.cpp
```diff
@@ -612,15 +613,16 @@ void AIEBaseInstrInfo::expandSpillPseudo(
...
-    const int BitsInByte = 8;
-
     unsigned MemorySize = 0;
     if (EI.MemSize) {
       MemorySize = EI.MemSize;
     } else {
-      assert(EI.SubRegIndex);
-      MemorySize = divideCeil(TRI.getSubRegIdxSize(EI.SubRegIndex), BitsInByte);
-      MemorySize = alignTo(MemorySize, SubRegOffsetAlign);
+      const TargetRegisterClass *RC = TRI.getMinimalPhysRegClass(Reg);
```
F-Stuckmann, line 620: nit: this could have been in a Constructor in `AIEPseudoExpandInfo`

llvm/lib/Target/AIE/aie2ps/AIE2PSRegisterInfo.td
F-Stuckmann: That would be pretty bad if the alignment is the same across all the subregs, nice work!

llvm/lib/Target/AIE/aie1/AIE1RegisterInfo.td
isoard-amd: I would rather keep this. Else we have to mark the related unit test as failing.

F-Stuckmann (dismissed):
LGTM. 
Let's worry about the minor cleanups after merging this PR

F-Stuckmann (dismissed):

llvm/lib/Target/AIE/aie2ps/AIE2PSRegisterInfo.cpp
```diff
@@ -125,7 +125,9 @@ bool AIE2PSRegisterInfo::eliminateFrameIndex(MachineBasicBlock::iterator II,
   // Select the base pointer (BP) and calculate the actual offset from BP
   // to the beginning of the object at index FI.
   int Offset = TFI->getFrameIndexReference(MF, FrameIndex, FrameReg).getFixed();
-
+  int64_t BaseOffset = 0;
+  if (!MI.memoperands_empty())
```
konstantinschwarz, line 129: Do you have a test case that showcases the wrong offset computation?

isoard-amd: I added a test (before the change, then updated it after the fix)

llvm/lib/Target/AIE/aie2ps/AIE2PSInstrInfo.cpp
isoard-amd: Thanks for doing this for me.

konstantinschwarz (approved):
LGTM, thanks for adding the test case!

PR #852, opened 2026-03-16
[AIEX] Combine repeating G_BUILD_VECTOR pattern into G_AIE_BROADCAST_VECTOR

konstantinschwarz:
Match a G_BUILD_VECTOR whose operands form a repeating sub-pattern of length Period (e.g., [a, b, a, b, ...] for Period=2). The Period elements are packed into a merged scalar and then broadcast via G_AIE_BROADCAST_VECTOR to avoid full scalarization. Only matches vectors >= 128 bits. Skips pure splats (Period=1).


llvm/lib/Target/AIE/AIECombinerHelper.cpp
```diff
@@ -5002,3 +5002,153 @@ bool llvm::matchVSelToUnmergeConcatOrCopy(MachineInstr &MI,
...
+/// ElemBits * Period == MergedBits must hold.
+/// Adding a new pattern (e.g., {32, 2, 64}) only requires an entry here;
+struct AlternatingBVPattern {
+  unsigned ElemBits;   // Source element size in bits.
+  unsigned Period;     // Number of consecutive elements that form one unit.
+  unsigned MergedBits; // Merged scalar size in bits (= ElemBits * Period).
+};
+
+/// Table of currently enabled merge patterns.
+/// Ordered by (ElemBits, Period) so that smaller periods are tried first.
```
martien-de-jong: Perhaps add 'for the same element size. Now I look at the table and see a Period 4 sitting between two Period 2 ones.

llvm/lib/Target/AIE/AIECombinerHelper.cpp
```diff
@@ -5002,3 +5002,153 @@ bool llvm::matchVSelToUnmergeConcatOrCopy(MachineInstr &MI,
...
+  const LLT DstVecTy = MRI.getType(DstVecReg);
+  const unsigned NumElts = DstVecTy.getNumElements();
+
+  // The number of elements must be a multiple of the period.
+  if (NumElts % Pat.Period != 0)
+    return false;
+
+  // Collect the period elements (operands 1..Period in the G_BUILD_VECTOR).
+  SmallVector<Register, 4> PeriodElts;
+  PeriodElts.reserve(Pat.Period);
```
martien-de-jong: nit: with the current table, this is a noop. Morever, it needs to fit a scalar register, so I don't think it will change soon.

llvm/lib/Target/AIE/AIECombinerHelper.cpp
```diff
@@ -5002,3 +5002,153 @@ bool llvm::matchVSelToUnmergeConcatOrCopy(MachineInstr &MI,
...
+
+//===----------------------------------------------------------------------===//
+// combine_alternating_build_vector
+//===----------------------------------------------------------------------===//
+
+namespace {
+
+/// Describes a supported merge pattern for G_BUILD_VECTOR operands.
+/// ElemBits * Period == MergedBits must hold.
+/// Adding a new pattern (e.g., {32, 2, 64}) only requires an entry here;
```
F-Stuckmann: nit: we already have a sawtooth pattern matching for vshuffles. We could reuse it

llvm/lib/Target/AIE/AIECombinerHelper.cpp
```diff
@@ -5002,3 +5002,153 @@ bool llvm::matchVSelToUnmergeConcatOrCopy(MachineInstr &MI,
...
+  SmallVector<Register, 4> PeriodElts;
+  PeriodElts.reserve(Pat.Period);
+  for (unsigned I = 0; I < Pat.Period; ++I)
+    PeriodElts.push_back(MI.getOperand(1 + I).getReg());
+
+  // Verify that the full operand list is a repetition of PeriodElts.
+  for (unsigned I = Pat.Period; I < NumElts; ++I) {
+    if (MI.getOperand(1 + I).getReg() != PeriodElts[I % Pat.Period])
+      return false;
+  }
```
martien-de-jong: nit: I think you can establish a period while checking, going from 1 to 2 to 4 at the appropriate period mismatches.  The other match parameters follow from checking Period * elemsize against 32 bits. A splat can be recognized by finding period==1

llvm/lib/Target/AIE/AIECombinerHelper.cpp
```diff
@@ -5002,3 +5002,153 @@ bool llvm::matchVSelToUnmergeConcatOrCopy(MachineInstr &MI,
...
+bool llvm::matchAlternatingBuildVector(
+    MachineInstr &MI, MachineRegisterInfo &MRI, const AIEBaseInstrInfo &TII,
+    AIEAlternatingBuildVectorMatchData &MatchInfo) {
+  assert(MI.getOpcode() == TargetOpcode::G_BUILD_VECTOR &&
+         "Expected a G_BUILD_VECTOR");
+
+  const Register DstVecReg = MI.getOperand(0).getReg();
+  const LLT DstVecTy = MRI.getType(DstVecReg);
+
+  // Only handle vectors >= 128 bits. Smaller vectors are kept in GPRs.
```
martien-de-jong, line 5019: Could still be a copy instead of a 32 bit immediate. Or can that be left to machine cse?

llvm/lib/Target/AIE/AIECombinerHelper.cpp
```diff
@@ -5002,3 +5002,153 @@ bool llvm::matchVSelToUnmergeConcatOrCopy(MachineInstr &MI,
...
+
+  const Register DstVecReg = MatchInfo.DstVecReg;
+  const LLT DstVecTy = MRI.getType(DstVecReg);
+  const unsigned NumElts = DstVecTy.getNumElements();
+  const unsigned ElemBits = DstVecTy.getElementType().getSizeInBits();
+  const unsigned Period = MatchInfo.PeriodElts.size();
+  const unsigned MergedBits = MatchInfo.MergedBits;
+
+  // Step 1: Pack the Period elements into a merged scalar of MergedBits.
+  //   packed = elem[0] | (elem[1] << ElemBits) | (elem[2] << 2*ElemBits) ...
```
martien-de-jong: Horner's scheme? (i.e. (((elem[3] << ElemBits) | elem[2]) << ElemBits)  | elem[1]) << ElemBits) | elem[0])
This reuses the shift constant

llvm/lib/Target/AIE/AIECombinerHelper.cpp
```diff
@@ -5002,3 +5002,153 @@ bool llvm::matchVSelToUnmergeConcatOrCopy(MachineInstr &MI,
...
+
+  Register PackedReg =
+      B.buildZExt(MergedScalarTy, MatchInfo.PeriodElts[0]).getReg(0);
+  for (unsigned I = 1; I < Period; ++I) {
+    Register EltExtReg =
+        B.buildZExt(MergedScalarTy, MatchInfo.PeriodElts[I]).getReg(0);
+    Register ShiftAmt = B.buildConstant(MergedScalarTy, I * ElemBits).getReg(0);
+    Register ShiftedReg =
+        B.buildShl(MergedScalarTy, EltExtReg, ShiftAmt).getReg(0);
+    PackedReg = B.buildOr(MergedScalarTy, PackedReg, ShiftedReg).getReg(0);
```
martien-de-jong: One of the elements may be flush with the signbit after shifting, so doesn't need zext.

llvm/lib/Target/AIE/AIECombinerHelper.cpp
konstantinschwarz: We can leave this to the legalizer, it turns small G_BUILD_VECTOR into bit arithmetic

llvm/lib/Target/AIE/AIECombinerHelper.cpp
konstantinschwarz: Done. Unfortunately there are other GlobalISel combines that optimize `(x << 8) << 8 -> x << 16` so we still end up with multiple shift constants

konstantinschwarz:
Thanks for the review, I think I applied all the comments

F-Stuckmann (approved):
Nice

martien-de-jong (approved):
Nice!. I think the optimizers can only combine the shift amounts if we have a constant zero element in the period, in which the extra constant load is cost-neutral

PR #853, opened 2026-03-17
[NFC][AIEX] Use explicit case labels instead of default in fifo_st switch

mludevid:


martien-de-jong (approved):
Beyond a shadow of doubt.

PR #854, opened 2026-03-17
Remove loops from aie2ps upd headers

F-Stuckmann:
Revert Loop Unrolling whitelist in favor of removing the loops in the aie2ps headers.

### DIFF
--- a/llvm/lib/Target/AIE/AIEBaseTargetTransformInfo.cpp
+++ b/llvm/lib/Target/AIE/AIEBaseTargetTransformInfo.cpp
@@ -185,12 +185,9 @@ void AIETTICommon::adjustUnrollingPreferences(Loop *L, ScalarEvolution &SE,
   }
 }
 
-static bool isVectorReshuffleLoop(const Loop *L);
-
 void AIETTICommon::applyLoopIdiomUnrolling(
     Loop *L, TTI::UnrollingPreferences &UP) const {
-  if (isScalarizedVectorOpIdiomLoop(L) || isScalarLoop(L) ||
-      isVectorReshuffleLoop(L)) {
+  if (isScalarizedVectorOpIdiomLoop(L) || isScalarLoop(L)) {
     UP.Threshold = LoopIdiomUnrollingThreshold;
   } else if (UnrollOnlyLoopsWithPragma && !AIELoopUtils::hasUnrollPragma(L)) {
     UP.Threshold = 0;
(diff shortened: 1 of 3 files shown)


llvm/lib/Target/AIE/AIEBaseTargetTransformInfo.cpp
```diff
@@ -252,48 +254,94 @@ bool AIETTICommon::isScalarizedVectorOpIdiomLoop(const Loop *L) const {
   return IsVectorLoopIdiom;
 }
 
+// Return true when V will become a compile-time constant once the loop
+// is fully unrolled. That is the case when V is either:
+//   - loop-invariant (already constant across iterations), or
+//   - an affine recurrence {start,+,step} in L (each unrolled iteration
```
SagarMaheshwari99: Do we need to restrict to affine recurrences? The add recurrence could be of any degree no?

llvm/lib/Target/AIE/AIEBaseTargetTransformInfo.cpp
```diff
@@ -252,48 +254,94 @@ bool AIETTICommon::isScalarizedVectorOpIdiomLoop(const Loop *L) const {
...
+// vector) are SCEVUnknown and correctly rejected.
+static bool isConstantAfterUnroll(const Value *V, const Loop *L,
+                                  ScalarEvolution &SE) {
+  const SCEV *S = SE.getSCEV(const_cast<Value *>(V));
+  if (isa<SCEVCouldNotCompute>(S))
+    return false;
+  if (SE.isLoopInvariant(S, L))
+    return true;
+  const auto *AR = dyn_cast<SCEVAddRecExpr>(S);
+  return AR && AR->isAffine() && AR->getLoop() == L;
```
SagarMaheshwari99: to make it broad : remove `AR->isAffine()` check.

llvm/lib/Target/AIE/AIEBaseTargetTransformInfo.cpp
```diff
@@ -252,48 +254,94 @@ bool AIETTICommon::isScalarizedVectorOpIdiomLoop(const Loop *L) const {
...
+}
+
+/// Return true when \p Ins inserts a value that was extracted from
+/// another vector, possibly through a bitcast.
+static bool isExtractInsertPair(const InsertElementInst *Ins) {
+  const Value *Val = Ins->getOperand(1);
+  if (const auto *BC = dyn_cast<BitCastInst>(Val))
```
SagarMaheshwari99: this peels exactly one bitcast.. it might be rare.. but if we have consecutive bitcasts.. the other ones would still remain. 

Maybe we should replace `if` with `while`.

llvm/lib/Target/AIE/AIEBaseTargetTransformInfo.cpp
```diff
@@ -252,48 +254,94 @@ bool AIETTICommon::isScalarizedVectorOpIdiomLoop(const Loop *L) const {
...
+  return isa<ExtractElementInst>(Val);
+}
+
+/// Return true when the instruction is zero-cost integer scaffolding:
+/// index arithmetic, comparisons, or casts that disappear after
+/// constant-folding the unrolled iterations.
+static bool isIntegerScaffolding(const Instruction &I) {
+  if (isa<ICmpInst>(I))
+    return true;
+  if (isa<BitCastInst>(I))
```
SagarMaheshwari99: We can get rid of this bitcast.. since we have a more tighter condition on line 295.

llvm/lib/Target/AIE/AIEBaseTargetTransformInfo.cpp
```diff
@@ -252,48 +254,94 @@ bool AIETTICommon::isScalarizedVectorOpIdiomLoop(const Loop *L) const {
...
       continue;
     }
-    if (isa<BinaryOperator>(I) && I.getType()->isIntegerTy())
-      continue;
-    if (isa<ICmpInst>(I))
-      continue;
-    if (isa<BitCastInst>(I))
-      continue;
-    if (isa<CastInst>(I) && I.getType()->isIntegerTy())
+
```
SagarMaheshwari99: Can we also add a scalar select?

llvm/test/CodeGen/AIE/opt/unroll_reshuffle_counterexample.ll
```diff
@@ -0,0 +1,72 @@
+; NOTE: Assertions have been autogenerated by utils/update_test_checks.py UTC_ARGS: --version 5
+
+; This file is licensed under the Apache License v2.0 with LLVM Exceptions.
+; See https://llvm.org/LICENSE.txt for license information.
+; SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+;
+; (c) Copyright 2026 Advanced Micro Devices, Inc. or its affiliates
+
+; RUN: opt -S -mtriple=aie2p -passes=loop-unroll %s -o - | FileCheck %s
+
```
SagarMaheshwari99: also for aie2ps..

llvm/lib/Target/AIE/AIEBaseTargetTransformInfo.cpp
```diff
@@ -252,48 +254,94 @@ bool AIETTICommon::isScalarizedVectorOpIdiomLoop(const Loop *L) const {
...
+    // Integer scaffolding (index math, comparisons, casts): allowed.
+    if (isIntegerScaffolding(I))
```
martien-de-jong: I would maintain a counter. My hunch is that this is just to accommodate the update and test of the loop counter.

F-Stuckmann:
We should not update our unrolling decisions but instead fix the headers that provide the loops instead of extracts/concats.
I reverted the unrolling logic update and fixed the headers instead.

SagarMaheshwari99:
Maybe remove NFC from the second commit message? Since semantically they may be equivalent.. but functionally there is still a change... (in the test).

SagarMaheshwari99 (approved):

PR #855, opened 2026-03-18
[AIE2PS][NFC] Fix upd_ext headers

F-Stuckmann:
Add missing breaks in switch statements.

As a nice side effect we also remove unnecessary branches from the IR :)


abnikant (approved):
LGTM!

PR #857, opened 2026-03-18
[AIEX][NFC] Add Verifier for Setup and Jump distance constraints to LoopEnd

khallouh:
Our verifier before this only checked region alignments. This checks also other byte distance constraints handled by the machine alignment pass.

### DIFF
--- a/llvm/lib/Target/AIE/AIEMachineAlignment.cpp
+++ b/llvm/lib/Target/AIE/AIEMachineAlignment.cpp
@@ -11,9 +11,12 @@
 #include "AIEMachineAlignment.h"
 #include "AIE.h"
 #include "AIEBundle.h"
+#include "Utils/AIELoopUtils.h"
 #include "Utils/AIEMachineBasicBlockUtils.h"
+#include "llvm/ADT/DenseMap.h"
 #include "llvm/CodeGen/MachineBasicBlock.h"
 #include "llvm/CodeGen/MachineInstrBuilder.h"
+#include "llvm/Support/Debug.h"
 
 using namespace llvm;
 using namespace AIEMachineBasicBlockUtils;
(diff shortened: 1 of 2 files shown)


llvm/lib/Target/AIE/AIEMachineAlignment.cpp
```diff
@@ -247,6 +250,115 @@ void verifyAlignment(MachineFunction &MF) {
...
+  for (MachineBasicBlock &MBB : MF) {
+    if (!TII.isZOLBody(MBB))
+      continue;
+    MachineBasicBlock *Preheader =
+        AIELoopUtils::getDedicatedFallThroughPreheader(MBB);
+    if (!Preheader)
+      continue;
+
+    auto LastInstr = MBB.getLastNonDebugInstr();
+    if (LastInstr == MBB.end())
```
katerynamuts, line 285: nit: I think this check is redundant in the case of ZOLs because `isZOLBody` already verifies that the last non-debug instruction is a `LEND`

llvm/lib/Target/AIE/AIEMachineAlignment.cpp
```diff
@@ -247,6 +250,115 @@ void verifyAlignment(MachineFunction &MF) {
...
+          continue;
+        unsigned BytesToLastJumpStart = 0;
+        bool FoundJump = false;
+        unsigned CurrentOffset = 0;
+        for (auto MI = OtherMBB.begin(), E = OtherMBB.end(); MI != E; ++MI) {
+          if (MI->isBundle() && TII.isJumpBundle(MI)) {
+            unsigned JumpPC = MBBStartOffset[&OtherMBB] + CurrentOffset;
+            if (JumpPC < LoopEndPC) {
+              FoundJump = true;
+              BytesToLastJumpStart = CurrentOffset;
```
katerynamuts, line 341: nit: maybe call it `BytesToClosestJumpStart` instead of `BytesToLastJumpStart` to make it more clear.

katerynamuts (approved):
LGTM

PR #859, opened 2026-03-19
Cleanup and remove unnecessary header files from our toolchain distribution

konstantinschwarz:


clang/cmake/caches/Peano-AIE.cmake
```diff
@@ -49,7 +49,11 @@ set(LLVM_TOOLCHAIN_TOOLS
 set(_llvm_distribution_components
   clang
   lld
-  clang-resource-headers
+  core-resource-headers
+  aiev2-resource-headers
```
mludevid, line 53: Q: Did you intend to not install the AIE1 headers by default?

mludevid (approved):
LGTM.

clang/cmake/caches/Peano-AIE.cmake
konstantinschwarz: Yes, we weren't installing the aie1 builtins and runtimes before, so I left out the headers as well.
The libraries are still enabled in Peano-AIE-runtime-libraries.cmake, so we get build coverage, just not installing them into the distribution

PR #860, opened 2026-03-20
[AIEX] Fix RegionEndEdges DAG mutator to lower hwloop setup latency and padd the setup instruction instead 

khallouh:
I ran full mllib suite on aie2p, PM size improvements across the board up to 80 bytes per kernel and some Compute Cycle Count impovements up to 1.9% on PADD2D and PADD3D.
image

image

### DIFF
--- a/llvm/lib/Target/AIE/AIEBaseInstrInfo.cpp
+++ b/llvm/lib/Target/AIE/AIEBaseInstrInfo.cpp
@@ -431,8 +431,12 @@ AIEBaseInstrInfo::getPostZOLRegionSizeInfo(MachineBasicBlock &MBB) const {
     if (MI.isDebugInstr())
       continue;
 
-    if (isZOLSetupBundle(&MI) && isLastZOLSetupBundleInMBB(&MI))
+    // We iterate backwards so this is the last setup bundle in the block.
+    if (isZOLSetupBundle(&MI)) {
+      BundleCount++;
+      Size += getAIEMachineBundleSize(MI);
       break;
+    }
 
     if (MI.isBundle())
       BundleCount++;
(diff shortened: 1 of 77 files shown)


llvm/lib/Target/AIE/AIEBaseInstrInfo.cpp
```diff
@@ -431,8 +431,11 @@ AIEBaseInstrInfo::getPostZOLRegionSizeInfo(MachineBasicBlock &MBB) const {
     if (MI.isDebugInstr())
       continue;
 
-    if (isZOLSetupBundle(&MI) && isLastZOLSetupBundleInMBB(&MI))
+    if (isZOLSetupBundle(&MI) && isLastZOLSetupBundleInMBB(&MI)) {
```
martien-de-jong: This looks silly. We iterate backwards. The first ZOLSetupBundle that we see is the last in the block.

llvm/lib/Target/AIE/AIEBaseInstrInfo.cpp
```diff
@@ -431,8 +431,11 @@ AIEBaseInstrInfo::getPostZOLRegionSizeInfo(MachineBasicBlock &MBB) const {
...
+      BundleCount++;
+      Size += getAIEMachineBundleSize(MI);
```
martien-de-jong, line 437: This interprets the entire bundle in which the setup instruction sits as contributing to bytes 'after' it. That sounds strange to me, and I hope we have evidence that it works for any alignment and any size of that bundle.

khallouh:
Note: to clear any confusion, the distance (in all targets including AIE2PS) is to be ensured to the end_of_loop which in the core model refers to the last instruction in the loop, which is not the PseudoLoopEnd in Peano which comes after.

llvm/lib/Target/AIE/AIEBaseInstrInfo.cpp
khallouh: I see but the hardware doesn't care where the actual setup sits inside the bundle. The evidence is a software stall rule in the architecture description, which I'll share privately.

llvm/lib/Target/AIE/AIEBaseInstrInfo.cpp
khallouh: Right, I was also confused when I saw this but forgot to clean it up. I'll update.

martien-de-jong (approved):
LGTM

PR #861, opened 2026-03-20
[AIEX][NFC] clarify MinTripCount==1 overload

F-Stuckmann:
Allow HW loops when MinTripCount ==1.

### DIFF
--- a/llvm/lib/Target/AIE/AIEBaseTargetTransformInfo.cpp
+++ b/llvm/lib/Target/AIE/AIEBaseTargetTransformInfo.cpp
@@ -35,6 +35,12 @@ static cl::opt<bool>
                 cl::desc("Enable true zero overhead hardware loops on AIE"),
                 cl::init(true), cl::Hidden);
 
+/// We currently override MinIterCountHLReject == 1 to
+/// a) remove loop guards and
+/// b) disable HW loop generation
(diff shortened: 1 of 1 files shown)


konstantinschwarz (changes requested):
I agree with this change.
However, we used the Min Tripcount == 1 as a poor hack to disable hardware loop generation. This was necessary to control code bloat in very size constrained loops.

We should have a separate pragma to disable hardware loops. Until then, we should not merge this change. Setting to Request changes to avoid accidentally merging this PR

F-Stuckmann:
I changed the PR to only add a comment explaining our `MinIterCountHLReject = 1` choice

andcarminati (approved):
LGTM.

llvm/lib/Target/AIE/AIEBaseTargetTransformInfo.cpp
```diff
@@ -35,6 +35,12 @@ static cl::opt<bool>
                 cl::desc("Enable true zero overhead hardware loops on AIE"),
                 cl::init(true), cl::Hidden);
 
+/// We currently override MinIterCountHLReject == 1 to
```
martien-de-jong, line 38: nit: overload?

konstantinschwarz (approved):

PR #871, opened 2026-03-26
[AIEX] fix reallocation bug in WAW Rewriter

F-Stuckmann:
Indentation bug. Clarified with early continue.

### DIFF
--- a/llvm/lib/Target/AIE/AIEWawRegRewriter.cpp
+++ b/llvm/lib/Target/AIE/AIEWawRegRewriter.cpp
@@ -373,9 +373,11 @@ bool AIEWawRegRewriter::reAllocate(OriginalAllocation &Candidates,
   BitVector UsedUnits(TRI->getNumRegUnits());
   for (auto &[MO, Org] : Candidates) {
     auto VReg = MO->getReg();
-    if (!replaceReg(VReg, Registers, UsedUnits))
-      LLVM_DEBUG(dbgs() << "Renaming " << printReg(VReg, TRI, 0, MRI)
-                        << " failed\n");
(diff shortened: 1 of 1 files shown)


llvm/lib/Target/AIE/AIEWawRegRewriter.cpp
```diff
@@ -362,9 +362,11 @@ bool AIEWawRegRewriter::reAllocate(OriginalAllocation &Candidates,
   BitVector UsedUnits(TRI->getNumRegUnits());
   for (auto &[MO, Org] : Candidates) {
     auto VReg = MO->getReg();
-    if (!replaceReg(VReg, Registers, UsedUnits))
-      LLVM_DEBUG(dbgs() << "Renaming " << printReg(VReg, TRI, 0, MRI)
-                        << " failed\n");
+    if (replaceReg(VReg, Registers, UsedUnits))
+      continue;
```
martien-de-jong, line 377: Since no tests change, I think we don't test this path.

F-Stuckmann: yeah that is concerning! lets see if I can llvm-reduce a unit-test

konstantinschwarz:
Were you able to find a test case for this bug? In any case, I think we should get this merged

F-Stuckmann:
I was not able to generate a test case that executes the code path.

llvm/lib/Target/AIE/AIEWawRegRewriter.cpp
F-Stuckmann: the codepath is obfuscated and i wasn't able to find a unit-test.

andcarminati (approved):

PR #875, opened 2026-03-26
[AIE1] Fix libc: Pass -mno-vitis-headers to try_compile (#1801)

F-Stuckmann:
somehow we missed aie1 in the runtime patch

### DIFF
--- a/libc/startup/baremetal/aie/crt1.cc
+++ b/libc/startup/baremetal/aie/crt1.cc
@@ -3,7 +3,12 @@
 // See https://llvm.org/LICENSE.txt for license information.
 // SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
 //
-// (c) Copyright 2024 Advanced Micro Devices, Inc. or its affiliates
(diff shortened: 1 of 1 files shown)


andcarminati (approved):

libc/startup/baremetal/aie/crt1.cc
```diff
@@ -3,7 +3,12 @@
 // See https://llvm.org/LICENSE.txt for license information.
 // SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
 //
-// (c) Copyright 2024 Advanced Micro Devices, Inc. or its affiliates
+// (c) Copyright 2024-2026 Advanced Micro Devices, Inc. or its affiliates
+#if __AIEARCH__ == 10
+#include "aiev1intrin.h"
```
martien-de-jong, line 8: discussed offline. We may want to have an aieintrin.h  that switches out into different archs.

martien-de-jong (approved):

PR #876, opened 2026-03-26
[AIEX] Remove KILL and IMPLICIT_DEF before post-RA scheduling

mludevid:
Meta instructions like `KILL` and `IMPLICIT_DEF` confuse the dependence analysis in `ScheduleDAGInstrs::buildEdges`. When a meta instruction defines a register, it absorbs the entry in the Defs/Uses maps, replacing the real instruction that was previously tracked. This breaks direct dependency edges between real instructions.

For example, given:

```
$r0 = MOV ...          (A)
$r0 = KILL $r0         (B)
$r0 = ADD ...          (C)
```

`buildEdges` walks backward and creates edges A→B and B→C, but never a direct A→C output dependence. Since B is a zero-latency meta instruction that emits no code, the 0+0=0 transitive latency allows A and C to be placed in the same VLIW bundle. On AIE's exposed pipeline this is a correctness bug — two real writes to the same register cannot coexist in the same cycle.

Fix: Erase all `KILL` and `IMPLICIT_DEF` instructions at the start of the post-RA scheduling pipeline in `AIEPostRASchedStrategy::enterFunction`. No downstream pass depends on these instructions being present — the `AsmPrinter` only emits debug comments for them, `AIEFinalizeBundle` already skips meta instructions, and liveness is invalidated after scheduling.


llvm/test/CodeGen/AIE/aie2/set.ll
```diff
@@ -107,17 +107,23 @@ define dso_local noundef <128 x i8> @_Z27test_set_v256uint4_1024_256iDv32_DU8_(i
...
-; CHECK-NEXT:    nopx // Delay Slot 5
+; CHECK-NEXT:    vlda wh5, [sp, #-64]; nopb ; nopxm // 32-byte Folded Reload
+; CHECK-NEXT:    nop
+; CHECK-NEXT:    j #.LBB2_6
+; CHECK-NEXT:    nop // Delay Slot 5
 ; CHECK-NEXT:    nop // Delay Slot 4
 ; CHECK-NEXT:    nop // Delay Slot 3
-; CHECK-NEXT:    vlda wh5, [sp, #-64] // 32-byte Folded Reload Delay Slot 2
+; CHECK-NEXT:    nop // Delay Slot 2
 ; CHECK-NEXT:    vmov wh4, wh5 // Delay Slot 1
```
mludevid, line 115: Notice that this was previously incorrectly compiled due to a missing data dependency for `wh5` between the load and the move.

llvm/test/CodeGen/AIE/aie2/schedule/meta_instrs.mir
konstantinschwarz: Looks like most of this test is obsolete now?

konstantinschwarz (commented):
PR looks good to me. @martien-de-jong any comments before merging?

llvm/test/CodeGen/AIE/aie2/schedule/meta_instrs.mir
mludevid: Yes it is. I initially removed the entire test file, but I decided against that to show that the instructions are being removed. I could simplify the test file if you'd like that.

konstantinschwarz: I think updating some of the comments would make sense. They specifically mention the need for NOPs between those meta instructions, but instead the meta instructions (including the NOPs) are eliminated

llvm/lib/Target/AIE/AIEMachineScheduler.cpp
```diff
@@ -724,6 +724,26 @@ void AIEPostRASchedStrategy::schedNode(SUnit *SU, bool IsTopNode) {
...
+  // output dependence. Since these are zero-latency meta instructions that
+  // emit no code, the transitive chain allows A and C to be placed in the
+  // same VLIW bundle, which is incorrect. IMPLICIT_DEF has the same problem
+  // when its def sits between two real writers of the same register.
+  // Removing KILL before VLIW scheduling is the same approach used by the
+  // Hexagon packetizer (HexagonVLIWPacketizer.cpp).
+  for (MachineBasicBlock &MBB : *MF)
+    for (MachineInstr &MI : llvm::make_early_inc_range(MBB))
+      if (MI.isKill() || MI.isImplicitDef())
+        MBB.erase(&MI);
```
andcarminati: It looks we overloaded `enterFunction` with too much responsibility. I don't mind a helper function do to this job ;-)

andcarminati: nit: `MI.eraseFromParent();`

llvm/test/CodeGen/AIE/aie2/schedule/meta_instrs.mir
mludevid: I removed the `undef.mir` test file that was not testing anything anymore and simplified the `meta_instrs.mir` test to only two tests that check that the two instructions are removed. I hope this is good to go now.

llvm/lib/Target/AIE/AIEMachineScheduler.cpp
mludevid: Done.

martien-de-jong (approved):

PR #879, opened 2026-03-27
[AIEX] Split wide concat+store into half-sized stores

abhinay-anubola:


llvm/lib/Target/AIE/AIECombine.td
```diff
@@ -531,18 +537,30 @@ def aie_postlegalizer_custom_shared_combines : GICombineGroup<[
...
 ]>;
 
-// Post-legalizer custom combines for AIE2P and more recent targets (AIE2PS, etc.).
-def aie2p_plus_postlegalizer_custom_combines : GICombineGroup<[
+// Post-legalizer custom combines for AIE2P
+def aie2p_postlegalizer_custom_combines : GICombineGroup<[
+  combine_extract_vector_assert_combine,
+  combine_extract_broadcast_to_scalar,
+  combine_vshift_chain_to_copy,
+  combine_pad_unpad_to_copy,
```
F-Stuckmann, line 568: nit: shouldn't the combiner group `aie2p_postlegalizer_custom_combines` be combined with `aie2p_additional_combines`?

llvm/lib/Target/AIE/AIECombine.td
```diff
@@ -531,18 +537,30 @@ def aie_postlegalizer_custom_shared_combines : GICombineGroup<[
...
+  combine_unpad_pad_to_copy,
+  combine_cse_vector_ops
+]>;
+
+// Post-legalizer custom combines for AIE2PS
+def aie2ps_postlegalizer_custom_combines : GICombineGroup<[
```
F-Stuckmann: `combine_extract_vector_assert_combine` and `combine_extract_broadcast_to_scalar` are shared between `aie2p` and `aie2ps`. We may want to move them also into a common combiner group

llvm/lib/Target/AIE/AIECombine.td
F-Stuckmann: Also should this be arch independent? why are we excluding it from aie2ps?

llvm/lib/Target/AIE/AIECombinerHelper.cpp
```diff
@@ -5199,6 +5199,59 @@ static void makeStoreDead(GStore *StMI, const TargetInstrInfo &TII,
...
+// a wide value that may require a more constrained register class.
+// Handles both G_STORE and G_AIE_OFFSET_STORE.
+//   store (G_CONCAT_VECTORS lo, hi), ptr
+// becomes:
+//   store hi, ptr + halfsize
+//   store lo, ptr
+bool llvm::matchSplitConcatStore(MachineInstr &StMI, MachineRegisterInfo &MRI,
+                                 BuildFnTy &MatchInfo) {
+  const Register DataReg = StMI.getOperand(0).getReg();
+  MachineInstr *const DataDef = MRI.getVRegDef(DataReg);
```
F-Stuckmann: nit: Rename DataDef to ConcatI, since that is what we are expecting

llvm/lib/Target/AIE/AIECombinerHelper.cpp
```diff
@@ -5199,6 +5199,59 @@ static void makeStoreDead(GStore *StMI, const TargetInstrInfo &TII,
...
+  const unsigned HalfSizeBytes = HalfTy.getSizeInBytes();
+  const Register PtrReg = StMI.getOperand(1).getReg();
+  const LLT PtrTy = MRI.getType(PtrReg);
+  const bool IsOffsetStore = StMI.getOpcode() != TargetOpcode::G_STORE;
+
+  MatchInfo = [=, &MRI, &StMI](MachineIRBuilder &B) {
+    MachineMemOperand *const MMO = *StMI.memoperands_begin();
+    // For G_AIE_OFFSET_STORE, incorporate the offset into the base pointer.
+    Register BasePtrReg = PtrReg;
+    if (IsOffsetStore) {
```
F-Stuckmann, line 5327: have tried running the combiner earlier and letting the address combiner automatically combine both parts for you?

Because you are missing the case where we have a postinc store

llvm/lib/Target/AIE/AIECombinerHelper.cpp
```diff
@@ -5199,6 +5199,59 @@ static void makeStoreDead(GStore *StMI, const TargetInstrInfo &TII,
...
+      !MRI.hasOneNonDBGUse(DataReg))
+    return false;
+
+  const Register LoReg = DataDef->getOperand(1).getReg();
+  const Register HiReg = DataDef->getOperand(2).getReg();
+  const LLT HalfTy = MRI.getType(LoReg);
+  const unsigned HalfSizeBytes = HalfTy.getSizeInBytes();
+  const Register PtrReg = StMI.getOperand(1).getReg();
+  const LLT PtrTy = MRI.getType(PtrReg);
+  const bool IsOffsetStore = StMI.getOpcode() != TargetOpcode::G_STORE;
```
F-Stuckmann: nit: this does not detect a offset. We could also match a postinc

llvm/lib/Target/AIE/AIECombinerHelper.cpp
```diff
@@ -5199,6 +5199,59 @@ static void makeStoreDead(GStore *StMI, const TargetInstrInfo &TII,
...
+
+  MatchInfo = [=, &MRI, &StMI](MachineIRBuilder &B) {
+    MachineMemOperand *const MMO = *StMI.memoperands_begin();
+    // For G_AIE_OFFSET_STORE, incorporate the offset into the base pointer.
+    Register BasePtrReg = PtrReg;
+    if (IsOffsetStore) {
+      Register OffsetReg = StMI.getOperand(2).getReg();
+      BasePtrReg = B.buildPtrAdd(PtrTy, PtrReg, OffsetReg).getReg(0);
+    }
+    // Store upper half at base + halfsize first.
```
F-Stuckmann, line 5331: if we have a postinc store we may actually want to reverse the order, so that the first store calculates the new postinc and the second store is an offset store.

llvm/lib/Target/AIE/AIECombinerHelper.cpp
martien-de-jong: The pattern has an explicit opcode list, so if it's not a store is an offset store. Still, I think it would be better to check usiong `== TargetOpcode::G_AIE_OFFSET_STORE`

llvm/lib/Target/AIE/AIECombinerHelper.cpp
```diff
@@ -5199,6 +5199,59 @@ static void makeStoreDead(GStore *StMI, const TargetInstrInfo &TII,
...
+// Split a store of a G_CONCAT_VECTORS result into two half-sized stores.
+// This allows each half to be stored independently, avoiding the creation of
+// a wide value that may require a more constrained register class.
+// Handles both G_STORE and G_AIE_OFFSET_STORE.
+//   store (G_CONCAT_VECTORS lo, hi), ptr
+// becomes:
+//   store hi, ptr + halfsize
+//   store lo, ptr
+bool llvm::matchSplitConcatStore(MachineInstr &StMI, MachineRegisterInfo &MRI,
+                                 BuildFnTy &MatchInfo) {
```
martien-de-jong, line 5290: I'm wondering whether we don't have this functionality for other architectures somewhere, e.g. splitting  an x register in wl and wh for AIE2. It would be good to unify then.

llvm/lib/Target/AIE/AIECombinerHelper.cpp
martien-de-jong: Again, this issue should have been resolved ages ago. The fact that it creeps up here is a sign that we are following different strategies for different architectures.

llvm/test/CodeGen/AIE/GlobalISel/combine-split-concat-store.mir
```diff
@@ -0,0 +1,131 @@
...
+# This file is licensed under the Apache License v2.0 with LLVM Exceptions.
+# See https://llvm.org/LICENSE.txt for license information.
+# SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+#
+# (c) Copyright 2026 Advanced Micro Devices, Inc. or its affiliates
+# RUN: llc -mtriple aie2p -run-pass=aie-postlegalizer-custom-combiner -verify-machineinstrs -o - %s | FileCheck %s
+# RUN: llc -mtriple aie2ps -run-pass=aie-postlegalizer-custom-combiner -verify-machineinstrs -o - %s | FileCheck %s
+
+---
+name:            split_concat_store_2048
```
martien-de-jong: So it seems we are doing 1024 and 2048 bits here. On AIE2 we those do not occur, but the code should probably be generalized for any dual input store, rather then making this target specific one.

llvm/lib/Target/AIE/AIECombinerHelper.cpp
```diff
@@ -5199,6 +5199,59 @@ static void makeStoreDead(GStore *StMI, const TargetInstrInfo &TII,
...
+  const Register DataReg = StMI.getOperand(0).getReg();
+  MachineInstr *const DataDef = MRI.getVRegDef(DataReg);
+
+  // Only match 2048-bit+ stores of two-operand G_CONCAT_VECTORS with a
+  // single use (the store itself).
+  if (!DataDef || DataDef->getOpcode() != TargetOpcode::G_CONCAT_VECTORS ||
+      DataDef->getNumOperands() != 3)
+    return false;
+  if (MRI.getType(DataReg).getSizeInBits() < 2048 ||
```
martien-de-jong: The 2048 should be a target-dependent number from TII

llvm/lib/Target/AIE/AIECombine.td
abhinay-anubola: `aie2p_additional_combines` are used as both in pre and post legalizer combines.

llvm/lib/Target/AIE/AIECombinerHelper.cpp
abhinay-anubola: I think `TargetOpcode::G_AIE_OFFSET_STORE` doesn't work because, `G_AIE_OFFSET_STORE` is a target-specific opcode

llvm/lib/Target/AIE/AIECombinerHelper.cpp
F-Stuckmann: could you reuse the existing combiner that calls `matchLoadStoreSplit`? It is currently an aie2 combiner

F-Stuckmann (approved):
For expediencies sake I am approving the PR, but we should think about reusing existing code and not fragmenting the combiners that much between architectures.

PR #880, opened 2026-03-27
Reduce differences to upstream in bf16 vs f16  GlobalISel code

jorickert:
This should make it easier to pull in upstream changes .

Instead of globally disabling Bf16 patterns for the GIsel, still allow them for selected targets


isoard-amd (approved):
Good. This align with `[SPIRV] Enable bfloat16 arithmetic` from upstream (https://github.com/llvm/llvm-project/pull/166031)

konstantinschwarz (approved):

PR #881, opened 2026-03-30
[AIE2PS] Fix VMAC scheduling bug

mludevid:
`VMAC_f_vmac_bfp16` reads an `EX` composite register through two sub-register operands (`ewl` at pipeline cycle 1, `ewh` at cycle 2). When the scheduler builds anti-dependency edges from these reads to a subsequent `VSHUFFLE` that overwrites the same `EX` register, it creates two edges with different signed latencies (-1 for `ewl`, 0 for `ewh`). Both edges get the same `SDep` register (the super-register from the DEF side), so `SUnit::addPred` considers them overlapping and deduplicates.

The deduplication compared latencies using `getLatency()`, which clamps negative values to 0. This made -1 and 0 indistinguishable, so whichever edge was added first survived. Since `ewl` (operand index 2) is always processed before `ewh` (operand index 5), the looser constraint (-1) won and the tighter one (0) was dropped. This allowed the `VSHUFFLE` to be placed one cycle too early.

Fix: Change `addPred`'s deduplication comparison from `getLatency()` to `getSignedLatency()`, and use `setSignedLatency()` for the update. This correctly distinguishes -1 from 0 and keeps the tighter constraint.


llvm/lib/Target/AIE/AIEBaseSubtarget.cpp
```diff
@@ -151,6 +151,21 @@ void AIEBaseSubtarget::adjustSchedDependency(
...
+    // making edges with signed latencies like -1 and 0 look identical. If the
+    // looser constraint (-1) was added first, the tighter one (0) won't
+    // replace it. Fix this by updating any existing overlapping edge that has
+    // a smaller signed latency than the one we're about to add.
+    for (SDep &PredDep : Use->Preds) {
+      if (PredDep.overlaps(Dep) && PredDep.getSignedLatency() < *Lat) {
+        getForwardEdge(*Use, PredDep).setSignedLatency(*Lat);
+        PredDep.setSignedLatency(*Lat);
+      }
+    }
```
martien-de-jong: As discussed, this is just facilitates comparing two edge latencies in a place that is designed to compute one.  Since the architecture decides to fill in signed latencies, the central comparison in generic code would not change when doing that comparison using getSignedLatency. I prefer to do it there, as one more step to get to a sane, signed, notion of latency.

mludevid: Done. I've also updated the PR description to reflect the new approach. I was hesitant to change LLVM generic code, but this is definitely cleaner.

martien-de-jong (approved):

PR #885, opened 2026-04-01
[AIE2PS] Improve G_PHI mapping for accumulators

andcarminati:
This is designed to solve the problem exposed by this case:

```
  bb.0:
    successors: %bb.1
    %0:_(s32) = G_CONSTANT i32 0
    %1:_(<16 x s32>) = G_AIE_BROADCAST_VECTOR %0(s32)
  bb.1:
    successors: %bb.2
    %2:_(<16 x s32>) = G_PHI %1(<16 x s32>), %bb.0, %1(<16 x s32>), %bb.0 <- will be mapped to vec because of the broadcast
    %3:_(<32 x s32>) = G_CONCAT_VECTORS %2(<16 x s32>), %2(<16 x s32>)
    %4:_(<32 x s32>) = G_INTRINSIC_W_SIDE_EFFECTS intrinsic(@llvm.aie2ps.ACC1024.accfloat.add.conf), %3(<32 x s32>), %3(<32 x s32>), 60
    %5:_(<16 x s32>), %6:_(<16 x s32>) = G_UNMERGE_VALUES %4(<32 x s32>)
  bb.2:
```

This can't be covered by the look through `G_PHI` engine that we already have, because we can't map `G_AIE_BROADCAST_VECTOR`  to acc bank.


llvm/lib/Target/AIE/aie2ps/AIE2PSRegisterBankInfo.cpp
```diff
@@ -883,6 +883,37 @@ AIE2PSRegisterBankInfo::getInstrMapping(const MachineInstr &MI) const {
...
+    // Only handle vector PHIs >= 512 bits
+    if (!(DstType.isVector() && DstType.getSizeInBits() >= 512))
+      return AIEBaseRegisterBankInfo::getInstrMapping(MI);
+
+    // Check if PHI result prefers accumulator bank
+    const auto PreferredRegBank =
+        getPreferredRegBankForVectorTy(MRI, TRI, DstReg);
+    PartialMappingIdx BankIdx;
+
+    if (PreferredRegBank && *PreferredRegBank == &AIE2PS::AccRegBank) {
```
F-Stuckmann, line 900: can we have TTI hooks for ACCRegBank and FIFORegBank so that this reg bank select feature can be enabled on all Targets?

andcarminati: This is part of `getInstrMapping` which is implemented by each AIE Target completely. A refactor in this moments if out of scope of this PR.

llvm/test/CodeGen/AIE/GlobalISel/regbankselect-phi-accumulator-aie2ps.mir
```diff
@@ -24,10 +24,8 @@ body: |
   ; CHECK-NEXT: bb.1:
   ; CHECK-NEXT:   successors: %bb.2(0x80000000)
   ; CHECK-NEXT: {{  $}}
-  ; CHECK-NEXT:   [[PHI:%[0-9]+]]:vregbank(<16 x s32>) = G_PHI [[AIE_BROADCAST_VECTOR]](<16 x s32>), %bb.0, [[AIE_BROADCAST_VECTOR]](<16 x s32>), %bb.0
-  ; CHECK-NEXT:   [[COPY:%[0-9]+]]:accregbank(<16 x s32>) = COPY [[PHI]](<16 x s32>)
-  ; CHECK-NEXT:   [[COPY1:%[0-9]+]]:accregbank(<16 x s32>) = COPY [[PHI]](<16 x s32>)
-  ; CHECK-NEXT:   [[CONCAT_VECTORS:%[0-9]+]]:accregbank(<32 x s32>) = G_CONCAT_VECTORS [[COPY]](<16 x s32>), [[COPY1]](<16 x s32>)
+  ; CHECK-NEXT:   [[PHI:%[0-9]+]]:accregbank(<16 x s32>) = G_PHI [[AIE_BROADCAST_VECTOR]](<16 x s32>), %bb.0, [[AIE_BROADCAST_VECTOR]](<16 x s32>), %bb.0
```
niwinanto, line 27: Q: After PHI elimination, we introduce copies again right? Or are we getting benefit in terms of better inner loop? Also, what happens if live-ins coming from different basic blocks, can we have a test?

F-Stuckmann (commented):
LGTM

llvm/test/CodeGen/AIE/GlobalISel/regbankselect-phi-accumulator-aie2ps.mir
andcarminati: The copies will be outside the block. This basically removes copies from and back vectors from important kernels - enabling pipelining. Live-ins?

martien-de-jong (approved):

PR #886, opened 2026-04-02
[AIEX] Fix ABI size of *_unaligned data types 

konstantinschwarz:


isoard-amd (approved):
nit: we could have used `alignas`?

Either way, good find!

konstantinschwarz:
> nit: we could have used `alignas`?
> 
> Either way, good find!

But to match the ABI, I think we also need to guarantee sizeof(type) == 4

PR #888, opened 2026-04-02
feat: add tosa.select folding

ge28boj:


mlir/lib/Dialect/Tosa/Transforms/TosaFolders.cpp
```diff
@@ -1438,6 +1438,81 @@ struct TosaFoldConstantEqual
...
+    matchPattern(onFalseVal, m_Constant(&onFalseAttr));
+
+    if (foldSplatOrSingleUseOnly && !isa<SplatElementsAttr>(predAttr) &&
+        !isa<SplatElementsAttr>(onTrueAttr) &&
+        !isa<SplatElementsAttr>(onFalseAttr) && !predVal.hasOneUse() &&
+        !onTrueVal.hasOneUse() && !onFalseVal.hasOneUse()) {
+      return rewriter.notifyMatchFailure(
+          op, "Select will only be folded if at least one input is splat or "
+              "has a single use");
+    }
```
jorickert: is it inteded that we check that any input is splat or single use and not all?

mlir/lib/Dialect/Tosa/Transforms/TosaFolders.cpp
```diff
@@ -1438,6 +1438,81 @@ struct TosaFoldConstantEqual
...
 
+template <typename ValType>
+DenseElementsAttr foldSelectWithValueType(DenseElementsAttr predAttr,
+                                          DenseElementsAttr onTrueAttr,
+                                          DenseElementsAttr onFalseAttr,
+                                          TensorType outputType) {
+  auto targetShape = outputType.getShape();
+  auto predShape = predAttr.getType().getShape();
+  auto onTrueShape = onTrueAttr.getType().getShape();
+  auto onFalseShape = onFalseAttr.getType().getShape();
```
jorickert, line 1449: check for static shape?

mlir/lib/Dialect/Tosa/Transforms/TosaFolders.cpp
ge28boj: updated, should check if all, not any

jorickert (approved):
Thanks

PR #893, opened 2026-04-02
[AIEX][AIE2PS] Finish refactor of postlegalizer combiners

andcarminati:
This PR also enables more combiners for AIE2PS.

### DIFF
--- a/llvm/lib/Target/AIE/AIECombine.td
+++ b/llvm/lib/Target/AIE/AIECombine.td
@@ -542,9 +542,13 @@ def aie2_postlegalizer_custom_combines : GICombineGroup<[
 ]>;
 
 // Post-legalizer custom combines for AIE2P and more recent targets (AIE2PS, etc.).
-def aie2p_plus_postlegalizer_custom_combines : GICombineGroup<[
+def aie2p_plus_postlegalizer_custom_shared_combines : GICombineGroup<[
   combine_extract_vector_assert_combine,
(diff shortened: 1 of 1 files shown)


llvm/lib/Target/AIE/AIECombine.td
```diff
@@ -544,7 +544,11 @@ def aie2_postlegalizer_custom_combines : GICombineGroup<[
 // Post-legalizer custom combines for AIE2P and more recent targets (AIE2PS, etc.).
 def aie2p_plus_postlegalizer_custom_combines : GICombineGroup<[
   combine_extract_vector_assert_combine,
-  combine_extract_broadcast_to_scalar
+  combine_extract_broadcast_to_scalar,
+  combine_vshift_chain_to_copy,
```
F-Stuckmann, line 548: shouldn't these combiners be also backported to aie2? Can we have a generic aie_postlegalizer_custom_shared_combines?

andcarminati: I feel unsafe, as the biggest difference between aie2 and newer variants is exactly the vector handling. We may generate non-legalizable/non-selectable code in the worst case.

F-Stuckmann: could we rename it to aie2p_common_postlegalizer_custom_combines

F-Stuckmann (dismissed):

F-Stuckmann (approved):

PR #896, opened 2026-04-02
[MC] Replace assertions with diagnostics in immediate encoder

### LINKED ISSUE #858: AIE2: llvm-mc crashes on SP-relative instructions with assertion in getSImmOpValueXStep
## Summary

`llvm-mc` crashes with an assertion failure when assembling SP-relative load/store instructions on the AIE2 target. The crash occurs in `getSImmOpValueXStep` in `AIEBaseMCCodeEmitter.h`.

## Reproduction

Minimal repro (any of these crash):

```sh
echo 'vlda wl0, [sp, #0]' | llvm-mc --triple=aie2 --filetype=obj -o /dev/null
echo 'vst amll0, [sp, #0]' | llvm-mc --triple=aie2 --filetype=obj -o /dev/null
echo 'paddb [sp], #4' | llvm-mc --triple=aie2 --filetype=obj -o /dev/null
```

Negative offsets work for some variants:
```sh
# This works:
echo 'vlda wl0, [sp, #-32]' | llvm-mc --triple=aie2 --filetype=obj -o /dev/null
# This crashes:
echo 'vlda wl0, [sp, #0]' | llvm-mc --triple=aie2 --filetype=obj -o /dev/null
```

Non-SP-relative variants of the same instructions assemble correctly:
```sh
# These all work fine:
echo 'vlda wl0, [p2, #0]' | llvm-mc --triple=aie2 --filetype=obj -o /dev/null
echo 'vlda amll0, [p2, #0]' | llvm-mc --triple=aie2 --filetype=obj -o /dev/null
echo 'vst amll0, [p2, #0]' | llvm-mc --triple=aie2 --filetype=obj -o /dev/null
```

## Assertion

```
llvm-mc: AIEBaseMCCodeEmitter.h:142:
void llvm::(anonymous)::getSImmOpValueXStep(...) [with int N = 12; unsigned int step = 32;
int offset = 0; bool isNegative = true; bool isUnsigned = false]:
Assertion `Imm >= Min && Imm <= Max && "can not represent value in the given immediate type range!"' failed.
```

The `isNegative = true` template parameter suggests the encoder variant only accepts negative offsets. For `paddb [sp]`, the assertion fires even with `#0`:

```
getSImmOpValueXStep [with int N = 12; unsigned int step = 32; int offset = 0;
bool isNegative = false; bool isUnsigned = false]:
Assertion `(Imm & hexValue) == 0 && "Value must be divisible by step!"` failed.
```

## Affected instructions

All SP-relative (`[sp, $imm]`) variants appear affected:
- `vlda $dst, [sp, $imm]` (vector and am-class spill loads)
- `vst $src, [sp, $imm]` (vector and am-class spill stores)
- `paddb [sp], $imm` (B-port pointer advance)
- `padda [sp], $imm` works only with multiples of 32

## Environment

- Branch: `aie-public` (commit from March 2026)
- Build: Debug+Asserts
- Triple: `aie2`

## Context

Found while developing an ISA-level test generator that exercises all AIE2 instruction encodings via `llvm-mc`. The SP-relative spill instructions are the only class that consistently crashes the assembler.

FIM43-Redeye:
Replaces the two `assert`s in `getSImmOpValueXStep` with `report_fatal_error` calls that print the offending value and the valid range.

Before this change, assembling an out-of-range SP-relative offset (e.g. `vlda wl0, [sp, #0]`) crashed with an opaque assertion. Now it produces a clear error message.

Fixes #858


FIM43-Redeye:
Rebased onto latest aie-public and addressed the Copilot feedback:

- `report_fatal_error` now passes `GenCrashDiag=false` so the error exits cleanly without a crash backtrace
- Test updated from `not --crash` to `not` to match the clean-exit behavior

This fixes a real crash we hit when SP-relative immediates don't fit (#858) -- the assertion fires in release+asserts builds and the encoder silently miscompiles in pure release builds. The diagnostic makes the failure mode clear in both cases.

Would appreciate a review when someone has a chance.

konstantinschwarz (dismissed):
LGTM, thanks for fixing this!

konstantinschwarz (approved):

PR #912, opened 2026-04-08
[AIE2PS][GISel] Legalize G_FNEG (scalar XOR + vector broadcast/XOR)

katerynamuts:
This change legalizes G_FNEG for AIE2PS in GlobalISel.

Scalars (S16, S32, S64) stay on the generic path: lowerFor so LegalizerHelper lowers to G_XOR with the correct per-type sign mask.

Vectors (V16S32, V32S32, V32S16, V64S8) use custom legalization in AIELegalizerHelper::legalizeG_FNEG: build a scalar sign constant, G_AIE_BROADCAST_VECTOR, then G_XOR. This avoids the long G_AIE_ADD_VECTOR_ELT_HI chains produced by the generic splat lowering. Wider-than–basic-vector types (e.g. <32 x s32>) broadcast once at the legal 512-bit chunk width and reuse that mask for each unmerged chunk before G_CONCAT_VECTORS.


llvm/lib/Target/AIE/AIELegalizerHelper.cpp
```diff
@@ -1474,6 +1474,68 @@ bool AIELegalizerHelper::legalizeG_FMUL(LegalizerHelper &Helper,
...
+  assert(DstTy.isVector() && "Expected vector type for custom G_FNEG");
+
+  const unsigned EltSize = DstTy.getScalarSizeInBits();
+  const APInt SignMask = APInt::getSignMask(EltSize);
+
+  // G_AIE_BROADCAST_VECTOR takes a 32-bit (or 64-bit) scalar. For s8/s16
+  // element types the hardware VBCST instruction packs the lower bits of a
+  // 32-bit source into every lane, so extend the mask to i32.
+  const LLT ScalarTy = LLT::scalar(EltSize < 32 ? 32 : EltSize);
+  Register ScalarMask =
```
mludevid: Nit: `const`

llvm/lib/Target/AIE/aie2ps/AIE2PSLegalizerInfo.cpp
```diff
@@ -234,6 +234,10 @@ AIE2PSLegalizerInfo::AIE2PSLegalizerInfo(const AIE2PSSubtarget &ST)
 
   getActionDefinitionsBuilder(G_FABS).customFor({S16, S32, S64}).scalarize(0);
 
+  // Lower to G_XOR + per-lane sign mask (LegalizerHelper G_FNEG case).
+  getActionDefinitionsBuilder(G_FNEG).lowerFor(
+      {S16, S32, S64, V16S32, V32S32, V32S16, V64S8});
```
mludevid: What about `V64S16` and `V128S8`?

katerynamuts: I added them, thanks for noticing this

llvm/lib/Target/AIE/AIELegalizerHelper.cpp
```diff
@@ -1474,6 +1474,68 @@ bool AIELegalizerHelper::legalizeG_FMUL(LegalizerHelper &Helper,
...
+
+  const Register DstReg = MI.getOperand(0).getReg();
+  const Register SrcReg = MI.getOperand(1).getReg();
+  const LLT DstTy = MRI.getType(DstReg);
+  assert(DstTy.isVector() && "Expected vector type for custom G_FNEG");
+
+  const unsigned EltSize = DstTy.getScalarSizeInBits();
+  const APInt SignMask = APInt::getSignMask(EltSize);
+
+  // G_AIE_BROADCAST_VECTOR takes a 32-bit (or 64-bit) scalar. For s8/s16
```
mludevid: Nit: I don't think this code is ever called for 64-bit scalars. This might be misleading.

mludevid (dismissed):
LGTM

mludevid (approved):

PR #915, opened 2026-04-08
[AIEX] WAW allow cross bank Reg Copies to be renamed in loops

F-Stuckmann:
Allow WAW Renamer to rename cross bank copies, since they cannot be copy coalesced.


andcarminati (approved):
LGTM.

PR #916, opened 2026-04-08
[AIEX] Generic Loop Pragmas

F-Stuckmann:
Added generic loop pragmas, so that adding loop pragmas becomes as easy as adding the aie-hint in cpp and telling the backend how to handle it.


F-Stuckmann:
I am still debating on whether to make the hints clang native, even if we then have more diffs, because we would then need less plumbing.

clang/test/CodeGen/aie/looppragma-aie-hint.c
```diff
@@ -0,0 +1,159 @@
...
+// CHECK-NEXT:    br label %[[FOR_COND]], !llvm.loop [[LOOP7:![0-9]+]]
+// CHECK:       [[FOR_END]]:
+// CHECK-NEXT:    [[TMP6:%.*]] = load i32, ptr [[S]], align 4
+// CHECK-NEXT:    ret i32 [[TMP6]]
+//
+int test_multi_hint(int *p, int n) {
+  int s = 0;
+  #pragma clang loop min_iteration_count(3) max_iteration_count(6)
+  #pragma clang loop aie_hint(no_predication)
+  #pragma clang loop aie_hint(target_ii, 8)
```
andcarminati: What is the difference between this and the existing pipeliner pragma?

clang/test/CodeGen/aie/looppragma-aie-hint.c
```diff
@@ -0,0 +1,159 @@
+// NOTE: Assertions have been autogenerated by utils/update_cc_test_checks.py UTC_ARGS: --check-globals all --version 5
+// --- looppragma-aie-hint.c ------------------------------------------------///
+//
+// This file is licensed under the Apache License v2.0 with LLVM Exceptions.
+// See https://llvm.org/LICENSE.txt for license information.
+// SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+//
+// (c) Copyright 2026 Advanced Micro Devices, Inc. or its affiliates
+// --------------------------------------------------------------------------///
+// RUN: %clang --target=aie2 -S -emit-llvm %s -o - | FileCheck %s
```
andcarminati: Better aie2ps.

clang/test/CodeGen/aie/peel-aie-hint.c
```diff
@@ -0,0 +1,137 @@
...
+// CHECK-NEXT:    [[ADD]] = add nsw i32 [[TMP4]], [[S_04]]
+// CHECK-NEXT:    [[INC]] = add nuw nsw i32 [[I_05]], 1
+// CHECK-NEXT:    [[EXITCOND_NOT:%.*]] = icmp eq i32 [[INC]], [[N]]
+// CHECK-NEXT:    br i1 [[EXITCOND_NOT]], label %[[FOR_COND_CLEANUP]], label %[[FOR_BODY]], !llvm.loop [[LOOP6:![0-9]+]]
+//
+int bool_hint_peel(int n, int *p) {
+  int s = 0;
+  #pragma clang loop min_iteration_count(28)
+  #pragma clang loop max_iteration_count(37)
+  #pragma clang loop aie_hint(no_predication)
```
andcarminati: What is the meaning of "no_predication"?

clang/include/clang/Basic/Attr.td
```diff
@@ -4369,6 +4369,19 @@ def LoopHint : Attr {
   let HasCustomParsing = 1;
 }
 
+
+/// AIE-specific generic loop hint pragma.
```
andcarminati: I see aie here in the comment. Apart of that, the implementation seems to be pretty generic. Also, `specific generic` seems to be +1-1 ;-).

clang/lib/CodeGen/CGLoopInfo.cpp
```diff
@@ -464,6 +464,19 @@ MDNode *LoopInfo::createMetadata(
     LoopProperties.push_back(MDNode::get(Ctx, {Vals, NumMD}));
   }
 
+  // Emit generic AIE loop hint metadata.
+  for (const auto &[Key, Value] : Attrs.AIEHints) {
+    const std::string MDName = "llvm.loop.hint." + Key;
```
andcarminati, line 469: Should we use aiehint or make this really generic?

clang/test/CodeGen/aie/peel-aie-hint.c
```diff
@@ -0,0 +1,137 @@
...
+// CHECK-NEXT:    [[INC]] = add nuw nsw i32 [[I_05]], 1
+// CHECK-NEXT:    [[EXITCOND_NOT:%.*]] = icmp eq i32 [[INC]], [[N]]
+// CHECK-NEXT:    br i1 [[EXITCOND_NOT]], label %[[FOR_COND_CLEANUP]], label %[[FOR_BODY]], !llvm.loop [[LOOP13:![0-9]+]]
+//
+int multi_hint_peel(int n, int *p) {
+  int s = 0;
+  #pragma clang loop min_iteration_count(28)
+  #pragma clang loop max_iteration_count(37)
+  #pragma clang loop aie_hint(no_predication)
+  #pragma clang loop aie_hint(target_ii, 5)
```
andcarminati: Could we add tests for ill-formed pragmas? Mostly to check parsing and diagnostic.

clang/lib/CodeGen/CGLoopInfo.h
```diff
@@ -93,6 +93,10 @@ struct LoopAttributes {
 
   /// Value for whether the loop is required to make progress.
   bool MustProgress;
+
+  /// Generic AIE loop hints: pairs of (key, optional integer value).
+  /// Emitted as llvm.loop.hint.<key> metadata.
+  llvm::SmallVector<std::pair<std::string, std::optional<int64_t>>, 2> AIEHints;
```
andcarminati: Could we have an even more generic implementation? Something to be used like:
```
#pragma clang loop aie_hint(use_pipeliner, pre)
```

andcarminati:
Maybe something like "codegen_hints" or "target_hints", because the target will use this to guide optimizations and code generation.

clang/test/CodeGen/aie/peel-aie-hint.c
F-Stuckmann: placeholder for anything you want to add

clang/lib/CodeGen/CGLoopInfo.h
F-Stuckmann: so the value should be a string, right?

clang/lib/CodeGen/CGLoopInfo.cpp
F-Stuckmann: I will check how much work it is it make it fully generic

F-Stuckmann:
1) Moved unit-tests to aie2ps
2) allow strings as keys

I kept aie hints, since we are the only ones using them and we could reuse them in the optimizer, so they aren't really only codegen hints

clang/lib/Parse/ParsePragma.cpp
```diff
@@ -1581,6 +1581,54 @@ bool Parser::HandlePragmaLoopHint(LoopHint &Hint) {
       // Argument is a constant expression with an integer type.
       Hint.ValueExpr = R.get();
     }
+  } else if (OptionInfo && OptionInfo->isStr("aie_hint")) {
```
andcarminati: I see that we turn this into a .hint. so, why not just use "hint"? We could put in a shape that could be upstreamed.

andcarminati:
The diagnostic seems odd for `#pragma clang loop aie_hint()` (string?) : 
```
error: missing argument; expected an integer value
```

andcarminati:
Should we accept negative numbers?
```
error: invalid value '-5'; must be positive
    8 |    #pragma clang loop aie_hint(foo, -5)
      |                                     ^
1 error generated.
```

F-Stuckmann:
> The diagnostic seems odd for `#pragma clang loop aie_hint()` (string?) :
> 
> ```
> error: missing argument; expected an integer value
> ```

negative numbers are not natively supported, so I would rather keep it that way instead of introducing more upstream changes.

clang/lib/Parse/ParsePragma.cpp
F-Stuckmann: upstream already has hints that are tightly guarded, i.e. the allowed keys are limited. 

We however want to add generic keys and values and thus we should separate them instead of modifying hints upstream.

clang/lib/Parse/ParsePragma.cpp
```diff
@@ -1581,6 +1584,54 @@ bool Parser::HandlePragmaLoopHint(LoopHint &Hint) {
...
+      return false;
+    }
+    Hint.StateLoc =
+        IdentifierLoc::create(Actions.Context, Tok.getLocation(), KeyInfo);
+    PP.Lex(Tok);
+
+    // Optionally parse a comma followed by an integer or identifier value.
+    if (Tok.is(tok::comma)) {
+      PP.Lex(Tok);
+      if (IdentifierInfo *ValIdent = Tok.getIdentifierInfo()) {
```
martien-de-jong: This doesn't sound like an arbitrary string. Can I add "LC + 4 * (K + 42)" or even "LC+4*K" ?

clang/test/CodeGen/aie/looppragma-aie-hint.c
```diff
@@ -127,14 +129,60 @@ int test_int_hint(int *p, int n) {
...
+// CHECK-NEXT:    [[INC:%.*]] = add nsw i32 [[TMP5]], 1
+// CHECK-NEXT:    store i32 [[INC]], ptr [[I]], align 4
+// CHECK-NEXT:    br label %[[FOR_COND]], !llvm.loop [[LOOP12:![0-9]+]]
+// CHECK:       [[FOR_END]]:
+// CHECK-NEXT:    [[TMP6:%.*]] = load i32, ptr [[S]], align 4
+// CHECK-NEXT:    ret i32 [[TMP6]]
+//
+int test_string_hint(int *p, int n) {
+  int s = 0;
+  #pragma clang loop hint(use_pipeliner, pre)
```
martien-de-jong, line 185: So I'd like to have a slightly less restrictive string here. In particular containing interpunction characters, possibly using some escapes for syntactic delimitors.
Pertinent example: clang loop hint (config, "pipelinesolve(NS=4,II=7)")

clang/test/CodeGen/aie/aie-hint-errors.c
```diff
@@ -0,0 +1,32 @@
...
+
+  // expected-error@+1 {{missing argument; expected a hint key identifier}}
+  #pragma clang loop hint()
+  for (int i = 0; i < n; i++) { s += p[i]; }
+
+  // expected-error@+1 {{expected expression}}
+  #pragma clang loop hint(key, )
+  for (int i = 0; i < n; i++) { s += p[i]; }
+
+  // expected-error@+1 {{invalid value '0'; must be positive}}
```
andcarminati: Could you check?

andcarminati: Hou about `#pragma clang loop hint(key, false)`?

F-Stuckmann: i added it

andcarminati:
What happens with:
```
#pragma clang loop hint(key, 0)
#pragma clang loop hint(key, foo)
```

?

clang/lib/CodeGen/CGLoopInfo.cpp
```diff
@@ -658,6 +656,9 @@ void LoopInfoStack::push(BasicBlock *Header, clang::ASTContext &Ctx,
         Value = ValueAPS.getSExtValue();
       } else if (!GenericHint->getValueStr().empty()) {
         Value = GenericHint->getValueStr().str();
+      } else {
+        // Key-only hint (e.g. hint(no_predication)) defaults to true.
+        Value = static_cast<int64_t>(1);
```
andcarminati, line 661: Nice!

clang/lib/Parse/ParsePragma.cpp
F-Stuckmann: yeah you can add that if you want

andcarminati (dismissed):
LGTM.

clang/lib/Parse/ParsePragma.cpp
```diff
@@ -1603,15 +1604,41 @@ bool Parser::HandlePragmaLoopHint(LoopHint &Hint) {
...
       PP.Lex(Tok);
-      if (IdentifierInfo *ValIdent = Tok.getIdentifierInfo()) {
-        // String-typed value (e.g. hint(key, ident)).
+      // Check for true/false keywords or identifiers first so they map to
+      // integer 1/0 instead of being treated as string-valued hints.
+      const bool IsTrueKw = Tok.is(tok::kw_true);
+      const bool IsFalseKw = Tok.is(tok::kw_false);
+      IdentifierInfo *ValIdent = Tok.getIdentifierInfo();
+      const bool IsTrueIdent = ValIdent && ValIdent->isStr("true");
+      const bool IsFalseIdent = ValIdent && ValIdent->isStr("false");
```
martien-de-jong, line 1630: Are these reachable? Is the spelling different from the keywords?

clang/lib/Sema/SemaStmtAttr.cpp
```diff
@@ -132,12 +133,19 @@ static Attr *handleLoopHintAttr(Sema &S, Stmt *St, const ParsedAttr &A,
...
     StringRef HintKey = StateLoc->Ident->getName();
-    if (ValueExpr && !ValueExpr->isValueDependent()) {
-      if (S.CheckLoopHintExpr(ValueExpr, St->getBeginLoc(),
-                              /*AllowZero=*/true))
-        return nullptr;
-    }
+    // If the value is a string literal, extract its content and treat it
+    // as a string-valued hint rather than an expression.
     StringRef ValueStr;
+    if (ValueExpr) {
```
martien-de-jong, line 139: so a string literal is actually part of an expression. Are there string operators in that expression syntax? Because then the string recognition above eats away a term of a larger expression.

F-Stuckmann: in the case of string operators in an expression, we parse a string, not an expression.

clang/lib/Parse/ParsePragma.cpp
F-Stuckmann: They are needed for pre C23 standards

martien-de-jong (approved):
Looks good  to me. I'd say we can merge this and iron out small flaws in follow ups.

PR #919, opened 2026-04-08
[AutoBump] Merge with fixes of 5aa4979c (Feb 04) (22)

jorickert:


llvm/lib/CodeGen/MachineScheduler.cpp
```diff
@@ -437,73 +548,85 @@ ScheduleDAGInstrs *PostMachineScheduler::createPostMachineScheduler() {
...
+
+  PreservedAnalyses PA = getMachineFunctionPassPreservedAnalyses();
+  PA.preserveSet<CFGAnalyses>();
+  PA.preserve<SlotIndexesAnalysis>();
+  PA.preserve<LiveIntervalsAnalysis>();
+  return PA;
 }
 
-bool PostMachineScheduler::runOnMachineFunction(MachineFunction &mf) {
-  bool SchedulingRequired = mf.getSubtarget().forcePostRAScheduling();
```
jorickert, line 478: Need a review here: The PostMachineSchedulerLegacy::runOnMachineFunction checks MF.getSubtarget().forcePostRAScheduling(); , but PostMachineSchedulerPass::run( does not, not sure if we should check MF.getSubtarget().forcePostRAScheduling() there too

konstantinschwarz: @martien-de-jong @andcarminati thoughts?

andcarminati: This commit is not enabling the new pass framework for AIE targets, the backend still configures the pipeline using the legacy infra. We use:

```
addPass(&PostMachineSchedulerID);
```

and 

```
char &llvm::PostMachineSchedulerID = PostMachineSchedulerLegacy::ID;
```
In this way, it is safe to keep as is. As discussed with @martien-de-jong, we need to carefully configure our pass pipeline to make use of this and we will take the proper actions to make it work correctly.

martien-de-jong (approved):
LGTM

PR #932, opened 2026-04-10
[PromoteMem2Reg] Replace lifetime.start/end with store undef before promotion

isoard-amd:
When PromoteMemToReg promotes an alloca, it erases all lifetime intrinsics via removeIntrinsicUsers(). This loses the semantic information that the alloca's content is undefined after lifetime.start/end, causing the SSA renaming pass to create spurious loop-carried phi nodes that carry a stale value from the previous iteration across a back-edge.

Fix this by replacing each lifetime.start/end with an actual store of undef to the alloca, instead of simply deleting it. The standard SSA renaming then sees a real definition of undef at that point, and back-edge phi nodes are naturally resolved to undef — eliminating spurious loop-carried dependences without any special-casing in the renaming algorithm.


isoard-amd:
I don't think it is fully ready, we should check that the byte-length in the lifetime intrinsic match the type store size.

isoard-amd:
We are attempting to push this change upstream https://github.com/llvm/llvm-project/pull/191909 (mostly to get more eyeballs on it)

konstantinschwarz (approved):

PR #936, opened 2026-04-14
[AIEX] PostPipeliner New Heuristics

F-Stuckmann:
Add a new EffectiveHeight priority component and DeferNonCritical placement modifier to the postpipeliner.

Add PlacementModifier as a new mechanism parallel to PriorityComponent. While PriorityComponents control which node is picked  next, PlacementModifiers adjust which cycle a picked node lands in. The first modifier, DeferNonCritical, nudges non-critical  nodes (EffectiveHeight == 0) one cycle later so their earliest modulo slot stays free for critical-path nodes.


llvm/lib/Target/AIE/AIEPostPipeliner.cpp
```diff
@@ -455,6 +455,35 @@ void PostPipeliner::computeRecMII() {
...
+  for (int K = NInstr - 1; K >= 0; K--) {
+    const SUnit &SU = DAG->SUnits[K];
+    int MaxEH = 0;
+    for (const SDep &Dep : SU.Succs) {
+      if (Dep.getKind() != SDep::Data)
+        continue;
+      const int S = Dep.getSUnit()->NodeNum;
+      const int Latency = Dep.getSignedLatency();
+      const bool IsLoopCarried = S >= NInstr;
+      const int SuccIdx = IsLoopCarried ? S - NInstr : S;
```
martien-de-jong: IsLoopCarried and  SuccIdx can be moved in the if statement below. Here they are confusing.

F-Stuckmann: I refactored the whole function to make it clearer what we are actually looking at

llvm/lib/Target/AIE/AIEPostPipeliner.cpp
```diff
@@ -455,6 +455,38 @@ void PostPipeliner::computeRecMII() {
...
+      // Skip edges not on the critical path: the successor is already
+      // pushed later by another predecessor, so this edge has slack.
+      const bool HasSlack = Info[K].Earliest + Latency < Info[S].Earliest;
+      if (HasSlack)
+        continue;
+
+      // Accumulate height along this critical edge.
+      const bool IsLoopCarried = S >= NInstr;
+      const int SuccIdx = IsLoopCarried ? S - NInstr : S;
+      const int Candidate = Latency + Info[SuccIdx].EffectiveHeight;
```
martien-de-jong: If we cross the iteration boundary, we take the effective height from something that may be very early in the basic block. I guess it will usually be zero since that node has not been processed yet, but it looks iffy. I think that sinks of the first iteration should be considered height 0 rather than using a default initial value from an unprocessed node.

andcarminati (dismissed):
LGTM.

martien-de-jong (approved):

andcarminati (approved):
LGTM.

llvm/lib/Target/AIE/AIEPostPipeliner.cpp
```diff
@@ -1089,8 +1123,27 @@ class ConfigStrategy : public PostPipelinerStrategy {
...
+    PlacementSize
+  };
+  static std::string getPlacementName(PlacementModifier Mod) {
+    switch (Mod) {
+    case PlacementModifier::DeferNonCritical:
+      return "Defer";
+    default:
+      break;
+    }
+    return "PlacementSize - Illegal";
```
martien-de-jong, line 1145: nit: PlacementSize ?

PR #938, opened 2026-04-15
[AIEX] Fix domination errors in AIESuperRegRewriter for dead sub-lanes

SagarMaheshwari99:
1. rewriteFullCopy skipped sub-copies when the source lane was dead, leaving the destination sub-register with no definition at all. Fix by emitting an IMPLICIT_DEF for those lanes instead of skipping them.

2. rewriteOperandsToSubRegs could produce uses of sub-VRegs that have no reaching definition. This happens when a composite register's main live interval spans a region (e.g. via a PHI) where a specific sub-lane is not actually live. After decomposition the new sub-VReg inherits the use but has no def to dominate it. Fix by checking sub-lane liveness at each use site and marking non-live uses as undef.


llvm/lib/Target/AIE/AIESuperRegUtils.cpp
```diff
@@ -113,8 +113,17 @@ void rewriteFullCopy(MachineInstr &CopyMI, LiveIntervals &LIS,
...
                         << TRI.getSubRegIndexName(SubRegIdx) << "\n");
+      MachineInstr *ImplDef =
+          BuildMI(*CopyMI.getParent(), CopyMI, CopyMI.getDebugLoc(),
+                  TII.get(TargetOpcode::IMPLICIT_DEF))
+              .addReg(DstReg, RegState::Define | AdditionalFlags, SubRegIdx)
+              .getInstr();
+      AdditionalFlags = 0;
+      LLVM_DEBUG(dbgs() << "        to " << *ImplDef);
+      LIS.InsertMachineInstrInMaps(*ImplDef);
+      RegistersToRepair.insert(DstReg);
```
andcarminati: I guess we end here because of some coalescing of edc registers, right?

llvm/lib/Target/AIE/AIESuperRegUtils.cpp
```diff
@@ -215,19 +224,31 @@ createSubRegisterVRegs(Register Reg, const SmallSet<int, 8> &SubRegs,
...
     MachineRegisterInfo &MRI, const AIEBaseRegisterInfo &TRI,
-    const TargetInstrInfo &TII, VirtRegMap &VRM) {
+    const TargetInstrInfo &TII, VirtRegMap &VRM, LiveIntervals &LIS) {
   for (MachineOperand &RegOp : make_early_inc_range(MRI.reg_operands(Reg))) {
     LLVM_DEBUG(dbgs() << printReg(RegOp.getReg(), &TRI, 0, &MRI)
                       << "  Changing " << *RegOp.getParent());
     int SubReg = RegOp.getSubReg();
     assert(SubReg);
+
+    bool MarkUseAsUndef = false;
```
andcarminati: Could we create a lambda for this? It would be nice to have a lazy evaluation here, like:

```
    if (RegOp.isDef()) {
      RegOp.setIsUndef(false);
    } else if (CheckUndef()) {
      RegOp.setIsUndef(true);
    }
```

llvm/lib/Target/AIE/AIESuperRegUtils.cpp
```diff
@@ -215,19 +224,31 @@ createSubRegisterVRegs(Register Reg, const SmallSet<int, 8> &SubRegs,
...
+    if (RegOp.isUse() && !RegOp.isUndef() && LIS.hasInterval(Reg)) {
+      SlotIndex UseIdx = LIS.getInstructionIndex(*RegOp.getParent());
+      LaneBitmask UseLaneMask = TRI.getSubRegIndexLaneMask(SubReg);
+      LaneBitmask LiveLanes = getLiveLanesAt(UseIdx, Reg, LIS);
+      if ((LiveLanes & UseLaneMask).none())
```
andcarminati: With the lambda, you could just return `(LiveLanes & UseLaneMask).none()`

andcarminati:
Is is possible to add the existing reproducer as a test?

llvm/lib/Target/AIE/AIESuperRegUtils.cpp
```diff
@@ -113,8 +113,17 @@ void rewriteFullCopy(MachineInstr &CopyMI, LiveIntervals &LIS,
   SmallSet<Register, 8> RegistersToRepair;
   for (int SubRegIdx : CopySubRegs) {
     if ((LiveSrcLanes & TRI.getSubRegIndexLaneMask(SubRegIdx)).none()) {
-      LLVM_DEBUG(dbgs() << "        Skip undefined subreg "
+      LLVM_DEBUG(dbgs() << "        IMPLICIT_DEF for undefined subreg "
                         << TRI.getSubRegIndexName(SubRegIdx) << "\n");
+      MachineInstr *ImplDef =
+          BuildMI(*CopyMI.getParent(), CopyMI, CopyMI.getDebugLoc(),
```
andcarminati: Could we refactor this `BuildMI` to use in bot cases, adding what is missing afterwards?

llvm/lib/Target/AIE/AIESuperRegUtils.cpp
SagarMaheshwari99: yes.

llvm/lib/Target/AIE/AIESuperRegUtils.cpp
SagarMaheshwari99: done..

llvm/lib/Target/AIE/AIESuperRegUtils.cpp
SagarMaheshwari99: done

SagarMaheshwari99:
> Is is possible to add the existing reproducer as a test?

Added a reduced test.

llvm/lib/Target/AIE/AIESuperRegUtils.cpp
SagarMaheshwari99: done.

andcarminati:
LGTM. I guess you have `undef` in the files. You can use `undefined` instead, like in
```
    // The full copy is decomposed into one MI per sub-register. Only the first
    // of those MIs gets the `undef` flag on its def, which tells the verifier
    // "this is a partial def of DstReg; the other sub-lanes are undefined
    // here". Subsequent MIs are additional partial defs of an
    // already-partially-live DstReg, so they must not carry `undef`.
```

This trigger alarm in the code checker.

andcarminati (approved):
LGTM. We need to see how to merge.

SagarMaheshwari99:
closing this, Konstantin will open a new PR, that identifies and solves a bug introduced before the super reg rewriter.. that would fix the domination errors.

PR #939, opened 2026-04-15
[AIEX] Postpipeliner Solver 

F-Stuckmann:
This PR shifts the Solver to prototyping Software and gives a best effort estimation without the need to have prior knowledge about the Problem.

Increase Solver capabilities by allowing 

- NS+1 search
- retry the search with known constraints
- have generic try Postpipeliner pragma instead of proving II


F-Stuckmann:
we should keep --aie-postpipeliner-target-ii

llvm/lib/Target/AIE/AIEPostPipeliner.cpp
```diff
@@ -1469,15 +1469,17 @@ bool PostPipeliner::tryApproaches() {
...
     }
-    if (NS == MinTripCount) {
-      // Only try this at the boundary case
-      if (solve(Data, NS + 1, true)) {
-        return true;
-      }
+    // Let's try SEF solution.
+    if (solve(Data, NS + 1, true)) {
+      return true;
+    }
```
martien-de-jong, line 1479: I would still consider MinTripCount. NS+1 is futile if MinIterCount is insufficient. SEF really only helps for the boundary case.

llvm/lib/Target/AIE/AIEInterBlockScheduling.cpp
```diff
@@ -618,7 +618,14 @@ SchedulingStage InterBlockScheduling::updateScheduling(BlockState &BS) {
...
     if (PostSWP.isPostPipelineCandidate(*BS.TheBlock)) {
-      BS.FixPoint.II = PostSWP.getResMII(*BS.TheBlock);
+      // A CLI --aie-postpipeliner-target-ii is a hard limit: start at
+      // exactly that II (bypassing --aie-postpipeliner-maxii) and let
+      // updatePipelining one-shot it. A pragma-driven TargetII is a soft
+      // hint: start at ResMII and iterate normally; the solver fallback at
+      // II == TargetII is handled inside the post-pipeliner.
+      BS.FixPoint.II = PostSWP.isTargetIIHardLimit()
+                           ? PostSWP.getTargetII()
+                           : PostSWP.getResMII(*BS.TheBlock);
```
martien-de-jong, line 628: I think we should simplify. When we drive an example from the command line, we want to say where we start, where we stop. Orthogonal to that, we want to say which approaches to enable. In my branch I have introduced a MinII CLI.

llvm/lib/Target/AIE/AIEInterBlockScheduling.cpp
```diff
@@ -632,11 +639,17 @@ SchedulingStage InterBlockScheduling::updatePipelining(BlockState &BS) {
...
+  // II, even if it exceeds --aie-postpipeliner-maxii. If that attempt
+  // failed, do not try any other II. A pragma-driven TargetII keeps the
+  // normal iteration (ResMII..MaxII).
+  if (!BS.getPostSWP().isTargetIIHardLimit()) {
+    // Otherwise try a larger II.
+    // We cut off at larger IIs to prevent excessive compilation time.
+    if (++BS.FixPoint.II <= PostPipelinerMaxII &&
+        ++BS.FixPoint.IITries <= PostPipelinerMaxTryII) {
+      return SchedulingStage::Pipelining;
+    }
```
martien-de-jong, line 652: So, a one-shot attempt can be build from MinII and MaxII and enabling the algorithms that you want to act on it.

llvm/lib/Target/AIE/AIEBaseInstrInfo.h
```diff
@@ -406,6 +406,10 @@ struct AIEBaseInstrInfo : public TargetInstrInfo {
     llvm_unreachable("Target didn't implement getNumReservedDelaySlots");
   }
 
+  /// Convert an InstrStage's getUnits() value to an FU bit position.
+  /// Default: identity, matching AIE's FUNCUNIT_REPRESENTATION(x) = (x).
+  virtual unsigned getFuncUnitIndex(uint64_t Units) const { return Units; }
```
martien-de-jong, line 411: Why is this virtual?

llvm/lib/Target/AIE/AIEPostPipeliner.cpp
F-Stuckmann: Adding the resource constraints allows us to fallback to the conservative application of the SEF. 
I will keep it

llvm/lib/Target/AIE/AIEPostPipeliner.cpp
```diff
@@ -1538,13 +1539,41 @@ bool PostPipeliner::solve(const SolverData &Data, int NS, bool SEFStage) {
...
+// Register \p MI's per-cycle FU footprint into \p Data so
+// SWPSolver::resourceConflicts can forbid same-cycle co-occupation.
+static void addResourceUses(SolverData &Data, int Id, const MachineInstr *MI,
+                            const AIEBaseInstrInfo *TII) {
+  const InstrItineraryData *Itin =
+      MI->getMF()->getSubtarget().getInstrItineraryData();
+  if (!Itin || Itin->isEmpty()) {
+    return;
+  }
+  const unsigned SchedClass = MI->getDesc().getSchedClass();
```
martien-de-jong, line 1551: No. TII->getSchedClass().

llvm/lib/Target/AIE/AIEPostPipeliner.cpp
```diff
@@ -1538,13 +1539,41 @@ bool PostPipeliner::solve(const SolverData &Data, int NS, bool SEFStage) {
...
+    assert(IS.getNextCycles() >= 0 &&
+           "Negative NextCycles breaks cumulative offset");
+    const unsigned FUIndex = TII->getFuncUnitIndex(IS.getUnits());
+    for (unsigned C = 0; C < IS.getCycles(); C++) {
+      const int Off = Cycle + C;
+      Data.addResourceUse(Id, Off, FUIndex, IsRequired);
+    }
+    Cycle += IS.getNextCycles();
+  }
+}
```
martien-de-jong, line 1568: And I'm not too happy about this parallel implementation of resource queries.
Perhaps we can pusblish anyStage() in AIEHazardRecogniser.

llvm/lib/Target/AIE/AIEPostPipeliner.cpp
```diff
@@ -1538,13 +1539,41 @@ bool PostPipeliner::solve(const SolverData &Data, int NS, bool SEFStage) {
...
+  }
+  const unsigned SchedClass = MI->getDesc().getSchedClass();
+  int Cycle = 0;
+  for (const InstrStage &IS : Itin->getStages(SchedClass)) {
+    const bool IsRequired = IS.getReservationKind() == InstrStage::Required;
+    const bool IsReserved = IS.getReservationKind() == InstrStage::Reserved;
+    assert(IsRequired != IsReserved &&
+           "ReservationKind must be exactly one of Required/Reserved");
+
+    assert(IS.getNextCycles() >= 0 &&
```
martien-de-jong, line 1559: getNextCycles is unsigned and asserts itself.

llvm/lib/Target/AIE/AIEPostPipeliner.cpp
F-Stuckmann: we use the current pattern all across the codebase. 
Should I be changing them all across the AIE target, when we have TII available?

llvm/lib/Target/AIE/AIESWPSolver.h
```diff
@@ -69,6 +70,15 @@ class Instruction {
...
 };
 
+// Per-cycle FU fingerprint of an instruction. BitVector (not uint64_t)
+// so the model scales past 64 FUs (AIE2PS has 106). See fuConflict().
+struct ResourceUseEntry {
+  int Instr;
+  int CycleOffset; // relative to the instruction's issue cycle
+  BitVector Required;
+  BitVector Reserved;
+};
```
martien-de-jong, line 80: You're reimplmenting one component of FuncUnitWrapper. anyStage() in AIEHazardRecognizer yields the entire FuncUnitWrapper, which has a conflict() (overlap()?) method.
Even more direct, in your instruction x instruction matrix, you could generate a full scoreboard for one instruction (II + pipelinedepth) and check each other instruction for the modulo distance where they conflict in this scoreboard.

F-Stuckmann: talked offline: we do not need worry too much about adding too many constraints since they are cheap. We can reuse the AIEHazardRecognizer and add any conflict()

PR #940, opened 2026-04-15
[AIEX] Fix empty unreachable blocks surviving to assembly.

SagarMaheshwari99:
BranchFolding failed to eliminate empty MBBs from unreachable IR blocks because: (1) it ran before DeadMachineInstructionElim stripped the dead instructions that made the block empty, and (2) even when it did encounter such blocks, it skipped them—empty blocks with no successors didn't match the `isSuccessor(FallThrough)` check, and the `FallThrough == MF.end()` case was an unimplemented TODO.

Fix BranchFolding to handle both cases, and add a BranchFolding run after DCE in the AIE2 pipeline so it can clean up blocks that only become empty after dead code elimination.

### DIFF
--- a/llvm/lib/CodeGen/BranchFolding.cpp
+++ b/llvm/lib/CodeGen/BranchFolding.cpp
@@ -4,6 +4,9 @@
 // See https://llvm.org/LICENSE.txt for license information.
 // SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
 //
+// Modifications (c) Copyright 2026 Advanced Micro Devices, Inc. or its
+// affiliates
+//
 //===----------------------------------------------------------------------===//
 //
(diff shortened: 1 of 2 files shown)


martien-de-jong (commented):
I guess it will allow linking, but I think the code will still be broken. The 'unreachable' that starts this chain of events is early in llvm code, after it has inferred that calling a function would definitely cause undefined behaviour. That unreachable leaks out through inlining.
Since the branch was definitely emitted, the block definitely existed entering the MCLayer. I would rather emit that empty unreachable block rather than remove the branch to it.

andcarminati:
We fixed the root case of the problem here: https://github.com/Xilinx/llvm-aie/pull/941. Should we close this?

SagarMaheshwari99:
> We fixed the root case of the problem here: #941. Should we close this?

yes

PR #941, opened 2026-04-15
[LoopIterCountAssumptions] Prevent assume(false) with inconsistent metadata

andcarminati:
Fix bug where inconsistent loop iteration count metadata would cause
insertion of @llvm.assume(i1 false), leading to unreachable instructions
in the entry block.

The issue occurred when the pass evaluated the loop condition at iteration
(N-1) where N is the metadata-specified iteration count. If this evaluation
constant-folded to false (e.g., evaluating "7 < 4" when metadata claims 8
iterations but the loop only runs 4 times), the pass would create
@llvm.assume(i1 false), which is semantically incorrect.


F-Stuckmann (commented):
Could we also have a check that we never insert a assume false?

martien-de-jong (approved):
LGTM

PR #943, opened 2026-04-15
[AIEX] Allow Loop Specific command line options

F-Stuckmann:
Allow Loop Specific Command line arguments.

We set up a `LoopOptionOverrides` for a loop and then querry a command line option to provide either the default or the overwrittten value.

I am unsure if strings are really needed.


andcarminati:
Maybe disconnect this PR from "command line options". **The mechanism is pretty generic. -> please, do not consider this comment.**

llvm/lib/Target/AIE/Utils/AIELoopOptionOverrides.h
```diff
@@ -0,0 +1,78 @@
...
+/// Usage:
+///   LoopOptionOverrides Overrides(MBB);
+///   bool Aggressive = Overrides.get(AggressiveReAlloc);
+///   RewriteMode Mode = Overrides.get(RegRewriteMode);
+class LoopOptionOverrides {
+  /// Prefix stripped from metadata keys to derive the cl::opt ArgStr.
+  static constexpr StringLiteral Prefix{"llvm.loop.hint."};
+
+  /// Integer-valued overrides (covers bool, int, unsigned, and enum types).
+  StringMap<int64_t> IntVals;
```
andcarminati, line 53: String alternative?

andcarminati: Additional `StringVals`?

F-Stuckmann: sorry that was in the last commit, i was unsure if it was needed and wanted to be able to cut it off if it was out of scope

llvm/lib/Target/AIE/Utils/AIELoopOptionOverrides.h
```diff
@@ -62,15 +67,21 @@ class LoopOptionOverrides {
...
+  /// enum, and std::string types.
   template <typename T> T get(const cl::opt<T> &Opt) const {
-    auto It = IntVals.find(Opt.ArgStr);
-    if (It != IntVals.end())
-      return static_cast<T>(It->second);
+    if constexpr (std::is_same_v<T, std::string>) {
+      auto It = StrVals.find(Opt.ArgStr);
+      if (It != StrVals.end())
+        return It->second;
+    } else {
```
andcarminati, line 86: A type check will not hurt.

llvm/lib/Target/AIE/Utils/AIELoopOptionOverrides.h
```diff
@@ -62,15 +67,21 @@ class LoopOptionOverrides {
 
   /// Return the effective value for a cl::opt. Looks up Opt.ArgStr in the
   /// overrides; falls back to the cl::opt value. Works for bool, integral,
-  /// and enum types (enums are encoded as i64 in metadata).
+  /// enum, and std::string types.
   template <typename T> T get(const cl::opt<T> &Opt) const {
-    auto It = IntVals.find(Opt.ArgStr);
-    if (It != IntVals.end())
-      return static_cast<T>(It->second);
+    if constexpr (std::is_same_v<T, std::string>) {
```
andcarminati, line 82: If the user pass the command line, we should honor it. You can use:
```
if (Opt.getNumOccurrences() > 0)
    return Opt;
```

llvm/lib/Target/AIE/Utils/AIELoopOptionOverrides.cpp
```diff
@@ -39,6 +39,10 @@ LoopOptionOverrides::LoopOptionOverrides(const MDNode *LoopID) {
       IntVals[OptName] = *Val;
       LLVM_DEBUG(dbgs() << "Loop override: " << OptName << " = " << *Val
                         << " (int)\n");
+    } else if (const auto Val = AIELoopUtils::getMetadataStringValue(*MD)) {
+      StrVals[OptName] = Val->str();
+      LLVM_DEBUG(dbgs() << "Loop override: " << OptName << " = " << *Val
+                        << " (string)\n");
     }
```
andcarminati, line 46: Please, add a TODO considering metadata without parameters (like your `no_predication`) example.

llvm/lib/Target/AIE/Utils/AIELoopOptionOverrides.h
F-Stuckmann: good idea

llvm/lib/Target/AIE/Utils/AIELoopOptionOverrides.cpp
F-Stuckmann: I fixed it in the frontend PR. loop hints will now always generate a value (true, just like cmd would)

andcarminati (approved):
Nice work! Please, squash the fixup commits before merging.

PR #944, opened 2026-04-15
[AIE2PS] Mark all aie2ps intrinsics as having default attributes

konstantinschwarz:
Without these, LLVM needs to be conservative for some intrinsics and assume they never return. Adding the default attributes enables more optimizations, similar to what was done for AIE2p.

### DIFF
--- a/llvm/include/llvm/IR/IntrinsicsAIE2PS.td
+++ b/llvm/include/llvm/IR/IntrinsicsAIE2PS.td
@@ -17,7 +17,7 @@ let TargetPrefix = "aie2ps" in {
 
 //===----------------------------------------------------------------------===//
 class AIE2PSEventIntrinsic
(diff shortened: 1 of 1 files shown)


F-Stuckmann:
How does the QoR look like?

konstantinschwarz:
Mostly flat. This was specifically targeting MX6 kernels, for which this is pretty effective (~6% improvement).
The regressions are in topk1 and topk2, in the fp16 variants. Looking at the code, I think there are a few missing peephole patterns for fp16 operations that can bring us back the lost performance.

konstantinschwarz:
@F-Stuckmann can we merge this? I can follow-up on the topk1 and topk2 regressions

F-Stuckmann (approved):
I like it

PR #948, opened 2026-04-16
[AIE2PS] Add missing caccfloat cascade intrinsic declarations

F-Stuckmann:
The ADF accessors.h header instantiates CASCADE_READ(caccfloat, 8/16) and CASCADE_WRITE(caccfloat, 8) for __AIE_ARCH__==22, expanding to calls to get_scd_v8caccfloat, get_scd_v16caccfloat, and put_mcd(v8caccfloat). These intrinsic declarations were present in the AIE2 compat header but missing from the AIE2PS compat header, causing compile errors.


clang/test/CodeGen/aie/aie2ps/aie2ps-caccfloat-cascade-compat.cpp
```diff
@@ -0,0 +1,43 @@
+//===- aie2ps-caccfloat-cascade-compat.cpp --------------------------------===//
```
SagarMaheshwari99, line 1: what purpose does this test serve? (just that the declarations are available? then probably not much..)

We don't normally include tests just for declarations present in compat header.

SagarMaheshwari99 (approved):

clang/test/CodeGen/aie/aie2ps/aie2ps-caccfloat-cascade-compat.cpp
F-Stuckmann: you are right, it doesn't provide additional information.

PR #952, opened 2026-04-17
[AIE2PS] A package of Conv2D improvements.

andcarminati:
This improves Conv2d. Some fine tuning is happening now.


llvm/lib/Target/AIE/AIECombinerHelper.cpp
```diff
@@ -1203,6 +1203,297 @@ void llvm::applyAddVecEltUndef(MachineInstr &MI, MachineRegisterInfo &MRI,
...
+
+  // Verify the POSTINC's pointer output has only one use (the PTR_ADD)
+  if (!MRI.hasOneNonDBGUse(PtrAddBase))
+    return false;
+
+  // POSTINC_LOAD has: def0 (data), def1 (pointer), use0 (base ptr), use1
+  // (offset) POSTINC_STORE has: def0 (pointer), use0 (data), use1 (base ptr),
+  // use2 (offset)
+  const unsigned PtrOutIdx = IsPostIncLoad ? 1 : 0;
+  const unsigned OffsetIdx = IsPostIncLoad ? 3 : 3;
```
andcarminati: check this.

llvm/lib/Target/AIE/AIECombinerHelper.cpp
```diff
@@ -1203,6 +1203,292 @@ void llvm::applyAddVecEltUndef(MachineInstr &MI, MachineRegisterInfo &MRI,
...
+  assert(MI.getOpcode() == TargetOpcode::G_PTR_ADD && "Expected G_PTR_ADD");
+
+  // This is the PTR_ADD that follows the POSTINC_LOAD
+  const Register PtrAddDst = MI.getOperand(0).getReg();
+  const Register PtrAddBase = MI.getOperand(1).getReg();
+  const Register PtrAddOffset = MI.getOperand(2).getReg();
+
+  LLVM_DEBUG(dbgs() << "Checking POSTINC_LOAD+PTR_ADD pattern for: " << MI);
+
+  // Get s32 source for PTR_ADD offset (only from TRUNC, not ZEXTLOAD)
```
andcarminati: The benefits of enabling zextload here will be checked after this PR.

llvm/lib/Target/AIE/aie2ps/AIE2PSLegalizerInfo.cpp
```diff
@@ -340,7 +340,10 @@ AIE2PSLegalizerInfo::AIE2PSLegalizerInfo(const AIE2PSSubtarget &ST)
           [=](const LegalityQuery &Query) {
             return std::pair(0, LLT::scalar(32));
           })
-      .legalFor({{S32, S32}, {P0, S32}})
+      // Pointer selects are not supported in hardware - convert to s32
```
F-Stuckmann: would this also be beneficial in other AIE architectures?

andcarminati: Definitely. We could also beckport it.

andcarminati:
QoR data:
image
image

llvm/lib/Target/AIE/aie2ps/AIE2PSLegalizerInfo.cpp
andcarminati: I left out to handle in a next PR.

F-Stuckmann: lets not forget about it

llvm/test/CodeGen/AIE/GlobalISel/legalize-select.mir
```diff
@@ -5,26 +5,59 @@
 # SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
 #
 # (c) Copyright 2023-2026 Advanced Micro Devices, Inc. or its affiliates
-# RUN: llc -mtriple aie -run-pass=legalizer %s -verify-machineinstrs -o - | FileCheck %s
-# RUN: llc -mtriple aie2 -run-pass=legalizer %s -verify-machineinstrs -o - | FileCheck %s
-# RUN: llc -mtriple aie2p -run-pass=legalizer %s -verify-machineinstrs -o - | FileCheck %s
-# RUN: llc -mtriple aie2ps -run-pass=legalizer %s -verify-machineinstrs -o - | FileCheck %s
+# RUN: llc -mtriple aie -run-pass=legalizer %s -verify-machineinstrs -o - | FileCheck --check-prefix=AIE %s
```
F-Stuckmann, line 8: why do we need all the different prefixes?

llvm/lib/Target/AIE/aie2ps/AIE2PSLegalizerInfo.cpp
andcarminati: We also need to test other opportunities related to the last combiner.

llvm/lib/Target/AIE/aie2ps/AIE2PSPostLegalizerFinalCombiner.cpp
```diff
@@ -0,0 +1,96 @@
...
+#undef GET_GICOMBINER_CONSTRUCTOR_INITS
+{
+}
+} // end anonymous namespace
+
+std::unique_ptr<Combiner> createAIE2PSPostLegalizerFinalCombinerImpl(
+    MachineFunction &MF, CombinerInfo &CInfo, const TargetPassConfig *TPC,
+    GISelKnownBits &KB, GISelCSEInfo *CSEInfo, const AIEBaseSubtarget &STI,
+    MachineDominatorTree *MDT, const LegalizerInfo *LI) {
+  static AIE2PSPostLegalizerFinalCombinerImplRuleConfig RuleConfig;
```
F-Stuckmann, line 86: sorry I wasn't able to compleltly delete the boilerplate code in the combiners :(

llvm/test/CodeGen/AIE/GlobalISel/legalize-select.mir
andcarminati: As we legalize differently for aie2ps, we need different lines. We could also unify the different ones, then we need to update again in the next backport. I see this as lengthy but easier to update after.

llvm/lib/Target/AIE/aie2ps/AIE2PSPostLegalizerFinalCombiner.cpp
andcarminati: No problem, when we see the opportunity (time) again we may simplify.

llvm/lib/Target/AIE/aie2ps/AIE2PSLegalizerInfo.cpp
F-Stuckmann: could you add a todo?

andcarminati: Done!

llvm/lib/Target/AIE/AIECombinerHelper.cpp
```diff
@@ -1203,6 +1203,168 @@ void llvm::applyAddVecEltUndef(MachineInstr &MI, MachineRegisterInfo &MRI,
...
+    const LLT SrcTy = MRI.getType(SrcReg);
+    if (SrcTy == LLT::scalar(32))
+      return SrcReg;
+  }
+
+  // Case 2: G_ZEXTLOAD (loads s16, zero-extends to s20)
+  if (DefMI->getOpcode() == TargetOpcode::G_ZEXTLOAD) {
+    if (!DefMI->memoperands_empty()) {
+      const MachineMemOperand *MMO = *DefMI->memoperands_begin();
+      if (MMO && MMO->getMemoryType() == LLT::scalar(16)) {
```
F-Stuckmann: why is this 16 bit?

llvm/lib/Target/AIE/AIECombinerHelper.cpp
```diff
@@ -1203,6 +1203,168 @@ void llvm::applyAddVecEltUndef(MachineInstr &MI, MachineRegisterInfo &MRI,
...
+  }
+
+  // Verify insertion point dominates the final use
+  if (!Helper.dominates(*InsertionPoint, MI))
+    return false;
+
+  // Build the transformation
+  MatchInfo = [=, &MRI, &MI](MachineIRBuilder &B) {
+    // Set insertion point right after the dominated definition
+    B.setInsertPt(*InsertionPoint->getParent(),
```
andcarminati: Fix the insertion point, this can be a phi node.

llvm/lib/Target/AIE/AIECombinerHelper.cpp
```diff
@@ -1365,6 +1370,125 @@ bool llvm::matchChainedPtrAddWithTruncOrZextLoad(MachineInstr &MI,
...
+  }
+
+  // Verify insertion point dominates the POSTINC
+  if (!Helper.dominates(*InsertionPoint, *PostIncMI))
+    return false;
+
+  // Build the lambda that will perform the transformation
+  MatchInfo = [=, &MRI](MachineIRBuilder &B) {
+    // Set insertion point right after the dominated definition
+    B.setInsertPt(*InsertionPoint->getParent(),
```
andcarminati: Fix this in case we have a phi node.

andcarminati:
In this version, we also take care of the insertion point to not place instructions between phi nodes.

llvm/lib/Target/AIE/AIECombinerHelper.cpp
andcarminati: This is related to the observed pattern. We may generalize after further testing.

llvm/lib/Target/AIE/AIECombinerHelper.cpp
```diff
@@ -1203,6 +1203,179 @@ void llvm::applyAddVecEltUndef(MachineInstr &MI, MachineRegisterInfo &MRI,
...
+/// 2. G_ZEXTLOAD of s16 -> returns the s20 register (already zero-extended)
+/// Returns std::nullopt if the pattern doesn't match
+static std::optional<Register> getS32SourceFromS20(Register S20Reg,
+                                                   MachineRegisterInfo &MRI) {
+  MachineInstr *DefMI = MRI.getVRegDef(S20Reg);
+  if (!DefMI)
+    return std::nullopt;
+
+  const LLT S20Ty = MRI.getType(S20Reg);
+  if (S20Ty != LLT::scalar(20))
```
F-Stuckmann: nit: I think we already have S20 nativly in the AIECombineHelper

llvm/lib/Target/AIE/AIECombinerHelper.cpp
```diff
@@ -1203,6 +1203,179 @@ void llvm::applyAddVecEltUndef(MachineInstr &MI, MachineRegisterInfo &MRI,
...
+    return std::nullopt;
+
+  // Case 1: G_TRUNC s32 -> s20
+  if (DefMI->getOpcode() == TargetOpcode::G_TRUNC) {
+    const Register SrcReg = DefMI->getOperand(1).getReg();
+    const LLT SrcTy = MRI.getType(SrcReg);
+    if (SrcTy == LLT::scalar(32))
```
F-Stuckmann: CHECK: wouldn't we be also capable of truncating anything larger than s20?

llvm/lib/Target/AIE/AIECombinerHelper.cpp
```diff
@@ -1203,6 +1203,179 @@ void llvm::applyAddVecEltUndef(MachineInstr &MI, MachineRegisterInfo &MRI,
...
+      return SrcReg;
+  }
+
+  // Case 2: G_ZEXTLOAD (loads s16, zero-extends to s20)
+  if (DefMI->getOpcode() == TargetOpcode::G_ZEXTLOAD) {
+    if (!DefMI->memoperands_empty()) {
+      const MachineMemOperand *MMO = *DefMI->memoperands_begin();
+      if (MMO && MMO->getMemoryType() == LLT::scalar(16)) {
```
F-Stuckmann: nit: why would we only limit ourselves to i16?

andcarminati: Please, see your comment a bit up.

llvm/lib/Target/AIE/AIECombinerHelper.cpp
andcarminati: Maybe yes, but we don't see a real case in our code base. Also, the truncation will be still there, and we are bypassing it and getting the src value directly.

llvm/lib/Target/AIE/AIECombinerHelper.cpp
andcarminati: Please, keep in mind that this PR addresses  specific patterns, any further generalization may come after with extensive testing.

llvm/lib/Target/AIE/AIECombinerHelper.cpp
F-Stuckmann: same here S32 is already a native type

llvm/lib/Target/AIE/AIECombinerHelper.cpp
F-Stuckmann: s16 is already a native type

llvm/lib/Target/AIE/AIECombinerHelper.cpp
```diff
@@ -1203,6 +1204,471 @@ void llvm::applyAddVecEltUndef(MachineInstr &MI, MachineRegisterInfo &MRI,
...
+  }
+
+  // Case 2: G_ZEXTLOAD (loads s16, zero-extends to s20)
+  // Skip this case if OnlyTruncs is true
+  if (!OnlyTruncs && DefMI->getOpcode() == TargetOpcode::G_ZEXTLOAD) {
+    if (!DefMI->memoperands_empty()) {
+      const MachineMemOperand *MMO = *DefMI->memoperands_begin();
+      if (MMO && MMO->getMemoryType() == LLT::scalar(16)) {
+        // The s20 value from ZEXTLOAD is already zero-extended from s16
+        // We can use it directly in s32 arithmetic (upper 12 bits are zero)
```
F-Stuckmann: `getS32SourceFromS20` returns an s20 register here, but the name promises s32. The caller then has to check the type and `G_ZEXT` to s32 separately (line ~1520). Instead, consider building the `G_ZEXT` inside this function (it would need a `MachineIRBuilder` parameter) so it always returns an s32 register, matching the function name and keeping the extension logic in one place.

llvm/test/CodeGen/AIE/GlobalISel/combine-chained-ptradd-trunc.mir
```diff
@@ -0,0 +1,235 @@
...
+    %0:_(p0) = COPY $p0
+    %1:_(p0) = COPY $p1
+    %2:_(s32) = COPY $r0
+    %3:_(s20) = G_ZEXTLOAD %1(p0) :: (load (s8))
+    %4:_(s20) = G_TRUNC %2(s32)
+    %5:_(p0) = G_PTR_ADD %0, %3(s20)
+    %6:_(p0) = G_PTR_ADD %5, %4(s20)
+    $p0 = COPY %6(p0)
+    PseudoRET implicit $lr, implicit $p0
+...
```
F-Stuckmann, line 235: nit: Could you add a negative test with the two PTR_ADDs in different basic blocks? The implementation rejects cross-block chains (AIECombinerHelper.cpp:1462) but there's no coverage for that path here.

llvm/lib/Target/AIE/AIECombinerHelper.cpp
```diff
@@ -1203,6 +1204,471 @@ void llvm::applyAddVecEltUndef(MachineInstr &MI, MachineRegisterInfo &MRI,
...
+  } else {
+    return false;
+  }
+
+  // Verify insertion point dominates the final use
+  if (!Helper.dominates(*InsertionPoint, MI))
+    return false;
+
+  // Build the transformation
+  MatchInfo = [=, &MRI, &MI](MachineIRBuilder &B) {
```
F-Stuckmann, line 1507: The insertion-point setup (PHI-skipping), G_ADD, and G_TRUNC construction are duplicated with the apply lambda in `matchPostIncLoadStorePtrAddWithTrunc` (~line 1641). Could you extract a shared helper, e.g., `buildCombinedS20Offset(MachineIRBuilder &B, MachineInstr *InsertionPoint, Register Src1, Register Src2, MRI)` that handles: set insertion point after the dominated def, optionally `G_ZEXT` s20 operands to s32, build `G_ADD`, build `G_TRUNC` to s20, and return the s20 result register?

F-Stuckmann (commented):
Review comments for the chained PTR_ADD combiner.

llvm/test/CodeGen/AIE/GlobalISel/legalize-select.mir
```diff
@@ -5,26 +5,59 @@
...
+# RUN: llc -mtriple aie2 -run-pass=legalizer %s -verify-machineinstrs -o - | FileCheck --check-prefix=AIE2 %s
+# RUN: llc -mtriple aie2p -run-pass=legalizer %s -verify-machineinstrs -o - | FileCheck --check-prefix=AIE2P %s
```
F-Stuckmann, line 10: I think we should implement these combiners for all architectures, not only aie2ps.
These changes are generic enough

llvm/lib/Target/AIE/AIECombinerHelper.cpp
```diff
@@ -1203,6 +1203,179 @@ void llvm::applyAddVecEltUndef(MachineInstr &MI, MachineRegisterInfo &MRI,
...
+  const Register SecondOffset = MI.getOperand(2).getReg();
+
+  // Get s32 source for second offset (from TRUNC or ZEXTLOAD)
+  auto SecondS32Opt = getS32SourceFromS20(SecondOffset, MRI);
+  if (!SecondS32Opt)
+    return false;
+  const Register SecondS32Reg = *SecondS32Opt;
+
+  // Check if base comes from another G_PTR_ADD
+  MachineInstr *FirstPtrAddMI = MRI.getVRegDef(SecondPtrAddBase);
```
F-Stuckmann, line 1467: I find these kinds of functions difficult to read, because it blurs the line between high level algorithm and implementation. Could we refactor it?

llvm/lib/Target/AIE/AIECombinerHelper.cpp
```diff
@@ -1203,6 +1203,179 @@ void llvm::applyAddVecEltUndef(MachineInstr &MI, MachineRegisterInfo &MRI,
...
+
+    const LLT S32 = LLT::scalar(32);
+    const LLT S20 = LLT::scalar(20);
+
+    // Handle the case where one or both values are s20 (from ZEXTLOAD)
+    // We need to extend them to s32 before adding
+    Register FirstS32Extended = FirstS32Reg;
+    Register SecondS32Extended = SecondS32Reg;
+
+    if (MRI.getType(FirstS32Reg) == S20) {
```
F-Stuckmann: nit: this should never happen, because the name suggests S32Reg. we could fix this by always returning S32 in getS32SourceFromS20

llvm/lib/Target/AIE/aie2ps/AIE2PSPostLegalizerFinalCombiner.cpp
```diff
@@ -0,0 +1,96 @@
...
+
+namespace {
+
+#define GET_GICOMBINER_TYPES
+#include "AIE2PSGenPostLegalizerGIFinalCombiner.inc"
+#undef GET_GICOMBINER_TYPES
+
+class AIE2PSPostLegalizerFinalCombinerImpl
+    : public AIECombinerBase<AIE2PSPostLegalizerFinalCombinerImplRuleConfig> {
+protected:
```
F-Stuckmann: Nit: empty `protected:` section -- looks like a copy-paste leftover from the custom combiner (which has protected members). Remove the stale `protected:` label.

llvm/lib/Target/AIE/AIEPostLegalizerFinalCombiner.cpp
```diff
@@ -0,0 +1,113 @@
...
+  else
+    return false; // No final combiner for other architectures
+
+  return Impl->combineMachineInstrs();
+}
+
+char AIEPostLegalizerFinalCombiner::ID = 0;
+INITIALIZE_PASS_BEGIN(AIEPostLegalizerFinalCombiner, DEBUG_TYPE,
+                      "AIE Post Legalizer Final Combiner", false, false)
+INITIALIZE_PASS_DEPENDENCY(TargetPassConfig)
```
F-Stuckmann, line 103: `MachineDominatorTreeWrapperPass` is listed in `getAnalysisUsage` as required (line 58) but missing from the `INITIALIZE_PASS_DEPENDENCY` list. Add `INITIALIZE_PASS_DEPENDENCY(MachineDominatorTreeWrapperPass)` here.

llvm/lib/Target/AIE/AIEPostLegalizerFinalCombiner.cpp
```diff
@@ -0,0 +1,113 @@
...
+  const auto &ST = MF.getSubtarget<AIEBaseSubtarget>();
+  const auto *LI = ST.getLegalizerInfo();
+  auto *KB = &getAnalysis<GISelKnownBitsAnalysis>().get(MF);
+  auto *MDT = &getAnalysis<MachineDominatorTreeWrapperPass>().getDomTree();
+
+  CombinerInfo CInfo(/*AllowIllegalOps*/ true, /*ShouldLegalizeIllegal*/ false,
+                     /*LegalizerInfo*/ nullptr, EnableOpt, F.hasOptSize(),
+                     F.hasMinSize());
+
+  const Triple &TT = ST.getTargetTriple();
```
F-Stuckmann, line 87: The triple check is unnecessary -- only `AIE2PSPassConfig::addPreRegBankSelect` adds this pass to the pipeline, so it will never run on a non-AIE2PS target. Simplify to a direct call to `createAIE2PSPostLegalizerFinalCombinerImpl` without the guard.

F-Stuckmann (commented):
Review comments.

F-Stuckmann:
Nit: commit `814c553` title says "postlegalizer custom combiner pass" but the pass is named "final combiner" everywhere in the code. Suggest: `[AIE2PS] Add a new postlegalizer final combiner pass`.

llvm/lib/Target/AIE/AIECombinerHelper.cpp
```diff
@@ -1203,6 +1204,471 @@ void llvm::applyAddVecEltUndef(MachineInstr &MI, MachineRegisterInfo &MRI,
...
+
+  // Get the definitions of both s32/s20 source registers
+  MachineInstr *FirstS32DefMI = MRI.getVRegDef(FirstS32Reg);
+  MachineInstr *SecondS32DefMI = MRI.getVRegDef(SecondS32Reg);
+
+  if (!FirstS32DefMI || !SecondS32DefMI)
+    return false;
+
+  // Check dominance: we need one to dominate the other
+  Register DominatingReg, DominatedReg;
```
F-Stuckmann: `DominatingReg` and `DominatedReg` are declared and assigned but never read. Only `InsertionPoint` is actually used from this dominance-check block. The lambda uses `FirstS32Reg` / `SecondS32Reg` directly (G_ADD is commutative so ordering is irrelevant). Compare with `matchPostIncLoadStorePtrAddWithTrunc` (line 1655) which does use them in its `buildAdd`. Remove the dead variables.

llvm/lib/Target/AIE/AIECombinerHelper.cpp
F-Stuckmann: "upper 12 bits are zero" is misleading at this point in the code. The value here is s20, not s32. `G_ZEXTLOAD s16 -> s20` zeros only the upper 4 bits (bits 16-19). The 12 zeroed bits (bits 20-31) only appear after the subsequent `G_ZEXT s20 -> s32` built by the caller. Suggest rephrasing to "upper 4 bits are zero (bits 16-19); the caller extends to s32".

llvm/lib/Target/AIE/AIECombine.td
```diff
@@ -352,6 +359,18 @@ def combine_split_concat_store : GICombineRule<
          [{ return matchSplitConcatStore(*${root}, MRI, (const AIEBaseInstrInfo &)B.getTII(), ${matchinfo}); }]),
   (apply [{ Helper.applyBuildFnNoErase(*${root}, ${matchinfo}); }])>;
 
+def combine_chained_ptradd_with_trunc : GICombineRule<
```
F-Stuckmann: Nit: The rule is named `combine_chained_ptradd_with_trunc` but the implementation (`matchChainedPtrAddWithTruncOrZextLoad`) now also handles G_ZEXTLOAD patterns. The C++ function name is correct but the .td rule name is stale. Consider `combine_chained_ptradd_with_s20_offsets` or similar to reflect the broader scope.

F-Stuckmann (commented):
Review comments.

llvm/lib/Target/AIE/AIECombinerHelper.cpp
```diff
@@ -1376,6 +1381,131 @@ bool llvm::matchChainedPtrAddWithTruncOrZextLoad(MachineInstr &MI,
...
+
+    // Build G_TRUNC to s20
+    const LLT S20 = LLT::scalar(20);
+    const Register CombinedS20 = MRI.createGenericVirtualRegister(S20);
+    B.buildTrunc(CombinedS20, CombinedS32);
+
+    // Update the POSTINC (LOAD or STORE) to use the combined offset and output
+    // to PtrAddDst Note: We don't use Observer here since it's not available in
+    // the lambda context
+    PostIncMI->getOperand(OffsetIdx).setReg(CombinedS20); // Update offset
```
F-Stuckmann, line 1658: CHECK: this should have been done by the global combiner, right?

andcarminati: I think no. We have a ptradd in between with a non con constant offset.

llvm/lib/Target/AIE/AIECombinerHelper.cpp
```diff
@@ -1203,6 +1204,471 @@ void llvm::applyAddVecEltUndef(MachineInstr &MI, MachineRegisterInfo &MRI,
...
```
F-Stuckmann, line 1658: **Missing GISelChangeObserver notification and transient SSA violation.**

The comment says the Observer is not available, but it is accessible via `B.getObserver()` (see `MachineIRBuilder.h:385`). Other combiners in this file use it (e.g. lines 1099, 1117, 1872).

Additionally, `applyBuildFn` erases the root G_PTR_ADD *after* this lambda returns, so between the `setReg(PtrAddDst)` here and the erase, `PtrAddDst` has two definitions (PostIncMI + the not-yet-erased G_PTR_ADD) -- a transient SSA violation.

Suggested fix: switch to `applyBuildFnNoErase` in the `.td`, erase MI inside the lambda *before* the `setReg` calls, and notify the observer:

```cpp
MI.eraseFromParent();

GISelChangeObserver *Observer = B.getObserver();
if (Observer)
  Observer->changingInstr(*PostIncMI);
PostIncMI->getOperand(OffsetIdx).setReg(CombinedS20);
PostIncMI->getOperand(PtrOutIdx).setReg(PtrAddDst);
if (Observer)
  Observer->changedInstr(*PostIncMI);
```

F-Stuckmann (commented):
Review comments.

llvm/lib/Target/AIE/AIECombinerHelper.cpp
```diff
@@ -1203,6 +1204,471 @@ void llvm::applyAddVecEltUndef(MachineInstr &MI, MachineRegisterInfo &MRI,
...
+                                                const AIEBaseInstrInfo &TII,
+                                                BuildFnTy &MatchInfo) {
+  assert(MI.getOpcode() == TargetOpcode::G_PTR_ADD && "Expected G_PTR_ADD");
+
+  // This is the PTR_ADD that follows the POSTINC_LOAD
+  const Register PtrAddDst = MI.getOperand(0).getReg();
+  const Register PtrAddBase = MI.getOperand(1).getReg();
+  const Register PtrAddOffset = MI.getOperand(2).getReg();
+
+  LLVM_DEBUG(dbgs() << "Checking POSTINC_LOAD+PTR_ADD pattern for: " << MI);
```
F-Stuckmann: Nit: The function handles both `G_AIE_POSTINC_LOAD` and `G_AIE_POSTINC_STORE`, but the debug message only mentions LOAD. Should be `"POSTINC_LOAD/STORE+PTR_ADD"`.

llvm/test/CodeGen/AIE/GlobalISel/combine-postincmem-ptradd-trunc.mir
```diff
@@ -0,0 +1,125 @@
...
+    %1:_(s32) = COPY $r0
+    %2:_(s32) = COPY $r1
+    %8:_(<16 x s32>) = COPY $x0
+    %3:_(s20) = G_TRUNC %1(s32)
+    %4:_(s20) = G_TRUNC %2(s32)
+    %5:_(p0) = G_AIE_POSTINC_STORE %8(<16 x s32>), %0, %3(s20) :: (store (<16 x s32>))
+    %6:_(p0) = G_PTR_ADD %5, %4(s20)
+    $p0 = COPY %6(p0)
+    PseudoRET implicit $lr, implicit $p0
+...
```
F-Stuckmann, line 157: The negative test for multiple uses (`test_postincload_ptradd_multiple_uses_should_not_combine`) only covers the LOAD variant. Consider adding a corresponding STORE test where the STORE's pointer output (`PtrOutIdx = 0`) has multiple uses, to verify the guard works for the different operand layout.

llvm/lib/Target/AIE/AIECombine.td
```diff
@@ -352,6 +359,18 @@ def combine_split_concat_store : GICombineRule<
...
+  (defs root:$root, build_fn_matchinfo:$matchinfo),
+  (match (wip_match_opcode G_PTR_ADD): $root,
+  [{ return matchChainedPtrAddWithTruncOrZextLoad(*${root}, MRI, Helper, ${matchinfo}); }]),
+  (apply [{ Helper.applyBuildFn(*${root}, ${matchinfo}); }])>;
+
+def combine_postincloadstore_ptradd_with_trunc : GICombineRule<
+  (defs root:$root, build_fn_matchinfo:$matchinfo),
+  (match (wip_match_opcode G_PTR_ADD): $root,
+  [{ return matchPostIncLoadStorePtrAddWithTrunc(*${root}, MRI, Helper, (const AIEBaseInstrInfo &)B.getTII(), ${matchinfo}); }]),
```
F-Stuckmann: Nit: C-style cast `(const AIEBaseInstrInfo &)B.getTII()`. `static_cast<const AIEBaseInstrInfo &>(B.getTII())` is more idiomatic. (This pattern is used throughout the file, so not blocking -- but worth fixing if touching this line.)

llvm/lib/Target/AIE/AIECombinerHelper.cpp
andcarminati: Now we have ;-)

llvm/lib/Target/AIE/AIECombinerHelper.cpp
andcarminati: The name is misleading and I corrected it. Also, we cannot create instructions in a match context - just in the apply phase. I updated the names/comments accordingly.

llvm/test/CodeGen/AIE/GlobalISel/combine-chained-ptradd-trunc.mir
andcarminati: Done!

llvm/test/CodeGen/AIE/GlobalISel/legalize-select.mir
andcarminati: Yes! We can do it in subsequent PR, this is quite big as is.

F-Stuckmann (commented):
Review comments.

llvm/lib/Target/AIE/AIECombine.td
```diff
@@ -409,7 +428,8 @@ def aie2p_additional_combines : GICombineGroup<[
   combine_unpad_unmerge,
   combine_pad_unpad_fusion,
   combine_concat_unpad_fusion,
-  combine_flatten_nested_concat
+  combine_flatten_nested_concat,
+  combine_chained_ptradd_with_trunc
```
F-Stuckmann: The match/apply logic is already covered by `combine-chained-ptradd-trunc.mir`. Since this commit's stated purpose is to enable opportunities for the Global Pointer Combiner downstream, would it be worth adding an `.ll` end-to-end test (aie2p and/or aie2ps) that exercises a chain of PTR_ADDs being collapsed pre-legalize and shows the resulting Global Pointer Combiner behavior? It would help lock in the intended interaction.

F-Stuckmann (commented):
Review comments.

llvm/lib/Target/AIE/AIEPostLegalizerFinalCombiner.cpp
andcarminati: This is the same in all other combiner passes.

llvm/lib/Target/AIE/AIEPostLegalizerFinalCombiner.cpp
andcarminati: We can keep as a preparation to enable for other targets.

llvm/lib/Target/AIE/AIECombine.td
F-Stuckmann: The match/apply logic is already covered by `combine-chained-ptradd-trunc.mir`. Since this commit's stated purpose is to enable opportunities for the Global Pointer Combiner downstream, would it be worth adding an `.ll` end-to-end test (aie2p and/or aie2ps) that exercises a chain of PTR_ADDs being collapsed pre-legalize and shows the resulting Global Pointer Combiner behavior? It would help lock in the intended interaction.

F-Stuckmann (commented):
Review comments.

llvm/lib/Target/AIE/AIECombine.td
andcarminati: I will keep because it is a very common indeed.

llvm/lib/Target/AIE/AIECombine.td
andcarminati: I can say that the behavior is always captured locally with the current combiner.

andcarminati: Big tests here will not lock the changes. If we see benefit of changing somethin here because some benchmark, we will change this test without fear. Smaller tests narrow the behavior and protect unintended changes because we really understand them.

llvm/lib/Target/AIE/AIECombinerHelper.cpp
andcarminati: The name was misleading.

andcarminati: I refactored a bit.

llvm/lib/Target/AIE/AIECombinerHelper.cpp
andcarminati: I changed the comment.

llvm/lib/Target/AIE/AIECombinerHelper.cpp
```diff
@@ -1203,6 +1205,467 @@ void llvm::applyAddVecEltUndef(MachineInstr &MI, MachineRegisterInfo &MRI,
...
+    // skipping any PHI nodes that immediately follow it.
+    setInsertPtAfterInstr(B, InsertionPoint);
+
+    // Build G_ADD of the two s32 values
+    const LLT S32 = LLT::scalar(32);
+    const Register CombinedS32 = MRI.createGenericVirtualRegister(S32);
+    B.buildAdd(CombinedS32, DominatingReg, DominatedReg);
+
+    // Build G_TRUNC to s20
+    const LLT S20 = LLT::scalar(20);
```
F-Stuckmann: nit: S20 is now a native type. no need to recreate it.

llvm/lib/Target/AIE/AIECombinerHelper.cpp
```diff
@@ -1203,6 +1205,467 @@ void llvm::applyAddVecEltUndef(MachineInstr &MI, MachineRegisterInfo &MRI,
...
+  if (!PostIncS32DefMI || !PtrAddS32DefMI)
+    return false;
+
+  // Check dominance: we need one to dominate the other
+  Register DominatingReg, DominatedReg;
+  MachineInstr *InsertionPoint = nullptr;
+
+  if (Helper.dominates(*PostIncS32DefMI, *PtrAddS32DefMI)) {
+    // PostInc s32 dominates PtrAdd s32
+    DominatingReg = PostIncS32Reg;
```
F-Stuckmann: nit: we duplicate this dominance check, we can reduce code duplication by having a helper function

llvm/lib/Target/AIE/AIECombinerHelper.cpp
```diff
@@ -1203,6 +1205,467 @@ void llvm::applyAddVecEltUndef(MachineInstr &MI, MachineRegisterInfo &MRI,
...
+    UnmergeMI->eraseFromParent();
+    BitcastMI->eraseFromParent();
+  };
+
+  return true;
+}
+
+/// Get an s32/s20 value from an s20 register that comes from either:
+/// 1. G_TRUNC of s32 -> returns the original s32 register
+/// 2. G_ZEXTLOAD of s16 -> returns the s20 register (already zero-extended)
```
F-Stuckmann, line 1370: nit: we always EXT the s20 register, we can as well return S32 here. If we don't use the EXT, DCE will remove it for us.

llvm/lib/Target/AIE/AIECombinerHelper.cpp
```diff
@@ -1203,6 +1204,170 @@ void llvm::applyAddVecEltUndef(MachineInstr &MI, MachineRegisterInfo &MRI,
...
+  // If we try to merge PADDs from different blocks, we may end-up de-hoisting
+  // PADDs as ADD inside loops.
+  if (MI.getParent() != FirstPtrAddMI->getParent())
+    return false;
+
+  const Register FirstPtrAddBase = FirstPtrAddMI->getOperand(1).getReg();
+  const Register FirstOffset = FirstPtrAddMI->getOperand(2).getReg();
+
+  // Get s32 source for first offset (from TRUNC or ZEXTLOAD)
+  auto FirstS32Opt = getSourceFromS20(FirstOffset, MRI);
```
F-Stuckmann: nit: we don't actually get s32 back. it could also be s20

llvm/lib/Target/AIE/AIECombinerHelper.cpp
```diff
@@ -1203,6 +1204,170 @@ void llvm::applyAddVecEltUndef(MachineInstr &MI, MachineRegisterInfo &MRI,
...
+
+  // Verify insertion point dominates the final use
+  if (!Helper.dominates(*InsertionPoint, MI))
+    return false;
+
+  // Build the transformation
+  MatchInfo = [=, &MRI, &MI](MachineIRBuilder &B) {
+    // Set insertion point right after the dominated definition
+    // Be careful to not insert between phi nodes.
+    MachineBasicBlock *InsertPtMBB = InsertionPoint->getParent();
```
F-Stuckmann: nit: this insertionPoint is duplicated. why is this not located in the find InsertionPoint mechanic starting in line 1315 and 1621

llvm/lib/Target/AIE/aie2p/AIE2PInstrInfo.cpp
```diff
@@ -1921,6 +1921,14 @@ unsigned AIE2PInstrInfo::getGenericBroadcastVectorOpcode() const {
   return AIE2P::G_AIE_BROADCAST_VECTOR;
 }
 
+unsigned AIE2PInstrInfo::getGenericPostIncLoadOpcode() const {
+  return AIE2P::G_AIE_POSTINC_LOAD;
```
F-Stuckmann, line 1925: another incentive to finally unify the opcode spaces.

llvm/lib/Target/AIE/AIECombinerHelper.cpp
```diff
@@ -1204,6 +1205,168 @@ void llvm::applyAddVecEltUndef(MachineInstr &MI, MachineRegisterInfo &MRI,
...
+/// \param OriginalID The intrinsic ID to check for splitting
+/// \return The split intrinsic ID if supported, std::nullopt otherwise
+///
+/// NOTE: This list may be extended in the future with additional intrinsics
+/// after proper benchmarking to ensure the split version provides performance
+/// benefits over the original wide intrinsic.
+static std::optional<Intrinsic::ID>
+getSplitIntrinsic(Intrinsic::ID OriginalID) {
+  switch (OriginalID) {
+  case Intrinsic::aie2ps_I512_v64_acc32_srs:
```
F-Stuckmann, line 1229: nit: This is target specific, it should be in TII

llvm/lib/Target/AIE/AIECombinerHelper.cpp
```diff
@@ -1204,6 +1205,168 @@ void llvm::applyAddVecEltUndef(MachineInstr &MI, MachineRegisterInfo &MRI,
...
+  // 6. Verify UNMERGE produces exactly 2 results
+  if (UnmergeMI->getNumDefs() != 2)
+    return false;
+
+  // 7. Get the two unmerge output registers
+  const Register LoReg = UnmergeMI->getOperand(0).getReg();
+  const Register HiReg = UnmergeMI->getOperand(1).getReg();
+
+  // 8. Verify both outputs are used by G_STORE
+  auto HasStoreUse = [&MRI](Register Reg) {
```
F-Stuckmann: nit: why is this not a single Non Debug use check? Besides the Opcode where is the difference between the checks 3, 5 and 8 ?

llvm/lib/Target/AIE/AIECombinerHelper.cpp
```diff
@@ -1204,6 +1205,168 @@ void llvm::applyAddVecEltUndef(MachineInstr &MI, MachineRegisterInfo &MRI,
...
+    return false;
+  };
+
+  if (!HasStoreUse(LoReg) || !HasStoreUse(HiReg))
+    return false;
+
+  // 9. Extract intrinsic operands (first operand after the intrinsic ID)
+  // For G_INTRINSIC_W_SIDE_EFFECTS: operand 0 = def, 1 = ID, 2+ = inputs
+  // For G_INTRINSIC: operand 0 = def, 1 = ID, 2+ = inputs
+  const Register AccReg = MI.getOperand(2).getReg();
```
F-Stuckmann, line 1315: This is only valid for the current aie2ps split intrinsics. I think we should bundle the Intrinsics and the OperandIndices.

llvm/lib/Target/AIE/AIECombinerHelper.cpp
```diff
@@ -1204,6 +1205,168 @@ void llvm::applyAddVecEltUndef(MachineInstr &MI, MachineRegisterInfo &MRI,
...
+  const Register ShiftReg = MI.getOperand(3).getReg();
+  const Register SignReg = MI.getOperand(4).getReg();
+
+  // 10. Derive types from the IR (no hardcoded types!)
+  const LLT OrigAccTy = MRI.getType(AccReg);
+  const LLT OrigIntrOutTy = MRI.getType(IntrinsicOutReg);
+
+  // Calculate split types by dividing by 2
+  const LLT AccHalfTy = OrigAccTy.divide(2);
```
F-Stuckmann, line 1324: nit: do we check if we have an even bitwidth?

llvm/lib/Target/AIE/AIECombinerHelper.cpp
```diff
@@ -1204,6 +1205,168 @@ void llvm::applyAddVecEltUndef(MachineInstr &MI, MachineRegisterInfo &MRI,
...
+  // clean up dead instructions.
+  MatchInfo = [=, &MI, &MRI](MachineIRBuilder &B) {
+    // Step 1: Unmerge the accumulator into two halves
+    const Register AccLoReg = MRI.createGenericVirtualRegister(AccHalfTy);
+    const Register AccHiReg = MRI.createGenericVirtualRegister(AccHalfTy);
+    B.buildUnmerge({AccLoReg, AccHiReg}, AccReg);
+
+    // Step 2: Create two split intrinsics using the ID from getSplitIntrinsic
+    const bool HasSideEffects =
+        (Opcode == TargetOpcode::G_INTRINSIC_W_SIDE_EFFECTS);
```
F-Stuckmann, line 1338: are the brackets here for formatting?

llvm/lib/Target/AIE/AIECombinerHelper.cpp
andcarminati: No, we should differentiate between target opcode and target instrinsic. Target opcodes may collide, so we use TII. Please, see mayMIShiftElements and buildWidenMulScalarAsVector functions.

llvm/lib/Target/AIE/AIECombinerHelper.cpp
andcarminati: There is a refactor as last commit.

llvm/lib/Target/AIE/AIECombinerHelper.cpp
andcarminati: Yes, this is valid of aie2ps and we may extend for a general use when the cases appear.

llvm/lib/Target/AIE/AIECombinerHelper.cpp
andcarminati: There is no need. We are handling valid user written code based on valid vector types. Optimizer will not generate aie instrinsic calls on random vector sizes.

llvm/lib/Target/AIE/AIECombinerHelper.cpp
andcarminati: I rewrote check number 8 to be consistent with the other checks.

llvm/lib/Target/AIE/AIECombinerHelper.cpp
andcarminati: I changed the type to just `Offset`.

llvm/lib/Target/AIE/AIECombinerHelper.cpp
andcarminati: The usage is sligthly different.

llvm/lib/Target/AIE/AIECombinerHelper.cpp
F-Stuckmann: the other checks only guarantee that the first user is our target one, not that we only have one user

F-Stuckmann (approved):
I think @katerynamuts is finally moving the global combiners as a standalone pass, so that the finalCombiner may not be necessary :)

PR #953, opened 2026-04-17
[AIEX] Do not allow global combiner to cross scheduling barriers.

### LINKED ISSUE #950: OffsetCombiner reorders loads across side-effecting instructions
The AIE Post-Legalizer Custom Combiner's `OffsetCombiner` hoists a `G_LOAD` past an `aie2.acquire` (or other side-effecting instruction) when an earlier load with the same base pointer exists. This breaks lock-protected memory accesses.

## Reproducer

The mlir-aie test `add_one_ctrl_packet` triggers this. After building the test:

```
cd mlir-aie/build/test/npu-xrt/add_one_ctrl_packet/peano/aie_arch.mlir.prj
llc -mtriple=aie2 -O2 -print-after=aie-postlegalizer-custom-combiner main_core_0_2.opt.ll
```

Look for `arrayidx46` / `arrayidx74` in the output: loop 3/4's loads of `input[3..7]` end up above the `acq` for lock 50.

The kernel runs four lock-gated inner loops, each adding 1 to its window of `input`. Expected output `7,8,9,10,11,12,13,14`; observed (after the hoist) `7,8,9,8,9,10,11,12` — loops 3 and 4 only see +2 instead of +4 because their loads are observed before loops 1 and 2's stores have landed.

Reproduced on NPU1 silicon.

## Root cause

`OffsetCombiner` inherits from `PointerModifierCombiner`. Its `isReorderCandidate` (`llvm/lib/Target/AIE/AIEGlobalCombinerPtrMods.cpp:397`) checks same input pointer, same MBB, post-inc partner, store-source dominance — but never checks whether intervening instructions impose memory ordering constraints.

When `isReorderCandidate` returns true, `AIEGlobalCombiner::reorderCombinerInsertions` calls `copyInsertionPoint` (`llvm/lib/Target/AIE/AIEGlobalCombiner.cpp:101`), which silently moves the OffsetCombiner's `InsertionPoint` to the `PostIncCombiner`'s position. The combined `G_AIE_OFFSET_LOAD` is then emitted at that earlier position regardless of what sits between.

Note that `OffsetCombiner::getInstructionsToMove` (`AIEGlobalCombinerPtrMods.cpp:328`) returns only the `G_PTR_ADD` in the move set, never the load itself — so the framework's existing `isMovable` walk over DAG Preds never validates the load's relocation either.

## Trace evidence

`-debug-only=global-combiner` shows the reorder being applied:

```
Reordering 
Offset[80] (%48:_,%114:_,) [1,1,1,] 
{Combiner Offset
  %48:_(p0) = nuw G_PTR_ADD %1:_, %47:_(s20)
  %114:_(s32) = G_LOAD %48:_(p0) :: (load from %ir.reverse.arrayidx46)
    Insertion Point: %72:_(s32) = G_LOAD %1:_(p0) :: (load from @input_buffer)
    to Move Up: %47:_(s20) = G_CONSTANT i20 12
}
```

Between `%72` (loop 2's first load) and `%114` (loop 3's load) sit loop 2's stores and the `G_INTRINSIC_W_SIDE_EFFECTS @llvm.aie2.acquire` for lock 50. The reorder traverses both.

## Proposed fix

In `OffsetCombiner::isReorderCandidate`, before returning true, walk the MBB between the two InsertionPoints and reject if any intervening instruction is a barrier per `TargetInstrInfo::isGlobalMemoryObject` (the same predicate `ScheduleDAGInstrs::buildEdges` uses to identify barrier candidates):

```cpp
const MachineInstr *NewIP = PostIncCombiner->CombinerData.InsertionPoint;
const MachineInstr *OldIP = CombinerData.InsertionPoint;
for (auto It = std::next(NewIP->getIterator());
     It != OldIP->getIterator(); ++It) {
  if (TII->isGlobalMemoryObject(&*It))
    return false;
}
```

Happy to send a PR.

F-Stuckmann:
This PR does not allow the global combiner to violate scheduling barriers. 

This PR fixes https://github.com/Xilinx/llvm-aie/issues/950


F-Stuckmann:
I have a hunch that we are missing that check somewhere else too.

andcarminati:
Nice, could you also provide some QoR data?

F-Stuckmann:
yeah, that's why it's not ready for review yet ;)

F-Stuckmann:
I was never this happy about a no-performance-improvement warning :) 


andcarminati (approved):
LGTM.

PR #954, opened 2026-04-17
[AIE2PS] Skip zero-bank instructions in bank-based slot materialization

F-Stuckmann:
When a multi-slot pseudo has no memory bank info (e.g. FIFO pops with unknown-size and no addrspace), skip it in the bank-based strategy instead of aborting the entire strategy. This allows instructions with valid addrspace annotations to be assigned slots based on their memory bank, while zero-bank instructions fall through to slot-pressure materialization.

This PR only improves performance with a pipeliner heuristic update (https://github.com/Xilinx/llvm-aie/pull/956).

I will be providing the combined QoR numbers.


F-Stuckmann:
Traditionally this indication of a Addrspace cast that we loose somewhere in the source code.

PR #959, opened 2026-04-21
[AIE] Expand also `uses` in expandCopyBundles

andcarminati:
We can still have cycles in copy bundles if we only expand defs.

### DIFF
--- a/llvm/test/CodeGen/AIE/aie2p/ra/unallocated-supereg-all-bundles.mir
+++ b/llvm/test/CodeGen/AIE/aie2p/ra/unallocated-supereg-all-bundles.mir
@@ -88,21 +88,19 @@ body:             |
   ; CHECK-NEXT:   [[COPY34:%[0-9]+]].sub_hi_dim_then_sub_dim_count:eds = COPY [[COPY33]].sub_hi_dim_then_sub_dim_count
   ; CHECK-NEXT:   [[COPY34:%[0-9]+]].sub_hi_dim_then_sub_dim_size:eds = COPY [[COPY33]].sub_hi_dim_then_sub_dim_size
   ; CHECK-NEXT:   [[COPY34:%[0-9]+]].sub_hi_dim_then_sub_dim_stride:eds = COPY [[COPY33]].sub_hi_dim_then_sub_dim_stride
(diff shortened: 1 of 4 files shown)


andcarminati:
This is a safe solution. In the future, we may strip out from the copy bundles only sub copies from or to full reg classes (no sub register).

llvm/test/CodeGen/AIE/aie2p/ra/unallocated-supereg-all-bundles.mir
```diff
@@ -88,21 +88,19 @@ body:             |
   ; CHECK-NEXT:   [[COPY34:%[0-9]+]].sub_hi_dim_then_sub_dim_count:eds = COPY [[COPY33]].sub_hi_dim_then_sub_dim_count
   ; CHECK-NEXT:   [[COPY34:%[0-9]+]].sub_hi_dim_then_sub_dim_size:eds = COPY [[COPY33]].sub_hi_dim_then_sub_dim_size
   ; CHECK-NEXT:   [[COPY34:%[0-9]+]].sub_hi_dim_then_sub_dim_stride:eds = COPY [[COPY33]].sub_hi_dim_then_sub_dim_stride
-  ; CHECK-NEXT:   undef [[COPY35:%[0-9]+]].sub_lo_dim:eds = COPY [[COPY34]].sub_lo_dim {
-  ; CHECK-NEXT:     internal [[COPY35]].sub_hi_dim_then_sub_dim_count:eds = COPY [[COPY34]].sub_hi_dim_then_sub_dim_count
-  ; CHECK-NEXT:     internal [[COPY35]].sub_hi_dim_then_sub_dim_size:eds = COPY [[COPY34]].sub_hi_dim_then_sub_dim_size
```
F-Stuckmann, line 93: do you know why we now drop internal?

F-Stuckmann (approved):
LGTM

llvm/test/CodeGen/AIE/aie2p/ra/unallocated-supereg-all-bundles.mir
andcarminati: We are unbundling those copies, so they are not internal anymore.

andcarminati:
No QoR regressions!

PR #961, opened 2026-04-22
[AIEX] Postpipeliner SideEffectFree Fix

F-Stuckmann:
Instructions that modify FIFO states are not side effect free. 

Fix this in the sideEffectFree check of the postpipeliner.


llvm/lib/Target/AIE/AIEPostPipeliner.cpp
```diff
@@ -183,6 +184,17 @@ bool isSideEffectFree(MachineInstr *MI) {
...
     return false;
   }
+
+  // FIFO operations modify persistent hardware state (the FIFO
+  // position register). Executing an extra copy corrupts the FIFO
+  // state and is thus not side effect free.
+  const auto &TRI = static_cast<const AIEBaseRegisterInfo &>(
+      *MI->getMF()->getSubtarget().getRegisterInfo());
+  for (const MachineOperand &Def : MI->defs()) {
+    if (Def.isReg() && TRI.isFifoPhysReg(Def.getReg()))
```
martien-de-jong: llvm::any_of?

martien-de-jong: To be refined. I don't think every instruction accessing the register performs a side effect. For now, for the SEF stage purpose, I think it is fine.

martien-de-jong (dismissed):
Approved as a hotfix. I think the instruction should have an implicit-def or something to represent the updated state.

llvm/lib/Target/AIE/AIEPostPipeliner.cpp
F-Stuckmann: This is actually not optimal. 
If we speculativly execute the fill and pop we can speculativly execute them.

F-Stuckmann: Turns out it was masking a greedy zero initialization issue of the fifo regs.

andcarminati (approved):
LGTM. This solves this problem and also other that are related to sub/super registers liveness.

PR #966, opened 2026-04-24
[AIEX] Warn and emit optimization remark for loads without memory ban…

F-Stuckmann:
…k in static materializer

When the static materializer encounters a multi-slot pseudo load without a memory bank annotation, emit both a WithColor::warning (always visible on stderr) and a MachineOptimizationRemarkMissed (available via -Rpass-missed=aie-multi-slot-pseudo). Both carry function name, block name, and debug location so front-end engineers can trace the issue back to source.


andcarminati:
Could we have a test?

andcarminati (approved):
LGTM.

PR #976, opened 2026-04-30
[AIE] ReservedRegsLICM: extend hoisting/sinking to nested loops and save/restore brackets

andcarminati:
This series extends the ReservedRegsLICM pass in three ways:

1. Nested loop support. The pass previously only processed the innermost
   loop. It now iterates over all loops in the nest (innermost first via
   reverse preorder), allowing loop-invariant control register assignments
   to be hoisted or sinked out of outer loops as well.

2. Correctness fix for multi-block loops. processForExitSink could
   incorrectly sink a register's last def to the exit block in loops where
   the register is defined in one conditional branch but not another. On
   the path that skips the def the register is live at the latch entry,
   carrying a value from a predecessor; sinking the def would expose the
   wrong value on that path. A LatchEntryLive pre-computation (walking the
   latch backward) guards sinking with !LatchEntryLive.contains(reg).

3. Save/restore bracket transformation. Some loops protect a reserved
   register with a save/restore bracket:

     vreg = COPY $reserved_reg    ; save pre-loop value
     $reserved_reg = MOVX imm     ; set loop-invariant value
     ... uses of $reserved_reg ...
     $reserved_reg = COPY vreg    ; restore pre-loop value

   The existing hoisting/sinking cannot handle this because the register
   is a loop livein (the save reads it) and has two defs (set + restore).
   The new findSaveRestoreBracket / processSaveRestoreBracket functions
   detect and transform the pattern as a unit: save+set are moved to the
   preheader (in order, so the save captures the pre-loop value before the
   set overwrites it) and the restore is sinked to the exit block. The loop
   body then always sees the loop-invariant value with no per-iteration
   save/restore overhead. Safety checks ensure no uses of the register
   appear before the save, after the restore, or in non-latch blocks of
   the loop.

Tests are added for all three cases, including a regression test that
verifies brackets with uses outside the [set, restore] window are
correctly rejected.

### DIFF
--- a/llvm/lib/Target/AIE/ReservedRegsLICM.cpp
+++ b/llvm/lib/Target/AIE/ReservedRegsLICM.cpp
@@ -341,36 +341,44 @@ static void moveInstruction(const CandidateInfo &Cand,
 
 void ReservedRegsLICM::processForExitSink(MachineLoop &L,
                                           const BitVector &ReservedLiveins) {
-  // Pre-compute what's live at the entry of the latch block by walking it
+  // Sink candidates must come from the exiting block (the block that has the
+  // edge to the exit block), not the latch. When the latch and the exiting
+  // block differ, instructions in the latch are not on the exit path: sinking
+  // them to the exit block would insert a def that was never executed on that
+  // path, corrupting the register's exit value.
+  MachineBasicBlock *ExitingBlock = L.getExitingBlock();
+  assert(ExitingBlock);
+
+  // Pre-compute what's live at the entry of the exiting block by walking it
   // fully backward. For multi-block loops with conditional paths, a candidate
   // register may be defined in some branches but not others, making it live at
-  // the latch entry on those paths. If the register is live at the latch entry
-  // (i.e., used before its first def in the latch), sinking its last def to
-  // the exit block would expose the wrong value on those paths.
-  assert(L.getLoopLatch());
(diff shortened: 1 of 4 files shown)


andcarminati:
QoR:
image

Avg reduction of:  -0.17%

konstantinschwarz:
I had a competing local implementation, so I asked AI to compare the two implementations.
It found two bugs in this implementation, I added the reproducers
[reserved-reg-licm-livein-bug.mir.txt](https://github.com/user-attachments/files/27374206/reserved-reg-licm-livein-bug.mir.txt)
[reserved-reg-licm-latch-vs-exiting-bug.mir.txt](https://github.com/user-attachments/files/27374207/reserved-reg-licm-latch-vs-exiting-bug.mir.txt)

andcarminati:
> I had a competing local implementation, so I asked AI to compare the two implementations. It found two bugs in this implementation, I added the reproducers [reserved-reg-licm-livein-bug.mir.txt](https://github.com/user-attachments/files/27374206/reserved-reg-licm-livein-bug.mir.txt) [reserved-reg-licm-latch-vs-exiting-bug.mir.txt](https://github.com/user-attachments/files/27374207/reserved-reg-licm-latch-vs-exiting-bug.mir.txt)

Hi, thank you very much for catching those cases. The implementation covers those cases now.

PR #979, opened 2026-04-30
[AIE] Fold brcond(xor(setcc, 1)) into PseudoJZ

konstantinschwarz:
The existing GlobalISel `matchNotCmp` combine tries to pull the negation into the icmp itself. However, if the icmp has other users, it bails out.
Here, we look for such remaining negation patterns, and pull them into the conditional branch itself


martien-de-jong (approved):
Pity we need the replication. How much of this could be shared if we would introduce universal register class aliases?

konstantinschwarz:
> Pity we need the replication. How much of this could be shared if we would introduce universal register class aliases?

The patterns are 100% identical. I was debating if I should add them to AIEBaseInstrPatterns.td, but that would force every architecture to use the same opcode names.

PR #981, opened 2026-05-04
[AIEX] Hazard Recognizer drops Load Store MemoryBank conflicts

F-Stuckmann:
I have a conflict() refactor in here. We previously had multiple different conflict locations and i unified them in this PR.

### DIFF
--- a/llvm/lib/Target/AIE/AIEHazardRecognizer.cpp
+++ b/llvm/lib/Target/AIE/AIEHazardRecognizer.cpp
@@ -745,7 +745,10 @@ unsigned AIEHazardRecognizer::computeScoreboardDepth() const {
 
 MemoryBankBits
 AIEHazardRecognizer::getMemoryBanks(const MachineInstr *MI) const {
-  if (!(MI->mayLoad() || MI->mayStore()))
+  // Bank tracking is load-only: stores cannot collide on banks because
+  // AIE has only one store slot per VLIW cycle, and load-vs-store uses separate
+  // HW ports.
+  if (!MI->mayLoad())
     return 0;
 
   if (MI->memoperands_empty())
--- a/llvm/unittests/Target/AIE/HazardRecognizerTest.cpp
+++ b/llvm/unittests/Target/AIE/HazardRecognizerTest.cpp
@@ -4,7 +4,7 @@
 // See https://llvm.org/LICENSE.txt for license information.
 // SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
 //
-// (c) Copyright 2023-2024 Advanced Micro Devices, Inc. or its affiliates
+// (c) Copyright 2023-2026 Advanced Micro Devices, Inc. or its affiliates
 //
 //===----------------------------------------------------------------------===//
 #include "AIE2.h"
--- a/llvm/unittests/Target/AIE/HazardRecognizerTest.cpp
+++ b/llvm/unittests/Target/AIE/HazardRecognizerTest.cpp
@@ -67,6 +67,13 @@ static const llvm::InstrStage MockStages[] = {
     {6, 2, 0, Req},
     {6, 3, 0, Req},
 
+    // Class 13: store-port resource 7 at cycle 4 (matches mem cycle 5)
+    {0, 16, 4, Req},
+    {1, 7, 1, Req},
+    // Class 14: store-port resource 7 at cycle 5 (matches mem cycle 6)
+    {0, 16, 5, Req},
+    {1, 7, 1, Req},
+
 };
 
 class MockItineraries : public InstrItineraryData {
@@ -90,12 +97,18 @@ public:
       return {MockStages + 6, 2};
     case 12:
       return {MockStages + 8, 4};
+    // Class 13: store-port resource at mem-access cycle 5.
+    case 13:
+      return {MockStages + 16, 2};
+    // Class 14: store-port resource at mem-access cycle 6.
+    case 14:
+      return {MockStages + 18, 2};
     default:
       return {MockStages, 1};
     }
   }
   bool isEndMarker(unsigned SchedClass) const override {
-    return SchedClass > 12;
+    return SchedClass > 14;
   }
   bool isEmpty() const override { return false; }
   std::optional<unsigned> getOperandCycle(unsigned SchedClass,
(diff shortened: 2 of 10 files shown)


andcarminati:
Could we verify if two stores may collide if they are one cycle apart? For example, we have stores accessing memory on cycle 5, 6 or even 7. If a 5-cycle store is scheduled just before a 6-cycle store, they will stall the memory interface. Could we still handle store banks separately?

F-Stuckmann:
I think we already handle this because already use the same resources.
Nonetheless i will add a unit-test to verify

llvm/lib/Target/AIE/AIEHazardRecognizer.cpp
```diff
@@ -125,22 +125,103 @@ FuncUnitWrapper &FuncUnitWrapper::operator|=(const FuncUnitWrapper &Other) {
...
+
+  if ((Slots & Other.Slots) != 0 || (Conflicts & Other.Slots) != 0 ||
+      (Slots & Other.Conflicts) != 0)
+    return ConflictKind::Slot;
+
+  if (Required.overlap(Other.Required))
+    return ConflictKind::FuncUnitRequired;
+
+  if (Reserved.overlap(Other.Required) || Required.overlap(Other.Reserved))
+    return ConflictKind::FuncUnitReserved;
```
martien-de-jong: nit: This only describes the first conflict, which implies an importance ordering. I would hold slot and resource unit conflicts more important than memory bank and memory object ones.

llvm/lib/Target/AIE/AIEHazardRecognizer.cpp
```diff
@@ -559,56 +640,73 @@ bool AIEHazardRecognizer::checkConflict(
...
+bool findConflict(const ResourceScoreboard<FuncUnitWrapper> &Scoreboard,
+                  const InstrItineraryData *ItinData, const MIResourceParams &P,
+                  int DeltaCycles, std::optional<int> FUDepthLimit,
+                  OnConflictT OnConflict) {
   // Verify format hazards
-  FuncUnitWrapper EmissionCycle(SlotSet, ConflictSet);
-  if (EmissionCycle.conflict(Scoreboard[DeltaCycles]))
+  FuncUnitWrapper EmissionCycle(P.SlotSet, P.ConflictSet);
+  if (EmissionCycle.conflict(Scoreboard[DeltaCycles])) {
+    OnConflict(EmissionCycle, Scoreboard[DeltaCycles]);
```
martien-de-jong: ScoreBoard[DeltaCycles] invokes a method, and may not be properly optimized away

llvm/lib/Target/AIE/AIEHazardRecognizer.cpp
```diff
@@ -559,56 +640,73 @@ bool AIEHazardRecognizer::checkConflict(
...
                      ThisCycle.dump(); dbgs() << "\n");
+          OnConflict(ThisCycle, Scoreboard[StageCycle]);
+          Found = true;
           return true;
         }
         return false;
-      };
+      },
+      FUDepthLimit);
+  return Found;
```
martien-de-jong: why not `return anyStage(...)` ?

llvm/lib/Target/AIE/AIEHazardRecognizer.cpp
```diff
@@ -97,13 +96,13 @@ void FuncUnitWrapper::clearResources() {
   Reserved.clear();
   Slots = 0;
   Conflicts = 0;
-  MemoryBanks = 0;
+  LoadBanks = 0;
```
martien-de-jong: I think this name change serves no purpose. They are still memory banks, even if the stores don't conflict on them.

llvm/lib/Target/AIE/AIEHazardRecognizer.cpp
```diff
@@ -534,7 +533,9 @@ void AIEHazardRecognizer::EmitInstruction(SUnit *SU, int DeltaCycles) {
     if (AIE::MachineBundle::isNoHazardMetaInstruction(SelectedOpcode))
       continue;
 
-    emitInScoreboard(TII->get(SelectedOpcode), getMemoryBanks(&BundledMI),
+    const MemoryBankBits LoadBI =
+        BundledMI.mayLoad() ? getMemoryBanks(&BundledMI) : 0;
```
martien-de-jong: getMemoryBanks is local, checking mayLoad and mayStore. Why not change that?

llvm/lib/Target/AIE/AIEPostPipeliner.cpp
```diff
@@ -1504,8 +1504,8 @@ SolverData PostPipeliner::createSolverData() {
     auto SlotKind = TII->getSlotKind(MI->getOpcode());
 
     const uint64_t MemoryBanks = HR.getMemoryBanks(MI);
```
martien-de-jong, line 1506: If we would have changed getMemoryBanks, this change would not have been necessary.

llvm/lib/Target/AIE/AIEHazardRecognizer.cpp
F-Stuckmann: I went with that

llvm/lib/Target/AIE/AIEHazardRecognizer.cpp
F-Stuckmann: I deleted the NFC refactor commit, it was just adding noise

llvm/unittests/Target/AIE/HazardRecognizerTest.cpp
```diff
@@ -615,3 +628,56 @@ TEST(HazardRecognizer, objectConflictHazard) {
...
+  EXPECT_TRUE(HR.hazard(14, -1, /*SlotSet=*/0b1, /*MemoryBanks=*/0,
+                        /*ObjectsBits=*/0, /*MemoryAccessCycle=*/{6}));
+
+  // Same staggered case, resource alone (no slot).
+  EXPECT_TRUE(HR.hazard(14, -1, /*SlotSet=*/0b0, /*MemoryBanks=*/0,
+                        /*ObjectsBits=*/0, /*MemoryAccessCycle=*/{6}));
+
+  // Same cycle without slot, resources don't overlap -> no hazard.
+  EXPECT_FALSE(HR.hazard(14, 0, /*SlotSet=*/0b0, /*MemoryBanks=*/0,
+                         /*ObjectsBits=*/0, /*MemoryAccessCycle=*/{6}));
```
martien-de-jong, line 682: As discussed also expect no hazard at -2.

martien-de-jong (approved):

PR #982, opened 2026-05-04
[AIEX] Refactor shuffle clang headers to remove branches.

F-Stuckmann:
This PR refactors the individual clang headers into a common aie common header. 

While doing so, any branches in the code selection from these extract/inserts are removed, since they are bad for performance.

Awaiting QoR.


llvm/lib/Target/AIE/Utils/AIEIRUtils.cpp
```diff
@@ -31,6 +31,49 @@ std::optional<Instruction *> instCombineDemandedBits(InstCombiner &IC,
...
+    return std::nullopt;
+  const unsigned LaneElems = LaneBits / ElemBits;
+  const unsigned NumElems = VecTy->getNumElements();
+  if (LaneElems == 0 || NumElems % LaneElems != 0)
+    return std::nullopt;
+  const unsigned NumLanes = NumElems / LaneElems;
+
+  // VEXTBCST encodes the index modulo the number of lanes (the upper bits of
+  // the 6-bit immediate field are documented as zero-extended).
+  const unsigned LaneIdx = IdxC->getZExtValue() & (NumLanes - 1);
```
martien-de-jong, line 63: Are we assuming powers of 2 here?

llvm/lib/Target/AIE/Utils/AIEIRUtils.cpp
```diff
@@ -31,6 +31,49 @@ std::optional<Instruction *> instCombineDemandedBits(InstCombiner &IC,
...
+
+  // Build a mask that broadcasts elements [Base, Base+LaneElems) across all
+  // NumLanes quadrants of the 512-bit result.
+  SmallVector<int, 64> Mask;
+  Mask.reserve(NumElems);
+  for (unsigned Lane = 0; Lane < NumLanes; ++Lane)
+    for (unsigned E = 0; E < LaneElems; ++E)
+      Mask.push_back(Base + E);
+
+  return new ShuffleVectorInst(Vec, Mask);
```
martien-de-jong, line 74: Long time ago I saw a bare new(). Shouldn't this go through some instruction factory?

llvm/test/CodeGen/AIE/opt/instcombine-aie2p-vextract-broadcast128.ll
```diff
@@ -18,8 +18,7 @@ declare <16 x i32> @llvm.aie2p.vextract.broadcast128.I512(<16 x i32>, i32)
 ; Constant idx 0 -- should fold to shufflevector picking lanes 0..3 of %v.
 define <4 x i32> @test_vextbcst128_const_idx0(<16 x i32> %v) {
 ; CHECK-LABEL: @test_vextbcst128_const_idx0(
-; CHECK-NEXT:    [[BCST:%.*]] = call <16 x i32> @llvm.aie2p.vextract.broadcast128.I512(<16 x i32> [[V:%.*]], i32 0)
-; CHECK-NEXT:    [[R:%.*]] = shufflevector <16 x i32> [[BCST]], <16 x i32> poison, <4 x i32> <i32 0, i32 1, i32 2, i32 3>
+; CHECK-NEXT:    [[R:%.*]] = shufflevector <16 x i32> [[BCST:%.*]], <16 x i32> poison, <4 x i32> <i32 0, i32 1, i32 2, i32 3>
```
martien-de-jong, line 21: Confusing use of the BCST regexp definition. I think it intends to match %v ?

martien-de-jong (commented):
Looks good in general. It's a real pity we have to reason our way through SimplifyCFG. It would be nice to have a few sample of .ll to .s examples that show the last few commits.

konstantinschwarz:
What is driving this PR? Did you see any issues in the final generated code that motivates this change?

F-Stuckmann:
The motivating example was Slice.

We had multiple branches inside of a single MBB loop, which disabled our regular loop handling and expanded the II from 2 to ~16

PR #985, opened 2026-05-04
[AIE2PS] Port AIE2P's modifyStoreFlush post-select optimization

konstantinschwarz:
This turns an II=5 to an II=2 loop in conv2d_mx6's conversion function


andcarminati (approved):
LGTM.

PR #987, opened 2026-05-05
[AIEX] Fix invalid remat after super-reg rewrite/unbundle

niwinanto:
Greedy register allocation can produce an invalid rematerialization after the AIE register-rewriter passes run, because those passes modify the IR in place but leave the LiveInterval of the VRM "Original" ancestor stale. `SplitKit::defFromParent` walks back through `VRM.getOriginal()` and uses that stale `OrigLI` to pick a remat candidate, which can clone an instruction whose semantics no longer match the destination's register class.


llvm/test/CodeGen/AIE/aie2p/ra/staged-ra-stale-remat.mir
```diff
@@ -98,23 +100,25 @@ body:             |
   ; CHECK-NEXT: {{  $}}
   ; CHECK-NEXT:   [[MOV_RLC_imm11_pseudo2:%[0-9]+]]:er = MOV_RLC_imm11_pseudo 1
   ; CHECK-NEXT:   [[AND1:%[0-9]+]]:er = AND [[COPY]], [[MOV_RLC_imm11_pseudo2]]
-  ; CHECK-NEXT:   [[MOV_PD_imm11_pseudo4:%[0-9]+]]:eds = MOV_PD_imm11_pseudo 128
```
niwinanto: This was an illegal copy that we have fixed now.

llvm/lib/Target/AIE/AIEUnallocatedSuperRegRewriter.cpp
```diff
@@ -229,13 +229,24 @@ bool AIEUnallocatedSuperRegRewriter::runOnMachineFunction(MachineFunction &MF) {
...
+
   LLVM_DEBUG(dbgs() << "Expanding copy bundles...\n");
   expandCopyBundles(Info.ExpandableRegs, MRI, Indexes, LIS, VRM, LRM);
 
   LLVM_DEBUG(dbgs() << "Performing register rewrites...\n");
   rewriteCandidates(Info.RewritableRegs, MRI, TRI, VRM, LRM, LIS, Indexes,
                     DebugVars);
 
+  // Prevent SplitKit from rematerializing through stale ancestor LIs.
+  LLVM_DEBUG(dbgs() << "Clearing stale split-from mappings...\n");
```
konstantinschwarz: nit: this will emit debug prints twice, once here, once in `clearStaleSplitFromMappings` itself.
Drop the print here?

konstantinschwarz (dismissed):

andcarminati:
Hi Niwin, could you please share the current QoR status of this fix?

llvm/test/CodeGen/AIE/aie2p/ra/staged-ra-stale-remat.mir
```diff
@@ -1,11 +1,11 @@
...
 # This file is licensed under the Apache License v2.0 with LLVM Exceptions.
 # See https://llvm.org/LICENSE.txt for license information.
 # SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
 #
 # (c) Copyright 2026 Advanced Micro Devices, Inc. or its affiliates
-
-# RUN: llc -mtriple=aie2p -start-before=greedy %s -o /dev/null --filetype=obj
-
+# RUN: : not --crash llc -mtriple=aie2p -start-before=greedy %s -o /dev/null --filetype=obj
+# RUN: llc -mtriple=aie2p -start-before=greedy -stop-before=virtregrewriter %s -o - | FileCheck %s
```
andcarminati, line 8: Check: with -stop-before=virtregrewrite we currently prevent the crash, right?

niwinanto: You are right, we do the wrong remat in greedy and crash during code emission. Reason I stopped before `virtregrewriter ` is to minimize the change to MIR(still use virtual registers) so that wrong remat is clear. Also, I kept the run line to reproduce the crash.

niwinanto:
@andcarminati 
Core_compute_Insn_Count for the changed kernels(AIE2PS)
```
Topk1D_int8_0,3077,3346,PASS,REGR(+8.74%)
Topk2D_int8_0,178758,177934,PASS,IMPR(-0.46%)
Conv2D_Transpose_fp16_1,4729,4595,PASS,IMPR(-2.83%)
Conv2D_Transpose_bf16_AIE2_1,4728,4594,PASS,IMPR(-2.83%)
Conv2D_Transpose_fp16_0,59339,56981,PASS,IMPR(-3.97%)
Conv2D_Transpose_bf16_AIE2_0,59333,56975,PASS,IMPR(-3.97%)
Average diff,+0.00%,-0.01%,,-0.01%
```
Topk1d tests are fragile, as their inputs change from run to run and they strongly affect the execution profile, generated code is literally same between 2 runs. And regarding the improvements it reports, RA has not really reclaimed all the regressions but it changed the allocation and that act as an enabler for the pipeliner and we have better code in the end.

On the other side, Core_PM regress for these kernels.
```
Conv2D_Transpose_bf16_AIE2_1,4800,5024,PASS,REGR(+4.67%)
Conv2D_Transpose_fp16_1,4800,5024,PASS,REGR(+4.67%)
Conv2D_Transpose_bf16_AIE2_0,6896,7120,PASS,REGR(+3.25%)
Conv2D_Transpose_fp16_0,6896,7120,PASS,REGR(+3.25%)
```
No change to Core_StackSize.

andcarminati (approved):
LGTM. Thanks for fixing this!

PR #681, opened 2025-10-17
[AIE2P] do not add defaultintrinsics for conv_fp32_bf16

F-Stuckmann:
[hotfix] This PR does not allow LICM to hoist Conv_fp32_bf16 outside of loops.
Keeping the conversion inside of the loops reduces some Stacks sizes.

PR #988, opened 2026-05-05
[AIE2P]/[AIE2PS] Mark float conversion intrinsics as having default attributes

khallouh:
Without these, LLVM needs to be conservative for some intrinsics and assume they never return. Adding the default attributes enables more optimizations.


konstantinschwarz:
@F-Stuckmann attempted this before but was running into unfortunate performance regressions: https://github.com/Xilinx/llvm-aie/pull/681
I still think this is the right change, but we need to mitigate the performance regressions before this can be merged

PR #989, opened 2026-05-05
[AIE2PS] Fix shouldCoalesce to block coalescing in both directions for ePSRFLdF

khallouh:
The shouldCoalesce hook was only blocking coalescing when a register was
being written INTO sub_ptr/sub_avail of ePSRFLdF (e.g., %25.sub_ptr = COPY %src).
However, it failed to block coalescing when a register was READING from
sub_ptr/sub_avail and the reader gets merged back into the composite
(e.g., %dst = COPY %25.sub_ptr). This second case causes the reader's
earlier definitions to become sub_ptr definitions, extending the live
range of the ePSRFLdF register back to the entry block.
Fix by checking both SubReg and DstSubReg to block coalescing in both
directions, preventing the artificial extension of the FIFO load state
register's live range.

### DIFF
--- a/llvm/lib/Target/AIE/aie2ps/AIE2PSRegisterInfo.cpp
+++ b/llvm/lib/Target/AIE/aie2ps/AIE2PSRegisterInfo.cpp
@@ -601,9 +601,12 @@ bool AIE2PSRegisterInfo::shouldCoalesce(
   // disjoint control-flow regions can share the same two physical registers
   // (plfr0, plfr1) without interference and without any spilling.
 
-  if (AIE2PS::ePSRFLdFRegClass.hasSubClassEq(NewRC) &&
-      SubReg != AIE2PS::NoSubRegister && SubReg != AIE2PS::sub_fifo) {
-    return false;
+  if (AIE2PS::ePSRFLdFRegClass.hasSubClassEq(NewRC)) {
+    // Block coalescing into sub_ptr or sub_avail from either direction
+    if ((SubReg != AIE2PS::NoSubRegister && SubReg != AIE2PS::sub_fifo) ||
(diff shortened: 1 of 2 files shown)


konstantinschwarz (approved):
Makes sense, thanks!

llvm/lib/Target/AIE/aie2ps/AIE2PSRegisterInfo.cpp
```diff
@@ -601,9 +601,12 @@ bool AIE2PSRegisterInfo::shouldCoalesce(
   // disjoint control-flow regions can share the same two physical registers
   // (plfr0, plfr1) without interference and without any spilling.
 
-  if (AIE2PS::ePSRFLdFRegClass.hasSubClassEq(NewRC) &&
-      SubReg != AIE2PS::NoSubRegister && SubReg != AIE2PS::sub_fifo) {
-    return false;
+  if (AIE2PS::ePSRFLdFRegClass.hasSubClassEq(NewRC)) {
```
andcarminati, line 604: I just noticed that this PR degrades Conv2D to:
```
.LBB13_12:                              // %for.body159.i
                                        //   Parent Loop BB13_11 Depth=1
                                        // =>  This Inner Loop Header: Depth=2
	vlda	 x2, [p3], #64;		nopb	;		nops	;		or	 r24, r19, r19;		vshuffle	x0, x6, x6, r5;		vmac	dm0, dm0, x0, x2, r8
	nopa	;		vldb.popx	 x6, [p0, lf0, r24];		nops	;		nopxm	;		nopv	
.L_LEnd11:
	vlda	 x4, [p3], #64;		nopb	;		padds.3d	 [p0], d1;		nopx	;		mov	r19, r24;		vmac	dm1, dm1, x0, x4, r8
```

Could we fix it ASAP? 
@khallouh 
@konstantinschwarz

PR #990, opened 2026-05-06
[AIE2PS] Revert shouldCoalesce due to performance degradation

andcarminati:
The reverted change seems to be good, but it exposed a clash with Conv2D. I am reverting it until we have a better understanding of the problem and a narrowed solution.

FYI @khallouh .


mludevid (approved):
LGTM. Reverts #989.

khallouh:
We need to make the hook smart enough to only block coalescing when we expect spills since that would be more expensive than the scalar moves causing the regression. We can come back to this later.

PR #991, opened 2026-05-06
[AIE2PS] Fix fifo pop pseudo to model srfifo_uf dependence

khallouh:
Baseline test shows wrong scheduling due to previously missing implicit def srfifo_uf register on fifo pop pseudos, which this PR adds.


andcarminati (approved):
LGTM.

PR #993, opened 2026-05-08
Reorder scheduling order of MBBs

F-Stuckmann:
Provide accurate scoreboard information for epilogues by scheduling its' prologues beforehand.

Awaiting QoR.

PR #994, opened 2026-05-08
[AIE] Improve AA when scheduling outer-loop pipelined epilogues

andcarminati:
This is in synergy with https://github.com/Xilinx/llvm-aie/pull/993 (may also conflict in the tests.)

We consider that we can only pipeline an outer loop when we have no memory-carried dependencies, we should flexible AA checks in the Post-RA. Sometimes, even with alias information, AA is very conservative, and it is better to relax AA just in the cases we trust.


llvm/lib/Target/AIE/Utils/AIELoopUtils.cpp
```diff
@@ -54,6 +54,15 @@ bool hasIIPragma(const MachineBasicBlock &LoopBlock) {
       .has_value();
 }
 
+bool isOuterLoopPipelined(const MachineBasicBlock &LoopLatch) {
+  // This metadata should be inserted by outerloop pipeliner.
+  // We just check the availability. No other pass is expected
+  // to insert this metadata.
+  return getLoopMetadata(getLoopID(LoopLatch),
+                         "llvm.loop.hint.aie_outerloop_pipeliner_success")
```
F-Stuckmann, line 62: check: we are using hints now for state tracking?

llvm/lib/Target/AIE/AIEInterBlockScheduling.h
```diff
@@ -207,6 +207,10 @@ class BlockState {
   // This holds an instance of the PostPipeliner for candidate loops.
   std::unique_ptr<PostPipeliner> PostSWP;
 
+  // This holds the information if this is an epilogue of a loop that was
+  // optimized by the outer-loop pipeliner.
+  bool IsEpilogueOfOuterPipelinedLoop = false;
```
F-Stuckmann, line 212: why is this not a BlockState?

llvm/lib/Target/AIE/AIEMachineScheduler.cpp
```diff
@@ -95,6 +95,12 @@ static cl::opt<bool> PreSchedFollowsSkipPipeliner(
     "aie-presched-follows-skip-pipeliner", cl::init(true),
     cl::desc("Don't run the prescheduler if the pipeliner is skipped"));
 
+static cl::opt<bool>
+    PostSchedIgnoreMemoryDeps("aie-postsched-ignore-memory-deps",
```
F-Stuckmann: nit: since we are only checking outer loop pipelined loops, could the argument reflect that? 

Reading this description would indicate that this also benefits the postpipeliner

F-Stuckmann (commented):
Is this intended for production use?
This seems like a hack to an AA shortcomings.
I think value tracking here would also work, similar to what we do in 2d/3d iteration AA checking

llvm/lib/Target/AIE/Utils/AIELoopUtils.cpp
andcarminati: Yeah, loop pipeliner is expected to replace the metadata. Also, we don't want to pipeline the same loop twice while testing.

llvm/lib/Target/AIE/AIEInterBlockScheduling.h
andcarminati: We have several places where we check for Epilogue and this this a loop/graph classification. The proposed flag means something different.

llvm/lib/Target/AIE/AIEMachineScheduler.cpp
andcarminati: This is a general flag used in `mayAlias` with a general meaning. I will mark this as hidden to avoid confusion.

andcarminati:
> Is this intended for production use?
> This seems like a hack to an AA shortcomings.
> I think value tracking here would also work, similar to what we do in 2d/3d iteration AA checking

This should be a contract between outer-pipeliner and its users. Outer pipelining should be explicitly requested by the developer. I also wish value tracking and 2d/3d iteration AA checking could be a silver bullet, but it is not, unfortunately.
I hope we could remove, soon. but it is not just value tracking...

llvm/lib/Target/AIE/Utils/AIELoopUtils.cpp
andcarminati: Also, see the PR merged yesterday: https://github.com/Xilinx/llvm-aie/pull/992 with that metadata already included.

llvm/lib/Target/AIE/AIEMachineScheduler.cpp
```diff
@@ -1571,6 +1577,20 @@ SUnit &AIEPostRASchedStrategy::addFixedSUnit(MachineInstr &MI, bool IsTop) {
...
 bool AIEScheduleDAGMI::mayAlias(SUnit *SUa, SUnit *SUb, bool UseTBAA) {
   BlockState &BS = getSchedImpl()->getInterBlock().getBlockState(getBB());
+
+  // We never skip AA check. Except for epilogues of outer-pipelined loops.
+  // This is a pre-condition of the optimization (sometimes restrict information
+  // may not help).
+  // TODO: also consider #pragma loop hints here. We should pre-load
+  // this information in the BlockState to avoid parsing metadata every
+  // call to this function (really heavy cost).
+  const bool SkipAACheck = PostSchedIgnoreMemoryDeps.getNumOccurrences() > 0
```
F-Stuckmann: If we want this to also hook up into loop hints, you need to access it with the LoopOptionOverride
```
// Build per-loop option overrides from !llvm.loop metadata.
Overrides = AIE::LoopOptionOverrides(*MBB);
Overrides.get(PostSchedIgnoreMemoryDeps)
```

F-Stuckmann:
Doesn't the postpipeliner already have this analysis information available when construction the 2 Iteration DAG?

llvm/lib/Target/AIE/Utils/AIELoopUtils.cpp
```diff
@@ -54,6 +54,15 @@ bool hasIIPragma(const MachineBasicBlock &LoopBlock) {
       .has_value();
 }
 
+bool isOuterLoopPipelined(const MachineBasicBlock &LoopLatch) {
+  // This metadata should be inserted by outerloop pipeliner.
+  // We just check the availability. No other pass is expected
+  // to insert this metadata.
+  return getLoopMetadata(getLoopID(LoopLatch),
```
F-Stuckmann, line 61: Check: This cannot live in a BlockState in Interblock because the outer loop pipelining is done outside of the MachineScheduler?

llvm/test/CodeGen/AIE/aie2ps/end-to-end/conv2d_int8_outerloop_pipelined.ll
F-Stuckmann, line 43: wait a moment is AA failing us because we do not have restricts?

llvm/lib/Target/AIE/AIEMachineScheduler.cpp
andcarminati: Please, see that we have a TODO here. Also according to existing comment, we should *NOT* call this here. `mayAlias` is a hot function called for every pair of memory instructions several times X the number of times we try a schedule in the PostSWP + normal scheduling. If we need to parse the metadata every call, it is just insane.

andcarminati:
> Doesn't the postpipeliner already have this analysis information available when construction the 2 Iteration DAG?

We have an analysis for single MBBs where we first probe that the pointers are the *same*. Different case.

llvm/test/CodeGen/AIE/aie2ps/end-to-end/conv2d_int8_outerloop_pipelined.ll
andcarminati: Not generally.

llvm/lib/Target/AIE/Utils/AIELoopUtils.cpp
andcarminati: The OuterPipeliner accepts to transform a loop (still in IR form) when the user requests this using a pragma (or command line). The pipeliner drops the original metadata and insert another one to inform that this loop was transformed. This blocks the same optimization twice and also can inform the rest of the compilation flow that this loop was in fact trasnformed.

andcarminati: So we annotate the loop to enable specific transformation/schedulig steps. For example, if we decide to split fixed regions, this will give us one idea where to apply.

llvm/lib/Target/AIE/AIEMachineScheduler.cpp
andcarminati: LoopOptionOverride > done!

llvm/lib/Target/AIE/AIEInterBlockScheduling.cpp
```diff
@@ -1172,11 +1178,21 @@ void BlockState::classify() {
...
   // We use the classification engine as a place to determine if this block
   // is the epilogue of an outerloop pipelined loop.
   IsEpilogueOfOuterPipelinedLoop =
       AIELoopUtils::isOuterLoopPipelined(*TheBlock);
+
+  // We never skip AA check. Except for epilogues of outer-pipelined loops.
+  // This is a pre-condition of the optimization (sometimes restrict information
+  // may not help).
+  auto Overrides = AIE::LoopOptionOverrides(*TheBlock);
+  IsSafeToIgnoreMemDeps = IsEpilogueOfOuterPipelinedLoop
```
F-Stuckmann, line 1193: CHECK: is it intended that we cannot enable the strict AA anymore once a loop is Outer Loop pipelined?

F-Stuckmann (approved):
I have a hunch we may not need this after all the dust has settled.

PR #996, opened 2026-05-12
[MLIR][TOSA] Make tosa const prop dialect agnostic

tr00:
Removes the restrictive `isa<ConstOp>` check such that we can match on constants from other dialects as well during ConstProp.
Adds a few test cases using the arith dialect (onnx is not registered in tosa) which demonstrate cross-dialect constant propagation.

### DIFF
--- a/mlir/test/Dialect/Tosa/constant-fold-constant-like.mlir
+++ b/mlir/test/Dialect/Tosa/constant-fold-constant-like.mlir
@@ -1,0 +1,74 @@
+// RUN: mlir-opt --split-input-file --tosa-layerwise-constant-fold %s | FileCheck %s
+
+// CHECK-LABEL: @reciprocal_fold_arith_const
+func.func @reciprocal_fold_arith_const() -> tensor<f32> {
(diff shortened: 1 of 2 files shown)


jorickert (approved):

gerion-amd (approved):

mlir/test/Dialect/Tosa/constant-fold-constant-like.mlir
```diff
@@ -0,0 +1,74 @@
+// RUN: mlir-opt --split-input-file --tosa-layerwise-constant-fold %s | FileCheck %s
+
+// CHECK-LABEL: @reciprocal_fold_arith_const
+func.func @reciprocal_fold_arith_const() -> tensor<f32> {
+  // CHECK: %[[RES:.*]] = "tosa.const"() <{value = dense<2.500000e-01> : tensor<f32>}> : () -> tensor<f32>
```
gerion-amd, line 5: Nit: Maybe switch "check lines" -> code with the other way around. This is more consistent with the other tests.

tr00: I looked this up in llvm and the other const prop tests also have the format: check-label, func, check lines, code...

PR #997, opened 2026-05-12
[AIE] Outer loop pipeliner

andcarminati:
This PR is a working in progress.


llvm/lib/Target/AIE/AIE2TargetMachine.cpp
```diff
@@ -78,8 +78,12 @@ TargetPassConfig *AIE2TargetMachine::createPassConfig(PassManagerBase &PM) {
 }
 
 bool AIE2PassConfig::addPreISel() {
-  if (TM->getOptLevel() != CodeGenOptLevel::None)
+  if (TM->getOptLevel() != CodeGenOptLevel::None) {
+    // Outer loop pipelining must run before hardware loop formation so that
+    // the pipelined loop structure is visible to the hardware loop pass.
     addPass(createHardwareLoopsLegacyPass());
+    addPass(createAIEOuterLoopPipelinerPass());
```
konstantinschwarz, line 83: Isn't this doing the exact opposite (running outer-loop pipelining *after* hardware loop pass?)

llvm/test/CodeGen/AIE/aie2p/outer-loop-pipelining-metadata.ll
```diff
@@ -0,0 +1,187 @@
...
+;
+; Precedence rules (from LoopOptionOverrides):
+;   1. Explicit command-line setting (highest)
+;   2. Per-loop metadata override
+;   3. cl::opt default value (lowest)
+;
+; Because the global flag is at its default (false), the metadata value is
+; authoritative for each loop independently.
+
+; RUN: llc -mtriple=aie2p -O2 -stop-after=aie-outer-loop-pipeliner \
```
konstantinschwarz: cosmetic: my eyes are trained to find the RUN lines towards the very top of the file. If possible, could you move them up?

llvm/test/CodeGen/AIE/aie2p/outer-loop-pipelining.ll
```diff
@@ -0,0 +1,325 @@
...
+;       br inner.header.cd
+;
+;   [inner.header.cd]        <- cloned inner loop
+;       br cooldown.exit
+;
+;   [cooldown.exit]          <- stores only (no loads)
+;       store result.cd
+;       br exit
+
+; RUN: llc -mtriple=aie2p -O2 -aie-enable-outer-loop-pipelining \
```
konstantinschwarz: Same here

llvm/lib/Target/AIE/AIEOuterLoopPipeliner.cpp
```diff
@@ -0,0 +1,1411 @@
...
+
+static cl::opt<bool> SplitPrologueAtConversion(
+    "aie-outer-loop-pipelining-split-prologue",
+    cl::desc("Pipeline only loads and their direct conversion intrinsics; "
+             "leave post-conversion instructions in the outer header"),
+    cl::init(true), cl::Hidden);
+
+static cl::opt<bool> EnableOuterLoopHardwareLoop(
+    "aie-outer-loop-hw-loop",
+    cl::desc("Convert downcounting outer loops to JNZD hardware loops after "
```
konstantinschwarz, line 87: The HardwareLoop pass can create JNZD loops for us, but it is disabled by default (and currently not possible to enable due to early bail-out). Should we leave this decision to the existing target hooks (`isHardwareLoopProfitable`)

llvm/lib/Target/AIE/AIEOuterLoopPipeliner.cpp
```diff
@@ -0,0 +1,1411 @@
...
+                                      Value *AdjustedTripCount);
+
+  // Returns true if the outer latch has a downcounting icmp pattern that can
+  // be replaced by @llvm.loop.decrement.reg.
+  bool isOuterLoopDowncounting(const LoopStructure &LS) const;
+
+  // Returns true if I is a CallInst whose return type is a 2048-bit vector.
+  // These are the "anchor" instructions that define the split point between
+  // Part 1 (pipelined) and Part 2 (stays in outer.header + cooldown.entry).
+  static bool produces2048BitVector(const Instruction *I);
```
konstantinschwarz: I could imagine we'll end up with multiple valid strategies for finding good split points. Should we prepare for this by introducing a Strategy class that provides a common interface (e.g. `collectPrologueInstructionsForSplit`) and hides all the other helpers?

llvm/lib/Target/AIE/AIEOuterLoopPipeliner.cpp
```diff
@@ -0,0 +1,1411 @@
...
+             L->getHeader()->printAsOperand(dbgs(), false); dbgs() << "\n");
+  // Build per-loop option overrides from !llvm.loop.hint.* metadata.
+  AIE::LoopOptionOverrides Overrides(L->getLoopID());
+  // Skip unless globally enabled OR the loop explicitly opts in via metadata.
+  if (!Overrides.get(EnableOuterLoopPipelining)) {
+    LLVM_DEBUG(dbgs() << "    Outer loop pipelining not enabled\n");
+    return false;
+  }
+  LoopStructure LS;
+  if (!analyzeLoopStructure(L, LS) || !isProfitableToRotate(LS, Overrides) ||
```
konstantinschwarz: nit: for easier debugging, put each condition into its own `if`?

llvm/lib/Target/AIE/AIEOuterLoopPipeliner.cpp
```diff
@@ -0,0 +1,1411 @@
...
+    if (BB->getTerminator()->getNumSuccessors() > 1) {
+      LLVM_DEBUG(dbgs() << "    Complex control flow in prologue\n");
+      return false;
+    }
+  for (BasicBlock *BB : LS.EpilogueBlocks)
+    if (BB->getTerminator()->getNumSuccessors() > 1) {
+      LLVM_DEBUG(dbgs() << "    Complex control flow in epilogue\n");
+      return false;
+    }
+  return true;
```
konstantinschwarz: Is there an expected `LoopStructure` state that we could assert here?

llvm/lib/Target/AIE/AIEOuterLoopPipeliner.cpp
```diff
@@ -0,0 +1,1411 @@
...
+  collectEpilogueStores(LS, EpilogueStores);
+  if (EpilogueStores.empty()) {
+    LLVM_DEBUG(dbgs() << "    No stores in epilogue\n");
+    return false;
+  }
+  return true;
+}
+
+bool AIEOuterLoopPipeliner::isSafeToReorderMemoryOps(const LoopStructure &LS) {
+  // TODO: Add alias/dependence analysis to verify that moving prologue loads
```
konstantinschwarz, line 515: Aren't we moving prologue loads *after* epilogue stores?

llvm/lib/Target/AIE/AIEOuterLoopPipeliner.cpp
```diff
@@ -0,0 +1,1411 @@
...
+
+  LLVM_DEBUG(dbgs() << "    Split-prologue: " << Out.size()
+                    << " Part-2 instructions (stay in outer.header + "
+                       "cooldown.entry)\n");
+}
+
+bool AIEOuterLoopPipeliner::performTransformation(
+    LoopStructure &LS, const AIE::LoopOptionOverrides &Overrides) {
+  BasicBlock *Preheader = LS.OuterLoop->getLoopPreheader();
+  if (!Preheader)
```
konstantinschwarz: Haven't we checked this before? See also comment about assertable state after loop analysis

llvm/lib/Target/AIE/AIEOuterLoopPipeliner.cpp
```diff
@@ -0,0 +1,1411 @@
...
+    return false;
+
+  // Step 1: Collect the data-load chain instructions from the outer header.
+  // Hardware-loop setup calls (set.loop.iterations) are excluded.
+  // In split-prologue mode, only loads + address chains + ups conversions are
+  // collected; post-conversion instructions remain in the outer header.
+  SmallVector<Instruction *, 16> PInsts;
+  if (Overrides.get(SplitPrologueAtConversion)) {
+    if (!collectPrologueInstructionsForSplit(LS, PInsts)) {
+      LLVM_DEBUG(dbgs() << "    Split-prologue: no load→conversion pairs, "
```
konstantinschwarz: nit: avoid Unicode in debug prints

llvm/lib/Target/AIE/AIEOuterLoopPipeliner.cpp
```diff
@@ -0,0 +1,1411 @@
...
+  SmallPtrSet<Instruction *, 32> Visited;
+  SmallVector<Instruction *, 16> Worklist;
+  auto Seed = [&](Value *V) {
+    auto *I = dyn_cast<Instruction>(V);
+    if (!I || isa<PHINode>(I) || I->getParent() != LS.OuterHeader)
+      return;
+    if (Visited.insert(I).second)
+      Worklist.push_back(I);
+  };
+  // Seed from values used inside the inner loop.
```
konstantinschwarz, line 552: It's not clear to me what `Seed`'ing means in this context

llvm/lib/Target/AIE/AIEOuterLoopPipeliner.cpp
```diff
@@ -0,0 +1,1411 @@
...
+  bool isProfitableToRotate(const LoopStructure &LS,
+                            const AIE::LoopOptionOverrides &Overrides);
+  bool isSafeToReorderMemoryOps(const LoopStructure &LS);
+  void collectPrologueLoads(const LoopStructure &LS,
+                            SmallVectorImpl<LoadInst *> &Loads) const;
+  void collectEpilogueStores(const LoopStructure &LS,
+                             SmallVectorImpl<StoreInst *> &Stores) const;
+  // Return the single epilogue block (outer latch or the last epilogue block).
+  BasicBlock *getEpilogueBlock(const LoopStructure &LS) const {
+    return LS.EpilogueInLatch ? LS.OuterLatch : LS.EpilogueBlocks.back();
```
konstantinschwarz: Again, would be nice if we could assert the valid / expected states after analysis

llvm/lib/Target/AIE/AIEOuterLoopPipeliner.cpp
```diff
@@ -0,0 +1,1411 @@
...
+                            BasicBlock *FromBlock) const;
+  // Collect the data-load chain instructions from the outer header that feed
+  // the inner loop. Does NOT include hardware-loop setup calls
+  // (@llvm.set.loop.iterations) — those stay in the outer header.
+  void collectPrologueInstructions(const LoopStructure &LS,
+                                   SmallVectorImpl<Instruction *> &Out) const;
+  bool performTransformation(LoopStructure &LS,
+                             const AIE::LoopOptionOverrides &Overrides);
+
+  // Step 1: Clone data-load chain into a warm-up block before the outer loop.
```
konstantinschwarz: Not sure how useful these `Step X:` comments are.
e.g., `clonePrologueAsWarmUp` is `Step 1` here, but `Step 3` on the caller side

llvm/lib/Target/AIE/AIEOuterLoopPipeliner.cpp
```diff
@@ -0,0 +1,1411 @@
...
+// Replace all uses of I inside the inner loop with phi_I, then erase I.
+void AIEOuterLoopPipeliner::createPipelinedPHIs(
+    const LoopStructure &LS, BasicBlock *WarmUp,
+    const SmallVectorImpl<Instruction *> &PInsts,
+    const ValueToValueMapTy &WarmUpVMap, const ValueToValueMapTy &EpiVMap) {
+  Instruction *InsertPt = &*LS.OuterHeader->getFirstInsertionPt();
+
+  SmallVector<std::pair<Instruction *, PHINode *>, 8> Replacements;
+  for (Instruction *I : PInsts) {
+    if (I->getType()->isVoidTy())
```
konstantinschwarz, line 673: This looks strange. Why the check for void type?

llvm/lib/Target/AIE/AIEOuterLoopPipeliner.cpp
```diff
@@ -0,0 +1,1411 @@
...
+      continue;
+    auto WIt = WarmUpVMap.find(I);
+    auto EIt = EpiVMap.find(I);
+    if (WIt == WarmUpVMap.end() || EIt == EpiVMap.end())
```
konstantinschwarz: What does it mean if one of the maps doesn't have a mapping, while the other does? Should this be an assert?

llvm/test/CodeGen/AIE/aie2p/outer-loop-pipelining.ll
```diff
@@ -0,0 +1,325 @@
...
+  %cmp.outer = icmp sgt i32 %N, 1
+  br i1 %cmp.outer, label %outer.header, label %exit
+
+outer.header:
+  %i = phi i32 [ 0, %entry ], [ %i.next, %outer.latch ]
+  %a.ptr = phi ptr [ %a, %entry ], [ %a.ptr.next, %outer.latch ]
+  %b.ptr = phi ptr [ %b, %entry ], [ %b.ptr.next, %outer.latch ]
+  %c.ptr = phi ptr [ %c, %entry ], [ %c.ptr.next, %outer.latch ]
+  ; Prologue: loads for the inner loop
+  %v0 = load i32, ptr %a.ptr, align 4
```
konstantinschwarz, line 111: Can you add a test case where the load does not directly use the `%a.ptr`, but has a GEP in between?

konstantinschwarz (commented):
Nice work, the overall structure makes sense to me and is easy to follow

llvm/lib/Target/AIE/AIE2TargetMachine.cpp
andcarminati: The comment is part of a previous approach, I need to change it. It helps with the loop structure.

llvm/lib/Target/AIE/AIEOuterLoopPipeliner.cpp
andcarminati: Now this is a standalone pass, but we may transfer the responsibility to save a bit of code. I tried to isolate the first implementation as much as possible

llvm/lib/Target/AIE/AIEOuterLoopPipeliner.cpp
andcarminati: You discovered my idea ;-). I am planning to have a set of strategies, maybe each target can install different ones.

andcarminati:
In the newer version, we have a better pointer handling. We avoid reversing the loads and the pointer updates, which can create a mess for the pointer combiner.

llvm/test/CodeGen/AIE/aie2p/outer-loop-pipelining-ptr-lifting.ll
```diff
@@ -0,0 +1,170 @@
+; NOTE: Assertions have been autogenerated by utils/update_llc_test_checks.py UTC_ARGS: --version 5
```
konstantinschwarz: nit: CHECK lines were modified after initial generation, should drop the comment

llvm/lib/Target/AIE/AIEOuterLoopPipeliner.cpp
```diff
@@ -0,0 +1,1557 @@
...
+        }
+        // Operands from outer header (PHIs) or preheader are OK - they're
+        // already available before the inner loop.
+      }
+    }
+
+    // Additional check: if any instruction in the chain has uses by other
+    // epilogue instructions (outside the chain), we cannot lift this chain.
+    // This excludes chains used by stores, loop control (icmp), etc.
+    if (CanLift) {
```
konstantinschwarz, line 1171: I'm trying to come up with examples to break this.
e.g., thinking about a case where an epilogue instruction has no other uses within the epilogue, but it has other unmodelled side-effects (e.g. reads global registers written within the epilogue).
I don't think that is a typical case for pointer updates, but should we be conservative and forbid lifting instructions with side-effects?

llvm/lib/Target/AIE/AIEOuterLoopPipeliner.cpp
andcarminati: Void-typed instructions (stores, side-effect-only intrinsics) don't  produce values, so there's nothing to merge via a PHI node. Each execution path simply runs its own cloned copy independently. I included a comment.

llvm/lib/Target/AIE/AIEOuterLoopPipeliner.cpp
andcarminati: I cleaned the comments.

llvm/lib/Target/AIE/AIEOuterLoopPipeliner.cpp
andcarminati: To be safe here, I gated intrinsics to aie pointer adds.

andcarminati:
Hi @konstantinschwarz, thank you for taking time to review. I addressed all your comments as new commits to keep the full story/

llvm/test/CodeGen/AIE/aie2ps/end-to-end/conv2d_opt_outerloop_blocks.ll
```diff
@@ -0,0 +1,346 @@
...
+declare { ptr, i20 } @llvm.aie2ps.add.2d(ptr, i20, i20, i20, i20) #0
+
+; Function Attrs: nocallback noduplicate nofree nosync nounwind willreturn
+declare void @llvm.set.loop.iterations.i32(i32) #3
+
+; Function Attrs: nocallback noduplicate nofree nosync nounwind willreturn
+declare i1 @llvm.loop.decrement.i32(i32) #3
+
+; Function Attrs: mustprogress
+define dso_local void @_Z6conv2dILh1EL5act_t0EaaaLb0ELb1EEvPT1_PT2_PT3_R13conv2d_paramsS6_S6_.for.body.i(i32 %0, ptr %add.ptr3, ptr %ofm, ptr %psum_0_tdm, ptr %psum_1_tdm, ptr %ifm, ptr %add.ptr.i, <64 x i8> %1, i32 %conv10.i, i20 %2, i20 %3, i20 %4, i20 %5, i20 %6, i32 %7, i32 %or25.i.i.i, i32 %8, i20 %9, i20 %10, i20 %11, i20 %12, i20 %13, i20 %14, i20 %15, i20 %16, i20 %17, i20 %18, i20 %19, i32 %conv91.i, i32 %20, i20 %idx.ext.i216.i, i20 %21, i20 %22, i20 %23, i32 %or22.i.i.i) #4 {
```
F-Stuckmann: nit: can we demangle the function name?

llvm/lib/Target/AIE/AIEOuterLoopPipeliner.cpp
```diff
@@ -0,0 +1,856 @@
...
+  LS.InnerExit = LS.InnerLoop->getExitBlock();
+  if (!LS.InnerPreheader || !LS.InnerExit || !LS.InnerLatch) {
+    LLVM_DEBUG(dbgs() << "    Inner loop missing preheader/exit/latch\n");
+    return false;
+  }
+  for (BasicBlock *BB : L->blocks()) {
+    if (LS.InnerLoop->contains(BB))
+      continue;
+    if (DT->dominates(BB, LS.InnerPreheader) && BB != LS.InnerPreheader)
+      LS.PrologueBlocks.push_back(BB);
```
F-Stuckmann: CHECK: this would also add control flow?

llvm/lib/Target/AIE/AIEOuterLoopPipeliner.cpp
```diff
@@ -0,0 +1,856 @@
...
+//       |
+//   [outer.header]  <---------\  <- PHIs: v0/v1 from warm-up or epilogue
+//       |                      |     set.loop.iterations stays here
+//   [inner loop]               |
+//       |                      |
+//   [outer.latch]  ------------/  <- stores + loads for NEXT iteration
+//       |  (false branch)
+//   [cooldown.entry]               <- set.loop.iterations (cloned)
+//       |
+//   [inner loop clone]             <- uses v0.epi/v1.epi from last epilogue
```
martien-de-jong, line 29: I guess we pipeline that loop twice and get the same schedule in both cases.

andcarminati: Yes!

llvm/lib/Target/AIE/AIEOuterLoopPipeliner.cpp
```diff
@@ -0,0 +1,856 @@
...
+  bool performTransformation(LoopStructure &LS);
+
+  // Step 1: Clone data-load chain into a warm-up block before the outer loop.
+  BasicBlock *
+  clonePrologueAsWarmUp(const LoopStructure &LS,
+                        const SmallVectorImpl<Instruction *> &PInsts,
+                        ValueToValueMapTy &WarmUpVMap);
+
+  // Step 2: Clone data-load chain into the epilogue (outer latch), using the
+  // NEXT-iteration pointer values so the loads prefetch for the next iteration.
```
martien-de-jong, line 247: I think the general description would be to translate phi values to the appropriate predecessor. We don't need to fix the decision as to where we cut the warm-up phase in this comment.

llvm/lib/Target/AIE/AIEOuterLoopPipeliner.cpp
```diff
@@ -0,0 +1,856 @@
...
+  SmallVector<Loop *, 4> TopLevelLoops(LI->begin(), LI->end());
+  for (Loop *L : TopLevelLoops)
+    Changed |= runOnLoop(L);
+  return Changed;
+}
+
+bool AIEOuterLoopPipeliner::runOnLoop(Loop *L) {
+  LLVM_DEBUG(dbgs() << "  Considering: ";
+             L->getHeader()->printAsOperand(dbgs(), false); dbgs() << "\n");
+  LoopStructure LS;
```
martien-de-jong: It is not clear to me whether we find suitable loops that are nested in outer loops. I would expect a recursion here to find and transform suitable loops.

llvm/lib/Target/AIE/AIEOuterLoopPipeliner.cpp
```diff
@@ -0,0 +1,856 @@
...
+
+bool AIEOuterLoopPipeliner::analyzeLoopStructure(Loop *L, LoopStructure &LS) {
+  if (L->getSubLoops().size() != 1) {
+    LLVM_DEBUG(dbgs() << "    Not exactly one subloop\n");
+    return false;
+  }
+  LS.OuterLoop = L;
+  LS.InnerLoop = L->getSubLoops()[0];
+  LS.OuterHeader = L->getHeader();
+  LS.OuterLatch = L->getLoopLatch();
```
martien-de-jong: Is it necessary to cache the headers separately from the loops? Perhaps to keep track when repairing LoopInfo?

llvm/lib/Target/AIE/AIEOuterLoopPipeliner.cpp
```diff
@@ -0,0 +1,856 @@
...
+    LLVM_DEBUG(dbgs() << "    No single outer latch\n");
+    return false;
+  }
+  LS.InnerPreheader = LS.InnerLoop->getLoopPreheader();
+  LS.InnerHeader = LS.InnerLoop->getHeader();
+  LS.InnerLatch = LS.InnerLoop->getLoopLatch();
+  LS.InnerExit = LS.InnerLoop->getExitBlock();
+  if (!LS.InnerPreheader || !LS.InnerExit || !LS.InnerLatch) {
+    LLVM_DEBUG(dbgs() << "    Inner loop missing preheader/exit/latch\n");
+    return false;
```
martien-de-jong: nit: return `optional<LoopStructure>` ?

llvm/lib/Target/AIE/AIEOuterLoopPipeliner.cpp
```diff
@@ -0,0 +1,856 @@
...
+  }
+  for (BasicBlock *BB : L->blocks()) {
```
martien-de-jong, line 413: comment what we're checking here?

llvm/lib/Target/AIE/AIEOuterLoopPipeliner.cpp
```diff
@@ -0,0 +1,1556 @@
...
+  return performTransformation(LS, Overrides);
+}
+
+bool AIEOuterLoopPipeliner::analyzeLoopStructure(Loop *L, LoopStructure &LS) {
+  if (L->getSubLoops().size() != 1) {
+    LLVM_DEBUG(dbgs() << "    Not exactly one subloop\n");
+    return false;
+  }
+  LS.OuterLoop = L;
+  LS.InnerLoop = L->getSubLoops()[0];
```
F-Stuckmann: nit: You are setting up the Constructor of LoopStructure, can you move this logic to LoopStructures constructor? Or a factory function?

llvm/lib/Target/AIE/AIEOuterLoopPipeliner.cpp
```diff
@@ -0,0 +1,856 @@
...
+    else if (DT->dominates(LS.InnerExit, BB) && BB != LS.InnerExit)
+      LS.EpilogueBlocks.push_back(BB);
+  }
+  if (LS.PrologueBlocks.empty() && LS.OuterHeader == LS.InnerPreheader) {
+    LS.PrologueInHeader = true;
+    LLVM_DEBUG(dbgs() << "    Prologue is in outer header\n");
+  }
+  if (LS.EpilogueBlocks.empty() && LS.OuterLatch == LS.InnerExit) {
+    LS.EpilogueInLatch = true;
+    LLVM_DEBUG(dbgs() << "    Epilogue is in outer latch\n");
```
martien-de-jong: nit: Inner prologue, Inner epilogue.

llvm/lib/Target/AIE/AIEOuterLoopPipeliner.cpp
```diff
@@ -0,0 +1,856 @@
...
+  }
+  if (!LS.PrologueInHeader && LS.PrologueBlocks.empty()) {
+    LLVM_DEBUG(dbgs() << "    No prologue found\n");
+    return false;
+  }
+  if (!LS.EpilogueInLatch && LS.EpilogueBlocks.empty()) {
+    LLVM_DEBUG(dbgs() << "    No epilogue found\n");
+    return false;
+  }
```
martien-de-jong, line 438: I don't see how this case can arise for a properly nested loop.

llvm/lib/Target/AIE/AIEOuterLoopPipeliner.cpp
```diff
@@ -0,0 +1,1556 @@
...
+  assert(LS.InnerPreheader == LS.OuterHeader &&
+         "Prologue must be in outer header");
+  assert(LS.InnerExit == LS.OuterLatch && "Epilogue must be in outer latch");
+
+  LLVM_DEBUG(dbgs() << "    Linear structure: outer.header -> inner loop -> "
+                       "outer.latch\n");
+  return true;
+}
+
+bool AIEOuterLoopPipeliner::isInnerLoopHardwareLoop(
```
F-Stuckmann, line 445: nit: Can we move this to AIELoopUtils.h?

llvm/lib/Target/AIE/AIEOuterLoopPipeliner.cpp
```diff
@@ -0,0 +1,856 @@
...
+  for (BasicBlock *BB : LS.PrologueBlocks)
+    if (BB->getTerminator()->getNumSuccessors() > 1) {
+      LLVM_DEBUG(dbgs() << "    Complex control flow in prologue\n");
+      return false;
+    }
+  for (BasicBlock *BB : LS.EpilogueBlocks)
+    if (BB->getTerminator()->getNumSuccessors() > 1) {
+      LLVM_DEBUG(dbgs() << "    Complex control flow in epilogue\n");
+      return false;
+    }
```
martien-de-jong: I would say that having a single loop latch would suffice. That's carrying the backedge, across which you copy the warm-up stage.

llvm/lib/Target/AIE/AIEOuterLoopPipeliner.cpp
```diff
@@ -0,0 +1,856 @@
...
+  return true;
+}
+
+bool AIEOuterLoopPipeliner::isInnerLoopHardwareLoop(
+    const LoopStructure &LS) const {
```
martien-de-jong, line 446: I wouldn't mind refactoring this to a general 'isHardwareLoop(Loop *)` in AIELoopUtils.h

llvm/lib/Target/AIE/AIEOuterLoopPipeliner.cpp
```diff
@@ -0,0 +1,856 @@
...
+        Loads.push_back(L);
+  };
+  if (LS.PrologueInHeader)
+    Collect(LS.OuterHeader);
+  else
+    for (BasicBlock *BB : LS.PrologueBlocks)
+      Collect(BB);
+}
+
+void AIEOuterLoopPipeliner::collectEpilogueStores(
```
martien-de-jong, line 468: I am a bit confused here. Conceptually, we run a two-stage pipeline for the outer loop. The freedom that we have is to decide what to put in the first stage, and we limit that to the loads from the prologue. The second stage is everything else in the prologue, the inner loop, and anything in the epilogue. How can we limit ourselves to the stores here?

llvm/lib/Target/AIE/AIEOuterLoopPipeliner.cpp
```diff
@@ -0,0 +1,856 @@
...
+  }
+}
+
+bool AIEOuterLoopPipeliner::isProfitableToRotate(const LoopStructure &LS) {
+  if (!isInnerLoopHardwareLoop(LS)) {
+    LLVM_DEBUG(dbgs() << "    Inner loop is not a hardware loop\n");
+    return false;
+  }
+  std::optional<int64_t> MinTC = llvm::getMinTripCount(LS.OuterLoop, SE);
+  if (!MinTC || *MinTC < (int64_t)OuterLoopPipeliningMinTripCount) {
```
martien-de-jong: should we protect against people setting the command line option to 1? I guess we should discriminate between legal and profitable.

llvm/lib/Target/AIE/AIEOuterLoopPipeliner.cpp
```diff
@@ -0,0 +1,856 @@
...
+    return true;
+  SmallVector<LoadInst *, 8> PrologueLoads;
+  collectPrologueLoads(LS, PrologueLoads);
+  if (PrologueLoads.size() < OuterLoopPipeliningMinLoads) {
+    LLVM_DEBUG(dbgs() << "    Not enough loads in prologue\n");
+    return false;
+  }
+  SmallVector<StoreInst *, 8> EpilogueStores;
+  collectEpilogueStores(LS, EpilogueStores);
+  if (EpilogueStores.empty()) {
```
martien-de-jong, line 507: I think any operations that can run in parallel with the loads from the prologue would qualify. In particular, VMACs or reductor trees or conversions.

llvm/lib/Target/AIE/AIEOuterLoopPipeliner.cpp
martien-de-jong: We want to move them 'next to' the stores. If none of the loads can move past any of the stores, the gain is low.

llvm/lib/Target/AIE/AIEOuterLoopPipeliner.cpp
martien-de-jong: your grow instruction trees from it?

llvm/lib/Target/AIE/AIEOuterLoopPipeliner.cpp
```diff
@@ -0,0 +1,856 @@
...
+  // OrigEpiInsts: the set of instructions that were in the epilogue block
+  // BEFORE clonePrologueIntoEpilogue inserted the prologue load clones.
+  void
+  peelLastIterationEpilogue(const LoopStructure &LS,
+                            const SmallPtrSetImpl<Instruction *> &OrigEpiInsts);
+
+  // Step 6: Adjust the outer loop trip count from N to N-1.
+  void adjustLoopBound(const LoopStructure &LS);
+
+  // Step 7: Drop itercount.range metadata (trip count changed).
```
martien-de-jong: We don't drop it, we repair.

llvm/lib/Target/AIE/AIEOuterLoopPipeliner.cpp
```diff
@@ -0,0 +1,856 @@
...
+    BasicBlock *Orig = InnerBlocks[I];
+    BasicBlock *Clone = ClonedInnerBlocks[I];
+    for (Instruction &I : *Orig) {
+      Instruction *CloneI = I.clone();
+      if (!CloneI->getType()->isVoidTy())
+        CloneI->setName(I.getName() + ".cd");
+      CloneI->insertInto(Clone, Clone->end());
+      CoolVMap[&I] = CloneI;
+    }
+  }
```
martien-de-jong, line 833: I can't see where you repair LoopInfo, adding the extra loop while we pledge to preserve LoopInfo.

llvm/lib/Target/AIE/AIEOuterLoopPipeliner.cpp
```diff
@@ -0,0 +1,1556 @@
...
+  return false;
+}
+
+// Returns true if the intrinsic ID is a safe pointer increment intrinsic
+// (2D/3D addressing) that can be lifted from epilogue to prologue.
+// These intrinsics have no side effects (IntrNoMem) and are pure pointer
+// computations. Any other intrinsic may have unknown side effects and should
+// not be lifted.
+static bool isSafePointerIncrementIntrinsic(Intrinsic::ID IID) {
```
F-Stuckmann, line 133: nit: shouldn't this be in TII?

andcarminati: No TII access here. TII is available in CodeGen, what is ahead of this pass.

llvm/lib/Target/AIE/AIEOuterLoopPipeliner.cpp
andcarminati: I am removing the preserved part. If this info is relevant after, the pass manager will recompute it.

llvm/lib/Target/AIE/AIEOuterLoopPipeliner.cpp
andcarminati: Yes, this is something that I would like to explore.

llvm/lib/Target/AIE/AIEOuterLoopPipeliner.cpp
andcarminati: Maybe the names are a bit confusing. This should be clarified.

llvm/lib/Target/AIE/AIEOuterLoopPipeliner.cpp
andcarminati: Yes! That is a good idea.

llvm/lib/Target/AIE/AIEOuterLoopPipeliner.cpp
andcarminati: This function is only used to profitability test. We may relax this condition if needed.

andcarminati:
Remaining points will be addressed in follow-up PRs.

F-Stuckmann (approved):
let's get this merged and then we can properly clean up in a follow up PR.

llvm/lib/Target/AIE/AIEOuterLoopPipeliner.cpp
martien-de-jong: TTI then? And don't we have similar functionality for our 2d/3d alias extension?

llvm/lib/Target/AIE/AIEOuterLoopPipeliner.cpp
```diff
@@ -0,0 +1,1570 @@
...
+    "aie-outer-loop-hw-loop",
+    cl::desc("Convert downcounting outer loops to JNZD hardware loops after "
+             "outer loop pipelining"),
+    cl::init(false), cl::Hidden);
+
+// Type alias for split strategy predicates.
+// A split strategy is a function that identifies "anchor" instructions
+// which define the split point between Part 1 (pipelined) and Part 2
+// (stays in outer.header + cooldown.entry).
+using SplitStrategy = std::function<bool(const Instruction *)>;
```
martien-de-jong, line 95: Somehow, it doesn't sound like a strategy.

llvm/lib/Target/AIE/AIEOuterLoopPipeliner.cpp
```diff
@@ -0,0 +1,1570 @@
...
+}
+
+namespace {
+
+// Loop structure for outer loop pipelining.
+// We only support the linear structure:
+//   outer.header (prologue) → inner loop → outer.latch (epilogue)
+// where prologue instructions are in the outer header and epilogue
+// instructions are in the outer latch (no separate blocks).
+struct LoopStructure {
```
martien-de-jong, line 157: Can we please make it a class, adding public" if necessary?

PR #999, opened 2026-05-13
[AIE] Consolidate itinerary queries over register supersets

martien-de-jong:
Lo-priority

Generated reg itin pairs tend to split register sets in their subsets, all yielding the same itinerary.
By recognising those equivalences and reverting to the largest superset we gain several advantages:
1- There are less entries to check. This saves table size and compile time
   spent in table lookup.
2- For reverse-lookup purposes, i.e. finding register classes compatible with
   physical register occurrences, we end up at wider sets with more allocation
   freedom.


llvm/utils/TableGen/AIEItineraryEquivalence.h
```diff
@@ -0,0 +1,289 @@
...
+
+  /// Cycles until the start of the next stage (cycle delta between consecutive
+  /// stages).
+  int TimeInc = 0;
+
+  /// Timing model record name (e.g., "NoItinerary", "Reserved").
+  StringRef TimingModelName;
+
+  bool operator==(const StageCycleInfo &Other) const {
+    return Cycles == Other.Cycles && UnitNames == Other.UnitNames &&
```
andcarminati: nit: check cheaper first, also, could we have early returns? Cycles -> TimingModelName -> UnitNames? Maybe we can save some table build time.

llvm/utils/TableGen/AIEItineraryEquivalence.h
```diff
@@ -0,0 +1,289 @@
...
+           TimeInc == Other.TimeInc && TimingModelName == Other.TimingModelName;
+  }
+
+  bool operator<(const StageCycleInfo &Other) const {
+    if (Cycles != Other.Cycles)
+      return Cycles < Other.Cycles;
+    if (UnitNames != Other.UnitNames)
```
andcarminati: This could be the last test.

llvm/utils/TableGen/AIEItineraryEquivalence.h
```diff
@@ -0,0 +1,289 @@
...
+      return It->first == It->second;
+    return true;
+  }
+
+  /// Get the signature for a given itinerary.
+  ///
+  /// \param ItinName The name of the itinerary to look up.
+  /// \returns The signature for the itinerary. If not found, returns an empty
+  ///          signature.
+  ItinerarySignature getSignature(StringRef ItinName) const {
```
andcarminati: optional?

llvm/utils/TableGen/AIEVariableInstrItineraryEmitter.cpp
```diff
@@ -60,17 +68,351 @@ class AIEVariableInstrItineraryEmitter {
...
+    ArrayRef<const CodeGenRegisterClass *> RCs) {
+  if (RCs.empty())
+    return nullptr;
+
+  if (RCs.size() == 1)
+    return RCs[0];
+
+  // First check that all input classes have compatible attributes.
+  for (size_t I = 1; I < RCs.size(); ++I) {
+    if (!haveCompatibleAttributes(*RCs[0], *RCs[I])) {
```
andcarminati, line 161: Check: this is intended to speedup the whole strategy by discarding obvious cases.

martien-de-jong: For me it's correctness. I shouldn't merge reg classes that aren't equivalent in e.g. spill size. But perhaps there are sanity rules that I don't know of.

llvm/utils/TableGen/AIEItineraryEquivalence.h
martien-de-jong: Actually, unused function...

llvm/utils/TableGen/AIEVariableInstrItineraryEmitter.cpp
martien-de-jong: More detail: I check the result class only against RC[0].

llvm/test/CodeGen/AIE/aie2ps/end-to-end/conv2d_opt_outerloop_pipelining_out_mode_0.ll
```diff
@@ -80,28 +80,28 @@ define void @conv2d_opt_outerloop_out_mode_0(
 ; ASM-NEXT:    lda r4, [p3, dj0]; nopx
 ; ASM-NEXT:    nop
 ; ASM-NEXT:    mova dj1, #18
-; ASM-NEXT:    lda.u8 r16, [p3, dj1]
```
martien-de-jong, line 83: I had hoped this to be NFC, but we get slightly different itineraries in the prescheduler, which is basing itself on the register class of the virtual registers, not on the actual register.  I don't see any changes in II or prologue epilogue bundles, just small register allocation and scheduling differences.

llvm/test/CodeGen/AIE/aie2ps/end-to-end/conv2d_opt_outerloop_pipelining_out_mode_1.ll
```diff
@@ -164,26 +164,25 @@ define void @conv2d_opt_outerloop_out_mode_1(
...
-; ASM-NEXT:    lda dn6, [p3], #-8; vldb.pop.3d x1, [p0, lf0, r24, d0]; add r1, r7, r28; mov p2, p1
-; ASM-NEXT:    lda.s8 r6, [p3, dj3]; movs m3, r7; add r5, r1, r22; vbcst.32 x0, r18
-; ASM-NEXT:    lda dj6, [p3, #0]; movs dc6, dc3; or r10, r8, r0; mov p3, p1
-; ASM-NEXT:    padda [p3], m3; movs dc2, dc3; add r18, r3, #-1; mov m3, r1
-; ASM-NEXT:    padda [p2], m3; movs dj7, r22; movx r0, #7; mov m3, r5
-; ASM-NEXT:    padda [p1], m3; movs dc5, dc3; lshl r0, r3, r0; mov r5, p6
-; ASM-NEXT:    movs m3, r0; vldb.128 wl2, [p2, dj7]; add r0, r20, #-1; mov r3, p1
-; ASM-NEXT:    movs dj3, r28; vldb.128 wl4, [p1, #16]; movx r12, #264; addm.nc r1, r0, #-1
-; ASM-NEXT:    vlda.ups.2x cml0, s0, upssign1, [p5], #64; vldb x8, [p3, dj3]; movx crsrsmode, #0; vshuffle x10, x10, x1, r4; movs p6, p5
-; ASM-NEXT:    vlda.ups.2x cmh0, s0, upssign1, [p5], #64; vldb x6, [p2, #64]; movx r16, #16; mov s1, r6; movs dc1, dc3
```
mludevid: I'm surprised that we are gaining one cycle here. Have you looked into this?

llvm/test/CodeGen/AIE/aie2ps/end-to-end/gemm_int8_0_v2.ll
```diff
@@ -80,57 +80,58 @@ define dso_local void @gemm_int8_0_v2(ptr noalias %p_a, ptr noalias %p_b, ptr no
...
+; CHECK-NEXT:    lda r28, [p6, #0]; sel.eqz r8, r22, r3, r27; addm.nc r5, r5, #-1
+; CHECK-NEXT:    lda r30, [p6, #4]; vldb x6, [p1], m4; movxm p6, ##_ZN3aie6detail19transpose_bits_implILj8EaLj64EE13shuffle_modesE; movs dc1, dc0
+; CHECK-NEXT:    padda [p6], m3; movs dc5, dc0; movx r3, #16; vbcst.16 x0, r17
+; CHECK-NEXT:    lda r2, [p6, #0]; vldb.3d x8, [p1], d1; movx r17, #15; vshift x10, x4, x0, r3
+; CHECK-NEXT:    lda r4, [p6, #4]; vsel.32 x4, x2, x4, r17
+; CHECK-NEXT:    lda m3, [p5], #-4; or r25, r10, r10; vsel.32 x2, x2, x10, r17
+; CHECK-NEXT:    lda dn3, [p5, #0]; movxm p6, ##_ZN3aie6detail19transpose_bits_implILj8EaLj64EE13shuffle_modesE
+; CHECK-NEXT:    lda dj3, [p5, #-4]; movxm p5, #.LBB0_1
+; CHECK-NEXT:    vlda.ups.2x cmh0, s0, upssign1, [p4], #64; or r29, r12, r12; vshuffle x6, x6, x0, r24
+; CHECK-NEXT:    padda [p6], m3; vldb x6, [p0], #64; movx r12, #776; vshuffle x1, x6, x0, r26
```
mludevid: Here on the other side we are losing one cycle. I'm also surprised by this regression.

llvm/test/CodeGen/AIE/aie2ps/end-to-end/gemm_int8_psum_0_v2.ll
```diff
@@ -104,38 +104,40 @@ define dso_local void @gemm_int8_psum_0_v2(ptr noalias %p_a, ptr noalias %p_b, p
...
+; CHECK-NEXT:    lda r28, [p6, #0]; sel.eqz r8, r22, r3, r27; addm.nc r5, r5, #-1
+; CHECK-NEXT:    lda r30, [p6, #4]; vldb x6, [p1], m4; movxm p6, ##_ZN3aie6detail19transpose_bits_implILj8EaLj64EE13shuffle_modesE; movs dc1, dc0
+; CHECK-NEXT:    padda [p6], m3; movs dc5, dc0; movx r3, #16; vbcst.16 x0, r17
+; CHECK-NEXT:    lda r2, [p6, #0]; vldb.3d x8, [p1], d1; movx r17, #15; vshift x10, x4, x0, r3
+; CHECK-NEXT:    lda r4, [p6, #4]; vsel.32 x4, x2, x4, r17
+; CHECK-NEXT:    lda m3, [p5], #-4; or r25, r10, r10; vsel.32 x2, x2, x10, r17
+; CHECK-NEXT:    lda dn3, [p5, #0]; movxm p6, ##_ZN3aie6detail19transpose_bits_implILj8EaLj64EE13shuffle_modesE
+; CHECK-NEXT:    lda dj3, [p5, #-4]; movxm p5, #.LBB0_1
+; CHECK-NEXT:    vlda bmhh0, [p4], #64; vldb x6, [p0], #64; or r29, r12, r12; vshuffle x6, x6, x0, r24
+; CHECK-NEXT:    padda [p6], m3; movs dc4, dc0; movx r12, #776; vshuffle x10, x6, x0, r26
```
mludevid, line 140: As you mentioned, due to the merged itineraries the pre-scheduler has access to more specific itineraries. Most of the time that leads to better schedules, but we do see some regressions, for example here. As far as I can tell this regression is caused by 2 `paddb` instructions having changed to `padda` instructions although we are bound on the A slot in this basic block. Why did the multi-slot materialization take this worse decision?

llvm/utils/TableGen/AIEItineraryEquivalence.h
```diff
@@ -0,0 +1,255 @@
...
+    for (const Record *Stage : ItinData->getValueAsListOfDefs("Stages")) {
+      StageCycleInfo Info;
+      Info.Cycles = static_cast<int>(Stage->getValueAsInt("Cycles"));
+      Info.TimeInc = static_cast<int>(Stage->getValueAsInt("TimeInc"));
+      if (Stage->getValue("Units")) {
+        for (const Record *Unit : Stage->getValueAsListOfDefs("Units"))
+          Info.UnitNames.push_back(Unit->getName());
+      }
+      if (Stage->getValue("TM")) {
+        const Record *TM = Stage->getValueAsDef("TM");
```
mludevid: What's TM? Should this be "Kind"?

llvm/utils/TableGen/AIEItineraryEquivalence.h
```diff
@@ -0,0 +1,255 @@
...
+/// This class builds and maintains a mapping from itinerary names to their
+/// canonical representative. Itineraries with identical signatures map to
+/// the same representative, which is the first itinerary encountered with
+/// that signature.
+class ItineraryEquivalenceMap {
+  /// Maps itinerary name to its representative (canonical) itinerary name.
+  std::map<StringRef, StringRef> EquivalenceMap;
+
+  /// Maps itinerary name to its signature.
+  std::map<StringRef, ItinerarySignature> ItineraryToSignature;
```
mludevid: This field is never read, just written to. Can this be removed?

llvm/utils/TableGen/AIEItineraryEquivalence.h
```diff
@@ -0,0 +1,255 @@
...
+             const CodeGenProcModel &ItinModel) {
+    // Process all schedule classes to build the equivalence map.
+    for (const CodeGenSchedClass &SchedClass :
+         SchedModels.explicitSchedClasses()) {
+      // Some schedule classes have no itinerary entry; skip them.
+      if (SchedClass.Index >= ItinModel.ItinDefList.size())
+        continue;
+      const Record *ItinData = ItinModel.ItinDefList[SchedClass.Index];
+      if (!ItinData)
+        continue;
```
mludevid, line 176: How can this case happen?

llvm/utils/TableGen/AIEItineraryEquivalence.h
```diff
@@ -0,0 +1,255 @@
...
+    return ItinName;
+  }
+
+  /// Check if an itinerary is the representative of its equivalence class.
+  bool isRepresentative(StringRef ItinName) const {
+    auto It = EquivalenceMap.find(ItinName);
+    if (It != EquivalenceMap.end())
+      return It->first == It->second;
+    return true;
+  }
```
mludevid: As far as I can tell `isRepresentative` is dead code.

llvm/utils/TableGen/AIEVariableInstrItineraryEmitter.cpp
```diff
@@ -60,17 +68,351 @@ class AIEVariableInstrItineraryEmitter {
...
+    : Records(R), Target(R), CDP(R),
+      SchedModels(CDP.getTargetInfo().getSchedModels()),
+      RegBank(Target.getRegBank()) {
   CurrentNamespace = Target.getName().str();
+
+  // Find the itinerary model and build equivalence map.
+  if (SchedModels.hasItineraries()) {
+    for (const CodeGenProcModel &ProcModel : SchedModels.procModels()) {
+      if (ProcModel.hasItineraries()) {
+        assert(ItinModel == nullptr && "Multiple Itin models");
```
mludevid: Use `printFatalError` instead.

llvm/utils/TableGen/AIEVariableInstrItineraryEmitter.cpp
```diff
@@ -60,17 +68,351 @@ class AIEVariableInstrItineraryEmitter {
...
+    bool AllValid = true;
+    for (const size_t GI : GroupIndices) {
+      const StringRef RCName = WorkSet[GI].OperandRCs[DiffPos].second;
+      const Record *RCDef = Records.getDef(RCName);
+      if (!RCDef) {
+        AllValid = false;
+        break;
+      }
+      CodeGenRegisterClass *RC = RegBank.getRegClass(RCDef);
+      if (!RC) {
```
mludevid, line 244: Nit: `getRegClass` never returns `nullptr` so this is dead code.

llvm/utils/TableGen/AIEVariableInstrItineraryEmitter.cpp
```diff
@@ -60,17 +68,351 @@ class AIEVariableInstrItineraryEmitter {
...
+                                       const CodeGenRegisterClass &B);
+
+  // Find a register class whose members are exactly the union of the input
+  // classes and has compatible attributes. Returns nullptr if none exists.
+  const CodeGenRegisterClass *
+  findUnionClass(ArrayRef<const CodeGenRegisterClass *> RCs);
+
+  // Try to merge a group of Multiplicity variants in WorkSet at a given
+  // operand position. Tries all candidate subsets to find one with a valid
+  // union class. On success, mutates WorkSet and returns true.
```
mludevid: "all candidate subsets": I think this is not true, we are only trying to merge adjacent itineraries with a sliding window.

llvm/utils/TableGen/AIEVariableInstrItineraryEmitter.cpp
```diff
@@ -60,17 +68,351 @@ class AIEVariableInstrItineraryEmitter {
...
+    for (size_t I = Start; I < Start + NumNeeded; ++I)
+      GroupIndices.push_back(CandidateIndices[I]);
+
+    // Collect the register classes at the differing position.
+    std::vector<const CodeGenRegisterClass *> RCs;
+    bool AllValid = true;
+    for (const size_t GI : GroupIndices) {
+      const StringRef RCName = WorkSet[GI].OperandRCs[DiffPos].second;
+      const Record *RCDef = Records.getDef(RCName);
+      if (!RCDef) {
```
mludevid: Nit: I think this is unreachable as well.

llvm/utils/TableGen/AIEVariableInstrItineraryEmitter.cpp
```diff
@@ -60,17 +68,351 @@ class AIEVariableInstrItineraryEmitter {
...
+  ItineraryEquivalenceMap ItinEquivMap;
 };
 
 } // namespace
 
 AIEVariableInstrItineraryEmitter::AIEVariableInstrItineraryEmitter(
     const RecordKeeper &R)
-    : Records(R), Target(R) {
+    : Records(R), Target(R), CDP(R),
+      SchedModels(CDP.getTargetInfo().getSchedModels()),
```
mludevid: Please try getting the scheduling models from target directly instead of from the `COdeGenDAGPattern` which is otherwise unused and could thus be removed.

llvm/utils/TableGen/AIEItineraryEquivalence.h
martien-de-jong: 100%

llvm/test/CodeGen/AIE/aie2ps/end-to-end/gemm_int8_psum_0_v2.ll
martien-de-jong: I think because we don't run static materialization here, so we materialize on very local postscheduler view.

llvm/utils/TableGen/AIEItineraryEquivalence.h
martien-de-jong: Good question, but this is checked in many other places as well.

llvm/utils/TableGen/AIEVariableInstrItineraryEmitter.cpp
martien-de-jong: FYI: This fails for synthetic union sets, since these are not added to Records. They are found as union and then percolate into future worksets. Work around is to fall back to linear search trhough the reg class if Records.getDef() can't find it.

llvm/utils/TableGen/AIEVariableInstrItineraryEmitter.cpp
```diff
@@ -60,17 +67,371 @@ class AIEVariableInstrItineraryEmitter {
...
+      for (size_t Idx : Indices)
+        WorkSet.push_back(InstrData.Variants[Idx]);
+
+      // Try N-tuple consolidation for N from WorkSet.size() down to 2.
+      // The outer fixpoint ensures that a successful merge is followed by
+      // another pass on the updated WorkSet.
+      bool Changed = true;
+      while (Changed) {
+        Changed = false;
+        for (unsigned Multiplicity = WorkSet.size(); Multiplicity >= 2;
```
mludevid: Nit: No need to keep trying multiplicities if you already found something. Check `Changed`.

mludevid (dismissed):
LGTM.

mludevid (dismissed):

mludevid (approved):

PR #1000, opened 2026-05-13
[CI] Make nightly wheel version PEP 440 / PyPI compliant

hunhoffe:
## Problem                                                                                                                                                  
Today's nightly wheel version is `X.Y.Z.{datetime}+{hash}`. The `+hash` local version segment is rejected by PyPI on upload, so the wheels can only ship to GitHub Releases and the `llvm-aie` PyPI namespace  stays unclaimed.                                                                                                                                                     
                                                                                                                                                              
## Change                                                                                                                                                   
Three small edits, scoped to `.github/workflows/`:                                                                                                          
1. **Version → `X.Y.Z.{datetime}`** (4-segment PEP 440 release).  PyPI-uploadable. Bare `pip install llvm-aie` works without `--pre`, matching the project's posture (nightlies are the shipped artifact). Pinnable as `llvm-aie==X.Y.Z.{datetime}` for reproducible builds downstream.                                                                                                                                              
                                                                                                                                                              
2. **Embed the source SHA inside the wheel** as `commit.txt`. PEP 440 no longer lets the hash ride in the version string, so this keeps it recoverable from the unpacked tree. `long_description` still links the GitHub commit URL.                                                                                                                             
                                                                                                                                                              
3. **`DATETIME` granularity hour → minute.** With `+hash` removed, two uploaded runs in the same hour would produce identical wheel filenames and silently overwrite on GitHub Releases. Minute granularity makes practical collisions impossible.                                                                                                       
                                                                                                                                                              
The `get_wheel_version` step is also adjusted — it previously assumed  a `+` was present in the version and would crash without one.                                                                                               
                                                                                                                                                              
## Notes for reviewers                                                                                                                                      
- Sample wheel name: `llvm_aie-19.0.0.202605131342-py3-none-manylinux_2_28_x86_64.whl`                                                                                         
- A future stable release would need to bump the major (e.g. `20.0.0`), since `19.0.0.{datetime}` sorts above `19.0.0` under PEP 440. Happy to switch to a `.devN` scheme instead if the project would prefer that ordering property — the trade-off is `--pre` becomes required for bare `pip install`.            
- This PR does not add a PyPI publish step but should make that step easier in the future.

Thanks for considering — let me know what you'd like adjusted!


konstantinschwarz (approved):
This change looks good to me! I assume we'll need to synchronize with the consumers of our packages to use the new names for version pinning

hunhoffe:
Hi! Would a maintainer be willing to kick off the AMD AIE Distro workflow against this branch? It's workflow_dispatch-only and doesn't run on PRs, so the version-string changes haven't actually been exercised in CI yet. A single Linux x86_64 matrix entry is enough to verify — I'd like to confirm the wheel filename, embedded commit.txt, and the patched get_wheel_version step all behave as intended before this merges.

PR #1001, opened 2026-05-13
[AIE2PS] Fix alignment of 64-bit spills and improve CSR spills of L registers

khallouh:


llvm/lib/Target/AIE/aie2ps/AIE2PSInstrInfo.cpp
```diff
@@ -540,9 +540,9 @@ AIE2PSInstrInfo::getSpillPseudoExpandInfo(const TargetRegisterInfo &TRI,
 
   switch (MI.getOpcode()) {
   case AIE2PS::ST_R_SPILL:
-    return {{AIE2PS::ST_dms_sts_scalar_spill}};
+    return {{AIE2PS::ST_dms_sts_scalar_spill, AIE2PS::NoSubRegister, 8}};
```
martien-de-jong: Why 8 here?

niwinanto (dismissed):
LGTM!

llvm/lib/Target/AIE/aie2ps/AIE2PSInstrInfo.cpp
khallouh: It is not needed anymore. Dropped it.

llvm/test/CodeGen/AIE/aie2ps/bfp16_13_mul.ll
```diff
@@ -20,15 +20,16 @@ target triple = "aie2ps"
...
 ; CHECK-NEXT:    vmul.f dm0, ex0, ey2, ey3, r8
-; CHECK-NEXT:    nop
+; CHECK-NEXT:    lda r9:r8, [sp, #-64] // 8-byte Folded Reload
 ; CHECK-NEXT:    ret lr
 ; CHECK-NEXT:    nop // Delay Slot 5
 ; CHECK-NEXT:    nop // Delay Slot 4
 ; CHECK-NEXT:    nop // Delay Slot 3
 ; CHECK-NEXT:    nop // Delay Slot 2
-; CHECK-NEXT:    mov r8, r0 // Delay Slot 1
+; CHECK-NEXT:    paddxm [sp], #-64 // Delay Slot 1
```
niwinanto: We are regressing now, would be nice to see why

llvm/test/CodeGen/AIE/aie2ps/end-to-end/conv2d_int8_outerloop_pipelined.ll
```diff
@@ -67,12 +67,12 @@ define dso_local void @conv2d(i32 %0, ptr %add.ptr3, ptr %cond, ptr %cond.i, ptr
...
-; CHECK-NEXT:    lda m2, [p6], #-4; mov s1, r3
-; CHECK-NEXT:    lda dj2, [p6], #-4; or r28, r8, r8; mov dj3, #0
-; CHECK-NEXT:    lda dj6, [p6], #-4; movs dc2, dj4; or r30, r5, r5; mov r5, dj4
-; CHECK-NEXT:    lda dn2, [p6, #0]; movs dc6, dj4; or r8, r7, r7; mov r7, dj4
+; CHECK-NEXT:    lda r26, [p6], #-4; mov dj4, #0
+; CHECK-NEXT:    lda r22, [p6], #-4; mov s1, r3
+; CHECK-NEXT:    lda m2, [p6], #-4; mov dj3, #0
+; CHECK-NEXT:    lda dj2, [p6], #-4; mov dc2, dj4
+; CHECK-NEXT:    lda dj6, [p6], #-4; movs dc6, dj4; or r30, r5, r5; mov r5, dj4
+; CHECK-NEXT:    lda dn2, [p6, #0]; st r9:r8, [sp, #-64]; or r8, r7, r7; mov r7, dj4 // 8-byte Folded Spill
```
andcarminati: For some reason we are spilling more. luckily we could pack the stores within available bundles and we also have more unused stack because of the alignment (64 bytes). Do you have an idea?

llvm/test/CodeGen/AIE/aie2ps/bfp16_13_mul.ll
khallouh: Yes, determineCalleeSaved is not smart enough to mark only the gpr that is needed for saving, once we have L registers in the CSR list, whenever a subregister is required it tries to save the entire super-register. Moreover, when saving the L register, you see that it spills to memory even though it could copy it to another L register but it doesn't at the moment. The reason being that the spilling to register of CSRs handling in our AIEBaseFrameLowering is only fitted to GPRs. I will change all that, then we would have more improvement and there should not be any regressions anymore.

llvm/test/CodeGen/AIE/aie2ps/end-to-end/conv2d_int8_outerloop_pipelined.ll
khallouh: yes, see comment above

llvm/test/CodeGen/AIE/aie2ps/bfp16_13_mul.ll
khallouh: Sadly there is no single L to L move instruction so the second improvement I mentioned won't work even if we support it. I was confused by moves involving exponents which looked similar. Anyways, I will just fix the regressions.

khallouh:
Updated. There are no regressions anymore including in mllib L1 (full run). Also conv2 and gemm don't change in the tests. The improvements in those kernels are mostly in the setup and main functions. I am currently running mllib again (reduced run) to get updated improvement numbers since they are supposed to increase with the last commit.

khallouh:
There are only improvements, zero regressions
image

llvm/lib/Target/AIE/aie2ps/AIE2PSFrameLowering.cpp
```diff
@@ -85,13 +85,74 @@ void AIE2PSFrameLowering::determineCalleeSaves(MachineFunction &MF,
...
+  // When both L registers and their sub-GPRs are in the CSR list, we need to
+  // decide whether to save as L register or individual GPRs.
+  //
+  // Strategy:
+  // - If only one GPR of the pair is used: save just that GPR
+  // - If both GPRs are used AND function has calls: use L register save
+  //   (stack spill is required, 1 L spill is more efficient than 2 GPR spills)
+  // - If both GPRs are used AND no calls: use individual GPR saves
+  //   (allows GPR-to-GPR spilling via scratch registers)
+  static const MCPhysReg LRegs[] = {AIE2PS::l4, AIE2PS::l5, AIE2PS::l6,
```
niwinanto: Nit: CalleeSavedLRegs maybe?

llvm/lib/Target/AIE/aie2ps/AIE2PSFrameLowering.cpp
```diff
@@ -85,13 +85,74 @@ void AIE2PSFrameLowering::determineCalleeSaves(MachineFunction &MF,
...
+    bool EvenMarked = SavedRegs.test(EvenGPR);
+    bool OddMarked = SavedRegs.test(OddGPR);
+
+    if (!LRegMarked && !EvenMarked && !OddMarked)
+      continue;
+
+    // Determine if both subregisters actually need saving.
+    // LRegMarked alone doesn't mean both - check individual GPR marks.
+    bool BothNeeded =
+        (EvenMarked && OddMarked) || (LRegMarked && !EvenMarked && !OddMarked);
```
niwinanto, line 134: Do we have both L and R at the same time? Can we assert?

llvm/lib/Target/AIE/aie2ps/AIE2PSFrameLowering.cpp
```diff
@@ -85,13 +85,74 @@ void AIE2PSFrameLowering::determineCalleeSaves(MachineFunction &MF,
...
+        SavedRegs.reset(OddGPR);
+    }
+  }
 
   // If there is a frame pointer (dynamic stack allocation), p7 will be used as
   // a frame pointer. The register allocator will not be able to see the
   // redefinition of p7 as the prologue and the epilogue are emitted after the
   // register allocation. Thus, we make sure to spill p7 at the beginning of the
   // function body and restore it at the end by adding it in SavedRegs.
+  Register FPReg = STI.getRegisterInfo()->getFrameRegister(MF);
```
niwinanto: Nit: Why don`t we directly use already available TRI?

llvm/lib/Target/AIE/aie2ps/AIE2PSFrameLowering.cpp
```diff
@@ -85,13 +85,74 @@ void AIE2PSFrameLowering::determineCalleeSaves(MachineFunction &MF,
...
+  // decide whether to save as L register or individual GPRs.
+  //
+  // Strategy:
+  // - If only one GPR of the pair is used: save just that GPR
+  // - If both GPRs are used AND function has calls: use L register save
+  //   (stack spill is required, 1 L spill is more efficient than 2 GPR spills)
+  // - If both GPRs are used AND no calls: use individual GPR saves
+  //   (allows GPR-to-GPR spilling via scratch registers)
+  static const MCPhysReg LRegs[] = {AIE2PS::l4, AIE2PS::l5, AIE2PS::l6,
+                                    AIE2PS::l7};
```
niwinanto:
```suggestion
const MCPhysReg *CSRegs = MF.getRegInfo().getCalleeSavedRegs();
for (unsigned i = 0; CSRegs[i]; ++i) {
  MCPhysReg Reg = CSRegs[i];
  if (!AIE2PS::eLRegClass.contains(Reg))
    continue;
  ...
```
Since we use multiple calling conventions, would be nice to derive this list than hardcoding. At this moment it has no correctness impact, but it can change in the future.

llvm/lib/Target/AIE/aie2ps/AIE2PSFrameLowering.cpp
```diff
@@ -85,13 +85,74 @@ void AIE2PSFrameLowering::determineCalleeSaves(MachineFunction &MF,
...
+      continue;
+
+    // Determine if both subregisters actually need saving.
+    // LRegMarked alone doesn't mean both - check individual GPR marks.
+    bool BothNeeded =
+        (EvenMarked && OddMarked) || (LRegMarked && !EvenMarked && !OddMarked);
+
+    if (BothNeeded) {
+      // Both subregisters need saving.
+      if (MFI.hasCalls()) {
```
niwinanto, line 144: I don`t fully understand why we have to do special handling with a call in the function, since we are working on callee saved registers. Would be nice to include a comment for future refernce.

llvm/lib/Target/AIE/aie2ps/AIE2PSFrameLowering.cpp
```diff
@@ -85,13 +85,74 @@ void AIE2PSFrameLowering::determineCalleeSaves(MachineFunction &MF,
...
+        // Use L register save. Stack spill is required (scratch regs
+        // clobbered by calls), so 1 L spill is more efficient than 2 GPR
+        // spills.
+        SavedRegs.set(LReg);
```
niwinanto, line 148: Some of these resets seems redundant as they never set in the first place. Could you please double check?

llvm/lib/Target/AIE/aie2ps/AIE2PSFrameLowering.cpp
khallouh: I didn't add this line, I just moved it down

llvm/lib/Target/AIE/aie2ps/AIE2PSFrameLowering.cpp
khallouh: It is a hardcoded list anyway as part of the ABI. I don't think it will change for this target, this is already in an aie2ps target specific hook. We won't be sharing this with another target. If the same list is needed somewhere else in the future, we can factor it out for reuse.

llvm/lib/Target/AIE/aie2ps/AIE2PSFrameLowering.cpp
khallouh: We would like to avoid unnecessary checks so resetting here is fine. There is no problem in resetting it when it wasn't set.

llvm/lib/Target/AIE/aie2ps/AIE2PSFrameLowering.cpp
khallouh: As discussed offline, when there is calls we mark the L register so that we get a single spill instead of 2. When there are no calls, we prefer marking the subregsiters since they can be copied to non CSR registers instead of spilled to memory (There is no move instruction between L registers). For the call case we have no choice but to spill anyway since we don't know which registers the callee is going to use. I'll add this as a comment.

llvm/lib/Target/AIE/aie2ps/AIE2PSFrameLowering.cpp
khallouh: The check `!EvenMarked && !OddMarked` is meant for the case where only one GPR is touched, since we have its super-register in the CSR list it will also be marked by the generic  `determineCalleeSaves` for saving, but we don't want to save the entire L just the gpr. This is then handled in the else case where we reset the L and mark the single GPR that is actually needed.
```
    } else {
      // Only one GPR needs saving - clear L and keep only the needed GPR.
      SavedRegs.reset(LReg);
      if (EvenMarked || (LRegMarked && !OddMarked))
        SavedRegs.set(EvenGPR);
      else
        SavedRegs.reset(EvenGPR);
      if (OddMarked || (LRegMarked && !EvenMarked))
        SavedRegs.set(OddGPR);
      else
        SavedRegs.reset(OddGPR);
    }
```

khallouh:
I addressed the comments and fixed up the test for the call case to reflect realistic MIR (not using CSR registers in the PseudoRET)

llvm/lib/Target/AIE/aie2ps/AIE2PSFrameLowering.cpp
```diff
@@ -85,13 +85,80 @@ void AIE2PSFrameLowering::determineCalleeSaves(MachineFunction &MF,
...
+  static const MCPhysReg CalleeSavedLRegs[] = {AIE2PS::l4, AIE2PS::l5,
+                                               AIE2PS::l6, AIE2PS::l7};
+  for (MCPhysReg LReg : CalleeSavedLRegs) {
+    // Get the two GPR subregisters of this L register
+    MCPhysReg EvenGPR = TRI->getSubReg(LReg, AIE2PS::sub_l_even);
+    MCPhysReg OddGPR = TRI->getSubReg(LReg, AIE2PS::sub_l_odd);
+
+    // Check what's marked for saving by the base determineCalleeSaves.
+    // This already reflects which registers are actually clobbered.
+    bool LRegMarked = SavedRegs.test(LReg);
```
andcarminati: Could we have const in all possible booleans?

llvm/lib/Target/AIE/aie2ps/AIE2PSFrameLowering.cpp
```diff
@@ -85,13 +85,80 @@ void AIE2PSFrameLowering::determineCalleeSaves(MachineFunction &MF,
...
+        SavedRegs.reset(OddGPR);
+    }
+  }
 
   // If there is a frame pointer (dynamic stack allocation), p7 will be used as
   // a frame pointer. The register allocator will not be able to see the
   // redefinition of p7 as the prologue and the epilogue are emitted after the
   // register allocation. Thus, we make sure to spill p7 at the beginning of the
   // function body and restore it at the end by adding it in SavedRegs.
+  Register FPReg = STI.getRegisterInfo()->getFrameRegister(MF);
```
andcarminati: nit: const

llvm/lib/Target/AIE/aie2ps/AIE2PSFrameLowering.cpp
```diff
@@ -85,13 +85,80 @@ void AIE2PSFrameLowering::determineCalleeSaves(MachineFunction &MF,
...
+
+    // Check what's marked for saving by the base determineCalleeSaves.
+    // This already reflects which registers are actually clobbered.
+    bool LRegMarked = SavedRegs.test(LReg);
+    bool EvenMarked = SavedRegs.test(EvenGPR);
+    bool OddMarked = SavedRegs.test(OddGPR);
+
+    if (!LRegMarked && !EvenMarked && !OddMarked)
+      continue;
+
```
andcarminati, line 123: Could we reset all 3 regs, and then just set what we need below?

andcarminati:
This looks good, thank you for fixing this issue! I I just left some comments.

llvm/lib/Target/AIE/aie2ps/AIE2PSFrameLowering.cpp
niwinanto: It means, there could be cases L register and corresponding R register marked at the same time, right?

llvm/lib/Target/AIE/aie2ps/AIE2PSFrameLowering.cpp
niwinanto: +1

llvm/lib/Target/AIE/aie2ps/AIE2PSFrameLowering.cpp
niwinanto: yes, since we have it right now lets use it.

llvm/lib/Target/AIE/aie2ps/AIE2PSFrameLowering.cpp
khallouh: Exactly, but only one of them. This is a big limitation of the generic determineCalleeSaves as it will happily mark all super registers for saving even if only of their subregisters is actually used. So far this has not been an issue as our CSR list only contained atomic registers.

niwinanto: Nit: You could add an assert `assert((!EvenMarked || LRegMarked) && (!OddMarked || LRegMarked) &&
       "sub-reg mark without L pair mark violates invariant");`

llvm/lib/Target/AIE/aie2ps/AIE2PSFrameLowering.cpp
```diff
@@ -85,13 +85,78 @@ void AIE2PSFrameLowering::determineCalleeSaves(MachineFunction &MF,
...
+        SavedRegs.set(OddGPR);
+      }
+    } else {
+      // Only one GPR needs saving
+      if (EvenMarked || (LRegMarked && !OddMarked))
+        SavedRegs.set(EvenGPR);
+      else if (OddMarked || (LRegMarked && !EvenMarked))
+        SavedRegs.set(OddGPR);
+      else
+        SavedRegs.reset(OddGPR);
```
niwinanto: Do we need this reset as we reset all 3 at the begining?

llvm/lib/Target/AIE/aie2ps/AIE2PSFrameLowering.cpp
khallouh: Update: The list should not change for aie2ps but we have a test `save_partial_L` that overrides the calee saved list to exclude $l4. Once I added the assert below, it triggered it
 ```
assert((!EvenMarked || LRegMarked) && (!OddMarked || LRegMarked) &&
    "sub-reg mark without L pair mark violates invariant"); 
```

Comment (khallouh) [reply llvm/lib/Target/AIE/aie2ps/AIE2PSFrameLowering.cpp]: So I will take your suggestion to derive the list

File: llvm/lib/Target/AIE/aie2ps/AIE2PSFrameLowering.cpp (hunk 13 above)
Comment (khallouh) [reply llvm/lib/Target/AIE/aie2ps/AIE2PSFrameLowering.cpp]: No, I had forgot to remove this one

**Review by niwinanto — DISMISSED** [review]
LGTM! Thanks for addressing the comments.

File: llvm/lib/Target/AIE/aie2ps/AIE2PSFrameLowering.cpp [hunk 14]
```diff
@@ -85,13 +85,84 @@ void AIE2PSFrameLowering::determineCalleeSaves(MachineFunction &MF,
...
+        // spills.
+        SavedRegs.set(LReg);
+      } else {
+        // No calls: use individual GPRs for GPR-to-GPR copy.
+        SavedRegs.set(EvenGPR);
+        SavedRegs.set(OddGPR);
+      }
+    } else if (EvenMarked) {
+      SavedRegs.set(EvenGPR);
+    } else if (OddMarked) {
```
niwinanto: Now it looks really good, may be one more nit to assert on `else` case

niwinanto (approved):

llvm/lib/Target/AIE/aie2ps/AIE2PSFrameLowering.cpp
khallouh: Made the `else if` an `else`

PR #1002, opened 2026-05-14
[AIE] Improve AA in MaxLatencyFinder

andcarminati:
__Change 1: Pass Alias Analysis to RegionEndEdges__

Threads `AAResults*` through `RegionEndEdges` → `MaxLatencyFinder` → `findEarliestRef()` → `depends()`. This enables the use of `MachineInstr::mayAlias()` to determine if two memory operations actually alias, instead of conservatively assuming all memory ops depend on each other.

__Change 2: Use isSafeToIgnoreMemDeps in dependency checks__

Queries `BlockState::isSafeToIgnoreMemDeps()` and skips memory dependency checks entirely when it returns true. This mirrors the pattern used in `AIEScheduleDAGMI::mayAlias()` and allows pipelined loops to avoid unnecessary memory constraints.

The second change is safe because the fall-through block (of the an outerloop pipelined block) will be a new copy of one iteration if the complete loop. If we have no memory-carried dependencies in the loop, they will be absent also in the epilogue to prologue back edge and and also epilogue to peeled iteration edge. Also, the second change is intended to be effective (test changes) once we merge https://github.com/Xilinx/llvm-aie/pull/997, to this PR should be merge after.

### DIFF
--- a/llvm/lib/Target/AIE/AIEMaxLatencyFinder.cpp
+++ b/llvm/lib/Target/AIE/AIEMaxLatencyFinder.cpp
@@ -171,16 +171,17 @@ MaxLatencyFinder::MaxLatencyFinder(
     const AIEPostRASchedStrategy *const Scheduler,
     const AIEBaseInstrInfo *const TII,
     const InstrItineraryData *const Itineraries,
-    const MCRegisterInfo *const TRI, MachineBasicBlock *const CurBB)
+    const MCRegisterInfo *const TRI, MachineBasicBlock *const CurBB,
+    AAResults *AA)
     : Scheduler(Scheduler), TII(TII), Itineraries(Itineraries), TRI(TRI),
-      CurBB(CurBB), InterBlock(true) {}
+      CurBB(CurBB), InterBlock(true), AA(AA) {}
 
 // This is called from different contexts, so we need some case analysis
 // If we have a basic block, we are in a regular MachineScheduler invocation,
 // and we will be able to retrieve its strategy,
 // Otherwise we are an abstract region; Scheduler will be nullptr, which
 // will not be derefenced.
-MaxLatencyFinder::MaxLatencyFinder(ScheduleDAGInstrs *DAG)
+MaxLatencyFinder::MaxLatencyFinder(ScheduleDAGInstrs *DAG, AAResults *AA)
     : Scheduler(DAG->getBB()
                     ? static_cast<AIEScheduleDAGMI *>(DAG)->getSchedImpl()
                     : nullptr),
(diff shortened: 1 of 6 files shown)


llvm/lib/Target/AIE/AIEMaxLatencyFinder.cpp
```diff
@@ -132,8 +133,13 @@ static bool depends(const MachineInstr &Src, const MachineInstr &Dst,
...
 
-    if (AA && !IsPartWordStore(Src)) {
-      return Src.mayAlias(AA, Dst, true);
+    if (!IsPartWordStore(Src)) {
+
+      // If it's safe to ignore memory dependencies, skip memory checks.
+      if (SafeToIgnoreMemDeps)
+        return false;
+
+      return AA && Src.mayAlias(AA, Dst, true);
```
martien-de-jong: no AA says no alias?

andcarminati: Good catch! I am going to fix this

llvm/lib/Target/AIE/AIEMaxLatencyFinder.cpp
```diff
@@ -191,7 +197,10 @@ MaxLatencyFinder::MaxLatencyFinder(ScheduleDAGInstrs *DAG, AAResults *AA)
       InterBlock(InterBlockLatency && CurBB &&
                  isBottomRegion(DAG->ExitSU.getInstr()) &&
                  Scheduler->successorsAreScheduled(CurBB)),
-      AA(AA) {}
+      AA(AA), SafeToIgnoreMemDeps(Scheduler && CurBB &&
+                                  Scheduler->getInterBlock()
+                                      .getBlockState(CurBB)
+                                      .isSafeToIgnoreMemDeps()) {}
```
martien-de-jong: This is getting overwhelming. Following isSafeToIgnoreMemDeps(), I see it's wired to is outerlooppipelined metadata, and I can't find why that says my dependences are safe.

andcarminati: This is a current assumption that we need to make here to get a proper compact schedule. Basically we consider that a loop is a candidate to be outerloop pipelined if there is no memory carried dependencies. The related metadata will be attached to the loop/epilogue but the outerloop pipeliner so we catch here. Basically it says that the epilogue stores will not alias to the loads related to the peeled iteration. We will further analyze why AA is too conservative in some cases, and happily drop this.

martien-de-jong: Could you summarize this in a comment in the code?

andcarminati: Yes, done!

F-Stuckmann (approved):

PR #1003, opened 2026-05-14
[AIEX] Drop stderr warning for missing memory bank

F-Stuckmann:
Keep only the missed-optimization remark for loads without an assigned memory bank, removing the duplicate stderr warning emitted via WithColor::warning.

### DIFF
--- a/llvm/lib/Target/AIE/AIEMultiSlotInstrMaterializer.cpp
+++ b/llvm/lib/Target/AIE/AIEMultiSlotInstrMaterializer.cpp
@@ -23,7 +23,6 @@
 #include "llvm/CodeGen/MachineInstr.h"
 #include "llvm/CodeGen/MachineOptimizationRemarkEmitter.h"
 #include "llvm/Support/ErrorHandling.h"
-#include "llvm/Support/WithColor.h"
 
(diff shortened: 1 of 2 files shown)


andcarminati (approved):

llvm/lib/Target/AIE/AIEMultiSlotInstrMaterializer.cpp
```diff
@@ -110,7 +92,7 @@ class SlotMapping {
 
   /// \return whether a Slot can be assigned to \b MI and assign it in the
   /// mapping.
-  bool isSlotAssignable(const MachineInstr &MI, const AIEHazardRecognizer &HR) {
+  bool isSlotAssignable(MachineInstr &MI, const AIEHazardRecognizer &HR) {
```
SagarMaheshwari99, line 95: why remove const?

F-Stuckmann: The optimization remarks could modify the MachineFunction and I am deeply suspicious of dropping the const to set up 
` MachineOptimizationRemarkEmitter MORE(MF, nullptr);`

PR #1005, opened 2026-05-18
[AIE] Fix findPrologueEpilogue assert on loops with multiple non-loop preds

erwei-xilinx:
## Summary

Fix `llc` assertion failure on legitimate non-pipelined single-MBB loops whose loop header has multiple non-loop predecessors.

`AIELoopUtils::findPrologueEpilogue` (introduced in 5fbf2933) asserted that a single-MBB loop has at most one non-loop predecessor / successor. That invariant only holds for **pipelined** loops, which was fine when the only caller was `PipelineExtractor`'s constructor (gated by pipelining filters that guarantee uniqueness).

007425c ("[AIEX] Emit unified pipeliner optimization remarks for all loop types") added a second caller in `InterBlockScheduling::emitLoopRemarks` that runs on **every** single-MBB loop, pipelined or not. Non-pipelined loops can legitimately have multiple non-loop predecessors — e.g. an outer `do-while` loop reached both from a function-entry block and from an end-of-iteration cleanup block, or a halt-spin self-edge combined with normal entry paths. So the assertion now fires on legal CFGs.

## Fix

Make `findPrologueEpilogue` return `nullptr` for the ambiguous side instead of asserting. The remark-emitter call site already has `if (!PrologueMBB || !EpilogueMBB) continue;`. The pipeliner call site keeps its own `assert(PrologueMBB && EpilogueMBB && "Pipelined loop must have a unique prologue and epilogue")`, so pipelined code that ever violates the invariant still fails loudly at the only site where it actually matters.

## Crash signature

```
llc: llvm/lib/Target/AIE/Utils/AIELoopUtils.cpp:275:
  Assertion `!Prologue && "Single-MBB loop has multiple non-loop predecessors"' failed.

Running pass 'PostRA Machine Instruction Scheduler' on function '@core_0_2'
  llvm::AIELoopUtils::findPrologueEpilogue
  llvm::AIE::InterBlockScheduling::emitLoopRemarks
```

## Impact

Currently blocking the mlir-air Ryzen AI CI on **both** NPU Phoenix (amd8845hs, npu1) and NPU Strix (amdhx370, npu2) runners. Reference failing run: [Xilinx/mlir-air actions run 26049388494](https://github.com/Xilinx/mlir-air/actions/runs/26049388494).

**NPU Phoenix (npu1) — 9 failing tests:**
- `xrt/02_mul_shim_1x1/run_peano.lit`
- `xrt/03_mul_L1L2_1x1/run_peano.lit`
- `xrt/06_add_shim_bf16/run_peano.lit`
- `xrt/07_extern_linalg/run_npu1_peano.lit`
- `xrt/28_gemm_loop_nest_bf16/run_peano.lit`
- `xrt/29_gemm_4_level_tiling_extern_vec_4x4_bf16/run_peano.lit`
- `xrt/34_cascade_vecadd/run_npu1_peano.lit`
- `xrt/36_cascade_vecmat_i32/run_npu1_peano.lit`
- `xrt/38_cascade_vecmat_transform_2x4_i32/run_npu1_peano.lit`

**NPU Strix (npu2) — 11 failing tests** (superset of the npu1 list with `_npu2_peano` variants, plus 2 npu2-only tests: `03_mul_L1L2_1x1/run_peano_elf.lit` and `30_mul_rtp_1x1/run_npu2_peano.lit`).

All 9 npu1 tests verified to pass on real NPU Phoenix hardware with this patch applied to nightly `21.0.0.2026051801+55604435`. The npu2 failures share the same root cause (same llc binary, same `findPrologueEpilogue` assertion in the same crash stack), so the fix should unblock them as well.

## Bisect

| nightly | result |
|---|---|
| `2026051501+f4933ef7` (May 15) | OK — last good |
| `2026051601+55604435` (May 16) | CRASH — first bad |

## Test

Added `llvm/test/CodeGen/AIE/aie2/schedule/postpipeliner/loop-remark-multi-pred.mir`: minimal MIR with a single-MBB loop reached from two non-loop predecessors (function entry and end-of-iteration cleanup). Verified to fail on `aie-public` HEAD without the fix and pass with it.

### DIFF
--- a/llvm/lib/Target/AIE/Utils/AIELoopUtils.cpp
+++ b/llvm/lib/Target/AIE/Utils/AIELoopUtils.cpp
@@ -267,20 +267,37 @@ std::pair<MachineBasicBlock *, MachineBasicBlock *>
 findPrologueEpilogue(const MachineBasicBlock &LoopBB) {
   assert(isSingleMBBLoop(&LoopBB) && "Expected a single-MBB loop");
 
+  // Return nullptr for the ambiguous side if the loop has multiple non-loop
+  // predecessors/successors. A valid do-while style outer loop may take its
+  // backedge from a different block than the one that enters it, producing
+  // two non-loop predecessors. Callers that require uniqueness (e.g. the
+  // pipeliner's PipelineExtractor) must check for nullptr themselves.
   MachineBasicBlock *Prologue = nullptr;
   MachineBasicBlock *Epilogue = nullptr;
+  bool MultiplePrologues = false;
+  bool MultipleEpilogues = false;
   for (MachineBasicBlock *P : LoopBB.predecessors()) {
     if (P == &LoopBB)
       continue;
-    assert(!Prologue && "Single-MBB loop has multiple non-loop predecessors");
+    if (Prologue) {
+      MultiplePrologues = true;
+      break;
+    }
     Prologue = P;
   }
   for (MachineBasicBlock *S : LoopBB.successors()) {
     if (S == &LoopBB)
       continue;
-    assert(!Epilogue && "Single-MBB loop has multiple non-loop successors");
+    if (Epilogue) {
+      MultipleEpilogues = true;
+      break;
+    }
     Epilogue = S;
   }
+  if (MultiplePrologues)
+    Prologue = nullptr;
+  if (MultipleEpilogues)
+    Epilogue = nullptr;
   return {Prologue, Epilogue};
 }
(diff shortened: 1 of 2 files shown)


erwei-xilinx:
Cc: @F-Stuckmann

martien-de-jong (approved):

martien-de-jong:
Thanks for fixing this.

PR #1006, opened 2026-05-20
[AIE] Enable HL by default in AIEOuterLoopPipeliner

andcarminati:
JNZD is easier to fit in a schedule, as we are now scheduling epilogues in top-down to the end.

### DIFF
--- a/llvm/lib/Target/AIE/AIEOuterLoopPipeliner.cpp
+++ b/llvm/lib/Target/AIE/AIEOuterLoopPipeliner.cpp
@@ -149,6 +149,21 @@ static bool isSafePointerIncrementIntrinsic(Intrinsic::ID IID) {
 
 namespace {
 
+// Cached result of latch exit condition analysis.
+// Populated once by canAdjustLoopBound() and reused by adjustLoopBound() and
+// getDowncountingInfo() to avoid repeating the same icmp pattern-matching.
+// All pointer fields are guaranteed non-null when the struct is returned.
+// Latch exit condition analysis result, computed once by analyzeLoopStructure()
+// and stored in LoopStructure::Bound. All pointer fields guaranteed non-null.
+struct LatchBoundInfo {
+  ICmpInst *Cmp = nullptr;           // latch exit icmp condition
+  Value *Limit = nullptr;            // loop-invariant limit operand
+  unsigned LimitIdx = 0;             // index of Limit in Cmp's operand list
+  int64_t Step = 0;                  // non-zero induction step (from add)
+  BinaryOperator *Counter = nullptr; // add instr feeding the icmp (phi+step)
+  PHINode *OldIV = nullptr;          // counting PHI in outer header
+};
+
 // Loop structure for outer loop pipelining.
 // We only support the linear structure:
 //   outer.header (prologue) → inner loop → outer.latch (epilogue)
@@ -196,6 +211,10 @@ struct LoopStructure {
   BranchInst *getLatchBranch() const {
     return cast<BranchInst>(OuterLatch->getTerminator());
   }
+
+  // Latch exit condition analysis; populated by analyzeLoopStructure().
+  // Always valid: LoopStructure is only returned after Bound is set.
+  LatchBoundInfo Bound;
 };
 
 // Holds the pre-validated components of a downcounting outer loop exit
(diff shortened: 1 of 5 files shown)


llvm/lib/Target/AIE/AIEOuterLoopPipeliner.cpp
martien-de-jong: I think we heavily rely on the pre-condition that it *is* a downcounting loop. Perhaps mention that in a comment.

martien-de-jong: nit: could this underflow?

martien-de-jong: I'm wondering why we don't use 'prepare-for-pipelining' in the regular hardware loop creation. The only reason I can find is that at this point we are actually pipelining the outer loop. Is there a path that can stop outer loop pipelining for a loop with the pragma, and if so, can we predict that in hardware loop creation itself?

martien-de-jong: Nit: you could cache the anchors needed for replacement in a match structure that you return.

andcarminati: It make sense.

andcarminati: Yes, I plan to transfer the responsibility to the original hardware loop pass in the future.

andcarminati: We check min iter count. To underflow we need to have wrong metadata here, not our fault.

llvm/lib/Target/AIE/AIEOuterLoopPipeliner.cpp
```diff
@@ -298,13 +308,17 @@ class AIEOuterLoopPipeliner : public FunctionPass {
...
+                                      const DowncountingInfo &Info);
 
-  // Returns true if the outer latch has a downcounting icmp pattern that can
-  // be replaced by @llvm.loop.decrement.reg.
-  bool isOuterLoopDowncounting(const LoopStructure &LS) const;
+  // Returns the downcounting pattern components if the outer latch has the
+  // canonical downcounting icmp pattern that can be replaced by
+  // @llvm.loop.decrement.reg, or std::nullopt otherwise.
+  std::optional<DowncountingInfo>
+  getDowncountingInfo(const LoopStructure &LS) const;
```
martien-de-jong, line 350: Oh, danke sch"on!

llvm/lib/Target/AIE/AIEOuterLoopPipeliner.cpp
```diff
@@ -931,38 +946,27 @@ Value *AIEOuterLoopPipeliner::adjustLoopBound(const LoopStructure &LS) {
...
-    // Detect counter direction from the add that produces the counter value.
-    if (auto *Add = dyn_cast<BinaryOperator>(Counter)) {
-      if (Add->getOpcode() == Instruction::Add) {
-        if (auto *C = dyn_cast<ConstantInt>(Add->getOperand(1)))
-          IncrementLimit =
-              C->isNegative(); // decrement counter → increment limit
-      }
+  // Determine the induction step. If the counter is an add with a constant
+  // operand, extract the step directly (handles any step magnitude).
+  // Otherwise, infer a unit step from the comparison predicate.
```
martien-de-jong: I don't see the 'otherwise' path. Moreover, that sounds like another implicit precondition.

andcarminati: Please, see the last commit, we removed any implicit preconditions. Not we have `canAdjustLoopBound` that parses everything that is necessary just once, or rejects before any transformation.

llvm/lib/Target/AIE/AIEOuterLoopPipeliner.cpp
```diff
@@ -149,6 +149,18 @@ static bool isSafePointerIncrementIntrinsic(Intrinsic::ID IID) {
...
+// Cached result of latch exit condition analysis.
+// Populated once by canAdjustLoopBound() and reused by adjustLoopBound() and
+// getDowncountingInfo() to avoid repeating the same icmp pattern-matching.
+struct LatchBoundInfo {
+  ICmpInst *Cmp;           // latch exit icmp condition
+  Value *Limit;            // loop-invariant limit operand
+  unsigned LimitIdx;       // index of Limit in Cmp's operand list
+  int64_t Step;            // induction step (from add const, or +=1 inferred)
+  BinaryOperator *Counter; // add instr feeding the icmp (null for plain PHI)
+  PHINode *OldIV;          // counting PHI feeding Counter (null if not found)
```
andcarminati: Checking this.

llvm/lib/Target/AIE/AIEOuterLoopPipeliner.cpp
```diff
@@ -905,31 +930,22 @@ void AIEOuterLoopPipeliner::peelLastIterationEpilogue(
...
-
-  ICmpInst *Cmp = dyn_cast<ICmpInst>(BI->getCondition());
-  assert(Cmp && "Outer latch condition must be icmp");
+// Validates the latch exit condition and caches all discovered components in
+// LS.LatchBound. Returns true if the loop bound can be adjusted (Step != 0).
+// All subsequent uses of the latch icmp pattern (adjustLoopBound,
+// getDowncountingInfo) read from LS.LatchBound without re-scanning.
+bool AIEOuterLoopPipeliner::canAdjustLoopBound(LoopStructure &LS) const {
+  if (LS.LatchBound)
+    return true; // Already computed.
```
martien-de-jong: CHECK: We have no way to see whether we had a total rejection earlier, but I assume we wouldn't be here. But then I wonder whether we could assert that it has been computed.

martien-de-jong (dismissed):

martien-de-jong (approved):

PR #1007, opened 2026-05-21
Martien.maxlatencyfinder

martien-de-jong:
This PR reuses InterBlockEdges to get a better estimate of MaxLatency.

The InterBlock logic has been cleaned up. We gather all scheduling regions for all blocks first, construct all InterBlockEdges between the current block's Bot region and the Top regions of its successors in BlockState::PerSuccEdges.
With that in place, MaxLatencyFinder was updated to register either the scheduled depth or the static depth of the successor nodes and then use the latencies corrected with that depth to compute the ExitSU edges.

I have integrated Fabian's PR to fix the lock delays, This boils down to copying the tests and the refactorings in TII / SubTarget, since the dag mutator does all the heavy lifting for MaxLatencyFinder now.


llvm/lib/Target/AIE/AIEBaseSubtarget.cpp
martien-de-jong, line 1: This is probably a dead end. With separate DAG instance hanging around, I would like to have more control and visibility on the dag mutators to call, rather than have the machine scheduler calling them. 
Since we have a buildGraph() virtual in the scheduler strategy anyway, we could also directly call the 'AIE' dag mutators as part of that buildGraph().

llvm/test/CodeGen/AIE/aie2/end-to-end/Add2D-red.ll
```diff
@@ -76,20 +76,14 @@ define void @add2d(ptr noalias %params, ptr noalias %ifm1_data, ptr noalias %ifm
...
-; ASM-NEXT:    mova r0, #-4; movxm le, #.L_LEnd0
-; ASM-NEXT:    and r0, r2, r0
-; ASM-NEXT:    mova r2, #-2; add r0, r0, #-4
-; ASM-NEXT:    lshl r0, r0, r2; mov crSRSSign, r6
+; ASM-NEXT:    vlda.3d.ups.s32.d8 cm7, s1, [p2], d0; movx r0, #-4; mov crUPSSign, r4
+; ASM-NEXT:    vlda.ups.s32.d8 cm0, s1, [p1], m1; movxm ls, #.LBB0_3
+; ASM-NEXT:    vlda.3d.ups.s32.d8 cm6, s1, [p2], d0; movxm le, #.L_LEnd0
+; ASM-NEXT:    vlda.ups.s32.d8 cm4, s1, [p1], m1; and r0, r2, r0; mov s1, r3
+; ASM-NEXT:    vlda.3d.ups.s32.d8 cm5, s1, [p2], d0; add r0, r0, #-4; mov r2, #-2
+; ASM-NEXT:    vlda.3d.ups.s32.d8 cm3, s1, [p2], d0; lshl r0, r0, r2; mov crSRSSign, r6
```
martien-de-jong, line 86: This is one of the best improvements. The loads got the 7 cycles load latency, whereas the dependence was on the pointer increments. Therefore, the loads got pushed away from the end of the block without reason.

llvm/test/CodeGen/AIE/aie2ps/end-to-end/conv2d_int8_outerloop_pipelined.ll
```diff
@@ -64,20 +65,16 @@ define dso_local void @conv2d(i32 %0, ptr %add.ptr3, ptr %cond, ptr %cond.i, ptr
...
-; CHECK-NEXT:    vlda.ups.2x cml0, s0, upssign1, [p1, #0]; movs dc3, dj4; mov r23, m5
-; CHECK-NEXT:    padda [p1], m4; movs dc7, dj4; add r0, r0, #-1; mov p6, p0
+; CHECK-NEXT:    lda r20, [p6], #-4; mov dj4, #0
+; CHECK-NEXT:    lda m6, [p6], #-4; mov s1, r3
+; CHECK-NEXT:    lda r1, [p6], #-4; mov dj3, #0
+; CHECK-NEXT:    lda r26, [p6], #-4; movs dc2, dj4; or r28, r8, r8; mov dc6, dj4
+; CHECK-NEXT:    lda r22, [p6], #-4; movs dc0, dj4; or r30, r5, r5; mov r5, dj4
+; CHECK-NEXT:    lda m2, [p6], #-4; movs dc1, dj4; or r8, r7, r7; mov r7, dj4
+; CHECK-NEXT:    lda dj2, [p6], #-4; movs dc5, dj4; mov r23, m5
+; CHECK-NEXT:    lda dj6, [p6], #-4; movs dc3, dj4; mov dj1, r31
```
andcarminati: Crazy. I guess p6 was responsible for this huge pessimistic latency.

llvm/lib/Target/AIE/AIEInterBlockScheduling.cpp
```diff
@@ -839,16 +839,17 @@ void InterBlockScheduling::enterRegion(MachineBasicBlock *BB,
...
-        RegionEnd == BB->end() ? ArrayRef<MachineBundle>(BS.BottomInsert)
-                               : ArrayRef<MachineBundle>();
-    BS.addRegion(BB, RegionBegin, RegionEnd, TopFixedBundles, BotFixedBundles);
+    BS.addRegion(BB, RegionBegin, RegionEnd);
+    // Fixed bundles result from loop pipelining and are set separately on the
+    // region, after the instructions have been physically inserted into the
+    // block by emitInterBlockTop / emitInterBlockBottom.
+    if (RegionBegin == BB->begin() && !BS.TopInsert.empty())
+      BS.getCurrentRegion().setTopFixedBundles(BS.TopInsert);
+    if (RegionEnd == BB->end() && !BS.BottomInsert.empty())
```
F-Stuckmann: CHECK: Do we have the guarantee that a Region cannot have a top and bottom fixed region?
We have that check for loops, but do we also get that guarantee for shared epilog/prologues?

Looking at this code, it is allowed. Do we interleave them?

llvm/lib/Target/AIE/AIEBaseInstrInfo.cpp
```diff
@@ -1025,6 +1025,77 @@ int AIEBaseInstrInfo::getCoreResumeCycleAfterReleaseLoad() const {
...
+
+  // Ensure the memory operation retires before the core stalls. For load-only
+  // instructions, use the lock-ordering cycle (write-back stage) rather than
+  // the bare memory-access cycle, since the load data must be fully committed
+  // before the lock can observe it. nullopt means no backward lock-ordering
+  // constraint for this class (e.g. vmovx ss has mayLoadOrStore but does not
+  // access memory).
+  const bool IsLoadOnly = MemMI.mayLoad() && !MemMI.mayStore();
+  std::optional<int> OptLastMemCycle =
+      getLastMemoryCycleForLockOrdering(MemMI.getDesc().SchedClass, IsLoadOnly);
```
mludevid: I think we need to use `getSchedClass` here as well.

llvm/lib/Target/AIE/AIEBaseSubtarget.cpp
```diff
@@ -219,32 +216,32 @@ class LockDelays : public ScheduleDAGMutation {
         if (PredEdge.getKind() != SDep::Order || !LdSt->mayLoadOrStore()) {
           continue;
         }
-        // Ensure memory operation happens before the core stalls
-        auto OptLastMemCycle =
-            TII->getLastMemoryCycle(LdSt->getDesc().SchedClass);
-        assert(!ExactLatencies || OptLastMemCycle);
```
mludevid, line 225: Q: We dropped this assert. Is that intentional? Is this due to missing memory cycle for stream instructions?

llvm/lib/Target/AIE/AIEBaseInstrInfo.cpp
```diff
@@ -1025,6 +1025,77 @@ int AIEBaseInstrInfo::getCoreResumeCycleAfterReleaseLoad() const {
...
+  if (!OptLastMemCycle)
+    return std::nullopt;
+
+  return *OptLastMemCycle - getCoreStallCycleAfterLock() + 1;
+}
+
+std::optional<int>
+AIEBaseInstrInfo::getLockOrderingDelay(const MachineInstr &Pred,
```
mludevid: Isn't this dead code?

llvm/lib/Target/AIE/AIEBaseInstrInfo.cpp
```diff
@@ -1025,6 +1025,77 @@ int AIEBaseInstrInfo::getCoreResumeCycleAfterReleaseLoad() const {
...
+
+  const unsigned SchedClass = getSchedClass(MemMI.getDesc(), MemMI.operands(),
+                                            MemMI.getMF()->getRegInfo());
+
+  // TM ops hold no lckLdaRsrc/lckStRsrc tokens, so the resume-cycle formula
+  // does not apply; only same-bundle issue must be prevented.
+  if (isTMAccessSchedClass(SchedClass))
+    return 1;
+
+  // some Instructions have mayLoadOrStore but dont access memory, i.e. stream
```
mludevid: Typos: `some instructions have mayLoadOrStore but don't access memory`

llvm/lib/Target/AIE/AIEDataDependenceHelper.h
```diff
@@ -53,6 +58,101 @@ class DataDependenceHelper : public ScheduleDAGInstrs {
...
+  // The successor block feeding instructions after the boundary.
+  MachineBasicBlock *Succ = nullptr;
+  // The boundary between Pred and Succ nodes.
+  std::optional<unsigned> Boundary;
+  // When true, memory edges crossing the boundary are suppressed.
+  bool SafeToIgnoreMemDeps = false;
+
+  /// We can add the same instruction on both sides of the boundary.
+  /// We maintain explicit maps to retrieve the corresponding SUnit.
+  using IndexMap = std::map<MachineInstr *, unsigned>;
```
F-Stuckmann, line 95: nit: Could the comment clarify that MIs in the following MBBs will always have Indices >= Boundary?

llvm/lib/Target/AIE/AIEInterBlockScheduling.cpp
```diff
@@ -832,26 +868,125 @@ BlockState &InterBlockScheduling::getBlockState(MachineBasicBlock *BB) {
...
+    SE.buildEdges();
+  }
+}
+
+void InterBlockScheduling::recordPostDepths(MachineBasicBlock *BB) {
+  for (const auto &SEPtr : getBlockState(BB).getPerSuccEdges()) {
+    InterBlockEdges &SE = *SEPtr;
+    MachineBasicBlock *SuccBB = SE.getSucc();
+    if (!SuccBB)
+      continue;
```
F-Stuckmann, line 1042: CHECK: this is a check for not entering other Functions?

llvm/lib/Target/AIE/AIEDataDependenceHelper.cpp
```diff
@@ -72,4 +72,49 @@ void DataDependenceHelper::dumpDot(raw_ostream &OS,
...
+}
+
+bool InterBlockEdges::isPostBoundaryNode(SUnit *SU) const {
+  return Boundary ? SU->NodeNum >= *Boundary : false;
+}
+
+void InterBlockEdges::recordPostDepth(MachineInstr *MI, int Depth) {
+  const auto Found = SuccMap.find(MI);
+  if (Found == SuccMap.end())
+    return;
```
F-Stuckmann, line 142: Check: This can happen if we are in loop scheduling, right? Can we add a comment?

llvm/lib/Target/AIE/AIEInterBlockScheduling.cpp
```diff
@@ -1300,17 +1367,29 @@ void BlockState::setBlockProperties() {
                               : Overrides.get(PostSchedIgnoreMemoryDeps);
 }
 
+InterBlockEdges *BlockState::getLoopSelfEdge() {
+  for (auto &SEPtr : PerSuccEdges)
+    if (SEPtr->getSucc() == TheBlock)
+      return SEPtr.get();
+  return nullptr;
+}
```
mludevid, line 1503: Nit: Reuse the const implementation.
```
InterBlockEdges *BlockState::getLoopSelfEdge() {
  return const_cast<InterBlockEdges *>(
      std::as_const(*this).getLoopSelfEdge());
}
```

llvm/lib/Target/AIE/AIEDataDependenceHelper.h
```diff
@@ -53,6 +58,101 @@ class DataDependenceHelper : public ScheduleDAGInstrs {
...
+                  MachineBasicBlock *Pred = nullptr,
+                  MachineBasicBlock *Succ = nullptr)
+      : DataDependenceHelper(Context, true, true), Pred(Pred), Succ(Succ),
+        SafeToIgnoreMemDeps(SafeToIgnoreMemDeps) {}
+
+  MachineBasicBlock *getPred() const { return Pred; }
+  MachineBasicBlock *getSucc() const { return Succ; }
+
+  /// Add a Node to the DAG.
+  void addNode(MachineInstr *);
```
F-Stuckmann: nit: Aren't we missing MI?

llvm/lib/Target/AIE/AIEDataDependenceHelper.cpp
```diff
@@ -72,4 +72,49 @@ void DataDependenceHelper::dumpDot(raw_ostream &OS,
   OS << "}\n";
 }
 
+void InterBlockEdges::addNode(MachineInstr *MI) {
+  if (auto Index = initSUnit(*MI)) {
```
F-Stuckmann: nit: can we reduce nesting by early returning if we don't have an Index?

llvm/lib/Target/AIE/AIEInterBlockScheduling.h
```diff
@@ -251,9 +202,10 @@ class BlockState {
   }
   void addRegion(MachineBasicBlock *BB, MachineBasicBlock::iterator RegionBegin,
                  MachineBasicBlock::iterator RegionEnd) {
-    assert((Kind == BlockType::Loop &&
-            FixPoint.Stage == SchedulingStage::GatheringRegions) ||
-           FixPoint.Stage == SchedulingStage::Scheduling);
+    // During the gathering pass addRegion is called once per region
+    // (multi-region blocks call it multiple times).  For newly-created
+    // blocks encountered during scheduling, Regions must be empty.
+    assert(Regions.empty() || FixPoint.NumIters == 0);
```
mludevid: Nit: This assert is weaker now. The problem is that at this point we no longer know if we are gathering regions or not. Maybe could pass that information in? But having an additional parameter just for an assert is also weird.

llvm/lib/Target/AIE/AIEDataDependenceHelper.h
```diff
@@ -53,6 +58,101 @@ class DataDependenceHelper : public ScheduleDAGInstrs {
...
+///   PreRegionLength — total number of bundles in the pre-boundary region.
+///
+///   PostRegionLength — total number of bundles in the post-boundary region,
+///   used to represent the depth of the artificial ExitSU node.
+class InterBlockEdges : public DataDependenceHelper {
+  // The predecessor block feeding instructions before the boundary.
+  MachineBasicBlock *Pred = nullptr;
+  // The successor block feeding instructions after the boundary.
+  MachineBasicBlock *Succ = nullptr;
+  // The boundary between Pred and Succ nodes.
```
F-Stuckmann: nit: can we clarify that the Boundary is the Index of the first Succ Node.

llvm/lib/Target/AIE/AIEInterBlockScheduling.cpp
```diff
@@ -832,20 +868,77 @@ BlockState &InterBlockScheduling::getBlockState(MachineBasicBlock *BB) {
...
+
+    // Pre-boundary: free instructions of the current region.
+    for (MachineInstr *MI : CurRegion.getFreeInstructions())
+      SE.addNode(MI);
+
+    SE.markBoundary();
+
+    // Post-boundary: always use getFreeInstructions() as the single source
+    // of node identity.  Empty regions signify empty basic blocks; in that
+    // case no post-boundary nodes are added.
```
mludevid: Q: Is this correct? If we have a successor that is empty, shouldn't we fall back to using the default safety distance? Because we don't check the successor of that empty successor, which might have restrictions we need to enforce.

llvm/lib/Target/AIE/AIEDataDependenceHelper.cpp
```diff
@@ -72,4 +72,49 @@ void DataDependenceHelper::dumpDot(raw_ostream &OS,
...
+    IndexMap &TheMap = Boundary ? SuccMap : PredMap;
+    TheMap.emplace(MI, *Index);
+  }
+}
+
+void InterBlockEdges::markBoundary() { Boundary = SUnits.size(); }
```
F-Stuckmann: nit: can we assert that we do not mark the boundary multiple times?

llvm/lib/Target/AIE/AIEDataDependenceHelper.h
```diff
@@ -53,6 +58,101 @@ class DataDependenceHelper : public ScheduleDAGInstrs {
...
+  std::optional<unsigned> Boundary;
+  // When true, memory edges crossing the boundary are suppressed.
+  bool SafeToIgnoreMemDeps = false;
```
F-Stuckmann, line 91: nit: I don't think we are using SafeToIgnoreMemDeps=true anywhere

llvm/lib/Target/AIE/AIEBaseSubtarget.cpp
```diff
@@ -298,7 +298,7 @@ class RegionEndEdges : public ScheduleDAGMutation {
       // Extend the edge latency if MI requires delay slots. This makes sure
       // there are at least getNumDelaySlots() cycles between MI and ExitSU.
       if (DelaySlots) {
-        assert(EdgeLatency < DelaySlots);
+        assert(EdgeLatency <= DelaySlots + 1);
```
mludevid, line 387: Please improve the commit message of this "fixup" commit.

llvm/lib/Target/AIE/AIEInterBlockScheduling.cpp
martien-de-jong: As far as this PR is concerned, this is an NFC. We create regions earlier, so we can't pass the inserts in the constructor. Streamlining the Fixed region boundaries is left to a followup PR.

llvm/lib/Target/AIE/AIEBaseSubtarget.cpp
martien-de-jong, line 387: This assert makes sure the new latency is not shorter than the old one. The old MaxLatencyFinder somehow nudged it to fit the limit, which was tighter than necessary.

mludevid:
I've finished a first review round. I've commented on a few nits, but I'm mostly worried about correctness issues at this point in cases where we don't have enough information about our successors. I'm still not convinced that we correctly fall back to using a sound safety margin, especially when one of our successors is an empty region, or other weird cases like that. Let's discuss this together next week.

llvm/lib/Target/AIE/AIEBaseSubtarget.cpp
mludevid: From what I could understand so far the "somehow" is because we used to cap all latencies to be at least 1, see `Latency = std::max(Latency - Earliest, 1);` in `llvm/lib/Target/AIE/AIEMaxLatencyFinder.cpp:262`. Now we allow latencies of 0.

llvm/lib/Target/AIE/AIEDataDependenceHelper.h
mludevid: Isn't this used for fixed instructions created by the SWP?

llvm/lib/Target/AIE/AIEBaseInstrInfo.cpp
F-Stuckmann: nope, DelayLock DAG Mutator also uses it

llvm/lib/Target/AIE/AIEBaseSubtarget.cpp
F-Stuckmann: yes

llvm/lib/Target/AIE/AIEInterBlockScheduling.cpp
martien-de-jong: That will happen automatically. Since no successor nodes are added, we will get a pure ExitEdge latency to the interblock ExitSU. We have replaced ExitSU with another ExitSU
I do need to check that RegionExitEdges of InterBlockEdges is not confusing the state of the predecessor with that of the successor. 
Although it sounds attractive to chain successors recursively...

llvm/lib/Target/AIE/AIEInterBlockScheduling.cpp
martien-de-jong: It's defensive AI programming I think. I find it strange if a PerSuccEdge doesn't have a successor block, since they are created by following the successor lists. On the other hand, I wouldn't want to find the blockstate of nullptr. not recording depth is conservative, retrieving it will default to 0.

llvm/lib/Target/AIE/AIEInterBlockScheduling.cpp
martien-de-jong: I think this would be lying to the const version.

llvm/lib/Target/AIE/AIEInterBlockScheduling.h
martien-de-jong: The assert is a bit out of place in the new situation. We call it once for every region, and the first region in the block will have empty. The other half of the assert would be described by InterBlock::GatheringRegions. I'll take it out.
There are two calls to this method, close to each other and  with explanatory comments. I'll take it out.

llvm/lib/Target/AIE/AIEBaseSubtarget.cpp
martien-de-jong: The assert fires on higher EdgeLatency, not a lower one. I think we got a true ExitSU latency from the successor block, effectively the delayslot latency, which is perfect.
Note that this is the nice thing about using InterBlockEdges. It just chains the worst-case maxlatency of the successor to ours, which accounts for empty or too short successors inductively.

llvm/lib/Target/AIE/AIEInterBlockScheduling.cpp
```diff
@@ -832,26 +868,125 @@ BlockState &InterBlockScheduling::getBlockState(MachineBasicBlock *BB) {
   return Blocks.at(BB);
 }
 
+void InterBlockScheduling::buildPerSuccEdges(MachineBasicBlock *BB) {
+  BlockState &BS = getBlockState(BB);
+  // The current region must have been gathered before we can build edges.
+  // Blocks that are just created (e.g. DedicatedExit) may have empty regions
+  // until enterRegion is called; skip building in that case.
+  if (BS.getRegions().empty())
```
F-Stuckmann: can we assert this? Also can we assert that IsGatheringPhase is false?

llvm/lib/Target/AIE/AIEDataDependenceHelper.h
F-Stuckmann: nope its a shortcut we took in OuterloopPipelining :/

llvm/lib/Target/AIE/AIEMaxLatencyFinder.cpp
```diff
@@ -74,193 +74,118 @@ bool MaxLatencyFinder::isBottomRegion(MachineInstr *ExitMI) {
...
+int MaxLatencyFinder::computeEffectiveLatency(MachineInstr &MI) {
+  int EffectiveLatency = 0;
+  int SuccNo = 0;
+  const auto &PerSuccEdges =
+      Scheduler->getInterBlock().getBlockState(CurBB).getPerSuccEdges();
+  for (auto &SEPtr : PerSuccEdges) {
+    InterBlockEdges &SE = *SEPtr;
+    LLVM_DEBUG(
+        dbgs() << "Processing InterBlockEdge: Pred="
+               << (SE.getPred() ? SE.getPred()->getNumber() : -1) << " Succ="
```
F-Stuckmann, line 109: Can we use printName instead of a number?

llvm/lib/Target/AIE/AIEDataDependenceHelper.cpp
martien-de-jong: Actually I think it's another case of defensive programming. I don't think the caller knows what it is doing if it wants to record the depth of a non-successor node.

llvm/lib/Target/AIE/AIEDataDependenceHelper.h
martien-de-jong: Interesting. Valid but uncommon.

llvm/lib/Target/AIE/AIEDataDependenceHelper.h
martien-de-jong: Perhaps not in this commit. We feed it from a similar flag from BlockState when creating PerSuccEdges, which gets it set from loop meta data. Please note that clangd doesn't look through make_unique to see the  constructor use.

llvm/lib/Target/AIE/AIEBaseInstrInfo.cpp
martien-de-jong: Yet another one.

llvm/lib/Target/AIE/AIEBaseInstrInfo.cpp
martien-de-jong: , e.g., stream

llvm/lib/Target/AIE/AIEInterBlockScheduling.cpp
```diff
@@ -832,26 +868,125 @@ BlockState &InterBlockScheduling::getBlockState(MachineBasicBlock *BB) {
...
+    return;
+  const MachineSchedContext &C = *Context;
+  const Region &CurRegion = BS.getCurrentRegion();
+  const bool SafeToIgnoreMemDeps = BS.isSafeToIgnoreMemDeps();
+
+  auto &PerSuccEdges = BS.getPerSuccEdges();
+  PerSuccEdges.clear();
+
+  for (MachineBasicBlock *SuccBB : BB->successors()) {
```
F-Stuckmann, line 1012: If we have multiple Regions within a MBB, the regular DAG constructor will create the correct order dependencies between the regions, right? 

bb.1: 
Region 1
Region 2
bb.2:
region 3

But the current algorithm gets Current Region and not the last one to build edges with. 
So either Region 1 or Region 2 get wrong edges. 


If we have a MBB with a conditional jump in Region 1 and an unconditional jump in region 2, I think this algorithm will be wrong
bb1: 
Region 1
jnz xxx, bb.3
---branch delay slots ----
Region 2
j bb.4
---branch delay slots ----
...
bb.3:
....
bb.4

So I think region 1 should only consider edges to (Region 2 + bb4) and BB3, while region 2 only needs a SuccEdges to bb.4

I don't think the current algorithm is correct for these two cases, especially since we work off of the currentRegion.

llvm/lib/Target/AIE/AIEMaxLatencyFinder.cpp
martien-de-jong: Currently, InterBlockScheduling uses numbers in other places. I've aligned with that. I'm fine with changing that globally, but perhaps not in this PR.

llvm/lib/Target/AIE/AIEInterBlockScheduling.cpp
martien-de-jong: AFAICS, empty blocks (and regions) are skipped by the scheduler, so they are not gathered. I think an InterBlockEdges with no successor nodes would be  perfectly legitimate. We can detect the case in  leaveBlock in IsGatheringPhase, and create an empty region an degenerate but correct PerSuccEdges. Furthermore, I   think that inserted dedicated epilogue might call on this function to do the just-in-time initialization as if it came from IsGatheringRegions.

llvm/lib/Target/AIE/AIEInterBlockScheduling.cpp
F-Stuckmann: Example 2 can happen if MachineBlockPlacement reshuffles some MBBs and removes a fallthrough MBB

martien-de-jong: Good catch. We should explicitly use the bottom region of this block. To wit, each PerSuccEdge is between a bottom region of the predecessor and the top region of the successor.

llvm/lib/Target/AIE/AIEBaseInstrInfo.cpp
mludevid: Can you share a reference? I still can't find any usages.

llvm/lib/Target/AIE/AIEInterBlockScheduling.cpp
mludevid: As far as I know this meyers const pattern is the common way to share code between const and non-const methods. But it's a nit, leave it as is if you like, I think the functions are small enough that duplicating the code is fine.

llvm/lib/Target/AIE/AIEInterBlockScheduling.cpp
mludevid: Yes, please verify this. Also, recursive successors would be awesome, but I think worth delaying for a potential follow-up.

martien-de-jong: To wit, DataDependenceHelper doesn't set a current block to the ScheduleDAGInstrs it creates. If that block is missing, MaxLatencyFinder returns MaxLatency. Hence, the exit edges of InterBlockEdges carry MaxLatency..
I will add an explicit test for this.

mludevid: Ah that might be right. I had missed that detail. Please add the test, but I agree this should be fine.

llvm/lib/Target/AIE/AIEInterBlockScheduling.cpp
```diff
@@ -832,20 +868,77 @@ BlockState &InterBlockScheduling::getBlockState(MachineBasicBlock *BB) {
...
+    // If Regions is empty, the block was empty during GatheringRegions (e.g.
+    // a newly-created dedicated exit block). The machine scheduler skips
+    // enterRegion for empty blocks so no region was captured. Create it now
+    // with correct free-instruction boundaries, excluding any fixed bundles.
+    if (BS.getRegions().empty()) {
+      const unsigned TopCount =
+          (RegionBegin == BB->begin()) ? BS.TopInsert.size() : 0u;
+      const unsigned BotCount =
+          (RegionEnd == BB->end()) ? BS.BottomInsert.size() : 0u;
+      BS.addRegion(BB, std::next(RegionBegin, TopCount),
```
F-Stuckmann: If we create Region when we create dedicatedExit blocks, then we can delete this whole code path.

llvm/lib/Target/AIE/AIEInterBlockScheduling.cpp
```diff
@@ -467,10 +480,21 @@ bool InterBlockScheduling::leaveBlock() {
...
   // to schedule next and with what parameters
   auto &BS = *CurrentBlockState;
+  if (IsGatheringPhase) {
+    // This is the first visit to this block. The region decomposition has been
+    // gathered. Now transition to Scheduling so the next pass actually
+    // schedules the gathered regions.
+    if (BS.Kind == BlockType::Loop) {
+      // For loops, also create the interblock edges between the top and the
+      // bottom region.
+      BS.initInterBlock(*Context, *HR);
```
F-Stuckmann, line 578: nit: can we rename this to BS.initLoopSchedule or even BS.trySetupPostPipeliner?

llvm/lib/Target/AIE/AIEMaxLatencyFinder.cpp
```diff
@@ -74,193 +74,118 @@ bool MaxLatencyFinder::isBottomRegion(MachineInstr *ExitMI) {
...
+      }
+
+      // For ExitSU the depth is the full length of the successor block's
+      // top region (all its cycles have elapsed before reaching ExitSU).
+      // For a regular instruction node the depth is its scheduled cycle
+      // within the block.
+      const int Depth = Succ->isBoundaryNode() ? SE.getPostRegionLength()
+                                               : SE.getPostDepthOr(Succ, 0);
+      const int EdgeLat = Dep.getSignedLatency();
+      const int Remaining = EdgeLat - Depth;
```
F-Stuckmann, line 137: We have this same calculation in llvm/lib/Target/AIE/AIEInterBlockScheduling.cpp:1238 and llvm/lib/Target/AIE/AIEInterBlockScheduling.cpp:614

llvm/lib/Target/AIE/AIEDataDependenceHelper.cpp
martien-de-jong: It does happen, in particular for Fixed nodes that enter in scheduled bundles. One of the lasts commits in this PR is dealing with Fixed nodes, putting the 'first iteration' copies into the PerSuccEdges, but there will still be some unrecorded nodes come in.

llvm/lib/Target/AIE/AIEDataDependenceHelper.cpp
```diff
@@ -72,4 +72,59 @@ void DataDependenceHelper::dumpDot(raw_ostream &OS,
...
+  IndexMap &TheMap = Boundary ? SuccMap : PredMap;
+  TheMap.emplace(MI, *Index);
+}
+
+void InterBlockEdges::markBoundary() {
+  assert(!Boundary.has_value());
+  Boundary = SUnits.size();
+}
+
+bool InterBlockEdges::mayAlias(SUnit *SUa, SUnit *SUb, bool TBAA) {
```
F-Stuckmann, line 90: nit: can we const this function and its arguments?

llvm/lib/Target/AIE/AIEDataDependenceHelper.cpp
```diff
@@ -72,4 +72,59 @@ void DataDependenceHelper::dumpDot(raw_ostream &OS,
...
+    const bool BIsPost = SUb->NodeNum >= *Boundary;
+    if (AIsPost != BIsPost)
+      return false;
+  }
+  return DataDependenceHelper::mayAlias(SUa, SUb, TBAA);
+}
+
+const SUnit *InterBlockEdges::getPreBoundaryNode(MachineInstr *MI) const {
+  const auto Found = PredMap.find(MI);
+  if (Found == PredMap.end()) {
```
F-Stuckmann, line 103: Can we refactor this Map finding and returning into its own private helper fucntions, since we duplicate code with recordPostDepth

llvm/lib/Target/AIE/AIEBaseInstrInfo.cpp
F-Stuckmann: yeah we can drop it now

llvm/lib/Target/AIE/AIEDataDependenceHelper.h
```diff
@@ -17,6 +17,9 @@
 
 #include "llvm/CodeGen/MachineScheduler.h"
 #include "llvm/CodeGen/ScheduleDAGInstrs.h"
+#include <limits>
```
F-Stuckmann: nit: can we remove limits import?

llvm/lib/Target/AIE/AIEDataDependenceHelper.h
```diff
@@ -53,6 +58,110 @@ class DataDependenceHelper : public ScheduleDAGInstrs {
...
+/// The class also provides a depth map (keyed by SUnit NodeNum) that represents
+/// the depth of an instruction after the boundary.
+///
+///   PostDepths — top-down cycle of each post-boundary node.  Populated by
+///   recordPostDepth(); queried by getPostDepth().
+///
+///   PostRegionLength — total number of bundles in the post-boundary region,
+///   used to represent the depth of the artificial ExitSU node.
+///
+/// In practice, SUnits and their dependences are invariant after first
```
F-Stuckmann, line 78: nit: dependencies

llvm/lib/Target/AIE/AIEDataDependenceHelper.cpp
martien-de-jong: It's overriding an existing interface. Supplying const to llvm types is limiting what you can do with it. This is not the first time I make this point.

llvm/lib/Target/AIE/AIEDataDependenceHelper.h
```diff
@@ -53,6 +58,110 @@ class DataDependenceHelper : public ScheduleDAGInstrs {
...
+  /// We maintain explicit maps to retrieve the corresponding SUnit.
+  using IndexMap = std::map<MachineInstr *, unsigned>;
+  IndexMap PredMap;
+  IndexMap SuccMap;
+
+  /// Depth (top-down cycle) of post-boundary SUnits, keyed by NodeNum.
+  std::map<unsigned, int> PostDepths;
+
+  /// Total number of bundles in the post-boundary region.
+  int PostRegionLength = 0;
```
F-Stuckmann: nit: Could we harmonize this with how we call it, i.e. PostRegionDepth. 
Length suggests to me that we are Querrying SuccMap.size().

llvm/lib/Target/AIE/AIEDataDependenceHelper.cpp
martien-de-jong: I will have a look. I assume you mean to pass in the map to search as a parameter?

llvm/lib/Target/AIE/AIEInterBlockScheduling.cpp
```diff
@@ -871,26 +935,138 @@ BlockState &InterBlockScheduling::getBlockState(MachineBasicBlock *BB) {
...
+        SE.recordPostDepth(SU.getInstr(), Depth);
+      }
+    } else {
+      int Cycle = 0;
+      for (const MachineBundle &Bundle : SBS.getTop().Bundles) {
+        for (MachineInstr *MI : Bundle.getInstrs())
+          SE.recordPostDepth(MI, Cycle);
+        ++Cycle;
+      }
+      SE.setPostRegionLength(Cycle);
```
F-Stuckmann: Do we even need to set the PostRegionDepth, if we also record it when we insert recordPostDepth ?

martien-de-jong: As discussed. We will drop this interface, we will track PostRegionMaxDepth, and we will add +1 in the caller to get to the depth of the next basic block

llvm/lib/Target/AIE/AIEInterBlockScheduling.h
```diff
@@ -240,25 +208,34 @@ class BlockState {
...
-  const InterBlockEdges &getBoundaryEdges() const {
-    assert(Kind == BlockType::Loop && BoundaryEdges);
-    return *BoundaryEdges;
+  /// Return the self-loop back-edge DAG from PerSuccEdges, or null if absent.
+  /// Populated by buildPerSuccEdges().
+  InterBlockEdges *getLoopSelfEdge();
+  const InterBlockEdges *getLoopSelfEdge() const;
+
+  /// Return the loop self-edge, asserting it is present.
+  InterBlockEdges &getBoundaryEdges() {
```
F-Stuckmann: Why is the assert here necessary? Couldn't we use `getLoopSelfEdge` directly?

llvm/lib/Target/AIE/AIEInterBlockScheduling.h
```diff
@@ -309,17 +286,6 @@ class BlockState {
...
-  Absolute,
-  // The bundles are accurate, but may shift at most one cycle
-  // due to alignment of a successor block
-  AccountForAlign,
-  // We don't have bundles for all successors
-  Conservative
-};
-
 class InterBlockScheduling {
   const MachineSchedContext *Context;
```
F-Stuckmann: nit: = nullptr; ?

llvm/lib/Target/AIE/AIEInterBlockScheduling.h
```diff
@@ -344,10 +316,12 @@ class InterBlockScheduling {
...
   /// latencies converged.
-  MachineInstr *latencyConverged(BlockState &BS) const;
+  MachineInstr *latencyConverged(BlockState &BS);
 
-  /// After finding the loops, determine the epilogue blocks.
-  void markEpilogueBlocks();
+  /// Derive the Kind for \p BS from the current CFG: Loop blocks are
+  /// returned unchanged; any other block is Epilogue if at least one
+  /// predecessor is a Loop block, and Regular otherwise.
+  BlockType classifyNonLoop(const BlockState &BS) const;
```
F-Stuckmann: nit: we are classifying every Block, not only loop blocks. use `classifyBlockState` ?

llvm/lib/Target/AIE/AIEInterBlockScheduling.cpp
```diff
@@ -871,26 +935,138 @@ BlockState &InterBlockScheduling::getBlockState(MachineBasicBlock *BB) {
...
+  const Region &Bot = BS.getBottom();
+  const bool SafeToIgnoreMemDeps = BS.isSafeToIgnoreMemDeps();
+
+  auto &PerSuccEdges = BS.getPerSuccEdges();
+
+  // When a specific successor is provided, only rebuild that one edge so that
+  // edges for unaffected successors are preserved and other successors with
+  // un-emitted BottomInsert clones are not touched.
+  if (UpdateSuccessor) {
+    auto It = llvm::find_if(PerSuccEdges, [UpdateSuccessor](const auto &SEPtr) {
```
F-Stuckmann: Can we replace the comment up top with a comment here saying:
// Erase UpdateSuccessor SuccEdge

llvm/lib/Target/AIE/AIEInterBlockScheduling.cpp
```diff
@@ -871,26 +935,138 @@ BlockState &InterBlockScheduling::getBlockState(MachineBasicBlock *BB) {
   return Blocks.at(BB);
 }
 
+void InterBlockScheduling::buildPerSuccEdges(
```
F-Stuckmann: I think this method would be structurally simpler to understand if SuccEdge had a proper constructor

InterBlockEdges(const MachineSchedContext &Context,
                  bool SafeToIgnoreMemDeps = false,
                  BlockState *Pred = nullptr,
                  MachineBasicBlock *Succ = nullptr);

So then method simplifies to: 

if (UpdateSuccessor) {
  deleteSuccEdge(UpdateSuccessor);
  PerSuccEdges.emplace_back(
        std::make_unique<InterBlockEdges>(C, SafeToIgnoreMemDeps, BS, UpdateSuccessor));
        return;
}


for (MachineBasicBlock *SuccBB : BB->successors()) 
  PerSuccEdges.emplace_back(
        std::make_unique<InterBlockEdges>(C, SafeToIgnoreMemDeps, BS, SuccBB));

llvm/lib/Target/AIE/AIEInterBlockScheduling.h
```diff
@@ -429,16 +400,44 @@ class InterBlockScheduling {
...
+
   std::optional<SWPEpilogueContext>
   getSWPEpilogueContext(MachineBasicBlock *MBB);
+
+  /// If \p CurrentBB is not already dedicated to \p LoopMBB (i.e. it has
+  /// more than one predecessor), split the edge from \p LoopMBB to
+  /// \p CurrentBB and return the new dedicated exit block. Otherwise returns
+  /// \p CurrentBB unchanged.
+  MachineBasicBlock *makeDedicatedLoopExit(MachineBasicBlock *LoopMBB,
+                                           MachineBasicBlock *CurrentBB);
```
F-Stuckmann: nit: CurrentMBB

llvm/lib/Target/AIE/AIEInterBlockScheduling.h
```diff
@@ -429,16 +400,44 @@ class InterBlockScheduling {
...
+
   AIEAlternateDescriptors &getSelectedAltDescs() { return SelectedAltDescs; }
 
+  const MachineSchedContext *getContext() const { return Context; }
+  bool isGatheringPhase() const { return IsGatheringPhase; }
+
   std::optional<SWPEpilogueContext>
   getSWPEpilogueContext(MachineBasicBlock *MBB);
+
+  /// If \p CurrentBB is not already dedicated to \p LoopMBB (i.e. it has
```
F-Stuckmann: maybe simpler: If \p LoopMBB is not the only Predecessor of \p CurrentMBB, create a dedicated Exit MBB

llvm/lib/Target/AIE/AIEInterBlockScheduling.cpp
```diff
@@ -909,19 +1085,29 @@ MachineBasicBlock *splitEdge(MachineBasicBlock *Pred, MachineBasicBlock *Succ) {
...
-// If the loop does not have a dedicated exit block, create one and return it.
-// If the current exit is already dedicated, simply return the current exit
-// instead.
-MachineBasicBlock *makeDedicatedLoopExit(MachineBasicBlock *LoopMBB,
-                                         MachineBasicBlock *CurrentBB) {
-  MachineBasicBlock *DedicatedExit = CurrentBB;
+} // namespace
+
+MachineBasicBlock *
+InterBlockScheduling::makeDedicatedLoopExit(MachineBasicBlock *LoopMBB,
```
F-Stuckmann, line 1129: nit: we could directly work on Blockstates here, since this method creates a fully set up Blockstate (except for the top fixed setting)

llvm/lib/Target/AIE/AIEInterBlockScheduling.h
martien-de-jong: We used to create BoundaryEdges on the spot. We now  reuse the PerSuccEdge, which could fail where the earlier version could not. of course we could have asserted in the caller as well, but this function connects the old local BoundaryEdges to the new implementation.

llvm/lib/Target/AIE/AIEInterBlockScheduling.h
martien-de-jong: Can do. But I don't have a default constructor, so it's not observable.

llvm/lib/Target/AIE/AIEDataDependenceHelper.cpp
martien-de-jong: Discussed offline. We would need to pass SuccMap or PredMap to a factored function, and still need to represent whether it was found or not. I'll leave it like this.

llvm/lib/Target/AIE/AIEDataDependenceHelper.h
martien-de-jong: I like dependence, an existing word, better. I'm also not changing to AIEDataDependencyHelper

llvm/lib/Target/AIE/AIEDataDependenceHelper.h
martien-de-jong: Done. Now it is PostRegionMaxDepth.

llvm/lib/Target/AIE/AIEInterBlockScheduling.h
martien-de-jong: Mais soit. It's also two years old....

llvm/lib/Target/AIE/AIEInterBlockScheduling.cpp
martien-de-jong: I will do it slightly different, because constructing the graph in InterBlockEdges requires BlockState internals.
I will separate the allocation of PerSuccEdges, then populate them with a separate InterBlockScheduling call.
When we need to reconstruct one, I clear that one and re-populate.

llvm/lib/Target/AIE/AIEInterBlockScheduling.cpp
```diff
@@ -276,10 +298,16 @@ void InterBlockScheduling::enterFunction(MachineFunction *MF) {
   if (LoopAware) {
     // Mark epilogues of the loops we found. This is only necessary if
     // we have created Loops in the first place, as indicated by LoopAware.
-    markEpilogueBlocks();
+    for (auto &[_, BS] : Blocks) {
```
F-Stuckmann, line 304: nit: i think this code path has become redundant with the new classifyBlock method

llvm/lib/Target/AIE/AIEInterBlockScheduling.cpp
```diff
@@ -480,6 +497,12 @@ class PipelineExtractor : public PipelineScheduleVisitor {
     MachineInstr *ToBeEmitted =
         InLoop ? MI : Loop.TheBlock->getParent()->CloneMachineInstr(MI);
     CurrentBundle.add(ToBeEmitted);
+
+    // Record the first-iteration prologue clone for each original instruction.
+    // try_emplace only inserts when the key is absent, so only the first
+    // occurrence (i.e. the first-iteration copy) is kept.
```
martien-de-jong, line 503: Note that we iterate in time order, not in iteration order. Regardless of that, the first copy of every instruction is the first-iteration copy, 
S0            // first iteration copies of S0
S1 S0       // first iteration copies of S1
S2 S1 S0  // first iteration copies of S2

llvm/lib/Target/AIE/AIEInterBlockScheduling.h
```diff
@@ -263,9 +212,23 @@ class BlockState {
   const Region &getTop() const { return Regions.back(); }
   Region &getTop() { return Regions.back(); }
   const Region &getBottom() const { return Regions.front(); }
+  /// Return the self-loop back-edge DAG from PerSuccEdges, or null if absent.
+  /// Populated by buildPerSuccEdges().
+  InterBlockEdges *getLoopSelfEdge();
+  const InterBlockEdges *getLoopSelfEdge() const;
+
+  /// Return the loop self-edge, asserting it is present.
   const InterBlockEdges &getBoundaryEdges() const {
```
mludevid: Nit: I find the preexisting name confusing. There is no mention of self-edges nor loops in the name. `getBoundaryEdges` sounds to me like a getter function for all edges to the successors.

F-Stuckmann (dismissed):

llvm/lib/Target/AIE/AIEInterBlockScheduling.cpp
```diff
@@ -849,6 +890,25 @@ void InterBlockScheduling::defineSchedulingOrder(MachineFunction *MF) {
 }
 
 MachineBasicBlock *InterBlockScheduling::nextBlock() {
+  if (IsGatheringPhase) {
+    if (NextInOrder < MBBSequence.size()) {
```
mludevid, line 935: Could this be an assert? How can this even happen? Silently skipping blocks doesn't seem the right thing to do if we ever run out of space in `MBBSequence`.

llvm/lib/Target/AIE/AIEDataDependenceHelper.cpp
```diff
@@ -111,17 +111,26 @@ bool InterBlockEdges::isPostBoundaryNode(SUnit *SU) const {
...
 
 void InterBlockEdges::clearPostDepths() {
-  Boundary = {};
   PostDepths.clear();
   PostRegionMaxDepth = 0;
 }
 
 void InterBlockEdges::clear() {
   clearDAG();
+  Boundary = {};
```
mludevid, line 120: Nit: Could you move this fixup to the commit that initially creates the clear function?

llvm/lib/Target/AIE/AIEDataDependenceHelper.cpp
```diff
@@ -111,17 +111,26 @@ bool InterBlockEdges::isPostBoundaryNode(SUnit *SU) const {
...
 
+void InterBlockEdges::recordPostDepth(int Depth) {
+  PostRegionMaxDepth = std::max(PostRegionMaxDepth, Depth);
+}
+
 void InterBlockEdges::recordPostDepth(MachineInstr *MI, int Depth) {
+  // Even if we can't find an index, our caller has found evidence
+  // that we have code at this depth. As a common example,
+  // consider delay slot bundles, that carry nops without index.
+  recordPostDepth(Depth);
```
mludevid, line 135: Nit: Can you move this refactoring to the commit where you created the recordPostDepth function?

llvm/lib/Target/AIE/AIEDataDependenceHelper.h
```diff
@@ -53,6 +57,116 @@ class DataDependenceHelper : public ScheduleDAGInstrs {
...
+/// pre/post boundary are suppressed via a mayAlias() override.
+///
+/// The class also provides a depth map (keyed by SUnit NodeNum) that represents
+/// the depth of an instruction after the boundary.
+///
+///   PostDepths — top-down cycle of each post-boundary node.  Populated by
+///   recordPostDepth(); queried by getPostDepth().
+///
+///   PostRegionLength — total number of bundles in the post-boundary region,
+///   used to represent the depth of the artificial ExitSU node.
```
mludevid: Documentation is not up to date anymore.

llvm/lib/Target/AIE/AIEInterBlockScheduling.h
```diff
@@ -429,16 +406,40 @@ class InterBlockScheduling {
...
+  /// Update PerSuccEdges of BB for the successor \p For
+  void updatePerSuccEdges(MachineBasicBlock *BB, MachineBasicBlock *For);
+
+  void buildGraph(InterBlockEdges &);
+
+  /// Clear and repopulate the PostDepths of every per-successor inter-block
+  /// edge for BB. For scheduled successors, records the actual scheduled cycle
+  /// of each instruction; for unscheduled ones, computes a static lower bound
+  /// from the inter-block DDG topology. Called from initializeBotScoreBoard
+  /// and MaxLatencyFinder::recordPerSuccDepths so both are self-contained.
```
mludevid, line 419: Documentation is out of date.

mludevid (commented):
Please update commit message of "[AIE][InterBlock] Compute MaxLatency from InterBlockEdgese". It has a typo and the additional commit information still mentions not giving any guarantees about correctness.

llvm/lib/Target/AIE/AIEInterBlockScheduling.h
martien-de-jong: Ok. I drop the wrapper, make a direct call.

llvm/lib/Target/AIE/AIEInterBlockScheduling.cpp
martien-de-jong: No, this is by design.  NextInOrder counts through the basic blocks. As soon as we reach the end, we reset it, leave the gathering phase and run into the scheduling phase.

llvm/lib/Target/AIE/AIEInterBlockScheduling.h
martien-de-jong: I removed getBoundaryEdges, inlining it.

llvm/lib/Target/AIE/AIEInterBlockScheduling.cpp
martien-de-jong: This is totally refactored. It now retains the DAG structure, just clears it, and updating a successor has become a separate function

F-Stuckmann (approved):

PR #1008, opened 2026-05-26
[AIE] Add combine_lag_ptr_add to AIE2PS and AIE2P

katerynamuts:
Add the lag-ptr-add pointer combine for AIE2PS and AIE2P targets.

The combine rewrites a G_PTR_ADD of a loop pre-increment pointer in
terms of the corresponding POSTINC instruction's post-increment def,
avoiding a redundant add:
- G_PTR_ADD %pre, C  →  COPY %post      (when C == step)
- G_PTR_ADD %pre, C  →  G_PTR_ADD %post, (C - step)  (otherwise)


llvm/lib/Target/AIE/AIECombinerHelper.cpp
```diff
@@ -1544,6 +1544,127 @@ bool llvm::matchChainedPtrAddWithNonConstOffsets(MachineInstr &MI,
...
+  auto [PreReg, C] = *Root;
+
+  for (MachineInstr &UseMI : MRI.use_nodbg_instructions(PreReg)) {
+    // UseMI (the POSTINC) must come before MI (the G_PTR_ADD root) in program
+    // order.
+    if (!Helper.dominates(UseMI, MI))
+      continue;
+    // Find a POSTINC instruction with PreReg as base pointer.
+    auto MaybeInfo = TII.getPtrPostIncOpInfo(UseMI.getOpcode());
+    if (!MaybeInfo)
```
niwinanto, line 1617: Nit: PtrPostIncOpInfo make it more easy to understand

llvm/lib/Target/AIE/AIECombinerHelper.cpp
```diff
@@ -1544,6 +1544,127 @@ bool llvm::matchChainedPtrAddWithNonConstOffsets(MachineInstr &MI,
...
+//   * require a compile-time constant offset on operand 2
+// On success, returns {PreReg (pre-increment register, operand 1),
+//                      C (constant offset)}; std::nullopt otherwise.
+static std::optional<std::pair<Register, int64_t>>
+getLagRootPreReg(const MachineInstr &MI, MachineRegisterInfo &MRI) {
+  Register DefReg = MI.getOperand(0).getReg();
+  for (const MachineInstr &UseMI : MRI.use_nodbg_instructions(DefReg)) {
+    if (UseMI.mayLoadOrStore() && !UseMI.hasUnmodeledSideEffects())
+      return std::nullopt;
+  }
```
niwinanto, line 1558: Do we need this block in this helper function? Can be part of the main loop in the combine where it checks useMI for other legalities.

llvm/lib/Target/AIE/AIECombinerHelper.cpp
```diff
@@ -1544,6 +1544,127 @@ bool llvm::matchChainedPtrAddWithNonConstOffsets(MachineInstr &MI,
...
+// rooted siblings.
+bool llvm::matchLagPtrAdd(MachineInstr &MI, MachineRegisterInfo &MRI,
+                          const AIEBaseInstrInfo &TII, CombinerHelper &Helper,
+                          LagPtrAddMatchInfo &MatchInfo) {
+  assert(MI.getOpcode() == TargetOpcode::G_PTR_ADD && "Expected G_PTR_ADD");
+
+  auto Root = getLagRootPreReg(MI, MRI);
+  if (!Root)
+    return false;
+  auto [PreReg, C] = *Root;
```
niwinanto, line 1608: Nit: I feel inlining this helper is more readable, what do you think. What do you think. Would be nice to update the name of variables, it can really helps for readability. 
```
  const Register PreReg = MI.getOperand(1).getReg();
  auto Step = getIConstantVRegValWithLookThrough(MI.getOperand(2).getReg(), MRI);
  if (!Step )
    return false;
  const int64_t StepSize = Step->Value.getZExtValue();
```

llvm/lib/Target/AIE/AIECombinerHelper.cpp
```diff
@@ -1544,6 +1544,127 @@ bool llvm::matchChainedPtrAddWithNonConstOffsets(MachineInstr &MI,
...
+        UseMI.getOperand(MaybeInfo->PrePtrUseIdx).getReg() != PreReg)
+      continue;
+    // Step must be a compile-time constant.
+    auto MaybeStep = getIConstantVRegValWithLookThrough(
+        UseMI.getOperand(MaybeInfo->ModifierUseIdx).getReg(), MRI);
+    if (!MaybeStep)
+      continue;
+    // Replacement offset NewC = C - step must fit in the original offset
+    // type; otherwise the G_CONSTANT built in apply silently wraps.
+    const int64_t S = MaybeStep->Value.getSExtValue();
```
niwinanto, line 1629: Nit: PostIncStepSize may be?

llvm/test/CodeGen/AIE/aie2ps/GlobalIsel/global-combiners/lag-reg-add-coalesce.mir
niwinanto, line 12: Nit: you can remove `--aie-global-ptr-mod-opt=true`, is enabled by default

niwinanto (commented):
In general PR looks good to me, few nits. How ever I am wondering why these cases are not really covered by `AIE Pointer Modifier Optimization`. @F-Stuckmann any thoughts?

F-Stuckmann:
This issue surfaces because `llvm/lib/Target/AIE/AIEClusterBaseAddress.cpp` currently forms pointer chains only within a single MBB. Previously, no pointer chains were formed at all when the base pointer was shared across multiple MBBs; I changed that so chains are now built within each MBB even when the base pointer is shared.

What's still missing is the selection step: rather than always chaining off the given base pointer, we need logic to decide which base pointer to use (or reuse) when several candidates exist.

katerynamuts:
This combiner is not needed when https://github.com/Xilinx/llvm-aie/pull/1013 is merged.

PR #1009, opened 2026-05-28
[AIEX] Implement foldImmediate Peephole optimization

khallouh:
Fold move-immediate + COPY pairs into a single move-immediate at the
consumer site:
    %c = MOVA 42       →    %a = MOVA 42
    %a = COPY %c
Guarded by hasOneNonDBGUse to avoid materializing constants multiple
times when they have several consumers, which would inflate register
pressure and perturb allocation.


khallouh:
Top improvements Topk1D_int8_0, LayerNormC8Part2_aie2_int8_0 and SinBf16_0 (7%, 5% and 3% respectively in Instruction Count) There are a few regressions but most are small, highest being 1.5% in BufferPadInnermost. These are mostly due to side effects on decisions made by the register allocator, so not really easily recoverable.
image
Edit: Topk1D_int8_0 number is not reliable because its runtime takes random paths in each run, which brings down our top improvement to 5% in LayerNormC8Part2_aie2_int8_0

llvm/lib/Target/AIE/AIEBaseInstrInfo.cpp
```diff
@@ -539,6 +555,65 @@ unsigned AIEBaseInstrInfo::getAIEMachineBundleSize(
...
+  // DefMI alive for other consumers and end up materializing the constant
+  // twice, inflating register pressure.
+  if (AIEFoldImmRequireOneUse && !MRI->hasOneNonDBGUse(Reg)) {
+    ++NumFoldImmBlockedMultiUse;
+    return false;
+  }
+
+  // Get the appropriate move-immediate opcode for the destination register
+  APInt ImmAPInt(32, ImmVal, /*isSigned=*/true);
+  unsigned NewOpc = getConstantMovOpcode(*MRI, DstReg, ImmAPInt);
```
martien-de-jong: Theoretically this could fail I think. Should we detect that and bail out?

llvm/test/CodeGen/AIE/aie2/GlobalISel/bfloat.ll
```diff
@@ -4,7 +4,7 @@
 ; See https://llvm.org/LICENSE.txt for license information.
 ; SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
 ;
-; (c) Copyright 2023-2024 Advanced Micro Devices, Inc. or its affiliates
+; (c) Copyright 2023-2026 Advanced Micro Devices, Inc. or its affiliates
```
martien-de-jong: Only copyright change?

llvm/lib/Target/AIE/AIEBaseInstrInfo.cpp
khallouh: I changed getConstantMovOpcode() from returning unsigned to returning
std::optional<unsigned> to allow callers to handle cases where no
suitable opcode exists for a given register class/immediate combination.

llvm/lib/Target/AIE/AIEBaseInstructionSelector.cpp
```diff
@@ -1059,7 +1059,8 @@ bool AIEBaseInstructionSelector::selectStartLoop(MachineInstr &I,
   if (auto Const =
           getIConstantVRegValWithLookThrough(I.getOperand(2).getReg(), MRI)) {
     auto OpCode = TII.getConstantMovOpcode(MRI, DstReg, Const->Value);
-    auto Mov = MIB.buildInstr(OpCode, {DstReg}, {})
+    assert(OpCode && "Failed to get constant mov opcode during ISel");
```
martien-de-jong, line 1062: Check: this would be a legilization/regbankselect failure.

martien-de-jong (dismissed):

llvm/lib/Target/AIE/AIEBaseInstrInfo.cpp
niwinanto, line 1: Please add some unit tests

llvm/lib/Target/AIE/AIEBaseInstrInfo.cpp
```diff
@@ -539,6 +555,65 @@ unsigned AIEBaseInstrInfo::getAIEMachineBundleSize(
...
+    return false;
+
+  // Check if DefMI is a move-immediate instruction
+  if (!isIConst(DefMI.getOpcode()))
+    return false;
+
+  // Get the immediate value from DefMI
+  // Move-immediate instructions have the format: Dst = MOV Imm
+  // The immediate is typically in operand 1
+  if (DefMI.getNumOperands() < 2 || !DefMI.getOperand(1).isImm())
```
niwinanto: For `isIConst` this has to be a hard assert.

khallouh: No, PeepholeOptimizer uses a different criterion: It populates ImmDefMIs based on MI.isMoveImmediate() (the TableGen isMoveImm flag) OR TII->getConstValDefinedInReg():
If an opcode has isMoveImm = true in TableGen but isn't in isIConst(), the peephole optimizer will still call foldImmediate with it. The soft return handles this gracefully.

khallouh: The control register moves or the hardware loop setup instructions for example they are not added to `isIConst() ` and we don't want to try folding them because the QoR impact is unclear. For an assert to not break our unit tests I would have to add those to the  isIConst(). 
If you try it later and it helps we can add all the isMoveImm instructions to the isIConst() function and assert here but that can be a follow-up

khallouh: In that case might as well remove the isIConst() function and check for isMoveImm in call sites

khallouh: Oh, did you mean asserting this? DefMI.getNumOperands() < 2 || !DefMI.getOperand(1).isImm() 
If yes, that makes sense, I will just remove it it will fail in any case if that's not true.

martien-de-jong (dismissed):

llvm/lib/Target/AIE/AIEBaseInstrInfo.cpp
khallouh: Actually, even asserting or removing this `DefMI.getNumOperands() < 2 || !DefMI.getOperand(1).isImm()` doesn't work. If we move a symbol for example it will be IConst but not MoveImm. I had crashes in tests where we move a symbol. I restored my initial change.

khallouh:
@niwinanto I tried addressing your comment but it doesn't work when we move symbol, I restored my initial change. I also added the mir unit test. Could you have a look again?

llvm/test/CodeGen/AIE/aie2ps/constant-folding.mir
```diff
@@ -21,15 +22,13 @@ body: |
...
+    ; CHECK-NEXT: [[MOV2:%[0-9]+]]:er = MOV_RLC_imm11_pseudo 42
+    ; CHECK-NEXT: $r0 = COPY [[MOV2]]
     ; CHECK-NEXT: PseudoRET implicit $lr, implicit $r0
     %1:mlockid_reg = COPY $r1
-    %2:er = MOV_RLC_imm11_pseudo 0
-    %3:ms2 = COPY %2:er
-    %0:mlockid_reg = MOV_alu_mv_alu_fx2flt %1:mlockid_reg, %3:ms2, implicit-def dead $srfpcnvfx2fl, implicit $crfpcnvfx2flmask
-    $r0 = COPY %0:mlockid_reg
+    %2:er = MOV_RLC_imm11_pseudo 42
+    %3:er = COPY %2:er
```
niwinanto: Why did you update the test input, this was one of the patterns I would like to have a test. Meaning, cross bank copy and that being used in some real instruction.

niwinanto: Also, could you extend the coverage, as there are many combinations and encodable bit constrains between the constants of register classes. Would be nice to add couple of such cases.

khallouh: The combine behaves the same whether the final consumer is a copy or a target instruction. Anyway, I updated to use a real instruction and added one more test with a different register class and a different immediate value. As for the immediate encodings we rely on `getConstantMovOpcode` which is used in a few places and should be tested enough already.

llvm/lib/Target/AIE/AIEBaseInstrInfo.cpp
niwinanto: `DefMI.getNumOperands() < 2` is already valid, but I see you have updated it already.

llvm/test/CodeGen/AIE/aie2ps/constant-folding.mir
niwinanto: Previous version was not testing a cross bank copy being removed as you updated the input. Current test includes it.

niwinanto (approved):
LGTM!

llvm/test/CodeGen/AIE/aie2ps/constant-folding.mir
niwinanto: And I expect, cross bank case is what going to benefit us as same bank copies will be optimized/coalesced away

niwinanto: These are some cross bank copy examples I tried
```
---
---
name: test_imm_constant_no_fold_large_imm_into_ptr
legalized: true
tracksRegLiveness: true
body: |
  bb.1.entry:
    %0:er = MOVXM_lng_cg 70000
    %1:ep = COPY %0:er
    $p0 = COPY %1:ep
    PseudoRET implicit $lr, implicit $p0
...
name: test_imm_control_reg
legalized: true
tracksRegLiveness: true
body: |
  bb.1.entry:
    %0:er = MOV_RLC_imm11_pseudo 5
    %1:mcrsat = COPY %0:er
    $crsat = COPY %1:mcrsat
    PseudoRET implicit $lr, implicit $crsat
...
```
I think we need more guards while folding cross-bank copies.

khallouh: As discussed offline, regarding the verifier errors on these examples (see below). I think the code is still correct, the verifier should not be complaining about `eP` used in `MOVXM_lng_cg `instead of `mMvSclDstCg` because `eP` is a subclass of `mMvSclDstCg`. Let's fix the verifier in a follow-up.
```
*** Bad machine code: Illegal virtual register for instruction ***
- function:    test_imm_constant_no_fold_large_imm_into_ptr
- basic block: %bb.0 entry (0x5d25114bfec8)
- instruction: %1:ep = MOVXM_lng_cg 70000
- operand 0:   %1:ep
Expected a mMvSclDstCg register, but got a eP register
LLVM ERROR: Found 1 machine code errors.
```

khallouh: You're actually right, the verifier shouldn't complain about eP but your other example of cross bank copy to control register might actually generate incorrect code, I'll add more guards.

khallouh: The bug was in getConstantMovOpcode https://github.com/Xilinx/llvm-aie/pull/1023

PR #1014, opened 2026-06-01
[AIEX] Recursively order nested epilogues in Phase 2 scheduling.

katerynamuts:
When an upstream epilogue's forward post_order walk reaches a nested epilogue
inside another pipelined loop whose epilogue branches back to its prologue,
the prologue can be pushed after that epilogue and freeze the wrong MBBSequence.
Process nested epilogues with their own root and guard against re-entrancy so
loop-back successors are scheduled before epilogue scheduling.

### DIFF
--- a/llvm/lib/Target/AIE/AIEInterBlockScheduling.cpp
+++ b/llvm/lib/Target/AIE/AIEInterBlockScheduling.cpp
@@ -475,13 +475,13 @@ public:
       : IBS(IBS), Emitted(Emitted), Processing(Processing), Push(Push) {}
 
   void operator()(MachineBasicBlock *EpilogueRoot) const {
-    if (!Processing.insert(EpilogueRoot).second)
+    if (Emitted.contains(EpilogueRoot) ||
+        !Processing.insert(EpilogueRoot).second)
       return;
 
     for (MachineBasicBlock *Sub : post_order(EpilogueRoot)) {
       if (Sub != EpilogueRoot &&
-          IBS.getBlockState(Sub).Kind == BlockType::Epilogue &&
-          !Emitted.contains(Sub) && !Processing.contains(Sub)) {
+          IBS.getBlockState(Sub).Kind == BlockType::Epilogue) {
         (*this)(Sub);
       }
       Push(Sub);
@@ -826,8 +826,6 @@ void InterBlockScheduling::defineSchedulingOrder(MachineFunction *MF) {
 
   for (MachineBasicBlock *MBB : post_order(MF)) {
     if (getBlockState(MBB).Kind != BlockType::Epilogue)
-      continue;
-    if (Emitted.contains(MBB))
       continue;
     ProcessEpilogue(MBB);
   }
--- a/llvm/lib/Target/AIE/AIEInterBlockScheduling.cpp
+++ b/llvm/lib/Target/AIE/AIEInterBlockScheduling.cpp
@@ -784,11 +784,41 @@ void InterBlockScheduling::defineSchedulingOrder(MachineFunction *MF) {
   // post-order. post_order(E) ends with E itself, so E is scheduled AFTER
   // everything reachable from it (its successors are already done -> precise
   // inter-block latency for E) and BEFORE its non-loop predecessors.
+  //
+  // When an outer epilogue's forward walk reaches a nested epilogue inside a
+  // cyclic region, post-order rooted at the outer epilogue can emit the inner
+  // epilogue before its prologue (loop-back). Recursively process nested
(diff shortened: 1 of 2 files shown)


llvm/lib/Target/AIE/AIEInterBlockScheduling.cpp
```diff
@@ -784,11 +784,41 @@ void InterBlockScheduling::defineSchedulingOrder(MachineFunction *MF) {
   // post-order. post_order(E) ends with E itself, so E is scheduled AFTER
   // everything reachable from it (its successors are already done -> precise
   // inter-block latency for E) and BEFORE its non-loop predecessors.
+  //
+  // When an outer epilogue's forward walk reaches a nested epilogue inside a
+  // cyclic region, post-order rooted at the outer epilogue can emit the inner
+  // epilogue before its prologue (loop-back). Recursively process nested
+  // epilogues with their own root so post_order picks the correct order.
+  SmallPtrSet<const MachineBasicBlock *, 16> Processing;
+  struct EpilogueProcessor {
```
F-Stuckmann: nit: can this be a class?

llvm/lib/Target/AIE/AIEInterBlockScheduling.cpp
```diff
@@ -784,11 +784,41 @@ void InterBlockScheduling::defineSchedulingOrder(MachineFunction *MF) {
...
+            IBS.getBlockState(Sub).Kind == BlockType::Epilogue &&
+            !Emitted.contains(Sub) && !Processing.contains(Sub)) {
+          (*this)(Sub);
+        }
+        Push(Sub);
+      }
+
+      Processing.erase(EpilogueRoot);
+    }
+  } ProcessEpilogue{*this, Emitted, Processing, Push};
```
F-Stuckmann: nit: this is a class boundary, can we have a separation here? 
Additionally Can we move the Class definition outside of `defineSchedulingOrder` ?

llvm/lib/Target/AIE/AIEInterBlockScheduling.cpp
```diff
@@ -784,11 +784,41 @@ void InterBlockScheduling::defineSchedulingOrder(MachineFunction *MF) {
...
+
   for (MachineBasicBlock *MBB : post_order(MF)) {
     if (getBlockState(MBB).Kind != BlockType::Epilogue)
       continue;
-    for (MachineBasicBlock *Sub : post_order(MBB))
-      Push(Sub);
+    if (Emitted.contains(MBB))
```
F-Stuckmann: check: outside of ProcessEpilogue we guard reentry of ProcessEpilogue, inside we also guard it. I think we can simplify this.

F-Stuckmann: This would do the trick: 
```
void operator()(MachineBasicBlock *EpilogueRoot) const {
      if (!Processing.insert(EpilogueRoot).second)
        return;
      if (Emitted.contains(MBB)) 
         return; 
    ........
}
```
Then you could also drop the condition in the for loop if `Emitted.contains(MBB)`.

llvm/lib/Target/AIE/AIEInterBlockScheduling.cpp
katerynamuts: it has only public data members and a public operator(), so I think struct is more appropriate here.

llvm/lib/Target/AIE/AIEInterBlockScheduling.cpp
katerynamuts: you're right, it is more consistent with the rest of the file. After moving outside, I've also changed struct to class.

llvm/lib/Target/AIE/AIEInterBlockScheduling.cpp
katerynamuts: good suggestion, thanks. Done

katerynamuts:
@F-Stuckmann thank you for reviewing the PR, I addressed all your comments and pushed fixup commits.

llvm/lib/Target/AIE/AIEInterBlockScheduling.cpp
martien-de-jong: Opinion: 'struct' exists in C++ for C compatibility only. A struct can be expressed in C++ by `class X { public:`

martien-de-jong (approved):
Perhaps we can revisit this later with a more direct optimisation metric, which is to maximize the flow through edges that have their successors scheduled when scheduling the predecessors. I don't think it will gain much performance, but it would clarify the algorithm.

PR #1015, opened 2026-06-01
[AIE] Fix scoreboard crash

martien-de-jong:
This fixes the scoreboard crash. I haven't been able to reproduce it yet on aie2ps. It requires a large basic block, and in the original it achieved that by spilling a large register in many parts.


mludevid (approved):
LGTM. If you are able to create a small mir reproducer that triggers the bug, that'd be nice.

PR #1017, opened 2026-06-05
[AIEX] Reject ZOL candidates that lower to libcalls

niwinanto:
Extend isAllowedInZOL and isLoweredToCall to block integer div/rem, i64 mul, frem, and f64 minnum/maxnum, and add no-rt-calls.ll coverage.

### DIFF
--- a/llvm/lib/Target/AIE/AIEBaseTargetTransformInfo.cpp
+++ b/llvm/lib/Target/AIE/AIEBaseTargetTransformInfo.cpp
@@ -110,6 +110,13 @@ bool AIETTICommon::isLoweredToCall(const Function *F) {
     return true;
   }
 
+  // fmin/fmax are legal for f32 but lower to a libcall for f64 (double).
+  if ((F->getIntrinsicID() == Intrinsic::minnum ||
+       F->getIntrinsicID() == Intrinsic::maxnum) &&
+      F->getReturnType()->getScalarType()->isDoubleTy()) {
+    return true;
+  }
+
   return !F->isIntrinsic();
 }
 
@@ -154,11 +161,22 @@ bool AIETTICommon::isAllowedInZOL(Instruction &I) {
   }
 
   case Instruction::FDiv:
+  case Instruction::FRem:
   case Instruction::FPToSI:
   case Instruction::FPToUI:
   case Instruction::UIToFP:
   case Instruction::SIToFP:
     return false;
+  case Instruction::SDiv:
+  case Instruction::UDiv:
+  case Instruction::SRem:
+  case Instruction::URem:
+    return false;
+  case Instruction::Mul:
+    // i32 MUL is legal; i64 MUL lowers to a __muldi3 libcall.
+    if (!Ty->isVectorTy() && Ty->getScalarType()->getScalarSizeInBits() > 32)
+      return false;
+    break;
   }
   return true;
 }
--- a/llvm/test/CodeGen/AIE/aie2ps/zol/no-rt-calls.ll
+++ b/llvm/test/CodeGen/AIE/aie2ps/zol/no-rt-calls.ll
@@ -5,7 +5,7 @@
 ; SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
 ;
 ; (c) Copyright 2025-2026 Advanced Micro Devices, Inc. or its affiliates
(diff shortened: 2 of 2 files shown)


llvm/test/CodeGen/AIE/aie2p/zol/no-rt-calls.ll
```diff
@@ -4,7 +4,7 @@
 ; See https://llvm.org/LICENSE.txt for license information.
 ; SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
 ;
-; (c) Copyright 2025 Advanced Micro Devices, Inc. or its affiliates
+; (c) Copyright 2025-2026 Advanced Micro Devices, Inc. or its affiliates
 ; RUN: llc -O2 --mtriple=aie2p %s -o - --stop-before=irtranslator | \
```
khallouh, line 8: Can we run this for aie2ps?

llvm/lib/Target/AIE/AIEBaseTargetTransformInfo.cpp
```diff
@@ -154,11 +161,22 @@ bool AIETTICommon::isAllowedInZOL(Instruction &I) {
...
+  case Instruction::FRem:
   case Instruction::FPToSI:
   case Instruction::FPToUI:
   case Instruction::UIToFP:
   case Instruction::SIToFP:
     return false;
+  case Instruction::SDiv:
+  case Instruction::UDiv:
+  case Instruction::SRem:
+  case Instruction::URem:
```
martien-de-jong, line 173: Better safe than sorry, I agree, but dependent on where we run in the pipeline this may also flag powers-of-two cases like if (i % 4 == 0) as illegal,

llvm/lib/Target/AIE/AIEBaseTargetTransformInfo.cpp
```diff
@@ -154,11 +161,22 @@ bool AIETTICommon::isAllowedInZOL(Instruction &I) {
...
+    return false;
+  case Instruction::Mul:
+    // i32 MUL is legal; i64 MUL lowers to a __muldi3 libcall.
+    if (Ty->getScalarType()->isIntegerTy(64))
```
martien-de-jong: Can we write this as 'size > 32' ?

niwinanto: Done!

llvm/lib/Target/AIE/AIEBaseTargetTransformInfo.cpp
niwinanto: Sure, this make sense, but I can see  InstCombine is handling such patterns. May be this wont occur in practice.

llvm/lib/Target/AIE/AIEBaseTargetTransformInfo.cpp
```diff
@@ -110,6 +110,13 @@ bool AIETTICommon::isLoweredToCall(const Function *F) {
     return true;
   }
 
+  // fmin/fmax are legal for f32 but lower to a libcall for f64 (double).
+  if ((F->getIntrinsicID() == Intrinsic::minnum ||
+       F->getIntrinsicID() == Intrinsic::maxnum) &&
```
martien-de-jong, line 115: I find these names strange. Are they straight float min and max?

niwinanto: True, that's how it is. No llvm opcode, we need to use the intrinsic, but have gMIR opcode

```
G_FMINNUM
^^^^^^^^^

Perform floating-point minimum on two values.

In the case where a single input is a NaN (either signaling or quiet),
the non-NaN input is returned.

The return value of (FMINNUM 0.0, -0.0) could be either 0.0 or -0.0.

G_FMAXNUM
^^^^^^^^^

Perform floating-point maximum on two values.

In the case where a single input is a NaN (either signaling or quiet),
the non-NaN input is returned.

The return value of (FMAXNUM 0.0, -0.0) could be either 0.0 or -0.0.
```

martien-de-jong: Ah, perhaps 'num' for "not-NotANumber', resolving an unordered case.

martien-de-jong (approved):

PR #1019, opened 2026-06-08
[AIEX] Skip spill slot decomposition for coalesced cross-reg slots

niwinanto:


llvm/lib/Target/AIE/AIESpillSlotOptimization.cpp
F-Stuckmann: nit: since we now only allow same sized StackSlots, can we assert that we never increase here?

F-Stuckmann, line 349: nit: I think the rejection of the  Creation of a Stackslot should be performed here, because we already reject SpillPoints, if we cannot reduce the size, i.e. all subregs are live and recreating them would not gain us anything.

F-Stuckmann (commented):
I wonder if we would not run into this issue if we would switch the ordering of StackColoring and this pass, because StackColoring tries to merge StackSlots with non overlapping liveranges.

llvm/lib/Target/AIE/AIESpillSlotOptimization.cpp
niwinanto: As I understands, `SizeReduction` check works on one register at a time, so it fits in the SpillPoint constructor. The conflict check is different: one register's lanes never land on the same offset, so a conflict can only show up when StackSlotColoring puts two different registers on the same stack slot. The constructor only sees one register, so it can't know about that. We only have the full picture after analysis, when all spills and reloads for the slot are collected in FrameIndexInfo — that's why the check runs there instead.

niwinanto: Done!

niwinanto:
> I wonder if we would not run into this issue if we would switch the ordering of StackColoring and this pass, because StackColoring tries to merge StackSlots with non overlapping liveranges.

Agreed, switching the order would likely avoid this. If this pass runs before StackSlotColoring, it would decompose each spill on its own frame index, so StackSlotColoring would then be merging already-decomposed, sub-register-sized slots and the overlap we hit here wouldn't occur. It's a meaningfully different path though — this pass currently assumes coloring has already run — so it needs a thorough performance analysis. Worth trying as a follow-up.

llvm/lib/Target/AIE/AIESpillSlotOptimization.cpp
F-Stuckmann, line 460: supernit: don't we already exit, if we cannot profitably decompose all the Spills?

llvm/lib/Target/AIE/AIESpillSlotOptimization.cpp
```diff
@@ -519,22 +521,54 @@ void AIESpillSlotOptimization::processReloadsForSlot(
...
-          return SP->DecomposeProfitable;
-        });
-    if (!AllSpillsProfitable)
+    // Decompose all-or-nothing, and only when safe and profitable:
+    //  - every spill must be profitable, else a partial rewrite splits the slot
+    //    across the original and replacement objects; and
+    //  - no coalesced lanes may conflict, which would give the replacement slot
+    //    the wrong size.
+    const bool CanDecomposeSlot =
+        llvm::all_of(Info.Spills,
```
F-Stuckmann, line 563: supernit: since we already refactored `hasConflictingLanes` out, can we also have a function call for `AllSpillsAreProfitable`?

F-Stuckmann (approved):
Let's see if we have any catastrophic QoR regressions, but otherwise only supernits left :)

niwinanto:
@F-Stuckmann Thank you so much for running the mllib. As per the mllib run, there is no performance regression.

PR #1022, opened 2026-06-09
Make reproducer/external_resources pipeline printing work with JSON options

gerion-amd:
Made with help of AI/Cursor.

### DIFF
--- a/mlir/include/mlir/Pass/PassOptions.h
+++ b/mlir/include/mlir/Pass/PassOptions.h
@@ -4,6 +4,9 @@
 // See https://llvm.org/LICENSE.txt for license information.
 // SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
 //
+// Modifications (c) Copyright 2026 Advanced Micro Devices, Inc. or its
+// affiliates
+//
 //===----------------------------------------------------------------------===//
 //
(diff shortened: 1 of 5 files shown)


mlir/include/mlir/Pass/PassOptions.h
```diff
@@ -67,7 +70,15 @@ static void printOptionValue(raw_ostream &os, const std::string &str) {
   const size_t escapeIndex =
       std::min({str.find_first_of('{'), str.find_first_of('\''),
                 str.find_first_of('"')});
-  const bool requiresEscape = spaceIndex < escapeIndex;
+  // A value that starts with a wrapper character ('{', '\'' or '"') is stripped
+  // of one matching outer pair by the option parser (extractArgAndUpdateOptions
+  // in PassRegistry.cpp) when it is read back. We must wrap it in an extra
+  // '{...}' so the value round-trips; otherwise e.g. a JSON value '{"k":1}'
+  // would be parsed back as '"k":1' (the outer braces silently dropped).
+  const bool startsWithWrapper =
```
jorickert, line 78: Please add a test with a string option "{" .

It seems like you may be too aggressive here.  extractArgAndUpdateOptions only strips if there is a surrounding {}, not just at the start

PR #1023, opened 2026-06-10
[AIEX] Properly constrain new dst after folding constants and reject classes with new valid opcode 

khallouh:


niwinanto (dismissed):

llvm/lib/Target/AIE/aie2ps/AIE2PSInstrInfo.cpp
```diff
@@ -1147,8 +1147,10 @@ AIE2PSInstrInfo::getConstantMovOpcode(MachineRegisterInfo &MRI,
     if (regClassMatches(AIE2PS::mMvSclDstRegClass, DstRegClass, Reg))
       return AIE2PS::MOV_scalar_imm11_pseudo;
   }
-  if (ImmSize <= 32)
-    return AIE2PS::MOVXM_lng_cg;
+  if (ImmSize <= 32) {
```
niwinanto: May be from a different angle, this can cause problem at instruction selection. If ISel want to create constant involving pointer, they fail to get the opcode. May be, we should add this restriction to the consumer.

khallouh: You are right, I now constrain in the foldmmediate instead. That works, it folds pointers after constraining to eP_as32Bit and rejects control registers and such since they don't have an intersection with any valid opcode returned by getConstantMovOpcode. I renamed the PR to reflect the new fix.

llvm/test/CodeGen/AIE/aie2ps/constant-folding.mir
```diff
@@ -68,3 +68,36 @@ body: |
...
+
+---
+name: test_imm_constant_fold_large_imm_into_ptr
+legalized: true
+tracksRegLiveness: true
+body: |
+  bb.1.entry:
+    ; CHECK-LABEL: name: test_imm_constant_fold_large_imm_into_ptr
+    ; CHECK: [[MOVXM_lng_cg:%[0-9]+]]:ep_as_32bit = MOVXM_lng_cg 70000
+    ; CHECK-NEXT: $p0 = COPY [[MOVXM_lng_cg]]
```
niwinanto, line 97: Just curious, do you see this copy getting eliminated in later passes?

niwinanto (approved):
Thank you for addressing this @khallouh. LGTM!

llvm/lib/Target/AIE/AIEBaseInstrInfo.cpp
niwinanto, line 1: Nit: you could update the commit name as well as you updated the PR title.

niwinanto (approved):

llvm/test/CodeGen/AIE/aie2ps/constant-folding.mir
khallouh: Yes, it is eliminated by the coalescer

PR #1025, opened 2026-06-11
[AIE] Fix instruction-select crash on unpack of a non-load value

hunhoffe:
Note: I'm just a peano user, not a dev -- agents helped copiously in this PR, so I'd appreciate a thorough review.

### Problem

Feeding the result of `acc.to_vector<int8>()` directly into `aie::unpack` crashes Peano with a SIGSEGV during instruction selection, on both AIE2 and AIE2P.
  
Minimal C++ reproducer (`clang++ --target=aie2p-none-unknown-elf -O2`):

```cpp
  #include <aie_api/aie.hpp>
  aie::vector<int16, 32> bad(aie::vector<int8, 32> a, aie::vector<int8, 64> b) {
    using MMUL = aie::mmul<4, 8, 8, int8, int8>;
    MMUL acc; acc.mac(a, b);
    aie::vector<int8, 32> proj = acc.to_vector<int8>(8);
    return aie::unpack(proj);            // <- proj feeds unpack directly
  } 
```
  
### Backtrace:
* `AIE2PInstructionSelector::getCombinedOpcodeUNPACKLoad`
* `AIE2PInstructionSelector::canCombineUNPACKLoad`
* `AIEBaseInstructionSelector::selectG_AIE_LOAD_UNPACK`
* `llvm::InstructionSelect::selectMachineFunction`

### Root cause

`selectG_AIE_LOAD_UNPACK` tries to fold a load directly into the unpack. It gates the fold on `MemOp.mayLoad()`, then calls `getLoadStoreSize(MemOp)`, which dereferences the instruction's first memory operand:

```
  return (*MI.memoperands_begin())->getSizeInBits().getValue();
```

The producer here is the SRS narrowing intrinsic (the lowering of `to_vector<int8>()`), a `G_INTRINSIC_W_SIDE_EFFECTS` that reports `mayLoad()` but carries no memory operand. `memoperands_begin() == memoperands_end()`, so the dereference reads an invalid pointer → segfault.

### Fix

Bail out of the combine when the candidate has no memory operand, so a non-load producer falls back to normal `VUNPACK` selection. Applied to both the AIE2 and AIE2P selectors (both had the identical latent crash).

### Testing

- New MIR regression tests for both targets (`inst-select-unpack-non-load.mir`); they SIGSEGV before this change and select cleanly after.
- Real load→unpack still folds into `VLDB_UNPACK` (existing `inst-select-*vldb_unpack.mir` tests unchanged).
- Full `llvm/test/CodeGen/AIE/{aie2,aie2p}/GlobalIS*el` suites pass locally (391/391).


llvm/lib/Target/AIE/AIE2InstructionSelector.cpp
```diff
@@ -640,7 +640,11 @@ AIE2InstructionSelector::getCombinedOpcodeUNPACKLoad(
            Intrinsic::aie2_unpack_I16_I8))
     return {};
 
-  if (!MemOp.mayLoad())
+  // The unpack may be fed by an op that reports mayLoad() but is not an actual
+  // memory access (e.g. the SRS narrowing intrinsic that lowers
+  // acc.to_vector<int8>()). Such ops carry no memory operand, so bail out
+  // before getLoadStoreSize() dereferences a non-existent one.
+  if (!MemOp.mayLoad() || MemOp.memoperands_empty())
```
martien-de-jong, line 647: Have you tried AIE2PS?  If it doesn't occur there, it may mean that the memory operand modeling/handling is better there.

hunhoffe: @martien-de-jong Verified — AIE2P crashes identically before this fix. The guard logic in `AIE2PInstructionSelector::getCombinedOpcodeUNPACKLoad` was `if (!MemOp.mayLoad()) return {};` with no `memoperands_empty()` check, so the same  null-deref SIGSEGV occurs. I confirmed this by running the new `inst-select-unpack-non-load.mir` test against an unpatched build (exit 139 / SIGSEGV) and a patched build (exit 0, selects `VUNPACK_mv_unpack_w_unpackSign1`).

hunhoffe:
Closing — this was independently fixed and generalized in 046b11f931ce and 37de47afb103, which moved the fix into the shared AIEBaseInstructionSelector::selectG_AIE_LOAD_UNPACK (guarding on isGenericMemOpcode()) and consolidated AIE2PS to reuse it. That covers AIE2, AIE2P, and AIE2PS, with equivalent regression tests already merged for all three targets.

PR #1026, opened 2026-06-11
[AIE] Legalize G_ROTL/G_ROTR by lowering to shifts

hunhoffe:
### Problem
  
A runtime bit-rotate — the standard `(x << k) | (x >> (32 - k))` idiom used by xoshiro/PCG/SplitMix PRNGs and many hashes — fails to compile for all AIE targets:
```
  fatal error: error in backend: unable to legalize instruction:
  %0:(s32) = G_ROTL %1:, %2:_(s32)
```

Reproducer (`clang++ --target=aie2p-none-unknown-elf -O2`):
```cpp
uint32_t rotl(uint32_t x, int k) { return (x << k) | (x >> (32 - k)); }
```

### Root cause

LLVM's middle-end canonicalizes the shift-or idiom into the generic G_ROTL/G_ROTR opcodes before the AIE backend sees the code. The AIE legalizers defined a .lower() rule for the closely-related {G_FSHL, G_FSHR} funnel shifts but never listed {G_ROTL, G_ROTR}, so a rotate fell through to "unable to legalize." This is a missing table entry, not a missing hardware feature — a rotate is FSHL(x, x, k) and decomposes into the same shifts + or.

### Fix

Add getActionDefinitionsBuilder({G_ROTL, G_ROTR}).lower(); to the AIE2, AIE2P, and AIE2PS legalizers. The generic lowerRotate helper expands the rotate into the shift/and/or sequence AIE already supports: `rotl(x,k) = (x << (k & 31)) | (x >>l ((0-k) & 31))`.

### Testing
- New legalizer test legalize-rotate.mir (G_ROTL + G_ROTR, runs on aie2 / aie2p / aie2ps); it crashes with "unable to legalize" before this change and lowers correctly after.
- Full llvm/test/CodeGen/AIE suite passes (no regressions).
- A runtime-rotate kernel built with this fix runs on an AMD Strix NPU and matches a numpy reference across
  several rotate amounts; the kernel object contains no native rotate instruction (confirming the shift+or lowering).


llvm/lib/Target/AIE/aie2ps/AIE2PSLegalizerInfo.cpp
```diff
@@ -470,6 +470,8 @@ AIE2PSLegalizerInfo::AIE2PSLegalizerInfo(const AIE2PSSubtarget &ST)
 
   getActionDefinitionsBuilder({G_FSHL, G_FSHR}).lower();
 
+  getActionDefinitionsBuilder({G_ROTL, G_ROTR}).lower();
```
martien-de-jong, line 473: I wouldn't mind merging all plain .lower entries in one getActionsDefinitionBuilder.

martien-de-jong (approved):

PR #1029, opened 2026-06-12
[AIEX] Stuckmann.runtimes.parallel

F-Stuckmann:
Speed up runtime building by

- Allowing each Architecture's runtime build to proceed in parallel
- Adding no-vitis-headers in  compiler ID steps

### DIFF
--- a/llvm/cmake/modules/LLVMExternalProjectUtils.cmake
+++ b/llvm/cmake/modules/LLVMExternalProjectUtils.cmake
@@ -1,3 +1,14 @@
+#===----------------------------------------------------------------------===#
+#
+# Part of the LLVM Project, under the Apache License v2.0 with LLVM Exceptions.
+# See https://llvm.org/LICENSE.txt for license information.
+# SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+#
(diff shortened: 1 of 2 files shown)


llvm/cmake/modules/LLVMExternalProjectUtils.cmake
```diff
@@ -397,9 +408,6 @@ function(llvm_ExternalProject_Add name source_dir)
     INSTALL_COMMAND ""
     STEP_TARGETS configure build
     BUILD_ALWAYS 1
-    USES_TERMINAL_CONFIGURE 1
-    USES_TERMINAL_BUILD 1
-    USES_TERMINAL_INSTALL 1
```
martien-de-jong, line 402: Are we sure this doesn't share a build directory across targets? In the grand scheme of things we may also be building the intel codegen for host and emulation flow. This will cause more differenes than switching from one AIE to another.

F-Stuckmann: Each Target themselves creates its own runtimes folders, so there are no overlaps.

martien-de-jong (approved):

PR #1031, opened 2026-06-15
[AIE2PS] Model ePSRFLdF spills with PLFR composite spill pseudos to preserve stack liveness

khallouh:
The previous approach expanding the spill early in RA did not maintain stack liveness information, which led to the first 2 out of 3 stores being marked as stores to a dead slot causing machine verifier errors.
As far as correctness is concerned, we have been safe since the 3 stores are atomic and stack slot coloring runs immediately afterwards, which is the only pass relying on cached stack liveness later.
This approach properly maintains the stack liveness and keeps the machine verifier happy.
This is achieved by Spill/reload of ePSRFLdF via VST_PLFR_SPILL and VLDA_PLFR_SPILL so reload
defines the composite as a whole (FIFO via a VEC1024 scratch), avoiding
sub-register liveness issues from piecewise expansion.
eliminateFrameIndex splits the pseudo into Y and scalar spill pseudos
and expands them.


llvm/lib/Target/AIE/aie2ps/AIE2PSRegisterInfo.cpp
```diff
@@ -327,6 +350,89 @@ bool AIE2PSRegisterInfo::eliminateFrameIndex(MachineBasicBlock::iterator II,
...
+    // Operand layouts:
+    //   VST_PLFR_SPILL  $fifo(use), $avail(use), $ptr(use), <FI>
+    //   VLDA_PLFR_SPILL $fifo(def, VEC1024 scratch), $plfr(full def), <FI>
+    // For the store the FIFO half has already been bounced into the VEC1024
+    // operand. For the reload the scalar halves are sub-registers of the
+    // composite $plfr def, and the FIFO half is loaded into the VEC1024 scratch
+    // and then bounced into $plfr's sub_fifo with a copy.
+    const bool IsStore = (Opc == AIE2PS::VST_PLFR_SPILL);
+    const int FifoSize =
+        static_cast<int>(TRI.getSpillSize(AIE2PS::VEC1024RegClass));
```
niwinanto: Nit: anyway, we are hardcoding `VEC1024RegClass`, why dont we hardcode 128?

llvm/test/CodeGen/AIE/aie2ps/spill/spill-reload-psrf.mir
```diff
@@ -21,15 +21,10 @@ body:             |
...
-    ; CHECK-NEXT: ST_R_SPILL $r24, %stack.0, implicit $sp :: (store (s32) into %stack.0 + 128, align 64)
-    ; CHECK-NEXT: ST_R_SPILL $p0, %stack.0, implicit $sp :: (store (s32) into %stack.0 + 132, basealign 64)
+    ; CHECK-NEXT: VST_PLFR_SPILL killed renamable $y0, $r24, $p0, %stack.0, implicit $sp :: (store (s1088) into %stack.0, align 64)
     ; CHECK-NEXT: PseudoJL 32, csr_aie2ps, implicit-def $lr
-    ; CHECK-NEXT: renamable $y0 = VLDA_Y_SPILL %stack.0, implicit $sp :: (load (s1024) from %stack.0, align 64)
-    ; CHECK-NEXT: dead renamable $lf0 = COPY killed renamable $y0
-    ; CHECK-NEXT: dead renamable $r24 = LDA_R_SPILL %stack.0, implicit $sp :: (load (s32) from %stack.0 + 128, align 64)
-    ; CHECK-NEXT: renamable $p0 = LDA_R_SPILL %stack.0, implicit $sp :: (load (s32) from %stack.0 + 132, basealign 64)
-    ; CHECK-NEXT: PseudoRET implicit $lr, implicit renamable $plfr0
+    ; CHECK-NEXT: dead renamable $y0, renamable $plfr0 = VLDA_PLFR_SPILL %stack.0, implicit $sp :: (load (s1088) from %stack.0, align 64)
```
niwinanto, line 26: Why don't we make the load semantically similar to store, meaning split $y0, $r24, $p0.

niwinanto, line 26: Also, `$y0` is marked as dead.

llvm/test/CodeGen/AIE/aie2ps/spill/spill-reload-psrf-ir.ll
```diff
@@ -0,0 +1,40 @@
...
+; Spill: bounce the FIFO buffer through x, store the scalars, then store both
+; FIFO-buffer halves (the vst land in the call's delay slots).
+; CHECK:         vmov lfl0, x6
+; CHECK:         vmov lfh0, x7
+; CHECK:         st r24, [sp, #-64]; vmov x0, lfl0
+; CHECK:         st p0, [sp, #-60]; vmov x1, lfh0
+; CHECK:         vst x0, [sp, #-192]
+; CHECK:         vst x1, [sp, #-128]
+; Reload: load both FIFO-buffer halves and the scalars back.
+; CHECK:         vlda x0, [sp, #-192]
```
niwinanto: We are not copying back to fifo registers from x register.

llvm/test/CodeGen/AIE/aie2ps/spill/spill-reload-psrf.mir
khallouh: The reason is sub-register liveness. On the store side, the pseudo only uses the sub-registers ($y0, $r24, $p0), so splitting them is fine -- you're just reading from existing live values. But on the load side, the pseudo defines the composite. If you were to split the load into three separate defs of the individual sub-registers ($y0, $r24, $p0), you would then need to reconstruct the composite $plfr from those pieces afterwards. That reconstruction would break the register allocator's understanding of the composite register's liveness (its sub-register liveness).

khallouh: The dead flag is an unavoidable artifact of deferring the scratch's use to eliminateFrameIndex but it's harmless with the current pipeline because none of the intervening passes (BranchFolding, TailDuplication, MachineCopyPropagation, MachineLateInstrsCleanup, AIESpillSlotOptimization) insert new instructions that define arbitrary physical registers based on dead-flag availability.

llvm/test/CodeGen/AIE/aie2ps/spill/spill-reload-psrf-ir.ll
khallouh: That was just missing check lines we do copy back, let me fix it

khallouh: And now you don't see an explicit vmov to fifo because the move is optimized away since we return in vector registers

llvm/lib/Target/AIE/aie2ps/AIE2PSRegisterInfo.cpp
```diff
@@ -327,6 +350,88 @@ bool AIE2PSRegisterInfo::eliminateFrameIndex(MachineBasicBlock::iterator II,
...
+        MIB.addMemOperand(MMO);
+      PendingSubs.push_back({MIB.getInstr(), SubOff, Encodable});
+    };
+
+    const unsigned FifoOpc =
+        IsStore ? AIE2PS::VST_Y_SPILL : AIE2PS::VLDA_Y_SPILL;
+    const unsigned ScalarOpc =
+        IsStore ? AIE2PS::ST_R_SPILL : AIE2PS::LDA_R_SPILL;
+
+    BuildSub(FifoOpc, FifoBounce, /*ByteOff=*/0, FifoSize, /*Encodable=*/true);
```
mludevid: Why do you hardcode that the vector spill is always encodable?

khallouh: I had missed that, fixed!

llvm/lib/Target/AIE/aie2ps/AIE2PSRegisterInfo.cpp
```diff
@@ -332,6 +355,94 @@ bool AIE2PSRegisterInfo::eliminateFrameIndex(MachineBasicBlock::iterator II,
...
+    // Operand layouts:
+    //   VST_PLFR_SPILL  $fifo(use), $avail(use), $ptr(use), <FI>
+    //   VLDA_PLFR_SPILL $fifo(def, VEC1024 scratch), $plfr(full def), <FI>
+    // For the store the FIFO half has already been bounced into the VEC1024
+    // operand. For the reload the scalar halves are sub-registers of the
+    // composite $plfr def, and the FIFO half is loaded into the VEC1024 scratch
+    // and then bounced into $plfr's sub_fifo with a copy.
+    const bool IsStore = (Opc == AIE2PS::VST_PLFR_SPILL);
+    const int FifoSize = 128;
+    const int ScalarSize = 4;
```
mludevid, line 375: Nit: Could you fetch these constants from the `TRI->getSpillSize`?

mludevid (approved):
LGTM. Thanks for addressing my concerns!

PR #1038, opened 2026-06-17
[AIEX] Disable foldImmediate peephole by default; rename opt-in flag

erwei-xilinx:
## Problem

The `foldImmediate` peephole added in e9815c34 collapses `MOVA reg, #imm` / `COPY dst, reg` pairs into a single `MOVA dst, #imm` at the consumer site. This caused a correctness failure on AIE2 (NPU1) hardware, surfaced as a wrong-output failure in the `vector_exp` programming example in mlir-air CI ([run 27560527882](https://github.com/Xilinx/mlir-air/actions/runs/27560527882)).

### Root cause

The peephole removes a pre-RA data edge (the `COPY`) that the post-RA scheduler was using to keep a `VBCST.16` write separated from downstream vector-register readers. Without that edge, the scheduler places the `VBCST.16` (which fills a full 512-bit x-register) immediately adjacent to a `VMOV_mv_w` that reads only the **lower wl sub-register** (`wl2`) of the same x-register:

```
; Failing schedule produced after foldImmediate:
mova r0, #24;  vbcst.16 x2, r1   ← x2 written (cycle 0)
vshuffle x4, x8, x4, r0          ← gap (cycle 1)
vmov wh4, wl2                     ← wl2 read (cycle 2) — RAW hazard
```

The `wl` forwarding bypass encoded in the AIE2 itinerary for `II_VMOV_W_WMH_WML` consumers (shared `MOV_Bypass` with `VBAND`/`VMOV` outputs) **does not extend to `VBCST` producers**: `VBCST` writes both halves through a broadcast path that lacks the wl-specific forwarding, so the 2-cycle output latency is insufficient and the hardware reads a stale value.

### Why the itinerary is hard to fix directly

`VBAND` (II_VBLOG) and `VBCST` both emit `MOV_Bypass` for their x-register output. `VBAND→wl` has a real hardware bypass; `VBCST→wl` does not. LLVM's itinerary system allows only one bypass class per operand, so we cannot distinguish them without a new bypass type (which would require touching many itinerary entries) or increasing `VBCST` output latency (which would break the validated `VBCST→VEXTRACT` bypass). This is left as a follow-up.

## Fix

Disable `foldImmediate` by default and rename the controlling flag from `-aie-disable-fold-imm` (double-negative, `cl::init(false)`) to `-aie-enable-fold-imm` (`cl::init(false)`, opt-in), consistent with other AIE feature flags (`aie-tail-call-optimization`, `aie-subreg-renaming`, etc.).

## Testing

- All 7 affected codegen tests updated via `update_llc_test_checks.py` / `update_mir_test_checks.py`
- `constant-folding.mir` gains a second `RUN` line with `-aie-enable-fold-imm` that preserves the positive test for the folding logic (using `NOFOLD`/`FOLD` FileCheck prefixes)
- Verified on local NPU1 (Phoenix / amd8845hs):
  - `mlir-air check-air-e2e-peano`: identical pass/fail count before and after (19 pre-existing failures, none new)
  - `mlir-air check-programming-examples-peano`: `vector_exp` goes from FAIL → PASS; no other changes
  - LLVM `check-llvm`: 18 failures (all pre-existing; 3 fewer than before this fix)

## Follow-up

Correct the AIE2 itinerary to accurately model the VBCST→wl latency (either a new bypass class separating VBCST output from VBLOG/VMOV output, or per-consumer output latency for VBCST wl readers), then re-enable the optimisation by default.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/llvm/lib/Target/AIE/AIEBaseInstrInfo.cpp
+++ b/llvm/lib/Target/AIE/AIEBaseInstrInfo.cpp
@@ -51,10 +51,24 @@ static cl::opt<bool> AIEFoldImmRequireOneUse(
     cl::desc("Only fold immediate into COPY when the constant has a single "
              "non-debug use (so the def can be DCE'd)."));
 
-static cl::opt<bool> AIEDisableFoldImm(
-    "aie-disable-fold-imm", cl::init(false), cl::Hidden,
-    cl::desc("Completely disable the AIE foldImmediate override (fall back to "
-             "default no-op behaviour)."));
+// Opt-in flag for the foldImmediate peephole (disabled by default).
+//
+// foldImmediate collapses MOVA reg, #imm / COPY dst, reg pairs into a single
+// MOVA dst, #imm at the consumer site.  While functionally correct, it removes
+// a pre-RA scheduling constraint: the original COPY gave the post-RA scheduler
+// a data edge that kept the MOVA far from its downstream consumers.  Without
+// that edge, the post-RA scheduler may place a VBCST.16 write (which fills an
+// entire x-register) immediately before a VMOV that reads only the lower-half
+// wl sub-register.  The wl forwarding bypass modelled in the AIE2 itinerary
+// for VMOV_W_WMH_WML consumers does not apply when the producer is a VBCST,
+// so the 2-cycle output latency is insufficient and the hardware reads a stale
+// value.  The correct fix requires a more precise itinerary (either a new
+// bypass class or increased VBCST output latency for wl readers); this flag
+// provides a safe fallback in the meantime.
+static cl::opt<bool> AIEEnableFoldImm(
+    "aie-enable-fold-imm", cl::init(false), cl::Hidden,
+    cl::desc("Enable the AIE foldImmediate peephole that collapses "
+             "MOVA+COPY pairs into a single MOVA at the consumer site."));
 
 static cl::opt<bool>
     NoCheapInstHoisting("aie-no-cheap-inst-hoising",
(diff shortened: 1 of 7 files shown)


erwei-xilinx:
Closing this in favour of a targeted workaround in mlir-air.

The root cause is that `foldImmediate` removes a pre-RA data edge that the post-RA scheduler was using to keep VBCST writes separated from VMOV wl readers, causing a RAW hazard that the current itinerary does not catch. Flipping the default here would regress the optimisation for all Peano users, not just those hitting this pattern.

Instead, mlir-air now passes `-mllvm -aie-disable-fold-imm` to the Peano clang++ invocation when compiling AIE2 external kernels. This scopes the workaround to mlir-air only and leaves the optimisation enabled for everyone else.

The proper long-term fix in llvm-aie is to correct the AIE2 itinerary so that the wl forwarding bypass is not incorrectly shared between VBLOG/VMOV outputs (which have the hardware bypass) and VBCST outputs (which do not).

khallouh:
Thanks @erwei-xilinx. This seems like a bug in the scheduling model rather than a problem with the constant folding optimization. In fact, this optimization yields some good improvements and we plan to extend it in the future to cover more registers classes and more instructions consuming immediates. 

Now, I'm not sure I exactly understand what's wrong with the schedule snippet you posted. As far as we can see, from the architecture description, VBCST instruction uses the bypass for its X register outputs, that is for both WL and WH parts, and that is also what we model in AIE2Schedule.td. 
```
InstrItinData<II_VBCST, [EmptyCycles<1>, InstrStage<1,  [W_WM_PORT]>],
              [2,1], [MOV_Bypass,NoBypass]>,
```
That being said, the failing schedule which you posted should be fine whether there is a bypass or not since the wl2 register is not being overwritten by any other instruction in between the broadcast and the vmov. The value is committed to the register in cycle 2 whether it is bypassed or not.

Also, you mention VBAND but I don't see this instruction in your snippet (reposted below).
```
; Failing schedule produced after foldImmediate:
mova r0, #24;  vbcst.16 x2, r1   ← x2 written (cycle 0)
vshuffle x4, x8, x4, r0          ← gap (cycle 1)
vmov wh4, wl2                     ← wl2 read (cycle 2) — RAW hazard
```
It would be great if you could share an LLVM IR reproducer of the bug and the assembly with and without `foldImmediate` optimization. Maybe this can be an easy fix and you can reenable the constant folding for your flow since this seems unrelated.

PR #1044, opened 2026-06-22
feat: Add canonicalization for Slice(Pad(X)) in TOSA

ehsan-toosi:
Fold a tosa.pad consumed by a tosa.slice so the slice reads directly from the pad's input, re-padding only the region the slice still covers. When the slice window lies entirely within the unpadded input, the pad is dropped.

PR #1046, opened 2026-06-22
feat: Add canonicalization for Pad(Slice(X)) only with ub.poison value in TOSA

ehsan-toosi:


mlir/lib/Dialect/Tosa/IR/TosaCanonicalizations.cpp
```diff
@@ -744,9 +745,108 @@ struct PadPadOptimization : public OpRewritePattern<tosa::PadOp> {
...
+                                PatternRewriter &rewriter) const override {
+    if (!padOp.getPadConst().getDefiningOp<ub::PoisonOp>())
+      return rewriter.notifyMatchFailure(padOp, "pad does not fill poison");
+
+    auto sliceOp = padOp.getInput1().getDefiningOp<tosa::SliceOp>();
+    if (!sliceOp)
+      return rewriter.notifyMatchFailure(padOp, "pad input must be a slice");
+    if (!sliceOp->hasOneUse())
+      return rewriter.notifyMatchFailure(
+          padOp, "preceding slice must have a single use");
```
mathmer-amd, line 766: NIT: Same as for the `Slice(Pad(X))`, I think you could extend the pattern to support the case where it has more than one use

mathmer-amd (dismissed):

mlir/lib/Dialect/Tosa/IR/TosaCanonicalizations.cpp
mathmer-amd: discussed offline, this could result in have 3 tosa ops, which may not always be beneficial

mlir/lib/Dialect/Tosa/IR/TosaCanonicalizations.cpp
```diff
@@ -744,9 +745,108 @@ struct PadPadOptimization : public OpRewritePattern<tosa::PadOp> {
   }
 };
 
+/// Fold a tosa.pad of a tosa.slice that fills with poison. Poison is
+/// don't-care, so it is backfilled from the data the slice dropped: the slice
```
gerion-amd, line 749: We discussed already offline that there are multiple meanings for poison, so either:
- undefined behavior that the compiler can handle as needed.
- undefined behavior that needs propagation (so using these values will also result in UB).

We go for the first approach here.

mlir/lib/Dialect/Tosa/IR/TosaCanonicalizations.cpp
```diff
@@ -744,9 +745,108 @@ struct PadPadOptimization : public OpRewritePattern<tosa::PadOp> {
...
+    // overlap is read by a widened slice and the remainder stays as poison pad.
+    ArrayRef<int64_t> sliceStart = sliceOp.getStart();
+    ArrayRef<int64_t> sliceSize = sliceOp.getSize();
+    SmallVector<int64_t> newStart(rank);
+    SmallVector<int64_t> newSize(rank);
+    SmallVector<int64_t> newPadding(2 * rank, 0);
+    bool needsSlice = false;
+    bool changed = false;
+    for (int64_t axis = 0; axis < rank; ++axis) {
+      const int64_t inDim = sliceInputType.getDimSize(axis);
```
gerion-amd, line 808: Nit: For code understanding, maybe you can do a similar picture here than in #1044 or at least link to it.

gerion-amd (dismissed):
Thanks!

mathmer-amd (approved):

gerion-amd (approved):

PR #1049, opened 2026-06-24
[NFC][AIE][INTERBLOCK] Don't overload BS.Fixpoint.II as pipelining indicator

martien-de-jong:
These are some final relics from an earlier series

Additionally, we don't try pipelining if we exceed the maximum II upfront. This save a bit of time and a lot of logging if you are debugging a specific pipeliner mode.

The maximum absolute II was set comfortably high, since time control is better done with the number of II tries.


F-Stuckmann (approved):
Let's hope maxii is not too small

PR #1050, opened 2026-06-24
[AIE] Tighten top-fixed -> ExitSU latency with cross-BB MaxLatencyFinder

mludevid:


llvm/lib/Target/AIE/AIEInterBlockScheduling.cpp
```diff
@@ -488,7 +494,28 @@ class PipelineExtractor : public PipelineScheduleVisitor {
...
+    auto &TopOrder = Epilogue->TopInsertSemanticOrder;
+    assert(TopOrder.empty() && "Epilogue already has a top semantic order.");
+    for (MachineInstr *OrigMI : Loop.getTop().getFreeInstructions()) {
+      const auto It = EpilogueLastIterClones.find(OrigMI);
+      if (It != EpilogueLastIterClones.end())
+        TopOrder.push_back(It->second);
+    }
+
     TimedRegion.clear();
+    InEpilogue = false;
```
martien-de-jong, line 543: I know this is my doing, but can I ask you to move the implementation of finish () down to after the earlier actions?

llvm/lib/Target/AIE/AIEMaxLatencyFinder.cpp
```diff
@@ -110,40 +112,55 @@ int MaxLatencyFinder::computeEffectiveLatency(MachineInstr &MI) {
                << (SE.getSucc() ? SE.getSucc()->getNumber() : -1) << "\n");
     LLVM_DEBUG(dbgs() << format("Successor %d PostRegionMaxDepth=%d\n", SuccNo,
                                 SE.getPostRegionMaxDepth()));
-    const SUnit *Pred = SE.getPreBoundaryNode(&MI);
-    if (!Pred) {
+    // A top-fixed instruction is represented as a BUNDLE root in the
+    // scheduling DAG, but the inter-block DDG holds the individual bundled
+    // instructions. Collect the pre-boundary nodes for every member so the
+    // bundle's effective latency is the max over its instructions.
+    SmallVector<const SUnit *, 4> Preds;
```
martien-de-jong: We have an accurate upper bound for the size of this vector. We can prevent any heap allocation.

mludevid: Done. I've used the same upper bound as we use in the hazard recognizer, 16 slots.

llvm/lib/Target/AIE/AIEInterBlockScheduling.cpp
mludevid: Done.

llvm/lib/Target/AIE/AIEMaxLatencyFinder.cpp
```diff
@@ -116,15 +120,19 @@ int MaxLatencyFinder::computeEffectiveLatency(MachineInstr &MI) {
     // scheduling DAG, but the inter-block DDG holds the individual bundled
     // instructions. Collect the pre-boundary nodes for every member so the
     // bundle's effective latency is the max over its instructions.
-    SmallVector<const SUnit *, 4> Preds;
+    std::array<const SUnit *, MaxBundleSlots> PredStorage;
+    unsigned NumPreds = 0;
```
martien-de-jong: Why not just use SmallVector<const SUnit *, MaxBundleSlots>, which would  care of the NumPreds admin. It also doesn't impose a true static limit, so 7 would be suitable.

mludevid: Done.

martien-de-jong (approved):

Issue #1053, opened 2026-06-25
[AIE2] -O2 schedules an argument-defining load into a call's delay slots (int->float->int miscompile)


atassis:
## Summary

On AIE2, at `-O2`, the post-RA scheduler sinks a scalar load that defines a soft-float helper call's argument register into that call's delay slots. AIE2 is an in-order exposed-pipeline VLIW with no hardware stall logic, a scalar load has a ~7-cycle def latency, and `jl` has only 5 delay slots, so the callee reads a stale argument register. The result is a silent wrong result, not a crash.

This regressed between the `llvm-aie` nightlies of 2026-06-23 (good, `cb664e8c`) and 2026-06-24 (bad, `fdf257de`). It is the root cause of the mlir-aie CI breakage tracked in Xilinx/mlir-aie#3219.

## Minimal reproducer (no hardware, no MLIR, just `llc` on tiny IR)

Kernel (exactly what the failing mlir-aie tests compile for the int case; `+ 1.0f` forces int->float->int even though the arrays are `int`):

```c
extern "C" void add_one(int* input, int* output, int tile_size) {
    for (int i = 0; i < tile_size; i++)
        output[i] = input[i] + 1.0f;
}
```

Optimized IR (frontend output is byte-identical between the good and bad compiler except the version string, so this is purely a backend regression):

```llvm
%1     = load i32, ptr %arrayidx
%conv  = sitofp i32 %1 to float
%add   = fadd float %conv, 1.0
%conv1 = fptosi float %add to i32
store i32 %conv1, ptr %arrayidx2
```

Repro command (same `-O2` the mlir-aie Peano path uses; target triple `aie2-none-unknown-elf`):

```
llc add_one.ll -O2 -o out.s
```

## The bug, concretely

The soft-float helper `__floatsisf` takes its integer argument in `r1` (verified with `float g(int* p){ return (float)(*p); }` -> `lda r1,[p0,#0]` then `jl #__floatsisf`, no intervening move).

GOOD (`cb664e8c`) -- the argument load is before the call, so `r1` is ready:

```
lshl r0, r17, r19          ; byte offset
mov  dj0, r0
lda  r1, [p6, dj0]         ; r1 = input[i]  (the call argument)
jl   #__floatsisf          ; reads r1 -- ready
nop x4 ; mov r22, r0       ; delay slots (no def of r1)
```

BAD (`fdf257de`) -- the same load is inside the call's delay slots:

```
[...]

Issue #1051, opened 2026-06-24
[AIE2/AIE2P] -O2 schedules a delayed-call argument too late after scalar lowering


thomthehound:
The AIE2/AIE2P optimized code path (`-O2`, versus `-O0`) can generate wrong code when a latency-bearing producer supplies a delayed-call argument.

A conversion-only kernel,

```cpp
output[i] = static_cast<int>(static_cast<float>(input[i]));
```

passes at `-O0` but fails all 1024 output elements at `-O2` on hardware, on both AIE2 and AIE2P.

In the `O2` object, the `r1` load supplying `__floatsisf` is scheduled at the tail of that call's delay region, while `__floatsisf` consumes `r1` immediately on entry. The first stale result is `0x70400`, which is the output-buffer address.

This appears to be a scheduler dependency gap for dynamically appended implicit call arguments.

This causes downstream CI breakage (see [MLIR-AIE #3219](https://github.com/Xilinx/mlir-aie/issues/3219)).

thomthehound:
Addressed by #1069
Thank you, @martien-de-jong !

PR #1054, opened 2026-06-25
[AIE][InterBlock] Keep a call's argument load out of its delay slots

atassis:
## Summary

On AIE2 at `-O2`, the post-RA loop-aware scheduler can place the load that defines a soft-float call's argument register into that call's delay slots. AIE2 is in-order with no interlocks; a scalar load has a multi-cycle def latency while `jl` has only 5 delay slots, so the callee reads a stale argument register, giving a silent wrong result. This is the root cause of #1053 (and the mlir-aie CI breakage in Xilinx/mlir-aie#3221 / Xilinx/mlir-aie#3219).

## Root cause

Call argument registers are moved onto a `DelayedSchedBarrier` after the `jl` (in `AIEPseudoBranchExpansion`) so that dependencies form with the barrier. The barrier is the region end (`ExitSU`). `RegionEndEdges::apply` removes all real `ExitSU` predecessors and re-adds artificial edges whose latency is `MaxLatency(MI)`. In loop-aware scheduling `MaxLatency` is capped (`getLatencyCap`, which can be ~0 during fixpoint convergence), so the edge from a producer of a barrier-used register (the call argument) loses its real latency. The producer is then free to sit in the call's delay-slot shadow.

`git bisect` points to d4581aae13 ("[AIE][InterBlock] Less conservative interblock scoreboard"): that change reduced the bottom-scoreboard conservatism that had been incidentally providing the spacing, which exposed this latent gap. The immediately preceding inter-block commit still schedules this case correctly.

## Fix

In `RegionEndEdges::apply`, floor the `ExitSU` edge latency by the real, uncapped producer latency for any register the region-end barrier actually uses. This keeps a value the call consumes at least its real latency before the call, i.e. outside the delay slots, independent of the loop-aware latency cap. Non-call regions are unaffected (their barrier has no register uses).

## Validation

- New regression test `call-arg-not-in-delay-slot.ll`: the `r1`-defining load is emitted before `jl #__floatsisf`.
- AIE scheduling lit suite (aie2 + aie2p + aie2ps) passes, and the fix leaves existing golden output unchanged.
- Minimal `int->float->int` reproducer is fixed at `-O2`. On device (NPU2): correct before the bad nightly, wrong after, correct at `-O0`, and correct with this fix.

Reproducer, bisect, ISA details, and on-device confirmation are in #1053.

Once this lands and a new nightly is published, the temporary pin in Xilinx/mlir-aie#3221 can be reverted.


llvm/test/CodeGen/AIE/aie2/schedule/call-arg-not-in-delay-slot.ll
```diff
@@ -0,0 +1,41 @@
...
+
+; Regression test: a load that defines a call's argument register must not be
+; scheduled into that call's delay slots. AIE2 is an in-order exposed-pipeline
+; machine with no interlocks; a scalar load has a multi-cycle def latency while
+; a jl has only 5 delay slots, so a load in the call's delay shadow leaves the
+; argument register stale when the callee reads it (a silent miscompile).
+;
+; The int->float->int kernel below calls the soft-float helper __floatsisf,
+; whose integer argument is in r1. Verify the r1-defining load is emitted before
+; the call, not in its delay slots.
```
martien-de-jong: Could you do me a favour, and directly supply the input to the postscheduler. which you can obtain by passing --stop-before=postmisched. Take the output file, give it a .mir extension and change the RUN line to include --start-before=postmisched.

Then make it auto-updateable by running utils/update_llc_test_checks.py on the file.

llvm/lib/Target/AIE/AIEBaseSubtarget.cpp
```diff
@@ -388,6 +388,30 @@ class RegionEndEdges : public ScheduleDAGMutation {
...
+            if (MO.isReg() && MO.isDef() && MO.getReg() &&
+                EndMI->readsRegister(MO.getReg(), TRI))
+              return true;
+          return false;
+        };
+        if (Itins && FeedsBarrierReg())
+          EdgeLatency = std::max<unsigned>(
+              EdgeLatency,
+              AIE::maxLatency(&MI, *TII, *Itins, /*IncludeStages=*/false));
+      }
```
martien-de-jong, line 414: Looks good. Some questions:
1- I think this only goes wrong in a Zero-Overhead Loop, since regular control flow would constitute another Bottom Region.
2- Is this triggered by the recent MaxLatencyFinder PR?

llvm/lib/Target/AIE/AIEBaseSubtarget.cpp
```diff
@@ -388,6 +388,30 @@ class RegionEndEdges : public ScheduleDAGMutation {
...
+          for (const MachineOperand &MO : MI.operands())
+            if (MO.isReg() && MO.isDef() && MO.getReg() &&
+                EndMI->readsRegister(MO.getReg(), TRI))
+              return true;
+          return false;
+        };
+        if (Itins && FeedsBarrierReg())
+          EdgeLatency = std::max<unsigned>(
+              EdgeLatency,
+              AIE::maxLatency(&MI, *TII, *Itins, /*IncludeStages=*/false));
```
martien-de-jong, line 413: nit: we can just use maxLatency, it will never be less than EdgeLatency.

llvm/test/CodeGen/AIE/aie2/schedule/call-arg-not-in-delay-slot.ll
martien-de-jong: If you don't feel comfortable with that, no problem.

thomthehound:
This addresses the same issue I reported in #1051. I will close that once this is ready.

I do have one small question about this fix: `FeedsBarrierReg()` appears to match any instruction in the region that defines a register read by `DelayedSchedBarrier`, rather than specifically the reaching definition of that barrier use.

Could an earlier definition of, say, `$r1` that is overwritten before the call therefore receive the uncapped latency floor unnecessarily? The semantic requirement seems to be the latency of the definition that actually supplies the call argument.

I had not yet finished my own attempt at a patch, much less validated one, but I am wondering whether we should add a postmisched MIR guard with an earlier long-latency `$r1` definition overwritten by a later short-latency one. The later definition should remain eligible for the final delay slot, and the earlier dead definition should not be constrained merely because the barrier also reads `$r1`. But, again, I am not certain how much of a concern this really is.

atassis:
@thomthehound hi there. I have missed your issues, that's why I did mine, sorry if I have disturbed you in any way. 
On the fix itself - you are right, the check fires on any instruction that writes a register the call reads, not just the one that really supplies the argument. But I don't think it can miscompile, because the floor only ever pushes a producer further from the call, never closer. So in your example the real (later, cheap) write to `r1` still floors to its own tiny latency and can stay in the last delay slot. It's only a dead earlier write to `r1` that gets shoved out for no reason. Worst case we waste a delay slot, we don't generate wrong code.
The exact fix would be to look at what genuinely feeds the call (ExitSU's real dependency edges) instead of "anything that writes `r1`". The thing is, the code just above (`removeExitSUPreds`) throws those edges away and rebuilds them, so I'd have to grab that info before it's wiped. I'll add a postmisched MIR test in the shape you described - dead write, live write, then the call - so the behavior is pinned. And, if you and @martien-de-jong think the extra precision is worth it I'm glad to switch the match over to the reaching def.

llvm/lib/Target/AIE/AIEBaseSubtarget.cpp
atassis: I went back and forth on this, but I think the max is load-bearing. EdgeLatency can already be higher than the uncapped maxLatency in a few places - the `-aie-latency-margin` path (a load has no delay slots, so it takes the UserLatencyMargin branch, and a few tests set that margin) and
the IncludeStages term in non-loop regions - so a plain assign would lower it there and could change existing schedules. The intent is a one-way floor: raise the edge when the loop cap shrank it, never lower it. Happy to drop the max if you see why those paths can't line up with
a barrier-feeding def.

llvm/lib/Target/AIE/AIEBaseSubtarget.cpp
atassis: 1. Yep, this matches what we see. The latency cap only applies in loop regions (it's gated on BlockType::Loop), so outside a loop the exit edge keeps the real latency and the load stays put. Our reproducer is a loop and we couldn't get a straight-line case to misbehave, consistent with
your read.
2. The trigger is actually d4581aae13, not the MaxLatencyFinder change. git bisect lands on d4581aae13 as first-bad, and the commit right before it (f8d932c9ab) still schedules this correctly on the same IR. That said, f8d932c9ab did relax the ExitSU assert from `< DelaySlots` to `<= DelaySlots + 1`, so it set up part of the gap; d4581aae13 is what exposed it by removing the conservative scoreboard cushion that had been keeping the load out of the shadow.

llvm/test/CodeGen/AIE/aie2/schedule/call-arg-not-in-delay-slot.ll
atassis: That's fine, went ahead and did it. Converted the test to a postmisched MIR test in 92cde73: dumped with --stop-before=postmisched, gave it the .mir extension, RUN line restarts at postmisched into FileCheck, and regenerated the checks with update_llc_test_checks.py - re-running it is idempotent, so it stays auto-updateable. The 92cde73 SHA may shift, though, if the branch gets rebased.

thomthehound:
> @thomthehound hi there. I have missed your issues, that's why I did mine, sorry if I have disturbed you in any way. On the fix itself - you are right, the check fires on any instruction that writes a register the call reads, not just the one that really supplies the argument. But I don't think it can miscompile, because the floor only ever pushes a producer further from the call, never closer. So in your example the real (later, cheap) write to `r1` still floors to its own tiny latency and can stay in the last delay slot. It's only a dead earlier write to `r1` that gets shoved out for no reason. Worst case we waste a delay slot, we don't generate wrong code. The exact fix would be to look at what genuinely feeds the call (ExitSU's real dependency edges) instead of "anything that writes `r1`". The thing is, the code just above (`removeExitSUPreds`) throws those edges away and rebuilds them, so I'd have to grab that info before it's wiped. I'll add a postmisched MIR test in the shape you described - dead write, live write, then the call - so the behavior is pinned. And, if you and @martien-de-jong think the extra precision is worth it I'm glad to switch the match over to the reaching def.

Thank you for looking into this. I perhaps find myself overly cautious given how unprotected this area seems to be by current regression tests.

atassis:
@martien-de-jong hi! Friendly reminder if this can be merged.

llvm/lib/Target/AIE/AIEBaseSubtarget.cpp
martien-de-jong: Nothing wrong with the fix, but finding a call in a ZOL body sounds wrong. We have a (fragile) mechanism that is rejecting ZOL for operations that are predicted to end up in runtime support calls, which is second-guessing legalization. I kind of promised to replace that mechanism with the reverse, which is lowering a ZOL loop to normal control flow when we find a call after code selection. I guess I need to start implementing that....

llvm/lib/Target/AIE/AIEBaseSubtarget.cpp
martien-de-jong: Ah, UserLatencyMargin. That's a testing option to be used with care. It breaks the consistency on the scheduler model. But ok. The max is obviously correct.

martien-de-jong:
About the earlier write to r1, that's connected to the second one with a WAW edge, and will not connect to ExitSU. The dag mutator is only modifying latencies that go over the edge of the region.

llvm/test/CodeGen/AIE/aie2/schedule/call-arg-not-in-delay-slot.mir
```diff
@@ -0,0 +1,299 @@
...
+    renamable $bmh1 = VADD_F killed renamable $bmh0, killed renamable $bmh1, renamable $r20, implicit-def dead $srfpflags, implicit $crfpmask
+    $x2 = VMOV_mv_x killed $bmh1
+    renamable $r1 = VEXTRACT_S32 killed renamable $x2, renamable $r18
+    JL &__fixsfsi, implicit-def $lr
+    DelayedSchedBarrier csr_aie2, implicit $r1, implicit-def $r0
+    renamable $r17 = nuw nsw ADD_add_r_ri killed renamable $r17, 1, implicit-def dead $srcarry
+    renamable $r1 = EQ renamable $r16, renamable $r17
+    $dj0 = MOV_mv_scl killed $r22
+    ST_dms_sts_idx killed renamable $r0, renamable $p7, killed renamable $dj0 :: (store (s32) into %ir.arrayidx2)
+    JZ killed renamable $r1, %bb.2
```
martien-de-jong, line 281: But this is not a zero-overhead loop. The call is now not in the last region of the block, and should not partake in latency capping

martien-de-jong: I'm getting convinced that this is a symptom of another bug. We should really not be capping the latency on this load, since it is not the bottom region. Without the fix, the load sits in the last delay slot, which can only be the case if it gets its latency from capping. The fix is undoing the capping by recomputing its latency locally. The DelaySlotBarrier is part of the next region, hence Maxlatency should return maxLatency, 
The Cap is norrmally safe since we check converge of LoopAware in 'latencyConverged' which is explicitly checking the latency between Top and Bottom region of the loop block. It is not checking the latency between any other region with its ExitSU. I'll be working on a proper fix.

martien-de-jong:
@atassis , I am assuming recent PRs merged to aie-public fix the root cause of this issue. Can you confirm and close the PR or report back your findings?

atassis:
@martien-de-jong hi there. I am sick at the moment, not able to properly review the current state of the work. Can I take a bit more time and answer you back in several days? I'll look through all commited work I have done against xilinx/AMD repos and make a proper action in each when I get better, if it sounds good for you

atassis:
@martien-de-jong you are right - I rebuilt current aie-public (3a4f47b4f, without this PR) and the int->float->int reproducer no longer miscompiles: the r1-defining lda is scheduled before jl #__floatsisf, delay slots are clean, and the regression FileCheck passes. Looks like b892afe21 ("Fix unscheduled Depth computation") covers it. So the fix here is redundant - closing this PR.

One thing: this scheduling case still has no regression test upstream (the concern @thomthehound raised). If it's useful I can send just the call-arg-not-in-delay-slot.ll / postmisched MIR guard as a tiny standalone PR so a future scheduler change can't silently put a call-argument load back in the delay slots.

Happy either way - thanks for the nudge to recheck.

PR #1057, opened 2026-06-25
[NFC][AIEX] Add IR verifier RUN line to outer-loop-pipelining tests

khallouh:
The outer-loop-pipelining unit tests run the pass via `llc -stop-after=aie-outer-loop-pipeliner`, which halts the codegen pipeline immediately after the pass and prints the IR. The codegen IR verifier only runs on the pass *input*, so malformed output (e.g. a PHI whose incoming block is not a predecessor) slips through as a silent FileCheck mismatch rather than an error.

Add a second RUN line that feeds the emitted MIR back through `llc -x mir -run-pass=none`, so the MIR parser re-runs the IR verifier on the pass output and the tests fail loudly on broken IR. Each verifier line mirrors its test's own pipeliner flags (the metadata/nested tests opt in via loop metadata rather than -aie-enable-outer-loop-pipelining).

The end-to-end tests under aie2ps/end-to-end already exercise the full pipeline (and thus the module verifier), so they are left unchanged.

### DIFF
--- a/llvm/test/CodeGen/AIE/aie2p/outer-loop-pipelining-gep.ll
+++ b/llvm/test/CodeGen/AIE/aie2p/outer-loop-pipelining-gep.ll
@@ -8,6 +8,11 @@
 ; RUN:     -stop-after=aie-outer-loop-pipeliner \
 ; RUN:     -o - %s 2>&1 | FileCheck %s
 
+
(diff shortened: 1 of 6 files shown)


F-Stuckmann (approved):
Thank you

Issue #995, opened 2026-05-09
ICE in mm_bfp.cc: unable to legalize instruction: G_BUILD_VECTOR <8 x s8>


AngryLoki:
clang version 21.0.0 (latest build adc9df1ae1b148af3ff6a7a615145deba6000ac7) fails to compile https://github.com/Xilinx/mlir-aie/blob/main/aie_kernels/aie2p/mm_bfp.cc:

Looks like it could be (at least) 2 separate issues:
```
$ .venv/lib/python3.13/site-packages/llvm-aie/bin/clang++ -O2 -std=c++20 \
  --target=aie2p-none-unknown-elf -DDIM_M=64 -DDIM_K=64 -DDIM_N=64 \
  -I .venv/lib/python3.13/site-packages/mlir_aie/include \
  -I /src/mlir-aie/aie_kernels \
  -c /src/mlir-aie/aie_kernels/aie2p/mm_bfp.cc -o /dev/null

fatal error: error in backend: unable to legalize instruction:
  %55:_(<8 x s8>) = G_BUILD_VECTOR %56:_(s8), %56:_(s8), ...
  (in function: matmul_vectorized_bfp16)
Pass: 'Legalizer'
```
<details><summary>Full log</summary>
<p>

```
In file included from /src/mlir-aie/aie_kernels/aie2p/mm_bfp.cc:12:
In file included from .venv/lib/python3.13/site-packages/mlir_aie/include/aie_api/aie.hpp:41:
In file included from .venv/lib/python3.13/site-packages/mlir_aie/include/aie_api/detail/abs.hpp:13:
In file included from .venv/lib/python3.13/site-packages/mlir_aie/include/aie_api/detail/../vector.hpp:12:
In file included from .venv/lib/python3.13/site-packages/mlir_aie/include/aie_api/detail/../concepts.hpp:17:
.venv/lib/python3.13/site-packages/mlir_aie/include/aie_api/detail/../types.hpp:30:9: warning: '__aie_register' macro redefined [-Wmacro-redefined]
   30 | #define __aie_register(...) chess_storage(__VA_ARGS__)
      |         ^
/src/ArtLite/.venv/lib/python3.13/site-packages/llvm-aie/lib/clang/21/include/aiebase_chess.h:15:9: note: previous definition is here
   15 | #define __aie_register(x)
      |         ^
In file included from /src/mlir-aie/aie_kernels/aie2p/mm_bfp.cc:12:
[...]

Issue #847, opened 2026-03-13
ICE at -O0 / wrong code at -O1,-O2 with BFP16-emulated MMUL<8,8,8> and shuffle_up_fill


RobinMelhuish:
# ICE at -O0 / wrong code at -O1,-O2 with BFP16-emulated MMUL<8,8,8> and shuffle_up_fill

## Summary

When compiling code that uses `aie::mmul<8,8,8>` with `-DAIE_API_EMULATE_BFLOAT16_MMUL_WITH_BFP16`, the compiler:
- **Crashes at `-O0`** with assertion failure in `AIE2PInstructionSelector::selectG_AIE_LOAD_STORE`
- **Generates wrong code at `-O1` and `-O2`** — element 0 of each 8-element group in the output is systematically corrupted

The bug is triggered when 9 BFP16-emulated MAC calls use `shuffle_up_fill`/`shuffle_down_fill` on activation vectors loaded from 3 different source pointers, with all 9 weight vectors pre-loaded.

## Compiler version

```
clang version 20.0.0 (https://github.com/Xilinx/llvm-aie b169186bd072791e958d65deb239f2ac384c30c2)
Target: aie2p-none-unknown-elf
```

## Minimal reproducer

```cpp
#define NOCPP
#include <stdint.h>
#include <aie_api/aie.hpp>

using MMUL = aie::mmul<8, 8, 8, bfloat16, bfloat16>;

extern "C" {
void repro(
    bfloat16 *src0, bfloat16 *src1, bfloat16 *src2,
    bfloat16 *wts, bfloat16 *dst,
    int32_t n
) {
    auto zvec = aie::zeros<bfloat16, 64>();

    auto w0 = aie::load_v<64>(wts);
    auto w1 = aie::load_v<64>(wts + 64);
    auto w2 = aie::load_v<64>(wts + 128);
    auto w3 = aie::load_v<64>(wts + 192);
    auto w4 = aie::load_v<64>(wts + 256);
    auto w5 = aie::load_v<64>(wts + 320);
    auto w6 = aie::load_v<64>(wts + 384);
    auto w7 = aie::load_v<64>(wts + 448);
    auto w8 = aie::load_v<64>(wts + 512);

    for (int i = 1; i < n - 1; i++) {
        int off = i * 64;
        MMUL acc;

        auto a = aie::load_v<64>(src0 + off - 64);
        auto b = aie::load_v<64>(src0 + off);
[...]

PR #1058, opened 2026-06-28
[AIE2P] Don't fold bfp16 conversions into the plain store-convert combine

atassis:
Addresses the **-O0 ICE** reported in #847. (Not a full fix for the issue --
see "Scope" below.)

## Summary

`aie::mmul<8, 8, 8, bfloat16>` compiled with
`-DAIE_API_EMULATE_BFLOAT16_MMUL_WITH_BFP16` crashes the AIE2P GlobalISel
instruction selector at `-O0`:

```
MachineOperand::getReg(): Assertion `isReg() && "This is not a register operand!"' failed.
  (anonymous namespace)::AIE2PInstructionSelector::selectG_AIE_LOAD_STORE
  llvm::InstructionSelect::selectMachineFunction
  Running pass 'InstructionSelect' on function '@...'
```

## Root cause

The bfp16 conversion intrinsics

- `llvm.aie2p.v64accfloat.to.v64bfp16ebs8`
- `llvm.aie2p.v64accfloat.to.v64bfp16ebs16`
- `llvm.aie2p.v64bfp16ebs8.to.v64bfp16ebs16`

produce an aggregate `{ <64 x i8> mantissa, <8 x i8> exponent }`, i.e. they
have **two** defs, so in MIR the source operand is at index 3:

```
%mant:vregbank(<64 x s8>), %exp:gprregbank(<8 x s8>) =
    G_INTRINSIC_W_SIDE_EFFECTS intrinsic(@llvm.aie2p.v64accfloat.to.v64bfp16ebs8), %acc(<64 x s32>)
   ; op0 = mantissa def, op1 = exponent def, op2 = intrinsic ID, op3 = source
```

`getCombinedOpcodeCONVStore` accepted these intrinsics for **every** store
kind. A fused store-convert opcode only exists for the FIFO push path
(`VST_PUSH_*_CONV_bfp16*`, selected by `selectVST_FIFO_CONV`, which correctly
reads the source from operand 3). On the plain vector-store path there is no
`VST.CONV.bfp16` opcode, so the combine returned a **bf16** store-convert
(`VST_CONV_bf16_fp32_*`), and `selectG_AIE_STORE_CONV` read the conversion
source from operand 2 -- correct only for the single-result
`accfloat -> bf16` conversions. For the two-def bfp16 conversions operand 2 is
the intrinsic-ID operand, so `getReg()` tripped the `isReg()` assertion.

The mantissa of a bfp16 value is just a vector held in registers; storing it
is an ordinary vector store, and the conversion itself is selected as a
register `VCONV` by `selectVCONVbfp16` (these intrinsics are already routed
there in `select()`).

## Fix

Reject the bfp16 conversion intrinsics in `getCombinedOpcodeCONVStore` when the
memory op is not the FIFO push intrinsic. This makes `canCombineCONVStore`
return false on the plain-store path, so `selectG_AIE_STORE_CONV` bails before
the bad operand access and the mantissa is selected as a plain vector store;
the FIFO store-convert path is unchanged.

## Test

`llvm/test/CodeGen/AIE/aie2p/GlobalIsel/inst-select-store-bfp16-no-conv-combine.mir`
selects a store of the `ebs8` and `ebs16` mantissa results and checks for the
register `VCONV` + plain `VST` (instead of crashing). It asserts on the
unpatched compiler and passes with the fix. The existing
`inst-select-fifo-store-conv.mir`, `inst-select-vconv-bfp16.mir`, and the
`gemm-bfp16` / `conv2d_bfp16_convert` end-to-end tests continue to pass
(full `aie2p/GlobalIsel` suite green).

## Validation

- The exact reproducer from #847 ICEs at `-O0` without the patch and compiles
  cleanly at `-O0`/`-O1`/`-O2` with it.
- The minimal `aie::mmul<8,8,8,bfloat16>` repro and the IRON `aie2p/mm.cc`
  bfp16 kernel also compile cleanly at all opt levels.
- On hardware: the mlir-air `bf16_x_bfp16` matmul, compiled with the patched
  back end, matches the CPU reference (output correlation 0.999975, threshold
  0.999) on an AIE2P (XDNA2) part.

## Scope (please read before treating this as a full fix for #847)

#847 reports two failures: the `-O0` ICE (above) **and** a separate `-O1`/`-O2`
"element-0 of each 8-group has a systematic +offset" miscompile that the issue
isolates to high register pressure + shuffles + 9 bfp16 MACs.

This change fixes **only the ICE**. I verified the miscompile is independent:
the `-O2` assembly of the issue's exact reproducer is **byte-identical with and
without this patch** (the buggy store-convert combine never reaches selection
at `-O2` for that code; the mantissa stays in registers). So the `-O1`/`-O2`
miscompile is a different bug and remains open. I have therefore not used a
closing keyword; #847 should stay open for the miscompile.

Unrelated: #995 ("ICE in mm_bfp.cc: unable to legalize G_BUILD_VECTOR
`<8 x s8>`") is a distinct legalizer failure, not touched here.


llvm/lib/Target/AIE/aie2p/AIE2PInstructionSelector.cpp
```diff
@@ -2612,6 +2612,22 @@ AIE2PInstructionSelector::getCombinedOpcodeCONVStore(
...
+  // store-convert and read the conversion source from operand 2, which is the
+  // intrinsic ID -- these conversions have two defs, shifting the source to
+  // operand 3 -- so getReg() trips the isReg() assertion in
+  // selectG_AIE_STORE_CONV. See the -O0 ICE in
+  // https://github.com/Xilinx/llvm-aie/issues/847.
+  if ((CombOpID == Intrinsic::aie2p_v64accfloat_to_v64bfp16ebs8 ||
+       CombOpID == Intrinsic::aie2p_v64accfloat_to_v64bfp16ebs16 ||
+       CombOpID == Intrinsic::aie2p_v64bfp16ebs8_to_v64bfp16ebs16) &&
+      MemOp.getOpcode() != AIE2P::G_INTRINSIC_W_SIDE_EFFECTS)
+    return {};
```
mludevid: Nit: `return std::nullopt;` for consistency.

mludevid (dismissed):
LGTM. Thanks for the fix!

llvm/lib/Target/AIE/aie2p/AIE2PInstructionSelector.cpp
atassis: @mludevid Done, thanks!

mludevid (approved):

PR #1060, opened 2026-06-28
[TableGen] Error on sub-register indices that overflow their register

atassis:
Depends on #1059. This is the TableGen check that would have caught #1059's bug
at build time, so the two go together: this branch includes #1059's one-line
`sub_bfp16_e` fix as its first commit (the verifier is a hard error and would
otherwise reject the unfixed AIE2P register info). Once #1059 lands, I will
rebase this down to just the verifier commit. I also intend to propose the
verifier upstream to llvm-project.

## What

A register defined with `CoveredBySubRegs` derives its sub-register lane masks
from the declared `Size`/`Offset` of its explicit `SubRegIndex`es. If a
`SubRegIndex` `Size` is wrong such that `offset + size` runs past the
register's own size, that sub-register gets an oversized lane mask. Nothing
diagnoses this today, so it surfaces as silently wrong sub-register liveness
and spilling: a wrong-code bug with no ICE and no machine-verifier error.

This adds a check, run when the register bank is built, that a register's
explicit sub-registers tile a contiguous bit range that fits within the
register's size.

## Why (a real bug)

This is exactly the `-O1`/`-O2` miscompile from #847: `sub_bfp16_e` (the 64-bit
bfp16 block exponent) was declared with `Size` 576, so the exponent's lane mask
spanned 576 bits instead of 64. Once register pressure forced the bfp16 MAC
operands to spill, the reload was corrupted (the first element of every
8-element block, governed by the shared exponent). TableGen accepted it without
a word; this check turns it into a build error. With #1059 applied, AIE2P
builds clean.

## Scope and false-positive handling

The check only applies to registers whose explicit sub-registers tile a
*contiguous* bit range. Strided/spaced register tuples (e.g. ARM's
`Tuples3DSpc = [dsub_0, dsub_2, dsub_4]`) intentionally place sub-registers at
non-adjacent offsets that extend past the nominal size; those have
non-contiguous ranges and are skipped.

I ran `-gen-register-info` with the check over every in-tree target. It is
clean on all of them (including AIE2P with #1059), and it fires on a register
deliberately given an oversized sub-register index.

## Test

`llvm/test/TableGen/subreg-index-overflow.td` defines a 64-bit pair whose high
half is mis-sized so the sub-registers cover 96 bits, and checks for the error.
The TableGen lit suite passes (374/374).

### DIFF
--- a/llvm/lib/Target/AIE/aie2p/AIE2PRegisterInfo.td
+++ b/llvm/lib/Target/AIE/aie2p/AIE2PRegisterInfo.td
@@ -60,7 +60,7 @@ def sub_hi_exp : SubRegIndex<32, 32>;
 def sub_bfp_exp : SubRegIndex<32,   0>;
 def sub_bfp_vec : SubRegIndex<256, 32>;
 
-def sub_bfp16_e : SubRegIndex<576,   512>;
+def sub_bfp16_e : SubRegIndex<64,   512>;
 def sub_bfp16_x : SubRegIndex<512, 0>;
 
 def sub_bfp576_lo : SubRegIndex<576,   0>;
--- a/llvm/test/TableGen/subreg-index-overflow-allowed.td
+++ b/llvm/test/TableGen/subreg-index-overflow-allowed.td
@@ -1,0 +1,37 @@
+// RUN: llvm-tblgen -gen-register-info -I %p/../../include %s -o - | FileCheck %s
+// Companion to subreg-index-overflow.td: patterns that look like an overflow
+// but are legitimate and must NOT be diagnosed by the size check.
+include "llvm/Target/Target.td"
+
+class MyReg<string n, list<Register> subs = []> : Register<n> {
+  let Namespace = "Test";
+  let SubRegs = subs;
+  let CoveredBySubRegs = 1;
+}
+class MyClass<int sz, list<ValueType> types, dag regs>
+  : RegisterClass<"Test", types, sz, regs> { let Size = sz; }
+
+def lo : SubRegIndex<32, 0>;
+def hi_gap : SubRegIndex<32, 64>; // leaves a [32,64) gap -> non-contiguous
+def lo2 : SubRegIndex<32, 0>;
+def hi2 : SubRegIndex<32, 32>;
+
+foreach i = 0-7 in { def R#i : MyReg<"r"#i>; }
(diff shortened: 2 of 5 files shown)


mludevid:
> and it fires on a register
deliberately given an oversized sub-register index.

@atassis, could you elaborate on this? When does this happen?

This feature is very nice, but we usually try to keep changes to upstream code to a minimum because any diff in generic code has the potential to create conflicts when we bump upstream llvm. Have you started the upstreaming process? Once it has landed in upstream llvm we can evaluate if we want to cherry-pick the patch or if we are fine with waiting until the patch reaches us in a future upstream bump.

atassis:
@mludevid sure. "Deliberately" there is just the lit test - `subreg-index-overflow.td` mis-sizes a sub-register on purpose so it overflows, to prove the check fires. It never happens legitimately in a real backend; an oversized sub-register index is always a bug, and the only one in the wild is the sub_bfp16_e case this PR fixes.
And yes - upstreaming is started: [llvm/llvm-project#206346](https://github.com/llvm/llvm-project/pull/206346), already in review with @jurahul. Agree on keeping the fork diff minimal; once it lands upstream you can cherry-pick or just wait for the next bump.

mludevid:
Sounds good! I've subscribed to the upstream PR and I'll try to keep up to date with it. In any case, a friendly reminder when that PR lands would be appreciated. We can cherry-pick it then into our fork. Thanks for looking into this @atassis!

atassis:
@mludevid Hi there. Sorry for being late, got cold few days ago, slowly coming back to life. I'll try to bump all xilinx/amd related work in a few days, including this PR. Thanks for merging another PR, btw!

mludevid:
Hey @atassis! No worries, hope you're feeling better! I'll also try go get back to your other bfp16 PR this week.

atassis:
So I have bumped my PR in llvm-project, but maybe we should let this PR to sit for a while?  So if my llvm work lands in upstream- I'll re-review the whole work again and close or fix this PR in place?

mludevid:
> So I have bumped my PR in llvm-project, but maybe we should let this PR to sit for a while? So if my llvm work lands in upstream- I'll re-review the whole work again and close or fix this PR in place?

@atassis, sounds good!

atassis:
@mludevid hi there
My work has just landed onto LLVM, so the latest upstream should carry the verifier there.
So I am closing this PR then as agreed. thanks for attention!

PR #1061, opened 2026-06-28
Martien.maxlatencyfinder fixcrash

martien-de-jong:
This fixes a crash that was observed in a consumer CI pipeline, The crash was dereferencing a nullptr parent basic block of a prologue instruction in and InterBlockEdges while building the edges.
It has been fixed by temporarily putting them the prologue instruction into the block where they should ultimately land, and removing them again before we definitely insert the bundles.

### DIFF
--- a/llvm/lib/Target/AIE/AIEInterBlockScheduling.cpp
+++ b/llvm/lib/Target/AIE/AIEInterBlockScheduling.cpp
@@ -978,7 +978,10 @@ BlockState &InterBlockScheduling::getBlockState(MachineBasicBlock *BB) {
 }
 
 void InterBlockScheduling::buildGraph(InterBlockEdges &DAG) {
-  const BlockState &BS = getBlockState(DAG.getPred());
+
+  MachineBasicBlock *PredBB = DAG.getPred();
+
+  const BlockState &BS = getBlockState(PredBB);
   const Region &Bot = BS.getBottom();
 
   // Pre-boundary: free instructions of the current region.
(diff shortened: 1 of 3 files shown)


llvm/test/CodeGen/AIE/aie2ps/schedule/postpipeliner/crash-bottom-insert-register-dep.mir
```diff
@@ -0,0 +1,186 @@
+# This file is licensed under the Apache License v2.0 with LLVM Exceptions.
+# See https://llvm.org/LICENSE.txt for license information.
+# SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+#
+# (c) Copyright 2026 Advanced Micro Devices, Inc. or its affiliates
+
+# XFAIL: *
+# RUN: llc -mtriple=aie2ps -O2 --start-before=postmisched %s -o /dev/null
+
+# Variant of crash-bottom-insert-semantic-order.mir that exercises the
```
F-Stuckmann: Can you add the comment of the Commit message here about what this tests, i.e. VLDA without MachineMemOperand.

Besides the commit message it is difficult to understand differentiates this test from the insert-semantic-order test

llvm/lib/Target/AIE/AIEInterBlockScheduling.cpp
```diff
@@ -1018,6 +1018,15 @@ void InterBlockScheduling::buildPerSuccEdges(MachineBasicBlock *BB) {
   for (MachineBasicBlock *SuccBB : BB->successors()) {
     DEBUG_BLOCKS(dbgs() << "Building InterBlockEdge: Pred=" << BB->getNumber()
                         << " Succ=" << SuccBB->getNumber() << "\n");
+    // BottomInsertSemanticOrder instructions are prologue clones that will
+    // eventually be inserted at the end of SuccBB by emitInterBlockBottom.
+    // Temporarily place them in SuccBB now so buildGraph can access
+    // MI->getParent() during dependency analysis without crashing.
+    BlockState &SuccBS = getBlockState(SuccBB);
```
F-Stuckmann: nit: wouldn't this also apply to the TopInsert?

mludevid (approved):

PR #1065, opened 2026-06-30
[AIE2PS] Fix VEC1024/ACC1024/ACC2048 spill encodability threshold (was overly conservative)

khallouh:
eliminateFrameIndex used isEncodableAsNegativeInt<9, 128> to decide whether VLDA/VST_Y/CM/DM_SPILL can encode the SP offset as an immediate. These pseudos expand into native dmx x/bm store/load instructions whose offset field is c16n_step64 (step 64, not 128), so the correct threshold is <9, 64>. Using step 128 incorrectly accepts offsets in the [-65536, -32768) window and passes them through as immediates to the expanded x/bm spill, producing an out-of-range encoding error at object-emission time.

Add mid-offset tests (frame ~50 KB, offset -50240) that land in the distinguishing window: too large for step 64 but in range for step 128. With the fix these correctly emit indexed addressing; without it they would produce an unencodable immediate.

### DIFF
--- a/llvm/test/CodeGen/AIE/aie2ps/eliminate-frame-index.mir
+++ b/llvm/test/CodeGen/AIE/aie2ps/eliminate-frame-index.mir
@@ -196,7 +196,7 @@ body:             |
     ; CHECK-NEXT: VST_dmx_sts_bm_spill $bmhl0, -128, implicit $sp :: (store (s512) into %stack.0 + 128)
     ; CHECK-NEXT: VST_dmx_sts_bm_spill $bmhh0, -64, implicit $sp :: (store (s512) into %stack.0 + 192)
     $dm0 = VLDA_DM_SPILL %stack.0, implicit $sp :: (load (s2048) from %stack.0, align 64)
-    VST_DM_SPILL $dm0, %stack.0, implicit $sp :: (store (s1024) into %stack.0, align 64)
+    VST_DM_SPILL $dm0, %stack.0, implicit $sp :: (store (s2048) into %stack.0, align 64)
 ...
 
 ---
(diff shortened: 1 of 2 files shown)


llvm/test/CodeGen/AIE/aie2ps/eliminate-frame-index.mir
```diff
@@ -965,6 +965,127 @@ body:             |
...
+    ; CHECK-NEXT: $dj0 = MOVXM_lng_cg -49920
+    ; CHECK-NEXT: VST_dmx_sts_bm_st_idx $bmhl0, $p0, killed $dj0 :: (store (s512) into %stack.0 + 128)
+    ; CHECK-NEXT: $dj0 = MOVXM_lng_cg -49856
+    ; CHECK-NEXT: VST_dmx_sts_bm_st_idx $bmhh0, killed $p0, killed $dj0 :: (store (s512) into %stack.0 + 192)
+    ; CHECK-NEXT: frame-destroy PADDXM_pstm_sp_imm -50304, implicit-def $sp, implicit $sp
+    ; CHECK-NEXT: PseudoRET implicit $lr
+    renamable $p0 = PseudoFI %stack.0, implicit $sp
+    ADJCALLSTACKUP 0, 0, implicit-def $sp, implicit $sp
+    $dm0 = VLDA_DM_SPILL %stack.0, implicit $sp :: (load (s2048) from %stack.0, align 64)
+    VST_DM_SPILL $dm0, %stack.0, implicit $sp :: (store (s1024) into %stack.0, align 64)
```
mludevid: Nit: the MMO size is wrong.

mludevid (dismissed):
LGTM.

mludevid (approved):

PR #1068, opened 2026-07-03
[AIEX] update_ore_test_checks: don't chain asm update for remark-only…

F-Stuckmann:
… tests

With --update-llc-checks, update_ore_test_checks.py stripped the remark RUN lines and always invoked update_llc_test_checks.py. For a test whose only RUN lines are remark runs, that handed the sibling script a file with no RUN lines, which crashed in update_llc_test_checks.py main() with an UnboundLocalError (output_type is only bound inside the per-RUN-line loop).

Fix it structurally at the orchestration layer: only delegate to the asm/MIR update script when the test actually has a non-remark RUN line to regenerate (has_asm_run_lines). A remark-only test now skips the asm step and just regenerates its remark CHECK lines.

Add a regression test (x86-ore-chain-update-no-asm) covering a remark-only test updated with --update-llc-checks, including UTC_ARGS idempotency.


llvm/test/tools/UpdateTestChecks/update_ore_test_checks/x86-ore-chain-update-no-asm.test
```diff
@@ -0,0 +1,12 @@
...
+## asm-update step must be skipped because there is nothing to regenerate;
+## previously it stripped the remark RUN line and crashed the sibling
+## update_llc_test_checks.py with an UnboundLocalError on an empty run list.
+
+# RUN: cp -f %S/Inputs/x86-ore-chain-update-no-asm.ll %t.ll && %update_ore_test_checks --update-llc-checks %t.ll
+# RUN: diff -u %S/Inputs/x86-ore-chain-update-no-asm.ll.expected %t.ll
+
+## Idempotency via UTC_ARGS (no --update-llc-checks on the command line):
+# RUN: %update_ore_test_checks %t.ll
+# RUN: diff -u %S/Inputs/x86-ore-chain-update-no-asm.ll.expected %t.ll
```
martien-de-jong, line 12: Silly that diff fails on difference. diff should fail on equality.
if diff a b
then
  echo "files are equal"
fi

martien-de-jong (approved):

PR #1070, opened 2026-07-03
[AIE] Enforce release→VST separation in lock resume delay 

katerynamuts:


llvm/lib/Target/AIE/AIEBaseInstrInfo.cpp
khallouh, line 1: Can you explain the problem with co-issueing? If the CoreResumeCycle is correct and the FirstIssueMemCycle is bigger, then we should still preserve the ordering even if they are co-issued or even if the store is issued before the release, right? Or do all our stores use `lckStRsrc` resource in exactly the same cycle as the release?

llvm/lib/Target/AIE/AIEBaseInstrInfo.cpp
```diff
@@ -1074,7 +1074,12 @@ AIEBaseInstrInfo::getLockResumeDelay(const MachineInstr &Lock,
                                   ? ReleaseResumeCycle
                                   : getCoreResumeCycleAfterLock();
 
-  return CoreResumeCycle - *OptFirstMemCycle + 1;
+  const bool IsStoreOnly = MemMI.mayStore() && !MemMI.mayLoad();
```
martien-de-jong: nit: We have precondition isLoad || isStore after the first early return, so !isLoad is equivalent to IsStoreOnly. 
Given the important of IsLoad and IsStore in this code I would get them once at the start of the function, and also run that early return off of them.

llvm/lib/Target/AIE/AIEBaseInstrInfo.cpp
```diff
@@ -1074,7 +1074,12 @@ AIEBaseInstrInfo::getLockResumeDelay(const MachineInstr &Lock,
...
+  const int FirstIssueMemCycle =
+      isRelease(Lock.getOpcode()) && IsStoreOnly
+          ? std::min(*OptFirstMemCycle, ReleaseResumeCycle)
+          : *OptFirstMemCycle;
+  return CoreResumeCycle - FirstIssueMemCycle + 1;
```
martien-de-jong, line 1083: I had expected the test to change in this commit, and I didn't recognize the problem in the precommit.  Are the references in that first commit up-to-date with the code?

llvm/lib/Target/AIE/AIEBaseInstrInfo.cpp
katerynamuts: done but I left  `const bool IsStoreOnly = IsStore && !IsLoad;` for readability.

khallouh (approved):

PR #1045, opened 2026-06-22
[AIEX] Refactor outerloop Pipeliner

F-Stuckmann:
NFC refactors for clarity. This is not exhaustive and just a first walk through.

PR #1072, opened 2026-07-09
[AIEX] OuterLoopPipeliner Bug fix 1 

F-Stuckmann:
@khallouh the first bug fix for the OuterLoopPipeliner. You can see that i only discovered it after extensive refactors. 

I don't think properly porting the bug fix into nice 2 single commits is worth the effort, because the [NFC refactor](https://github.com/Xilinx/llvm-aie/pull/1045) will clean up this intermediate state anyway again. 

We may want to squash the refactors though.


martien-de-jong:
Can we at least have a description of the bug it fixes and a pointer to where it is fixed? This is unreviewable. I think squashing makes sense both jere and in the NFC PR.

F-Stuckmann:
I am splitting the OLP refactor into NFC parts that reimplement the bugs and the fixes on top of the refactor. that will be clearer to review.

PR #1074, opened 2026-07-09
[AIE2PS] Remove dead code patching spill offsets based on MMO

khallouh:
The patch is no longer needed with the safer approach https://github.com/Xilinx/llvm-aie/pull/1031


mludevid (approved):
LGTM.

khallouh:
As requested, I've run a few mllib kernels for safety and there were no issues

PR #1075, opened 2026-07-10
[AIEX] Optimization guide update

F-Stuckmann:
Added sections for

- shuffled storing
- avoid peeling iterations out of loops

### DIFF
--- a/PERF_OPT_GUIDE.md
+++ b/PERF_OPT_GUIDE.md
@@ -809,6 +809,38 @@ for (int i = 0; i < count * 8; i++) {
 Multiply the trip count by the former unroll factor and set
 `AIE_LOOP_RANGE` to guarantee the compiler the new minimum.
 
+### Avoid manually peeling the first iteration
+
+A related anti-pattern is **hand-peeling the first loop iteration** --
+pre-loading inputs and issuing the first computation *before* the loop, then
+starting the loop at the second iteration. This steals iterations from the
+pipeliner, which is the toolchain's expert at loop peeling: it already
+constructs an optimal prologue and decides how many iterations to peel.
+Hand-peeling cannot lower the achieved II and only takes that decision away
+from the component best equipped to make it.
+
+Prefer a single loop from `i = 0` with pre-zeroed accumulators. Seeding an
+accumulator with `zero + first_product` is numerically identical to the
+peeled version, so no correctness is lost:
+
+```cpp
+// BAD: first iteration peeled before the loop, which starts at i = 1.
+acc = first_product(...);                       // duplicated body
+for (unsigned i = 1; i < n; ++i) acc = mac(acc, ...);
+
+// GOOD: single body from i = 0, accumulator pre-zeroed.
+acc = aie::zeros<...>();
+for (unsigned i = 0; i < n; ++i) acc = mac(acc, ...);
+```
+
+**Exception -- deliberate peeling in nested loops.** When an inner
+software-pipelined loop's prologue/epilogue is too short to hide a
+long-latency instruction in the *outer* loop, peeling a few inner iterations
+up into the outer loop lets them run in parallel with that outer-loop latency.
+This is a deliberate scheduling trade, not accidental duplication -- justify it
+with a measured II/latency improvement (Section 14), since the default guidance
+above still holds for the common single-loop case.
+
 ### AIE_LOOP_UNROLL for controlled unrolling
 When the loop body is too small for the pipeliner to fill all VLIW
 slots, controlled unrolling can help by giving the pipeliner a larger
--- a/PERF_OPT_GUIDE.md
+++ b/PERF_OPT_GUIDE.md
@@ -326,7 +326,13 @@ warning plus an `-Rpass-missed=aie-multi-slot-pseudo` remark. See
 Section 14 for how to read these hints.
 
 
-## 4. Sub-32-Bit Store Limitations
+## 4. Store Width and Layout
+
+How a hot loop writes its results out can have a large effect on
+throughput. The following sections cover three independent techniques for
+making stores cheaper.
+
+### Sub-32-bit stores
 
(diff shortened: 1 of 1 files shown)


PERF_OPT_GUIDE.md
```diff
@@ -565,6 +565,10 @@ for (int i = 0; i < param.count; i++) { ... }
 If you set the minimum too low, the pipeliner may produce suboptimal code. If you set it too high
 (higher than the actual runtime count), the program will be incorrect.
 
+**Deriving a safe minimum from the tiling.** The minimum you assert must be
+one the caller's tiling *guarantees*, not a wish -- if the runtime count can
+fall below it, the program is incorrect.
```
martien-de-jong: This repeats the previous sentence in a less clear wording. 'Tiling' may or may not be in the scope of the programmer, and the section doesn't actually explain how to derive that minimum,

PERF_OPT_GUIDE.md
```diff
@@ -809,6 +875,28 @@ for (int i = 0; i < count * 8; i++) {
 Multiply the trip count by the former unroll factor and set
 `AIE_LOOP_RANGE` to guarantee the compiler the new minimum.
 
+### Avoid manually peeling the first iteration
```
mludevid, line 872: I've seen kernels where peeling iterations from the front or the back is indeed useful in nested loops where the prologue and epilogue of the SWP inner-loop is not enough to hide some large latency in the outer loop. In those cases it makes sense to peel some iterations into the outer loop to run them in parallel with that large latency instruction from the outer loop. Maybe we should mention this special case here.

PERF_OPT_GUIDE.md
```diff
@@ -809,6 +813,28 @@ for (int i = 0; i < count * 8; i++) {
...
+
+A related anti-pattern is **hand-peeling the first loop iteration** --
+pre-loading inputs and issuing the first computation *before* the loop, then
+starting the loop at the second iteration. This duplicates the loop body into a
+hand-written prologue, inflating program memory (PM) and lowering performance of the prologue. 
+It does not lower the achieved II, because the pipeliner already builds an optimal prologue.
```
martien-de-jong: I would shift the focus a bit. I wouldn't mention PM increase or prologue efficiency, I would say that you're stealing minitercount from the pipeliner, who is our top expert in loop peeling.
I think it's also worth finding out whether non-zero iv inits hamber hwloop creation.

PERF_OPT_GUIDE.md
```diff
@@ -326,7 +326,15 @@ warning plus an `-Rpass-missed=aie-multi-slot-pseudo` remark. See
...
 
 
-## 4. Sub-32-Bit Store Limitations
+## 4. Store Width and Layout
+
+The width and layout of stores in a hot loop's write-out path have a
+large effect on throughput. Three techniques matter: keep every store at
+least 32 bits wide, consider letting the store's *addressing* produce the
+destination layout instead of shuffling data into place first, and fold any
+output scaling into the conversion that already runs before the store.
```
martien-de-jong: I don't know what you mean here. I think you should elaborate with a few examples.

Oh, reading on I find the elaboration. Perhaps rewrite to less detailed intro, announcing sections on the subjects.

PERF_OPT_GUIDE.md
```diff
@@ -378,6 +386,60 @@ packed[3] = d;
 aie::store_v(ofm_ptr, packed);  // single 32-bit store
 ```
 
+### Consider storing to the destination layout directly instead of shuffling
+
+When results are computed in one register layout but memory needs another,
+the natural code shuffles/interleaves into the target layout and stores
+contiguously. If the shuffle exists *only* to feed a contiguous store, it
```
martien-de-jong, line 391: I think it's difficult to have guidelines here. 'shuffle' can be built from insert, extract + insert, real shuffles. insert and extract on 4x8 integer vectors takes two alu per element, whereas consecutive stores may be coalesced. And you probably don't want to separately store 8x32. Then there's the random collection of shuffle patterns that work quite well.

PERF_OPT_GUIDE.md
```diff
@@ -378,6 +386,60 @@ packed[3] = d;
...
+achieved II on the actual target and memory layout.
+
+Two constraints gate whether the direct-store form is worth measuring:
+
+- **Each sub-vector store must stay >=32 bits wide** (the rule above). If a
+  sub-vector would be narrower than 32 bits, the read-modify-write penalty
+  almost always outweighs the saved shuffles -- keep the shuffle.
+- **Extract indices must be compile-time constants.** A runtime `extract<>`
+  index generates a branchy dynamic extract. Hand-unroll the stores over
+  the statically known count and guard with a `static_assert` so a layout
```
martien-de-jong, line 410: aie2ps has a register lane operand in insert and extract

PERF_OPT_GUIDE.md
F-Stuckmann: boy let's hope AI understands all the special cases we are adding

PERF_OPT_GUIDE.md
F-Stuckmann: i'll just delete it, i haven't seen the AI struggle to determine the correct minimum iteration count.

PERF_OPT_GUIDE.md
F-Stuckmann: we always create hwloops when we can get SCEV.

PERF_OPT_GUIDE.md
F-Stuckmann: the guide specifically asks for empirical decision metrics. This is just a trick to apply.

PERF_OPT_GUIDE.md
F-Stuckmann: added a condition to the statement, since this also covers aie2p and aie2

mludevid (approved):

mludevid:
Please squash before merging, thanks!

PR #1078, opened 2026-07-10
[AIE] Add inner-loop versioning to pipeline small-trip-count loops

mludevid:
This PR adds inner-loop versioning for AIE targets so the post-pipeliner can software-pipeline inner loops even when their statically known minimum trip count is too small. A loop marked with the aie-loop-versioning hint is split into two runtime-guarded copies: a pipelined fast copy and a verbatim, un-pipelined slow copy. A trip-count check chooses between them at runtime, so small trip counts stay correct while large ones get the pipelined schedule.


andcarminati:
Could we merge the first commit in advance?

llvm/lib/Target/AIE/Utils/AIELoopUtils.h
```diff
@@ -35,6 +35,19 @@ bool hasIIPragma(const MachineBasicBlock &LoopBlock);
 
 bool isOuterLoopPipelined(const MachineBasicBlock &LoopLatch);
 
+/// Loop-hint key that enables AIE loop versioning. It is set by the IR
+/// versioning pass on the pipelined ("fast") copy of a versioned loop.
+constexpr StringLiteral LoopVersioningHintKey =
```
andcarminati, line 47: Maybe we could clarify this. Is this flag really enabling versioning as the comment says? Maybe it is just signaling the fast copy.  If you see outer loop pipeliner, we separate the flag that request the pipeliner from the flag that marks the loop as pipelined. If this marks a loop as versioned, we could rename to "llvm.loop.hint.aie-loop-versioned". Hints can be seen as pure loop metadata.

andcarminati: What happens if we run the versioner twice?

llvm/lib/Target/AIE/aie2ps/AIE2PSInstrInfo.cpp
```diff
@@ -1366,6 +1366,10 @@ AIE2PSInstrInfo::getZOLSupport() const {
   return Result;
 }
 
+std::optional<unsigned> AIE2PSInstrInfo::getLoopVersionThresholdOpcode() const {
+  return AIE2PS::PseudoLoopVersionThreshold;
```
andcarminati, line 1385: An easier alternative could be a simple scalar move with a fixup type for the post pipeliner. We just need to recognize the fixup and replace by a constant.

llvm/lib/Target/AIE/AIEInnerLoopVersioning.cpp
```diff
@@ -0,0 +1,363 @@
...
+  // computed / expanded there.
+  Value *expandTripCount(Loop *L, Instruction *InsertPt) const;
+  // Clone L into a slow copy and route a runtime guard to either copy.
+  bool versionLoop(Loop *L);
+  // Merge loop-defined values used after the loop across the two copies.
+  void addExitPHIs(Loop *FastLoop, Loop *SlowLoop, BasicBlock *ExitBlock,
+                   ValueToValueMapTy &VMap,
+                   ArrayRef<Instruction *> DefsUsedOutside) const;
+  // Rebuild L's loop id without any metadata entry whose key is in Keys.
+  void stripLoopMetadata(Loop *L, ArrayRef<StringRef> Keys) const;
```
andcarminati: I would love to see this in some sort of utils.

llvm/lib/Target/AIE/AIEInnerLoopVersioning.cpp
```diff
@@ -0,0 +1,363 @@
...
+  // Remove the versioning hint so the postpipeliner leaves the slow copy alone.
+  void stripVersioningHint(Loop *L) const;
```
andcarminati: I strong recommend two flags: to enable and to identify versioned loops. We should avoid problems related to double run of the pass.

llvm/lib/Target/AIE/AIEInnerLoopVersioning.cpp
```diff
@@ -0,0 +1,363 @@
...
+
+  LLVM_DEBUG(dbgs() << "  Versioning loop "; Header->printAsOperand(dbgs());
+             dbgs() << "\n");
+
+  // Build the runtime check in the (soon to be) guard block. The threshold is
+  // produced by the thin intrinsic (a placeholder); the postpipeliner patches
+  // it later. Take the slow copy when the trip count is below the threshold.
+  //
+  // The placeholder is the maximal unsigned value (-1 == UINT32_MAX), not 0, so
+  // that an un-patched guard fails safe: the unsigned compare then selects the
```
andcarminati: As we pipeline loops as a first step in the scheduler, drop loops belonging to failed cases should be feasible as an extension.

llvm/lib/Target/AIE/AIEInnerLoopVersioning.cpp
andcarminati: Maybe e could share some code with the outerloop pipeliner.

llvm/lib/Target/AIE/AIEBaseInstrInfo.cpp
```diff
@@ -421,6 +422,28 @@ void AIEBaseInstrInfo::adjustTripCount(MachineInstr &MI, int Adjustment) const {
...
+  assert(isLoopVersionThresholdDef(MI));
+  // The threshold rides in the pseudo's immediate operand, which materializes
+  // into a narrow scalar move (simm10 on AIE2, c11s elsewhere). NStages is a
+  // modulo-schedule stage count and thus tiny in practice; enforce that it fits
+  // the narrowest such field (deliberately the AIE2 simm10 width for all
+  // subtargets) so a pathological schedule fails loudly here instead of
+  // silently emitting a truncated immediate. This is checked unconditionally,
+  // not just under assert: a truncated threshold would select the pipelined
+  // fast copy for trip counts too small for the schedule, a miscompile.
+  if (!isInt<10>(MinTripCount))
```
F-Stuckmann, line 442: I think we should have a solution for itercount > 10 that does not crash

llvm/test/CodeGen/AIE/aie2ps/loop-version-threshold.ll
```diff
@@ -0,0 +1,24 @@
...
+; the post-RA scheduler materializes it into a concrete scalar move.
+
+; SELECT-LABEL: name: threshold
+; SELECT: PseudoLoopVersionThreshold 9
+
+; MOV-LABEL: name: threshold
+; MOV: MOV_alu_mv_mv_mv_cg_or 9
+define i32 @threshold() {
+entry:
+  %t = call i32 @llvm.aie2ps.loop.version.threshold(i32 9)
```
F-Stuckmann, line 19: nit: if you want an upper limit on itercount, you may want to have a dedicated test for it

llvm/lib/Target/AIE/AIEPostPipeliner.cpp
```diff
@@ -152,14 +154,24 @@ bool PostPipeliner::isPostPipelineCandidate(MachineBasicBlock &LoopBlock) {
   // count is checked before accepting the schedule.
   using namespace AIELoopUtils;
   auto ParsedMinTripCount = getMinTripCount(LoopBlock);
-  if (!ParsedMinTripCount) {
```
F-Stuckmann, line 155: could we instead of deleting this code section and adding nesting levels have a early exit for the LoopVersioning?

llvm/lib/Target/AIE/aie2ps/AIE2PSInstrInfo.cpp
F-Stuckmann: this is a deliberate stability choice, because otherwise we have to be very certain that we correctly identified the correct branching instruction

mludevid: The pseudo is required for correctness. We don't want the MachineCSE to modify any of these guards, so we need to hide the constant behind a pseudo.

llvm/lib/Target/AIE/AIEBaseInstrInfo.cpp
mludevid: Nonono, this is 2^10. I don't see how we would ever run into an innner loop with more than 1000 stages hahaha.

llvm/lib/Target/AIE/AIEPostPipeliner.cpp
```diff
@@ -1760,6 +1777,137 @@ void PostPipeliner::updateTripCount() const {
...
+  if (Guard->succ_size() != 2)
+    return nullptr;
+  return Guard;
+}
+
+// The register tested by the guard's conditional branch (the trip-count compare
+// result), or an invalid register if none. The branch is the instruction that
+// both targets a block and reads a register.
+static Register findGuardConditionReg(MachineBasicBlock &Guard) {
+  Register CondReg;
```
F-Stuckmann: could we add more asserts about what structure the guard is, i.e. MachineInstr ?
We also assume a specific layout structure, which we may also want to guard in asserts.

llvm/lib/Target/AIE/AIEPostPipeliner.cpp
```diff
@@ -1760,6 +1777,137 @@ void PostPipeliner::updateTripCount() const {
...
+    // Start scanning just above the compare: register allocation may reuse the
+    // compare's result register for one of its source operands (e.g. the
+    // threshold), so scanning from the compare itself would spuriously return
+    // the compare as its own operand's reaching def.
+    MachineInstr *Def =
+        findReachingDef(*Cmp.getParent(), std::next(Cmp.getReverseIterator()),
+                        MO.getReg(), TRI, Visited);
+    if (Def && TII->isLoopVersionThresholdDef(*Def)) {
+      if (Threshold && Threshold != Def)
+        return nullptr; // ambiguous
```
F-Stuckmann: Martien would comment to tell your AI to remove trailing comments ;)

llvm/lib/Target/AIE/AIEPostPipeliner.cpp
```diff
@@ -1760,6 +1777,137 @@ void PostPipeliner::updateTripCount() const {
...
+  // placeholder with the actual stage count. Because the pseudo is
+  // non-CSE-able, this binds to exactly this loop's guard.
+  //
+  // If any step fails to recognize the guard -- e.g. a future MIR pass reshaped
+  // the guard->preheader region -- bail without patching. This is safe: the IR
+  // pass seeds the placeholder with the maximal unsigned threshold, so an
+  // un-patched guard always selects the slow (verbatim) copy and the pipelined
+  // fast copy is simply never entered. We emit a debug message so the lost
+  // pipelining opportunity is diagnosable, but we do not abort the compile.
+  auto Bail = [&](const char *Msg) {
```
F-Stuckmann: nit: maybe rename to include MSG in the lamda?

llvm/lib/Target/AIE/AIEPostPipeliner.cpp
```diff
@@ -1760,6 +1777,137 @@ void PostPipeliner::updateTripCount() const {
...
+
+  Register CondReg = findGuardConditionReg(*Guard);
+  if (!CondReg.isValid())
+    return Bail("guard conditional branch not found");
+
+  SmallPtrSet<MachineBasicBlock *, 8> Visited;
+  MachineInstr *Cmp =
+      findReachingDef(*Guard, Guard->rbegin(), CondReg, TRI, Visited);
+  if (!Cmp)
+    return Bail("guard compare not found");
```
F-Stuckmann: I find it confusing that we return void

llvm/lib/Target/AIE/AIEPostPipeliner.cpp
```diff
@@ -1760,6 +1777,137 @@ void PostPipeliner::updateTripCount() const {
   TII->adjustTripCount(*TripCountDef, -Delta);
 }
 
+// Reaching definition of Reg seen from Use, scanning backwards into
+// predecessors -- the threshold pseudo may sit in a dominating block.
+static MachineInstr *
+findReachingDef(MachineBasicBlock &BB, MachineBasicBlock::reverse_iterator Use,
```
F-Stuckmann: Do we have to reimplement reachingDef? I have a suspicion we may not need to reimplement it

llvm/lib/Target/AIE/AIEInnerLoopVersioning.cpp
```diff
@@ -0,0 +1,363 @@
...
+//===----------------------------------------------------------------------===//
+//
+// Inner loop versioning for AIE.
+//
+// A loop carrying the llvm.loop.hint.aie-loop-versioning hint is split into two
+// copies guarded by a runtime trip-count check, so the postpipeliner can
+// pipeline the fast copy as aggressively as it likes while small trip counts
+// stay correct on the slow copy:
+//
+//   [preheader / guard]  if (trip < threshold) -> slow else -> fast
```
F-Stuckmann: In the previous commit you have an explicit Layout structure for loop versioning. You may want to harmonize this diagram with that layout structure

llvm/lib/Target/AIE/AIEInnerLoopVersioning.cpp
mludevid: I think this would be pretty involved. As far as I know, modifying the list of basic blocks during scheduling would require some serious infrastructure changes, e.g. updating our inter-block DAGs. I made sure to emit a warning to the user, because we don't want to be loop versioning if we cannot software pipeline, but I think this is something the user will have to deal with, either by making the necessary changes to enable software pipelining or by removing the loop versioning pragma.

llvm/lib/Target/AIE/AIEInnerLoopVersioning.cpp
```diff
@@ -0,0 +1,363 @@
...
+//
+// (c) Copyright 2026 Advanced Micro Devices, Inc. or its affiliates
+//
+//===----------------------------------------------------------------------===//
+//
+// Inner loop versioning for AIE.
+//
+// A loop carrying the llvm.loop.hint.aie-loop-versioning hint is split into two
+// copies guarded by a runtime trip-count check, so the postpipeliner can
+// pipeline the fast copy as aggressively as it likes while small trip counts
```
andcarminati: Fast copy is confusing here. The loop is the same, pipeliner will make it faster.

llvm/lib/Target/AIE/AIEInnerLoopVersioning.cpp
```diff
@@ -0,0 +1,363 @@
...
+//       |           |
+//    [slow]       [fast]   fast keeps the hint (postpipeliner pipelines it);
+//       \           /      slow is the verbatim, un-pipelined loop.
+//        \         /
+//         [exit]
+//
+// The threshold is produced by a thin per-subtarget intrinsic that lowers to a
+// non-CSE-able pseudo, so the postpipeliner has a stable handle to patch. Its
+// placeholder value is the maximal unsigned value (-1 == UINT32_MAX): because
+// the guard compare is unsigned, `trip < UINT32_MAX` is true for every trip
```
F-Stuckmann: I would recommend that fallthrough is the high iteration count case. 
FYI: I am working on making prologues movable to predecessors. 
this way we can speculatively execute parts of the postpipeliner prologue and hide startup times for the high iteration count.

llvm/lib/Target/AIE/AIEInnerLoopVersioning.cpp
```diff
@@ -0,0 +1,363 @@
...
+
+class AIEInnerLoopVersioning : public FunctionPass {
+public:
+  static char ID;
+  AIEInnerLoopVersioning() : FunctionPass(ID) {}
+
+  bool runOnFunction(Function &F) override;
+
+  void getAnalysisUsage(AnalysisUsage &AU) const override {
+    // We deliberately do not force LoopSimplify/LCSSA: that canonicalization
```
andcarminati, line 116: Maybe this could be moved to an earlier position, as this is part of the contract of the pass.

llvm/lib/Target/AIE/AIEInnerLoopVersioning.cpp
```diff
@@ -0,0 +1,363 @@
...
+
+bool AIEInnerLoopVersioning::versionLoop(Loop *L) {
+  // We need simplify form for a unique preheader and a single merge point for
+  // the two copies' results. Rather than force global LoopSimplify/LCSSA (which
+  // would perturb non-hinted loops seen by later passes), canonicalize just
+  // this hinted candidate on demand.
+  simplifyLoop(L, DT, LI, SE, /*AC=*/nullptr, /*MSSAU=*/nullptr,
+               /*PreserveLCSSA=*/false);
+  formLCSSARecursively(*L, *DT, LI, SE);
+  if (!L->isLoopSimplifyForm()) {
```
F-Stuckmann: could you add a requirement so that we can assert it?

llvm/lib/Target/AIE/AIEInnerLoopVersioning.cpp
```diff
@@ -0,0 +1,363 @@
...
+  OrigTerm->eraseFromParent();
+
+  DT->changeImmediateDominator(ExitBlock, GuardBB);
+  addExitPHIs(L, SlowLoop, ExitBlock, VMap, DefsUsedOutside);
+  formDedicatedExitBlocks(SlowLoop, DT, LI, nullptr, /*PreserveLCSSA=*/true);
+  formDedicatedExitBlocks(L, DT, LI, nullptr, /*PreserveLCSSA=*/true);
+
+  // The fast copy keeps the hint and gets pipelined; the slow copy is the
+  // verbatim un-pipelined fallback.
+  stripVersioningHint(SlowLoop);
```
F-Stuckmann: you may also want to strip the itercount in the slowloop

mludevid: Why? Imagine we can only guarnatee min iter count of 2 and we can generate a pretty good 2 stage pipeline for the "slow case" as well? Is there a reason we would want to prevent that?

llvm/lib/Target/AIE/AIEInnerLoopVersioning.cpp
```diff
@@ -0,0 +1,363 @@
...
+  StringRef getPassName() const override { return "AIE Inner Loop Versioning"; }
+
+private:
+  LoopInfo *LI = nullptr;
+  DominatorTree *DT = nullptr;
+  ScalarEvolution *SE = nullptr;
+  Intrinsic::ID ThresholdIID = Intrinsic::not_intrinsic;
+
+  // True when the loop carries the versioning enable hint.
+  bool isVersioningEnabled(const Loop *L) const;
```
andcarminati: maybe `isIterCountVersioningEnabled`

llvm/lib/Target/AIE/AIEInnerLoopVersioning.cpp
```diff
@@ -0,0 +1,363 @@
...
+  LLVMContext &Ctx = Header->getContext();
+  Type *I32Ty = Type::getInt32Ty(Ctx);
+
+  Value *TripCount = expandTripCount(L, GuardBB->getTerminator());
+  if (!TripCount) {
+    LLVM_DEBUG(dbgs() << "  Trip count not expandable\n");
+    return false;
+  }
+
+  LLVM_DEBUG(dbgs() << "  Versioning loop "; Header->printAsOperand(dbgs());
```
andcarminati: The code below could be refactored as a helper class, like `LoopVersioning`:
```
LoopVersioningl) LoopVer(L, DT, SE...);
LoopVer.versionLoop();
```

In `versionLoop`, we could have a sequence of method calls for each step. 

This can make the steps a bit more clear.

llvm/test/CodeGen/AIE/aie2ps/loop-version-threshold.ll
mludevid: Since the upper limit on itercount is unreachable in practice, I don't think a dedicated test is needed.

llvm/lib/Target/AIE/aie2ps/AIE2PSInstrInfo.cpp
andcarminati: But MachineCSE will not touch a fixup because it is not an immediate. You can always replace the fixup by an immediate in postmisched. If we cannot pipeline, we may assume MAX_INT there. Just an alternative solution.

andcarminati: But you can keep the current solution ;-)

mludevid: I still feel the pseudo is the nicer solution here. Besides the MachineCSE guarantee, isNotDuplicable + hasSideEffects also keeps the instruction from being rematerialized or moved around, and it gives the post-pipeliner a single opcode to recognize, so we're sure we're patching the right move.

llvm/lib/Target/AIE/AIEInnerLoopVersioning.cpp
```diff
@@ -0,0 +1,363 @@
...
+      L->getLoopID(), AIELoopUtils::LoopVersioningHintKey);
+  return Hint && *Hint > 0;
+}
+
+bool AIEInnerLoopVersioning::runOnFunction(Function &F) {
+  if (skipFunction(F))
+    return false;
+
+  Triple TT(F.getParent()->getTargetTriple());
+  ThresholdIID = getThresholdIntrinsic(TT);
```
andcarminati: This pretty cheap to query. I don't mind if we remove this field.

llvm/lib/Target/AIE/AIEInnerLoopVersioning.cpp
```diff
@@ -0,0 +1,363 @@
...
+INITIALIZE_PASS_DEPENDENCY(ScalarEvolutionWrapperPass)
+INITIALIZE_PASS_END(AIEInnerLoopVersioning, DEBUG_TYPE,
+                    "AIE Inner Loop Versioning", false, false)
+
+llvm::FunctionPass *llvm::createAIEInnerLoopVersioningPass() {
+  return new AIEInnerLoopVersioning();
+}
+
+/// Pick the subtarget-specific loop-version-threshold intrinsic.
+static Intrinsic::ID getThresholdIntrinsic(const Triple &TT) {
```
andcarminati: I am pretty sure @F-Stuckmann will ask this to be moved to `llvm/lib/Target/AIE/Utils/AIEIRUtils.h`

llvm/lib/Target/AIE/AIEInnerLoopVersioning.cpp
```diff
@@ -0,0 +1,363 @@
...
+  // Collect enabled loops before mutating, so cloning does not disturb
+  // iteration.
+  SmallVector<Loop *, 4> Candidates;
+  for (Loop *L : LI->getLoopsInPreorder())
+    if (isVersioningEnabled(L))
+      Candidates.push_back(L);
+
+  bool Changed = false;
+  for (Loop *L : Candidates) {
+    if (versionLoop(L)) {
```
andcarminati: As we may fail, we could call this `tryVersionLoop`

andcarminati:
general nit: can use `const` everywhere.

llvm/lib/Target/AIE/AIEInnerLoopVersioning.cpp
andcarminati: Also, it would be nice to have a clear separation between profitability check and transformation ;-).

llvm/lib/Target/AIE/AIEInnerLoopVersioning.cpp
```diff
@@ -0,0 +1,363 @@
...
+  // small minimum (e.g. 1) does not make the hardware-loop profitability gate
+  // reject the ZOL the postpipeliner needs. The slow copy keeps whatever it
+  // had, as it stays un-pipelined either way.
+  stripLoopMetadata(L, StringRef("llvm.loop.itercount.range"));
+
+  return true;
+}
+
+void AIEInnerLoopVersioning::addExitPHIs(
+    Loop *FastLoop, Loop *SlowLoop, BasicBlock *ExitBlock,
```
andcarminati: Those fast and slow terms are confusing to me. They are not fast and slow, they are the same at this point. Could we agree on a more descriptive name, like LowIterationLoop and HighIterationLoop? You can chose another name.

llvm/lib/Target/AIE/AIEInnerLoopVersioning.cpp
```diff
@@ -0,0 +1,363 @@
...
+    ValueToValueMapTy &VMap, ArrayRef<Instruction *> DefsUsedOutside) const {
+  PHINode *PN;
```
andcarminati: Please, initialize this pointer.

llvm/lib/Target/AIE/AIEInnerLoopVersioning.cpp
```diff
@@ -0,0 +1,363 @@
...
+
+  SmallVector<Instruction *, 8> DefsUsedOutside = findDefsUsedOutsideOfLoop(L);
+
+  // Split off an empty preheader for the fast (original) copy, then clone the
+  // loop into the slow copy dominated by the guard block.
+  BasicBlock *FastPH = SplitBlock(GuardBB, GuardBB->getTerminator(), DT, LI,
+                                  nullptr, Header->getName() + ".ph");
+  ValueToValueMapTy VMap;
+  SmallVector<BasicBlock *, 8> SlowBlocks;
+  Loop *SlowLoop = cloneLoopWithPreheader(FastPH, GuardBB, L, VMap,
```
andcarminati: Logically, I was thinking the other way around: Versioning is the creation of a "faster" copy of the loop. Here we create a "slower" copy, and we make the original ''faster". Is this intended to simplify metadata handling?

andcarminati:
Could we have .td selection patterns for the intrinsics? It may reduce some code.

mludevid:
> Could we have .td selection patterns for the intrinsics? It may reduce some code.

Instruction selection of intrinsics with side effects is not handled by patterns. I'm gonna leave the selection code as is.

llvm/lib/Target/AIE/Utils/AIELoopUtils.h
mludevid: Done. I've created a dedicated flag that signals we've run loop versioning.

llvm/lib/Target/AIE/AIEInnerLoopVersioning.cpp
mludevid: Done.

llvm/lib/Target/AIE/AIEPostPipeliner.cpp
mludevid: We don't crash on incorrect layouts. So I think an assert is too harsh. I was emitting warnings when bailing. I'm planning to emit optimization remarks instead. I hope that addresses your concerns.

llvm/lib/Target/AIE/AIEPostPipeliner.cpp
mludevid: Done.

llvm/lib/Target/AIE/AIEPostPipeliner.cpp
mludevid: I've rewrote this to explicitly return void.

llvm/lib/Target/AIE/AIEInnerLoopVersioning.cpp
mludevid: Done.

llvm/lib/Target/AIE/AIEInnerLoopVersioning.cpp
mludevid: I've renamed everything to high iteration and low iteration count.

llvm/lib/Target/AIE/AIEInnerLoopVersioning.cpp
mludevid: I don't think this makes a difference. The delay slots of the loop version guard are paid by both control flow, so there is no runtime difference there. The real difference is in the jump to the exit block. Whichever version gets placed by the machine block placement directly before the exit block saves an unconditional jump and the corresponding delay slots. But this is not decided by this code.

llvm/lib/Target/AIE/AIEInnerLoopVersioning.cpp
mludevid: Done.

llvm/lib/Target/AIE/AIEInnerLoopVersioning.cpp
mludevid: Done.

llvm/lib/Target/AIE/AIEInnerLoopVersioning.cpp
mludevid: Done.

llvm/lib/Target/AIE/AIEInnerLoopVersioning.cpp
mludevid: Done.

llvm/lib/Target/AIE/AIEInnerLoopVersioning.cpp
mludevid: No, the metadata is copied over. I don't think it makes a difference since both copies of the loop are identical. I've changed it so that the high iteration count loop is the cloned one. I hope this is more intuitive.

llvm/lib/Target/AIE/AIEInnerLoopVersioning.cpp
mludevid: Done.

llvm/lib/Target/AIE/AIEInnerLoopVersioning.cpp
mludevid: Done.

llvm/lib/Target/AIE/AIEInnerLoopVersioning.cpp
mludevid: I've added an assert that the loop is in LCSSA form. I hope that's enough.

llvm/lib/Target/AIE/AIEPostPipeliner.cpp
mludevid: Done.

llvm/lib/Target/AIE/AIEPostPipeliner.cpp
mludevid: I'm not sure what we could reuse for this. If you have a specific suggestion please let me know. I'm not aware of any.

llvm/lib/Target/AIE/AIEInnerLoopVersioning.cpp
mludevid: I've moved `dropLoopMetadata` into aie loop utils.

llvm/lib/Target/AIE/AIEPostPipeliner.cpp
mludevid: I've removed the extra nesting levels.

llvm/lib/Target/AIE/AIEInnerLoopVersioning.cpp
mludevid: Done.

llvm/lib/Target/AIE/AIEInnerLoopVersioning.cpp
mludevid: Where to? The file header? Personally I think it makes sense to keep it here.

llvm/lib/Target/AIE/Utils/AIEIRUtils.cpp
```diff
@@ -38,6 +46,35 @@ bool isHardwareLoopDecrement(const Instruction *I) {
...
+  if (TT.isAIE2())
+    return Intrinsic::aie2_loop_version_threshold;
+  if (TT.isAIE2P())
+    return Intrinsic::aie2p_loop_version_threshold;
+  if (TT.isAIE2PS())
+    return Intrinsic::aie2ps_loop_version_threshold;
+  return Intrinsic::not_intrinsic;
+}
+
+void dropLoopMetadata(Loop &L, ArrayRef<StringRef> Keys) {
```
F-Stuckmann: CHECK: we are dropping the Key metadata, correct? You may want to rename the argument to clarify the intention.

llvm/lib/Target/AIE/AIEBaseInstrInfo.cpp
```diff
@@ -421,6 +422,26 @@ void AIEBaseInstrInfo::adjustTripCount(MachineInstr &MI, int Adjustment) const {
   Imm.setImm(Imm.getImm() + Adjustment);
 }
 
+bool AIEBaseInstrInfo::isLoopVersionThresholdDef(const MachineInstr &MI) const {
+  auto Opcode = getLoopVersionThresholdOpcode();
```
F-Stuckmann: nit: you can cost the opcode

llvm/lib/Target/AIE/AIEBaseInstrInfo.cpp
```diff
@@ -436,8 +436,8 @@ void AIEBaseInstrInfo::setLoopVersionThreshold(MachineInstr &MI,
   // unconditionally, not just under assert: a truncated threshold would select
   // the fast copy for trip counts too small for the schedule, a miscompile.
   if (!isInt<10>(MinTripCount))
-    report_fatal_error(
-        "AIE loop versioning: threshold does not fit the scalar-move immediate");
+    report_fatal_error("AIE loop versioning: threshold does not fit the "
```
andcarminati: Just a matter of taste ;-). This fatal is in TII, but the message resembles something emitted by the loop versioning pass.

llvm/lib/Target/AIE/AIEPostPipeliner.cpp
```diff
@@ -1760,9 +1776,141 @@ void PostPipeliner::updateTripCount() const {
...
+  // -> compare -> the pseudo feeding it), then patch its placeholder with the
+  // stage count. If any step fails to recognize the guard (e.g. a later MIR
+  // pass reshaped the region), bail without patching: the IR pass seeds the
+  // placeholder with the maximal threshold, so an un-patched guard keeps the
+  // loop on the low-trip-count copy. Diagnosable via debug output, never fatal.
+  MachineOptimizationRemarkEmitter ORE(*Preheader->getParent(), nullptr);
+  // A fall-through preheader may be empty, so pick a DebugLoc defensively.
+  const DebugLoc GuardLoc =
+      Preheader->empty() ? DebugLoc() : Preheader->front().getDebugLoc();
+  auto bailWithMsg = [&](StringRef Msg) {
```
andcarminati: nit `BailWithMsg `

llvm/lib/Target/AIE/AIEPostPipeliner.cpp
```diff
@@ -1760,9 +1776,141 @@ void PostPipeliner::updateTripCount() const {
   TII->adjustTripCount(*TripCountDef, -Delta);
 }
 
-int PostPipeliner::getFinalMinTripCount() const {
-  const int Delta = NStages - 1;
-  return MinTripCount - Delta;
+// Reaching definition of Reg seen from Use, scanning backwards into
+// predecessors -- the threshold pseudo may sit in a dominating block.
+static MachineInstr *
+findReachingDef(MachineBasicBlock &BB, MachineBasicBlock::reverse_iterator Use,
```
andcarminati: This could live in a utility file.

llvm/lib/Target/AIE/AIEPostPipeliner.cpp
```diff
@@ -1760,9 +1776,141 @@ void PostPipeliner::updateTripCount() const {
...
+// both targets a block and reads a register.
+static Register findGuardConditionReg(MachineBasicBlock &Guard) {
+  Register CondReg;
+  for (MachineInstr &MI : Guard) {
+    bool TargetsBlock = false;
+    Register RegUse;
+    for (const MachineOperand &MO : MI.operands()) {
+      if (MO.isMBB())
+        TargetsBlock = true;
+      else if (MO.isReg() && MO.isUse() && MO.getReg().isPhysical())
```
andcarminati: Why checking if it is physical?

llvm/lib/Target/AIE/AIEPostPipeliner.cpp
```diff
@@ -1760,9 +1776,141 @@ void PostPipeliner::updateTripCount() const {
...
+        RegUse = MO.getReg();
+    }
+    if (TargetsBlock && RegUse.isValid())
```
andcarminati: I would restructure this function in such a way we stop the search if we hit this if. Or simply a return here. Alternatively, we are looking to the second last instruction (before the barrier), why do we need a search? Also, If we keep the search for simplicity, why not a reverse traversal of the MBB?

llvm/lib/Target/AIE/AIEPostPipeliner.cpp
```diff
@@ -1760,9 +1776,141 @@ void PostPipeliner::updateTripCount() const {
...
+  if (!Visited.insert(&BB).second)
+    return nullptr;
+  for (MachineBasicBlock *Pred : BB.predecessors())
+    if (MachineInstr *Found =
+            findReachingDef(*Pred, Pred->rbegin(), Reg, TRI, Visited))
+      return Found;
+  return nullptr;
+}
+
+MachineBasicBlock *PostPipeliner::findVersionGuard() const {
```
andcarminati: nit: AIELoopUtils with a nice comment?

llvm/lib/Target/AIE/AIEPostPipeliner.cpp
andcarminati: Let's say that this is a simplified version for **when you know that you have a single definition** that dominates uses. General reaching defs for non-ssa code require a bit more expensive fixed-point dataflow analysis that could return you a set of definitions. Maybe findUniqueReachingDef?

andcarminati: The fact that we cannot hoist the intrinsic here (properties of the instruction), your search will stop in the same block of the instruction that uses the register discovered by `findGuardConditionReg`. Basically, def and use will be in the same block, could we aggressively simplify? I feel we have a complex code to deal with a simple problem.

llvm/test/CodeGen/AIE/aie2/inner-loop-versioning-swp.ll
```diff
@@ -0,0 +1,58 @@
+; This file is licensed under the Apache License v2.0 with LLVM Exceptions.
+; See https://llvm.org/LICENSE.txt for license information.
+; SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+;
+; (c) Copyright 2026 Advanced Micro Devices, Inc. or its affiliates
+;
+; End-to-end: a versioned loop that the post-pipeliner can schedule gets its
```
andcarminati, line 7: We may use two sets of folders: loop versioning exclusive folder (IR tests - we need to create) and end-to-end (this folder exists).

llvm/lib/Target/AIE/AIEInnerLoopVersioning.cpp
andcarminati: The comments only, to help to describe the pass pre and post conditions. Please no header files.

llvm/lib/Target/AIE/AIEInnerLoopVersioning.cpp
```diff
@@ -0,0 +1,367 @@
...
+  Instruction *OrigTerm = GuardBB->getTerminator();
+  Builder.SetInsertPoint(OrigTerm);
+  Builder.CreateCondBr(TakeLow, LowPH, HighLoop->getLoopPreheader());
+  OrigTerm->eraseFromParent();
+
+  DT.changeImmediateDominator(ExitBlock, GuardBB);
+  addExitPHIs(HighLoop, ExitBlock, VMap, DefsUsedOutside);
+  formDedicatedExitBlocks(HighLoop, &DT, &LI, nullptr, /*PreserveLCSSA=*/true);
+  formDedicatedExitBlocks(&L, &DT, &LI, nullptr, /*PreserveLCSSA=*/true);
+
```
andcarminati, line 392: Could we extract all the code below as a member function related to metadata handling, like `updateLoopsMetadata`? You can use more descriptive methods for the previous steps also. This is a suggestion to improve readability - if you agree, of course.

llvm/lib/Target/AIE/AIEInnerLoopVersioning.cpp
```diff
@@ -0,0 +1,367 @@
...
+  // Consume the request hint on both copies so a second run of the pass does
+  // not re-version either. Mark the high-trip-count copy as versioned (carrying
+  // the request's value) so the postpipeliner recognizes it as the pipelined
+  // copy; the low copy stays the verbatim un-pipelined fallback.
+  int64_t HintValue = *AIELoopUtils::getLoopHintInt(
```
andcarminati: const?

llvm/lib/Target/AIE/AIEInnerLoopVersioning.cpp
```diff
@@ -0,0 +1,367 @@
...
+  // TripCount = BackedgeTakenCount + 1.
+  const SCEV *TC = SE.getTripCountFromExitCount(BEC);
+  SCEVExpander Exp(SE, InsertPt->getDataLayout(), "lver.tc");
+  if (!Exp.isSafeToExpandAt(TC, InsertPt))
+    return nullptr;
+  return Exp.expandCodeFor(TC, TC->getType(), InsertPt);
+}
+
+bool AIELoopVersioner::isProfitable() const {
+  // The runtime guard compares the trip count in i32, and the pipelined
```
andcarminati, line 301: Could we have a test rejected by this function?

llvm/lib/Target/AIE/AIEInnerLoopVersioning.cpp
```diff
@@ -0,0 +1,367 @@
...
+  // rationale). Seed it with -1 (UINT32_MAX): besides failing safe, -1 fits the
+  // narrow scalar-move immediate the pseudo materializes into, unlike a literal
+  // INT32_MAX, so the fallback move is emitted intact.
+  IRBuilder<> Builder(GuardBB->getTerminator());
+  Triple TT(Header->getModule()->getTargetTriple());
+  Intrinsic::ID ThresholdIID = AIEIRUtils::getLoopVersionThresholdIntrinsic(TT);
+  Value *Threshold =
+      Builder.CreateIntrinsic(ThresholdIID, {}, {Builder.getInt32(-1)},
+                              /*FMFSource=*/nullptr, "lver.threshold");
+  if (TripCount->getType() != I32Ty)
```
andcarminati: We guarantee to not be greater than 32, so we may have 16bit or 8bit, right? Could we have a test?

llvm/lib/Target/AIE/AIEInnerLoopVersioning.cpp
```diff
@@ -0,0 +1,367 @@
...
+    TripCount = Builder.CreateZExtOrTrunc(TripCount, I32Ty, "lver.tc.i32");
+  Value *TakeLow = Builder.CreateICmpULT(TripCount, Threshold, "lver.low");
+  GuardBB->setName(Header->getName() + ".lver.guard");
+
+  SmallVector<Instruction *, 8> DefsUsedOutside = findDefsUsedOutsideOfLoop(&L);
```
andcarminati: Is it possible to move this a bit more close to the user?

llvm/lib/Target/AIE/AIEInnerLoopVersioning.cpp
```diff
@@ -0,0 +1,367 @@
...
+    ArrayRef<Instruction *> DefsUsedOutside) const {
+  PHINode *PN = nullptr;
+  // Give every value used after the loop a merge PHI fed by the original copy.
+  for (Instruction *Inst : DefsUsedOutside) {
+    for (auto I = ExitBlock->begin(); (PN = dyn_cast<PHINode>(I)); ++I)
+      if (PN->getIncomingValue(0) == Inst)
+        break;
+    if (PN)
+      continue;
+    PN = PHINode::Create(Inst->getType(), 2, Inst->getName() + ".lver");
```
andcarminati: We are enforcing LCSSA everywhere. Here we catch a possible LCSSA violation, how is this possible? If we don't have a phi node and we need to create a new one we are definitely out of LCSSA. I think we could assert the property and simplify the code.

andcarminati: I guess this works as well:
```
void AIELoopVersioner::addExitPHIs(
    Loop *ClonedLoop, BasicBlock *ExitBlock, ValueToValueMapTy &VMap,
    ArrayRef<Instruction *> DefsUsedOutside) const {
  for (PHINode &PN : ExitBlock->phis()) {
    assert(PN.getNumIncomingValues() == 1 &&
           PN.getIncomingBlock(0) == L.getExitingBlock() &&
           "expected single-edge LCSSA phi");
    Value *OrigValue = PN.getIncomingValue(0);
    Value *ClonedValue = VMap.lookup(OrigValue);
    PN.addIncoming(ClonedValue ? ClonedValue : OrigValue,
                   ClonedLoop->getExitingBlock());
  }
}
```

I can't find a counter example.

llvm/lib/Target/AIE/AIEInnerLoopVersioning.cpp
```diff
@@ -0,0 +1,367 @@
...
+  // Replace the guard's fall-through with the trip-count branch: below the
+  // threshold run the original (low) copy, otherwise the pipelined high copy.
+  Instruction *OrigTerm = GuardBB->getTerminator();
+  Builder.SetInsertPoint(OrigTerm);
+  Builder.CreateCondBr(TakeLow, LowPH, HighLoop->getLoopPreheader());
+  OrigTerm->eraseFromParent();
+
+  DT.changeImmediateDominator(ExitBlock, GuardBB);
+  addExitPHIs(HighLoop, ExitBlock, VMap, DefsUsedOutside);
+  formDedicatedExitBlocks(HighLoop, &DT, &LI, nullptr, /*PreserveLCSSA=*/true);
```
andcarminati, line 389: Also, I guess we don't need to call formDedicatedExitBlocks here, simplifyLoop called previously guarantees this. If you see a potential need, we could create a test.

llvm/lib/Target/AIE/Utils/AIEIRUtils.cpp
```diff
@@ -38,6 +46,35 @@ bool isHardwareLoopDecrement(const Instruction *I) {
...
+  MDNode *LoopID = L.getLoopID();
```
F-Stuckmann: nit: const

llvm/lib/Target/AIE/Utils/AIEIRUtils.cpp
```diff
@@ -38,6 +46,35 @@ bool isHardwareLoopDecrement(const Instruction *I) {
...
+  if (!LoopID)
+    return;
```
F-Stuckmann, line 62: nit: we could also early exit if Keys is empty

llvm/lib/Target/AIE/AIEInnerLoopVersioning.cpp
```diff
@@ -0,0 +1,367 @@
...
+  auto &LI = getAnalysis<LoopInfoWrapperPass>().getLoopInfo();
+  auto &DT = getAnalysis<DominatorTreeWrapperPass>().getDomTree();
+  auto &SE = getAnalysis<ScalarEvolutionWrapperPass>().getSE();
+
+  LLVM_DEBUG(dbgs() << "AIEInnerLoopVersioning: " << F.getName() << "\n");
+
+  // Collect enabled loops before mutating, so cloning does not disturb
+  // iteration.
+  SmallVector<Loop *, 4> Candidates;
+  for (Loop *L : LI.getLoopsInPreorder())
```
F-Stuckmann: is the order relevant?

llvm/lib/Target/AIE/AIEInnerLoopVersioning.cpp
```diff
@@ -0,0 +1,367 @@
...
+    return nullptr;
+  }
+  return ExitBlock;
+}
+
+bool AIELoopVersioner::tryVersionLoop() {
+  // Read-only profitability gate first, so an unprofitable loop is rejected
+  // without any IR mutation.
+  if (!isProfitable()) {
+    LLVM_DEBUG(dbgs() << "  Not profitable to version\n");
```
F-Stuckmann: nit: please mention the loop that is not profitable. 
We may also want to emit optimiazation remarks, about loopVersion success/skipping.

llvm/lib/Target/AIE/AIEInnerLoopVersioning.cpp
```diff
@@ -0,0 +1,367 @@
...
+  /// Materialize the loop trip count at InsertPt, or nullptr if it cannot be
+  /// computed / expanded there.
+  Value *expandTripCount(Instruction *InsertPt) const;
+  /// Merge loop-defined values used after the loop across the two copies.
+  void addExitPHIs(Loop *ClonedLoop, BasicBlock *ExitBlock,
+                   ValueToValueMapTy &VMap,
+                   ArrayRef<Instruction *> DefsUsedOutside) const;
+  /// Remove the versioning request hint so the postpipeliner leaves \p Loop
+  /// alone.
+  void stripVersioningHint(Loop &Loop) const;
```
F-Stuckmann: nit: Why do we strip loop version hints? The postpipeliner specifically only checks if we have versioned metadata and not the versioning request. I don't think removing the loop version request is necessary if we didn't loopversion.

llvm/lib/Target/AIE/AIEInnerLoopVersioning.cpp
```diff
@@ -0,0 +1,367 @@
...
+  // without any IR mutation.
+  if (!isProfitable()) {
+    LLVM_DEBUG(dbgs() << "  Not profitable to version\n");
+    stripVersioningHint(L);
+    return false;
+  }
+
+  BasicBlock *ExitBlock = canonicalizeAndCheckStructure();
+  if (!ExitBlock) {
+    // Versioning bailed but the request hint is still on the loop. Consume it
```
F-Stuckmann: nit: Instead of a comment, could you just print the debug message?

llvm/lib/Target/AIE/AIEPostPipeliner.cpp
mludevid: Agreed. I simplified things a lot!

llvm/lib/Target/AIE/AIEPostPipeliner.cpp
mludevid: Skipping, because I removed this function during the simplification of the "patching the loop guard up" logic.

llvm/lib/Target/AIE/AIEPostPipeliner.cpp
mludevid: Removed during the simplification.

llvm/lib/Target/AIE/AIEPostPipeliner.cpp
mludevid: Again, this was removed during the simplification.

llvm/lib/Target/AIE/AIEPostPipeliner.cpp
mludevid: Moved to utils and added the comment.

llvm/lib/Target/AIE/AIEInnerLoopVersioning.cpp
mludevid: We have negative tests now.

llvm/lib/Target/AIE/AIEInnerLoopVersioning.cpp
mludevid: Done.

llvm/lib/Target/AIE/AIEInnerLoopVersioning.cpp
mludevid: We do need this, because due to the duplication of the loop, they now share their exit block, so we need to create new dedicated exit blocks.

llvm/lib/Target/AIE/AIEInnerLoopVersioning.cpp
mludevid: Not really because we only care about inner loops.

llvm/lib/Target/AIE/AIEPostPipeliner.cpp
```diff
@@ -152,12 +156,20 @@ bool PostPipeliner::isPostPipelineCandidate(MachineBasicBlock &LoopBlock) {
   // count is checked before accepting the schedule.
   using namespace AIELoopUtils;
   auto ParsedMinTripCount = getMinTripCount(LoopBlock);
-  if (!ParsedMinTripCount) {
+  MinTripCount = ParsedMinTripCount.value_or(0);
+
+  // A versioned loop only runs when its trip count clears the runtime guard, so
+  // any minimum is acceptable here; the guard threshold is patched with the
```
F-Stuckmann: can we instead of having this wall of comment have a condition LLVM_DEBUG announcing that iter count is searched by loop versioning for MBB?

llvm/lib/Target/AIE/AIEPostPipeliner.cpp
```diff
@@ -1760,9 +1776,63 @@ void PostPipeliner::updateTripCount() const {
...
+// The threshold placeholder of \p Guard, or nullptr if the block does not hold
+// exactly one. See PseudoLoopVersionThreshold for why scanning the guard block
+// alone is sufficient.
+static MachineInstr *findVersionThreshold(MachineBasicBlock &Guard,
+                                          const AIEBaseInstrInfo &TII) {
+  MachineInstr *Threshold = nullptr;
+  for (MachineInstr &MI : Guard.instrs()) {
+    if (!TII.isLoopVersionThresholdDef(MI))
+      continue;
+    if (Threshold)
```
F-Stuckmann: nit: this would indicate a duplicated inserted Threshold, could you add debug print to notify the user that something unexpected happened?

llvm/lib/Target/AIE/AIEPostPipeliner.cpp
```diff
@@ -1760,9 +1776,63 @@ void PostPipeliner::updateTripCount() const {
...
+
+  MachineBasicBlock *Guard = AIELoopUtils::getVersionGuardBlock(*Preheader);
+  if (!Guard) {
+    BailWithMsg("guard block not found");
+    return;
+  }
+
+  MachineInstr *Threshold = findVersionThreshold(*Guard, *TII);
+  if (!Threshold) {
+    BailWithMsg("guard threshold pseudo not found");
```
F-Stuckmann, line 1839: If we have 2 thresholds in the guard MBB, how do we materialize the 2 or more thresholds Pseudo instructions?

llvm/lib/Target/AIE/AIEPostPipeliner.cpp
F-Stuckmann: we could also assert here that we never encounter more than a single Threshold in the Guard MBB?

llvm/lib/Target/AIE/AIEPostPipeliner.cpp
```diff
@@ -1760,9 +1776,63 @@ void PostPipeliner::updateTripCount() const {
...
+    return;
+
+  // If the guard is no longer recognizable (e.g. a later MIR pass reshaped the
+  // region), bail without patching: the IR pass seeds the placeholder with the
+  // maximal threshold, so an un-patched guard keeps the loop on the
+  // low-trip-count copy. Diagnosable via debug output, never fatal.
+  MachineOptimizationRemarkEmitter ORE(*Preheader->getParent(), nullptr);
+  // A fall-through preheader may be empty, so pick a DebugLoc defensively.
+  const DebugLoc GuardLoc =
+      Preheader->empty() ? DebugLoc() : Preheader->front().getDebugLoc();
```
F-Stuckmann: CHECK: what happens if the first instruction in the preheader does not have a debug location but some instructions later on? Why are we not searching for the closest debug location within Preheader?

llvm/lib/Target/AIE/AIEPostPipeliner.h
```diff
@@ -374,9 +379,11 @@ class PostPipeliner {
   // Modify the tripcount to run StageCount-1 less iterations.
   void updateTripCount() const;
 
-  void materializePipeline(PipelineScheduleVisitor &Visitor);
+  /// For a versioned loop, patch the guard's placeholder threshold pseudo with
```
F-Stuckmann, line 384: supernit: could you move this function below materializePipeline, so that the diff makes it clear that it replaces getFinalMinTripCount ?

llvm/lib/Target/AIE/AIEInnerLoopVersioning.cpp
```diff
@@ -0,0 +1,411 @@
...
+  // i32.
+  const SCEV *BEC = SE.getBackedgeTakenCount(&L);
+  const SCEV *TC = SE.getTripCountFromExitCount(BEC);
+  SCEVExpander Exp(SE, InsertPt->getDataLayout(), "lver.tc");
+  if (!Exp.isSafeToExpandAt(TC, InsertPt))
+    return nullptr;
+  return Exp.expandCodeFor(TC, TC->getType(), InsertPt);
+}
+
+bool AIELoopVersioner::isProfitable() const {
```
F-Stuckmann: nit: this isn't a profitablility check but rather a hasVersionableTripCount() check

llvm/lib/Target/AIE/AIEInnerLoopVersioning.cpp
```diff
@@ -0,0 +1,411 @@
...
+           PN.getIncomingBlock(0) == L.getExitingBlock() &&
+           "expected a single-edge LCSSA PHI");
+    Value *OrigValue = PN.getIncomingValue(0);
+    Value *ClonedValue = VMap.lookup(OrigValue);
+    PN.addIncoming(ClonedValue ? ClonedValue : OrigValue,
+                   ClonedLoop->getExitingBlock());
+  }
+}
+
+void AIELoopVersioner::updateLoopsMetadata(Loop &HighLoop) const {
```
F-Stuckmann: nit: const is lying here, we are changing L itself with the HighLoop. Could you also pass L so that it is clear which loop metadata change?

llvm/lib/Target/AIE/AIEInnerLoopVersioning.cpp
```diff
@@ -0,0 +1,411 @@
...
+  Triple TT(Header->getModule()->getTargetTriple());
+  Intrinsic::ID ThresholdIID = AIEIRUtils::getLoopVersionThresholdIntrinsic(TT);
+  Value *Threshold =
+      Builder.CreateIntrinsic(ThresholdIID, {}, {Builder.getInt32(-1)},
+                              /*FMFSource=*/nullptr, "lver.threshold");
+  // Exact in both directions: isProfitable() bounded the trip count to 32 bits.
+  if (TripCount->getType() != I32Ty)
+    TripCount = Builder.CreateZExtOrTrunc(TripCount, I32Ty, "lver.tc.i32");
+  Value *TakeLow = Builder.CreateICmpULT(TripCount, Threshold, "lver.low");
+  GuardBB->setName(Header->getName() + ".lver.guard");
```
F-Stuckmann: nit: for clarity could you also move the Intrinsic Insertion and GuardBB creation into a helper function, so that tryVersionLoop remains on the same Abstraction level?

F-Stuckmann:
We are missing unit-tests for conditional loops, e.g: 

```
define i32 @unguarded_exit_phi(ptr %a, i32 %n, i1 %enter) {
entry:
  br i1 %enter, label %loop, label %exit
loop:
  %i = phi i32 [ 0, %entry ], [ %i.next, %loop ]
  %x = load i32, ptr %a, align 4
  %i.next = add i32 %i, 1
  %c = icmp slt i32 %i.next, %n
  br i1 %c, label %loop, label %exit, !llvm.loop !0
exit:
  %result = phi i32 [ -1, %entry ], [ %x, %loop ]
  ret i32 %result
}
```

llvm/lib/Target/AIE/AIEPostPipeliner.cpp
mludevid: The loop versioning is designed in a way that multiple thresholds can't happen. In any case I think it's a good idea to print a warning, but instead of asserting I think we can fail safe by not patching up any threshold, which will always default to taking the verbatim version of the loop.

llvm/lib/Target/AIE/AIEPostPipeliner.cpp
mludevid: We can't ever have more than 1 threshold pseudo instruction in an MBB. If we did we fail safe and don't path up any of them.

llvm/lib/Target/AIE/AIEPostPipeliner.cpp
mludevid: I changed this to find the first available debug location.

mludevid:
> We are missing unit-tests for conditional loops, e.g:
> 
> ```
> define i32 @unguarded_exit_phi(ptr %a, i32 %n, i1 %enter) {
> entry:
>   br i1 %enter, label %loop, label %exit
> loop:
>   %i = phi i32 [ 0, %entry ], [ %i.next, %loop ]
>   %x = load i32, ptr %a, align 4
>   %i.next = add i32 %i, 1
>   %c = icmp slt i32 %i.next, %n
>   br i1 %c, label %loop, label %exit, !llvm.loop !0
> exit:
>   %result = phi i32 [ -1, %entry ], [ %x, %loop ]
>   ret i32 %result
> }
> ```

I've added this test. It works as expected, the loop versioning guard is inserted between entry and the two versions of the loop and the two dedicated exit blocks are placed between the loops and the exit block.

llvm/lib/Target/AIE/AIEPostPipeliner.cpp
```diff
@@ -1777,16 +1776,21 @@ void PostPipeliner::updateTripCount() const {
...
+// sufficient.
 static MachineInstr *findVersionThreshold(MachineBasicBlock &Guard,
                                           const AIEBaseInstrInfo &TII) {
   MachineInstr *Threshold = nullptr;
   for (MachineInstr &MI : Guard.instrs()) {
     if (!TII.isLoopVersionThresholdDef(MI))
       continue;
-    if (Threshold)
+    if (Threshold) {
+      LLVM_DEBUG(dbgs() << "AIE loop versioning: multiple threshold pseudos in "
```
F-Stuckmann, line 1790: supernit: can this be an assert?

llvm/lib/Target/AIE/AIEInnerLoopVersioning.cpp
F-Stuckmann: nit: we could also directly return a true/false for the modification. We never evaluate Result::Canonicalized, so i think we can delete the Result enum

F-Stuckmann: we have dedicated exits, could you add them into the diagram?

llvm/lib/Target/AIE/AIEInnerLoopVersioning.cpp
```diff
@@ -359,7 +346,7 @@ AIELoopVersioner::Result AIELoopVersioner::tryVersionLoop() {
   formDedicatedExitBlocks(HighLoop, &DT, &LI, nullptr, /*PreserveLCSSA=*/true);
```
F-Stuckmann, line 389: nit: can we also use rename the exists to something more meaningful than  %exit.loopexit.loopexit, %exit.loopexit.loopexit1 , so incorporate the .lver.high and .lver.low in before the loop exit suffix?

llvm/lib/Target/AIE/Utils/AIELoopUtils.cpp
```diff
@@ -93,6 +101,13 @@ getDedicatedFallThroughPreheader(const MachineBasicBlock &LoopBlock) {
   return Candidate;
 }
 
+MachineBasicBlock *getVersionGuardBlock(const MachineBasicBlock &Preheader) {
+  if (Preheader.pred_size() != 1)
+    return nullptr;
+  MachineBasicBlock *Guard = *Preheader.pred_begin();
+  return Guard->succ_size() == 2 ? Guard : nullptr;
```
martien-de-jong, line 108: nit: this would be getGuardBlock. A VersionGuardBlock would imply having the version guard intrinsic.

llvm/lib/Target/AIE/AIEPostPipeliner.cpp
```diff
@@ -152,12 +156,20 @@ bool PostPipeliner::isPostPipelineCandidate(MachineBasicBlock &LoopBlock) {
   // count is checked before accepting the schedule.
   using namespace AIELoopUtils;
   auto ParsedMinTripCount = getMinTripCount(LoopBlock);
-  if (!ParsedMinTripCount) {
+  MinTripCount = ParsedMinTripCount.value_or(0);
+
+  // A versioned loop only runs when its trip count clears the runtime guard, so
+  // any minimum is acceptable here; the guard threshold is patched with the
+  // required minimum after scheduling (see updateVersionGuard).
+  IsVersionedGuarded = isLoopVersioned(LoopBlock);
```
martien-de-jong: IsVersionGuarded?

llvm/lib/Target/AIE/Utils/AIELoopUtils.cpp
martien-de-jong: Sorry. You already know it's version guarded from the loop annotation.  Still, this routine is more generic than finding a versioned GuardBlock.

llvm/lib/Target/AIE/AIEPostPipeliner.cpp
```diff
@@ -1760,6 +1776,65 @@ void PostPipeliner::updateTripCount() const {
   TII->adjustTripCount(*TripCountDef, -Delta);
 }
 
+// The threshold placeholder of \p Guard, or nullptr if the block does not hold
+// exactly one. See PseudoLoopVersionThreshold for why scanning the guard block
+// alone is sufficient.
+static MachineInstr *findVersionThreshold(MachineBasicBlock &Guard,
+                                          const AIEBaseInstrInfo &TII) {
+  MachineInstr *Threshold = nullptr;
+  for (MachineInstr &MI : Guard.instrs()) {
```
martien-de-jong: We could scan in reverse. it will be in the last few instructions and in most cases we will terminate early.

llvm/include/llvm/IR/IntrinsicsAIE2.td
andcarminati, line 970: Please, define once this as a class and reuse, no need to duplicate definitions.

llvm/lib/Target/AIE/AIEPostPipeliner.cpp
andcarminati: Yes please, if we reach this we probably versioned twice.

llvm/lib/Target/AIE/AIEPostPipeliner.cpp
andcarminati: nit: ThresholdMI

andcarminati: nit: GuardMBB.

andcarminati: nit: ThresholdMI.

andcarminati: nit: GuardMBB.

Please. propagate meaningful names, it helps to follow the code.

andcarminati:
Please, include some e2e tests as a pre commit.

llvm/lib/Target/AIE/AIEInnerLoopVersioning.cpp
andcarminati, line 92: Please, command line option to be able to test this before patching the code (minitercount 1)? Also, I guess we need to prevent double runs from this perspective, detection of (check already inserted instrinsics).

andcarminati: nit: Versioner?

andcarminati: nit: IsCanonicalized

andcarminati, line 186: Again, a command line is option, provided that you check that the loop was not versioned before.

andcarminati: Maybe you already disabled versioning with metadata of the non-pipelineable version.

andcarminati, line 470: please, disable versioning in the low case.

andcarminati: This function could be inlined in the previous one, I see no readability benefit, as you already call dropLoopMetadata in the caller side of this function.

andcarminati: nit: HeaderBB.

llvm/lib/Target/AIE/Utils/AIEIRUtils.cpp
```diff
@@ -38,6 +46,35 @@ bool isHardwareLoopDecrement(const Instruction *I) {
   return getIntrinsicID(I) == Intrinsic::loop_decrement;
 }
 
+Intrinsic::ID getLoopVersionThresholdIntrinsic(const Triple &TT) {
+  if (TT.isAIE2())
+    return Intrinsic::aie2_loop_version_threshold;
+  if (TT.isAIE2P())
+    return Intrinsic::aie2p_loop_version_threshold;
+  if (TT.isAIE2PS())
+    return Intrinsic::aie2ps_loop_version_threshold;
```
martien-de-jong, line 55: is there a way to share this intrinsic across subtargets? I guess these are explicitly pushed into a namespace somehow.

llvm/lib/Target/AIE/AIEPostPipeliner.cpp
mludevid: Done.

llvm/lib/Target/AIE/AIEInnerLoopVersioning.cpp
mludevid: Done.

mludevid: Done.

llvm/lib/Target/AIE/AIEInnerLoopVersioning.cpp
mludevid: Done.

llvm/lib/Target/AIE/Utils/AIELoopUtils.cpp
mludevid: Done.

llvm/lib/Target/AIE/AIEPostPipeliner.cpp
mludevid: Done.

llvm/lib/Target/AIE/AIEPostPipeliner.cpp
mludevid: Another comment asked for an assert that we don't have multiple thresholds. That requires scanning the entire block so the scan order become irrelevant. Skipping.

llvm/include/llvm/IR/IntrinsicsAIE2.td
mludevid: Done.

llvm/lib/Target/AIE/AIEPostPipeliner.cpp
mludevid: Done.

mludevid: Done.

mludevid: Done.

mludevid: Done.

llvm/lib/Target/AIE/AIEInnerLoopVersioning.cpp
mludevid: I've added two: one to disable the pass entirely and one to force loop versioning for loops with a min iteration count smaller than a user defined threshold, e.g. setting the limit to 1 will automatically loop version any loop that we mark with min iteration count of 1.

mludevid: Done.

mludevid: Done.

llvm/lib/Target/AIE/Utils/AIEIRUtils.cpp
mludevid: I went with andreu's proposal to define one intrinsic class defintion and reuse it in every target. Sharing the actual intrinsic is not easily possible due to different namespaces.

llvm/lib/Target/AIE/AIEInnerLoopVersioning.cpp
mludevid: Done.

mludevid: Done. I've added a new loop marker to identify these cases. The only way to trigger this is with the new cli option, without it we can never run into this case because we strip the loop versioning hints.

mludevid: Done.

F-Stuckmann (approved):
LGTM

andcarminati (approved):
LGTM. Thanks for addressing the suggestions.

PR #1084, opened 2026-07-13
[AIE] AA: trace pointers to noalias function arg roots

mludevid:


llvm/lib/Target/AIE/AIEBaseAliasAnalysis.cpp
```diff
@@ -604,6 +616,16 @@ AliasResult AIEBaseAAResult::alias(const MemoryLocation &LocA,
                         /*No virtually unrolled*/) == AliasResult::NoAlias)
     return AliasResult::NoAlias;
 
+  // Distinct noalias function-arg roots. Complements getUnderlyingObjectAIE
+  // (GEP/cast/AIE intrinsics): also walks phi/select and LoadInst for the
+  // io_buffer pattern (data pointer loaded from inside a noalias arg).
+  if (DisambiguateNoAliasArgRoots) {
+    const Argument *RootA = traceToNoAliasArgRoot(LocA.Ptr);
```
andcarminati, line 649: What if we have one pointer with restrict and other without? One restrict is enough if I understand the C standard correctly.

llvm/lib/Target/AIE/AIEBaseAliasAnalysis.cpp
```diff
@@ -653,6 +701,82 @@ static std::optional<int64_t> getGEPConstantOffset(const Value *V) {
   return cast<ConstantInt>(GEP->getOperand(1))->getSExtValue();
 }
 
+// Trace V to a noalias function argument through GEP/cast/phi/select/load.
+// Load look-through assumes the AIE io_buffer model: pointers loaded from a
+// noalias arg's storage point to buffers disjoint from other noalias args.
+static const Argument *
```
andcarminati, line 707: What about optional pair of argument and noalias flag? This could help with the case with a single restrict generating a single noalias.

llvm/lib/Target/AIE/AIEBaseAliasAnalysis.cpp
```diff
@@ -711,9 +835,21 @@ AliasResult AIE::aliasAcrossVirtualUnrolls(const MachineInstr *MIA,
       if (aliasAIEIntrinsic(ValueBasePairA->first, ValueBasePairB->first,
                             ValueBasePairA->second, ValueBasePairB->second,
                             UnrollLevelMIA,
-                            UnrollLevelMIB) == AliasResult::MayAlias) {
-        return AliasResult::MayAlias;
+                            UnrollLevelMIB) == AliasResult::NoAlias)
+        continue;
+
+      // Fall back to a noalias-arg-root analysis. This handles the common
```
andcarminati, line 841: if you are interested in a innermost loop pointer disambiguation, you can simply use the loop overloads that disable AA in the post machine scheduler.

mludevid: Thanks for bringing up this pointer disambiguation option, but we do need this for outer-loops as well.

llvm/lib/Target/AIE/AIEBaseAliasAnalysis.cpp
mludevid: I think this would be a nice follow-up indeed!

andcarminati (approved):
LGTM. We should solve the undef gate that was disturbing you in this PR.

PR #1091, opened 2026-07-14
[AIEX][NFC] Relocate outer-loop-pipeliner tests under outer-loop-pipeliner

F-Stuckmann:
Since these tests are end-to-end, move them to outer loop-pipeliner


andcarminati (changes requested):
Hi @F-Stuckmann. E2E are intended to ll -> assembly output. If we just run a pass and catch the output it is not E2E.

F-Stuckmann:
Right, I moved them to a new directory outer-loop-pipeliner

andcarminati (approved):

PR #1098, opened 2026-07-14
[AIEX][NFC] Outer Loop Pipeliner Refactor

F-Stuckmann:
This is the big NFC refactor PR. 
I split the refactor into logical components. 

Note the first commit may conflict, so i will rebase on aie-public once all of my interim PRs are merged.


llvm/lib/Target/AIE/AIEOuterLoopPipeliner.cpp
```diff
@@ -1707,16 +1946,16 @@ void AIEOuterLoopPipeliner::convertOuterLoopToHardwareLoop(
...
+  // critical: peelLastIteration clones all instructions currently in the outer
+  // latch.  By erasing OldCounter and OldIV here first, the peel step never
+  // sees them and never creates live last-iteration clones that would keep
+  // OldCounter alive.
   if (OldCounter) {
     // If OldIV is a pure counting PHI (no uses outside OldCounter), break
     // the cycle so both the PHI and the add can be deleted.
     if (OldIV->hasOneUse()) {
-      int LatchIdx = OldIV->getBasicBlockIndex(LS.OuterLatch);
+      int LatchIdx = OldIV->getBasicBlockIndex(SteadyLS.getOuterLatch());
```
andcarminati: I was expecting a code reduce with this refactor ;-).

llvm/lib/Target/AIE/AIEOuterLoopPipeliner.cpp
```diff
@@ -1458,137 +1728,112 @@ void AIEOuterLoopPipeliner::collectPartTwoInstructions(
...
+  if (SplitMode && collectPeeledForSplit(OrigLS)) {
+    LLVM_DEBUG(dbgs() << "    Split-prologue: pipelining "
+                      << OrigLS.peeledInsts().size()
+                      << " peeled instructions\n");
+    // The kept set forward-tracks from the peeled set, so collect it while both
+    // still live on the original nest.
+    collectKeptInstructions(OrigLS);
   } else {
-    collectPrologueInstructions(LS, PInsts);
+    if (SplitMode)
```
andcarminati: We are calling for a warning here outside debug mode. You can simplify this code even more.

llvm/lib/Target/AIE/AIEOuterLoopPipeliner.cpp
F-Stuckmann: there is still a lot i don't like, for example the region handling, the JNZD loop conversion just to name the most obvious deletion options

llvm/lib/Target/AIE/AIEOuterLoopPipeliner.cpp
F-Stuckmann: already deleted later on

llvm/lib/Target/AIE/AIEOuterLoopPipeliner.cpp
```diff
@@ -925,189 +1204,175 @@ void AIEOuterLoopPipeliner::createPipelinedPHIs(
...
 }
 
-// Validates the latch exit condition and caches all discovered components in
-// LS.LatchBound. Returns true if the loop bound can be adjusted (Step != 0).
-// All subsequent uses of the latch icmp pattern (adjustLoopBound,
-// getDowncountingInfo) read from LS.LatchBound without re-scanning.
-std::optional<LatchBoundInfo>
-AIEOuterLoopPipeliner::canAdjustLoopBound(const LoopStructure &LS) const {
-  auto *BI = dyn_cast<BranchInst>(LS.OuterLatch->getTerminator());
+bool LoopStructure::tryAdjustLoopBound() {
```
andcarminati: What happened with the old name? The new name is completely misleading, this is a profitability check, we are not trying to adjust anything here.

PR #1104, opened 2026-07-15
[AIEX] add Bundle counts to outerloop pipeliner unit-tests

F-Stuckmann:
…rtranslator MIR checks

Switch the six remaining structural outer-loop-pipeliner tests from -stop-after=aie-outer-loop-pipeliner with hand-written IR CHECKs to -stop-after=irtranslator with checks generated by
update_mir_test_checks.py, matching the convention already used by exit-no-phi/latch-accumulate/shared-exit. This lets these tests be regenerated with the repository's auto-update scripts going forward.


andcarminati (changes requested):
Hi, IR tests should stay as IR, no matter the work that may need to update them. End-to-end is ok.

llvm/test/CodeGen/AIE/aie2p/outer-loop-pipeliner/outer-loop-pipelining-gep.ll
```diff
@@ -1,77 +1,222 @@
...
 ; This file is licensed under the Apache License v2.0 with LLVM Exceptions.
 ; See https://llvm.org/LICENSE.txt for license information.
 ; SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
 ;
 ; (c) Copyright 2026 Advanced Micro Devices, Inc. or its affiliates
 ;
 ; RUN: llc -mtriple=aie2p -O2 -aie-enable-outer-loop-pipelining \
-; RUN:     -stop-after=aie-outer-loop-pipeliner \
-; RUN:     -o - %s 2>&1 | FileCheck %s
+; RUN:     -stop-after=irtranslator \
```
andcarminati: This is a pass test. It should test the pass output given an input.

F-Stuckmann:
we see literally see the same thing after IRtranslator as in IR, 
except that we need a huge machinery now to enable auto updating.

khallouh:
Could we take the opportunity to fix the tests to run only the outer loop pipeliner? Those tests are running many passes leading up to the OLP, which makes it hard to isolate the changes done by the OLP. Add on that the fact that we are missing check lines since we only use CHECK instead of CHECK-NEXT. So they may well partially catch regressions but they don't really show the full transformation performed by the OLP. These are the 2 things that are bothering me about those tests.

F-Stuckmann:
not having auto updatability will bite us really hard, we already have 3+ cleanups/features in the pipeline and we should not  update checklines by hand, that is a waste of our lifetimes.

andcarminati (approved):
LGTM. Thanks for addressing the comments.

PR #1108, opened 2026-07-15
[AIEX] update_ore_test_checks: restore remark RUN lines in place

F-Stuckmann:
when chaining Optimization Remark updates with llc/mir updates, runlines would move. 

This PR holds the runlines steady.


khallouh:
Could we have a test on AIE, not only x86?

F-Stuckmann:
added aie unit-tests

llvm/test/tools/UpdateTestChecks/update_ore_test_checks/aie-ore-remark-first.test
```diff
@@ -0,0 +1,17 @@
...
+# REQUIRES: aie-registered-target
+## Regression test: when a remark RUN line precedes another RUN line, the
+## chained --update-llc-checks must keep it in place rather than moving it
+## past the following RUN line.
+
+# RUN: cp -f %S/Inputs/aie-ore-remark-first.ll %t.ll && %update_ore_test_checks --update-llc-checks %t.ll
+# RUN: diff -u %S/Inputs/aie-ore-remark-first.ll.expected %t.ll
+
+## Verify idempotency via UTC_ARGS (no --update-llc-checks on the command line):
+# RUN: %update_ore_test_checks %t.ll
```
khallouh, line 16: Could you corrupt the remark lines before this second pass to verify that they are actually updated to match the initial ones?

llvm/test/tools/UpdateTestChecks/update_ore_test_checks/Inputs/aie-ore-remark-first.ll.expected
```diff
@@ -0,0 +1,67 @@
+; NOTE: Assertions have been autogenerated by utils/update_ore_test_checks.py UTC_ARGS: --update-llc-checks
```
khallouh, line 1: I guess these mixed llc and remark tests should only be updated with update_ore_test_checks.py because update_llc_checks.py would strip the remark lines, right?

F-Stuckmann: I don't think we strip unmatched checklines, otherwise you could never have multiple check-prefixes.
You cannot have two update scripts running on the same file, so if you want ORE you have to go through the ore script.

llvm/test/tools/UpdateTestChecks/update_ore_test_checks/aie-ore-remark-first.test
F-Stuckmann: added the two step commit pattern to clarify the code change

F-Stuckmann: the checks already exist in https://github.com/Xilinx/llvm-aie/blob/aie-public/llvm/test/tools/UpdateTestChecks/update_ore_test_checks/x86-ore-basic-ll.test

khallouh: Yes, but they don't really test idempotency either because the check lines are there. Since we tried this locally and it works, fine with me.

khallouh (approved):
LGTM. Thanks for fixing this.

PR #1112, opened 2026-07-15
[AIEX][NFC] OuterLoopPipeliner Refactor

F-Stuckmann:
Pure NFC OLP refactor, which included 3 todos with dump behaviour to remain NFC. 

These todos as well as the bugs will be removed in the next PR.

### DIFF
--- a/llvm/lib/Target/AIE/AIEOuterLoopPipeliner.cpp
+++ b/llvm/lib/Target/AIE/AIEOuterLoopPipeliner.cpp
@@ -291,19 +291,19 @@ private:
                            const ValueToValueMapTy &EpiVMap);
 
   // Create the cool-down region (peeled epilogue for last iteration):
-  //   cooldown.entry: set.loop.iterations (cloned) + Part2Insts clones
-  //   inner loop clone: uses last epilogue load values + Part2 results
+  //   cooldown.entry: set.loop.iterations (cloned) + PartTwoInsts clones
+  //   inner loop clone: uses last epilogue load values + PartTwo results
   //   cooldown.exit: epilogue stores only (no loads, no prologue clones)
   // Redirects the outer latch's false branch to cooldown.entry.
   // OrigEpiInsts: the set of instructions that were in the epilogue block
   // BEFORE clonePrologueIntoEpilogue inserted the prologue load clones.
-  // Part2Insts: 2048-bit producing intrinsics and their descendants that must
+  // PartTwoInsts: 2048-bit producing intrinsics and their descendants that must
   // be cloned into cooldown.entry so the cloned inner loop has correct initial
   // accumulator values. Empty when not in split-prologue mode.
   void
   peelLastIterationEpilogue(const LoopStructure &LS,
                             const SmallPtrSetImpl<Instruction *> &OrigEpiInsts,
-                            const SmallVectorImpl<Instruction *> &Part2Insts);
+                            const SmallVectorImpl<Instruction *> &PartTwoInsts);
 
   // Validates the latch exit condition, computes the induction step, and
   // caches all discovered components in LS.LatchBound for later use by
@@ -363,9 +363,10 @@ private:
   // plus all their forward-reachable descendants within the prologue.
   // These instructions stay in outer.header and are also cloned into
   // cooldown.entry so the cloned inner loop has correct initial values.
-  void collectPart2Instructions(const LoopStructure &LS,
-                                const SmallPtrSetImpl<Instruction *> &Part1Set,
-                                SmallVectorImpl<Instruction *> &Out) const;
+  void
+  collectPartTwoInstructions(const LoopStructure &LS,
+                             const SmallPtrSetImpl<Instruction *> &PartOneSet,
+                             SmallVectorImpl<Instruction *> &Out) const;
 };
 
 } // end anonymous namespace
@@ -774,7 +775,7 @@ void AIEOuterLoopPipeliner::createPipelinedPHIs(
 //   etc.
 void AIEOuterLoopPipeliner::peelLastIterationEpilogue(
     const LoopStructure &LS, const SmallPtrSetImpl<Instruction *> &OrigEpiInsts,
-    const SmallVectorImpl<Instruction *> &Part2Insts) {
+    const SmallVectorImpl<Instruction *> &PartTwoInsts) {
   Function *F = LS.OuterHeader->getParent();
   LLVMContext &Ctx = F->getContext();
 
@@ -854,21 +855,21 @@ void AIEOuterLoopPipeliner::peelLastIterationEpilogue(
   // outer.header and must also appear in cooldown.entry so the cloned inner
   // loop has correct initial accumulator values.
   // IMPORTANT: This must happen BEFORE cloning inner loop instructions and
-  // remapping, so that CoolVMap[Part2_inst] is set when the cloned inner loop
+  // remapping, so that CoolVMap[PartTwo_inst] is set when the cloned inner loop
   // PHIs (which reference Part 2 results from outer.header) are remapped.
   // CoolVMap already maps Part 1 PHI nodes → their latch values (= epilogue
   // clones), so remapping Part 2 clones automatically uses the last epilogue
   // values for Part 1 operands.
-  SmallVector<Instruction *, 16> Part2Clones;
-  for (Instruction *I : Part2Insts) {
+  SmallVector<Instruction *, 16> PartTwoClones;
+  for (Instruction *I : PartTwoInsts) {
     Instruction *Clone = I->clone();
     if (!Clone->getType()->isVoidTy())
       Clone->setName(I->getName() + ".cd");
     Clone->insertInto(CoolEntry, CoolEntry->end());
     CoolVMap[I] = Clone;
-    Part2Clones.push_back(Clone);
+    PartTwoClones.push_back(Clone);
   }
-  for (Instruction *Clone : Part2Clones)
+  for (Instruction *Clone : PartTwoClones)
     RemapInstruction(Clone, CoolVMap,
                      RF_NoModuleLevelChanges | RF_IgnoreMissingLocals);
 
@@ -1133,8 +1134,8 @@ bool AIEOuterLoopPipeliner::collectPrologueInstructionsForSplit(
     return false;
 
   // Find all descendants of anchors within the prologue (Part 2 set).
-  SmallPtrSet<Instruction *, 32> Part2Set;
-  Part2Set.insert(Anchors.begin(), Anchors.end());
+  SmallPtrSet<Instruction *, 32> PartTwoSet;
+  PartTwoSet.insert(Anchors.begin(), Anchors.end());
   SmallVector<Instruction *, 16> DescWorklist(Anchors.begin(), Anchors.end());
   while (!DescWorklist.empty()) {
     Instruction *I = DescWorklist.pop_back_val();
@@ -1142,7 +1143,7 @@ bool AIEOuterLoopPipeliner::collectPrologueInstructionsForSplit(
       auto *UI = dyn_cast<Instruction>(U);
       if (!UI || isa<PHINode>(UI) || !LS.isInPrologue(UI))
         continue;
-      if (Part2Set.insert(UI).second)
+      if (PartTwoSet.insert(UI).second)
         DescWorklist.push_back(UI);
     }
   }
@@ -1155,7 +1156,7 @@ bool AIEOuterLoopPipeliner::collectPrologueInstructionsForSplit(
   for (Instruction &I : *LS.OuterHeader) {
     if (isa<PHINode>(&I) || I.isTerminator())
       continue;
-    if (Part2Set.count(&I))
+    if (PartTwoSet.count(&I))
       continue;
     // Exclude set.loop.iterations / start.loop.iterations — they stay in
     // outer.header and are cloned separately into cooldown.entry.
@@ -1295,23 +1296,23 @@ bool AIEOuterLoopPipeliner::liftEpiloguePointerUpdatesToPrologue(
   return true;
 }
 
-void AIEOuterLoopPipeliner::collectPart2Instructions(
-    const LoopStructure &LS, const SmallPtrSetImpl<Instruction *> &Part1Set,
+void AIEOuterLoopPipeliner::collectPartTwoInstructions(
+    const LoopStructure &LS, const SmallPtrSetImpl<Instruction *> &PartOneSet,
     SmallVectorImpl<Instruction *> &Out) const {
   // Find anchors: instructions matching any split strategy that are direct
   // users of Part 1 instructions (or transitively reachable from Part 1 within
   // the prologue).
-  SmallPtrSet<Instruction *, 32> Part2Set;
+  SmallPtrSet<Instruction *, 32> PartTwoSet;
   SmallVector<Instruction *, 16> Worklist;
 
   // Seed: forward-track from Part 1 instructions to find anchor instructions.
-  for (Instruction *P1 : Part1Set) {
+  for (Instruction *P1 : PartOneSet) {
     for (User *U : P1->users()) {
       auto *UI = dyn_cast<Instruction>(U);
       if (!UI || isa<PHINode>(UI) || !LS.isInPrologue(UI))
         continue;
-      if (!Part1Set.count(UI) && isAnchorInstruction(UI)) {
-        if (Part2Set.insert(UI).second)
+      if (!PartOneSet.count(UI) && isAnchorInstruction(UI)) {
+        if (PartTwoSet.insert(UI).second)
           Worklist.push_back(UI);
       }
     }
@@ -1324,14 +1325,14 @@ void AIEOuterLoopPipeliner::collectPart2Instructions(
       auto *UI = dyn_cast<Instruction>(U);
       if (!UI || isa<PHINode>(UI) || !LS.isInPrologue(UI))
         continue;
-      if (!Part1Set.count(UI) && Part2Set.insert(UI).second)
+      if (!PartOneSet.count(UI) && PartTwoSet.insert(UI).second)
         Worklist.push_back(UI);
     }
   }
 
-  // Emit Part2Set in program order.
+  // Emit PartTwoSet in program order.
   for (Instruction &I : *LS.OuterHeader)
-    if (Part2Set.count(&I))
+    if (PartTwoSet.count(&I))
       Out.push_back(&I);
 
   LLVM_DEBUG(dbgs() << "    Split-prologue: " << Out.size()
@@ -1394,11 +1395,11 @@ bool AIEOuterLoopPipeliner::performTransformation(
   // Part 2 = matched producing intrinsics + descendants that stay in
   // outer.header and must also be cloned into cooldown.entry.
   // Must be done while Part 1 instructions still exist for forward-tracking.
-  SmallVector<Instruction *, 16> Part2Insts;
+  SmallVector<Instruction *, 16> PartTwoInsts;
   if (Overrides.get(SplitPrologue) && !PInsts.empty()) {
-    SmallPtrSet<Instruction *, 32> Part1Set;
-    Part1Set.insert(PInsts.begin(), PInsts.end());
-    collectPart2Instructions(LS, Part1Set, Part2Insts);
+    SmallPtrSet<Instruction *, 32> PartOneSet;
+    PartOneSet.insert(PInsts.begin(), PInsts.end());
+    collectPartTwoInstructions(LS, PartOneSet, PartTwoInsts);
   }
 
   // Create pipelined PHI nodes in the outer header for each data-load
@@ -1429,11 +1430,12 @@ bool AIEOuterLoopPipeliner::performTransformation(
       convertOuterLoopToHardwareLoop(LS, WarmUp, AdjustedTripCount, *Info);
 
   // Create the cool-down region (peeled last iteration):
-  //   set.loop.iterations + Part2 clones + inner loop clone + epilogue stores.
+  //   set.loop.iterations + PartTwo clones + inner loop clone + epilogue
+  //   stores.
   // OrigEpiInsts ensures only original epilogue instructions are copied into
   // cooldown.exit (not the prologue clones inserted earlier).
   // Redirects outer latch false branch to cooldown.entry.
-  peelLastIterationEpilogue(LS, OrigEpiInsts, Part2Insts);
+  peelLastIterationEpilogue(LS, OrigEpiInsts, PartTwoInsts);
 
   // Adjust itercount metadata to reflect the reduced trip count.
   updateLoopMetadata(LS);
--- a/llvm/lib/Target/AIE/AIEOuterLoopPipeliner.cpp
+++ b/llvm/lib/Target/AIE/AIEOuterLoopPipeliner.cpp
@@ -347,12 +347,10 @@ class OrigLoopStructure : public LoopStructure {
   // Returns the bottom-block stores.
   SmallVector<StoreInst *, 16> collectBottomStores() const;
 
-  // Worklist closure over pipelineable neighbours (forward via users, backward
-  // via operands), each added to Set once; Set must already hold the seeds.
-  void forwardClosure(SmallVectorImpl<Instruction *> &Worklist,
-                      SmallPtrSetImpl<Instruction *> &Set) const;
-  void backwardClosure(SmallVectorImpl<Instruction *> &Worklist,
-                       SmallPtrSetImpl<Instruction *> &Set) const;
+  // Returns the pipelineable closure of Seeds: users if TraverseUsers,
+  // otherwise operands.
+  SmallPtrSet<Instruction *, 32> collectClosure(ArrayRef<Instruction *> Seeds,
+                                                bool TraverseUsers) const;
 
   // The stage-1 set: the split points (IsSplitPoint, reachable from the
   // top-block loads) and their top-block descendants. Empty if none is found.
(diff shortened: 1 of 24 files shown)


llvm/lib/Target/AIE/AIEOuterLoopPipeliner.cpp
```diff
@@ -291,19 +291,19 @@ class AIEOuterLoopPipeliner : public FunctionPass {
                            const ValueToValueMapTy &EpiVMap);
 
   // Create the cool-down region (peeled epilogue for last iteration):
-  //   cooldown.entry: set.loop.iterations (cloned) + Part2Insts clones
-  //   inner loop clone: uses last epilogue load values + Part2 results
+  //   cooldown.entry: set.loop.iterations (cloned) + PartTwoInsts clones
+  //   inner loop clone: uses last epilogue load values + PartTwo results
```
khallouh: You also rename other things without reserving a separate commit to each renaming action, I would not have this Part1/Part2 rename as a separate commit then.

F-Stuckmann: they were easily dividable and separating them as a separate commit helps readability. 
I find the monster commits more problematic where i do a lot of stuff within a commit

llvm/lib/Target/AIE/AIEOuterLoopPipeliner.cpp
```diff
@@ -275,17 +419,46 @@ struct LoopStructure {
...
+    }
+    setOuterPreheader(Peel);
   }
 
-  // Latch exit condition analysis; populated by analyzeLoopStructure().
-  // Always valid: LoopStructure is only returned after Bound is set.
-  LatchBoundInfo Bound;
+  /// \return true if latch exit condition are valid and the loop bound can be
+  /// adjusted (Step != 0).
+  bool tryAdjustLoopBound();
```
andcarminati: Oh no, our old tryAdjustLoopBound.

llvm/lib/Target/AIE/AIEOuterLoopPipeliner.cpp
```diff
@@ -275,17 +419,46 @@ struct LoopStructure {
...
+      if (Succ != getOuterHeader())
+        return Succ;
+    llvm_unreachable("Outer latch must have an exit successor");
+  }
+
+  // Splice a freshly created peel block in as this nest's preheader: repoint
+  // the header PHIs' incoming edge from the current preheader to Peel, then
+  // record Peel. The current preheader is read before it is overwritten, so the
+  // two steps must stay in this order.
+  void adoptPeelAsPreheader(BasicBlock *Peel) {
```
andcarminati: nit: `PeeledBlock` ?

llvm/lib/Target/AIE/AIEOuterLoopPipeliner.cpp
```diff
@@ -76,296 +107,486 @@ static cl::opt<unsigned> OuterLoopPipeliningMinTripCount(
...
+class LoopStructure {
+protected:
+  BasicBlock *InnerExit = nullptr;
+
+  SmallVector<BasicBlock *, 4> InnerLoopBlocks;
+
+  MDNode *OuterLoopID = nullptr;
+
+  // The top region in program order; its entry is the top block.
+  BlockRegion TopRegion;
```
andcarminati, line 240: Could it be PologueRegion and EpilogueRegion?

llvm/lib/Target/AIE/AIEOuterLoopPipeliner.cpp
```diff
@@ -76,296 +107,486 @@ static cl::opt<unsigned> OuterLoopPipeliningMinTripCount(
...
+// The original loop, built from LoopInfo. Owns the LoopInfo loops and the
+// stage-0 / stage-1 instruction lists. Only obtainable validated, via
+// analyze().
+class OrigLoopStructure : public LoopStructure {
+  Loop *OuterLoop = nullptr;
+  Loop *InnerLoop = nullptr;
+
+  // The stage-0 / stage-1 split of the top-block instructions, in program
+  // order.
+  SmallVector<Instruction *, 16> Stage0Insts;
```
andcarminati: You can increase this. Kernels in IR are huuuuuuge.

llvm/lib/Target/AIE/AIEOuterLoopPipeliner.cpp
```diff
@@ -76,296 +107,486 @@ static cl::opt<unsigned> OuterLoopPipeliningMinTripCount(
...
+  // A pipelineable value also clonable into the prefetch/last-iteration sites;
+  // excludes terminators and hardware-loop setup.
+  bool isPipelineCandidate(const Instruction *I) const;
+
+  // Returns true if the inner loop is a hardware (JNZD) loop, i.e. its latch is
+  // controlled by an @llvm.loop.decrement intrinsic.
+  bool isInnerLoopHardwareLoop() const;
+
+  // Returns the top-block loads.
+  SmallVector<LoadInst *, 8> collectTopLoads() const;
```
andcarminati: You can increase this as well.

llvm/lib/Target/AIE/AIEOuterLoopPipeliner.cpp
```diff
@@ -76,296 +107,486 @@ static cl::opt<unsigned> OuterLoopPipeliningMinTripCount(
...
+  // Each source value mapped to its clone in this LS.
+  RemapTable CloneMap;
+
+public:
+  // Deep-clone Src into "<name>.<Suffix>" IR; internal references are remapped,
+  // edges leaving the LS stay at Src's externals for the caller to rewire.
+  CloneLoopStructure(const LoopStructure &Src, const Twine &Suffix);
+
+  // Tag for the last-iteration skeleton constructor below.
+  struct LastIterSkeletonTag {};
```
andcarminati: What is this?

llvm/lib/Target/AIE/AIEOuterLoopPipeliner.cpp
```diff
@@ -408,1187 +641,1266 @@ bool AIEOuterLoopPipeliner::runOnLoop(Loop *L) {
...
+  const bool TripCountTooLow = !MinTC || *MinTC < (int64_t)MinTripCount;
+  if (TripCountTooLow) {
     LLVM_DEBUG(dbgs() << "    Trip count too low\n");
     return false;
   }
 
-  SmallVector<LoadInst *, 8> PrologueLoads;
-  collectPrologueLoads(LS, PrologueLoads);
+  SmallVector<StoreInst *, 8> BottomStores = collectBottomStores();
+  // TODO: Confirm whether a store-free bottom can ever be profitable.
```
andcarminati, line 798: Not really necessary. I was targeting an specific case and  now we know that this could be general.

F-Stuckmann: I agree, but i am just NFC refactoring here.

llvm/lib/Target/AIE/AIEOuterLoopPipeliner.cpp
F-Stuckmann: If we have more complex OLP cases, it is difficult to call the TopRegion a PrologueRegion, because it may also contain control flow.

llvm/lib/Target/AIE/AIEOuterLoopPipeliner.cpp
F-Stuckmann: it gets reverted again

llvm/lib/Target/AIE/AIEOuterLoopPipeliner.cpp
F-Stuckmann: either way it got renamed to Stage0, which more closely describes the concept we are implementing

llvm/lib/Target/AIE/AIEOuterLoopPipeliner.cpp
F-Stuckmann: a dumb way of using explicit constructors. I deleted it.

llvm/lib/Target/AIE/AIEOuterLoopPipeliner.cpp
```diff
@@ -408,1187 +635,1264 @@ bool AIEOuterLoopPipeliner::runOnLoop(Loop *L) {
...
-  }
+bool OrigLoopStructure::canAdjustLoopBound() {
+  auto *BI = dyn_cast<BranchInst>(getBottom()->getTerminator());
+  if (!BI || !BI->isConditional())
+    return false;
+  auto *Cmp = dyn_cast<ICmpInst>(BI->getCondition());
+  if (!Cmp)
+    return false;
+
+  std::optional<LimitCounterSplit> Split =
```
andcarminati: nit: single use structs can be replaced by tuples.

llvm/lib/Target/AIE/AIEOuterLoopPipeliner.cpp
```diff
@@ -408,1187 +635,1264 @@ bool AIEOuterLoopPipeliner::runOnLoop(Loop *L) {
...
+static PHINode *findCountingPhi(BinaryOperator *Add, const BasicBlock *Header) {
+  for (Value *Op : Add->operands()) {
+    auto *PHI = dyn_cast<PHINode>(Op);
+    if (PHI && PHI->getParent() == Header)
+      return PHI;
+  }
+  return nullptr;
+}
+
+namespace {
```
andcarminati: With a tuple we can remove completely this namespace.

llvm/lib/Target/AIE/AIEOuterLoopPipeliner.cpp
```diff
@@ -408,1187 +635,1264 @@ bool AIEOuterLoopPipeliner::runOnLoop(Loop *L) {
...
+
+void CloneLoopStructure::installPreheader(BasicBlock *NewPreheader) {
+  BasicBlock *OldPreheader = getPreheader();
+  reroutePhiIncomings(getTop(), OldPreheader, NewPreheader, PhiEdge::Repoint);
+  OuterPreheader = NewPreheader;
+}
+
+// The `add Counter, C` (with constant C) that defines Counter, or nullptr.
+static BinaryOperator *getConstantStepAdd(Value *Counter) {
+  auto *Add = dyn_cast<BinaryOperator>(Counter);
```
andcarminati, line 1499: We could apply an early return here whenever possible.

llvm/lib/Target/AIE/AIEOuterLoopPipeliner.cpp
```diff
@@ -408,1187 +635,1264 @@ bool AIEOuterLoopPipeliner::runOnLoop(Loop *L) {
...
+    Loop *L, const AIE::LoopOptionOverrides &Overrides) {
+  if (!Overrides.get(EnableOuterLoopPipelining)) {
+    LLVM_DEBUG(dbgs() << "    Not pipelining: not enabled (flag/metadata)\n");
+    return false;
   }
-  if (!L->getLoopLatch()) {
-    LLVM_DEBUG(dbgs() << "    No single outer latch\n");
-    return std::nullopt;
+
+  std::unique_ptr<OrigLoopStructure> LS = OrigLoopStructure::analyze(L);
```
andcarminati: `OrigLoopStructure::analyze` is not doing something that helps to clean/understand the code. This function instantiates and checks, then return. Then we check here again. We can instantiate here and perform a single check.

llvm/lib/Target/AIE/AIEOuterLoopPipeliner.cpp
F-Stuckmann: nit: we should also fully delete this whole section in a followup PR

llvm/lib/Target/AIE/AIEOuterLoopPipeliner.cpp
andcarminati: See that we could use the same pattern as below for the check, i.e. `isProfitableToRotate`.

llvm/lib/Target/AIE/AIEOuterLoopPipeliner.cpp
```diff
@@ -8,51 +8,80 @@
...
+#include "Utils/AIEIRUtils.h"
 #include "Utils/AIELoopOptionOverrides.h"
 #include "Utils/AIELoopUtils.h"
+#include "llvm/ADT/ArrayRef.h"
+#include "llvm/ADT/STLExtras.h"
+#include "llvm/ADT/STLFunctionalExtras.h"
+#include "llvm/ADT/SmallPtrSet.h"
+#include "llvm/ADT/SmallVector.h"
+#include "llvm/ADT/Twine.h"
 #include "llvm/Analysis/AliasAnalysis.h"
```
andcarminati: I guess, we can remove alias support, as we are not using it anymore.

llvm/lib/Target/AIE/AIEOuterLoopPipeliner.cpp
```diff
@@ -8,51 +8,80 @@
...
 #include "llvm/Analysis/DependenceAnalysis.h"
 #include "llvm/Analysis/LoopInfo.h"
 #include "llvm/Analysis/ScalarEvolution.h"
+#include "llvm/CodeGen/TargetPassConfig.h"
 #include "llvm/IR/BasicBlock.h"
+#include "llvm/IR/CFG.h"
 #include "llvm/IR/Constants.h"
 #include "llvm/IR/Dominators.h"
```
andcarminati, line 73: We also can remove Dominators.

llvm/lib/Target/AIE/AIEOuterLoopPipeliner.cpp
```diff
@@ -408,1187 +635,1264 @@ bool AIEOuterLoopPipeliner::runOnLoop(Loop *L) {
...
+        Worklist.push_back(UI);
     }
-  for (StoreInst *S : EpilogueStores)
-    if (S->isVolatile() || S->isAtomic()) {
-      LLVM_DEBUG(dbgs() << "    Unsafe: volatile/atomic epilogue store\n");
-      return false;
+  }
+}
+
+void OrigLoopStructure::backwardClosure(
```
andcarminati: These two functions are almost the same, could we unify? User derives from Value.

llvm/lib/Target/AIE/AIEOuterLoopPipeliner.cpp
andcarminati: But it diverges from the following checks!

andcarminati:
Nice clean-up @F-Stuckmann. I left some comments that you could address, mostly to reduce code and improve a bit more clarity. Thanks!

llvm/lib/Target/AIE/AIEOuterLoopPipeliner.cpp
F-Stuckmann: This is a deliberate factory function, so that data ownership is crystal clear and we don't need awkward concepts to represent data ownership/validity and we delete walls of documentation.

F-Stuckmann: the other checks are profitability checks, that are independent of validity checks (which analyze does)

F-Stuckmann: I renamed it to clarify the factory method nature of the static method.

andcarminati (approved):
LGTM.

llvm/test/CodeGen/AIE/aie2p/outer-loop-pipeliner/outer-loop-pipelining-split-prologue.ll
```diff
@@ -175,11 +175,11 @@ define void @split_prologue_basic(ptr noalias %a_ptr_init,
 ; ASM-NEXT:    nopa ; nopb ; nops ; nopx ; vmov x0, bmll0; nopv
 ; ASM-NEXT:    nopa ; nopb ; nops ; nopx ; vextract.64 r5:r4, x0, #0, vaddsign1; nopv
 ; ASM-NEXT:    nopa ; nopb ; nops ; nopxm ; nopv
-; ASM-NEXT:  .LBB0_6: // %inner.header.cd
+; ASM-NEXT:  .LBB0_6: // %steady.inner.header.lastiter
```
khallouh, line 178: This is called steady and lastiter at the same time. I still find that confusing tbh.
I understand that this is cloned from the steady-state loop above, but we could always rename the blocks with more meaningful names.

llvm/test/CodeGen/AIE/aie2p/outer-loop-pipeliner/outer-loop-pipelining-split-prologue.ll
```diff
@@ -165,7 +165,7 @@ define void @split_prologue_basic(ptr noalias %a_ptr_init,
 ; ASM-NEXT:    nop // Delay Slot 3
 ; ASM-NEXT:    mov p4, p3 // Delay Slot 2
 ; ASM-NEXT:    padda [p4], m0; st r5, [p2, #0]; mov p2, p3 // Delay Slot 1
-; ASM-NEXT:  // %bb.5: // %cooldown.entry
+; ASM-NEXT:  // %bb.5: // %lastiter.prologue
```
khallouh, line 168: Nit: Why do we start with lastiter in this `lastiter.prologue` but `steady.inner.header.lastiter` below has lastiter at the end? This could be harmonized.

khallouh: What you call `lastiter.prologue` is just the entry block/ preheader of the inner loop which will be pipelined or not by the postscheduler, why is it called prologue? I would reserve the prologue/epilogue naming for the outerloop which is relevant for this pipeliner. We risk overloading these terms so much that it becomes difficult to follow what we are talking about.

llvm/lib/Target/AIE/AIEOuterLoopPipeliner.cpp
```diff
@@ -8,27 +8,53 @@
...
-// Outer loop pipelining for AIE: overlaps the data-load chain (prologue) of
-// outer iteration i+1 with the inner loop + store chain (epilogue) of
-// iteration i.
+// Outer loop pipelining for AIE: overlaps the peeled chain (prologue) of outer
+// iteration i+1 with the inner loop + store chain (epilogue) of iteration i.
+//
+// The prologue (the outer header) splits into peeled and kept instructions by
+// what the transform does with each:
+//   Peeled = the load/address chain: loads plus the pointer/address arithmetic
+//            feeding the inner loop. Peeled out of the steady header — copied
```
khallouh, line 17: Nit: It is not only the loads and pointer/address arithmetic that can be peeled. We do peel everything leading up to the anchors in split mode.

llvm/lib/Target/AIE/AIEOuterLoopPipeliner.cpp
```diff
@@ -8,27 +8,53 @@
...
+//       |                       |
+//   [steady.latch]  -----------/  <- stores + loads for NEXT iteration
+//       |  (exit branch)
+//   [lastiter.prologue]        <- set.loop.iterations (cloned)
 //       |
-//   [inner loop clone]             <- uses v0.epi/v1.epi from last epilogue
+//   [steady.inner.*.lastiter]  <- inner loop clone, uses last epilogue values
 //       |
-//   [cooldown.exit]                <- stores only, no loads
+//   [lastiter.epilogue]        <- stage 1
```
khallouh, line 57: Nit: stage 1 is not confined to this `lastiter.epilogue` It also technically includes the inner loop and Part2 instructions

llvm/lib/Target/AIE/AIEOuterLoopPipeliner.cpp
```diff
@@ -151,24 +176,50 @@ static bool isSafePointerIncrementIntrinsic(Intrinsic::ID IID) {
...
-// instructions without special-casing the block count.
+// exit last). Today the prologue and epilogue are each a single block; the
+// wrapper lets callers test membership and iterate instructions through one
+// interface that would extend to multi-block regions without changing them.
 class BlockRegion {
   SmallVector<BasicBlock *, 4> Blocks;
 
+  // No edge enters a non-entry block and none leaves a non-exit block: the run
+  // is single-entry/single-exit. Vacuously true for a single-block region.
+  [[maybe_unused]] bool isSingleEntrySingleExit() const {
```
khallouh, line 187: This assumes that Blocks.front() is the entry block and Blocks.back() is the exit block, it would be nice to assert this and mention it.

llvm/lib/Target/AIE/AIEOuterLoopPipeliner.cpp
```diff
@@ -193,45 +238,144 @@ struct LatchBoundInfo {
...
+  // The peeled / kept split of the prologue instructions, in program order.
+  // Populated on the original nest only; clones never store these (see
+  // IsOrigLS).
+  SmallVector<Instruction *, 16> PeeledInsts;
+  SmallVector<Instruction *, 16> KeptInsts;
+
+  // Snapshot of the epilogue (outer latch) contents taken before
+  // clonePrologueIntoEpilogue inserts prefetch clones, so the last-iteration
+  // epilogue can be filtered back to the original stores only.
+  SmallPtrSet<Instruction *, 32> EpilogueSnapshot;
```
khallouh, line 291: Why do we still need a snapshot now that we keep OrigLS intact?

llvm/lib/Target/AIE/AIEOuterLoopPipeliner.cpp
khallouh: You're renaming them to PeeledSet and KeptSet in the next commit anyway

llvm/lib/Target/AIE/AIEOuterLoopPipeliner.cpp
```diff
@@ -849,44 +889,36 @@ void AIEOuterLoopPipeliner::populateVMapFromPHIs(ValueToValueMapTy &VMap,
...
+  if (!isInnerLoopHardwareLoop()) {
     LLVM_DEBUG(dbgs() << "    Inner loop is not a hardware loop\n");
     return false;
   }
-  std::optional<int64_t> MinTC =
-      llvm::getMinTripCount(OrigLS.getOuterLoop(), SE);
-  if (!MinTC ||
-      *MinTC < (int64_t)Overrides.get(OuterLoopPipeliningMinTripCount)) {
+  std::optional<int64_t> MinTC = llvm::getMinTripCount(getOuterLoop(), &SE);
+  if (!MinTC || *MinTC < (int64_t)MinTripCount) {
```
khallouh, line 900: Nit: These are 2 Min Trip Count which mean different things I would call MinTripCount something like MinTripCountRequirement or MinProfitableMinTripCount

llvm/lib/Target/AIE/AIEOuterLoopPipeliner.cpp
```diff
@@ -1324,24 +1357,38 @@ void AIEOuterLoopPipeliner::cloneInnerLoopIntoLastIter(
...
-    const LoopStructure &LastIterLS, const LoopStructure &SteadyLS,
+    const LoopStructure &OrigLS, const LoopStructure &SteadyLS,
+    const LoopStructure &LastIterLS, const ValueToValueMapTy &SteadyVMap,
     ValueToValueMapTy &LastIterVMap) const {
-  const SmallPtrSetImpl<Instruction *> &EpiSnapshot =
-      SteadyLS.epilogueSnapshot();
-  SmallVector<Instruction *, 16> EpiInsts;
-  for (Instruction &I : *SteadyLS.getOuterLatch()) {
+  // Read the pristine original epilogue: OrigLS is never mutated by the
+  // prefetch-cloning steps, so its latch holds the full last-iteration work
```
khallouh, line 1364: Nice!

llvm/lib/Target/AIE/AIEOuterLoopPipeliner.cpp
```diff
@@ -8,79 +8,76 @@
...
-//       |                       |        set.loop.iterations
-//       |                       |
-//   [outer.inner.*]             |  <- inner loop
-//       |                       |
-//   [outer.latch]  ------------/  <- Epilogue
+//   [top]  <------------------\  <- stage 0 + stage 1 (+ set.loop.iterations)
+//       |                      |
+//   [inner.*]                  |  <- inner loop
+//       |                      |
+//   [bottom]  ----------------/   <- stage 1 + latch
```
khallouh, line 32: This is the original CFG. As I understand, your stage 0 and 1 notions refer to previously peeled and kept instructions from the outer header. So stage 1 can't be in bottom (epilogue) until we pipeline

llvm/test/CodeGen/AIE/aie2p/outer-loop-pipeliner/outer-loop-pipelining-split-prologue.ll
khallouh: Oh, you renamed them again in a later commit

llvm/test/CodeGen/AIE/aie2p/outer-loop-pipeliner/outer-loop-pipelining-split-prologue.ll
khallouh: Ok, you renamed again later

llvm/lib/Target/AIE/AIEOuterLoopPipeliner.cpp
khallouh: I see you removed the snapshot use in a later commit

llvm/lib/Target/AIE/AIEOuterLoopPipeliner.cpp
khallouh, line 1: The commit message does not mention a major part of the changes which is all of the scaffolding to reproduce the live-out bug. I understand that the refactor helped you find the bug since you tried to delete the original LS whereas we were transforming in place before. I see that you remap the live-outs to steady state instead of LastIter to reproduce the buggy behavior, which is what it would have taken to fix the bug independantly of the refactor.

khallouh: This commit has some crucial changes of the refactor and undoes much of the previous renamings which got their own commits. It would have been nice to split it or at least be more explicit about the changes in the commit message to help tracking and review. That been said, the code itself is quite clear and well-documented.

PR #1121, opened 2026-07-16
[AIEX] Use MemoryObjectsBits only for loads

andcarminati:
Use a similar approach as memory banks.


khallouh:
How about the hazards for which the stores were included in the first place? Have they become irrelevant? I'd be interested to see QoR impact across the board.

andcarminati:
> How about the hazards for which the stores were included in the first place? Have they become irrelevant? I'd be interested to see QoR impact across the board.

Good question. The original implementation was created as an internal request to solve a stall problem related to two loads in tightly coupled loop:
```
   ; CHECK-NEXT:    .p2align 4
   ; CHECK-NEXT:  .LBB0_1: // %for.body
   ; CHECK-NEXT:    // =>This Inner Loop Header: Depth=1
-  ; CHECK-NEXT:    vldb wl2, [p0], #64; vlda wh2, [p0, #32]; vst.srs.s16.s32 bmh0, s0, [p1, #32]; nopxm ; nopv
+  ; CHECK-NEXT:    vldb wh2, [p0, #32]; nopa ; vst.srs.s16.s32 bml0, s0, [p1], #64; nopxm ; vmul cm0, x2, x0, r0
   ; CHECK-NEXT:  .L_LEnd0:
-  ; CHECK-NEXT:    nopb ; nopa ; vst.srs.s16.s32 bml0, s0, [p1], #64; nopxm ; vmul cm0, x2, x0, r0
+  ; CHECK-NEXT:    vldb wl2, [p0], #64; nopa ; vst.srs.s16.s32 bmh0, s0, [p1, #32]; nopxm ; nopv
   ; CHECK-NEXT:  // %bb.2: // %for.cond.cleanup
+  ; CHECK-NEXT:    nopb ; nopa ; vst.srs.s16.s32 bml0, s0, [p1], #64; nopxm ; vmul cm0, x2, x0, r0
```

In this case, before, we had an II=2 pipeline, packing two loads on the same pointer in the same slot, leading to a slower code due to wait cycles (bank conflicts).  With this change, we have an II=2, but with the loads in different slot, leading to a faster code (no stalls). 

We basically never evaluated the store impact because we just implemented this as similar approach as memory bank bits. Recently we dropped store support for memory bank bits.  Also, loads and stores are not sharing memory ports and we cannot have two stores in a slot. Even though, I will provide QoR numbers accordingly ;-).

F-Stuckmann (dismissed):
Nice, I like that you added ORE remarks so that we don't have to count bundles anymore

F-Stuckmann:
Could you add QoR numbers?

khallouh:
> > How about the hazards for which the stores were included in the first place? Have they become irrelevant? I'd be interested to see QoR impact across the board.
> 
> Good question. The original implementation was created as an internal request to solve a stall problem related to two loads in tightly coupled loop:
> 
> ```
>    ; CHECK-NEXT:    .p2align 4
>    ; CHECK-NEXT:  .LBB0_1: // %for.body
>    ; CHECK-NEXT:    // =>This Inner Loop Header: Depth=1
> -  ; CHECK-NEXT:    vldb wl2, [p0], #64; vlda wh2, [p0, #32]; vst.srs.s16.s32 bmh0, s0, [p1, #32]; nopxm ; nopv
> +  ; CHECK-NEXT:    vldb wh2, [p0, #32]; nopa ; vst.srs.s16.s32 bml0, s0, [p1], #64; nopxm ; vmul cm0, x2, x0, r0
>    ; CHECK-NEXT:  .L_LEnd0:
> -  ; CHECK-NEXT:    nopb ; nopa ; vst.srs.s16.s32 bml0, s0, [p1], #64; nopxm ; vmul cm0, x2, x0, r0
> +  ; CHECK-NEXT:    vldb wl2, [p0], #64; nopa ; vst.srs.s16.s32 bmh0, s0, [p1, #32]; nopxm ; nopv
>    ; CHECK-NEXT:  // %bb.2: // %for.cond.cleanup
> +  ; CHECK-NEXT:    nopb ; nopa ; vst.srs.s16.s32 bml0, s0, [p1], #64; nopxm ; vmul cm0, x2, x0, r0
> ```
> 
> In this case, before, we had an II=2 pipeline, packing two loads on the same pointer in the same slot, leading to a slower code due to wait cycles (bank conflicts). With this change, we have an II=2, but with the loads in different slot, leading to a faster code (no stalls).
> 
> We basically never evaluated the store impact because we just implemented this as similar approach as memory bank bits. Recently we dropped store support for memory bank bits. Also, loads and stores are not sharing memory ports and we cannot have two stores in a slot. Even though, I will provide QoR numbers accordingly ;-).

Thanks for the clarification. I think we could have 2 stores in different bundles with different access cycles, which would hit the same memory bank in the same cycle and cause a conflict (stall). Was this case taken into account before the store case drop in this PR?

andcarminati:
>  I think we could have 2 stores in different bundles with different access cycles, which would hit the same memory bank in the same cycle and cause a conflict (stall). Was this case taken into account before the store case drop in this PR?

This case could theoretically exist. Although, in common cases that I usually see in pipeline kernels, we have similar store chains, so they will be apart from one cycle. I will check QoR to see if we need more intervention, because separating loads and stores memory bits would require a lot of contract changes, including post pipeliner.

andcarminati:
> > I think we could have 2 stores in different bundles with different access cycles, which would hit the same memory bank in the same cycle and cause a conflict (stall). Was this case taken into account before the store case drop in this PR?
> 
> This case could theoretically exist. Although, in common cases that I usually see in pipeline kernels, we have similar store chains, so they will be apart from one cycle. I will check QoR to see if we need more intervention, because separating loads and stores memory bits would require a lot of contract changes, including post pipeliner.

Wait, I will bring a compact solution for this! Thanks for pointing this out.

andcarminati:
> > > I think we could have 2 stores in different bundles with different access cycles, which would hit the same memory bank in the same cycle and cause a conflict (stall). Was this case taken into account before the store case drop in this PR?
> > 
> > 
> > This case could theoretically exist. Although, in common cases that I usually see in pipeline kernels, we have similar store chains, so they will be apart from one cycle. I will check QoR to see if we need more intervention, because separating loads and stores memory bits would require a lot of contract changes, including post pipeliner.
> 
> Wait, I will bring a compact solution for this! Thanks for pointing this out.

I found the missing piece. While trying to create a test case, I found that stores always use (2p and 2ps) `DM_ADS` port and this port conflict will always keep the stores in a distance that implicates no bank conflicts. For a bank conflict to occur, we should have a resource conflict altogether, what we already prevent. In this case, we don't need anything else in this PR.

khallouh:
> `DM_ADS`

Is this because the difference between the use cycle of that resource between any 2 store instructions is the same as the difference in their memory access cycles? Could we assert this somewhere in case this assumption changes?

andcarminati:
As discussed with @khallouh, now we have a complete separation between loads and stores. I can't craft a test for the current targets as port conflicts are dominating object bits considering store chains. However, this may be useful for the future.

khallouh:
> > `DM_ADS`
> 
> Is this because the difference between the use cycle of that resource between any 2 store instructions is the same as the difference in their memory access cycles? Could we assert this somewhere in case this assumption changes?

I actually checked this and it holds for most instructions except for ST_TM and and part word stores

llvm/test/CodeGen/AIE/aie2ps/schedule/vst-memory-cycles.mir
```diff
@@ -0,0 +1,89 @@
...
+#   This conflicts! Both access memory at cycle 5.
+#
+#   Solution: Insert NOP between them:
+#   - CONV VST (MemCycles=6) issued at cycle 0 -> memory access at cycle 5
+#   - NOP at cycle 1
+#   - Basic VST (MemCycles=5) issued at cycle 2 -> memory access at cycle 6
+#   No conflict.
+
+# RUN: llc -mtriple=aie2ps -run-pass=postmisched %s -o - \
+# RUN:   | FileCheck %s
```
khallouh: Nit: same regression test for aie2p?

llvm/test/CodeGen/AIE/aie2ps/schedule/vst-memory-cycles.mir
```diff
@@ -0,0 +1,89 @@
...
+  }
+...
+
+---
+# Test 1: Memory cycles conflict - CONV VST (MemCycles=6) followed by basic VST (MemCycles=5)
+# Both stores access the same memory object (%ir.ptr).
+# CONV VST issued at cycle 0 -> memory access at cycle 5 (0+6-1)
+# Basic VST issued at cycle 1 -> memory access at cycle 5 (1+5-1) - CONFLICT!
+# The scheduler must insert a NOP to shift the basic VST to cycle 2,
+# so its memory access occurs at cycle 6 instead of cycle 5.
```
khallouh: Can you mention that the NOP is inserted by the normal hazard on ADS resource instead of Object Bits?

llvm/test/CodeGen/AIE/aie2ps/schedule/vst-memory-cycles.mir
khallouh, line 1: Thanks for adding the same feature for stores and this store vs store test. Could you add (pre-commit) one for load vs store which is the point of this PR to remove the artificial hazard?

llvm/test/CodeGen/AIE/aie2ps/schedule/vst-memory-cycles.mir
```diff
@@ -0,0 +1,89 @@
...
+  bb.0.entry:
+    liveins: $bmll0, $wl0, $p0, $p1
+    ; CHECK-LABEL: name: no_conflict_different_objects
+    ; The scheduler may reorder - but no NOP needed between them
+    ; CHECK: VST_dmw_sts_w_st_idx_imm killed $wl0, killed $p1, 0
+    ; CHECK-NEXT: VST_CONV_bf16_fp32_dmw_sts_srs_bf16_idx_imm killed $bmll0, killed $p0, 0
+    ; CONV VST (MemCycles=6) to %ir.ptr1
+    VST_CONV_bf16_fp32_dmw_sts_srs_bf16_idx_imm $bmll0, killed $p0, 0, implicit-def $srf2fflags, implicit $crf2fmask, implicit $crfpconvsat, implicit $crrnd :: (store (<16 x s16>) into %ir.ptr1, addrspace 6)
+    ; Basic VST (MemCycles=5) to different %ir.ptr2 - no conflict
+    VST_dmw_sts_w_st_idx_imm $wl0, killed $p1, 0 :: (store (<8 x s32>) into %ir.ptr2, addrspace 6)
```
khallouh: I just found that bank information is thrown away for stores after noticing that these 2 stores access the same bank and yet are marked "no conflict" which is counterintuitive.  Could we make the banks different in this test or add another one?

khallouh (commented):
LGTM. Left some comments mostly asking for clarification and more testing. And it would be nice to see some QoR numbers for focus kernels on the 2 targets.

khallouh:
> > > `DM_ADS`
> > 
> > 
> > Is this because the difference between the use cycle of that resource between any 2 store instructions is the same as the difference in their memory access cycles? Could we assert this somewhere in case this assumption changes?
> 
> I actually checked this and it holds for most instructions except for ST_TM and and part word stores

I tried to craft a test out of these exceptions, but we can't since TM accesses a different bank anyway and there is no conflict and partword stores are always kept away by 7 cycles due to the read-modify-write nature.

llvm/test/CodeGen/AIE/aie2ps/schedule/vst-memory-cycles.mir
andcarminati: This test actually slipped in, I will remove it because conflicts here are solved by resource conflicts.

llvm/test/CodeGen/AIE/aie2ps/schedule/vst-memory-cycles.mir
andcarminati: Test removed, not actually testing the feature in this PR.

andcarminati:
Hi @F-Stuckmann and @khallouh , new update: split loads and stores and included unit tests.

khallouh (changes requested):
I'm fine with a GTEST to replace MIR unit test, but I think there are some important gaps in the testing:
1. getMemoryObjectsBits() classification is never exercised. The three new tests build MemoryObjectPair by hand, so they only test the conflict() comparison, not the logic introduced by this PR. We don't test that a real load populates only Load, a real store only Store, this is the core behavior of this PR.
2. We don't test the "both bitmaps filled" case (e.g. partword-stores), to verify it conflicts with both a later load and a later store to the same object.
3. No test that the 2 bitmaps ncoexist independently within one cycle. I suggest that in a single scoreboard cycle, we emit a load to object A and a store to object B, then checks all four combinations independently:

    - load probe on A → conflict
    - store probe on B → conflict
    - load probe on B → no conflict
    - store probe on A → no conflict
    
This proves the Load and Store bitmaps are tracked and matched independently, which is the real invariant behind the change.

llvm/lib/Target/AIE/AIEHazardRecognizer.cpp
```diff
@@ -121,13 +125,17 @@ FuncUnitWrapper &FuncUnitWrapper::operator|=(const FuncUnitWrapper &Other) {
...
   return *this;
 }
 
 bool FuncUnitWrapper::conflict(const FuncUnitWrapper &Other) const {
+  // Load-Load and Store-Store conflicts are checked independently.
+  // Loads and stores don't conflict with each other on AIE (separate HW ports).
   if ((Slots & Other.Slots) != 0 || (MemoryBanks & Other.MemoryBanks) != 0 ||
-      (MemObjectsBits & Other.MemObjectsBits) != 0 ||
+      (LoadMemObjectsBits & Other.LoadMemObjectsBits) != 0 ||
+      (StoreMemObjectsBits & Other.StoreMemObjectsBits) != 0 ||
```
F-Stuckmann, line 138: StoreMemObjectBits cannot overlap because they can never be issued in the same bundle. 
Additionally, if we would have different access-latencies, so that 2 stores in different bundles could interfere, we already detect that in Required conflicts.

For partword stores, we already implement LoadMemObjectsBits for the word load and block Required resources from the load access cycle until the store is completed. Therefore a Load/store cannot overlap with partwordstores because of the blocked required resources alone

We can completely drop the Store MemObjectbits.

khallouh: I'm not sure I get what you mean by "we already detect that in Required conflicts", we should not make assumptions about which resources/ports are used based on current sched model of aie2ps, this is supposed to be target agnostic.
For partword-stores, that's my point, we should make sure we check conflicts with both loads and stores, which this PR does, only the test is missing.
So I disagree that we can drop the Store MemObjectbits. And for the same reasons, we shouldn't have dropped it for MemoryBanks either.

andcarminati:
QoR evaluation: -0.10% in core compute insn cycles (current focus - maybe not too representative).

```
| Core_Compute_Insn_Count | mllib_20072026-AIE2ps_peano_aie2ps | mllib_20072026-AIE2ps-memobj_peano_aie2ps | sim_success | Total diff |
| --- | --- | --- | --- | --- |
| MaxPool2D_bf16_2 |  |  | FAIL |  |
| Pad2D_bf16_0 | 16595 | 16593 | PASS | SAME(-0.01%) |
| AvgPool2D_bfloat16_3 | 8249 | 8246 | PASS | SAME(-0.04%) |
| AvgPool2D_bfloat16_0 | 7505 | 7502 | PASS | SAME(-0.04%) |
| AvgPool2D_bfloat16_1 | 7217 | 7214 | PASS | SAME(-0.04%) |
| Conv2D_DW_bf16_3 | 1979 | 1978 | PASS | SAME(-0.05%) |
| Conv2D_DW_bf16_0 | 1978 | 1977 | PASS | SAME(-0.05%) |
| AvgPool2D_bfloat16_2 | 5873 | 5870 | PASS | SAME(-0.05%) |
| AvgPool2dVariant_bf16_2 | 5531 | 5527 | PASS | SAME(-0.07%) |
| SoftMaxRowMajor_bf16_0 | 4495 | 4491 | PASS | SAME(-0.09%) |
| AvgPool2dVariant_bf16_0 | 4409 | 4405 | PASS | SAME(-0.09%) |
| GEMM_Bfp16_opt_1 | 3255 | 3252 | PASS | SAME(-0.09%) |
| AvgPool2dVariant_bf16_6 | 4333 | 4329 | PASS | SAME(-0.09%) |
| AvgPool2dVariant_bf16_5 | 3973 | 3969 | PASS | IMPR(-0.10%) |
| AvgPool2dVariant_bf16_1 | 3126 | 3122 | PASS | IMPR(-0.13%) |
| GEMM_Bfp16_opt_0 | 1994 | 1991 | PASS | IMPR(-0.15%) |
| AvgPool2dVariant_bf16_4 | 2516 | 2512 | PASS | IMPR(-0.16%) |
| Conv2D_DW_bf16_1 | 4534 | 4526 | PASS | IMPR(-0.18%) |
| AvgPool2dVariant_bf16_3 | 1979 | 1975 | PASS | IMPR(-0.20%) |
| Conv2D_DW_bf16_2 | 2533 | 2524 | PASS | IMPR(-0.36%) |
| Average diff | +0.00% | -0.10% |  | -0.10% |
| Diff stdev | 0.00 | 0.08 |  | 0.08 |
| Quantile #1 | +0.00% | -0.15% |  | -0.15% |
| Quantile #2 | +0.00% | -0.09% |  | -0.09% |
| Quantile #3 | +0.00% | -0.05% |  | -0.05% |

```

```
| Core_Compute_Cycle_Count | mllib_20072026-AIE2ps_peano_aie2ps | mllib_20072026-AIE2ps-memobj_peano_aie2ps | sim_success | Total diff |
| --- | --- | --- | --- | --- |
| MaxPool2D_bf16_2 |  |  | FAIL |  |
| Pad2D_bf16_0 | 46264 | 46262 | PASS | SAME(-0.00%) |
| AvgPool2D_bfloat16_3 | 21003 | 21000 | PASS | SAME(-0.01%) |
| AvgPool2D_bfloat16_0 | 20414 | 20411 | PASS | SAME(-0.01%) |
| AvgPool2D_bfloat16_1 | 19089 | 19086 | PASS | SAME(-0.02%) |
| Conv2D_DW_bf16_3 | 5715 | 5714 | PASS | SAME(-0.02%) |
| Conv2D_DW_bf16_0 | 6194 | 6193 | PASS | SAME(-0.02%) |
| AvgPool2D_bfloat16_2 | 15167 | 15164 | PASS | SAME(-0.02%) |
| AvgPool2dVariant_bf16_2 | 10513 | 10509 | PASS | SAME(-0.04%) |
| SoftMaxRowMajor_bf16_0 | 21244 | 21240 | PASS | SAME(-0.02%) |
| AvgPool2dVariant_bf16_0 | 9071 | 9067 | PASS | SAME(-0.04%) |
| GEMM_Bfp16_opt_1 | 21655 | 21652 | PASS | SAME(-0.01%) |
| AvgPool2dVariant_bf16_6 | 17891 | 17887 | PASS | SAME(-0.02%) |
| AvgPool2dVariant_bf16_5 | 14399 | 14395 | PASS | SAME(-0.03%) |
| AvgPool2dVariant_bf16_1 | 8264 | 8260 | PASS | SAME(-0.05%) |
| GEMM_Bfp16_opt_0 | 9917 | 9914 | PASS | SAME(-0.03%) |
| AvgPool2dVariant_bf16_4 | 9704 | 9700 | PASS | SAME(-0.04%) |
| Conv2D_DW_bf16_1 | 13450 | 13446 | PASS | SAME(-0.03%) |
| AvgPool2dVariant_bf16_3 | 6251 | 6247 | PASS | SAME(-0.06%) |
| Conv2D_DW_bf16_2 | 11065 | 11056 | PASS | SAME(-0.08%) |
| Average diff | +0.00% | -0.03% |  | -0.03% |
| Diff stdev | 0.00 | 0.02 |  | 0.02 |
| Quantile #1 | +0.00% | -0.04% |  | -0.04% |
| Quantile #2 | +0.00% | -0.02% |  | -0.02% |
| Quantile #3 | +0.00% | -0.02% |  | -0.02% |

```

llvm/lib/Target/AIE/AIEHazardRecognizer.cpp
F-Stuckmann: We are duplicating logic here. 
I am not making any assumptions here i am reading in our Instrinfo.td files, how we treat partwordstores.
Every single architecture we support blocks everything from the load to the store cycle for partwordstores, so no instruction that also accesses the  PART_WORD_STORE FuncUnit resource may interfere in the load/store window.
(Fun fact: We may still store/load to tile memory)

This reserved/required blocking is a necessary correctness implementation, because we cannot trust AA in the case of partwordstores.

If you have encountered an architecture where we violated that contract, we have to get it fixed ASAP. 
MemObj or MemBanks cannot and will not work correct with PartWordStores, because they can only block the immediate access cycles, but partwordstores have to block all the cycles between load and store!

andcarminati:
> I'm fine with a GTEST to replace MIR unit test, but I think there are some important gaps in the testing:
> 
> 1. getMemoryObjectsBits() classification is never exercised. The three new tests build MemoryObjectPair by hand, so they only test the conflict() comparison, not the logic introduced by this PR. We don't test that a real load populates only Load, a real store only Store, this is the core behavior of this PR.
> 2. We don't test the "both bitmaps filled" case (e.g. partword-stores), to verify it conflicts with both a later load and a later store to the same object.
> 3. No test that the 2 bitmaps ncoexist independently within one cycle. I suggest that in a single scoreboard cycle, we emit a load to object A and a store to object B, then checks all four combinations independently:
>    
>    * load probe on A → conflict
>    * store probe on B → conflict
>    * load probe on B → no conflict
>    * store probe on A → no conflict
> 
> This proves the Load and Store bitmaps are tracked and matched independently, which is the real invariant behind the change.

Included in a separated commit, also with mock testing.

llvm/lib/Target/AIE/AIEHazardRecognizer.cpp
khallouh: Sure, partword stores are modeled conservatively and for the rest we don't conflict in aie2ps due to that DM_ADS resource which happens to be used at the same intervals as the MemoryCycles between all instructions pairs. We need to make sure that any 2 stores don't conflict, not only partword stores. That's why I asked the question and Andreu split into 2 separate bitmaps to solve this and decouple the load and store. 
I think the logic is not duplicate, because:

- If we drop it and merge both bitmaps (pre PR state) we will try to avoid virtual bank conflicts between load and store and hurt InstCount with no clear benefit
- If we reserve the bitmap to load/load conflicts and ignore the stores as in MemoryBankBits, we might overlook a conflict between 2 stores in an architecture where this DM_ADS does not come to the rescue in the right cycle intervals.

llvm/unittests/Target/AIE/HazardRecognizerTest.cpp
```diff
@@ -843,3 +863,284 @@ TEST(HazardRecognizer, loadStoreObjectNoConflict) {
...
+  // Store instruction should populate the Store bitmap.
+  EXPECT_NE(Result.Store, 0u) << "Store instruction should set Store bits";
+
+  // For a pure store (!mayLoad), Load should be zero.
+  // For a RMW store (mayLoad), Load may be non-zero.
+  if (!StoreMI->mayLoad()) {
+    EXPECT_EQ(Result.Load, 0u)
+        << "Pure store instruction should NOT set Load bits";
+  }
+}
```
khallouh: This is dead code since the MF above has no pure stores

khallouh: Maybe you can add a normal store instruction to MF and do something like this:
```
TEST_F(GetMemoryObjectsBitsTest, PartWordStorePopulatesBothLoadAndStoreBits) {
  MachineFunction *MF = getMachineFunction(M.get(), "test_load_store");
  const MachineBasicBlock &MBB = MF->front();

  // The part-word store is the only op that both loads and stores (RMW).
  const MachineInstr *RMWStoreMI = nullptr;
  for (const MachineInstr &MI : MBB) {
    if (MI.mayLoad() && MI.mayStore()) {
      RMWStoreMI = &MI;
      break;
    }
  }
  ASSERT_NE(RMWStoreMI, nullptr) << "Could not find part-word RMW store";

  const auto *TII =
      static_cast<const AIEBaseInstrInfo *>(MF->getSubtarget().getInstrInfo());
  AIEAlternateDescriptors AltDescs;
  AIEHazardRecognizer HR(TII, MF->getSubtarget().getInstrItineraryData(),
                         AltDescs, /*IsPreRA=*/false);

  MemoryObjectPair Result = HR.getMemoryObjectsBits(RMWStoreMI);

  // A RMW store must populate BOTH bitmaps with the same object.
  EXPECT_NE(Result.Load, 0u) << "RMW store should set Load bits";
  EXPECT_NE(Result.Store, 0u) << "RMW store should set Store bits";
  EXPECT_EQ(Result.Load, Result.Store)
      << "RMW store should use the same object in both bitmaps";
}
```
```
TEST_F(GetMemoryObjectsBitsTest, PureStorePopulatesOnlyStoreBits) {
  MachineFunction *MF = getMachineFunction(M.get(), "test_load_store");
  const MachineBasicBlock &MBB = MF->front();

  // The full-word store is a pure store (stores but does not load).
  const MachineInstr *PureStoreMI = nullptr;
  for (const MachineInstr &MI : MBB) {
    if (MI.mayStore() && !MI.mayLoad()) {
      PureStoreMI = &MI;
      break;
    }
  }
  ASSERT_NE(PureStoreMI, nullptr) << "Could not find pure store instruction";

  const auto *TII =
      static_cast<const AIEBaseInstrInfo *>(MF->getSubtarget().getInstrInfo());
  AIEAlternateDescriptors AltDescs;
  AIEHazardRecognizer HR(TII, MF->getSubtarget().getInstrItineraryData(),
                         AltDescs, /*IsPreRA=*/false);

  MemoryObjectPair Result = HR.getMemoryObjectsBits(PureStoreMI);

  // Pure store should populate only the Store bitmap.
  EXPECT_NE(Result.Store, 0u) << "Pure store should set Store bits";
  EXPECT_EQ(Result.Load, 0u) << "Pure store should NOT set Load bits";
}

```

llvm/unittests/Target/AIE/HazardRecognizerTest.cpp
```diff
@@ -681,3 +757,390 @@ TEST(HazardRecognizer, storeStoreDifferentMemoryCycleSerializes) {
...
+
+/// Test that load and store to different objects get different object numbers.
+TEST_F(GetMemoryObjectsBitsTest, LoadAndStoreGetDifferentObjectNumbers) {
```
khallouh: Nit: This does not test what its name claims. It only checks that LoadResult.Load != 0 and StoreResult.Store != 0; it never compares the bit positions. %a and %b could be assigned the same object number and the test would still pass. To match the name, we would gave to assert the two bitmaps occupy different bits (e.g. Load bit index ≠ Store bit index).

khallouh (approved):
LGTM

PR #1123, opened 2026-07-16
[AIE] Pointer-coupled locks with alias-aware scheduling delays

katerynamuts:
Adds pointer-coupled lock intrinsics for AIE2P/AIE2PS and makes lock stall/resume
spacing alias-aware when the lock is tied to a specific buffer.

- **Intrinsics & Clang**: New `*_ptr` lock builtins carry a buffer pointer through
  to LLVM intrinsics (replacing ignored pointer arguments in the lock headers).
- **GISel / MIR**: Lower `*_ptr` locks to ordinary ACQ/REL while preserving the
  buffer as an **annotative** memoperand for alias tagging (not a real DM access).
- **Scheduling**: Use alias analysis in `LockDelays` (same-block) and
  `AIERegMemEventTracker` / `EmitFixedSUnits` (pipelined prologue/epilogue) to
  skip stall/resume delays when the lock buffer is disjoint from nearby memory ops.
  Id-only locks remain conservative.

## Test plan

- [x] `aie2p-lock-ptr-alias.ll` / `aie2ps-lock-ptr-alias.ll` — LockDelays,
      same-block acquire/release vs load/store (same vs disjoint buffer)
- [x] `pipelined-lock-ptr-alias.mir` (aie2p + aie2ps) — RegMemEventTracker via
      pipelined epilogue REL_ptr and prologue ACQ_ptr cases
- [x] Pre-commit test expectations document conservative (pre-fix) scheduling;
      follow-up commit updates disjoint-case CHECKs after alias-aware behavior
      
##  QoR results
no new failures     
      
Outlier: ThresholdedRelu

Metric                    private   locks-ptr   Δ%
──────────────────────────────────────────────────
Core_Compute_Insn_Count       783       1,564   +99.7%
Core_Cycle_Count           17,182      12,820   −25.4%

Core_PMSize_absolute
Core_StackSize_absolute


clang/lib/Headers/aie2p/aie2p_locks.h
```diff
@@ -35,25 +35,29 @@ release(unsigned id, int val, int cond) {
...
+}
 INTRINSIC(void)
 acquire_equal(const void *a, unsigned id, unsigned val, int cond) {
-  return acquire_equal(id, val, cond);
+  return __builtin_aie2p_acquire_cond_ptr(a, id, val, cond);
 }
 INTRINSIC(void)
 acquire_greater_equal(const void *a, unsigned id, unsigned val) {
-  return acquire_greater_equal(id, val);
+  return __builtin_aie2p_acquire_ptr(a, id, -val);
```
khallouh, line 48: why do we change the sign of val, passing -val instead?

F-Stuckmann:
Quick Question: why do we need to update AIERegMemEventTracker if we also update the DAG construction? 
Wouldn't the latencies edges provide us enough distance for scheduling?

llvm/test/CodeGen/AIE/aie2p/GlobalIsel/inst-select-locks.mir
```diff
@@ -185,3 +185,153 @@ body:             |
...
+    ; CHECK-NEXT: {{  $}}
+    ; CHECK-NEXT: [[COPY:%[0-9]+]]:er = COPY $r0
+    ; CHECK-NEXT: [[COPY1:%[0-9]+]]:er = COPY $r1
+    ; CHECK-NEXT: ACQ_mLockId_reg [[COPY]], [[COPY1]] :: (load unknown-size from %stack.0, align 1)
+    ; CHECK-NEXT: PseudoRET implicit $lr
+    %ptr:ptrregbank(p0) = G_FRAME_INDEX %stack.0
+    %0:gprregbank(s32) = COPY $r0
+    %1:gprregbank(s32) = COPY $r1
+    G_INTRINSIC_W_SIDE_EFFECTS intrinsic(@llvm.aie2p.acquire.ptr), %ptr:ptrregbank(p0), %0:gprregbank(s32), %1:gprregbank(s32) :: (load unknown-size from %stack.0)
+    PseudoRET implicit $lr
```
khallouh, line 210: Can we have tests showing how the MMO being attached? This only has ISel test no IRtranslator test for the getTgtMemIntrinsic change

llvm/test/CodeGen/AIE/aie2p/aie2p-lock-ptr-alias.ll
```diff
@@ -34,10 +34,9 @@ define void @acq_ptr_before_store_same(ptr %p, i32 %id, i32 %val) {
...
 ; CHECK-LABEL: acq_ptr_before_store_disjoint:
 ; CHECK:       // %bb.0:
-; CHECK-NEXT:    nopa ; nopb ; nops ; acq r0, r1; nopm ; nopv
-; CHECK-NEXT:    ret lr
-; CHECK-NEXT:    nop // Delay Slot 5
-; CHECK-NEXT:    nop // Delay Slot 4
+; CHECK-NEXT:    nopa ; nopb ; nops ; ret lr; nopm ; nopv
+; CHECK-NEXT:    nopx // Delay Slot 5
+; CHECK-NEXT:    acq r0, r1 // Delay Slot 4
 ; CHECK-NEXT:    nop // Delay Slot 3
```
khallouh: Nice test update

llvm/test/CodeGen/AIE/aie2p/schedule/postpipeliner/pipelined-lock-ptr-alias.mir
```diff
@@ -0,0 +1,319 @@
...
+
+--- |
+  target triple = "aie2p"
+
+  define void @epilogue_release_ptr_same_buffer() {
+  entry:
+    %store_buf = alloca i32, align 4
+    %lock_buf = alloca i32, align 4
+    unreachable
+  }
```
khallouh, line 28: Do you need this IR function definition in this MIR test?

khallouh: Same question for other 3 copies below which only differ in name

llvm/test/CodeGen/AIE/aie2ps/schedule/postpipeliner/pipelined-lock-ptr-alias.mir
```diff
@@ -0,0 +1,315 @@
...
+    liveins: $p0, $p1, $r0, $m0, $x0
+
+    $x0 = VADD_16 $x0, $x0, implicit $crbf8conf, implicit $crfp8conf
+    $p1 = VST_dmx_sts_x_st_pstm_nrm_imm $x0, $p1, 64 :: (store (s32) into %ir.store_buf)
+    PseudoLoopEnd <mcsymbol .L_LEnd0>, %bb.1
+
+  bb.2:
+    liveins: $p0, $p1, $r18, $x0
+    REL_mLockId_imm 49, renamable $r18 :: (load unknown-size from %ir.store_buf, align 1)
+    $p1 = VST_dmx_sts_x_st_pstm_nrm_imm $x0, $p1, 64 :: (store (s32) into %ir.store_buf)
```
khallouh, line 140: The REL tests have an interacting VST in the same block and in a different block (loop and epilogue). Can we have the same for ACQ since the tests there only exercise interblock?

llvm/test/CodeGen/AIE/aie2ps/schedule/postpipeliner/pipelined-lock-ptr-alias.mir
```diff
@@ -0,0 +1,315 @@
...
+    successors: %bb.1(0x7c000000), %bb.2(0x04000000)
+    liveins: $p0, $p1, $r0, $m0, $x0
+
+    $x0 = VADD_16 $x0, $x0, implicit $crbf8conf, implicit $crfp8conf
+    $p1 = VST_dmx_sts_x_st_pstm_nrm_imm $x0, $p1, 64 :: (store (s32) into %ir.store_buf)
+    PseudoLoopEnd <mcsymbol .L_LEnd0>, %bb.1
+
+  bb.2:
+    liveins: $p0, $p1, $r18, $x0
+    REL_mLockId_imm 49, renamable $r18 :: (load unknown-size from %ir.lock_buf, align 1)
```
khallouh, line 208: What happens if we have release in prologue and acquire in epilogue? Could we test this?

llvm/test/CodeGen/AIE/aie2ps/aie2ps-lock-ptr-alias.ll
khallouh, line 1: Nit: It would be nice to have an end-to-end test like this with a pipelined loop since these are all single block and you are touching interblock scheduling. For example, I would extract something from the benchmark driving this change.

llvm/test/CodeGen/AIE/aie2ps/aie2ps-lock-ptr-alias.ll
```diff
@@ -0,0 +1,261 @@
...
+; CHECK-NEXT:    st.s8 r2, [p0, #0]
+; CHECK-NEXT:    nop
+; CHECK-NEXT:    ret lr
+; CHECK-NEXT:    nop // Delay Slot 5
+; CHECK-NEXT:    nop // Delay Slot 4
+; CHECK-NEXT:    nop // Delay Slot 3
+; CHECK-NEXT:    nop // Delay Slot 2
+; CHECK-NEXT:    nop // Delay Slot 1
+  call void @llvm.aie2ps.acquire.ptr(ptr %p, i32 %id, i32 %val)
+  store volatile i8 %bval, ptr %p
```
khallouh: Q: I see these are all volatile in this file. Was it strictly needed?

llvm/lib/Target/AIE/AIEBaseSubtarget.cpp
khallouh, line 1: Did you think about how this optimization would affect the latencies to streaming instruction with 0 and 1 streams, as these also take a pointer and are mayLoad and mayStore? Maybe we could add a test.

llvm/lib/Target/AIE/AIEBaseInstructionSelector.cpp
```diff
@@ -1628,3 +1629,56 @@ bool AIEBaseInstructionSelector::selectVST_FIFO_Flush3D(
...
+
+// Ptr-coupled locks cannot be handled by a single TableGen Pat<> on *_ptr:
+// - eP:$ptr in the input but not on ACQ/REL is rejected ("dead named input").
+// - PatInaccessibleMem skips the mayLoad mismatch but not the dead-operand
+// check.
+// - ACQ/REL have mayLoad=false, so GISel patterns cannot merge the annotative
+// MMO.
+// Save the MMO, rewrite to the id-only intrinsic for existing Pat<> rules,
+// select, then attach the saved MMO to the lock MI in C++.
+bool AIEBaseInstructionSelector::selectLockPtrIntrinsic(MachineInstr &I) {
```
andcarminati, line 1674: I was wondering if it is better to handle everything with C++ code. I guess we have more code just to work around the automatic selection: TRI hooks + select + prepare + attach.

andcarminati: Just an idea, of course there are pros and cons considering each approach.

katerynamuts:
> Quick Question: why do we need to update AIERegMemEventTracker if we also update the DAG construction? Wouldn't the latencies edges provide us enough distance for scheduling?

@F-Stuckmann  as far as I understand AIERegMemEventTracker covers a different class of constraints than the DAG, e.g., prologue/epilogue where lock↔mem pairs are not connected by DAG edges at all.

clang/lib/Headers/aie2p/aie2p_locks.h
katerynamuts: The -val isn’t ptr-specific, acquire_greater_equal does the same thing for the non-ptr overload. There’s only one acq / __builtin_aie2p_acquire. The sign of the value operand selects the mode:
+val → wait until lock count == val (acquire_equal)
-val → wait until lock count >= val (acquire_greater_equal)

llvm/lib/CodeGen/MachineInstr.cpp
```diff
@@ -1387,10 +1390,10 @@ bool MachineInstr::isDead(const MachineRegisterInfo &MRI,
   return wouldBeTriviallyDead();
 }
 
-static bool MemOperandsHaveAlias(const MachineFrameInfo &MFI,
-                                 BatchAAResults *AA, bool UseTBAA,
-                                 const MachineMemOperand *MMOa,
-                                 const MachineMemOperand *MMOb) {
+bool MachineInstr::memOperandsHaveAlias(const MachineFrameInfo &MFI,
```
andcarminati: I don't think we need to turn this into a member function. The function itself is quite stable and it does no access any MI field (why it is static). I can argue in favor of a small code duplication (copy this function somewhere in the target) instead of a core code change that may case bump conflics.

llvm/lib/Target/AIE/AIEBaseSubtarget.cpp
```diff
@@ -221,6 +230,10 @@ class LockDelays : public ScheduleDAGMutation {
         if (PredEdge.getKind() != SDep::Order || !LdSt->mayLoadOrStore()) {
           continue;
         }
+        if (TII->isLock(MI->getOpcode()) &&
+            !TII->mayLockOrderWithMemOp(*MI, *LdSt,
```
andcarminati: Turn this into a small lambda and reuse below?

llvm/test/CodeGen/AIE/aie2p/schedule/postpipeliner/pipelined-lock-ptr-alias.mir
katerynamuts: yes, we need them because each MIR function needs a matching IR function definition so %ir.store_buf / %ir.lock_buf resolve. Alias analysis uses those to decide lock spacing. The difference between cases is which %ir.* name each MIR instruction uses, not the IR body.

llvm/lib/Target/AIE/AIERegMemEventTracker.h
```diff
@@ -99,6 +110,18 @@ class AIERegMemEventTracker {
...
+  // = earliest in program). Used for ACQ stall (mem must finish before lock).
+  int getMaxAliasingLastMemCycleForLock(const MachineInstr &Lock, bool IsStore,
+                                        BatchAAResults *BatchAA) const;
+
+  // Among aliasing mem ops keyed by FIRST mem stage (backward coords: max Cycle
+  // = earliest in program). Used for lock resume (mem must start after resume).
+  int getMaxAliasingFirstMemCycleForLock(const MachineInstr &Lock,
+                                         BatchAAResults *BatchAA) const;
+
+  void addPerInstructionLockOrdMemCycle(int LockOrdCycle, MachineInstr *MI);
```
andcarminati, line 138: Maybe a comment here as well.

llvm/lib/CodeGen/MachineInstr.cpp
andcarminati: We could also locally rewrite this function to use directly AAResults instead of batch, so we may simplify a bit AIERegMemEventTracker.

andcarminati:
I see a potential problem related to part-word stores. They Internally perform a __load-modify-store__ (RMW) sequence. In this way, we may have imprecise size information in MMOs (< 4 bytes), so we can have aliasing that AA doesn't correctly model.

The Risk

If:

1. A __part-word store__ instruction accesses address range `[A, A+1]` (1 byte)
2. A __lock__ has an MMO for address range `[A+4, A+8]` (disjoint)
3. AA says "no alias" based on the MMO sizes

Then the lock would __skip__ the scheduling delay against the part-word store, even though:

- The part-word store internally reads a full 4-byte word containing `[A, A+3]`
- The AA might be wrong if the lock buffer overlaps the *wider* RMW region

Some useful tests:

```
; Test for part-word store with disjoint buffers and lock alias analysis
; This test checks if the compiler correctly handles part-word stores with
; disjoint lock buffers - they should keep conservative spacing.

declare void @llvm.aie2p.acquire.ptr(ptr, i32, i32)

; Forward: acquire_ptr on bufA, part-word store to disjoint bufB
; Expected: Conservative spacing should be kept (part-word stores have RMW semantics)
define void @acq_ptr_before_part_word_store_disjoint(ptr noalias %lock_buf, ptr noalias %store_buf, i32 %id, i32 %val, i8 %bval) {
  call void @llvm.aie2p.acquire.ptr(ptr %lock_buf, i32 %id, i32 %val)
  store volatile i8 %bval, ptr %store_buf
  ret void
}

; Backward: st.s8 to bufB, acquire_ptr on disjoint bufA
; Expected: Conservative stall spacing should be kept (part-word stores have RMW semantics)
define void @part_word_store_before_acq_ptr_disjoint(ptr noalias %store_buf, ptr noalias %lock_buf, i32 %id, i32 %val, i8 %bval) {
  store volatile i8 %bval, ptr %store_buf
  call void @llvm.aie2p.acquire.ptr(ptr %lock_buf, i32 %id, i32 %val)
  ret void
}

; For comparison - same buffer version (should definitely keep spacing)
define void @acq_ptr_before_part_word_store_same(ptr %p, i32 %id, i32 %val, i8 %bval) {
  call void @llvm.aie2p.acquire.ptr(ptr %p, i32 %id, i32 %val)
  store volatile i8 %bval, ptr %p
  ret void
}

; For comparison - full word store with disjoint (spacing elided is OK here)
define void @acq_ptr_before_store_disjoint(ptr noalias %lock_buf, ptr noalias %store_buf, i32 %id, i32 %val) {
  call void @llvm.aie2p.acquire.ptr(ptr %lock_buf, i32 %id, i32 %val)
  store volatile i32 %val, ptr %store_buf
  ret void
}

```

Maybe this is enough to fix:

```
bool AIEBaseInstrInfo::mayLockOrderWithMemOp(const MachineInstr &Lock,
                                             const MachineInstr &Mem,
                                             BatchAAResults *AA,
                                             bool UseTBAA) const {
  ...
  // Part-word memory operations have special RMW semantics — alias analysis
  // on the narrow store may miss overlap with the wider load portion.
  if (isPartWordMemoryInst(Mem))
    return true;
  
  for (const MachineMemOperand *MemMMO : Mem.memoperands())
    if (MachineInstr::memOperandsHaveAlias(MFI, AA, UseTBAA, LockMMO, MemMMO))
      return true;

  return false;
}
```

llvm/test/CodeGen/AIE/aie2ps/aie2ps-lock-ptr-alias.ll
katerynamuts: yes, it is needed because otherwise the stores can removed or moved as unused or unrelated to the lock. volatile keeps them in place so we actually test lock↔memory scheduling.

llvm/lib/Target/AIE/AIERegMemEventTracker.cpp
```diff
@@ -405,9 +476,52 @@ unsigned AIERegMemEventTracker::getSafeOperandsDistanceFromBottom(
...
+      // For release locks, pick the resume cycle by whether the region is
+      // load-only or contains stores (RMW counts as store). Conservative when
+      // the region mixes an early load with a later store: we pair the larger
+      // store-side resume with the earliest cycle (the load's), over-estimating
+      // the load's constraint by up to (ResumeAfterReleaseStore -
+      // ResumeAfterReleaseLoad) cycles.
+      bool HasAliasingStore = false;
+      bool HasAliasingLoad = false;
+      if (BatchAA) {
+        auto hasAliasingMem = [&](const auto &MemMap) {
```
andcarminati: nit:
```
        auto hasAliasingMem = [&](const auto &MemMap) {
          for (const auto &[_, MemOps] : MemMap) {
            for (const MachineInstr *MemOp : MemOps) {
              if (TII->mayLockOrderWithMemOp(MI, *MemOp, &*BatchAA))
                return true;
            }
          }
          return false;
        };
```

llvm/lib/Target/AIE/AIEBaseInstructionSelector.cpp
katerynamuts: I was also thinking about it but I like the current hybrid approach because otherwise we would duplicate the eight lock Pat<> rules per target without really reducing complexity.

llvm/lib/Target/AIE/AIEBaseSubtarget.cpp
katerynamuts: this is actually a good point! I've added tests and I see that that ptr-coupled lock spacing doesn't distinguish
same-port vs disjoint-port for FIFO pop/push. As far as I can see it is because lock ptr MMOs are unknown-size and FIFO pop/push MMOs are also unknown-size. AA returns MayAlias when both MMOs are unknown-size, so we always orders locks against loop FIFO memory ops even when IR pointers are noalias. I think we should keep in mind and fix it in a follow-up PR.

katerynamuts:
> I see a potential problem related to part-word stores. They Internally perform a **load-modify-store** (RMW) sequence. In this way, we may have imprecise size information in MMOs (< 4 bytes), so we can have aliasing that AA doesn't correctly model.
> 
> The Risk
> 
> If:
> 
> 1. A **part-word store** instruction accesses address range `[A, A+1]` (1 byte)
> 2. A **lock** has an MMO for address range `[A+4, A+8]` (disjoint)
> 3. AA says "no alias" based on the MMO sizes
> 
> Then the lock would **skip** the scheduling delay against the part-word store, even though:
> 
> * The part-word store internally reads a full 4-byte word containing `[A, A+3]`
> * The AA might be wrong if the lock buffer overlaps the _wider_ RMW region
> 
> Some useful tests:
> 
> ```
> ; Test for part-word store with disjoint buffers and lock alias analysis
> ; This test checks if the compiler correctly handles part-word stores with
> ; disjoint lock buffers - they should keep conservative spacing.
> 
> declare void @llvm.aie2p.acquire.ptr(ptr, i32, i32)
> 
> ; Forward: acquire_ptr on bufA, part-word store to disjoint bufB
> ; Expected: Conservative spacing should be kept (part-word stores have RMW semantics)
> define void @acq_ptr_before_part_word_store_disjoint(ptr noalias %lock_buf, ptr noalias %store_buf, i32 %id, i32 %val, i8 %bval) {
>   call void @llvm.aie2p.acquire.ptr(ptr %lock_buf, i32 %id, i32 %val)
>   store volatile i8 %bval, ptr %store_buf
>   ret void
> }
> 
> ; Backward: st.s8 to bufB, acquire_ptr on disjoint bufA
> ; Expected: Conservative stall spacing should be kept (part-word stores have RMW semantics)
> define void @part_word_store_before_acq_ptr_disjoint(ptr noalias %store_buf, ptr noalias %lock_buf, i32 %id, i32 %val, i8 %bval) {
>   store volatile i8 %bval, ptr %store_buf
>   call void @llvm.aie2p.acquire.ptr(ptr %lock_buf, i32 %id, i32 %val)
>   ret void
> }
> 
> ; For comparison - same buffer version (should definitely keep spacing)
> define void @acq_ptr_before_part_word_store_same(ptr %p, i32 %id, i32 %val, i8 %bval) {
>   call void @llvm.aie2p.acquire.ptr(ptr %p, i32 %id, i32 %val)
>   store volatile i8 %bval, ptr %p
>   ret void
> }
> 
> ; For comparison - full word store with disjoint (spacing elided is OK here)
> define void @acq_ptr_before_store_disjoint(ptr noalias %lock_buf, ptr noalias %store_buf, i32 %id, i32 %val) {
>   call void @llvm.aie2p.acquire.ptr(ptr %lock_buf, i32 %id, i32 %val)
>   store volatile i32 %val, ptr %store_buf
>   ret void
> }
> ```
> 
> Maybe this is enough to fix:
> 
> ```
> bool AIEBaseInstrInfo::mayLockOrderWithMemOp(const MachineInstr &Lock,
>                                              const MachineInstr &Mem,
>                                              BatchAAResults *AA,
>                                              bool UseTBAA) const {
>   ...
>   // Part-word memory operations have special RMW semantics — alias analysis
>   // on the narrow store may miss overlap with the wider load portion.
>   if (isPartWordMemoryInst(Mem))
>     return true;
>   
>   for (const MachineMemOperand *MemMMO : Mem.memoperands())
>     if (MachineInstr::memOperandsHaveAlias(MFI, AA, UseTBAA, LockMMO, MemMMO))
>       return true;
> 
>   return false;
> }
> ```

@andcarminati actually, we don't what to be conservative for part-word stores when the user guarantees through restrict that the memories are disjoint. I think these tests are not fully representative:

> ; Forward: acquire_ptr on bufA, part-word store to disjoint bufB
> ; Expected: Conservative spacing should be kept (part-word stores have RMW semantics)
> define void @acq_ptr_before_part_word_store_disjoint(ptr noalias %lock_buf, ptr noalias %store_buf, i32 %id, i32 %val, i8 %bval) {
>   call void @llvm.aie2p.acquire.ptr(ptr %lock_buf, i32 %id, i32 %val)
>   store volatile i8 %bval, ptr %store_buf
>   ret void
> }
> 
> ; Backward: st.s8 to bufB, acquire_ptr on disjoint bufA
> ; Expected: Conservative stall spacing should be kept (part-word stores have RMW semantics)
> define void @part_word_store_before_acq_ptr_disjoint(ptr noalias %store_buf, ptr noalias %lock_buf, i32 %id, i32 %val, i8 %bval) {
>   store volatile i8 %bval, ptr %store_buf
>   call void @llvm.aie2p.acquire.ptr(ptr %lock_buf, i32 %id, i32 %val)
>   ret void
> }

because here we set noalias to lock_buf and store_buf, so the user gives this information and we can remove the edge between the lock and the part-word store. If we don't pass noalias than we remain conservative and keep the gap. I added the tests to show what we get in different scenarios.

katerynamuts:
@khallouh @andcarminati thank you for reviewing the PR, I addressed all your comments in follow-up commits; please take another look when you have a chance.

llvm/lib/Target/AIE/AIEBaseInstrInfo.h
```diff
@@ -271,6 +271,10 @@ struct AIEBaseInstrInfo : public TargetInstrInfo {
   virtual unsigned getAddrIntrinsic3D() const {
     llvm_unreachable("Target didn't implement getAddrIntrinsic3D");
   }
+  virtual std::optional<unsigned>
```
andcarminati: We have a special maker that could be used here instead of optional: `Intrinsic::not_intrinsic`

llvm/lib/Target/AIE/AIEBaseInstructionSelector.cpp
```diff
@@ -1628,3 +1629,56 @@ bool AIEBaseInstructionSelector::selectVST_FIFO_Flush3D(
...
+  const MachineMemOperand &OrigMMO = *I.memoperands()[0];
+  SavedMMO = MF->getMachineMemOperand(&OrigMMO, OrigMMO.getOffset(),
+                                      OrigMMO.getSize());
+
+  std::optional<unsigned> IdOnlyIntrinsic =
+      TII.getIdOnlyLockIntrinsic(cast<GIntrinsic>(I).getIntrinsicID());
+  if (!IdOnlyIntrinsic)
+    return false;
+
+  unsigned IntrOpIdx = I.getNumExplicitDefs();
```
andcarminati: const?

llvm/lib/Target/AIE/AIEBaseInstructionSelector.cpp
```diff
@@ -1628,3 +1629,56 @@ bool AIEBaseInstructionSelector::selectVST_FIFO_Flush3D(
...
+// Save the MMO, rewrite to the id-only intrinsic for existing Pat<> rules,
+// select, then attach the saved MMO to the lock MI in C++.
+bool AIEBaseInstructionSelector::selectLockPtrIntrinsic(MachineInstr &I) {
+  MachineMemOperand *SavedMMO = nullptr;
+  MachineBasicBlock *MBB = I.getParent();
+  if (!prepareLockPtrIntrinsic(I, SavedMMO))
+    return false;
+  if (!selectImpl(I, *CoverageInfo))
+    return false;
+  return attachLockPtrMemOperand(*MBB, SavedMMO);
```
andcarminati, line 1682: Should we assert SavedMMO instead of re-checking it in attachLockPtrMemOperand?

llvm/lib/Target/AIE/AIEBaseInstructionSelector.cpp
```diff
@@ -1628,3 +1629,56 @@ bool AIEBaseInstructionSelector::selectVST_FIFO_Flush3D(
...
+  if (!SavedMMO)
+    return false;
+
+  for (MachineInstr &MI : MBB) {
+    if (TII.isLock(MI.getOpcode()) && MI.memoperands_empty()) {
+      MI.addMemOperand(*MF, SavedMMO);
+      return true;
+    }
+  }
+  return false;
```
andcarminati: Should this return be an unreachable? I see some contracts alongside the path that leads to here.

llvm/lib/CodeGen/MachineInstr.cpp
```diff
@@ -1390,10 +1387,10 @@ bool MachineInstr::isDead(const MachineRegisterInfo &MRI,
   return wouldBeTriviallyDead();
 }
 
-bool MachineInstr::memOperandsHaveAlias(const MachineFrameInfo &MFI,
-                                        BatchAAResults *AA, bool UseTBAA,
-                                        const MachineMemOperand *MMOa,
-                                        const MachineMemOperand *MMOb) {
+static bool MemOperandsHaveAlias(const MachineFrameInfo &MFI,
```
andcarminati, line 1390: Could we completely revert all changes related to this file? Why upper case?

llvm/lib/Target/AIE/AIEBaseSubtarget.cpp
```diff
@@ -225,14 +222,18 @@ class LockDelays : public ScheduleDAGMutation {
                    TII->getDoneLatency(MI->getOpcode()))) {
         continue;
       }
+      auto ordersWithMemOp = [&](const MachineInstr &LockOrDone,
```
andcarminati: nit: `OrdersWithMemOp `

andcarminati:
> because here we set noalias to lock_buf and store_buf, so the user gives this information and we can remove the edge between the lock and the part-word store. If we don't pass noalias than we remain conservative and keep the gap. I added the tests to show what we get in different scenarios.

Please, keep in mind that AA may give wrong NoAlias (in practice) for part-word stores. noalias is not a guarantee for par-word stores, since we need to access adjacent memory locations not accounted by AA.

llvm/lib/CodeGen/MachineInstr.cpp
katerynamuts: I did it in this follow-up [commit ](https://github.com/Xilinx/llvm-aie/pull/1123/changes/0caaf6829b2f1d63f56f42634fc0016fc3dc1ed1), I will squash it to the original commit to avoid confusion.

katerynamuts:
> > because here we set noalias to lock_buf and store_buf, so the user gives this information and we can remove the edge between the lock and the part-word store. If we don't pass noalias than we remain conservative and keep the gap. I added the tests to show what we get in different scenarios.
> 
> Please, keep in mind that AA may give wrong NoAlias (in practice) for part-word stores. noalias is not a guarantee for par-word stores, since we need to access adjacent memory locations not accounted by AA.

since AA is unreliable for part-word stores I implemented it so that part-word stores no longer use AA but rely only user's annotations: no user annotation (no restrict) - stay conservative.

llvm/lib/Target/AIE/AIEBaseInstructionSelector.cpp
```diff
@@ -1628,3 +1629,55 @@ bool AIEBaseInstructionSelector::selectVST_FIFO_Flush3D(
...
+// select, then attach the saved MMO to the lock MI in C++.
+bool AIEBaseInstructionSelector::selectLockPtrIntrinsic(MachineInstr &I) {
+  MachineMemOperand *SavedMMO = nullptr;
+  MachineBasicBlock *MBB = I.getParent();
+  if (!prepareLockPtrIntrinsic(I, SavedMMO))
+    return false;
+  assert(SavedMMO && "expected memoperand saved from lock ptr intrinsic");
+  if (!selectImpl(I, *CoverageInfo))
+    return false;
+  return attachLockPtrMemOperand(*MBB, SavedMMO);
```
andcarminati, line 1682: If we prepare and select, is it possible for attachLockPtrMemOperand still fail?

llvm/lib/Target/AIE/AIERegMemEventTracker.cpp
```diff
@@ -405,9 +475,46 @@ unsigned AIERegMemEventTracker::getSafeOperandsDistanceFromBottom(
...
+      // For release locks, pick the resume cycle by whether the region is
+      // load-only or contains stores (RMW counts as store). Conservative when
+      // the region mixes an early load with a later store: we pair the larger
+      // store-side resume with the earliest cycle (the load's), over-estimating
+      // the load's constraint by up to (ResumeAfterReleaseStore -
+      // ResumeAfterReleaseLoad) cycles.
+      bool HasAliasingStore = false;
+      bool HasAliasingLoad = false;
+      if (AA) {
+        auto hasAliasingMem = [&](const auto &MemMap) {
```
andcarminati: nit `HasAliasingMem`

llvm/lib/Target/AIE/AIERegMemEventTracker.cpp
```diff
@@ -405,9 +475,46 @@ unsigned AIERegMemEventTracker::getSafeOperandsDistanceFromBottom(
...
+          for (const auto &[_, MemOps] : MemMap) {
```
andcarminati: We have some interleaving logic with getMaxAliasingFirstMemCycleForLock, do you see a way to simplify? I know that this code is hard to maintain indeed.

llvm/lib/Target/AIE/AIEBaseInstrInfo.cpp
```diff
@@ -1107,6 +1110,148 @@ AIEBaseInstrInfo::getLockStallDelay(const MachineInstr &MemMI) const {
...
+    V = GEP->getPointerOperand()->stripPointerCasts();
+  }
+
+  return isIdentifiedObject(V) ? V : nullptr;
+}
+
+/// Part-word stores may elide lock delays only when lock and store pointers
+/// trace to distinct IR objects that carry source-level noalias (restrict).
+static bool
+partWordStoreLockMayUseAliasAnalysis(const MachineMemOperand *LockMMO,
```
andcarminati, line 1132: I think this is optimistic. Do we have an use case where this optimistic assumption can help? My point is that, no AA information is 100% correct when dealing with part word stores. The AIE backend mitigates this by conservatively treating all part-word memory operations as potentially aliasing with all other memory operations, bypassing the standard alias analysis in critical scheduling paths.

andcarminati: Discussed offline: we should keep part-word stores as is - requirement.

llvm/lib/Target/AIE/AIEBaseInstructionSelector.cpp
katerynamuts: theoretically no, but to be on the safe side we have llvm_unreachable to catch any bug.

andcarminati: Ok, maybe the return type there could be void, I don't think boolean can help something.

llvm/lib/Target/AIE/AIERegMemEventTracker.cpp
katerynamuts: I tried my best to do it. I added a follow-up commit for the refactoring.

andcarminati (approved):
LGTM. Nice end-to-end work! Thanks for addressing the comments, I hope we will be able to move RegMenEventTracker semantics to InterBlock edges, so we can keep a single source of truth regarding locks.

llvm/lib/Target/AIE/AIEBaseSubtarget.cpp
```diff
@@ -216,11 +222,19 @@ class LockDelays : public ScheduleDAGMutation {
...
+          return true;
+        return TII->mayLockOrderWithMemOp(LockOrDone, Mem, AARes);
+      };
       for (auto &PredEdge : SU.Preds) {
         MachineInstr *LdSt = PredEdge.getSUnit()->getInstr();
         if (PredEdge.getKind() != SDep::Order || !LdSt->mayLoadOrStore()) {
           continue;
         }
+        if (!OrdersWithMemOp(*MI, *LdSt))
+          continue;
```
mludevid, line 237: @katerynamuts, question for a potential follow-up: Right now we are skipping these edges, which leaves them as they were before the DAG mutator, which as far as I understand means an artificial edge of latency 0 between the involved instructions. If the MMO proves the lock and the memory instructions are unrelated, shouldn't we be able to drop the edge?

PR #1155, opened 2026-07-20
[AIE2P] Range-check spill immediate offsets in eliminateFrameIndex

atassis:
`eliminateFrameIndex` has several spill case-groups, and two of them wrote the frame offset straight
into the instruction's immediate with no encodability check. Only the middle group checked, via
`isEncodableAsNegativeInt<9, 4>`. So the assembly printer was rejecting addresses that should never
have been formed.

@khallouh was right on the first version of this PR: the store was not the defect, and the indexed
form I originally proposed would have encoded cleanly and written 64 bytes to the wrong place. That
patch is gone. This one range-checks the offset and falls back to the indexed form only once the
offset is known not to fit.

Thresholds are per opcode rather than one number, because the field width differs by the form the
pseudo expands to. The composed pseudos are retagged to a concrete opcode in `expandPostRAPseudo`,
which runs after this pass, so they are resolved the same way here and checked against the field they
actually end up encoded in.

## The ePSRFLdF composite spill (folded in from #1276)

The `VLDA_PLFR_SPILL`/`VST_PLFR_SPILL` slot is `[sub_fifo 128B][sub_avail 4B][sub_ptr 4B]`. `sub_fifo`
moves through the FIFO path, whose SP offset must be a multiple of 64, and ePSRFLdF's spill alignment
guarantees it -- so a misaligned offset there means a **wrong frame object**, not a large one.

That case therefore must NOT take the indexed fallback, for exactly the reason you gave: the indexed
form would encode a truncated address cleanly. It is split into its own case group, asserts the
alignment, and routes a misaligned offset to the emitter, which rejects it. Only an aligned offset
that is out of range takes the register form.

This was open separately as #1276 and is folded in here, because it changes the same function that
this PR changes and completes the same story. Reviewing them apart would have meant reading
`eliminateFrameIndex` twice. #1276 is closed in favour of this.

## Tests

- `llvm/test/CodeGen/AIE/aie2p/eliminate-frame-index.mir` -- the range check per case-group.
- `llvm/test/CodeGen/AIE/aie2p/spill/spill-reload-psrf-prologepilog.mir` -- the composite spill,
  aligned in-range, aligned out-of-range, and the prologue/epilogue path.

Built with assertions on, since the composite case's guard is an `assert`.

### DIFF
--- a/llvm/lib/Target/AIE/aie2p/AIE2PInstrInfo.cpp
+++ b/llvm/lib/Target/AIE/aie2p/AIE2PInstrInfo.cpp
@@ -1222,6 +1222,22 @@ AIE2PInstrInfo::getRegOffsetSpillInstrInfoFromImmOffset(
     return {AIE2P::VLDA_dmx_lda_x_idx, AIE2P::MOVXM, &AIE2P::eDJRegClass};
   case AIE2P::VST_dmx_sts_x_spill:
     return {AIE2P::VST_dmx_sts_x_idx, AIE2P::MOVXM, &AIE2P::eDJRegClass};
+  case AIE2P::LDA_dmv_lda_q_spill:
+    return {AIE2P::LDA_dmv_lda_q_idx, AIE2P::MOVXM, &AIE2P::eDJRegClass};
+  case AIE2P::ST_dmv_sts_q_spill:
+    return {AIE2P::ST_dmv_sts_q_idx, AIE2P::MOVXM, &AIE2P::eDJRegClass};
+  case AIE2P::VLDA_128_dmv_lda_w_spill:
+    return {AIE2P::VLDA_128_dmv_lda_w_idx, AIE2P::MOVXM, &AIE2P::eDJRegClass};
+  case AIE2P::VST_128_dmv_sts_w_spill:
+    return {AIE2P::VST_128_dmv_sts_w_idx, AIE2P::MOVXM, &AIE2P::eDJRegClass};
+  case AIE2P::VLDA_dmw_lda_w_spill:
+    return {AIE2P::VLDA_dmw_lda_w_idx, AIE2P::MOVXM, &AIE2P::eDJRegClass};
+  case AIE2P::VST_dmw_sts_w_spill:
+    return {AIE2P::VST_dmw_sts_w_idx, AIE2P::MOVXM, &AIE2P::eDJRegClass};
+  case AIE2P::VLDA_dmx_lda_bm_spill:
+    return {AIE2P::VLDA_dmx_lda_bm_idx, AIE2P::MOVXM, &AIE2P::eDJRegClass};
+  case AIE2P::VST_dmx_sts_bm_spill:
+    return {AIE2P::VST_dmx_sts_bm_idx, AIE2P::MOVXM, &AIE2P::eDJRegClass};
   default:
     llvm_unreachable("Offset register spill instruction info un-implemented");
   }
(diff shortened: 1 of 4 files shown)


khallouh:
Hi @atassis, thanks for looking into this. I think the stack offset is rejected by the Assembly Printer because it is unaligned. We likely have a bug in the offset computation and using indexed store instead of immediate store here would just hide the bug and miscompile.

khallouh (changes requested):
Requesting changes because the bug is somewhere else, hence the asm printer rejecting the store address

atassis:
You are right, and worse than the abort: the hardware truncates the address, so the indexed store
would have encoded cleanly and written 64 bytes to the wrong place.

Confirming it, the selector already keys on alignment: the same `store <16 x i32>` lowers to `vst x0`
at align 64, `vst wl0`/`vst wh0` at align 32, and scalar stores at align 16. With
`StackAlignment = Align(64)` a frame object keeps its alignment in its SP offset, so a non-multiple of
64 means the frame object is wrong. The kernel that produced -184 does not reproduce on current
aie-public either.

So the misalignment fallback is dropped and only the out-of-range case is left, where the offset is a
multiple of 64 but outside the 9-bit field, which is what the `VST_PLFR_SPILL` group below already does
for its own range. A misaligned offset now asserts, and without assertions it falls through to the
immediate so the emitter still rejects it. Happy to drop the assert if you would rather not have it.

atassis:
You were right, and the store was not the defect. `eliminateFrameIndex` has three spill
case-groups and two of them wrote the frame offset into the immediate with no encodability
check at all; only the middle group checked, via `isEncodableAsNegativeInt<9,4>`. So the
asm printer was rejecting an address that should never have been formed, and the indexed
store I proposed would have encoded it cleanly and written to the wrong place.

Replaced the patch with the range check. Falling back to the indexed form is still how it
recovers, but only once the offset is known not to fit, which is what the second group
already did.

Thresholds are per opcode rather than one number: the field is a 9-bit negative scaled by
the access size, so 2K, 8K, 16K or 32K. 16 opcodes reproduce, 10 of them straight out of
`storeRegToStackSlot` / `loadRegFromStackSlot`.

atassis:
@khallouh this is ready for another look. The patch is not the one you reviewed.

You were right that the store was not the defect. `eliminateFrameIndex` has three spill case-groups and two of them wrote the frame offset into the immediate with no encodability check at all, so the assembly printer was rejecting an address that should never have been formed. The indexed store I first proposed would have encoded cleanly and written 64 bytes to the wrong place.

The current patch range-checks the offset and only falls back to the indexed form once it is known not to fit.

atassis:
Force-pushed: rebased onto current `aie-public` (7753da6c), and folded in #1276, which changed the same function.

@khallouh the range-check patch you were asked to re-look at on 2026-09-04 is unchanged in substance -- what is new is the `VLDA_PLFR_SPILL`/`VST_PLFR_SPILL` case, split out into its own group. That case is the one where your original point bites hardest: its slot must be 64-byte aligned, so a misaligned offset there is a wrong frame object, and the indexed fallback would encode a truncated address cleanly. It now asserts the alignment and routes a misaligned offset to the emitter instead, and only takes the register form for an aligned offset that is out of range.

Keeping it in one PR so `eliminateFrameIndex` only has to be read once.

Built with assertions on; `llvm/test/CodeGen/AIE/aie2p` is 518 passed / 0 failed, including the two tests here.

PR #1177, opened 2026-07-21
Bump required CXX standard to 20

philippjh:


amd-lauretta (approved):
Thanks for the update

PR #1187, opened 2026-07-22
[AIEX] Safeguard outer-loop pipeliner fallback from side effects (bug fix)

khallouh:


llvm/lib/Target/AIE/AIEOuterLoopPipeliner.cpp
```diff
@@ -895,6 +895,17 @@ void OrigLoopStructure::collectStages(
 }
 
 void OrigLoopStructure::collectStage0FromInnerLoop() {
+  // The fallback only follows SSA def-use edges from the inner loop. Do not
+  // pipeline when the top contains an effectful candidate: its consumer may be
+  // connected through target state rather than SSA, so moving its operands
+  // without moving the operation would be incorrect.
+  bool HasTopSideEffects = false;
```
andcarminati, line 904: Could we turn this as a profitability check? Something that we can call from `tryPipelineLoop`? We can bailout very early I would say.

khallouh: yeah but it should be called safety check rather than profitability check, since without it it is incorrect not suboptimal. That been said, I kept it in the collection logic because we may still wanna adjust this later to reproduce these effectful instructions where they should be and still have a correct pipeline rather than bail. It is just not useful at this moment for the known cases.

andcarminati: Maybe write in the comment that we are bailing out, but we could extend the logic here to cover those cases in some way.

andcarminati (approved):
LGTM. Thanks for catching this bug.

PR #1189, opened 2026-07-22
[AIE] Add speculative outer-loop pipelining

khallouh:


llvm/test/CodeGen/AIE/aie2p/outer-loop-pipeliner/outer-loop-pipelining-speculative-no-anchor.ll
```diff
@@ -0,0 +1,57 @@
...
+; With no stage split anchor, fallback collection places the load in stage 0.
+; Its stage-0 chain is side-effect-free, so explicitly requested speculative
+; mode is selected and no last-iteration region is created.
+
+declare void @llvm.set.loop.iterations.i32(i32)
+declare i1 @llvm.loop.decrement.i32(i32)
+
+; CHECK-LABEL: define void @speculative_no_anchor(
+; CHECK:       stage0.top:
+; CHECK:       load i32, ptr %a
```
F-Stuckmann: could you provide the full checklines? I would like see where call lands

khallouh: which call?

llvm/lib/Target/AIE/AIEOuterLoopPipeliner.cpp
```diff
@@ -112,6 +112,13 @@ static cl::opt<bool> SplitStages(
              "different strategies to reach more compact schedules"),
     cl::init(true), cl::Hidden);
 
+static cl::opt<bool> SpeculativeLastIteration(
+    "aie-outer-loop-pipelining-speculative",
+    cl::desc("Use speculative loads on the last iteration instead of creating "
+             "a cooldown region. Avoids code duplication and register "
```
F-Stuckmann: nit: we changed naming scheme. I don't understand the difference between last iteration and cooldown?

llvm/lib/Target/AIE/AIEOuterLoopPipeliner.cpp
```diff
@@ -1696,6 +1731,15 @@ bool AIEOuterLoopPipeliner::performTransformation(
     return false;
   }
 
+  // Speculation is opt-in and only allowed when the stage-0 chain is
+  // side-effect-free.
+  const bool Stage0IsSideEffectFree =
+      !llvm::any_of(OrigLS.stage0Insts(), [](const Instruction *I) {
```
F-Stuckmann: use llvm::none_of

llvm/test/CodeGen/AIE/aie2p/outer-loop-pipeliner/outer-loop-pipelining-speculative-no-anchor.ll
F-Stuckmann: setting the iterations.
Could you please show the whole code, i have a nagging feeling we are exiting OLP behaviour and the missing checklines are muddying what we are actually expecting here.  

Also could we have a clear separation of unit-test and then the change. i think we are dropping LastIter, so we are actually not OLPing anymore and instead speculatively executing stage0top and moving stage0top into the latch. 
If you want to keep the limited checks, i would at least expect to see where  `load i32, ptr %a` lives, but since this is actually not OLP behaviour anymore but peeling, i would prefer to have the full checklines.

llvm/lib/Target/AIE/AIEOuterLoopPipeliner.cpp
```diff
@@ -1696,6 +1731,15 @@ bool AIEOuterLoopPipeliner::performTransformation(
     return false;
   }
 
+  // Speculation is opt-in and only allowed when the stage-0 chain is
```
F-Stuckmann: nit: technically we have a 1 stage pipeline where we speculatively peel out stage0 instructions into the preheader. 

The enable options should also make clear that we are speculatively peeling and not OLPing anymore.

llvm/lib/Target/AIE/AIEOuterLoopPipeliner.cpp
```diff
@@ -1715,22 +1759,28 @@ bool AIEOuterLoopPipeliner::performTransformation(
...
+      convertOuterLoopToHardwareLoop(SteadyLS);
+
+    // Create the last-iteration region from the steady loop.
+    peelLastIteration(OrigLS, SteadyLS);
+
+    // Adjust itercount metadata to reflect the reduced trip count.
+    SteadyLS.updateLoopMetadata();
+  } else {
+    wireSteadyIntoExit(OrigLS, SteadyLS);
+    LLVM_DEBUG(dbgs() << "    Speculative last iteration: no last-iteration "
```
F-Stuckmann, line 1849: nit: update comment for speculative peeling. we should make it clear that we are now not OLPing anymore.

llvm/lib/Target/AIE/AIEOuterLoopPipeliner.cpp
```diff
@@ -1333,6 +1345,29 @@ void AIEOuterLoopPipeliner::wireLastIterIntoCFG(
                                                   LastIterLS.getTop());
 }
 
+void AIEOuterLoopPipeliner::wireSteadyIntoExit(
```
F-Stuckmann: nit: this is duplicated code from wireLastIterIntoCFG. 
Can we refactor the utilitiy and reuse wireLastIterIntoCFG and when we remap SteadyLS in https://github.com/Xilinx/llvm-aie/pull/1189/changes/333895b16fb2f5239e0b2f7d5c641c7048b4e493#diff-2a345fd214a7ed161996ef460fee9ef52d0b515704cddd51e42c5f4cefbe2f9eR1780 ?

llvm/lib/Target/AIE/AIEOuterLoopPipeliner.cpp
```diff
@@ -1777,7 +1777,13 @@ bool AIEOuterLoopPipeliner::performTransformation(
     // Adjust itercount metadata to reflect the reduced trip count.
     SteadyLS.updateLoopMetadata();
   } else {
+    // Exit live-outs must keep the original IV cycle alive when it is observed
+    // outside the loop, before JNZD removes the latch condition.
     wireSteadyIntoExit(OrigLS, SteadyLS);
+    const bool OuterIsHardwareLoop = EnableOuterLoopHardwareLoop &&
+                                     SteadyLS.latchCondition().isDowncounting();
+    if (OuterIsHardwareLoop)
```
F-Stuckmann, line 1846: nit: could we refactor the profitability check of the OLP out and instead do the JNZD conversion where it belongs: https://github.com/Xilinx/llvm-aie/blob/aie-public/llvm/lib/Target/AIE/AIEBaseTargetTransformInfo.cpp#L320
We can add another constraint that check if OLP is profitable and will run and then allow the default hardware loop conversion

F-Stuckmann: as a nice side effect we can start deleting duplicated code

F-Stuckmann (changes requested):
you are changing default behavior and causing performance regressions in `llvm/test/CodeGen/AIE/aie2ps/end-to-end/conv2d_opt_outerloop_pipelining_out_mode_1.ll` and `llvm/test/CodeGen/AIE/aie2ps/end-to-end/gemm_int8_psum_1.ll` [cherry-pick afb9ed6778167e2cd7941fb7f01f01e00115b6ba] for aie2ps. 

either 
a) drop commit 3
b) find a better heuristic that does not cause this much performance regressions across the board or 
c) add a loophint to enable speculative peeling instead of OLPing.

F-Stuckmann:
Heads up i merged the gemm OLP variants into aie-private. 
if you fix the conflict, you can see the performance loss in gemm_psum_1.

khallouh:
> you are changing default behavior and causing performance regressions in `llvm/test/CodeGen/AIE/aie2ps/end-to-end/conv2d_opt_outerloop_pipelining_out_mode_1.ll` and `llvm/test/CodeGen/AIE/aie2ps/end-to-end/gemm_int8_psum_1.ll` [cherry-pick [afb9ed6](https://github.com/Xilinx/llvm-aie/commit/afb9ed6778167e2cd7941fb7f01f01e00115b6ba)] for aie2ps.
> 
> either a) drop commit 3 b) find a better heuristic that does not cause this much performance regressions across the board or c) add a loophint to enable speculative peeling instead of OLPing.

Yes, I have already run the benchmarks and reported on such regressions offline. I decided to make speculative mode opt-in only in lean mode and via loop hint. I will update once my other PR for lean mode gets merged and address your other comments.

andcarminati:
Could you please attach a metadata stating that a loop was speculatively pipelined? This can be used by postmisched to trigger (or not) some specific optimizations.

llvm/test/CodeGen/AIE/aie2p/outer-loop-pipeliner/outer-loop-pipelining-speculative-no-anchor.ll
khallouh: I fully agree about showing the full lines, but there is a trade-off, either showing LLVM IR and having to add and update all check lines manually or going all the way after IRTranslator and getting that automatically from the scripts. I'm going for the latter to address your comment as was done in previous PRs.

llvm/lib/Target/AIE/AIEOuterLoopPipeliner.cpp
khallouh: It was called cooldown before, I'll rename to last iter

llvm/lib/Target/AIE/AIEOuterLoopPipeliner.cpp
khallouh: We are peeling and rotating. So this is software pipelining since the cross iteration overlap is what defines it.
Moreover, the logic lives in the OLP pass and is common with the standard mode. So I thought the name `aie-outer-loop-pipelining-speculative` was appropriate. 
I don't mention the front peeling because it is done anyway by the pipeliner in all cases. 
Feel free to suggest a name you like better.

llvm/lib/Target/AIE/AIEOuterLoopPipeliner.cpp
khallouh: I don’t agree that it is just peeling stage 0. It is software pipelining (OLP), specifically a 2-stage cross-iteration pipeline.
If we would be just peeling, there would be no repeated next-iteration computation and no loop-carried handoff.
That overlap between iterations is what defines pipelining, not whether an epilogue exists or the trip count of the steady loop.

llvm/lib/Target/AIE/AIEOuterLoopPipeliner.cpp
khallouh: Agreed. I'll extract the duplicated code

llvm/lib/Target/AIE/AIEOuterLoopPipeliner.cpp
khallouh: I agree that hardware-loop profitability and conversion should ultimately live in the generic HardwareLoops pass. Since that requires making OLP eligibility reusable and changing the input form OLP receives, I see it as a broader refactor than this speculative-mode enablement.

khallouh:
Closing in favor of https://github.com/Xilinx/llvm-aie/pull/1202

PR #1190, opened 2026-07-22
[AIEX][OLP] resolve WAR dependencies

F-Stuckmann:
To schedule epilogues of Steadystates in the OLP we can remove WriteAfterRead (WAR) dependencies by splitting VRegs and assigning separate PhysRegs to them.

This option has to be enabled by loop hint.


llvm/test/CodeGen/AIE/aie2ps/end-to-end/gemm_int8_psum_0.ll
```diff
@@ -164,28 +164,25 @@ define dso_local void @gemm_int8_psum_0(
...
+; CHECK-NEXT:    vlda bmhh5, [p4], #64; vldb.3d x8, [p0], d0; mov srssign0, r18; vmac dm2, dm2, x3, x10, r8
+; CHECK-NEXT:    vlda bmll1, [p4], #64; vsel.32 x2, x2, x1, r30; vmac dm1, dm1, x5, x6, r8
+; CHECK-NEXT:    vlda bmlh1, [p4], #64; vsel.32 x4, x4, x10, r30; vmac dm0, dm0, x3, x6, r8
+; CHECK-NEXT:    vlda bmhl1, [p4], #64
+; CHECK-NEXT:    vlda bmhh1, [p4], #64; vshuffle x10, x3, x0, r22
+; CHECK-NEXT:    vlda bmll0, [p4], #64; vst.srs.4x dm3, s0, srssign0, [p2], #64; vshuffle x1, x5, x0, r24
+; CHECK-NEXT:    vlda bmlh0, [p4], #64; vst.srs.4x dm2, s0, srssign0, [p2], m5; jnzd r1, r1, p5; vshuffle x10, x10, x0, r20
+; CHECK-NEXT:    vlda bmhl0, [p4], #64; vst.srs.4x dm1, s0, srssign0, [p2], #64; vshuffle x1, x1, x0, r26 // Delay Slot 5
+; CHECK-NEXT:    vlda bmhh0, [p4], #64; vst.2d.srs.4x dm0, s0, srssign0, [p2], d2; movx srssign0, #0; vmov cml3, cml4 // Delay Slot 4
+; CHECK-NEXT:    vmov cmh3, cmh4 // Delay Slot 3
```
andcarminati: Nice 3 cycle reduction.

llvm/lib/Target/AIE/Utils/AIERegAllocationUtils.cpp
```diff
@@ -0,0 +1,42 @@
...
+MCPhysReg AIERegAllocationUtils::findFreeNonOverlappingPhysReg(
+    const LiveInterval &LI, const TargetRegisterClass &RC,
+    ArrayRef<MCPhysReg> AllocationOrder, MCPhysReg AvoidPhysReg,
+    const BitVector &ReservedRegUnits, const TargetRegisterInfo &TRI,
+    LiveRegMatrix &LRM) {
+  for (MCPhysReg PhysReg : AllocationOrder) {
+    if (!RC.contains(PhysReg))
+      continue;
+
+    if (AvoidPhysReg && TRI.regsOverlap(PhysReg, AvoidPhysReg))
```
martien-de-jong: nit: Can't you add the regunits of the PhysReg to avoid to the ReservedRegUnits you pass in?
Alternatively, AvoidPhysreg might be a softer requirement, so keeping it as a fallback if no other register could be found. Perhaps that doesn't serve the purpose, but would give more semantics to this function that you intend for generic use.

llvm/lib/Target/AIE/Utils/AIERegAllocationUtils.h
```diff
@@ -0,0 +1,38 @@
...
+
+namespace llvm {
+
+class LiveInterval;
+class LiveRegMatrix;
+class TargetRegisterClass;
+class TargetRegisterInfo;
+
+namespace AIERegAllocationUtils {
+
```
martien-de-jong, line 30: This needs a description, especially since its in a generic utility header.

llvm/lib/Target/AIE/AIEWawRegRewriter.cpp
```diff
@@ -789,46 +790,33 @@ MCPhysReg AIEWawRegRewriter::getReplacementPhysReg(const Register VReg,
...
+  MCPhysReg PhysReg = AIERegAllocationUtils::findFreeNonOverlappingPhysReg(
+      LI, *RC, LRURegisters, MCRegister::NoRegister, BitVector(), *TRI, *LRM);
+  if (!PhysReg)
+    return MCRegister::NoRegister;
+
+  assert(llvm::is_contained(LRURegisters, PhysReg));
+
+  // Prefer a previously used killed register when it remains free.
+  if (std::optional<Register> KilledReg =
+          getKilledRegAtSingledDefPoint(VReg, *MRI);
```
martien-de-jong: I think this is an existing typo in the name.

llvm/lib/Target/AIE/AIEWawRegRewriter.cpp
```diff
@@ -789,46 +790,33 @@ MCPhysReg AIEWawRegRewriter::getReplacementPhysReg(const Register VReg,
...
     }
-    LLVM_DEBUG(dbgs() << "       Cannot assign " << printReg(VReg, TRI)
-                      << " to " << TRI->getName(PhysReg)
-                      << " due to interference\n");
   }
-  return MCRegister::NoRegister;
+
+  moveRegAndAliasesBack(PhysReg, LRURegisters, TRI);
+  for (MCRegUnit RU : TRI->regunits(PhysReg))
+    UsedUnits.set(RU);
```
martien-de-jong: This could also be a function in AIERegAllocationUtils

llvm/lib/Target/AIE/AIEBaseInstrInfo.h
```diff
@@ -1023,6 +1023,54 @@ struct AIEBaseInstrInfo : public TargetInstrInfo {
...
+  class CopyMaterializer {
+  public:
+    CopyMaterializer(const AIEBaseInstrInfo &TII, const TargetRegisterInfo &TRI,
+                     MachineBasicBlock &MBB, MachineBasicBlock::iterator MBBI,
+                     const DebugLoc &DL);
+    CopyMaterializer(const AIEBaseInstrInfo &TII,
+                     const TargetRegisterInfo &TRI);
+
+    CopyInstrBuilder buildInstr(const MCInstrDesc &Desc, Register DstReg);
+    bool copy(MCRegister DstReg, MCRegister SrcReg, bool KillSrc);
```
martien-de-jong: documentation of these methods please.

llvm/lib/Target/AIE/AIEBaseInstrInfo.cpp
```diff
@@ -811,22 +811,71 @@ static void collectSubRegs(MCRegister Reg, SmallSet<MCRegister, 8> &SubRegs,
...
+bool AIEBaseInstrInfo::CopyMaterializer::copy(MCRegister DstReg,
+                                              MCRegister SrcReg, bool KillSrc) {
+  return TII.materializeCopy(*this, DstReg, SrcReg, KillSrc);
+}
+
+std::optional<unsigned>
+AIEBaseInstrInfo::getCopyCost(const TargetRegisterInfo &TRI, MCRegister DstReg,
+                              MCRegister SrcReg) const {
+  CopyMaterializer M(*this, TRI);
+  if (!M.copy(DstReg, SrcReg, false))
```
martien-de-jong: As far as I understand it we create the instructions and discard them on destroying M. Is there a way to make the construction of the instructions table driven, and obtain the cost from that table entry?

llvm/lib/Target/AIE/aie2/AIE2TargetMachine.cpp
```diff
@@ -193,10 +194,12 @@ bool AIE2PassConfig::addRegAssignAndRewriteOptimized() {
     addPass(createAIESuperRegRewriter());
   }
   addPass(createGreedyRegisterAllocator());
-  if (EnableWAWRegRewrite) {
+  if (EnableWAWRegRewrite)
     addPass(createAIEWawRegRewriter());
+  if (EnableEpilogueRegRewrite)
+    addPass(createAIEEpilogueRegRewriter());
+  if (EnableWAWRegRewrite || EnableEpilogueRegRewrite)
```
martien-de-jong, line 203: Should we lift these kind of things to a base class?

llvm/lib/Target/AIE/AIEEpilogueRegRewriter.cpp
```diff
@@ -0,0 +1,89 @@
...
+    AU.addRequired<LiveIntervalsWrapperPass>();
+    AU.addPreserved<LiveIntervalsWrapperPass>();
+    AU.addRequired<LiveRegMatrixWrapperLegacy>();
+    AU.addPreserved<LiveRegMatrixWrapperLegacy>();
+    AU.addRequired<LiveDebugVariablesWrapperLegacy>();
+    AU.addPreserved<LiveDebugVariablesWrapperLegacy>();
+    MachineFunctionPass::getAnalysisUsage(AU);
+  }
+
+  bool runOnMachineFunction(MachineFunction &MF) override { return false; }
```
martien-de-jong: nit: already move the implementation out of the class body?

llvm/test/CodeGen/AIE/aie2/llc-pipeline-aie2.ll
```diff
@@ -231,6 +231,8 @@
 ; AIE-O123-NEXT:      AIE super-reg rewrite
 ; AIE-O123-NEXT:      Greedy Register Allocator
 ; AIE-O123-NEXT:      AIE waw-reg rewrite
+; AIE-O123-NEXT:      AIE epilogue register rewrite
+; AIE-O123-NEXT:      Live Stack Slot Analysis
```
martien-de-jong: Should this be preserved?

llvm/lib/Target/AIE/AIEEpilogueRegRewriter.cpp
```diff
@@ -66,9 +73,234 @@ class AIEEpilogueRegRewriter : public MachineFunctionPass {
     MachineFunctionPass::getAnalysisUsage(AU);
   }
 
-  bool runOnMachineFunction(MachineFunction &MF) override { return false; }
+  bool runOnMachineFunction(MachineFunction &MF) override;
+
+private:
+  std::optional<RewriteCandidate>
```
martien-de-jong: add docu.

llvm/lib/Target/AIE/AIEEpilogueRegRewriter.cpp
```diff
@@ -66,9 +73,234 @@ class AIEEpilogueRegRewriter : public MachineFunctionPass {
...
+                     VirtRegMap &VRM, LiveRegMatrix &LRM, LiveIntervals &LIS,
+                     LiveDebugVariables &DebugVars) const;
 };
 
+std::optional<RewriteCandidate> AIEEpilogueRegRewriter::collectCandidate(
+    Register Reg, MachineBasicBlock &MBB, MachineRegisterInfo &MRI,
+    const AIEBaseRegisterInfo &TRI, VirtRegMap &VRM, LiveIntervals &LIS) const {
+  SmallVector<MachineOperand *, 4> Defs;
+  SmallVector<MachineOperand *, 4> Uses;
+  for (MachineOperand &MO : MRI.reg_nodbg_operands(Reg)) {
```
martien-de-jong: I think we could collect all defs and uses of all candidates in a parallel sweep over the MBB's instructions. The defs would be added in order and uses and defs could not be in a different block.

llvm/lib/Target/AIE/AIEEpilogueRegRewriter.cpp
```diff
@@ -66,9 +73,234 @@ class AIEEpilogueRegRewriter : public MachineFunctionPass {
...
+    if (MO.getParent()->getParent() != &MBB)
+      continue;
+    if (MO.isDef())
+      Defs.push_back(&MO);
+    else if (MO.isUse())
+      Uses.push_back(&MO);
+  }
+  // A WAR hazard only exists if the register is both read and written here.
+  if (Defs.empty() || Uses.empty())
+    return std::nullopt;
```
martien-de-jong: The parallel scan would have to postpone this check.

llvm/lib/Target/AIE/AIEEpilogueRegRewriter.cpp
```diff
@@ -66,9 +73,234 @@ class AIEEpilogueRegRewriter : public MachineFunctionPass {
...
+    if (!CopyCost || *CopyCost > EpilogueCopyBudget - Spent) {
+      LLVM_DEBUG(dbgs() << "  skip "
+                        << printReg(Candidate.OldReg, &TRI, 0, &MRI)
+                        << ": copy cost does not fit the remaining budget\n");
+      continue;
+    }
+
+    // Block replaced registers from being selected again.
+    for (MCRegUnit Unit : TRI.regunits(NewPhys))
+      ReservedRegUnits.set(Unit);
```
martien-de-jong: You see? we can already reuse that proposed utility.

llvm/lib/Target/AIE/AIEEpilogueRegRewriter.cpp
```diff
@@ -66,9 +73,234 @@ class AIEEpilogueRegRewriter : public MachineFunctionPass {
...
+    if (!Overrides.get(EnableOLPWarRename))
+      continue;
+
+    SmallSet<Register, 16> SeenRegs;
+    for (MachineInstr &MI : MBB) {
+      for (const MachineOperand &Def : MI.defs()) {
+        if (!Def.getReg().isVirtual())
+          continue;
+
+        Register Reg = Def.getReg();
```
martien-de-jong: reuse above?

llvm/lib/Target/AIE/AIEEpilogueRegRewriter.cpp
```diff
@@ -66,9 +73,234 @@ class AIEEpilogueRegRewriter : public MachineFunctionPass {
...
+
+  return Candidates;
+}
+
+MCPhysReg AIEEpilogueRegRewriter::findReplacementPhysReg(
+    const RewriteCandidate &Candidate, MachineRegisterInfo &MRI,
+    const AIEBaseRegisterInfo &TRI, VirtRegMap &VRM, LiveIntervals &LIS,
+    LiveRegMatrix &LRM, BitVector &ReservedRegUnits) const {
+  // Repair copy slot isn't allocated yet; MBB end is a stable, conservative
+  // superset of the true [FirstDef, copy] range the new reg will occupy.
```
martien-de-jong: First define what Boundary represents, then say we start with an upper bound.

llvm/lib/Target/AIE/AIEEpilogueRegRewriter.cpp
```diff
@@ -66,9 +73,234 @@ class AIEEpilogueRegRewriter : public MachineFunctionPass {
...
+  const SlotIndex Boundary = LIS.getMBBEndIdx(Candidate.MBB);
+  const MachineOperand *FirstDefMO = Candidate.Defs.front();
+  const SlotIndex FirstDef = LIS.getInstructionIndex(*FirstDefMO->getParent())
+                                 .getRegSlot(FirstDefMO->isEarlyClobber());
+
+  // Defs live as one range up to the repair copy, and interference only
+  // consults the main range, so per-def subranges cannot change the result.
```
martien-de-jong: This comment is an implementation detail . I would first like to know the purpose of the code.

llvm/lib/Target/AIE/AIEEpilogueRegRewriter.cpp
```diff
@@ -66,9 +73,234 @@ class AIEEpilogueRegRewriter : public MachineFunctionPass {
...
+  LiveInterval ProspectiveLI(Candidate.OldReg, 0.0F);
+  VNInfo *MainValue =
+      ProspectiveLI.getNextValue(FirstDef, LIS.getVNInfoAllocator());
+  ProspectiveLI.addSegment(LiveRange::Segment(FirstDef, Boundary, MainValue));
+
+  LRM.invalidateVirtRegs();
```
martien-de-jong: It is unclear to me what the scope of this is. It looks as if this has effect on LRM which might be global to the entire run for all I know.

llvm/lib/Target/AIE/AIEEpilogueRegRewriter.cpp
```diff
@@ -66,9 +73,234 @@ class AIEEpilogueRegRewriter : public MachineFunctionPass {
...
+             << " in " << Candidate.MBB->getFullName() << '\n');
+
+  SmallSet<Register, 8> RegistersToRepair;
+  RegistersToRepair.insert(Candidate.OldReg);
+  AIESuperRegUtils::repairLiveIntervals(RegistersToRepair, VRM, LRM, LIS);
+
+  LiveInterval &NewLI = LIS.createAndComputeVirtRegInterval(NewReg);
+  LIS.shrinkToUses(&NewLI);
+
+  LRM.invalidateVirtRegs();
```
martien-de-jong: As discussed. This could be done once at the end when we have change. Then it would also clearly show the scope of the action.

llvm/lib/Target/AIE/aie1/AIE1InstrInfo.cpp
```diff
@@ -101,122 +101,133 @@ void AIEInstrInfo::copyPhysReg(MachineBasicBlock &MBB,
...
-                               bool RenamableDest, bool RenamableSrc) const {
-  MachineRegisterInfo &MRI = MBB.getParent()->getRegInfo();
-  const TargetRegisterInfo &TRI = *MRI.getTargetRegisterInfo();
+                               bool /* RenamableDest */,
+                               bool /* RenamableSrc */) const {
+  const TargetRegisterInfo &TRI =
+      *MBB.getParent()->getRegInfo().getTargetRegisterInfo();
+  CopyMaterializer M(*this, TRI, MBB, MBBI, DL);
+  if (!materializeCopy(M, DstReg, SrcReg, KillSrc))
+    assert(false && "unhandled case in copyPhysReg");
```
SagarMaheshwari99: `llvm_unreachable`

llvm/lib/Target/AIE/aie2p/AIE2PInstrInfo.cpp
```diff
@@ -627,254 +627,268 @@ void AIE2PInstrInfo::copyPhysReg(MachineBasicBlock &MBB,
...
-                                 bool RenamableDest, bool RenamableSrc) const {
-  MachineRegisterInfo &MRI = MBB.getParent()->getRegInfo();
-  const TargetRegisterInfo &TRI = *MRI.getTargetRegisterInfo();
+                                 bool /* RenamableDest */,
+                                 bool /* RenamableSrc */) const {
+  const TargetRegisterInfo &TRI =
+      *MBB.getParent()->getRegInfo().getTargetRegisterInfo();
+  CopyMaterializer M(*this, TRI, MBB, MBBI, DL);
+  if (!materializeCopy(M, DstReg, SrcReg, KillSrc))
+    llvm_unreachable("unhandled case in copyPhysReg");
```
SagarMaheshwari99: can you make this error better? See aie2ps.

llvm/lib/Target/AIE/AIE2InstrInfo.cpp
```diff
@@ -488,85 +488,92 @@ void AIE2InstrInfo::copyPhysReg(MachineBasicBlock &MBB,
...
-    BuildMI(MBB, MBBI, DL, get(AIE2::MOV_mv_scl),
-            TRI.getSubReg(DstReg, AIE2::sub_l_odd))
+    M.buildInstr(get(AIE2::MOV_mv_scl), TRI.getSubReg(DstReg, AIE2::sub_l_odd))
         .addReg(TRI.getSubReg(SrcReg, AIE2::sub_l_odd),
                 getKillRegState(KillSrc));
   } else if ((AIE2::eDRegClass.contains(SrcReg)) &&
              (AIE2::eDRegClass.contains(DstReg))) {
-    copyThroughSubRegs(MBB, MBBI, DL, DstReg, SrcReg, KillSrc);
+    if (!copyThroughSubRegs(M, DstReg, SrcReg, KillSrc))
+      return false;
```
SagarMaheshwari99: you can just do `return copyThroughSubRegs(...)`

llvm/lib/Target/AIE/aie2/AIE2TargetMachine.cpp
F-Stuckmann: This is the aie base class. We may want to rename it in time.

llvm/lib/Target/AIE/AIEEpilogueRegRewriter.cpp
F-Stuckmann: No we have to do it in place otherwise we would have to manually fix LiveIntervals, since we are changing them on each step and trying to access LRM lateron in the replacement step

F-Stuckmann:
I rebased this PR on top of https://github.com/Xilinx/llvm-aie/pull/1255 , since the data driven copy materialization is a nice standalone PR.

llvm/test/CodeGen/AIE/aie2ps/end-to-end/gemm_int8_psum_0.ll
```diff
@@ -164,28 +164,26 @@ define dso_local void @gemm_int8_psum_0(
...
+; CHECK-NEXT:    vlda bmhl6, [p4], #64; vmov.d dm3, dm4
+; CHECK-NEXT:    vlda bmhh6, [p4], #64; vshuffle x10, x3, x0, r22; vmov.d dm2, dm5
+; CHECK-NEXT:    vlda bmll7, [p4], #64; vst.srs.4x dm3, s0, srssign0, [p2], #64; vshuffle x1, x5, x0, r24
+; CHECK-NEXT:    vlda bmlh7, [p4], #64; vst.srs.4x dm2, s0, srssign0, [p2], m5; vshuffle x10, x10, x0, r20
+; CHECK-NEXT:    vlda bmhl7, [p4], #64; vst.srs.4x dm1, s0, srssign0, [p2], #64; jnzd r1, r1, p5; vshuffle x1, x1, x0, r26
+; CHECK-NEXT:    vlda bmhh7, [p4], #64; vst.2d.srs.4x dm0, s0, srssign0, [p2], d2; movx srssign0, #0; vmov.d dm1, dm6 // Delay Slot 5
+; CHECK-NEXT:    nop // Delay Slot 4
+; CHECK-NEXT:    nop // Delay Slot 3
+; CHECK-NEXT:    nop // Delay Slot 2
+; CHECK-NEXT:    vmov.d dm0, dm7 // Delay Slot 1
```
F-Stuckmann, line 186: this vmov is akward

F-Stuckmann: if we lower the threshold to 3, the epilog collapses to only 16 bundles

llvm/lib/Target/AIE/aie2/AIE2TargetMachine.cpp
```diff
@@ -195,10 +196,12 @@ bool AIE2PassConfig::addRegAssignAndRewriteOptimized() {
...
```
andcarminati, line 203: Are we deallocating something? If no, we may run the new pass before the rewriter, simplifying a bit here. Double check if you need the extra greedy run here.

llvm/lib/Target/AIE/Utils/AIERegAllocationUtils.cpp
```diff
@@ -0,0 +1,49 @@
...
+
+MCPhysReg AIERegAllocationUtils::findFreeNonOverlappingPhysReg(
+    const LiveInterval &LI, const TargetRegisterClass &RC,
+    ArrayRef<MCPhysReg> AllocationOrder, const BitVector &ReservedRegUnits,
+    const TargetRegisterInfo &TRI, LiveRegMatrix &LRM) {
+  for (MCPhysReg PhysReg : AllocationOrder) {
+    if (!RC.contains(PhysReg))
+      continue;
+
+    const bool HasReservedUnit =
```
andcarminati, line 30: maybe a direct if with a comment?

llvm/lib/Target/AIE/AIEEpilogueRegRewriter.cpp
```diff
@@ -0,0 +1,334 @@
...
+static cl::opt<bool> EnableOLPWarRename(
+    "aie-olp-war-rename", cl::Hidden, cl::init(false),
+    cl::desc("Enable WAR register renaming on blocks carrying the hint"));
+
+namespace {
+
+class RewriteCandidate {
+  const MachineInstr *LastUseMI = nullptr;
+  bool HasLateUse = false;
+
```
andcarminati, line 58: How we categorize public vs non-public here? It is a bit mixed.

llvm/lib/Target/AIE/AIEEpilogueRegRewriter.cpp
```diff
@@ -0,0 +1,334 @@
...
+public:
+  Register OldReg;
+  SmallVector<MachineOperand *, 4> Defs;
+  MachineBasicBlock *MBB = nullptr;
```
andcarminati: can you infer MBB from LastUseMI->getParent()?

andcarminati: In mean, we are scoped in a MBB, right?

llvm/lib/Target/AIE/AIEEpilogueRegRewriter.cpp
```diff
@@ -0,0 +1,334 @@
...
+                    << " candidate(s) in " << MBB.getFullName() << '\n');
+
+  BitVector ReservedRegUnits(TRI->getNumRegUnits());
+  unsigned Spent = 0;
+  bool Changed = false;
+  for (RewriteCandidate &Candidate : Candidates) {
+    if (Spent >= EpilogueCopyBudget) {
+      LLVM_DEBUG(dbgs() << "  skip " << printReg(Candidate.OldReg, TRI, 0, MRI)
+                        << ": epilogue copy budget exhausted\n");
+      continue;
```
andcarminati, line 269: Break? Or are you continuing because of the debug message?

llvm/lib/Target/AIE/AIEEpilogueRegRewriter.cpp
```diff
@@ -0,0 +1,334 @@
...
+  SmallVector<Register, 16> CandidateOrder;
+
+  // Scan the block once and group each virtual register's defs and uses.
+  // A block-end repair copy is safe only when every use precedes its first def.
+  for (MachineInstr &MI : MBB) {
+    for (MachineOperand &MO : MI.operands()) {
+      if (!MO.isReg() || !MO.getReg().isVirtual())
+        continue;
+
+      Register Reg = MO.getReg();
```
andcarminati, line 154: Register should be always passed by value, so you can use const.

llvm/lib/Target/AIE/AIEEpilogueRegRewriter.cpp
```diff
@@ -0,0 +1,334 @@
...
+      RewriteCandidate &Candidate =
+          CandidatesByReg.try_emplace(Reg, Reg, MBB).first->second;
+      if (MO.isUse()) {
+        Candidate.addUse(MI);
+        continue;
+      }
+      if (!MO.isDef())
```
andcarminati, line 161: Conditions are gated on the same boolean:
```
  bool isUse() const {
    assert(isReg() && "Wrong MachineOperand accessor");
    return !IsDef;
  }
```
So you can remove this if -> continue.

llvm/lib/Target/AIE/AIEEpilogueRegRewriter.cpp
```diff
@@ -0,0 +1,334 @@
...
+// Per-block opt-in: only rename WAR definitions in a block that carries this
+// hint (see AIE::LoopOptionOverrides).
+static cl::opt<bool> EnableOLPWarRename(
+    "aie-olp-war-rename", cl::Hidden, cl::init(false),
+    cl::desc("Enable WAR register renaming on blocks carrying the hint"));
+
+namespace {
+
+class RewriteCandidate {
+  const MachineInstr *LastUseMI = nullptr;
```
andcarminati, line 56: Maybe comment a bit about the semantics of each field.

llvm/lib/Target/AIE/AIEEpilogueRegRewriter.cpp
```diff
@@ -0,0 +1,334 @@
...
+    Defs.push_back(&Def);
+    return IsFirstDef;
+  }
+
+  void addUse(MachineInstr &MI) {
+    HasLateUse |= !Defs.empty();
+    LastUseMI = &MI;
+  }
+
+  bool isSafeToRewrite() const { return LastUseMI && !HasLateUse; }
```
andcarminati, line 80: Maybe check this name. It is safety or profitability?

llvm/lib/Target/AIE/AIEEpilogueRegRewriter.cpp
```diff
@@ -0,0 +1,334 @@
...
+
+  // Scan the block once and group each virtual register's defs and uses.
+  // A block-end repair copy is safe only when every use precedes its first def.
+  for (MachineInstr &MI : MBB) {
+    for (MachineOperand &MO : MI.operands()) {
+      if (!MO.isReg() || !MO.getReg().isVirtual())
+        continue;
+
+      Register Reg = MO.getReg();
+      RewriteCandidate &Candidate =
```
andcarminati, line 155: Maybe a lambda "GetOrCreateCandidate(...)"

llvm/test/CodeGen/AIE/aie2ps/ra/epilogue-reg-rewrite-independent.mir
```diff
@@ -0,0 +1,83 @@
...
+# See https://llvm.org/LICENSE.txt for license information.
+# SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+#
+# (c) Copyright 2026 Advanced Micro Devices, Inc. or its affiliates
+
+# Two enabled epilogues may reuse a replacement physical register when their
+# candidate live ranges do not overlap.
+# RUN: llc -mtriple=aie2ps --aie-wawreg-rewrite=false --aie-olp-war-rename \
+# RUN:   --aie-enable-epilogue-reg-rewrite --start-before=greedy \
+# RUN:   --stop-after=virtregrewriter %s -o - | FileCheck %s
```
khallouh, line 12: Q: Why do we need to pass 2 different options to enable?

llvm/test/CodeGen/AIE/aie2ps/ra/epilogue-reg-rewrite-independent.mir
```diff
@@ -0,0 +1,83 @@
+# NOTE: Assertions have been autogenerated by utils/update_mir_test_checks.py UTC_ARGS: --version 5
+# This file is licensed under the Apache License v2.0 with LLVM Exceptions.
+# See https://llvm.org/LICENSE.txt for license information.
+# SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+#
+# (c) Copyright 2026 Advanced Micro Devices, Inc. or its affiliates
+
+# Two enabled epilogues may reuse a replacement physical register when their
```
khallouh, line 8: 2 epilogues? I see no loop here

llvm/test/CodeGen/AIE/aie2ps/ra/epilogue-reg-rewrite-same-instr.mir
```diff
@@ -0,0 +1,45 @@
...
+# See https://llvm.org/LICENSE.txt for license information.
+# SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+#
+# (c) Copyright 2026 Advanced Micro Devices, Inc. or its affiliates
+
+# A definition that also reads its old value cannot be renamed: the repair
+# copy is inserted only after the instruction has executed.
+# RUN: llc -mtriple=aie2ps --aie-wawreg-rewrite=false --aie-olp-war-rename \
+# RUN:   --aie-enable-epilogue-reg-rewrite --start-before=greedy \
+# RUN:   --stop-after=aie-epilogue-reg-rewriter %s -o - | FileCheck %s
```
khallouh, line 12: These unit tests all disable aie-wawreg-rewrite but in the production compiler it is still enabled, can we test that behavior instead?

llvm/test/CodeGen/AIE/aie2ps/end-to-end/gemm_int8_psum_0.ll
khallouh: should we lower it? why was the threshold set to 4 initially?

khallouh: as discussed, maybe we could decide this dynamically, so this pass could be always beneficial and we can enable it by default instead of via hint. Since I think we have the same problem in most outer-loop pipelines

F-Stuckmann: for now the pass is only enabled by pragma, so we don't need any complicated cost functions.

llvm/lib/Target/AIE/Utils/AIERegAllocationUtils.cpp
```diff
@@ -0,0 +1,47 @@
...
+MCPhysReg AIERegAllocationUtils::findFreeNonOverlappingPhysReg(
+    const LiveInterval &LI, const TargetRegisterClass &RC,
+    ArrayRef<MCPhysReg> AllocationOrder, const BitVector &ReservedRegUnits,
+    const TargetRegisterInfo &TRI, LiveRegMatrix &LRM) {
+  for (MCPhysReg PhysReg : AllocationOrder) {
+    if (!RC.contains(PhysReg))
+      continue;
+
+    if (llvm::any_of(TRI.regunits(PhysReg), [&](MCRegUnit Unit) {
+          return !ReservedRegUnits.empty() && ReservedRegUnits.test(Unit);
```
martien-de-jong, line 31: I'm wondering whether the precheck for empty speeds things up.

llvm/lib/Target/AIE/AIEEpilogueRegRewriter.cpp
```diff
@@ -29,13 +34,16 @@
...
 
 #define DEBUG_TYPE "aie-epilogue-reg-rewriter"
 
-static cl::opt<unsigned> EpilogueCopyBudget(
-    "aie-epilogue-reg-rewrite-copy-budget", cl::Hidden, cl::init(4),
+static constexpr unsigned DefaultLongLatencyThreshold = 3;
+static cl::opt<unsigned> LongLatencyThreshold(
+    "aie-epilogue-reg-rewrite-long-latency-threshold", cl::Hidden,
+    cl::init(DefaultLongLatencyThreshold),
     cl::desc("Maximum materialized copy instructions per epilogue"));
```
martien-de-jong: I can't tie the description to the name. What's the unit? latency cycles? instruction count?

llvm/lib/Target/AIE/AIEEpilogueRegRewriter.cpp
```diff
@@ -46,32 +54,277 @@ static cl::opt<bool> EnableOLPWarRename(
...
+  // A block-end repair copy is safe only when every use precedes its first def.
+  for (MachineInstr &MI : MBB) {
+    for (MachineOperand &MO : MI.operands()) {
+      if (!MO.isReg() || !MO.getReg().isVirtual())
+        continue;
+
+      const Register Reg = MO.getReg();
+      RewriteCandidate &Candidate = GetOrCreateCandidate(Reg);
+      if (MO.isUse()) {
+        Candidate.addUse(MI);
```
martien-de-jong: nit: the addDef and addUse look complementary, but one adds an operand, the other an instruction. I could imagine addUse following the parent to the instruction or renaming to addUseMI.

llvm/lib/Target/AIE/AIEEpilogueRegRewriter.cpp
```diff
@@ -46,32 +54,277 @@ static cl::opt<bool> EnableOLPWarRename(
...
+        continue;
+      }
+
+      if (Candidate.addDef(MO))
+        CandidateOrder.push_back(Reg);
+    }
+  }
+
+  std::vector<RewriteCandidate> Candidates;
+  for (Register Reg : CandidateOrder) {
```
martien-de-jong: Why don't we list the Candidates themselves rather than their Regs?

llvm/lib/Target/AIE/AIEEpilogueRegRewriter.cpp
```diff
@@ -46,32 +54,277 @@ static cl::opt<bool> EnableOLPWarRename(
...
+    if (!NewPhys) {
+      LLVM_DEBUG(dbgs() << "  skip "
+                        << printReg(Candidate.oldReg(), TRI, 0, MRI)
+                        << ": no free non-overlapping physical register\n");
+      continue;
+    }
+
+    const std::optional<unsigned> CopyCost =
+        TII->getCopyCost(VRM->getPhys(Candidate.oldReg()), NewPhys);
+    if (!CopyCost || *CopyCost > LongLatencyThreshold - Spent) {
```
martien-de-jong: Why not maintain RemainingBudget in stead of Spent?

llvm/test/CodeGen/AIE/aie2ps/end-to-end/gemm_int8_psum_0.ll
```diff
@@ -810,7 +804,7 @@ attributes #3 = { nocallback nofree nosync nounwind willreturn memory(inaccessib
 !201 = distinct !{!201, !"_ZN3aie6detail11load_vectorILj16EL15aie_dm_resource0EiEEDaPKT1_"}
 !202 = distinct !{!202, !203, !"_ZN3aie6load_vILj16EL15aie_dm_resource0ETkNS_21DecoratedElemBaseTypeEiEENS_6vectorIN22aie_dm_resource_removeIT1_E4typeEXT_EEEPKS4_: %agg.result"}
 !203 = distinct !{!203, !"_ZN3aie6load_vILj16EL15aie_dm_resource0ETkNS_21DecoratedElemBaseTypeEiEENS_6vectorIN22aie_dm_resource_removeIT1_E4typeEXT_EEEPKS4_"}
-!204 = distinct !{!204, !205, !206}
+!204 = distinct !{!204, !205, !206, !237}
```
martien-de-jong, line 807: Add this in the original test?

llvm/test/CodeGen/AIE/aie2ps/ra/epilogue-reg-rewrite.mir
```diff
@@ -215,27 +220,29 @@ body:             |
...
+  ; PHYS-NEXT:   renamable $bmhl2, renamable $p0 = VLDA_dmx_lda_bm_ld_pstm_nrm_imm killed renamable $p0, 64
+  ; PHYS-NEXT:   renamable $bmhh2, dead renamable $p0 = VLDA_dmx_lda_bm_ld_pstm_nrm_imm killed renamable $p0, 64
+  ; PHYS-NEXT:   renamable $dm0 = COPY killed renamable $dm7
+  ; PHYS-NEXT:   renamable $dm5 = COPY killed renamable $dm1
+  ; PHYS-NEXT:   renamable $dm6 = COPY killed renamable $dm2
   ; PHYS-NEXT:   PseudoJ_jump_imm %bb.3
   ; PHYS-NEXT: {{  $}}
   ; PHYS-NEXT: bb.3:
-  ; PHYS-NEXT:   liveins: $dm0:0x000000030000000C, $dm5:0x000000030000000C, $dm6:0x000000030000000C, $r8, $x0, $x2
+  ; PHYS-NEXT:   liveins: $dm0, $dm5, $dm6, $r8, $x0, $x2
```
martien-de-jong, line 245: I assume this is because we lost the live lanes because we have full-reg copies now. The lane lane masks also cover the full register (I hope)

llvm/lib/Target/AIE/Utils/AIERegAllocationUtils.cpp
F-Stuckmann: This is an assert check, WAWRegRewriter may call findFreeNonOverlappingPhysReg with an empty bitvector and without the empty check we will run into an assert.

llvm/lib/Target/AIE/AIEEpilogueRegRewriter.cpp
```diff
@@ -46,32 +54,277 @@ static cl::opt<bool> EnableOLPWarRename(
...
+  DenseMap<Register, RewriteCandidate> CandidatesByReg;
+  SmallVector<Register, 16> CandidateOrder;
+  auto GetOrCreateCandidate = [&](Register Reg) -> RewriteCandidate & {
+    return CandidatesByReg.try_emplace(Reg, Reg, MBB).first->second;
+  };
+
+  // Scan the block once and group each virtual register's defs and uses.
+  // A block-end repair copy is safe only when every use precedes its first def.
+  for (MachineInstr &MI : MBB) {
+    for (MachineOperand &MO : MI.operands()) {
```
martien-de-jong: I have a feeling a lot of the magic in addDef and addUse could be eliminated if we split this in a loop over the uses (which always precede the defs) followed by a loop over the defs. Perhaps even scan backwards, because we then don't have to push defs for which we have already seen a use.

llvm/lib/Target/AIE/AIEEpilogueRegRewriter.cpp
```diff
@@ -46,32 +53,257 @@ static cl::opt<bool> EnableOLPWarRename(
...
+      GetOrCreateCandidate(MO.getReg()).Defs.push_back(&MO);
+    }
+
+    for (const MachineOperand &MO : MI.all_uses()) {
+      if (!MO.getReg().isVirtual())
+        continue;
+      RewriteCandidate &Candidate = GetOrCreateCandidate(MO.getReg());
+      Candidate.HasUse = true;
+      if (Candidate.Defs.empty())
+        Candidate.UseAfterDef = true;
```
martien-de-jong: I think we just have to record the use, and reject the def when we find the def in the reversed scan.

martien-de-jong (approved):

PR #1191, opened 2026-07-22
[AIEX] do not drop first metadata when updating metadata.

F-Stuckmann:
Do not drop first metadata when updating metadata in OLP and when updating Itercounts.


andcarminati (dismissed):
LGTM. Easy to stamp! Thanks for fixing this!

andcarminati (approved):

PR #1196, opened 2026-07-23
Add few types of widening to narrowing cast folding

nikolaos-amd:


mlir/lib/Dialect/Tosa/IR/TosaCanonicalizations.cpp
```diff
@@ -1720,6 +1720,54 @@ OpFoldResult EqualOp::fold(FoldAdaptor adaptor) {
...
+    // Only fold when the inner cast is strictly widening (the intermediate
+    // type must have more bits than the source type). This guarantees the
+    // first cast is information-preserving regardless of type categories.
+    // Same bit-width casts between different types (e.g., bf16 <-> f16, or
+    // same-sized float <-> int) are not information-preserving and must not
+    // be folded through.
+    unsigned srcBitWidth = srcElemTy.getIntOrFloatBitWidth();
+    unsigned midBitWidth = midElemTy.getIntOrFloatBitWidth();
+    if (midBitWidth <= srcBitWidth)
+      return failure();
```
jorickert: This check is both not strict enough (doing wrong folds) and also rejects valid folds. Only doing it on the bitwdith is not safe.

For example it will wrongly fold bf16 -> i32 -> fp32 to bf16 -> fp32, but it will reject the safe i8 -> i32 -> i8.

I would take some inspiration from `CastOp::fold` to check if the folding is valid or not.

Please also add tests for these cases

jorickert (changes requested):

mlir/lib/Dialect/Tosa/IR/TosaCanonicalizations.cpp
```diff
@@ -1720,6 +1720,54 @@ OpFoldResult EqualOp::fold(FoldAdaptor adaptor) {
...
+    if (srcElemTy == dstElemTy)
+      return failure();
+
+    // Only fold when the inner cast is strictly widening (the intermediate
+    // type must have more bits than the source type). This guarantees the
+    // first cast is information-preserving regardless of type categories.
+    // Same bit-width casts between different types (e.g., bf16 <-> f16, or
+    // same-sized float <-> int) are not information-preserving and must not
+    // be folded through.
+    unsigned srcBitWidth = srcElemTy.getIntOrFloatBitWidth();
```
jorickert: you need to check that its an int or float first, tosa allows quant types

mlir/lib/Dialect/Tosa/IR/TosaCanonicalizations.cpp
nikolaos-amd: I know of these cases and how they should be handled. Let me check if I did a change that broke them and add the test. Agreed that otherwise not strict enough.

mlir/lib/Dialect/Tosa/IR/TosaCanonicalizations.cpp
nikolaos-amd: thanks, I am rejecting those now

mlir/lib/Dialect/Tosa/IR/TosaCanonicalizations.cpp
```diff
@@ -1720,6 +1720,86 @@ OpFoldResult EqualOp::fold(FoldAdaptor adaptor) {
...
+
+    if (bothInt) {
+      // int -> int: require same signedness and strict widening.
+      if (srcIntTy.getSignedness() != midIntTy.getSignedness())
+        return failure();
+      if (midIntTy.getWidth() <= srcIntTy.getWidth())
+        return failure();
+    } else if (bothFloat) {
+      // float -> float: require strict widening (by total bit width).
+      if (midFloatTy.getWidth() <= srcFloatTy.getWidth())
```
jorickert: this check is not generally strict enough (but i think should be fine as long as tosa doe snot add more float types):
```
M.precision >= S.precision
M.minExponent <= S.minExponent
M.maxExponent >= S.maxExponent
``` 
should be safer

jorickert (dismissed):

mlir/test/Dialect/Tosa/tosa-canonicalizations.mlir
```diff
@@ -60,4 +64,189 @@ func.func @test_cast_lossy_conversion(%arg0: tensor<1x3x3xf32>) -> tensor<1x3x3x
...
+  %1 = tosa.cast %0 : (tensor<4xi64>) -> tensor<4xi8>
+  return %1 : tensor<4xi8>
+}
+// CHECK-SAME:  (%[[ARG0:.*]]: tensor<4xi1>) -> tensor<4xi8>
+// CHECK-NEXT:  tosa.cast %[[ARG0]] : (tensor<4xi1>) -> tensor<4xi8>
+// CHECK-NEXT:  return {{.*}} : tensor<4xi8>
+
+// -----
+
+// i8 -> i1 -> f16 is not folded because its the first conversion is a narrowing one.
```
ttjost, line 99: what about `i64 -> i32 -> i16`?

ttjost: I know we don't need. Just thought of mentioning 😄

nikolaos-amd: existing patterns handle the similar cases, I added one testcase for clarity.

jorickert (approved):

PR #1195, opened 2026-07-23
[AIE2PS] update Vec Regs for WAWRegRewriter

F-Stuckmann:
cherry-picked vec reg widing from @SagarMaheshwari99  and added missing reg classes.

PR #1198, opened 2026-07-24
[AIE2PS] Add BFP16 vectors to be eligible for register renaming.

SagarMaheshwari99:


F-Stuckmann (approved):
this will drop the two cherry-picks in https://github.com/Xilinx/llvm-aie/pull/1195

PR #1200, opened 2026-07-24
[AIEX] Loop hint template Evaluation

F-Stuckmann:
If template arguments are used in loop hints, we have to transform them once the template gets instantiated otherwise we crash.

### DIFF
--- a/clang/test/CodeGen/aie/generic-loop-hint-template-errors.cpp
+++ b/clang/test/CodeGen/aie/generic-loop-hint-template-errors.cpp
@@ -9,16 +9,18 @@
 //
 //===----------------------------------------------------------------------===//
 
-// Reproduce the crash: malformed dependent generic loop hint values stay
-// dependent without the fix and the evaluator asserts instead of diagnosing.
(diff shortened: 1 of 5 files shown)


andcarminati (approved):
LGTM.

clang/test/CodeGen/aie/generic-loop-hint-template-errors.cpp
```diff
@@ -0,0 +1,47 @@
...
+//
+// Modifications (c) Copyright 2026 Advanced Micro Devices, Inc. or its
+// affiliates
+//
+//===----------------------------------------------------------------------===//
+
+// Reproduce the crash: malformed dependent generic loop hint values stay
+// dependent without the fix and the evaluator asserts instead of diagnosing.
+
+// RUN: not --crash %clang_cc1 -triple aie2ps -std=c++20 -emit-llvm -o - %s 2>&1 | FileCheck %s --check-prefix=CRASH
```
martien-de-jong: I never understood this not --crash construct. lit should always report crashes as failures. I typically label crash status quos as XFAIL which is removed with the fix.

PR #1202, opened 2026-07-28
[AIE] Add speculative outer-loop pipelining

khallouh:
This supercedes https://github.com/Xilinx/llvm-aie/pull/1189


llvm/lib/Target/AIE/AIEOuterLoopPipeliner.cpp
```diff
@@ -1732,7 +1767,8 @@ bool AIEOuterLoopPipeliner::performTransformation(
     return SplitMode && isStage1SplitPoint(I);
   };
 
-  if (Overrides.get(LeanStage0Mode))
+  const bool LeanStage0 = Overrides.get(LeanStage0Mode);
```
andcarminati: nit: `useLeanStage0`.

llvm/lib/Target/AIE/AIEOuterLoopPipeliner.cpp
```diff
@@ -1741,6 +1777,20 @@ bool AIEOuterLoopPipeliner::performTransformation(
     return false;
   }
 
+  // Lean stage-0 defaults to speculation, regardless of side effects in that
```
andcarminati, line 1807: Do you think it is possible to have a something like a class where we take all configurations at once? 

like:

```
class OLPOpts {
    bool UseHardwareLoop;
    bool UseSpeculativeLastIterarion;
    bool UseLeanStage0Mode;
    .....
}
```

I am suggesting this because the number of options is growing, and we could centralize the configuration.

khallouh: I like this. It's done.

andcarminati:
When a pointer update (GEP) feeds an outer header PHI AND is used outside the loop, speculative mode may compute it one
extra time if the GEP is lifted to the top block. I will send to you a test case.

llvm/lib/Target/AIE/AIEOuterLoopPipeliner.cpp
```diff
@@ -1741,6 +1799,17 @@ bool AIEOuterLoopPipeliner::performTransformation(
...
+  // stage. Non-lean speculation still requires a side-effect-free stage-0
+  // chain.
+  const bool Stage0IsSideEffectFree =
+      llvm::none_of(OrigLS.stage0Insts(), [](const Instruction *I) {
+        return I->mayHaveSideEffects();
```
andcarminati: Could we have a helper function where we call return I->mayHaveSideEffects();? It will be useful to inject a common logic, like the upcoming TII hook to unblock gemm.

khallouh:
> When a pointer update (GEP) feeds an outer header PHI AND is used outside the loop, speculative mode may compute it one
> extra time if the GEP is lifted to the top block.

Actually, there is no bug in speculative mode. On the contrary, speculative would save us from hitting another latent bug. 
The ptr update would not be executed one extra time because it is in stage1. The actual bug is Independent of speculative. since we only follow SSA edges for stage0 and collect stage1 from the anchor, if there is no anchor, stage1 will remain empty and the lifted ptr update would not be copied to the bottom region so the remapping will resolve to poison.
So this is unrelated but I will fix it in this PR for convenience.

llvm/lib/Target/AIE/AIEOuterLoopPipeliner.cpp
khallouh: Done

khallouh:
> > When a pointer update (GEP) feeds an outer header PHI AND is used outside the loop, speculative mode may compute it one
> > extra time if the GEP is lifted to the top block.
> 
> Actually, there is no bug in speculative mode. On the contrary, speculative would save us from hitting another latent bug. The ptr update would not be executed one extra time because it is in stage1. The actual bug is Independent of speculative. since we only follow SSA edges for stage0 and collect stage1 from the anchor, if there is no anchor, stage1 will remain empty and the lifted ptr update would not be copied to the bottom region so the remapping will resolve to poison. So this is unrelated but I will fix it in this PR for convenience.

The bug fix changes the Conv2D mx6 tests. I don't know if it was passing before with OLP enabled. Since it is not related to the speculative feature, I dropped the 2 commits opened them here [[AIE] Preserve lifted pointer updates in peeled iterations by khallouh · Pull Request #1206 · Xilinx/llvm-aie](https://github.com/Xilinx/llvm-aie/pull/1206)

khallouh:
> > > When a pointer update (GEP) feeds an outer header PHI AND is used outside the loop, speculative mode may compute it one
> > > extra time if the GEP is lifted to the top block.
> > 
> > 
> > Actually, there is no bug in speculative mode. On the contrary, speculative would save us from hitting another latent bug. The ptr update would not be executed one extra time because it is in stage1. The actual bug is Independent of speculative. since we only follow SSA edges for stage0 and collect stage1 from the anchor, if there is no anchor, stage1 will remain empty and the lifted ptr update would not be copied to the bottom region so the remapping will resolve to poison. So this is unrelated but I will fix it in this PR for convenience.
> 
> The bug fix changes the Conv2D mx6 tests. I don't know if it was passing before with OLP enabled. Since it is not related to the speculative feature, I dropped the 2 commits opened them here [[AIE] Preserve lifted pointer updates in peeled iterations by khallouh · Pull Request #1206 · Xilinx/llvm-aie](https://github.com/Xilinx/llvm-aie/pull/1206)

Update: the bug fix is not responsible for the mx6 test updates, aie-public was just broken due to automerge, this should fix it [[AIE2PS] update check lines to fix CI by khallouh · Pull Request #1207 · Xilinx/llvm-aie](https://github.com/Xilinx/llvm-aie/pull/1207)

llvm/lib/Target/AIE/AIEOuterLoopPipeliner.cpp
```diff
@@ -1374,6 +1386,29 @@ void AIEOuterLoopPipeliner::wireLastIterIntoCFG(
...
+    const OrigLoopStructure &OrigLS, const CloneLoopStructure &SteadyLS) const {
+  BasicBlock *OrigExit = OrigLS.getExitBlock();
+  reroutePhiIncomings(OrigExit, OrigLS.getBottom(), SteadyLS.getBottom(),
+                      PhiEdge::Repoint,
+                      [&](Value *V) { return SteadyLS.cloneOf(V); });
+
+  // See wireLastIterIntoCFG: exit instructions may directly reference values
+  // defined in the original loop and must instead use their steady clones
+  // before the original is deleted.
+  for (Instruction &I : *OrigExit) {
```
andcarminati: Nit: Can we have just?
```
for (Instruction &I : make_range(OrigExit->getFirstNonPHI()->getIterator(),
                                  OrigExit->end())) {
  for (Use &Op : I.operands()) {
    Value *V = Op.get();
    if (Value *Mapped = ReplacementLS.cloneOf(V); Mapped != V)
      Op.set(Mapped);
  }
}
```
`if (!isa<Instruction>(V) && !isa<Argument>(V))` is redundant, we are not saving much in terms of hash...

llvm/lib/Target/AIE/AIEOuterLoopPipeliner.cpp
```diff
@@ -1760,37 +1831,52 @@ bool AIEOuterLoopPipeliner::performTransformation(
...
-    convertOuterLoopToHardwareLoop(SteadyLS);
-
-  // Create the last-iteration region from the steady loop.
-  peelLastIteration(OrigLS, SteadyLS);
-
-  // Adjust itercount metadata to reflect the reduced trip count.
-  SteadyLS.updateLoopMetadata();
+      Opts.UseHardwareLoop && SteadyLS.latchCondition().isDowncounting();
+
+  if (!UseSpeculativeLastIteration) {
```
andcarminati: This is very subtle, but I leave it to your taste to decide:
```
// Preparation: set up the latch condition for JNZD and/or peeling.
// - Non-speculative: adjust the loop bound from N to N-1 (peel one iteration)
// - Speculative: remap exit live-outs to SteadyLS before JNZD erases the IV
if (!UseSpeculativeLastIteration)
  SteadyLS.adjustLoopBound();
else
  remapExitForReplacement(OrigLS, SteadyLS);

// Convert to a JNZD hardware loop if applicable. Must run after bound
// adjustment (non-speculative) or exit remapping (speculative), and before
// peeling so the counting add/icmp are erased before the peel clones them.
if (OuterIsHardwareLoop)
  convertOuterLoopToHardwareLoop(SteadyLS);

// Finalization: complete the transformation and update metadata.
if (!UseSpeculativeLastIteration) {
  // Peel the last iteration into a separate region (no prefetch loads).
  peelLastIteration(OrigLS, SteadyLS);
  // Adjust itercount metadata to reflect the reduced trip count (N-1).
  SteadyLS.updateLoopMetadata();
} else {
  // Mark the loop as speculatively pipelined (no iteration peeled).
  SteadyLS.markSpeculativePipelining();
  LLVM_DEBUG(dbgs() << "    Speculative last iteration: no last-iteration "
                       "region\n");
}
```

andcarminati (approved):
LGTM. Nice work and thanks for addressing the comments.

PR #1203, opened 2026-07-28
[AIE2PS] Use vmov.d to copy full accumulators

andcarminati:
This helps some GEMM variants. We have similar implementation in aie2p.

### DIFF
--- a/llvm/lib/Target/AIE/aie2ps/AIE2PSInstrInfo.cpp
+++ b/llvm/lib/Target/AIE/aie2ps/AIE2PSInstrInfo.cpp
@@ -436,10 +436,8 @@ void AIE2PSInstrInfo::copyPhysReg(MachineBasicBlock &MBB,
                 getHiSubReg(TRI, SrcReg), KillSrc);
   } else if ((AIE2PS::ACC2048RegClass.contains(SrcReg)) &&
              (AIE2PS::ACC2048RegClass.contains(DstReg))) {
-    copyPhysReg(MBB, MBBI, DL, getLoSubReg(TRI, DstReg),
-                getLoSubReg(TRI, SrcReg), KillSrc);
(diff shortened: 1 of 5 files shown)


khallouh:
Is the optimal II reached in the tests due to using 1 instruction instead of 2 previously or the delay property is also needed?

khallouh:
Is it worth it to run QoR for this? Could we have regressions with the delayed instruction?

llvm/lib/Target/AIE/aie2ps/AIE2PSInstrInfo.cpp
```diff
@@ -436,10 +436,8 @@ void AIE2PSInstrInfo::copyPhysReg(MachineBasicBlock &MBB,
                 getHiSubReg(TRI, SrcReg), KillSrc);
   } else if ((AIE2PS::ACC2048RegClass.contains(SrcReg)) &&
              (AIE2PS::ACC2048RegClass.contains(DstReg))) {
-    copyPhysReg(MBB, MBBI, DL, getLoSubReg(TRI, DstReg),
-                getLoSubReg(TRI, SrcReg), KillSrc);
-    copyPhysReg(MBB, MBBI, DL, getHiSubReg(TRI, DstReg),
-                getHiSubReg(TRI, SrcReg), KillSrc);
+    BuildMI(MBB, MBBI, DL, get(AIE2PS::VMOV_D_vmov), DstReg)
```
F-Stuckmann, line 439: Let's see how this interacts with https://github.com/Xilinx/llvm-aie/pull/1190

andcarminati:
> Is it worth it to run QoR for this? Could we have regressions with the delayed instruction?

Here we go:
image
 
The 2 new failures mentioned are just timeouts, the code is the same as before, no vmov.d inserted.

I seem some marginal PM reduction as well.

khallouh (dismissed):
LGTM

khallouh (approved):

PR #1206, opened 2026-07-29
[AIE][OLP] Preserve lifted pointer updates in peeled iterations

khallouh:


llvm/test/CodeGen/AIE/aie2p/outer-loop-pipeliner/outer-loop-pipelining-speculative-ptr-liveout.ll
```diff
@@ -0,0 +1,143 @@
+; NOTE: Assertions have been autogenerated by utils/update_mir_test_checks.py UTC_ARGS: --version 5
```
andcarminati: We have some mir tests for the OLP and this is not a good practice. IR passes should be testes as IR ->IR (except e2e). We should avoid using CodeGen layer to test IR transformations. The down side that we cannot use the update scripts..

khallouh: Done

andcarminati (approved):
LGTM. Thanks for fixing this.

PR #1209, opened 2026-07-29
[AIE2P] Use indexed addressing for non-64-aligned ePSRFLdF spills

### LINKED ISSUE #1090: [AIE2P] Compiling a BFP16 matrix-multiplication kernel for AIE2P fails in the AIE2 assembly printer
## Summary

Compiling a BFP16 matrix-multiplication kernel for AIE2P with Peano/LLVM-AIE
fails in the AIE2 assembly printer:

```text
fatal error: error in backend:
immediate operand value -56 is not a multiple of 64
```

LLVM generates `vlda`/`vst` instructions that spill and reload the FIFO state
register `lfl0` using stack offset `-56`. These instructions require an
immediate divisible by 64, so the generated machine instruction cannot be
encoded.

This is a compilation-only failure; running on hardware is not required to
reproduce it.

## Environment

- Target: `aie2p-none-unknown-elf`
- LLVM-AIE version: `21.0.0`
- LLVM-AIE commit: `cb664e8cc3eb42a12e3ad3cee28729785ffa97a3`
- Optimization level: `-O2`
- Language mode: C++20
- BFP16 emulation enabled with:
  `AIE_API_EMULATE_BFLOAT16_MMUL_WITH_BFP16`

Compiler version:

```text
clang version 21.0.0
Target: aie2p-none-unknown-elf
Thread model: posix
InstalledDir: <llvm-aie-install>/bin
Build config: +assertions
```

## Compilation command

The failure can be reproduced with a command equivalent to:

```shell
clang++ \
  -O2 \
  -std=c++20 \
  --target=aie2p-none-unknown-elf \
  -Wno-parentheses \
  -Wno-attributes \
  -Wno-macro-redefined \
  -DNDEBUG \
  -I <mlir-aie-install>/include \
  -DAIE_API_EMULATE_BFLOAT16_MMUL_WITH_BFP16 \
  -DBIT_WIDTH=8 \
  -c mm_bfp.cc \
  -o mm_bfp.o
```

## Source pattern that triggers the failure

The affected implementation first converts BF16 input blocks into a local
BFP16 buffer:

```cpp
template <unsigned rowA, unsigned colA, unsigned colB, unsigned r, unsigned s,
          unsigned t, typename T_out>
void matmul_vectorized_2x2_bfp16_bf16_new(const bfloat16 *__restrict pA,
                                      const bfp16ebs8 *__restrict pB,
                                      T_out *__restrict pC) {
  const unsigned sizeA = r * s;
  const unsigned sizeB = s * t;
  const unsigned sizeC = r * t;
  static_assert(rowA % 2 == 0);
  static_assert(sizeA == 64);
  static_assert(sizeB == 64);
  static_assert(sizeC == 64);

  alignas(aie::vector_decl_align)
bfp16ebs8 converted_A[rowA * colA * sizeA / 8];

aie::block_vector_output_buffer_stream<bfp16ebs8, 64>
    pA_bfp16_out(converted_A);

pA_bfp16_out.seek(0);

// Store A in this order:
// row z/K0, row z+1/K0, row z/K1, row z+1/K1, ...
for (unsigned z = 0; z < rowA; z += 2) {
  for (unsigned i = 0; i < colA; ++i) {
    for (unsigned inner = 0; inner < 2; ++inner) {
      const bfloat16 *pA_block =
          pA + ((z + inner) * colA + i) * sizeA;

      aie::vector<bfloat16, 64> A_bf16 =
          aie::load_v<64>(pA_block);

      aie::accum<accfloat, 64> A_float;
      A_float = A_bf16;

      pA_bfp16_out.push(
          A_float.template to_vector<bfp16ebs8>());
    }
  }
}
  for (unsigned z = 0; z < rowA; z += 2) //rowA = 4
chess_prepare_for_pipelining chess_loop_range(rowA / 2, )
 {
    T_out *__restrict pC1 = pC + (z * colB + 0) * sizeC;
    T_out *__restrict pC2 = pC + ((z + 1) * colB + 0) * sizeC;

    for (unsigned j = 0; j < colB; j += 2) //colB = 16

      chess_flatten_loop 
      {

        aie::block_vector_input_buffer_stream<bfp16ebs8, 64> pA_bfp16(converted_A);
        pA_bfp16.seek(z * colA);

        aie::block_vector_input_buffer_stream<bfp16ebs8, 64> pB1bfp16(pB);
        aie::block_vector_input_buffer_stream<bfp16ebs8, 64> pB2bfp16(pB);
        // For non transposed matrix
        pB1bfp16.seek(j);
        pB2bfp16.seek(j + 1);
        aie::block_vector<bfp16ebs8, sizeB> B0;
        aie::block_vector<bfp16ebs8, sizeB> B1;

        aie::accum<accfloat, sizeC> accC00(aie::load_v<sizeC>(pC1));
        aie::accum<accfloat, sizeC> accC01(aie::load_v<sizeC>(pC1 + sizeC));
        aie::accum<accfloat, sizeC> accC10(aie::load_v<sizeC>(pC2));
        aie::accum<accfloat, sizeC> accC11(aie::load_v<sizeC>(pC2 + sizeC));



        for (unsigned i = 0; i < colA; ++i) //colA = 8
chess_prepare_for_pipelining chess_loop_range(colA, )
          {

            aie::block_vector<bfp16ebs8, 64> A0;
            aie::block_vector<bfp16ebs8, 64> A1;
            A0 = pA_bfp16.pop();
            A1 = pA_bfp16.pop();
            // For non transposed matrix
            B0 = pB1bfp16.pop_seek(colB - 1);
            B1 = pB2bfp16.pop_seek(colB - 1);
    

            accC00 = mac_8x8_8x8T(A0, B0, accC00);
            accC01 = mac_8x8_8x8T(A0, B1, accC01);
            accC10 = mac_8x8_8x8T(A1, B0, accC10);
            accC11 = mac_8x8_8x8T(A1, B1, accC11);
          }

        aie::store_v(pC1, accC00.template to_vector<T_out>());
        pC1 += sizeC;
        aie::store_v(pC1, accC01.template to_vector<T_out>());
        pC1 += sizeC;
        aie::store_v(pC2, accC10.template to_vector<T_out>());
        pC2 += sizeC;
        aie::store_v(pC2, accC11.template to_vector<T_out>());
        pC2 += sizeC;
      }
  }
}
```

This appears to cause the FIFO state register to be spilled.

## Actual behavior

Object-file generation fails while running:

```text
Running pass 'Function Pass Manager'
Running pass 'AIE2 Assembly Printer' on function
'matmul_vectorized_2x2_bfp16_bf16_new<...>'
```

Relevant backtrace:

```text
getSImmOpValueXStep<9, 64u, 0, true, false>
AIE2PMCCodeEmitter::getBinaryCodeForInstr
AIEBaseMCCodeEmitter::encodeInstruction
MCELFStreamer::emitInstToData
AIEBaseAsmPrinter::emitInstruction
AsmPrinter::emitFunctionBody
```

Full diagnostic:

```text
fatal error: error in backend:
immediate operand value -56 is not a multiple of 64
```

## Invalid assembly generated by LLVM

Compiling the same source with `-S` succeeds because textual assembly does not
perform binary instruction encoding.

The resulting assembly exposes the invalid compiler-generated spills and
reloads:

```asm
vlda lfh0, [sp, #-128]              // 64-byte Folded Reload
vlda lfl0, [sp, #-56]               // 4-byte Folded Reload
lda  p0,   [sp, #-56]               // 4-byte Folded Reload

vst  lfl0, [sp, #-56]               // 4-byte Folded Spill
st   p0,   [sp, #-56]               // 4-byte Folded Spill
```

In particular:

```asm
vlda lfl0, [sp, #-56]
vst  lfl0, [sp, #-56]
```

cannot be encoded because these instructions require the immediate operand to
be a multiple of 64:

```text
-128, -64, 0, 64, ...
```

However, `-56 % 64 != 0`.

The comments also indicate that LLVM treats the slot as a 4-byte folded
spill/reload while selecting vector load/store instructions with a 64-byte
immediate step.

## Related MLIR-AIE documentation

The MLIR-AIE asymmetric-tile-buffering example currently documents that
Peano's AIE2P backend cannot compile its BFP16 matmul kernels and therefore
uses the Chess compiler:

https://github.com/Xilinx/mlir-aie/blob/main/programming_examples/ml/block_datatypes/gemm_asymmetric_tile_buffering/README.md#examples

## Ask
Can peano perform manual allocation of internal kernel registers, similar to chess?

atassis:
Fixes #1090.

## Root cause

`AIE2PRegisterInfo::eliminateFrameIndex` handles `VLDA_PLFR_SPILL` / `VST_PLFR_SPILL` -- the
`ePSRFLdF` composite -- by taking the immediate path whenever the frame offset is in range. The
composite expands to three sub-spills laid out in the slot as

```
[sub_fifo (128B)] [sub_avail (4B)] [sub_ptr (4B)]
```

`sub_fifo` goes through the FIFO (`dmx fifohl`) path, whose SP offset must be a multiple of 64
(`c16n_step64`). The range check alone does not establish that: it assumed the slot was 64-byte
aligned. When the composite lands in an under-aligned slot -- which `-O2` stack-slot reuse does
produce, and which is what #1090 reports -- the FIFO offset is a multiple of 4 but not of 64, and the
immediate path emits a `vst/vlda lfl, [sp, #imm]` that cannot be encoded.

## Fix

Take the immediate path only when the offset really is a multiple of 64; otherwise fall back to
indexed (register-offset) addressing, which `expandSpillPseudo` already applies to every sub-spill and
which encodes any offset. This mirrors the existing `EX_SPILL` / `E_SPILL` fallback in the same
function -- same `MvSclOpcode` copy of SP into a fresh `ePRegClass` vreg, same
`expandSpillPseudo(..., SPReg, Offset)` call shape -- so it introduces no new pattern.

## Why the range check stays on the scalar step

Worth stating explicitly, because it looks like a copy-paste from the neighbouring `EX_SPILL` case and
it is not. The gate keeps `isEncodableAsNegativeInt<9, 4>` rather than the FIFO's own step 64:

`isEncodableAsNegativeInt<N, step>` is `isInt<N + Log2(step) + 1>`, so `<9,4>` is `isInt<12>` (+/-2048)
and `<9,64>` is `isInt<16>` (+/-32768). The immediate path has to encode all three sub-spills, and the
two scalar halves sit at `Offset+128` and `Offset+132` on the step-4 path. Widening the gate to the
FIFO's step would accept, say, `Offset = -4096`: `sub_fifo` encodes fine, but `sub_avail` lands at
-3968, outside the scalar +/-2048 range, and the immediate path emits an unencodable store -- the same
class of defect this patch removes.

So `<9,4>` is the tightest of the three sub-spill constraints, checked conservatively at the base. It
costs a 128-byte band (offsets in `[-2176, -2049]` fall to indexed addressing when they need not),
which seemed the right trade against re-opening the bug from the other side.

## Relationship to the aie2ps work

aie2p already models this composite as a late-split pseudo, so this is not a competing design to
4fb2354fed33 / 4617c73e5373 (@khallouh) -- those retrofitted that same shape onto aie2ps to fix a
stack-liveness/verifier problem aie2p does not have.

One difference remains that this patch does not address: the gate here is all-or-nothing for the whole
composite, so when only the FIFO half is misaligned the scalar halves get indexed addressing too
(~9 instructions per direction against ~7). aie2ps decides per sub-spill instead, in
`emitAndExpandSubSpill`. Porting that to aie2p looks tractable -- `AIE2PInstrInfo.cpp` already has the
sub-spill list -- but it is an optimisation on a correct fallback rather than part of this fix, so I
left it out. Happy to follow up if you would rather have it in one go.

Also disjoint from #1155, which is on `VST_dmx_sts_x_spill` (a bare 512-bit vector spill, not a
composite) and touches a different case block.

## Evidence

`llvm/test/CodeGen/AIE/aie2p/spill/spill-reload-psrf-prologepilog.mir` covers both the aligned slot
(immediate path, a guard against regressing the common case) and the under-aligned slot (indexed
fallback).

Checked that the test actually discriminates, rather than passing either way: reverting only the
`AIE2PRegisterInfo.cpp` hunk and rebuilding changes `libLLVM.so` and the test fails, with the unfixed
compiler emitting the immediate FIFO store the patch is there to prevent. Restoring the hunk makes it
pass again.


martien-de-jong (commented):
Could you add the test in a first commit, marking it as XFAIL, or check in the pre-existing reference as is the case?

atassis:
I was busy a bit with mlir-aie last week, will try to switch to this repo in next few days. I'll land your requested changes and also bump some other my PRs, too.
I also have some uncommited drafted changes for llvm-aie, is it ok for me to suggest few more PRs first, or finish working at current created ones?
I am trying not to overspam with stupid PRs, but sometimes I am not totally sure if my PRs are valid and need a feedback, don't want to create excess noise here and there

atassis:
So I took the second option rather than XFAIL: the file has two functions and the 64-byte-aligned one already passed before the fix, so `XFAIL: *` would have marked a passing case as expected-to-fail. Instead e5fd928 checks in the pre-fix reference and is green on its own, and c1692040 regenerates it — so the behaviour change is that commit's diff, with the FIFO sub-spill moving from `VST_dmx_sts_fifohl_spill $lfl0, -184` to the indexed form.

atassis:
@martien-de-jong Hi there! Can you review this again, please?

atassis:
Superseded by #1276, which carries this branch's commits plus the correction.

The misalignment fallback here is wrong for the reason @khallouh gave on #1155: the FIFO store's
512-bit access truncates a misaligned address instead of faulting, so routing it to the indexed
form writes to the wrong slot silently rather than failing at encode time. #1276 keeps only the
range fallback and asserts the alignment invariant instead.

PR #1215, opened 2026-07-31
[AIE2PS] Add more end-to-end tests

andcarminati:
Cases that we should improve.


llvm/test/CodeGen/AIE/aie2ps/end-to-end/add2d.ll
```diff
@@ -0,0 +1,197 @@
...
+; ASM-NEXT:    nop
+; ASM-NEXT:    nop
+; ASM-NEXT:    vadd dm5, dm3, dm4, r8
+; ASM-NEXT:    nop
+; ASM-NEXT:    nop
+; ASM-NEXT:    nop
+; ASM-NEXT:    nop
+; ASM-NEXT:    nop
+; ASM-NEXT:    vst.srs.4x cml5, s1, srssign0, [p0], #32
+; ASM-NEXT:    st p2, [p7, #0]
```
martien-de-jong, line 97: these are the invariant stores you were talking about on Friday?  Furthermore, could we restrict the pointer for the vector stores?

martien-de-jong (dismissed):

llvm/test/CodeGen/AIE/aie2ps/end-to-end/add2d.ll
andcarminati: Actually, in this case, the vector store is aliasing with a vector load. We can pack everything in 5 cycles with `AIE_LOOP_HINT(aie-safe-to-ignore-memory-deps, 1)`. I am just testing this change now.

martien-de-jong (approved):

PR #1216, opened 2026-08-02
[AIE] Register symbols referenced from a bundle slot

atassis:
## Problem

`MCStreamer::emitInstruction` scans an MCInst's own operands for expressions, and
a bundle's operands are its slot instructions, so a symbol referenced from a slot
never reached the symbol table. The relocation went out against index 0 and the
linker resolved the call to address 0 -- a silent hang.

## Fix

Register them in an ELF streamer subclass, as Hexagon does. This is the layer
both object-emitting paths pass through, so `AIEBaseAsmPrinter::EmitToStreamer`'s
slot walk goes with it.

## Test

New `MC/AIE/bundle-slot-symbol-reloc.s`, on aie2, aie2p and aie2ps. `MC/AIE` +
`CodeGen/AIE`: 2260 pass, 2 fail; both failures are pre-existing, identical with
the walk restored. aie1 registers the same streamer but its asm parser cannot
assemble a call.

### DIFF
--- a/llvm/lib/Target/AIE/AIEBaseAsmPrinter.cpp
+++ b/llvm/lib/Target/AIE/AIEBaseAsmPrinter.cpp
@@ -37,22 +37,6 @@ using namespace llvm;
 
 #define DEBUG_TYPE "asm-printer"
 
-void AIEBaseAsmPrinter::EmitToStreamer(MCStreamer &S, const MCInst &Inst) {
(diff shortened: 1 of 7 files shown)


martien-de-jong (approved):
Looks like the Right Thing. As I understand it this would fail in the assembler, succeed in the compiler. A proof that we don't use the assembler significantly in day-to-day development.

PR #676, opened 2025-10-07
[AIEX][NFC] Refactor Intrinsic Attributes

F-Stuckmann:
Harmonize the Intrinsics attributes: Use `DefaultAttrsIntrinsic` for all AIE intrinsics.

Core_Compute_Insn_Count

Core_PMSize_absolute

Core_StackSize_absolute

PR #1217, opened 2026-08-03
[AIE2PS] Add DefaultAttrsIntrinsic for v*bf16_to_v*accfloat

andcarminati:
Changes to unblock some OLP cases.

### DIFF
--- a/llvm/include/llvm/IR/IntrinsicsAIE2PS.td
+++ b/llvm/include/llvm/IR/IntrinsicsAIE2PS.td
@@ -37,6 +37,11 @@ class AIE2PSV32AccFloatToV32Float16 : DefaultAttrsIntrinsic<[llvm_v32f16_ty],
                                           [IntrReadMem, IntrInaccessibleMemOnly]>;
 class AIE2PSV32Float16ToV32AccFloat : DefaultAttrsIntrinsic<[llvm_v32f32_ty], [llvm_v32f16_ty], [IntrNoMem]>;
 
+// v16bfloat16 to v16accfloat (AIE2PS-specific with DefaultAttrsIntrinsic)
+class AIE2PSV16BfToV16AccFloat : DefaultAttrsIntrinsic<[llvm_v16f32_ty], [llvm_v16bf16_ty], [IntrNoMem]>;
+// v32bfloat16 to v32accfloat (AIE2PS-specific with DefaultAttrsIntrinsic)
+class AIE2PSV32BfToV32AccFloat : DefaultAttrsIntrinsic<[llvm_v32f32_ty], [llvm_v32bf16_ty], [IntrNoMem]>;
+
 // SRS
 class AIE2PSV32AccFloatToV32Float8 : DefaultAttrsIntrinsic<[llvm_v32i8_ty],
                                           [llvm_v32f32_ty, llvm_i32_ty],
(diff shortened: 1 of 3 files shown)


llvm/lib/Target/AIE/aie2ps/AIE2PSInstrInfo.cpp
```diff
@@ -2285,6 +2286,25 @@ AIE2PSInstrInfo::getWidenNarrowConversionPairs() const {
...
 
+bool AIE2PSInstrInfo::isOLPSideEffectFree(const Instruction &I) const {
+  if (auto *II = dyn_cast<IntrinsicInst>(&I)) {
+    switch (II->getIntrinsicID()) {
+    // These conversion intrinsics are marked as having side effects in their
+    // TableGen definitions to prevent LICM from hoisting them out of loops,
+    // which caused adverse benchmark effects. However, for outer loop
+    // pipelining purposes, they are pure arithmetic conversions that can be
+    // safely speculated - they have no actual side effects beyond reading
+    // their inputs and producing a result.
```
martien-de-jong: This is the road to chaos. We are lying to the entire compiler about the sideeffects of these nice functional operations because of a wrong optimization cost estimate of one single machine pass, which inherits side effects and then re-establishing the truth for another pass at LLVM IR, using an admitted layering violation. 
A) I think this function should be called isSideEffectFree, since it is not specific for OLP, which is also not a clear acronym to use in a high-level predicate.
B) LICM should have a different mechanism to reject candidates. If that can be made in a backward comptible way, it would be a valuable upstream contribution.
C) I guess that the cause of deterioration is not using the invariance of the expression in e.g. rematerializing spills. Perhaps those flaws should be addressed to alleviate the performance impact.

martien-de-jong: Note that the default TII::isSideEffectFree could use MachineInstr::isSideEffectFree, and a target modification could do an early return true based on it and dig deeper before returning false.

llvm/lib/Target/AIE/aie2ps/AIE2PSInstrInfo.cpp
mludevid: To use a flag such as hasSideEffects which has a very specific meaning and is needed for correctness to guide an optimization pass is a horrible precedent.

mludevid: I find this name very confusing. The natural question when reading this is: "What is special about these instructions that they don't have side effects only when running inside a pipelined outer loop?"
I know that that is not the intention of this function, but the name suggests this.

llvm/lib/Target/AIE/AIEOuterLoopPipeliner.cpp
mludevid: I see that `collectLeanStage0` also uses `TII`. I don't see a comment there about this. Was this discusses when adding that function?

mludevid (commented):
I think there are two fundamental problems that we are trying to solve here:

1. As far as I know the conv intrinsics are marked as hasSideEffects to model the output dependency of the status register. When we know that the status register is not read afterwards (which it mostly isn't) we want to be able to put the instruction in a SEF stage of the OLP for potential performance gains.

2. We currently don't have good heuristics to drive the decisions of LICM (which instructions to hoist) and OLP (which instructions to put into which stage). If the decision to have the conv intrinsics marked as hasSideEffect was driven by some optimization related to LICM as your PR suggests, that is a horrible precedent.

In this PR we are hijacking the hasSideEffect flag inside the OLP to do both things at the same time. I think this is a bad idea because it will only take 1 kernel in which we do need to read the status register after the loop to tip over this entire "infrastructure". Modeling the side effects of our intrinsics properly is required for correctness, we shouldn't use those concepts to tweak performance. At the very least I think we should separate these two concepts by:

- Creating a function that decides that these intrinsics are allowed to go into a SEF stage even if they actually have side effects. Maybe call it `ignoreSideEffects` to make it clear that the side effects are real, but that we are deciding to skip them. Ideally this would be driven by knowing that the status register is actually not being read. But even if we hardcode this for now I much rather have it in a concise function that exactly states what it does.

- Creating proper heuristics to guide LICM and OLP. For example, in OLP we could take into account things like the estimated register pressure on the input and output registers to estimate if it's better to keep an instruction in the same stage as its predecessor or its successor instruction.

llvm/lib/Target/AIE/AIEOuterLoopPipeliner.cpp
martien-de-jong, line 947: want to

martien-de-jong: nit: then and else seem to be alternative implementations, yet the names suggest that on collects only stage0, while the other collects all stages.

andcarminati:
Hi @mludevid, after discussing with @martien-de-jong, this is the cleanest solution, we are just considering the very original attributes of those intrinsics.

llvm/include/llvm/IR/IntrinsicsAIE2PS.td
mludevid, line 41: @andcarminati, I though you mentioned that we couldn't do this due to adverse effects in LICM. Was that not the case or how did you mitigate those?

mludevid (commented):
This breaks the established AIE API, but I think it makes sense to commit to this change. We are paying a lot in terms of performance by modeling the status registers dependencies properly when in the large majority of kernels the status registers are never read. I think we should:
1. Document this properly! This needs an API change to be correct. Until that change has landed I think we should at least have some documentation markdown file in our repo explaining this.
2. Investigate if we can apply this change to all intrinsics. I think having some intrinsics that do keep correctness for status registers and some that don't would be confusing. I'd rather apply this change to all intrinsics if possible.

mludevid:
Discussed offline with @andcarminati:

- For context: this PR now basically reverts this hotfix https://github.com/Xilinx/llvm-aie/pull/681 for AIE2PS.

- The change to make all intrinsics not have side-effects was made in https://github.com/Xilinx/llvm-aie/pull/676. That PR is marked as NFC, which I think is not correct due to loosing the status register dependencies. As I already mentioned in a previous comment I think this is the pragmatic thing to do, but this needs to be documented. Andreu agreed to open a separate PR to document this difference between the AIE API requirements and peano's behavior.

Other than that I think this PR is good to go, we revert back to a consistent definition of our intrinsics (at least for AIE2PS) which I think is very positive. And we gain the desired OLP improvements. I think we should also revert this hotfix for AIE2P, maybe @F-Stuckmann could look into reverting his PR? If there are significant regressions in LICM we should look into adding heuristics to it to make better decisions there.

@martien-de-jong, are you okay with this proposal? Can you give the final approval?

martien-de-jong (approved):
I agree with Marc that we should document that all bets are off if people start to use status values. It can be a single paragraph in the optimization guidelines, and we can connect it to the pragma that drives the optimization that would use it, e.g. outer loop pipelining. I guess that outer loop pipelining would deserve some special attention there anyway.

PR #1219, opened 2026-08-03
[AIEX] Waw aggressive realloc revert failure fix

F-Stuckmann:
When reverting an aggressive WAWRegRewriter we were forgetting to revert the liveins. 

Structurally i have a simplification NFC and then the fix.


llvm/lib/Target/AIE/AIEWawRegRewriter.cpp
```diff
@@ -673,9 +676,20 @@ bool AIEWawRegRewriter::renameMBBPhysRegs(const MachineBasicBlock *MBB) {
...
+    OriginalAllocation All;
+    All.reserve(FreedLiveIns.size() + Candidates.size());
+    All.insert(All.end(), FreedLiveIns.begin(), FreedLiveIns.end());
+    All.insert(All.end(), Candidates.begin(), Candidates.end());
+    return All;
+  };
 
   if (!reAllocate(Candidates, LRURegisters)) {
-    revertAllocation(Candidates);
+    revertAllocation(getRevertEntries());
```
martien-de-jong: nit: I think you could pass a range, which can be built using llvm::concat from the two component ranges.

martien-de-jong (approved):

PR #1221, opened 2026-08-04
Fix missed resource conflict in single-stage pipeline. 

### LINKED ISSUE #1012: [AIE2] Partial-word store scheduled into zero-overhead-loop end bundle is dropped on hardware
While building an open-source AIE2 emulator/validator we found what looks like an AIE2 codegen correctness bug in Peano, and wanted to flag it in case it's useful. Happy to help with a reduced test case or a fix if you'd like.

## Summary

For simple scalar loops with a sub-word (`int8_t`/`int16_t`) output store, Peano's unrolling places the final `st.s8`/`st.s16` in the **loop-end (`LE`) bundle** of a zero-overhead (hardware) loop. On real Phoenix (NPU1/AIE2) silicon that store is dropped on every back-edge -- only the final fall-through iteration commits -- so every Nth output element keeps its initial value. `xchesscc` compiles the same source correctly (it does not park the store at `LE`).

## Reproducer

```cpp
#include <stdint.h>
extern "C" void k(int8_t* __restrict out) {
    for (int i = 0; i < 128; i++)
        out[i] = i * i;        // int8
}
```

Built with `clang --target=aie2-none-unknown-elf -O2 -c k.cc -o k.o` and disassembled with `llvm-objdump -d --triple=aie2 k.o`, the core loop is a zero-overhead loop whose `LE` bundle is the `st.s8`:

```
00000020 <.LBB0_1>:                       ; loop start (LS)
      20: ... mul r1, r0, r0 ; mov dj0, r0
      30: st.s8 r1, [p0, dj0]
      ...
      7a: mul r2, r1, r1                   ; data for the 4th unrolled element
      7e: nop
00000080 <.L_LEnd0>:                       ; loop end (LE)
      80: nopb ; st.s8 r2, [p0, dj0] ; nops ; nopxm ; nopv   <-- parked at LE
```

Expected `out[i] = (i*i) & 0xff`. On hardware, every `i % 4 == 3` -- the unrolled element whose store lands in the `LE` bundle -- reads `0`, except the final iteration (which falls through with no back-edge):

| i | 3 | 7 | 11 | 15 | ... | 127 |
|---|---|---|----|----|-----|-----|
| expected | 9 | 49 | 121 | -31 | ... | 1 |
| **NPU** | **0** | **0** | **0** | **0** | ... | 1 |

## Why we think it's the LE-bundle store

- Only the element whose `st.s8` is the LE-bundle store is dropped; stores earlier in the loop body commit fine.
- It is specific to **partial-word** stores (`st.s8`/`st.s16`), which are read-modify-write and commit late (`AIE2Schedule.td`: `II_STHB` `MemoryCycles<[5,11]>`, i.e. E11). A full-word `st` in the same slot is fine.
- `xchesscc` does not place the partial-word store in the `LE` bundle and produces correct output.

This suggests the AIE2 scheduler should avoid placing a partial-word store in the `LE` bundle of a hardware loop (or account for its E11 commit against the back-edge fetch redirect), similar to how it already keeps the `ls`/`le`/`lc` loop-setup writes a fixed distance from `LE`.

## Environment

- Target: AIE2 / Phoenix (NPU1, Ryzen AI)
- Peano: `llvm-aie@0b72475` (clang 21.0.0git)
- Reproduced on real NPU hardware; `xchesscc` output is correct.

Glad to provide the full objdump, more reproducers, or a reduced test case -- just let us know what's most useful.

martien-de-jong:
Should fix #1012


llvm/test/CodeGen/AIE/aie2/schedule/postpipeliner/single-stage-resource-conflict.mir
```diff
@@ -0,0 +1,107 @@
...
+  ; CHECK-NEXT:    // =>This Inner Loop Header: Depth=1
+  ; CHECK-NEXT:    nopb ; nopa ; nops ; add r1, r0, #1; mov dj0, r0; nopv
+  ; CHECK-NEXT:    nopb ; st.s8 r2, [p0, dj0]; nops ; add r0, r0, #4; mov dj0, r1; nopv
+  ; CHECK-NEXT:    nopb ; nopa ; nops ; nopxm ; nopv
+  ; CHECK-NEXT:    nopb ; nopa ; nops ; nopxm ; nopv
+  ; CHECK-NEXT:    nopb ; nopa ; nops ; nopxm ; nopv
+  ; CHECK-NEXT:    nopa ; nopb ; nopx
+  ; CHECK-NEXT:    nop
+  ; CHECK-NEXT:    nop
+  ; CHECK-NEXT:    st.s8 r2, [p0, dj0]
```
martien-de-jong, line 32: The RMW region of this part-word store overlaps with that of the part-word store in the second cycle of this loop.

llvm/lib/Target/AIE/AIEPostPipeliner.cpp
```diff
@@ -1043,9 +1043,12 @@ bool PostPipeliner::scheduleOtherIterations(PostPipelinerStrategy &Strategy) {
...
   // to steady state by checking and inserting NStages. Since the steady
   // state is the busiest, we can shift the scoreboard by II after each stage.
+  // We repeat the resource schedule often enough to make the final one land
+  // after the conflict horizon of the first one.
+  const int PipelineDepth = HR.getPipelineDepth();
   ResourceScoreboard<FuncUnitWrapper> Resources;
-  Resources.config(0, II + HR.getPipelineDepth());
-  for (int S = 0; S < NStages; S++) {
+  Resources.config(0, 2 * II + PipelineDepth);
+  for (int Start = 0; Start < II + PipelineDepth; Start += II) {
```
andcarminati, line 1051: Elegant fix!

llvm/test/CodeGen/AIE/aie2p/schedule/postpipeliner/pipelined-lock-ptr-fifo-alias.ll
```diff
@@ -59,11 +59,12 @@ define void @pipelined_acquire_ptr_fifo_same(ptr noalias %io, i32 %lock_id) {
 ; CHECK-NEXT:    nopa ; nopb ; nops ; nopx ; mov r1, p0; nopv
 ; CHECK-NEXT:    nopa ; nopb ; nops ; nopxm ; nopv
 ; CHECK-NEXT:    nopa ; nopb ; nops ; nopxm ; nopv
-; CHECK-NEXT:    st.s8 r0, [p1, #0]; nopx
+; CHECK-NEXT:    st.s8 r0, [p1, #0]
 ; CHECK-NEXT:    nop
 ; CHECK-NEXT:    nop
 ; CHECK-NEXT:    nop
 ; CHECK-NEXT:    vextract.8 r0, x2, #0, vaddsign1
+; CHECK-NEXT:    nop
```
andcarminati, line 67: The problem was here before!

andcarminati (approved):
LGTM. Thanks to the original issue reporter @FIM43-Redeye and @martien-de-jong for the fix!

F-Stuckmann (approved):

PR #1223, opened 2026-08-05
[AIE][OLP] Add outer loop type selection for outer loop pipeliner

andcarminati:
Add command-line options to control whether the outer loop uses hardware
(JNZD) or software loop after pipelining:

- `-aie-outer-loop-type=[soft,hw,auto]`: Explicit loop type selection
  - soft: Always use software loop (original icmp-based latch)
  - hw: Always use hardware loop (JNZD conversion)
  - auto: Select based on pointer pressure (default)


llvm/test/CodeGen/AIE/aie2ps/end-to-end/loop_1_biased_conv_mx6x6_ebs16_unroll_impl_initial.ll
```diff
@@ -73,7 +73,7 @@ define dso_local void @loop_1_biased_conv_mx6x6_ebs16_unroll_impl_initial(i32 %c
 ; REMARKS-NEXT:   - Prologue:        bb.1.steady.stage1.top
 ; REMARKS-NEXT:   - PrologueBundles: '17'
 ; REMARKS-NEXT:   - Epilogue:        bb.3.steady.stage1.bottom.and.stage0.top
-; REMARKS-NEXT:   - EpilogueBundles: '27'
+; REMARKS-NEXT:   - EpilogueBundles: '28'
```
andcarminati, line 76: This specific loop will be optimized by the a next feature: fist iter mode..

llvm/lib/Target/AIE/AIEOuterLoopPipeliner.cpp
```diff
@@ -125,11 +125,25 @@ static cl::opt<bool> LeanStage0Mode(
...
+                          "Use hardware loop (JNZD conversion)"),
+               clEnumValN(OuterLoopType::Auto, "auto",
+                          "Auto-select based on pointer pressure")),
+    cl::init(OuterLoopType::Auto), cl::Hidden);
+
+static cl::opt<unsigned> OuterLoopPointerThreshold(
+    "aie-outer-loop-pointer-threshold",
+    cl::desc("Pointer pressure threshold for auto mode: if base pointer PHIs "
+             ">= threshold, use soft loop; otherwise use hw loop"),
+    cl::init(6), cl::Hidden);
```
martien-de-jong, line 147: six running pointers, one spare, one for jnzd. 
We may want to add more heuristics to estimate the 'spare'. Perhaps the description should be phrased in terms of estimated pointer register pressure rather than pointer PHI's.
Note that if the inner loop has pointer PHIs they will also have a contribution to the outer loop pressure. I guess it should be pretty simple to add inner loop phis that are not also catering for outer loop phis.

andcarminati: Innerloop recurrence phis are also part of the outer loop phis - if they are not restarting on every iteration, so naturally outer loop exposes this information. I will change a bit the description.

andcarminati: After rebasing, I see some regressions related to tests that are also in focus for the next pointer optimization round. Let's hold this a bit.

llvm/test/CodeGen/AIE/aie2ps/end-to-end/gemm_mx6_loop_for_body_2.ll
```diff
@@ -75,7 +75,7 @@ define dso_local void @gemm.if.else(ptr %add.ptr, ptr %tdm1) #5 {
 ; REMARKS-NEXT:   - Prologue:        bb.1.steady.stage1.top
 ; REMARKS-NEXT:   - PrologueBundles: '16'
 ; REMARKS-NEXT:   - Epilogue:        bb.3.steady.stage1.bottom.and.stage0.top
-; REMARKS-NEXT:   - EpilogueBundles: '24'
+; REMARKS-NEXT:   - EpilogueBundles: '29'
```
andcarminati, line 78: **On hold:** This is related to inefficient pointer handling. I will solve this before merging this PR.

llvm/lib/Target/AIE/AIEOuterLoopPipeliner.cpp
martien-de-jong: I meant that pointer definitions in the inner loop that are not part of the outer loop phi's will interfere with thos outer loop phis' and hence increase pressure.

martien-de-jong (approved):
Just a supernit at your discretion, the formal 'NumPointerPhis'  represents the more abstract pressure from the description, not the way it is estimated.

llvm/lib/Target/AIE/AIEOuterLoopPipeliner.cpp
```diff
@@ -150,7 +165,21 @@ struct OLPOpts {
             Overrides.hasOverride(SpeculativeLastIteration)
                 ? Overrides.get(SpeculativeLastIteration)
                 : UseLeanStage0),
-        UseHardwareLoop(Overrides.get(EnableOuterLoopHardwareLoop)) {}
+        LoopType(Overrides.get(OuterLoopTypeOpt)) {}
+
+  /// Determine whether to use hardware loop based on loop type and pointer
+  /// pressure. In auto mode, use hw loop if pointers < threshold.
+  bool shouldUseHardwareLoop(unsigned NumPointerPHIs) const {
```
F-Stuckmann, line 172: we should think about moving this to generic hardwareloops pass. This seems like a generic opportunity for nested loops.

martien-de-jong (approved):

PR #1229, opened 2026-08-07
Update list of test targets to execute

philippjh:


.github/workflows/amd-aie-tests.yml
```diff
@@ -27,7 +27,7 @@ jobs:
     name: Test AIE llvm,clang,lld
     uses: ./.github/workflows/llvm-project-tests.yml
     with:
-      build_target: check-llvm-aie check-clang-aie check-lld-aie
+      build_target: check-llvm builtins check-clang check-lld
```
khallouh: We already run non AIE tests in `amd-upstream-tests.yml` so we can keep `check-llvm-aie` and `check-lld-aie` to not run the upstream tests twice. `check-lld-aie` has a problem which I will fix. We can add builtins which was actually missing.

khallouh: Here is the PR https://github.com/Xilinx/llvm-aie/pull/1231 , this should fix CI on this one

khallouh (approved):
LGTM

PR #1231, opened 2026-08-07
[LLD][AIE] Make check-lld-aie actually run the AIE tests and fix the tests

khallouh:


jorickert (approved):

martien-de-jong (approved):
Do the right thing. Yay.

PR #1238, opened 2026-08-11
[AIE] Build the AIE runtimes (libc) in the public AIE test job

khallouh:
The test job builds an explicit target list, so the runtimes external projects configured by Peano-AIE.cmake (libc) were never compiled. And they are not covered by "builtins" which was recently added.

### DIFF
--- a/.github/workflows/llvm-project-tests.yml
+++ b/.github/workflows/llvm-project-tests.yml
@@ -97,7 +97,6 @@ jobs:
         with:
           fetch-depth: 250
       - name: Install requirements.txt
-        if: inputs.projects == 'mlir'
         run: |
           sudo apt-get update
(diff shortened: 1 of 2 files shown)


khallouh:
Parking as draft to missing python YAML dependency

.github/workflows/llvm-project-tests.yml
```diff
@@ -97,7 +97,6 @@ jobs:
         with:
           fetch-depth: 250
       - name: Install requirements.txt
-        if: inputs.projects == 'mlir'
```
khallouh, line 100: This would break Windows jobs. Do we need an OS condition? It worked because MLIR only ever runs on ubuntu-24.04.

PR #1239, opened 2026-08-11
[RFC][AIE2P] llvm-aie-run: an executable core model built on the MC decoder and itineraries

atassis:
Draft for discussion, not a merge request.

This is one half of a pair. The array and fabric model lives in a matching mlir-aie branch
(https://github.com/atassis/mlir-aie/tree/feat/aie-array-simulator) and the two execute together across a
versioned C ABI. Neither half does anything alone.

## What it is

An AIE2P core model that decodes through the target's own MC layer and reads its itineraries, rather than
carrying a second instruction description that would drift from the backend. It is exposed as
`llvm-aie-run`, which executes an ELF and reports per-opcode and per-register coverage, stalls, memory
regions, and explicit pass/fail/unknown verdicts.

The design rule is that unestablished behaviour faults instead of guessing. Where nothing in this tree
states what the hardware does, the model refuses rather than returning a plausible answer, so an
unmodelled corner surfaces as a refusal in a test rather than as a wrong number in a result.

88 files, +13,100 lines, no deletions. No existing behaviour changes.

## What it is not

Not a whole-array model and not silicon-calibrated. Core streams, cascade, neighbour bands and parts of
shim transport are open. No cycle count has been compared against hardware, so this makes no claim about
timing accuracy yet.

## Gates

- 50/50 `llvm-aie-run` tests here, 20/20 on the mlir-aie side
- 168 non-NOP AIE2P defs are issued by the kernels I build; 158 are covered
- Every opcode landed with mutation coverage: each modelled behaviour has a mutation that must turn the
  suite red, and a green control asserted after each restore

## What I would like an opinion on

1. The timing substrate. I read AIE2P's itineraries as a PORT model rather than a duration model:
   `AIE2PGenSchedule.td` has 4010 `InstrItinData` entries and only 9 real `InstrStage` uses (8x
   `LCKREQ` at 4 cycles, 1x `DONE_FU` at 3); everything else is `PrefixCycle` (a port, no time) or
   `EmptyCycles` (time, no port). So the model times straight-line compute flat at one bundle per
   cycle, with `PART_WORD_STORE` as the one buildable hazard. Is that reading right, and is there
   anything else in the ISA that carries real duration?
2. The C ABI boundary between this and the array model. Does that split belong here, or is it in the
   wrong place?
3. `llvm-aie-run` as a new tool. Is that the right home, or should this hang off an existing entry point?


atassis:
Paired with the mlir-aie half: https://github.com/Xilinx/mlir-aie/pull/3539

martien-de-jong:
Whoa. Interesting stuff. I don't want to leave you in the cold, but we currently don't have the bandwidth to review many external PRs, and definitely not PRs on this scale.  I can at least briefly answer your questions.
1- in AIE2PS we have VMAC instructions that use the same resource for two consecutive cycles. AFAIK, they are serializations of two halves of an input vector. That input vector also has distinct read stages for the low and the high part.  The LCK and PARTWORD resources are artificial, they serve to prevent race conditions that can not be described by latencies. Many FUNCUNITS correspond to actual ports, and will be claimed for one cycle. They are accompanied by corresponding operand latencies that redundantly code the register read/write timing.
2- I'm afraid I don't understand the question. Perhaps you can give some pointers into the code. I don't know what that C-ABI boundary is, or what you mean by 'the array'.
3- A simulator for me is a tool like objdump, providing an interface to a number of libraries. I would expect it to be a sibling of lvm-objdump in the tree.

atassis:
Sorry for the slow reply. I have been heads-down on kernel patches, among other things, and this deserved a considered answer rather than a quick one.

Understood on bandwidth, and I am not asking you to review the whole thing. Below is the answer to your question 2, which is the part I explained badly, plus what changes off the back of 1 and 3.

**1. Itineraries.** Your answer corrected an assumption: I had measured AIE2P's itineraries as a port model rather than a duration model and had taken that to generalise, but AIE2PS's fifty real `InstrStage<2, ...>` uses on the bfp16 MAC family are exactly the two-halves serialization you describe. The model now keeps one-bundle-per-cycle for AIE2P compute code, treats the two-cycle occupancy as AIE2PS-only, and handles `LCK` and `PARTWORD` as race guards rather than as timing.

**2. The C ABI boundary, with pointers.** My fault for using shorthand. The file is `llvm/lib/Target/AIE/Sim/aie_iss_c_abi.h`, 167 lines, and its header comment states all of this.

"The array" is the tile grid outside the core: DMAs, stream switches, locks, memtiles. That model lives in mlir-aie. This PR is only the core.

They cannot simply link. The core model needs the MC layer (disassembler, instruction info, subtarget) to decode bundles, and instruction semantics belong to this backend rather than to mlir-aie. Meanwhile the array model cannot be built against LLVM headers or bound to a particular `libLLVM`, because mlir-aie and Peano are separately versioned and separately built, and a Peano distribution ships `libLLVM.so` with no LLVM headers. So the boundary is a plain C ABI resolved with `dlopen`/`dlsym`, both sides vendor an identical copy of that header, and `AIE_ISS_ABI_VERSION` gates compatibility.

The consequence is the part worth judging: every memory access, lock operation, stream access and character the core performs leaves through host callbacks, so the ISS models the datapath only and holds no state beyond architectural registers. If that split is wrong, that is the most useful thing you could tell me, because everything else follows from it.

**3. Shape.** Agreed, and I will restructure to match: a library, with `llvm-aie-run` as a thin sibling of `llvm-objdump` rather than a tool that owns the model. The current layout is close to it already (`llvm/lib/Target/AIE/Sim` plus a 288-line driver), so it is mostly moving the seam and making the library usable on its own.

**On the size, since it is the obvious objection.** The diff reads as 13,472 lines, but 7,372 of those (55%) are tests and the model itself is 11 files. The design decisions live in `AIECoreState.h`, `AIEExecutor.h` and `aie_iss_c_abi.h`, about 550 lines between them. `AIE2PSemantics.cpp` is 3,456 lines of per-opcode semantics and is better skimmed than read.

**One concrete question, and it is separable from the RFC.** The model faults rather than guesses on anything the tree does not state, so its gaps are enumerable rather than vague: 345 of 1198 opcodes have semantics today and the rest fault by name. The group I cannot settle from the tree alone is the bfp16 MAC and block-format conversion family, for example

    VMAC_f_vmac_bfp_vmul_bfp_core_EX_EY   (and the EX_QEY / EY_QEX forms)
    VMUL_f_vmul_bfp_vmul_bfp_core_EX_EX   (and EX_EY / EX_QEY)
    VCONV_bfp16ebs16_ebs8
    VCONV_bfp16ebs16_fp32

What I cannot establish is the exact rounding and exponent behaviour: which shared exponent survives `ebs16 -> ebs8`, what happens to a mantissa that cannot be represented under the destination's exponent, and whether the `QEX`/`QEY` forms differ from `EX`/`EY` in anything beyond operand selection. The ISA description gives me the encoding but not the answer, and guessing would produce a model that is confidently wrong in exactly the place it matters most.

Would you be willing to answer ten of those in a batch? It is separable from reviewing the RFC and it unblocks the part of the model I cannot make progress on alone.

On why I would rather get this right than let it lapse: with the core backend defaulted to Peano, compilation no longer needs Chess, and simulation is the piece still reaching for Vitis. That is the gap this is aimed at. I will refresh the branch before asking anyone to look at it properly.

PR #1242, opened 2026-08-12
Reset MLIR EmitC, Linalg, and Affine dialect to upstream state to make syncing with upstream easier

jorickert:
The changes we did are not used/needed downstream anymore.


cferry-AMD (approved):
Thanks for taking care of this!

mgehre-amd (approved):
Thanks!

PR #1246, opened 2026-08-14
[AIEX][OLP] Establish LCSSA form before loop pipelining

andcarminati:
### Problem

The Outer Loop Pipeliner produces __poison values__ in exit phi nodes when:

- A function contains multiple loops that share a common exit block
- Loops exit through intermediate exit blocks (not directly to the shared exit)
- Values computed in the loop need to flow to the shared exit via phi nodes

This occurred because the pipeliner's transformation (cloning and remapping) did not properly maintain LCSSA form. When the original loop blocks were removed, references in exit phi nodes became dangling and resolved to poison.

### Root Cause

LCSSA (Loop Closed Static Single Assignment) form requires that all values used outside a loop are captured by phi nodes at the loop's exit. Without proper LCSSA form before the transformation:

1. Values exiting through intermediate blocks had no LCSSA phi
2. The pipeliner's remap logic couldn't connect exit live-outs to cloned definitions
3. After `removeFromCFG` deleted original blocks, exit phis referenced deleted values → poison

### Solution
__Add a single line at the start of `performTransformation`:__

```cpp
formLCSSARecursively(*OrigLS.getOuterLoop(), *DT, LI, SE);
```

This ensures proper LCSSA form is established before any transformation, creating the necessary phi nodes for all exit live-outs.

### Cleanup (Follow-up NFC)

The "non-phi exit live-out" workaround code (manual operand remapping in the exit block) is now dead and can be removed. This workaround was handling cases where no LCSSA phi existed - a situation that can no longer occur with LCSSA form guaranteed upfront.

### DIFF
--- a/llvm/test/CodeGen/AIE/aie2ps/end-to-end/gemm_mx6_loop_for_body_1.ll
+++ b/llvm/test/CodeGen/AIE/aie2ps/end-to-end/gemm_mx6_loop_for_body_1.ll
@@ -122,7 +122,7 @@ define dso_local void @gemm.if.then5(ptr %add.ptr, ptr %tdm1, i32 %cond3) #5 {
 ; ASM-NEXT:    lda dc0, [p1], #4; vmov bmlh0, bmll0
 ; ASM-NEXT:    lda dc4, [p1], #24; or r16, r10, r10; vmov cmh1, cml1
 ; ASM-NEXT:    lda dc1, [p1], #4; movs m7, dj3; movx r5, #0; vmov cmh0, cml0
-; ASM-NEXT:    lda dn0, [p2, #0]; movs m6, dj3; or r25, r5, r5; mov r7, p1
+; ASM-NEXT:    lda dn0, [p2, #0]; movs m6, dj3; or r25, r5, r5; mov r17, p1
 ; ASM-NEXT:    lda dc5, [p1, #0]; movs p1, p0; or r24, r5, r5; mov p0, p3
 ; ASM-NEXT:  .LBB0_1: // %steady.stage1.top
 ; ASM-NEXT:    // =>This Loop Header: Depth=1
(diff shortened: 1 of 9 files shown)


llvm/test/CodeGen/AIE/aie2ps/end-to-end/gemm_mx6_loop_for_body_1.ll
```diff
@@ -216,14 +216,14 @@ define dso_local void @gemm.if.then5(ptr %add.ptr, ptr %tdm1, i32 %cond3) #5 {
...
-; ASM-NEXT:    paddxm [sp], #-64; st dc4, [p0], #24; vmov fewl2, fewh2; vmac.f dm6, dm6, fex4, fey0, r8
-; ASM-NEXT:    st dc1, [p0, #0]; vmov fewl0, fewh10; vmac.f dm5, dm5, fex2, fey0, r8
-; ASM-NEXT:    st dc5, [p0, #4]; vmac.f dm1, dm1, fex2, fey0, r8
-; ASM-NEXT:    vmov fewl0, fewh8; vmac.f dm4, dm4, fex10, fey3, r8
-; ASM-NEXT:    vmac.f dm3, dm3, fex0, fey3, r8
+; ASM-NEXT:    padda [p5], #128; paddb [p3], #128; movs p0, r17; nopx ; mov m0, #32; vmac.f dm0, dm0, fex0, fey3, r8
+; ASM-NEXT:    lda p7, [sp, #-60]; paddb [p0], m0; movx r8, #772; vmov fewl4, fewh4; vmac.f dm2, dm2, fex8, fey3, r8 // 4-byte Folded Reload
+; ASM-NEXT:    lda p6, [sp, #-64]; or r10, r16, r16; mov m0, #-64; vmac.f dm7, dm7, fex4, fey0, r8 // 4-byte Folded Reload
+; ASM-NEXT:    paddxm [sp], #-64; st r1, [p0], m0; vmov fewl2, fewh2; vmac.f dm6, dm6, fex4, fey0, r8
+; ASM-NEXT:    st dc0, [p0], #4; vmov fewl0, fewh10; vmac.f dm5, dm5, fex2, fey0, r8
```
andcarminati, line 223: Nice, this test is also reproducing the bug.

khallouh (approved):
LGTM!

PR #1247, opened 2026-08-14
[AIEX] OLP pointer improvements

andcarminati:
This series of commits introduces a comprehensive pointer chain optimization pipeline for nested outer loops in AIE targets.

The work begins with the AIEOuterLoopPointerOptimizer, an IR-level pass that identifies loops with a specific structure: a prologue block followed by a single-block inner loop and an epilogue block. The pass performs several transformations: it canonicalizes GEPs to operate in the PHI's native address space (eliminating redundant addrspacecasts), converts non-i8 GEPs to i8-based GEPs for uniformity, links consecutive GEPs into chains where each uses the previous GEP as its base (enabling post-increment patterns), and hoists epilogue GEPs to the prologue when safe.

At the MIR level, the outer loop pipeliner was extended to promote derived pointer updates to stage-0 (the prologue). This allows pointer increments that depend on loop-carried PHIs to be computed earlier, positioning them for post-increment folding with memory operations.

A new pointer add reassociation combiner was added with a single-use guard to prevent creating expressions that would extend live ranges. This works in conjunction with a subsequent fix that explicitly skips offset combining when it would extend the live range of a G_PTR_ADD across loop boundaries.

Together, these optimizations prepare pointer chains at the IR level and then leverage them at the MIR level to enable post-increment loads and stores, reducing register pressure and improving code density in nested loop kernels.


llvm/test/CodeGen/AIE/aie2ps/end-to-end/gemm_int8_1_v2.ll
```diff
@@ -62,7 +62,7 @@ define dso_local void @gemm_int8_1_v2(ptr noalias %p_a, ptr noalias %p_b, ptr no
 ; REMARKS-NEXT:   - Prologue:        bb.1.steady.stage1.top
 ; REMARKS-NEXT:   - PrologueBundles: '12'
 ; REMARKS-NEXT:   - Epilogue:        bb.3.steady.stage1.bottom.and.stage0.top
-; REMARKS-NEXT:   - EpilogueBundles: '22'
+; REMARKS-NEXT:   - EpilogueBundles: '31'
```
andcarminati: This is my focus now.

andcarminati: the optimizer should look trough address space casts.

llvm/test/CodeGen/AIE/aie2ps/end-to-end/gemm_int8_1_v2.ll
```diff
@@ -111,114 +111,118 @@ define dso_local void @gemm_int8_1_v2(ptr noalias %p_a, ptr noalias %p_b, ptr no
...
-; CHECK-NEXT:    vst.srs.2x cmh1, s0, srssign1, [p7], #64; vshuffle x10, x1, x0, r20 // Delay Slot 5
-; CHECK-NEXT:    vst.srs.2x cml0, s0, srssign1, [p7], #64; vshuffle x1, x10, x0, r22 // Delay Slot 4
-; CHECK-NEXT:    mov p4, p7 // Delay Slot 3
-; CHECK-NEXT:    vst.srs.2x cmh0, s0, srssign1, [p4], #64; vshuffle x10, x3, x0, r24 // Delay Slot 2
-; CHECK-NEXT:    vshuffle x10, x10, x0, r26 // Delay Slot 1
+; CHECK-NEXT:    vst.srs.2x cml5, s0, srssign1, [p4], #64
+; CHECK-NEXT:    vst.srs.2x cmh5, s0, srssign1, [p4], #64
+; CHECK-NEXT:    vst.srs.2x cmh3, s0, srssign1, [p4, #64]; mov p6, p4
+; CHECK-NEXT:    vst.srs.2x cml3, s0, srssign1, [p6], #128; vsel.32 x4, x4, x10, r17
+; CHECK-NEXT:    vst.srs.2x cmh2, s0, srssign1, [p6, #64]; mov p7, p6
```
andcarminati: Investigate this

andcarminati: No pointer moves left.

llvm/lib/Target/AIE/AIEOuterLoopPointerOptimizer.cpp
```diff
@@ -0,0 +1,1033 @@
...
+// structure: prologue -> single-block inner loop -> epilogue.
+//
+// Optimizations:
+// 1. GEP Address Space Canonicalization: Keep GEPs in the PHI's canonical
+//    address space. This moves addrspacecast from before GEPs to point-of-use,
+//    enabling better GEP chain optimization.
+// 2. GEP Canonicalization: Convert non-i8 GEPs to i8-based GEPs for uniformity.
+//    This makes pointer arithmetic explicit and helps later optimizations.
+// 3. GEP Chain Linking: Link consecutive GEPs with the same base pointer to
+//    enable post-increment addressing patterns.
```
martien-de-jong, line 21: This serializes the address chains, possibly delaying the user of the last GEP, after the one using the first GEP. To know whether this is profitable, we need to know that we can't build this chained offset with an offset load.

llvm/lib/Target/AIE/AIEOuterLoopPointerOptimizer.cpp
```diff
@@ -0,0 +1,1033 @@
...
+// 4. GEP Hoisting: Move GEPs from bottom (epilogue) to top (prologue) block
+//    when safe, reducing code in the epilogue and improving scheduling.
```
martien-de-jong: I think this uses a lot of assumptions as to why it improves scheduling.

llvm/lib/Target/AIE/AIEOuterLoopPointerOptimizer.cpp
```diff
@@ -0,0 +1,1033 @@
...
+    cl::init(true), cl::Hidden);
+
+namespace {
+
+//===----------------------------------------------------------------------===//
+// Helper Functions
+//===----------------------------------------------------------------------===//
+
+/// Collect all GEPs from a basic block.
+static SmallVector<GetElementPtrInst *, 16> collectGEPs(BasicBlock *BB) {
```
martien-de-jong: static and anonymous namespace is overkill

llvm/lib/Target/AIE/AIEOuterLoopPointerOptimizer.cpp
```diff
@@ -0,0 +1,1033 @@
...
+/// Check if a GEP is i8-based with a single index.
+static bool isSimpleI8GEP(const GetElementPtrInst *GEP) {
+  return isSimpleGEP(GEP) && GEP->getSourceElementType()->isIntegerTy(8);
+}
+
+//===----------------------------------------------------------------------===//
+// GEP Canonicalization Helpers (non-i8 to i8-based)
+//===----------------------------------------------------------------------===//
+
+/// Replace a GEP with an i8-based version using the given byte offset.
```
martien-de-jong: nit: it doesn't actually replace it, it only builds a new one.

llvm/lib/Target/AIE/AIEOuterLoopPointerOptimizer.cpp
```diff
@@ -0,0 +1,1033 @@
...
+  return NewGEP;
+}
+
+/// Check if a GEP needs canonicalization to i8-based.
+static bool needsI8Canonicalization(GetElementPtrInst *GEP) {
+  // Skip GEPs that already use i8 as source element type
+  if (GEP->getSourceElementType()->isIntegerTy(8))
+    return false;
+  // We only handle simple GEPs with a single index
+  return GEP->getNumIndices() == 1;
```
martien-de-jong: isSimpleGEP() ?

martien-de-jong: Why not canonicalize multi-index to simple index?

llvm/lib/Target/AIE/AIEOuterLoopPointerOptimizer.cpp
```diff
@@ -0,0 +1,1033 @@
...
+static bool isChainLinkCandidate(GetElementPtrInst *GEP, int64_t &OutOffset) {
+  if (!isSimpleI8GEP(GEP))
+    return false;
+
+  auto *ConstIdx = dyn_cast<ConstantInt>(GEP->getOperand(1));
+  if (!ConstIdx)
+    return false;
+
+  OutOffset = ConstIdx->getSExtValue();
+  return OutOffset > 0; // Must be positive offset
```
martien-de-jong: Please no trailing line comments.
Why should it be positive?

llvm/lib/Target/AIE/AIEOuterLoopPointerOptimizer.cpp
```diff
@@ -0,0 +1,1033 @@
...
+
+/// Try to link a GEP to an existing chain, rewriting it to use delta offset.
+/// Returns the new GEP if linked, nullptr otherwise.
+static GetElementPtrInst *tryLinkToChain(GetElementPtrInst *GEP,
+                                         GEPChainState &ChainState,
+                                         int64_t CurrentOffset,
+                                         DominatorTree *DT) {
+  GetElementPtrInst *PrevGEP = ChainState.LastGEP;
+  int64_t PrevOffset = ChainState.LastOffset;
+  int64_t DeltaOffset = CurrentOffset - PrevOffset;
```
martien-de-jong: const const const

llvm/lib/Target/AIE/AIEOuterLoopPointerOptimizer.cpp
```diff
@@ -0,0 +1,1033 @@
...
+  Type *IdxTy = ConstIdx->getType();
+  Value *DeltaOffsetVal = ConstantInt::get(IdxTy, DeltaOffset);
+  Value *NewGEP =
+      Builder.CreateGEP(Builder.getInt8Ty(), PrevGEP, DeltaOffsetVal,
+                        GEP->getName() + ".chained");
+
+  if (auto *NewGEPInst = dyn_cast<GetElementPtrInst>(NewGEP))
+    NewGEPInst->setIsInBounds(GEP->isInBounds());
+
+  return cast<GetElementPtrInst>(NewGEP);
```
martien-de-jong: Why don't we draw this into the previous if statement?

llvm/lib/Target/AIE/AIEOuterLoopPointerOptimizer.cpp
```diff
@@ -0,0 +1,1033 @@
...
+    const auto *UI = dyn_cast<Instruction>(U);
+    if (!UI || UI->getParent() != BB)
+      continue;
+    if (isa<LoadInst>(UI) || isa<StoreInst>(UI))
+      return true;
+    // Also check intrinsics that may be memory operations
+    if (const auto *CI = dyn_cast<CallInst>(UI)) {
+      if (CI->mayReadOrWriteMemory())
+        return true;
+    }
```
martien-de-jong: I'm wondering why we are interested in other memory operations. If it's for postincrement possibilities, I thinkwe should only look for loads and stores.
Also note that the function's comment says load/store.

llvm/lib/Target/AIE/AIEOuterLoopPointerOptimizer.cpp
```diff
@@ -0,0 +1,1033 @@
...
+static bool hasValidHoistableBase(GetElementPtrInst *GEP, BasicBlock *Top) {
+  Value *PtrOp = GEP->getPointerOperand();
+  auto *PtrInst = dyn_cast<Instruction>(PtrOp);
+
+  if (!PtrInst)
+    return false; // Not an instruction
+  if (PtrInst->getParent() != Top)
+    return false; // Not defined in Top
+  if (isa<PHINode>(PtrInst))
+    return false; // PHI is input, not produced
```
martien-de-jong: These trailing comments are superfluous.

llvm/lib/Target/AIE/AIEOuterLoopPointerOptimizer.cpp
martien-de-jong: I guess you are expecting these memory operations to be expanded to load/store.

llvm/lib/Target/AIE/AIEOuterLoopPointerOptimizer.cpp
martien-de-jong: Perhaps just mention that you try to avoid stand-alone pointer updates by grouping them with related memory operations.

llvm/lib/Target/AIE/AIEOuterLoopPointerOptimizer.cpp
```diff
@@ -0,0 +1,1033 @@
...
+  }
+  return true;
+}
+
+/// Check if all GEP indices are available at the end of Top block.
+static bool areIndicesAvailableInTop(GetElementPtrInst *GEP, BasicBlock *Top,
+                                     DominatorTree *DT) {
+  for (unsigned OpIdx = 1; OpIdx < GEP->getNumOperands(); ++OpIdx) {
+    const Value *Idx = GEP->getOperand(OpIdx);
+    if (isa<Constant>(Idx))
```
martien-de-jong: nit: Constants will be ignored by the next if

llvm/lib/Target/AIE/AIEOuterLoopPointerOptimizer.cpp
```diff
@@ -0,0 +1,1033 @@
...
+    return false; // Not defined in Top
+  if (isa<PHINode>(PtrInst))
+    return false; // PHI is input, not produced
+
+  return true;
+}
+
+/// Check if hoisting would interfere with post-increment folding.
+/// Returns true if the pointer is NOT used by memory ops in Bottom.
+static bool canHoistWithoutMemoryInterference(GetElementPtrInst *GEP,
```
martien-de-jong: I think the name could be more focused on detecting uses of the pointeroperand. 'extend live range' would also hold for arbitrary uses in Bottom.

llvm/lib/Target/AIE/AIEOuterLoopPointerOptimizer.cpp
```diff
@@ -0,0 +1,1033 @@
...
+    return false;
+
+  return true;
+}
+
+//===----------------------------------------------------------------------===//
+// GEP Address Space Canonicalization Helpers
+//===----------------------------------------------------------------------===//
+
+/// Rebuild a GEP to use a new base pointer (possibly in a different AS).
```
martien-de-jong: AS -> AddressSpace

llvm/lib/Target/AIE/AIEOuterLoopPointerOptimizer.cpp
```diff
@@ -0,0 +1,1033 @@
...
+findRootCastsFromPHI(PHINode &PHI, BasicBlock *CastBB) {
+  SmallVector<AddrSpaceCastInst *, 4> RootCasts;
+  const unsigned CanonicalAS = PHI.getType()->getPointerAddressSpace();
+
+  for (User *U : PHI.users()) {
+    auto *ASC = dyn_cast<AddrSpaceCastInst>(U);
+    if (!ASC || ASC->getParent() != CastBB)
+      continue;
+    // Must be casting away from canonical AS
+    if (ASC->getSrcAddressSpace() != CanonicalAS)
```
martien-de-jong, line 345: This looks unlikely. Is this allowed?

llvm/lib/Target/AIE/AIEOuterLoopPointerOptimizer.cpp
```diff
@@ -0,0 +1,1033 @@
...
+    Worklist.push_back(L);
+
+  while (!Worklist.empty()) {
+    Loop *L = Worklist.pop_back_val();
+    // Add subloops to worklist
+    for (Loop *SubL : *L)
+      Worklist.push_back(SubL);
+
+    // Try to optimize this loop
+    Changed |= runOnLoop(L);
```
martien-de-jong, line 679: The first loop we process is an outer loop.

llvm/lib/Target/AIE/AIEOuterLoopPointerOptimizer.cpp
```diff
@@ -0,0 +1,1033 @@
...
+
+  return Changed;
+}
+
+bool AIEOuterLoopPointerOptimizer::runOnLoop(Loop *L) {
+  LLVM_DEBUG(dbgs() << "OLPO: Analyzing loop with header "
+                    << L->getHeader()->getName() << "\n");
+
+  // Try to build the loop structure
+  auto LS = LoopStructure::tryBuildFrom(L, *LI);
```
martien-de-jong: Please expand auto here,.

llvm/lib/Target/AIE/AIEOuterLoopPointerOptimizer.cpp
```diff
@@ -0,0 +1,1033 @@
...
+bool AIEOuterLoopPointerOptimizer::canonicalizeGEPAddressSpace(
+    LoopStructure &LS) {
+  LLVM_DEBUG(dbgs() << "OLPO: Canonicalizing GEP address spaces to PHI's AS\n");
+
+  bool Changed = false;
+
+  BasicBlock *Top = LS.getTop();
+
+  // Process top (prologue) block - PHIs are in Top, casts are in Top
+  Changed |= canonicalizeGEPAddressSpaceInBlock(Top, Top);
```
martien-de-jong: Can we reverse the comment to match the argument order?

llvm/lib/Target/AIE/AIEOuterLoopPointerOptimizer.cpp
```diff
@@ -0,0 +1,1033 @@
...
+}
+
+/// Canonicalize GEP address spaces in BB using PHIs from PHIBlock.
+/// BB is where the casts and GEPs are located.
+/// PHIBlock is where the PHIs are located (loop header).
+bool AIEOuterLoopPointerOptimizer::canonicalizeGEPAddressSpaceInBlock(
+    BasicBlock *BB, BasicBlock *PHIBlock) {
+  bool Changed = false;
+
+  // State used for processing each root cast
```
martien-de-jong: Why are they called root casts?

llvm/lib/Target/AIE/AIEOuterLoopPointerOptimizer.cpp
```diff
@@ -0,0 +1,1033 @@
...
+
+  bool Changed = false;
+
+  BasicBlock *Top = LS.getTop();
+
+  // Process top (prologue) block - PHIs are in Top, casts are in Top
+  Changed |= canonicalizeGEPAddressSpaceInBlock(Top, Top);
+
+  // Process bottom (epilogue) block - PHIs are in Top, casts are in Bottom
+  Changed |= canonicalizeGEPAddressSpaceInBlock(LS.getBottom(), Top);
```
martien-de-jong: Can't we pass LS, and do the parent check of the casts against top and bottom?  We now traverse the same phi nodes twice.

llvm/lib/Target/AIE/AIEOuterLoopPointerOptimizer.cpp
```diff
@@ -0,0 +1,1033 @@
...
+                            GEPCanonState &State) {
+  const unsigned CanonicalAS = PHI.getType()->getPointerAddressSpace();
+  const unsigned TargetAS = RootCast->getDestAddressSpace();
+
+  LLVM_DEBUG(dbgs() << "OLPO:   Found cast away from canonical: " << *RootCast
+                    << "\n"
+                    << "         Target AS: " << TargetAS << "\n");
+
+  // Initialize state for this root cast
+  State.clear();
```
martien-de-jong: As far as I can see, we only retain BB, which is, AFAIK, RootCast->Parent(). Why don't we construct a fresh GEPCanonState here?

llvm/lib/Target/AIE/AIEOuterLoopPointerOptimizer.cpp
```diff
@@ -0,0 +1,1033 @@
...
+// 2. GEP Canonicalization: Convert non-i8 GEPs to i8-based GEPs for uniformity.
+//    This makes pointer arithmetic explicit and helps later optimizations.
+// 3. GEP Chain Linking: Link consecutive GEPs with the same base pointer to
+//    enable post-increment addressing patterns.
+// 4. GEP Hoisting: Move GEPs from bottom (epilogue) to top (prologue) block
+//    when safe, reducing code in the epilogue and improving scheduling.
+//
+//===----------------------------------------------------------------------===//
+
+#include "AIE.h"
```
martien-de-jong, line 29: Do we need AIE.h?

llvm/lib/Target/AIE/AIEOuterLoopPointerOptimizer.cpp
```diff
@@ -0,0 +1,1033 @@
...
+  Loop *const OuterLoop = LS.getOuterLoop();
+  BasicBlock *const Preheader = LS.getPreheader();
+
+  for (GetElementPtrInst *GEP : GEPsToProcess) {
+    Value *const Index = GEP->getOperand(1);
+    Type *const SrcElemTy = GEP->getSourceElementType();
+    const uint64_t ElemSize = DL->getTypeAllocSize(SrcElemTy);
+
+    // If element size is 1, no mul needed - use index directly
+    if (ElemSize == 1) {
```
martien-de-jong: Do we really need I8 type, or do we just don't want multiplication factors? We could also rewrite the 'I8' selection to use Elemsize.

llvm/lib/Target/AIE/AIEOuterLoopPointerOptimizer.cpp
```diff
@@ -0,0 +1,1033 @@
...
+      Value *NewGEP = replaceGEPWithI8Based(GEP, Index);
```
martien-de-jong: This code should be shared with the rest of replacement. Conversely, the ElmSize=1 case is part of getting the scaling right. I would refactor all this scaling logic below into a separate function computing me ByteOffset. For ElemSize==1 it would return Index unchanged.

llvm/lib/Target/AIE/AIEOuterLoopPointerOptimizer.cpp
```diff
@@ -0,0 +1,1033 @@
...
+      Changed = true;
+      continue;
+    }
+
+    // Get or create byte offset with CSE
+    Value *ByteOffset = nullptr;
+    auto Key = std::make_pair(Index, ElemSize);
+    bool IsLoopInvariant = OuterLoop->isLoopInvariant(Index);
+
+    if (IsLoopInvariant) {
```
martien-de-jong: This condition should just select the map to use and the block to insert in.

llvm/lib/Target/AIE/AIEOuterLoopPointerOptimizer.cpp
```diff
@@ -0,0 +1,1033 @@
...
+  DenseMap<Value *, GEPChainState> BaseChains;
+
+  // Collect GEPs in program order from Top and Bottom
+  SmallVector<GetElementPtrInst *, 16> AllGEPs = collectGEPs(Top);
+  AllGEPs.append(collectGEPs(Bottom));
+
+  // Process GEPs in order
+  for (GetElementPtrInst *GEP : AllGEPs) {
+    // Check if GEP is still valid (might have been erased)
+    if (GEP->getParent() == nullptr)
```
martien-de-jong: You can't dereference an erased instruction.

llvm/lib/Target/AIE/AIEOuterLoopPointerOptimizer.cpp
```diff
@@ -0,0 +1,1033 @@
...
+    }
+
+    Value *Base = GEP->getPointerOperand();
+
+    // Look up the chain state for this base
+    auto It = BaseChains.find(Base);
+    if (It == BaseChains.end()) {
+      // This is the first GEP with this base - start a new chain
+      // But only if the base is a phi (start of chain) or an argument
+      if (isa<PHINode>(Base) || isa<Argument>(Base)) {
```
martien-de-jong: This looks a bit incomplete as a feasibility check. An Argument would be invariant in the outer loop and would introduce a copy. A PHI in Bottom would be awkward. I would have thought that you would require a GEP chain starting in a PHI leading back across the outer loop backedge.

llvm/lib/Target/AIE/AIEOuterLoopPointerOptimizer.cpp
andcarminati: Actually, this is considered a legal cast, despite of being strange.

llvm/lib/Target/AIE/AIEOuterLoopPointerOptimizer.cpp
andcarminati: As we use custom intrinsics for multidimensional memory address calculation, it is currently not necessary. I would prefer to keep the implementation simple as much as possible.

llvm/lib/Target/AIE/AIEOuterLoopPointerOptimizer.cpp
andcarminati: I removed this check.

llvm/lib/Target/AIE/AIEOuterLoopPointerOptimizer.cpp
andcarminati: I changed to canHoistInterfereWithPostIncFolding.

llvm/lib/Target/AIE/AIEOuterLoopPointerOptimizer.cpp
andcarminati: Yes, I changed the comment.

llvm/lib/Target/AIE/AIEOuterLoopPointerOptimizer.cpp
andcarminati: They are called "root" because:

1. They are __directly attached to the PHI__ (the root of the def-use chain)
2. They are the __starting point__ for the transformation - the algorithm walks from the root cast down through GEPs
3. There may be __chained GEPs__ below them, but the cast is where the address space divergence begins

I changed to `/// State for canonicalizing GEPs that use a cast from a PHI.`

llvm/lib/Target/AIE/AIEOuterLoopPointerOptimizer.cpp
andcarminati: Done!

llvm/lib/Target/AIE/AIEOuterLoopPointerOptimizer.cpp
andcarminati: We need because of the initialization function:

```
void llvm::initialize##passName##Pass(PassRegistry &Registry)
```

llvm/lib/Target/AIE/AIEOuterLoopPointerOptimizer.cpp
andcarminati: We want the multiplication outside the MBB after IRTranslator, as G_PTR_ADD is a plain pointer add, and the lowering/translation will require a mul/shift that will sit in the combiner path.

llvm/lib/Target/AIE/AIEOuterLoopPointerOptimizer.cpp
martien-de-jong: What I mean is that this is casting the PHI's result, so the PHI is the Src. I find it strange that we can represent the source addressspace separate from the addressspace of the source.

llvm/lib/Target/AIE/AIEOuterLoopPointerOptimizer.cpp
andcarminati: Done!

llvm/lib/Target/AIE/AIEOuterLoopPointerOptimizer.cpp
andcarminati: I simplified the code.

llvm/lib/Target/AIE/AIEOuterLoopPointerOptimizer.cpp
martien-de-jong: Sure, but the mul/shift come in if ElemSize != 1. This is a different condition from type == i8 which we enforce elsewhere. Just a nit.

llvm/lib/Target/AIE/AIEOuterLoopPointerOptimizer.cpp
andcarminati: I am using a ToErase set now.

llvm/lib/Target/AIE/AIEOuterLoopPointerOptimizer.cpp
andcarminati: isa<Argument>(Base) was intended for a case were we store counters back to function parameters, chaining can lead to small encodable offsets. For Phi nodes, you are correct, I will check parent against top.

martien-de-jong (dismissed):

martien-de-jong (approved):

PR #1248, opened 2026-08-14
[ArithToEmitC] Add lowering for arith.{min,max}{s,u}i

hunhoffe:
ArithToEmitC already lowers arith.select directly to emitc.conditional, and ShiftOpConversion emits a compare-and-select ternary to guard against UB on out-of-range shift amounts. Despite proving it can emit emitc.conditional, the pass had no pattern at all for arith.minsi/ maxsi/minui/maxui, even though min/max is exactly the same shape (`a < b ? a : b`) without any UB to guard against.

Add MinMaxOpConversion, modeled on CmpIOpConversion's signed/unsigned handling, and register it for all four integer min/max ops. Floating point min/max (minimumf/maximumf/minnumf/maxnumf) are intentionally left out of scope due to NaN-propagation semantics.


ehsan-toosi (approved):

ehsan-toosi:
Thanks. You can merge.

PR #1250, opened 2026-08-17
[AIE][OLP] Gate LCSSA restoration to downstream-exit live-outs

khallouh:
- Fixes regression in key gemm kernel (not the one in the test updates)
- Preserves fixed behavior in gemm_mx6 to avoid poison values

`performTransformation` unconditionally called `formLCSSARecursively`. That was
introduced to fix exit phis reading poison when multiple loops reach a common
exit through intermediate exit blocks.
The call is only needed for that structure. When every loop live-out is
consumed in the loop's *immediate* exit block, `reroutePhiIncomings` and the
non-phi exit-operand remap already produce correct code.

### DIFF
--- a/llvm/lib/Target/AIE/AIEOuterLoopPipeliner.cpp
+++ b/llvm/lib/Target/AIE/AIEOuterLoopPipeliner.cpp
@@ -1397,6 +1397,24 @@ void AIEOuterLoopPipeliner::remapExitForReplacement(
   reroutePhiIncomings(OrigExit, OrigLS.getBottom(), ReplacementLS.getBottom(),
                       PhiEdge::Repoint,
                       [&](Value *V) { return ReplacementLS.cloneOf(V); });
+
+  // Non-phi exit live-outs: when the latch dominates the exit there is no LCSSA
+  // phi, so LCSSA rematerializes the live-out as a plain instruction in the
+  // exit whose operands reference loop-internal values (e.g. the outer-header
+  // PHI and an inner-loop def). removeFromCFG deletes those originals, which
+  // would leave the exit reading poison. Remap each such operand to a live
+  // clone so the exit reads a defined value.
+  for (Instruction &I :
+       make_range(OrigExit->getFirstNonPHIIt(), OrigExit->end())) {
+    for (Use &Op : I.operands()) {
+      Value *V = Op.get();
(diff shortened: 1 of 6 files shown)


llvm/lib/Target/AIE/AIEOuterLoopPipeliner.cpp
```diff
@@ -1397,6 +1397,24 @@ void AIEOuterLoopPipeliner::remapExitForReplacement(
   reroutePhiIncomings(OrigExit, OrigLS.getBottom(), ReplacementLS.getBottom(),
                       PhiEdge::Repoint,
                       [&](Value *V) { return ReplacementLS.cloneOf(V); });
+
+  // Non-phi exit live-outs: when the latch dominates the exit there is no LCSSA
```
andcarminati, line 1401: I see some conceptual problems with this comment. LCSSA only inserts PHI nodes at loop exit blocks. It never creates "plain instructions" that recompute values. The LCSSA pass strictly adds PHI nodes with one incoming value per predecessor edge. Even when the latch dominates the exit, LCSSA would still insert a PHI node if a loop-defined value is used outside. The PHI would simply have a single incoming edge from the latch. The fact that the latch dominates doesn't cause LCSSA to behave differently (other than the PHI having fewer incoming values). See `[[PHI6:%[0-9]+]]:_(s32) = G_PHI [[ADD]](s32), %bb.4`

llvm/lib/Target/AIE/AIEOuterLoopPipeliner.cpp
```diff
@@ -1397,6 +1397,24 @@ void AIEOuterLoopPipeliner::remapExitForReplacement(
...
+  // phi, so LCSSA rematerializes the live-out as a plain instruction in the
+  // exit whose operands reference loop-internal values (e.g. the outer-header
+  // PHI and an inner-loop def). removeFromCFG deletes those originals, which
+  // would leave the exit reading poison. Remap each such operand to a live
+  // clone so the exit reads a defined value.
+  for (Instruction &I :
```
andcarminati, line 1407: Alternatively, can I read this as a fix for the cases where `needsLCSSAForExit` returns false and we skip formLCSSARecursively?

khallouh:
Closing in favor of https://github.com/Xilinx/llvm-aie/pull/1251

PR #1251, opened 2026-08-17
[AIE][OLP] Simplify trivial LCSSA PHIs to avoid register copies

andcarminati:
After restoring LCSSA form, simplify trivial PHI nodes at loop exits to prevent unnecessary register moves during code generation.


khallouh (approved):
LGTM

PR #1252, opened 2026-08-19
[AIEX] Bailout spill slots with untracked accesses in AIESpillSlotOptimization

niwinanto:
AIESpillSlotOptimization only understands spill/reload pseudos that isStoreToStackSlot()/isLoadFromStackSlot() recognize; analyzeFunction() collects its spill points through those hooks. The downstream target VST_PLFR_SPILL /VLDA_PLFR_SPILL pseudos have a non-standard operand layout (frame index at operand 3, multiple register operands, two defs on reload) and are therefore not recognized, so the pass cannot see or rewrite them. AIE2PS has the similar issue but the pass is not running since we miss to implement some hooks.

StackSlotColoring can coalesce such an untracked PLFR access onto a slot that also carries a decomposable spill (e.g. VST_DM_SPILL). The pass then decomposes and removes that slot while updating only the accesses it tracked, leaving the
PLFR access pointing at a dead stack object. prologepilog later asserts in eliminateFrameIndex ("Getting frame offset for a dead object?").

Detect frame-index operands the pass cannot rewrite (SlotsWithUntrackedAccess) and exclude those slots from decomposition, leaving them intact.

### DIFF
--- a/llvm/lib/Target/AIE/AIESpillSlotOptimization.cpp
+++ b/llvm/lib/Target/AIE/AIESpillSlotOptimization.cpp
@@ -196,6 +196,7 @@ private:
   // Phase 1: Analysis
   void analyzeFunction(MachineFunction &MF);
   bool isValidSpillSlot(int FI, MachineInstr &MI) const;
+  DenseSet<int> collectUntrackableAccesses(MachineFunction &MF) const;
 
   // Phase 2: Rewrite
   bool rewriteInstructions(MachineFunction &MF);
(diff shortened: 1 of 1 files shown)


llvm/lib/Target/AIE/AIESpillSlotOptimization.cpp
```diff
@@ -381,10 +394,44 @@ bool AIESpillSlotOptimization::isValidSpillSlot(int FI,
...
+
+/// Record slots that have an untracked access (see isTrackedAccess()), so
+/// rewriteInstructions() can skip decomposing them. Debug instructions are
+/// ignored so the decision does not depend on -g.
+void AIESpillSlotOptimization::collectUntrackedAccesses(MachineFunction &MF) {
+  for (MachineBasicBlock &MBB : MF) {
+    for (MachineInstr &MI : MBB) {
+      if (MI.isDebugInstr())
+        continue;
+      for (const MachineOperand &MO : MI.operands()) {
```
martien-de-jong, line 403: nit: we could also have an early continue for any LoadFromStackSlot/StoreToStackSlot. We are left with untrackable accesses, and we only have to check MO.isFI() on any of the operands.

llvm/lib/Target/AIE/AIESpillSlotOptimization.cpp
```diff
@@ -381,10 +394,44 @@ bool AIESpillSlotOptimization::isValidSpillSlot(int FI,
...
+      }
+    }
+  }
+}
+
 void AIESpillSlotOptimization::analyzeFunction(MachineFunction &MF) {
   LLVM_DEBUG(dbgs() << "=== Phase 1: Analysis ===\n");
 
   OrigSlotDecompositions.clear();
+  SlotsWithUntrackedAccess.clear();
```
martien-de-jong: SlotsWithUntrackableAccesses can be local to this function. How about

`const SlotsWithUntrackableAccess = collectUntrackedAccesses()`

niwinanto: It sounds better, done!

llvm/lib/Target/AIE/AIESpillSlotOptimization.cpp
niwinanto: Done!

martien-de-jong (approved):

PR #1253, opened 2026-08-20
Martien.modulo scoreboard

martien-de-jong:
This adds a true modulo mode to ResourceScoreboard. Refactoring the postpipeliner with this, removes all nitty gritty details of emulating a modulo scoreboard in a linear scoreboard, for instance the final check in scheduleOtherIterations where we bring the linear scoreboard up to steady state.
The PR is mostly NFC. The unit tests show slightly different schedules, which are caused by preventing infeasible schedules to be proposed to scheduleOtherIterations, avoiding conflicts in scheduleFirstIteration instead.
We gain readibility, we have less rejections in the final round and, last but not least, logging of the modulo scoreboard is much more concise and understandable.

### DIFF
--- a/llvm/include/llvm/CodeGen/ResourceScoreboard.h
+++ b/llvm/include/llvm/CodeGen/ResourceScoreboard.h
@@ -4,7 +4,7 @@
 // See https://llvm.org/LICENSE.txt for license information.
 // SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
 //
-// (c) Copyright 2023-2025 Advanced Micro Devices, Inc. or its affiliates
+// (c) Copyright 2023-2026 Advanced Micro Devices, Inc. or its affiliates
 //
 //===----------------------------------------------------------------------===//
 //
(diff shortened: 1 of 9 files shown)


llvm/lib/Target/AIE/AIEPostPipeliner.cpp
```diff
@@ -45,6 +45,11 @@ static cl::opt<int>
     HeuristicRuns("aie-postpipeliner-heuristic-runs",
                   cl::desc("Number of runs for heuristics that converge"),
                   cl::init(20), cl::Hidden);
+static cl::opt<std::string>
+    DisabledStrategyPrefix("aie-postpipeliner-disable-strategy",
+                           cl::desc("Disable all post-pipeliner strategies"
+                                    " whose name starts with this prefix"),
+                           cl::init(""), cl::Hidden);
```
martien-de-jong, line 52: We had an existing heuristic selection option, but that only controlled the 'config' ones by a magic index. We add a generic 'disable-by-name' which can also disable the non-table driven ones.

llvm/include/llvm/CodeGen/ResourceScoreboard.h
```diff
@@ -86,15 +101,38 @@ template <typename RC> class ResourceScoreboard {
...
+    Period = P;
+    Size = P;
+    LowestCycle = 0;
+    HighestCycle = P - 1;
+    Head = 0;
+    Cycles.assign(P, RC{});
+  }
+
   void advance() {
+    if (Period > 0)
```
martien-de-jong, line 129: Hmm. Actually looks awkward. I guess I would like to make this fatal.

andcarminati (approved):
LGTM. It brings in a nice simplification.

PR #1256, opened 2026-08-20
[AIEX] Guard matchMostlySequentialShuffleWithInsertions to only create legal G_AIE_UNPAD_VECTOR

konstantinschwarz:
This was missing a isLegalTypeToUnpad check, potentially creating illegal G_AIE_UNPAD_VECTOR that cannot be selected later.


niwinanto (approved):
LGTM!

PR #1259, opened 2026-08-21
[AIEX] Add sibling loop register alignment pass

andcarminati:
Add pass to align register allocations between sibling inner loops created by Outer Loop Pipeliner. When OLP creates steady and lastiter loops, they may receive different register allocations. This pass rewrites the peeled iteration loop to match the steady loop's registers where possible, reducing register divergence.

Enabled by default, controlled via -aie-sibling-loop-reg-align flag.

This is intended to enable further scheduling optimizations.


llvm/lib/Target/AIE/AIESiblingLoopRegAligner.cpp
```diff
@@ -0,0 +1,367 @@
...
+
+  LLVM_DEBUG(dbgs() << "Realigned " << printReg(VReg, TRI) << " from "
+                    << TRI->getName(CurrentPhys) << " to "
+                    << TRI->getName(TargetPhysReg) << "\n");
+  return true;
+}
+
+bool AIESiblingLoopRegAligner::alignInstructionPair(
+    MachineInstr *SteadyMI, MachineInstr *PeeledMI,
+    MachineBasicBlock *SteadyLoop, MachineBasicBlock *PeeledLoop,
```
martien-de-jong: Can't we use SteadyMI->getParent() and PeeledMI->getParent() ?

llvm/lib/Target/AIE/AIESiblingLoopRegAligner.cpp
```diff
@@ -0,0 +1,367 @@
...
+    unsigned &Aligned, unsigned &Skipped) {
+  bool Modified = false;
+  const unsigned NumOps =
+      std::min(SteadyMI->getNumOperands(), PeeledMI->getNumOperands());
```
martien-de-jong: Do we need to bother if they are not the same?

llvm/lib/Target/AIE/AIESiblingLoopRegAligner.cpp
```diff
@@ -0,0 +1,367 @@
...
+    return false;
+
+  const LiveInterval &LI = LIS->getInterval(VReg);
+
+  // Get the slot index range for the loop block
+  const SlotIndex LoopStart = LIS->getMBBStartIdx(LoopMBB);
+  const SlotIndex LoopEnd = LIS->getMBBEndIdx(LoopMBB);
+
+  // Check each segment of the live range
+  for (const LiveRange::Segment &Seg : LI) {
```
martien-de-jong: return all_of() ?

llvm/lib/Target/AIE/AIESiblingLoopRegAligner.cpp
```diff
@@ -0,0 +1,367 @@
...
+      return false;
+  }
+
+  return true;
+}
+
+bool AIESiblingLoopRegAligner::shouldAttemptAlignment(
+    const MachineOperand &SteadyOp, const MachineOperand &PeeledOp,
+    MachineBasicBlock *SteadyLoop, MachineBasicBlock *PeeledLoop) {
+  // Both operands must be register definitions
```
martien-de-jong: This is a silly breadth first comparison, AI.  Why not have a lambda IsLocalRegDefinition, and check them one by one.

llvm/lib/Target/AIE/AIESiblingLoopRegAligner.cpp
```diff
@@ -0,0 +1,367 @@
...
+}
+
+bool AIESiblingLoopRegAligner::tryReassignReg(Register VReg,
+                                              MCPhysReg TargetPhysReg) {
+  if (!VRM->hasPhys(VReg))
+    return false;
+
+  const MCPhysReg CurrentPhys = VRM->getPhys(VReg);
+  if (CurrentPhys == TargetPhysReg)
+    return true;
```
martien-de-jong: Nit: the return value is used to set Modified, whereas we haven't modified anything here.

llvm/lib/Target/AIE/AIESiblingLoopRegAligner.cpp
```diff
@@ -0,0 +1,367 @@
...
+
+  // Validate register class compatibility
+  const TargetRegisterClass *RC = MRI->getRegClass(VReg);
+  if (!RC->contains(TargetPhysReg))
```
martien-de-jong, line 255: CHECK: Does RC represent the smallest set that is compatible with all its uses?

llvm/lib/Target/AIE/AIESiblingLoopRegAligner.cpp
```diff
@@ -0,0 +1,367 @@
...
+        getPhysRegsIfAlignable(SteadyOp.getReg(), PeeledOp.getReg());
+    if (!PhysRegs)
+      continue;
+
+    const auto [SteadyPhys, PeeledPhys] = *PhysRegs;
+    const Register PeeledReg = PeeledOp.getReg();
+
+    if (tryReassignReg(PeeledReg, SteadyPhys)) {
+      Modified = true;
+      Aligned++;
```
martien-de-jong: Is there gain in trying the reverse alignment?

llvm/lib/Target/AIE/AIESiblingLoopRegAligner.cpp
```diff
@@ -0,0 +1,367 @@
...
+  bool Modified = false;
+  const unsigned NumOps =
+      std::min(SteadyMI->getNumOperands(), PeeledMI->getNumOperands());
+
+  for (unsigned I = 0; I < NumOps; ++I) {
+    const MachineOperand &SteadyOp = SteadyMI->getOperand(I);
+    MachineOperand &PeeledOp = PeeledMI->getOperand(I);
+
+    if (!shouldAttemptAlignment(SteadyOp, PeeledOp, SteadyLoop, PeeledLoop))
+      continue;
```
martien-de-jong: I think this loop can be restricted to numdefs.

llvm/lib/Target/AIE/AIESiblingLoopRegAligner.cpp
```diff
@@ -0,0 +1,367 @@
...
+SmallVector<std::pair<MachineInstr *, MachineInstr *>, 64>
+AIESiblingLoopRegAligner::buildInstructionMapping(
+    const AIELoopUtils::OuterLoopStructure &OLS) {
+  SmallVector<std::pair<MachineInstr *, MachineInstr *>, 64> Mapping;
+
+  auto SI = OLS.SteadyInner->begin();
+  auto PI = OLS.PeeledIterInner->begin();
+
+  for (; SI != OLS.SteadyInner->end() && PI != OLS.PeeledIterInner->end();
+       ++SI, ++PI) {
```
martien-de-jong, line 220: I think that we should not push pairs that have different opcodes. Something caused the difference, and forcing the algorithm in this situation can't improve the result.

llvm/lib/Target/AIE/AIESiblingLoopRegAligner.cpp
```diff
@@ -0,0 +1,367 @@
...
+//
+// (c) Copyright 2026 Advanced Micro Devices, Inc. or its affiliates
+//
+//===----------------------------------------------------------------------===//
+//
+// This pass aligns register allocations between sibling inner loops created
+// by the Outer Loop Pipeliner (OLP). OLP creates a "steady" inner loop and
+// a "peeled iteration" inner loop that execute the same code, but may receive
+// different register allocations. This pass rewrites the peeled iteration
+// loop's allocations to match the steady loop where possible.
```
martien-de-jong, line 15: Perhaps add that we do this to find more common code in the prologue generated by postpipelining.

llvm/lib/Target/AIE/AIESiblingLoopRegAligner.cpp
andcarminati: It may be an overkill indeed.

llvm/lib/Target/AIE/AIESiblingLoopRegAligner.cpp
andcarminati: You gave me an idea, if one direction fails, we may try the opposite.

llvm/lib/Target/AIE/AIESiblingLoopRegAligner.cpp
andcarminati: We already checked this in isValidForAlignment.

llvm/lib/Target/AIE/AIESiblingLoopRegAligner.cpp
andcarminati: The sibling loops have matching structure (same instruction patterns) If SteadyReg was allocated to SteadyPhys, and both registers have the same opcode, TargetPhysReg should be valid for PeeledReg's uses too

llvm/lib/Target/AIE/aie2/AIE2TargetMachine.cpp
```diff
@@ -200,6 +201,8 @@ bool AIE2PassConfig::addRegAssignAndRewriteOptimized() {
     addPass(createAIEWawRegRewriter());
     addPass(createGreedyRegisterAllocator());
   }
+  if (EnableSiblingLoopRegAlign)
+    addPass(createAIESiblingLoopRegAligner());
```
martien-de-jong: Could it be that Fabian is sharing this code between aie2 variants in the WAR resolver PR?

llvm/lib/Target/AIE/AIESiblingLoopRegAligner.cpp
```diff
@@ -0,0 +1,419 @@
...
+char AIESiblingLoopRegAligner::ID = 0;
+
+bool AIESiblingLoopRegAligner::initializeAnalyses(MachineFunction &Fn) {
+  MF = &Fn;
+  MRI = &Fn.getRegInfo();
+  TRI = static_cast<const AIEBaseRegisterInfo *>(MRI->getTargetRegisterInfo());
+
+  auto *VRMW = getAnalysisIfAvailable<VirtRegMapWrapperLegacy>();
+  auto *LRMW = getAnalysisIfAvailable<LiveRegMatrixWrapperLegacy>();
+  auto *LISW = getAnalysisIfAvailable<LiveIntervalsWrapperPass>();
```
martien-de-jong: I find to strange to use IfAvailable here. If they are not available the static configurations of our passes is wrong. AFAIK, IfAvailable is for things like expensive alias analysis, which is under control of optimization level.

llvm/lib/Target/AIE/AIESiblingLoopRegAligner.cpp
```diff
@@ -0,0 +1,419 @@
...
+
+  LLVM_DEBUG(dbgs() << "Realigned " << printReg(VReg, TRI) << " from "
+                    << TRI->getName(CurrentPhys) << " to "
+                    << TRI->getName(TargetPhysReg) << "\n");
+}
+
+SmallVector<ReassignmentCandidate, 32>
+AIESiblingLoopRegAligner::collectReassignmentCandidates(
+    const SmallVectorImpl<std::pair<MachineInstr *, MachineInstr *>> &InstMap,
+    const MachineBasicBlock *SteadyLoop, const MachineBasicBlock *PeeledLoop,
```
martien-de-jong: Curious: is this being called future-proof for the first-iteration peel?

PR #1268, opened 2026-08-26
[AIE] Pipeline single-stage ZOL loops and fix a memory-dependency sentinel

mludevid:
The post-pipeliner required iteration count metadata to know the minimum trip count. A ZOL cannot express zero iterations, so the body always runs at least once; use that guaranteed minimum of 1 when the metadata is absent.

getMaxAliasingMemCycle() returned 0 both for an aliasing memory operation at cycle 0 and for no aliasing operation at all, so checkMemoryDependency() could not tell the two apart and ordered against a store that does not exist. Return INT_MIN when nothing aliases, consistent with every other accessor in AIERegMemEventTracker.


llvm/test/CodeGen/AIE/aie2p/end-to-end/ore-hardware-loops.ll
```diff
@@ -34,7 +34,7 @@ define weak_odr dso_local void @convert_bf16_to_bfp16(ptr noalias %in, ptr noali
 ; REMARKS-NEXT:   - BasicBlock:      entry
 ; REMARKS-NEXT:   - RegionIndex:     '0'
 ; REMARKS-NEXT:   - BundleCount:     '20'
-; REMARKS-NEXT:   - ByteCount:       '144'
+; REMARKS-NEXT:   - ByteCount:       '160'
```
mludevid, line 37: These PM regressions are unfortunate.

llvm/test/CodeGen/AIE/aie2ps/end-to-end/loop_4_biased_conv_mx6x6_ebs16_unroll_impl_s2_accumulate.ll
```diff
@@ -56,7 +56,7 @@ define dso_local void @loop_4_biased_conv_mx6x6_ebs16_unroll_impl_s2_accumulate(
 ; REMARKS-NEXT:   - NS:              '3'
 ; REMARKS-NEXT:   - Loop:            bb.2.for.body82.i
 ; REMARKS-NEXT:   - Prologue:        bb.1.for.body.i254
-; REMARKS-NEXT:   - PrologueBundles: '29'
```
andcarminati, line 59: I guess OLP is not enabled for this case (missing metadata).

andcarminati:
I like this change! Could you please separate the two changes in two different commits?

llvm/test/CodeGen/AIE/aie2p/end-to-end/ore-hardware-loops.ll
andcarminati: Don't worry about this 16 bytes.

mludevid: Yeah, I won't. I was just surprised by this regression. I guess it's related to our alignment pass not working optimally when padding to meet distance requirements for the ZOL setup instructions.

mludevid:
> I like this change! Could you please separate the two changes in two different commits?

Agreed, I think it makes sense. I've pushed the changes.

PR #1269, opened 2026-08-26
[WIP][AIE2P] Add XFAIL test for dropped loop-carried accumulator update

hunhoffe:
-- do not merge! -- 

# [AIE2P] Loop-carried accumulator update is dropped at -O2

A reduction loop can lose its accumulation entirely: the accumulator is copied *into* a working register at the loop head and accumulated there, but never copied back to the loop-carried register. This branch adds an `XFAIL` regression test that captures the issue.

## The defect

The IR is correct: in `%for.body4.i` the SROA-split accumulator halves are well-formed loop-carried phis fed by the MAC result. The generated code can't honor them:
```
        vmov    bmll1, x0        ; preheader: cml1 = 0 -- the ONLY writes to cml1
.LBB0_2:
        vmov    cml3, cml1       ; copy IN
        vmac.f  dm3, dm3, x6, x8, r4
// %bb.6:
        jz      r20, #.LBB0_2    ; back edge, no copy-out -- next iter re-reads 0
```

## Trigger

The helper must be inlined into the function *declaring* the accumulator array. Out of line, the address escapes, the alloca isn't promotable, and the load/store round-trip keeps it correct. Inlining makes it promotable — legal, and the IR is correct — so the defect is purely in codegen. `noinline` on the helper or its caller avoids it.

### DIFF
--- a/llvm/test/CodeGen/AIE/aie2p/accumulator-loop-carried.ll
+++ b/llvm/test/CodeGen/AIE/aie2p/accumulator-loop-carried.ll
@@ -1,0 +1,300 @@
+; This file is licensed under the Apache License v2.0 with LLVM Exceptions.
+; See https://llvm.org/LICENSE.txt for license information.
+; SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+;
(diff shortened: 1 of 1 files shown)


hunhoffe:
Fixed in different PR.

PR #1271, opened 2026-08-27
[AIE2PS] Fix assembly parser for bfp16 ex operands

konstantinschwarz:
The instruction is defined using the corresponding el register, but round-trips using the ex register.


andcarminati (approved):
LGTM.

PR #1274, opened 2026-08-31
[AIEX] Do not query the physreg of a de-allocated VReg in WAW rewriting

mludevid:


llvm/test/CodeGen/AIE/aie2ps/ra/waw_reg_renaming_aggressive_livethrough.mir
```diff
@@ -0,0 +1,70 @@
...
+# high-latency instruction: getHighOutputLatencyRegs then calls VRM->getPhys()
+# on the now-unmapped VReg and feeds NoRegister to MCRegAliasIterator.
+#
+# %2 and %3 are live-through bb.1, which has no renaming candidates, so nothing
+# re-assigns them before bb.3 is processed. %3 is defined in bb.3 by a mayLoad
+# instruction, which is what triggers the unguarded lookup.
+#
+# Starts before greedy so the input shape is pinned and does not depend on
+# whether PeepholeOpt folds the sub-register COPY into the REG_SEQUENCE.
+#
```
SagarMaheshwari99, line 12: this should be after copy right?

mludevid: The peephole optimization is the downstream trigger of this bug. I've taken this part of the comment out as it is not relevant to the test.

andcarminati (approved):
It makes sense! Thank you for the fix!

PR #1276, opened 2026-08-31
[AIE2P] Narrow the ePSRFLdF composite spill fallback to the range case

atassis:
## Problem

`AIE2PRegisterInfo::eliminateFrameIndex`'s ePSRFLdF composite-spill fallback in #1209 treats a
misaligned in-range offset the same as an out-of-range one:
`isEncodableAsNegativeInt<9,4>(Offset) && Offset % 64 == 0`. ePSRFLdF's spill alignment is 64
bytes by TableGen (`AIE2PVector1076FifoRegisterClass = AIE2PRegisterClass<1088, 512, ...>`), so
a register-allocated slot always keeps the FIFO sub-spill's offset a multiple of 64, so a
misaligned offset means the frame object itself is wrong, and routing it to indexed addressing
does not fix that: the FIFO store's 512-bit access truncates a misaligned hardware address
instead of faulting, so it would silently write to the wrong 64-byte-aligned address instead of
failing at encode time (the same objection khallouh raised on #1155 for `VST_dmx_sts_x_spill`).

## Fix

Drop the misalignment branch and assert the invariant instead, keeping only the range fallback.
The assert has to gate nothing else: an earlier version of this patch asserted `IsAligned` and
also tested it in the `if`, so under `-DNDEBUG` the assert compiled away and the untested
conjunct did not, silently reintroducing the exact indexed-fallback routing this patch removes,
in exactly the case the assert exists to catch. CI builds with assertions on, so no test can
see that divergence. Fixed by asserting the expression directly and leaving the `if` on the
range check alone, so a misaligned offset in a release build now hits the encoder ("immediate
operand value is not a multiple of 64") instead of addressing the wrong slot.

## Test/Evidence

Replaced the hand-built under-aligned `fixedStack` case (now trips the assert) with
`test_epsrfldf_prologepilog_bigstack`: a 64-aligned composite pushed past the 12-bit range by a
filler object, the one fallback case that is reachable.

CodeGen/AIE + MC/AIE, `LLVM_ENABLE_ASSERTIONS=ON` (matches CI, rebuilt fresh against current
`upstream/aie-public`): 2256 passed, 5 XFAIL, 3 unsupported, 0 failed.

Also built an `LLVM_ENABLE_ASSERTIONS=OFF` (release/NDEBUG) `llc` and ran the old under-aligned
case both sides of this commit, since assertions-on CI cannot exercise this path at all. Before
this commit: `llc -run-pass=prologepilog` on the misaligned MIR exits 0 and silently emits
`VST_dmx_sts_fifohl_idx $lfl0, $p1, $dj0` (indexed addressing on offset -184), and
`-filetype=obj` encodes it, exit 0, no diagnostic. After this commit, same input, same NDEBUG
build: `LLVM ERROR: immediate operand value -184 is not a multiple of 64`, exit 1.

This stacks on #1209 (open); happy to rebase onto `aie-public` once that merges, or have both
reviewed together.


atassis:
Folding this into #1155 and closing here.

Both change the same function, `AIE2PRegisterInfo::eliminateFrameIndex`, and they are one story: #1155
adds the encodability check the spill case-groups were missing, and this PR is the part that says the
ePSRFLdF composite case must not take the indexed fallback at all when the offset is misaligned,
because that would encode a truncated address cleanly.

Reviewing them separately would have meant reading the same switch twice, and #1155 already carries
the review context for why the indexed form is the wrong recovery in that case.

Nothing is dropped: the narrowed composite-spill case and its
`spill/spill-reload-psrf-prologepilog.mir` test are both in #1155 now.

PR #1281, opened 2026-09-07
Fixing the <2 x i8> ABI incompatibility and legalize generic opcodes

khallouh:
The ABI requires to pass a <2 x i8> packed in consecutive bytes of a GPR. We were element-promoting it to <2 x i16>, one byte per 16-bit lane. Making getPreferredVectorAction return TypeWidenVector instead of the default TypePromoteInteger for fixed-length vectors narrower than a register fixes that: a <2 x i8> now travels as a packed <4 x i8>. <2 x i8> is currently the only type this affects.

That change pulled in the following chain of legalizer work:

- G_UNMERGE_VALUES splitting a <4 x s8> into two <2 x s8> now appears on every <2 x i8> argument.

- G_EXTRACT_VECTOR_ELT from a <2 x s8> had to move off moreElements() otherwise we get an infinite loop.

- The custom G_BITCAST lowering between S16 and V2S8 is gone; the bitcast is now a plain reinterpretation and legal. The old lowering miscompiled: converting a <2 x s8> to i16 kept element 0 and dropped element 1. First commit is a caller/callee reproducer where the caller writes the second byte into bits 31:16 and the callee returns bits 15:0.

- G_ZEXT from <2 x s8> to <2 x s16> needs a custom lowering now. Under promotion the source was a G_TRUNC of the very <2 x s16> being produced, so the combiner rewrote zext(trunc(x)) into a per-lane AND with 0xff and both artifacts were gone.

- G_BUILD_VECTOR into a <2 x s8> also needs a custom lowering. This is needed to lower a standalone insertelement into a <2 x i8> which only ever worked if there were a following extractelement, thanks to the artifact combiner. Any other case where the user was not an extractelement was crashing before.

- G_TRUNC from <2 x s16> and G_ANYEXT to <2 x s16> no longer reach the legalizer. We never legalized them, the artifact combiner folded them away if they would come from the IRTranslator, but otherwise we crashed (e.g. if we write them explicitly in LLVM IR). I added custom lowerings for both for completeness.


llvm/lib/Target/AIE/aie2/AIE2LegalizerInfo.cpp
```diff
@@ -217,8 +217,17 @@ AIE2LegalizerInfo::AIE2LegalizerInfo(const AIE2Subtarget &ST) : AIEHelper(ST) {
   // G_PTRTOINT), the only legal extension is S20 -> S32.
   // Extensions to types larger than 64 bits have to be broken down into
   // multiple parts.
-  getActionDefinitionsBuilder({G_ANYEXT, G_SEXT, G_ZEXT})
+  // A packed sub-register vector such as <2 x s8> reaches its scalar form
```
mludevid, line 220: Nit: Maybe also mention these AIE2 changes in the commit message.

llvm/lib/Target/AIE/aie2p/AIE2PISelLowering.cpp
```diff
@@ -196,5 +196,10 @@ AIE2PTargetLowering::getPreferredVectorAction(MVT VT) const {
   if (VT.is128BitVector())
     return TypeWidenVector;
 
+  // Keep sub-register vectors packed, e.g. v2i8 widens to v4i8 rather than
+  // being element-promoted to v2i16, matching aie2 and the in-memory layout.
+  if (VT.isFixedLengthVector() && VT.getFixedSizeInBits() < 32)
```
mludevid: The PR and commit descriptions only mention `<2 x i8>`, but this also catches `<1 x i8>` and `<1 x i16>`. Also, I think some more tests for these cases would be good to have.

llvm/test/CodeGen/AIE/aie2p/v2i8-pack.ll
```diff
@@ -101,3 +101,222 @@ entry:
...
+
+define <2 x i8> @bitcast_i16_to_v2i8(i16 %v) {
+; AIE2P-LABEL: bitcast_i16_to_v2i8:
+; AIE2P:       // %bb.0: // %entry
+; AIE2P-NEXT:    mova r0, #-8; ret lr; nopm
+; AIE2P-NEXT:    lshl r0, r1, r0 // Delay Slot 5
+; AIE2P-NEXT:    extend.u8 r1, r1 // Delay Slot 4
+; AIE2P-NEXT:    mova r2, #8; extend.u8 r0, r0 // Delay Slot 3
+; AIE2P-NEXT:    lshl r0, r0, r2 // Delay Slot 2
+; AIE2P-NEXT:    or r0, r1, r0 // Delay Slot 1
```
mludevid, line 113: Q: Why isn't this a simple move?

llvm/test/CodeGen/AIE/aie2p/v2i8-pack.ll
```diff
@@ -101,3 +101,222 @@ entry:
   store  i16 %r, ptr %p, align 2
   ret void
 }
+
+define <2 x i8> @bitcast_i16_to_v2i8(i16 %v) {
```
mludevid, line 105: Could you pre-commit these tests as well?

llvm/lib/Target/AIE/aie2/AIE2LegalizerInfo.cpp
```diff
@@ -470,7 +477,7 @@ AIE2LegalizerInfo::AIE2LegalizerInfo(const AIE2Subtarget &ST) : AIEHelper(ST) {
   getActionDefinitionsBuilder(G_BITCAST)
       .legalIf(
           LegalityPredicates::all(isLegalBitCastType(0), isLegalBitCastType(1)))
-      .customIf(typeInSet(0, {V2S8, S16}));
+      .legalForCartesianProduct({S16, V2S8});
   ;
```
mludevid: Nit: Stray `;`

llvm/lib/Target/AIE/aie2/AIE2LegalizerInfo.cpp
```diff
@@ -470,7 +477,7 @@ AIE2LegalizerInfo::AIE2LegalizerInfo(const AIE2Subtarget &ST) : AIEHelper(ST) {
   getActionDefinitionsBuilder(G_BITCAST)
       .legalIf(
           LegalityPredicates::all(isLegalBitCastType(0), isLegalBitCastType(1)))
-      .customIf(typeInSet(0, {V2S8, S16}));
+      .legalForCartesianProduct({S16, V2S8});
```
mludevid: Is a bitcast from `s16` to `s16` legal?

llvm/lib/Target/AIE/AIELegalizerHelper.cpp
```diff
@@ -215,6 +215,21 @@ bool AIELegalizerHelper::legalizeG_BUILD_VECTOR(LegalizerHelper &Helper,
 
   assert((EltSize == 8 || EltSize == 16 || EltSize == 32 || EltSize == 64) &&
          "non-existent integer size");
+  // A 16-bit vector lives packed in the low half of a scalar register, which
+  // it reaches through S16.
+  if (DstVecSize == 16) {
```
mludevid: Nit: I think this belongs in `packVector`.

llvm/test/CodeGen/AIE/aie2p/v2i8-pack.ll
khallouh: The bitcast is free. The cost comes from the 2x8 return convention and the existing combines don't recognize this pattern. We need a combine for this pattern below. Note: This is not a regression (base already has 5 instructions). I'll pre-commit the new tests as suggested to show this.
```
%0:_(<2 x s8>)  = G_BITCAST %1(s16)                        <- free
%4:_(s8), %5:_(s8) = G_UNMERGE_VALUES %0(<2 x s8>)         <- return lowering scalarises
%6:_(s8) = G_IMPLICIT_DEF
%3:_(<4 x s8>) = G_BUILD_VECTOR %4, %5, %6, %6             <- rebuilt for the return register
```

llvm/lib/Target/AIE/aie2/AIE2LegalizerInfo.cpp
khallouh: Done

llvm/lib/Target/AIE/aie2/AIE2LegalizerInfo.cpp
khallouh: s16 to s16 harmless in practice as verifier rejects it anyway, but I removed the cartesian product and made the types explicit.

llvm/lib/Target/AIE/AIELegalizerHelper.cpp
khallouh: Done

llvm/test/CodeGen/AIE/aie2p/v2i8-pack.ll
khallouh: I pre-committed the new tests except for build_v2i8_from_scalars  and insert_then_bitcast  which fail at the base

llvm/lib/Target/AIE/AIELegalizerHelper.cpp
mludevid: Nit: Fix comment.

llvm/lib/Target/AIE/AIELegalizerHelper.cpp
```diff
@@ -1834,43 +1868,24 @@ bool AIELegalizerHelper::legalizeG_CONCAT_VECTORS(LegalizerHelper &Helper,
...
-  }
+  const auto [DstReg, DstTy, SrcReg, SrcTy] = MI.getFirst2RegLLTs();
+  assert(DstTy == V2S16 && SrcTy == V2S8 &&
+         "Expected to legalize an extension of <2 x s8>");
+
+  // <2 x s8> keeps its two elements packed in the low half of a GPR, so
+  // widening the elements has to move the high byte up to bit 16.
+  auto Packed =
+      MIRBuilder.buildAnyExt(S32, MIRBuilder.buildBitcast(S16, SrcReg));
+  auto HighElt = MIRBuilder.buildAnd(
```
mludevid: For `G_ANYEXT` we don't need to clear the upper lane bits, so the and instructions are not needed.

llvm/test/CodeGen/AIE/GlobalISel/legalize-extract-vector-elt.mir
```diff
@@ -905,3 +871,31 @@ body: |
     %6:_(s32) = G_EXTRACT_VECTOR_ELT %4(<4 x s32>), %5(s32)
     $r0 = COPY %6(s32)
 ...
+---
```
mludevid: Please keep the test in the same place for a clean diff.

llvm/lib/Target/AIE/aie2p/AIE2PISelLowering.cpp
mludevid: This also accepts `v4i1`, which we don't support in the legalization logic. Maybe make the check more restrictive to still scalarize that case.

llvm/lib/Target/AIE/AIELegalizerHelper.cpp
```diff
@@ -215,12 +219,14 @@ bool AIELegalizerHelper::legalizeG_BUILD_VECTOR(LegalizerHelper &Helper,
 
   assert((EltSize == 8 || EltSize == 16 || EltSize == 32 || EltSize == 64) &&
          "non-existent integer size");
-  assert(DstVecSize >= 32 && DstVecSize <= 1024 &&
+
+  assert(DstVecSize >= 16 && DstVecSize <= 1024 &&
```
mludevid, line 223: Nit: `DstVecSize <= 1024` is already covered by the next assert.

mludevid (approved):
LGTM.

PR #1283, opened 2026-09-10
[AIE][NFC] Hook G_PTR_ADD offsets into the GPR file from the base RBI

niwinanto:
Give targets a shouldSelectPtrAddAsAddmNc override and consult it from requiresGPRRegBank so a G_TRUNC / G_PTRTOINT offset can stay in GPRs before the pointer add itself is mapped.


andcarminati (approved):
LGTM.

PR #1284, opened 2026-09-10
[AIE][GlobalISel] Fix return pre-lowering for multiple returns

konstantinschwarz:
Return CC assignments depend only on the return type, so they are shared by
every return of a function. Drop the vestigial RetVal from SavedRetCCState and
the RetVal==Val identity assert in lowerReturnVal, which wrongly assumed a
single return and crashed on functions with multiple differing returns.


andcarminati (approved):
LGTM. Impressive how long we survived with the current behavior.

PR #1285, opened 2026-09-10
[AIE] Implement mayBeEmittedAsTailCall

konstantinschwarz:
Return CI->isTailCall() so calls already marked tail become tail-call candidates. This lets CodeGenPrepare duplicate returns into tail-call blocks; add a CodeGenPrepare test and update the affected codegen checks.


andcarminati (approved):
LGTM.

PR #1287, opened 2026-09-10
[AIE] Set jumpIsExpensive to avoid splitting short-circuit conditions

konstantinschwarz:
Enable setJumpIsExpensive(true) so that short-circuit conditions such as 'a && b' are combined into a single guard instead of being split into multiple conditional branches, reducing the branch count. Add a dedicated test and update affected in-tree tests.


andcarminati (approved):
LGTM. Great to see that this option is helping.

PR #1288, opened 2026-09-10
[AIE] Add switch-to-OR-of-icmp lowering pass

konstantinschwarz:
Branches are expensive on AIE (mandatory delay slots), yet a switch whose several case values reach the same label lowers to one compare+branch per case cluster. Add a target IR pass, scheduled in addCodeGenPrepare for AIE2 and later, that rewrites small switches into chained OR-of-'icmp eq' conditions with a single conditional branch per destination.

The transform is gated by -aie-switch-or-chain (default on). The primary gate is -aie-switch-or-chain-max-dests (default 2), which caps the number of distinct case destinations: the OR-chain emits one conditional branch per destination, so it only beats the generic balanced bisection when there are few destinations. A secondary gate, -aie-switch-or-chain-max-cases (default 8), caps the total number of cases to keep the linear chain short. Adds a dedicated lit test.


llvm/lib/Target/AIE/AIESwitchLowering.cpp
```diff
@@ -0,0 +1,260 @@
...
+        IsLast ? DefaultBB
+               : BasicBlock::Create(Ctx, "sw.or.next", F, DefaultBB);
+
+    BranchInst *BI = Builder.CreateCondBr(ChainCond, Dest, FalseBB);
+
+    // Preserve profile info: the taken edge inherits this destination's summed
+    // case weight; the fall-through inherits the remaining destinations plus
+    // the default weight. Weights are saturated to the uint32 metadata range.
+    if (HasProf) {
+      uint64_t TrueW = DestWeight.lookup(Dest);
```
andcarminati: nit: const?

llvm/lib/Target/AIE/AIESwitchLowering.cpp
```diff
@@ -0,0 +1,260 @@
...
+  ++NumSwitchesLowered;
+  return true;
+}
+
+bool AIESwitchLowering::runOnFunction(Function &F) {
+  if (!EnableAIESwitchLowering)
+    return false;
+
+  // Collect candidate switches first; lowering mutates the CFG.
+  SmallVector<SwitchInst *, 4> Candidates;
```
andcarminati: We could have a function to collect the candidates to improve readability.

llvm/lib/Target/AIE/AIESwitchLowering.cpp
```diff
@@ -0,0 +1,260 @@
...
+  for (BasicBlock &BB : F) {
+    auto *SI = dyn_cast<SwitchInst>(BB.getTerminator());
+    if (!SI)
+      continue;
+    // Only handle reasonably small switches; larger ones are better left to the
+    // generic bisected compare-tree lowering (see MaxCasesForOrChain above for
+    // the linear-chain vs. bisection trade-off).
+    if (SI->getNumCases() == 0 || SI->getNumCases() > MaxCasesForOrChain)
```
andcarminati: is `SI->getNumCases() == 0` a feasible case?

llvm/lib/Target/AIE/AIESwitchLowering.cpp
```diff
@@ -0,0 +1,260 @@
...
+
+  // Rewire the PHIs of a successor \p Succ that was reached from the switch
+  // block \p BB so that the (single) incoming edge now originates from \p Pred.
+  // A switch may reference the same successor from several case values, which
+  // creates several predecessor edges (and therefore several incoming PHI
+  // entries) for BB. After this transform there is exactly one edge, so we keep
+  // one entry (retargeted to Pred) and drop the duplicates. The duplicated
+  // entries are required by the verifier to carry the same value, so dropping
+  // them preserves semantics.
+  auto RewirePHIs = [&](BasicBlock *Succ, BasicBlock *Pred) {
```
andcarminati: This is a candidate to be a separate function.

andcarminati (dismissed):
LGTM.

khallouh (approved):

abnikant (approved):

PR #1292, opened 2026-09-15
[AIE2][AIE2PS] Legalize loads and stores of 16-bit vectors

khallouh:


llvm/test/CodeGen/AIE/load-store-narrow-vector.ll
```diff
@@ -0,0 +1,149 @@
...
+; AIE2PS:       // %bb.0:
+; AIE2PS-NEXT:    lda.s16 r0, [p0, #0]; nopxm
+; AIE2PS-NEXT:    nop
+; AIE2PS-NEXT:    nop
+; AIE2PS-NEXT:    nop
+; AIE2PS-NEXT:    nop
+; AIE2PS-NEXT:    nop
+; AIE2PS-NEXT:    ret lr; mov r2, #-8
+; AIE2PS-NEXT:    lshl r2, r0, r2 // Delay Slot 5
+; AIE2PS-NEXT:    extend.u8 r0, r0 // Delay Slot 4
```
khallouh, line 37: We need a combine for this (see https://github.com/Xilinx/llvm-aie/pull/1281#discussion_r3969300820)

khallouh: This is not urgent as 2x8 is only used in a handful of non-focus kernels and even there it is not returned so we don't pay this cost of the return convention which we can't combine

llvm/test/CodeGen/AIE/load-store-narrow-vector.ll
```diff
@@ -0,0 +1,149 @@
+; NOTE: Assertions have been autogenerated by utils/update_llc_test_checks.py UTC_ARGS: --version 5
+;
+; This file is licensed under the Apache License v2.0 with LLVM Exceptions.
+; See https://llvm.org/LICENSE.txt for license information.
+; SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+;
+; (c) Copyright 2026 Advanced Micro Devices, Inc. or its affiliates
+; RUN: llc -O2 -mtriple=aie2 %s --verify-machineinstrs -o - | FileCheck %s --check-prefix=AIE2
+; RUN: llc -O2 -mtriple=aie2ps %s --verify-machineinstrs -o - | FileCheck %s --check-prefix=AIE2PS
```
mludevid, line 9: Nit: Why not share the check lines? They look identical to me.

llvm/lib/Target/AIE/aie2/AIE2LegalizerInfo.cpp
```diff
@@ -400,13 +400,12 @@ AIE2LegalizerInfo::AIE2LegalizerInfo(const AIE2Subtarget &ST) : AIEHelper(ST) {
...
             const LLT &Ty = Query.Types[0];
-            return Ty.isVector() &&
-                   (Ty.getSizeInBits() == 64 || Ty.getSizeInBits() == 32);
+            const unsigned Size = Ty.getSizeInBits();
+            return Ty.isVector() && (Size == 16 || Size == 32 || Size == 64);
           },
           [=](const LegalityQuery &Query) {
             const LLT Ty = Query.Types[0];
             const unsigned Size = Ty.getSizeInBits();
-            assert(Size % 32 == 0);
```
mludevid, line 409: Nit: Instead of removing the assert you could replace it by `assert(Size == 16 || Size == 32 || Size == 64)`. Same for AIE2PS.

mludevid (dismissed):
Looks good!

llvm/test/CodeGen/AIE/load-store-narrow-vector.ll
khallouh: Check lines differ in nops and registers, the are not identical

mludevid (approved):

PR #1296, opened 2026-09-15
[AIE2PS] Support 1024-bit fp32 vector fadd/fsub

konstantinschwarz:
Make 1024- and 2048-bit fp32 vector G_FADD/G_FSUB legal, map the 1024-bit case to the accumulator register bank (vadd.f/vsub.f run on CM/ACC1024 registers, not the generic vector bank), and add ISel patterns selecting VADD_f_vacc_half_fp/VSUB_f_vacc_half_fp. Adds legalize/regbankselect/ inst-select tests for both widths.


khallouh (approved):
LGTM

PR #1297, opened 2026-09-17
[AIEX][OLP] Fix poison from elided last-iter loads and assert surviving uses before loop deletion

khallouh:
The outer-loop pipeliner peels a last iteration and then deletes the original loop. `populateLastIterBottom` dropped **every** latch load on the grounds that a prefetch for an iteration that never runs is dead. That is only true for prefetches consumed through a header PHI.
When something that **outlives the peel** still reads the loaded value, the clone of that reader keeps pointing at the original load. `DeleteDeadBlocks` then rewrites the leftover use as **poison**. That is a silent miscompile, not a crash.
Two shapes hit this:
1. A latch instruction that uses the load and is cloned into the peeled bottom.
2. A live-out use through the exit PHI (the remap goes through `cloneOf`, which falls back to the original value).

**Fix:** elide a latch load only when no user outlives the peel. Prefetches consumed through a header PHI stay elided, so pipelined loops are unchanged.

**Guard against other flavors:** `removeFromCFG` now asserts, before deletion, that no surviving instruction still uses an original-loop value. A clone-map lookup cannot catch this by itself (out-of-loop values map to themselves; the maps fill incrementally). Walking the dead blocks once covers every remapping path into the peeled iteration, so the next dangling use fails at the violation instead of becoming poison.


andcarminati (approved):
LGTM.

PR #1299, opened 2026-09-18
[AIE2P] Fix bfp16ebs8/bfp16ebs16 scalar type definitions

### LINKED ISSUE #1232: `sizeof(bfp16ebs8) == 1`, not 9
In Peano `sizeof(bfp16ebs8) == 1`, in chess (supposedly, I don't have it), it is 9. What it affects in real scenarios is code like
https://github.com/Xilinx/mlir-aie/blob/f05e903b07efccb9752159cd4b5b85503bd9dffe/programming_examples/ml/block_datatypes/gemm_asymmetric_tile_buffering/config1/mm_bfp_mixed.cc#L56

Right now config1 kernel is only usable if allocated buffers are multiplied by 9 manually, and all pointer arithmetic on `bfp16ebs8*` is rewritten into byte arithmetic.

Repro:
```c++
#include <aie_api/aie.hpp>
#include <cstdint>
 
static_assert(sizeof(bfp16ebs8) == 1, "empty struct -> 1 byte pointer arithmetic");
 
static_assert(aie::block_vector<bfp16ebs8, 64>::memory_bytes() == 72, "64 bfp16ebs8 values really occupy 72 bytes");
 
extern "C" void f() {
    constexpr int N = 64;                       // number of bfp16ebs8 values
    alignas(64) uint8_t raw[N * 9];             // correct byte count
    aie::block_vector_output_buffer_stream<bfp16ebs8, 64> out((bfp16ebs8 *)raw);
    (void)out;
}
```

```
.venv/lib/python3.14/site-packages/llvm-aie/bin/clang++ \
  -O2 -std=c++20 -D__AIE_API_AIE_ADF_HPP__ \
  -I/src/ArtLite/.venv/lib/python3.14/site-packages/mlir_aie/include \
  --target=aie2p-none-unknown-elf -c repro.cc
```

clang version - nightly (f4a72c275b76da3bb773403133d2bc1e96cbd03b)

A quick side note: `config1` kernel works fine; after code cleanup (unspill, conv_even) it reaches 24.5 TFLOPS (with Peano), even more than 24.3 (with chess).

konstantinschwarz:
Correct the scalar bfp16ebs8 and bfp16ebs16 struct definitions to match their type specification and add a front-end test asserting their sizeof and alignment.

Fixes #1232


AngryLoki:
@konstantinschwarz , should https://github.com/Xilinx/aie_api/blob/main/include/aie_api/types.hpp#L49-L50 be updated as well (maybe bumping aie_api submodule here as well)?

konstantinschwarz:
hi @AngryLoki, 
update with what? Those definitions are guarded with 
```
#define __AIE_API_SCALAR_BFP_TYPES__          (__AIE_MODEL_VERSION__ >= 11000)
```

For AIE2p, we define
```
#define __AIE_MODEL_VERSION__ 11500
```
, i.e. we skip those definitions
```
#if (__AIE_ARCH__ == 21 && __AIE_API_SCALAR_BFP_TYPES__ == 0) || __AIE_ARCH__ != 21
struct bfp16ebs8  {};
struct bfp16ebs16 {};
#endif
```
, and use the ones provided by the compiler.

SagarMaheshwari99 (approved):

PR #1300, opened 2026-09-21
[AIE] Canonicalize constant offsets out of GEP index expressions

AngryLoki:
Memory operations whose address is computed as `gep(base, add(iv, C))` defeat the GlobalISel post-increment/offset load-store combiners, which only reason about direct pointer chains. This happens whenever the inliner or loop unrolling rewrites a strided pointer walk into index arithmetic: the constant offset ends up hidden inside the scalar index expression (often as `or disjoint` after InstCombine), and every memory operation then materializes its own index register, costing modifier register setup and program memory without changing the steady-state schedule.

Add<sup>[1]</sup> a late IR pass (running right before instruction selection) that rewrites each memory operation's address to
```
  gep(gep(base, {non-constant indices}), {merged constant offset})
```
by splitting plain adds, constant subtractions and `or disjoint` into terms + constant. Two shapes are left untouched on purpose: chains that already are a one-step-at-a-time pointer walk (merging them would force pointer copies and hurt pointer register pressure), and addresses already in canonical form.

Example: an unrolled 16-store zeroing loop<sup>[2]</sup> went from 38 bundles with an or-add offset table in the prologue (dj0-dj6 + r1-r6 materialization) to 35 bundles with a pure post-increment `[p1], #64` chain, identical to the manually unrolled source.

[1] Out of multiple similar passes, the closest existing one is AIEOutlineMemoryGEP (same pipeline slot, same iteration pattern over memory instructions), but it is focused on slightly different thing

[2] As encountered while checking asm produced by `aie::unroll_*` family vs manually unrolled
```c++
    for (; out < out_end;) {  
        aie::unroll_times<16>([&]() __aie_inline {  
            aie::store_v(out, zeros);
            out += 32;  
        });  
    }
```


llvm/test/CodeGen/AIE/aie2/end-to-end/Conv2D-red.ll
```diff
@@ -118,28 +118,24 @@ define dso_local void @conv2d.loop.nest(ptr %add.ptr6.i51, ptr %add.ptr5, ptr %c
...
-; ASM-NEXT:    padda.3d [p0], d1; mov m2, r15 // Delay Slot 2
-; ASM-NEXT:    padda.3d [p1], d2; mov r26, dc5 // Delay Slot 1
+; ASM-NEXT:    nopb ; nopa ; vst.srs.s16.s32 bmh7, s2, [p3, #32]; nopx ; mov p4, p3; nopv
+; ASM-NEXT:    nopa ; nopb ; nopx ; mov dc5, r26; vst.srs.s16.s32 bml7, s3, [p4], #64
+; ASM-NEXT:    vst.srs.s16.s32 bmh6, s3, [p4, #32]; mov p5, p3
+; ASM-NEXT:    vst.srs.s16.s32 bml6, s3, [p4], m4; paddb [p5], m4; mov m1, r30
+; ASM-NEXT:    vst.srs.s16.s32 bmh5, s3, [p5, #96]; mov p6, p5
+; ASM-NEXT:    padda [p3], m1; paddb [p6], #128; vst.srs.s16.s32 bml5, s3, [p5, #64]; mov dn5, r27
+; ASM-NEXT:    vst.srs.s16.s32 bmh4, s3, [p6, #32]; paddb [p3], m4; mov dj5, r28
+; ASM-NEXT:    padda [p4], m1; paddb [p3], #192; vst.srs.s16.s32 bml4, s3, [p6, #0]; mov p6, p3
```
andcarminati: Please note that we are increasing the number of pointer copies and standalone pointers adds, what can increase the pointer register pressure and also slot pressure (padd).

AngryLoki: Updated; now pass triggers only on genuinely-mixed indices or pure address-math links that neither a load nor a store uses; not affecting Conv2D-* kernels anymore.

PR #1302, opened 2026-09-22
[AIEX] Pair leaf CSR spills as L when stack spill is unavoidable

abnikant:
> Even/odd CSR GPRs form an L register, and there is no L-to-L move, so leaves with free scratches still copy each half into unused GPRs.
> Pair as one L store only when a stack spill is unavoidable: the function has calls, or a leaf does not have two unused scratch GPRs left.


khallouh (approved):
LGTM

PR #1303, opened 2026-09-22
[AIE] Prefer WAR-conflict-free slots for multi-slot pseudo instructions

andcarminati:
In the post-RA bottom-up scheduler, WAR (Anti) dependencies cause the Write (W) instruction to be scheduled before the Read (R) instruction. When W is a multi-slot pseudo, the hazard recognizer previously selected its alternate opcode without considering R's slot. If W was assigned the same slot as R, they could not be co-issued in the same VLIW bundle, forcing an extra cycle.

This change adds a slot-avoidance heuristic in AIEHazardRecognizer: before selecting an alternate opcode for a multi-slot pseudo, we compute a set of slots to avoid (AvoidSlots) based on the slots already assigned to unscheduled Anti predecessors whose slot is "committed" — i.e., all their data (RAW) successors have already been scheduled. The alternate opcode selection then prefers slots outside AvoidSlots, falling back to the unconstrained selection if no preferred slot is hazard-free.


llvm/lib/Target/AIE/AIEHazardRecognizer.cpp
```diff
@@ -388,19 +446,35 @@ AIEHazardRecognizer::getHazardType(SUnit *SU, int DeltaCycles) {
...
+    const SlotBits AvoidSlots =
+        WARSlotAvoidance ? computeWARAvoidSlots(SU, SelectedAltDescs, Formats)
+                         : SlotBits(0);
+
+    // Two passes over the alternates:
+    //   1. Prefer slots not in AvoidSlots (WAR-aware, most restrictive).
+    //   2. Accept any slot (original fallback, no restriction).
+    // When AvoidSlots == 0 both passes are identical; the first pass will
+    // always find a winner (if one exists) and we return before the second.
+    for (const SlotBits Mask : {AvoidSlots, SlotBits(0)}) {
```
andcarminati: Maybe give a name to `SlotBits(0)`

PR #1304, opened 2026-09-23
[AIE] Evaluate the versioning trip count in the guard's own type

mludevid:
The inner-loop versioning guard compares trip counts in i32, but we built the trip count with SCEV's default evaluation type, which is always one bit wider than the exit count. Whenever the exit count's range covers all-ones the +1 genuinely needs that extra bit, so the result was an i33 the guard could not hold and the loop was rejected.

That range is exactly what the rotated shape clang emits for for (int i = 0; i < n / 64; ++i) produces (exit count (n /u 64) - 1).

Evaluate the +1 in i32 directly. This also drops a dead i33 -> i32 trunc round-trip from the guard in the common case.


khallouh:
Can you update optimization guide with the versioning hint spec?

mludevid:
> Can you update optimization guide with the versioning hint spec?

Done.

llvm/test/CodeGen/AIE/aie2/inner-loop-versioning.ll
```diff
@@ -125,9 +124,61 @@ exit:
   ret void
 }
 
+; The rotated shape clang produces for "for (int i = 0; i < n / 64; ++i)": an
+; exit count of (n /u 64) - 1, whose range wraps through all-ones. Evaluating
```
khallouh: What's so special about 64? So does a loop with no-division at all. Its range also wraps through all-ones. This is kind of a red herring since before this PR even the simplest cases were being rejected by the versioner, such as:
```
    AIE_LOOP_HINT(aie-loop-versioning, 1)
    for (int i=0 ; i<N; i++){
        a = *p + i;
    }
```

khallouh: I think the actual problem was that trip count can be zero so SCEV infers the range starting at -1 for which the previous `getUnsignedRangeMax` yielded 0xFFFFFFFF and adding 1 to that doesn't fit in 32 bits.

llvm/lib/Target/AIE/AIEInnerLoopVersioning.cpp
```diff
@@ -154,16 +155,25 @@ class AIELoopVersioner {
...
-  /// Whether the trip count can be computed and held by the runtime guard.
-  /// Read-only: decided solely from the trip-count SCEV.
-  bool hasVersionableTripCount() const;
+  /// A trip count the runtime guard can hold, or the reason it cannot.
+  struct GuardTripCount {
+    /// The trip count in the i32 the guard compares in, null when unusable.
+    const SCEV *TC;
+    /// Remark name and user-facing reason, set only when TC is null.
+    StringRef RemarkName;
+    StringRef Reason;
```
khallouh: Please add these optimization remarks to the test
and nit: RejectionReason instead of just Reason

PERF_OPT_GUIDE.md
```diff
@@ -666,7 +666,30 @@ always executes at least N times, set `AIE_LOOP_RANGE(N,)` and avoid
 versioning entirely. Loop versioning should only be used when the trip
 count genuinely varies below the pipelining threshold at runtime.
 
-### The solution: VERSIONED_LOOP
+### Automatic versioning: the aie-loop-versioning hint
```
khallouh: Please document the gaps too such as the i64 induction variable case which can't be versioned with the new type check requiring i32 even if the range fits in 32 bits

khallouh: Same for the UINT32_MAX trip count case where we fall back to the cold copy after wrapping in the guard.

khallouh: And also the constant UINT32_MAX trip count case where we reject versioning altogether because the high copy would be otherwise cold i.e. :
`do { ... } while (++i != 0) `

llvm/lib/Target/AIE/AIEInnerLoopVersioning.cpp
```diff
@@ -286,27 +296,44 @@ bool AIEInnerLoopVersioning::runOnFunction(Function &F) {
...
 
-  const SCEV *TripCount = SE.getTripCountFromExitCount(BEC);
-  return SE.getUnsignedRangeMax(TripCount).getActiveBits() <= 32;
+  // The runtime guard compares the trip count in i32, and the pipelined
+  // high-trip-count copy can only become a zero-overhead loop if its trip count
+  // fits in a 32-bit register (see AIETTICommon::isHardwareLoopProfitable).
+  // Gate on the exit count's type, not its range: truncating an exit count that
+  // can exceed UINT32_MAX aliases a huge trip count onto a large i32 that
+  // passes the guard, running the high copy with a wrong ZOL count. Relaxing
+  // this to admit wide exit counts of narrow range must keep that range check.
```
khallouh: Nit: Put the future relaxation part of the comment in a separate line with a FIXME mention so that it is easily identifiable if the case comes up at some point in the benchmarks.

llvm/test/CodeGen/AIE/aie2/inner-loop-versioning-remarks.ll
```diff
@@ -30,8 +30,8 @@ exit:
   ret void
 }
 
-; CHECK: loop not versioned because its trip count is unknown or does not fit 32 bits
-define void @unprofitable(ptr noalias %a, ptr noalias %b, i64 %n) {
+; CHECK: loop not versioned because its trip count does not fit 32 bits
+define void @wide_trip_count(ptr noalias %a, ptr noalias %b, i64 %n) {
```
khallouh: The i64 cases with small ranges used to fit and be versioned and are not anymore. But we could not have ZOL out of them because of the truncation. Can you add a test for that case? ideally in a pre-commit to highlight the behavior narrowing in this patch

llvm/lib/Target/AIE/AIEInnerLoopVersioning.cpp
```diff
@@ -286,27 +296,44 @@ bool AIEInnerLoopVersioning::runOnFunction(Function &F) {
...
+  if (SE.getTypeSizeInBits(BEC->getType()) > 32)
+    return {nullptr, "WideTripCount", "its trip count does not fit 32 bits"};
+
+  // Evaluate the +1 in i32. SCEV's default evaluation type is always one bit
+  // wider than the exit count, and when the exit count's range covers all-ones
+  // (as the rotated `i < n` count of n - 1 does) the i33 result really needs
+  // that top bit, so the guard could not hold it. Wrapping in i32 instead costs
+  // only the single trip count of 2^32, which the guard routes to the fallback.
+  const SCEV *TC = SE.getTripCountFromExitCount(
+      BEC, Type::getInt32Ty(L.getHeader()->getContext()), &L);
```
khallouh: Can you mention in the commit message that u are now using this version of `getTripCountFromExitCount` taking the loop instead of the one-argument version earlier? This is a crucial detail of the fix, not only the i32 evaluation, earlier we were rejecting rotated loops because "the trip count might be zero" but that was not true because the loops were guarded and the previously used SCEV helper was taking a nullptr for the loop

llvm/lib/Target/AIE/AIEInnerLoopVersioning.cpp
khallouh: I like this btw! To mention the reason in the remarks

llvm/lib/Target/AIE/AIEInnerLoopVersioning.cpp
khallouh: Can you also mention this in the PR description? since it sounds like division by 64 is some kind of corner case which triggers the bug whereas we were almost always rejecting versioning because of the ignored guard

llvm/test/CodeGen/AIE/aie2/inner-loop-versioning.ll
mludevid: Done.

llvm/lib/Target/AIE/AIEInnerLoopVersioning.cpp
mludevid: Done.

PERF_OPT_GUIDE.md
mludevid: Done. I've added a limitations section in the documentation.

llvm/lib/Target/AIE/AIEInnerLoopVersioning.cpp
mludevid: Done.

llvm/test/CodeGen/AIE/aie2/inner-loop-versioning-remarks.ll
mludevid: Done. I've pre-commited a test to show the theoretical regression. But as you already mention, this can't really happen in practice due to the 32-bit limitation we have on ZOL anyways.

llvm/lib/Target/AIE/AIEInnerLoopVersioning.cpp
mludevid: Done.

llvm/test/CodeGen/AIE/aie2/inner-loop-versioning-remarks.ll
khallouh: If we have to use ZOL then yes, but in theory we can benefit from pipelining without ZOL

khallouh (approved):
LGTM. Thanks for addressing my comments.

